PR #237, opened 2018-08-03
Update Apache license to include Graph Protocol copyright specifics

fordN:
Boilerplate copyright specification was still hidden in the bottom of the Apache license. It has been updated to include the year and copyright owner.

### DIFF
--- a/LICENSE-APACHE
+++ b/LICENSE-APACHE
@@ -175,17 +175,6 @@ TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
 
 END OF TERMS AND CONDITIONS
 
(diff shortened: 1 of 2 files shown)


Zerim:
I think we should actually remove the boiler plate altogether from the license (since they are instructions on how to add the license). We already include the copyright year and the license info in our `README.md`.

May also want to add this block to our `REAMDE.md`:
```
Unless required by applicable law or agreed to in writing, software
distributed under the License is distributed on an "AS IS" BASIS,
WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
See the License for the specific language governing permissions and
limitations under the License.
```

PR #240, opened 2018-08-03
docs: Minor getting started improvements

Zerim:
- Consolidates build and compile steps in getting started
- Adds a note about `Store.set` and `Entity.unset` API to indicate that attributes not explicitly set (or unset) will not be overwritten.


html5cat:
Looks good! Fixed some of the links in ToC

PR #242, opened 2018-08-03
*: move from `tokio-core` to `tokio`

### LINKED ISSUE #93: Migrate from tokio-core to tokio only

### LINKED ISSUE #218: Store updates are only made after all events have been processed
The store updates should be done shortly after the event is processed by the mapping, if I understand correctly what @Jannis reported. I need to reproduce this and then figure out what we're doing wrong with our tokio streams.

leodasvacas:
Resolves #93
Resolves #218

Issue #218 was caused by our tokio reactor somehow running on a single thread, so when the `RuntimeHost` was processing events it would block all other tasks.

Now things are more async, which required making a bunch of things `Send` and `Sync`. This was done mostly by just adding the bounds, but the subgraph resolver needed a deeper refactoring, which was good because we're now using `failure` for it's errors.

The tricky part is the `WasmiModule` because it's not `Send`, so we spawn a dedicated OS thread to handle each `RuntimeHost` and then may call `wait()` inside host exported functions without causing a system-wide slowdown.


.buildkite/run-tests.sh
```diff
@@ -29,4 +29,4 @@ set_up
 trap tear_down EXIT
 
 # Run the test suite
-cargo test -- --test-threads=1
+cargo test
```
fordN: yay, parallel tests

store/postgres/tests/store.rs
```diff
@@ -28,15 +29,24 @@ fn postgres_test_url() -> String {
...
 where
-    T: FnOnce() -> () + panic::UnwindSafe,
+    T: FnOnce() -> () + Send + 'static + panic::UnwindSafe,
 {
-    insert_test_data();
-    let result = panic::catch_unwind(|| test());
-    remove_test_data();
-    result.expect("Failed to run test");
+    let _test_lock = TEST_MUTEX.lock().unwrap();
+    tokio::run(future::lazy(|| {
```
fordN: I am noticing tests pass that should be failing.  Example: If I have not set `THEGRAPH_STORE_POSTGRES_DIESEL_URL` all tests should fail, but are currently passing.

core/src/query/runner.rs
```diff
@@ -39,7 +36,7 @@ where
         let logger = self.logger.clone();
         let store = self.store.clone();
 
-        self.runtime.spawn(stream.for_each(move |query| {
+        tokio::spawn(stream.for_each(move |query| {
```
fordN: Now that we are no longer passing the `reactor::Handle` are we creating more threads than before?

store/postgres/tests/store.rs
fordN: I also tested the changes by using a valid `THEGRAPH_STORE_POSTGRES_DIESEL_URL`, but changing an assertion to force a test failure. Again all tests passed.  In both cases the process did log a panic, but the test did not fail.

core/src/query/runner.rs
leodasvacas: I don't think we're creating more, but we're certainly using more threads. The `spawn` we were calling before apparently always spawned on the same thread as the event loop, which was the main thread.

leodasvacas:
@fordN Thanks for the careful review! If a tokio worker thread panics it will only log the panic and drop the task, so I've changed the store tests to always panic on the main thread so that the test fails properly.

store/postgres/tests/store.rs
leodasvacas: Great catch!

fordN:
Looks good! 👍  `Graph-store-postgres` tests are now behaving as expected.

PR #254, opened 2018-08-09
Add complete set of supported filters to graphql schema, squash several filter query building bugs

### LINKED ISSUE #222: Query interface not showing all where filters
Not all filters show up in the GraphiQL docs on the right sidebar, in addition the queries that are not shown trigger inline errors in the query interface though they do work when run.

### LINKED ISSUE #252: Using `ends_with` or `not_ends_with` filters causes panic
Panicked at: "internal error: entered unreachable code"

fordN:
Resolves #222 and Resolves #252 

The combined schema that was being generated only included a subset of the supported filter fields,`equals` and `not equals`, which was manifesting in inconsistent user experience; filters could be used in a query, but did not show up in the schema docs, and triggered inline errors in the query interface. During research and testing several other issues with the filter queries were revealed. 

Summary of updates made:
* Add complete set of where filters to schema for scalar fields. Each scalar type receives a specific set of filter fields
* Update tests to check for full list of filters
* Fix bugs in building filtered queries:
    * `postgres/src/filter` -  Filter match expression within EndsWith arm was trying to match against StartsWith
    * `graphql/src/schema/ast` - match arms in `parse_field_as_filter()` were not in correct order making several arms unreachable
    * `graph/src/data/store/mod` - `from_query_value()` was missing an arm for List filters on a NamedType schema

### DIFF
--- a/graph/src/data/store/mod.rs
+++ b/graph/src/data/store/mod.rs
@@ -1,7 +1,12 @@
 use graphql_parser::query;
 use graphql_parser::schema;
 
+use bigdecimal::BigDecimal;
+use diesel::pg::Pg;
+use diesel::serialize::{self, Output, ToSql};
+use diesel::sql_types::{Bool, Float, Integer, Numeric, Text};
 use std::collections::{BTreeMap, HashMap};
+use std::io::Write;
 use std::iter::FromIterator;
 use std::ops::{Deref, DerefMut};
 use std::str::FromStr;
@@ -17,7 +22,7 @@ pub const BYTES_SCALAR: &str = "Bytes";
 pub const BIG_INT_SCALAR: &str = "BigInt";
 
 /// An attribute value is represented as an enum with variants for all supported value types.
-#[derive(Clone, Debug, Deserialize, Serialize, PartialEq)]
+#[derive(Clone, Debug, Deserialize, Serialize, PartialEq, AsExpression)]
 #[serde(untagged)]
 pub enum Value {
     String(String),
@@ -111,6 +116,55 @@ impl From<String> for Value {
 impl<'a> From<&'a String> for Value {
     fn from(value: &'a String) -> Value {
         Value::String(value.clone())
+    }
+}
+
+impl ToSql<Bool, Pg> for Value {
+    fn to_sql<W: Write>(&self, out: &mut Output<W, Pg>) -> serialize::Result {
+        match *self {
+            Value::Bool(ref b) => <bool as ToSql<Bool, Pg>>::to_sql(&b, out),
+            _ => unimplemented!(),
+        }
+    }
+}
+
+impl ToSql<Float, Pg> for Value {
+    fn to_sql<W: Write>(&self, out: &mut Output<W, Pg>) -> serialize::Result {
+        match *self {
+            Value::Float(ref f) => <f32 as ToSql<Float, Pg>>::to_sql(&f, out),
+            _ => unimplemented!(),
+        }
+    }
+}
+
+impl ToSql<Integer, Pg> for Value {
+    fn to_sql<W: Write>(&self, out: &mut Output<W, Pg>) -> serialize::Result {
+        match *self {
+            Value::Int(ref i) => <i32 as ToSql<Integer, Pg>>::to_sql(&i, out),
+            _ => unimplemented!(),
+        }
+    }
+}
+
+impl ToSql<Numeric, Pg> for Value {
+    fn to_sql<W: Write>(&self, out: &mut Output<W, Pg>) -> serialize::Result {
+        match *self {
+            Value::BigInt(ref number) => <BigDecimal as ToSql<Numeric, Pg>>::to_sql(
+                &BigDecimal::from_str(&number.to_string()).unwrap(),
+                out,
+            ),
+            _ => unimplemented!(),
+        }
+    }
+}
+
+impl ToSql<Text, Pg> for Value {
+    fn to_sql<W: Write>(&self, out: &mut Output<W, Pg>) -> serialize::Result {
+        match *self {
+            Value::String(ref s) => <String as ToSql<Text, Pg>>::to_sql(&s, out),
+            Value::Bytes(ref h) => <String as ToSql<Text, Pg>>::to_sql(&h.to_string(), out),
+            _ => unimplemented!(),
+        }
     }
 }
 
--- a/graph/src/data/store/mod.rs
+++ b/graph/src/data/store/mod.rs
@@ -45,6 +45,13 @@ impl Value {
                     .collect(),
             ),
 
+            (query::Value::List(values), NamedType(n)) => Value::List(
+                values
+                    .iter()
+                    .map(|value| Self::from_query_value(value, &NamedType(n.to_string())))
+                    .collect(),
+            ),
+
             (query::Value::String(s), NamedType(n)) => {
                 // Check if `ty` is a custom scalar type, otherwise assume it's
                 // just a string.
--- a/graph/src/data/store/mod.rs
+++ b/graph/src/data/store/mod.rs
@@ -1,10 +1,6 @@
 use graphql_parser::query;
 use graphql_parser::schema;
 
-use bigdecimal::BigDecimal;
-use diesel::pg::Pg;
-use diesel::serialize::{self, Output, ToSql};
-use diesel::sql_types::{Bool, Float, Integer, Numeric, Text};
 use std::collections::{BTreeMap, HashMap};
 use std::io::Write;
 use std::iter::FromIterator;
@@ -22,7 +18,7 @@ pub const BYTES_SCALAR: &str = "Bytes";
 pub const BIG_INT_SCALAR: &str = "BigInt";
 
 /// An attribute value is represented as an enum with variants for all supported value types.
-#[derive(Clone, Debug, Deserialize, Serialize, PartialEq, AsExpression)]
+#[derive(Clone, Debug, Deserialize, Serialize, PartialEq)]
 #[serde(untagged)]
 pub enum Value {
     String(String),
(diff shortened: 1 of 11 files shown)


store/postgres/src/filter.rs
```diff
@@ -317,8 +317,8 @@ fn store_filter_by_mode<'a>(
 
         StoreFilter::EndsWith(..) | StoreFilter::NotEndsWith(..) => {
             let (attribute, op, value) = match filter {
-                StoreFilter::StartsWith(attribute, value) => (attribute, " LIKE ", value),
-                StoreFilter::NotStartsWith(attribute, value) => (attribute, " NOT LIKE ", value),
+                StoreFilter::EndsWith(attribute, value) => (attribute, " LIKE ", value),
+                StoreFilter::NotEndsWith(attribute, value) => (attribute, " NOT LIKE ", value),
```
leodasvacas: Good catch, these new filters don't have unit tests, what do you think of taking the opportunity to add a test for an`EndsWith` query in `postgres/test/store.rs`?

graphql/src/schema/api.rs
```diff
@@ -212,19 +212,39 @@ fn field_scalar_filter_input_values(
...
+            "ends_with",
+            "not_ends_with",
+        ],
+        _ => vec!["", "not"],
+    }.into_iter()
+        .map(|filter_type| {
+            input_value(
+                &field.name,
+                filter_type,
+                Type::NamedType(field_type.name.to_owned()),
```
leodasvacas: It seems that for `in` and `not_in` the correct type here would be `[Type]` (a list of `Type`) rather than just `Type`, so that in the docs we get for example `deposit_in: [BigInt]` rather than `deposit_in: BigInt` which is what currently appears.

graphql/src/schema/api.rs
```diff
@@ -239,11 +239,13 @@ fn field_scalar_filter_input_values(
...
-            input_value(
-                &field.name,
-                filter_type,
-                Type::NamedType(field_type.name.to_owned()),
-            )
+            let value_type = match filter_type {
+                "in" | "not_in" => Type::ListType(Box::new(Type::NonNullType(Box::new(
+                    Type::NamedType(field_type.name.to_owned()),
+                )))),
+                _ => Type::NamedType(field_type.name.to_owned()),
```
leodasvacas: This code would look cleaner with `Type::NamedType(field_type.name.to_owned())` extracted to a variable.

graph/src/data/store/mod.rs
```diff
@@ -1,7 +1,12 @@
 use graphql_parser::query;
 use graphql_parser::schema;
 
+use bigdecimal::BigDecimal;
+use diesel::pg::Pg;
```
Jannis: I don't like seeing a dependency on Diesel in the `graph` crate. @leodasvacas @fordN I assume the only alternative would be to create `SqlValue(Value)` ad `SqlEntity(Entity)` newtypes in the `graph-store-postgres` crate and wrap entities in those before creating SQL statements?

graphql/src/schema/api.rs
```diff
@@ -212,19 +212,42 @@ fn field_scalar_filter_input_values(
...
+            "ends_with",
+            "not_ends_with",
+        ],
+        _ => vec!["", "not"],
+    }.into_iter()
+        .map(|filter_type| {
+            let field_type = Type::NamedType(field_type.name.to_owned());
+            let value_type = match filter_type {
+                "in" | "not_in" => {
+                    Type::ListType(Box::new(Type::NonNullType(Box::new(field_type))))
```
Jannis: I _think_ there needs to be another `Type::NonNullType` around the `Type::ListType` to disallow `null` being passed in as a value (anything from an empty list upwards is fine, but `null` makes no sense).

graphql/src/schema/ast.rs
```diff
@@ -25,14 +25,14 @@ pub(crate) fn parse_field_as_filter(key: &Name) -> (Name, FilterOp) {
         k if k.ends_with("_lt") => ("_lt", FilterOp::LessThan),
```
Jannis: This file doesn't seem to have changed other than reordering the match branches. Is there a reason for changing the order?

store/postgres/src/filter.rs
```diff
@@ -271,9 +271,72 @@ fn store_filter_by_mode<'a>(
         }
         // Is `attribute` equal to some `v` in `query_values`?
         StoreFilter::In(attribute, query_values) => {
-            query_values.into_iter().try_fold(query, |q, v| {
-                store_filter_by_mode(q, StoreFilter::Equal(attribute.clone(), v), FilterMode::Or)
-            })?
+            let op = " = ANY (";
+            match query_values[0].clone() {
```
Jannis: Is there anything that guarantees the list is not empty? Otherwise, this unconditional index access could panic.

store/postgres/src/filter.rs
```diff
@@ -271,9 +271,72 @@ fn store_filter_by_mode<'a>(
...
+                Value::Bool(_) => add_filter(
```
Jannis: Do we have a task to track query validation? Unless we validate the values in the query, we'll have to assume that `query_values` could be a list with values of different types and then the `bind::<Array<Bool>, _>(query_values)` will trigger `unimplemented!()` and panic.

Jannis: An easy check we could perform upfront is whether all values have the same type (can be checked using a simple reduction over all values).

Can we create a task to do this and allow filter generation to fail and return an error back to the client in case it does?

store/postgres/src/store.rs
```diff
@@ -236,6 +236,10 @@ impl BasicStore for Store {
                 .offset(range.skip as i64);
         }
 
+        {
+            debug!(self.logger, "find"; "Postgres query" => format!("{:?}", debug_query::<Pg, _>(&diesel_query)));
```
Jannis: This line is too long. `rustfmt` can't properly format the semicolon syntax. Do it manually:
```rust
debug!(self.logger, "find";
       "sql" => format("{:?}", debug_query::<Pg, _>(&diesel_query)));
```
Also note that we're using single-word labels for keys in the logged key/value pairs ("sql" instead of "Postgres query").

graph/src/data/store/mod.rs
leodasvacas: That's the only alternative I see. I think it's ok for `graph` to take the dependency, but ideally we'd re-export  diesel from `graph` and then `graph-store-postgres` would not depend directly on diesel, so we can control the diesel version and features from a single `Cargo.toml`. This is what was done for `slog` and `tokio`.

graph/src/data/store/mod.rs
```diff
@@ -114,6 +119,55 @@ impl<'a> From<&'a String> for Value {
     }
 }
 
+impl ToSql<Bool, Pg> for Value {
+    fn to_sql<W: Write>(&self, out: &mut Output<W, Pg>) -> serialize::Result {
+        match *self {
+            Value::Bool(ref b) => <bool as ToSql<Bool, Pg>>::to_sql(&b, out),
+            _ => unimplemented!(),
```
leodasvacas: `unimplemented` isn't the right description here, a custom panic message would be better.

graphql/src/schema/ast.rs
fordN: The order is important here because some of the arms' match strings are contained in others, so the previous order created essentially unreachable arms.  Example: if the .ends_with("_in") arm comes before .ends_with("_not_in") the "_not_in" arm will never be matched because "_not_in" ends_with "_in".         

For this reason I made sure:
-  "_not_starts_with" comes before "_starts_with"
-  "_not_ends_with" comes before "_ends_with" 
- "_in" comes before "_not_in"
-  "_contains" comes before "_not_contains".

Jannis: Oh, _of course_! Thanks for explaining, I should've seen that...

graph/src/data/store/mod.rs
Jannis: That's ok but we will very likely replace Postgres at some point and to avoid making that too hard, depending on it anywhere else but in `store/postgres` is a bad idea.

store/postgres/src/filter.rs
fordN: I created issue #272 to track validation of `query_values`.

store/postgres/src/filter.rs
fordN: This will also be resolved by issue #272.

store/postgres/src/filter.rs
```diff
@@ -271,9 +271,72 @@ fn store_filter_by_mode<'a>(
         }
         // Is `attribute` equal to some `v` in `query_values`?
         StoreFilter::In(attribute, query_values) => {
-            query_values.into_iter().try_fold(query, |q, v| {
-                store_filter_by_mode(q, StoreFilter::Equal(attribute.clone(), v), FilterMode::Or)
-            })?
+            let op = " = ANY (";
```
leodasvacas: Diesel query builder has support for this https://docs.diesel.rs/diesel/expression_methods/trait.ExpressionMethods.html#method.eq_any, I wonder if we can use it here.

store/postgres/src/models.rs
```diff
@@ -17,3 +23,61 @@ pub struct EntityTable {
...
+    pub fn new_array(values: Vec<Value>) -> Vec<Self> {
+        values.into_iter().map(|value| SqlValue(value)).collect()
+    }
+}
+
+impl ToSql<Bool, Pg> for SqlValue {
+    fn to_sql<W: Write>(&self, out: &mut Output<W, Pg>) -> serialize::Result {
+        match self.0 {
+            Value::Bool(ref b) => <bool as ToSql<Bool, Pg>>::to_sql(&b, out),
+            _ => unimplemented!(),
```
leodasvacas: `unimplemented` isn't the right description here, a custom panic message would be better.

store/postgres/src/filter.rs
fordN: I don't see a reasonable way through using diesel expression methods like `eq_any()` with our`jsonb` attributes right now.  

We could create a custom type that uses diesel-dynamic-schema to build the structs at runtime, but I think that would also require looking at the schema to see what attributes are on each entity.

store/postgres/src/models.rs
```diff
@@ -17,3 +23,61 @@ pub struct EntityTable {
...
+    pub fn new_array(values: Vec<Value>) -> Vec<Self> {
+        values.into_iter().map(|value| SqlValue(value)).collect()
+    }
+}
+
+impl ToSql<Bool, Pg> for SqlValue {
+    fn to_sql<W: Write>(&self, out: &mut Output<W, Pg>) -> serialize::Result {
+        match self.0 {
+            Value::Bool(ref b) => <bool as ToSql<Bool, Pg>>::to_sql(&b, out),
+            _ => panic!("Invalid input type, all items in array must be Bool"),
```
Jannis: `ToSql<Bool, Pg>` has nothing to do with arrays, so this message is confusing (same with the messages below). How about: "failed to convert non-boolean attribute value to boolean in SQL"?

store/postgres/src/store.rs
```diff
@@ -233,6 +233,11 @@ impl BasicStore for Store {
                 .offset(range.skip as i64);
         }
 
+        {
+            debug!(self.logger, "find";
+                "sql" => format!("{:?}", debug_query::<Pg, _>(&diesel_query)));
```
Jannis: The alignment here is off by a few spaces. I'm also not sure why we need a dedicated scope around the `debug!` statement.

store/postgres/src/store.rs
```diff
@@ -233,9 +233,10 @@ impl BasicStore for Store {
                 .offset(range.skip as i64);
         }
 
-        {
-            debug!(self.logger, "find"; "Postgres query" => format!("{:?}", debug_query::<Pg, _>(&diesel_query)));
-        }
+
+        debug!(self.logger, "find";
+            "sql" => format!("{:?}", debug_query::<Pg, _>(&diesel_query)));
```
fordN: I'm not clear on how the alignment is off here?  I'm using a single indent on the second line.

Jannis:
```
        debug!(self.logger, "find";
            "sql" => format!("{:?}", debug_query::<Pg, _>(&diesel_query)));
```
vs.
```
        debug!(self.logger, "find";
               "sql" => format!("{:?}", debug_query::<Pg, _>(&diesel_query)));
```
which one is aligned correctly?

fordN: nm. I see how it needs to be formatted.

PR #256, opened 2018-08-09
runtime: Handle all events, merge event streams

leodasvacas:
Fix regression reported by @Ford where the tokio upgrade caused us to only process one event, because waiting on one event stream permanently blocked all others.

This is done by merging all of the event streams and then waiting on the unified stream, allowing for receiving interleaved events which also improves performance.

### DIFF
--- a/runtime/wasm/src/host.rs
+++ b/runtime/wasm/src/host.rs
@@ -1,9 +1,8 @@
 use ethereum_types::Address;
-use futures::prelude::*;
 use futures::sync::mpsc::{channel, Receiver};
 use slog::Logger;
 use std::str::FromStr;
-use std::sync::{Arc, Mutex};
+use std::sync::Mutex;
 use std::thread;
 use uuid::Uuid;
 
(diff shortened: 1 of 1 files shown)


runtime/wasm/src/host.rs
```diff
@@ -107,7 +106,9 @@ impl RuntimeHost {
             // Load the mappings as a WASM module
             let module = WasmiModule::new(&logger, wasmi_config);
 
-            Self::subscribe_to_events(&logger, data_source, module, ethereum_adapter);
+            Self::subscribe_to_events(&logger, data_source, module, ethereum_adapter)
+                .wait()
```
Jannis: I find moving this wait + for_each out of subscribe_to_events odd and would prefer this to be part of the function.

leodasvacas: The bug was caused because of `wait` being called on multiple streams, now it's very visible that we wait on a single stream. So I prefer this form, even though this line doesn't read very well, I've added a comment to attempt to clarify what's happening, what do you think?

PR #257, opened 2018-08-09
Show debug logs in release builds

### LINKED ISSUE #238: Set maximum log level to `debug` in release builds
See https://docs.rs/slog/2.2.3/slog/#notable-details for how. Also we might want to re-export `slog` from `graph` so we can control this from a single `Cargo.toml`.

leodasvacas:
At least in this early stage we want all the logs we can get. First commit re-exports `slog` from `graph` so we can control this configuration from a single place.

Resolves #238

### DIFF
--- a/server/http/src/server.rs
+++ b/server/http/src/server.rs
@@ -1,22 +1,17 @@
-use futures::future;
-use futures::prelude::*;
 use futures::sync::mpsc::{channel, Receiver, Sender};
 use hyper;
 use hyper::Server;
 
-use slog;
 use std::error::Error;
 use std::fmt;
 use std::net::{Ipv4Addr, SocketAddrV4};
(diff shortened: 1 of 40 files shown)


server/http/src/server.rs
```diff
@@ -1,22 +1,17 @@
...
 use futures::sync::mpsc::{channel, Receiver, Sender};
 use hyper;
 use hyper::Server;
 
-use slog;
 use std::error::Error;
 use std::fmt;
 use std::net::{Ipv4Addr, SocketAddrV4};
-use std::sync::{Arc, Mutex};
+use std::sync::Mutex;
```
fordN: Why is `Arc` no longer needed here? included in prelude?

leodasvacas: exactly

PR #258, opened 2018-08-09
runtime: Remove block hash from `set`, `remove`, bind it transparently

### LINKED ISSUE #225: Block hash should be bound to store outside of mapping
We currently allow the user to set the `blockHash` for a store transaction *inside* of the mapping.

I believe this was a stop gap, because of engineering complexity injecting that context from the event stream into the WASM host external store functions.

This could lead to bad or broken behavior. The `blockHash` of a Store transaction (at least for our event-sourced data sources), should be determined by the block hash of the block/events being processed, not some arbitrary value set by the user which would allow them to write to the Store as of a different block.

leodasvacas:
This is straightforward since we never call two handlers in parallel for the same mapping.

Resolves #225
Companion PR graphprotocol/graph-cli#85


leodasvacas:
@Jannis afaict we never got to actually using them.

Jannis:
Great, ok, now that the issue exists to track the only remark I had, this should be good to go in.

PR #264, opened 2018-08-14
Export typeConversion.stringToH160(String): H160 function

Jannis:
This adds and exports the `stringToH160` type conversion function needed for graphprotocol/graph-cli#86 (and the corresponding PR: graphprotocol/graph-cli#87).


timmclean:
Only comment I would've had here is that h160ToString would seem like a natural addition? Otherwise, the implementation looks fine :+1:

Jannis:
@timmclean It would, if it wasn't for `bytesToString` and `bytesToHex`, which cover all byte arrays (H160, H256 etc.).

PR #265, opened 2018-08-14
docs: Add  collection filers, fix link

html5cat:


docs/graphql-api.md
```diff
@@ -88,6 +88,56 @@ query {
...
+{
+  applications(where: {deposit_gt:"10000000000"}) {
+    id
+    whitelisted
+    deposit
+  }
+}
+```
+
+Full list of parameter suffixes:
```
Jannis: I think we can do slightly better here. Some suffixes are only supported for specific types (e.g. `_contains` and `_not_contains`), so perhaps it would be good to use a support matrix here or something.

Here's the logic that creates filters for fields of the different types: https://github.com/graphprotocol/graph-node/blob/d4c721bcb166315081649fafb22eead946c89c52/graphql/src/schema/api.rs#L215

html5cat: For simplicity I'd rather not go with the matrix route, but definitely should include this info.. What if we add the link to the implementation for those who want to dig deeper?

Jannis: Ok. In that case I'd rather keep it simple (as you had it initially) and not point to a specific commit and file. The code can and will change, plus I don't think the code simple enough that we want to throw it at people right away. 😉

html5cat: I think it's pretty obvious what's happening there even without me knowing any Rust. Agree that pointing to part of code is "bad". Shall we just leave the link as is an see how it ages? 😄 Something tells me that part might not change *that* much

Jannis: Tricky... that commit isn't even part of master, so I wouldn't want to point to it at all.

html5cat: Agreed. Removed!

Jannis: Alright, let's get this in then 😄

PR #278, opened 2018-08-16
node: Test IPFS node connection on start up

### LINKED ISSUE #268: Improve IPFS unavailable error
Both of these errors have been seen
```
If IPFS is not running the node reports nothing.
```
```
If ipfs is not setup Error: connect ECONNREFUSED 127.0.0.1:5001
```
We should check to see if we can connect to IPFS on node startup and return a clear error if not.

We should ensure that error handling works correctly both when IPFS has been initialized and when it hasn't

timmclean:
Provide better error message to user if IPFS connection fails.

Resolves #268 

New error message looks like this:

```
Aug 16 15:46:47.190 INFO Connecting to IPFS node...
Aug 16 15:46:47.193 ERRO Failed to connect to IPFS: hyper client error 'an error occurred trying to connect: Connection refused (os error 111)'
Aug 16 15:46:47.193 ERRO Is there an IPFS node running at '127.0.0.1:5002'?
```

Normal operation:

```
Aug 16 15:47:29.025 INFO Connecting to IPFS node...
Aug 16 15:47:29.028 INFO Connected to IPFS node.
```

I included these two messages in case the client hangs trying to connect (I haven't observed that behaviour, but just in case).

### DIFF
--- a/node/src/main.rs
+++ b/node/src/main.rs
@@ -19,6 +19,8 @@ use std::env;
 use std::net::SocketAddr;
 use std::str::FromStr;
 use std::sync::Mutex;
+use std::thread;
+use std::time::Duration;
 
 use graph::components::forward;
 use graph::prelude::*;
@@ -108,7 +110,7 @@ fn async_main() -> impl Future<Item = (), Error = ()> + Send + 'static {
     let ethereum_ws = matches.value_of("ethereum-ws");
 
     let ipfs_socket_addr = SocketAddr::from_str(matches.value_of("ipfs").unwrap())
-        .expect("could not parse IPFS address, expected format is host:port");
+        .expect("could not parse IPFS address, expected format is ipaddress:port");
 
     debug!(logger, "Setting up Sentry");
 
@@ -128,13 +130,27 @@ fn async_main() -> impl Future<Item = (), Error = ()> + Send + 'static {
 
     info!(logger, "Starting up");
 
-    // Create system components
+    // Create and test IPFS client
+    info!(logger, "Connecting to IPFS node...");
     let resolver = Arc::new(
         IpfsClient::new(
             &format!("{}", ipfs_socket_addr.ip()),
             ipfs_socket_addr.port(),
         ).expect("Failed to start IPFS client"),
     );
+    let ipfs_test = resolver.cat("/ipfs/QmS4ustL54uo8FzR9455qaxZwuMiUhyvMcX9Ba8nUH4uVv/readme");
+    if let Err(e) = ipfs_test.concat2().wait() {
+        error!(logger, "Failed to connect to IPFS: {}", e);
+        error!(
+            logger,
+            "Is there an IPFS node running at '{}'?", ipfs_socket_addr
+        );
+        thread::sleep(Duration::from_millis(50)); // small delay to let logger finish writing
+        panic!();
+    }
+    info!(logger, "Connected to IPFS node.");
+
+    // Create subgraph provider
     let (mut subgraph_provider, subgraph_provider_events) = IpfsSubgraphProvider::new(
         logger.clone(),
         &format!("/ipfs/{}", subgraph_hash.clone()),


node/src/main.rs
```diff
@@ -128,13 +130,27 @@ fn async_main() -> impl Future<Item = (), Error = ()> + Send + 'static {
...
-    // Create system components
+    // Create and test IPFS client
+    info!(logger, "Connecting to IPFS node...");
     let resolver = Arc::new(
         IpfsClient::new(
             &format!("{}", ipfs_socket_addr.ip()),
             ipfs_socket_addr.port(),
         ).expect("Failed to start IPFS client"),
     );
+    let ipfs_test = resolver.cat("/ipfs/QmS4ustL54uo8FzR9455qaxZwuMiUhyvMcX9Ba8nUH4uVv/readme");
```
Jannis: We can't assume this file exists. We can't really assume any file exists.

node/src/main.rs
```diff
@@ -128,13 +130,27 @@ fn async_main() -> impl Future<Item = (), Error = ()> + Send + 'static {
...
+    if let Err(e) = ipfs_test.concat2().wait() {
+        error!(logger, "Failed to connect to IPFS: {}", e);
+        error!(
+            logger,
+            "Is there an IPFS node running at '{}'?", ipfs_socket_addr
+        );
+        thread::sleep(Duration::from_millis(50)); // small delay to let logger finish writing
```
Jannis: This is an arbitrary timeout that doesn't guarantee the message gets logged. Is there a way to flush the logger or to panic through the logger?

node/src/main.rs
```diff
@@ -108,7 +110,7 @@ fn async_main() -> impl Future<Item = (), Error = ()> + Send + 'static {
     let ethereum_ws = matches.value_of("ethereum-ws");
 
     let ipfs_socket_addr = SocketAddr::from_str(matches.value_of("ipfs").unwrap())
-        .expect("could not parse IPFS address, expected format is host:port");
+        .expect("could not parse IPFS address, expected format is ipaddress:port");
```
Jannis: I think host is fine here.

timmclean: A `SocketAddr` is an IP address + port, it doesn't do any DNS lookups, see here: https://doc.rust-lang.org/std/net/enum.SocketAddr.html

I could add support for that though!

node/src/main.rs
timmclean: Yeah, I know :(
I looked long and hard for a way to flush with `slog` but couldn't find anything. Without this, both the panic message and the logger messages are printed, but they overlap sometimes, so it's really just an issue of polish.

This is a general problem that applies to all panics in this function. Maybe I could wrap the call to async_main in [`catch_unwind`](https://doc.rust-lang.org/std/panic/fn.catch_unwind.html) and log any panic messages manually through `slog`?

leodasvacas: You can try dropping `logger` before panicking or creating and dropping an [AsyncGuard](https://docs.rs/slog-async/2.3.0/slog_async/struct.AsyncBuilder.html#method.build_with_guard), it's supposed to guarantee flushing.

leodasvacas: Panicking here in this function is not the best behavior anyways because it will not actually crashed the node, but just leave it hanged in some weird state.

timmclean:
Looking into the race conditions led to some interesting discussion, so I thought I'd write up the core issues here:

1.	**Panics leave the process in a partially operational state.** When a panic happens in one of the tokio workers, the other workers keep running like nothing happened. The panic info is written to stderr, but that's easy to miss since logging from other workers will continue.
2.	**Panics print to stderr, which conflicts with `slog_async`, occasionally leading to garbled output.** `slog_async` passes log records to a worker thread so that they can be processed asynchronously.  In our case, that means writing to stderr asynchronously. `panic!` writes to stderr immediately, regardless of the state of the `slog_async` worker thread, which causes conflicts.

(1) seems like an issue that needs to be solved prior to putting the project into production.  One approach would be to have uncaught panics in any worker thread trigger an orderly process exit (and presumably a supervisor would start a new process).  We could also look at relaunching workers, but then there is more risk of broken invariants.

Proposed interim partial solution to (1): change `main()` to use `runtime.block_on` instead of `tokio::run`, so that panics in `async_main` will end the process (gracefully) and make startup problems more visible.

I wanted to find a general solution for (2), but it was harder than expected.  `slog_async` only has one way of flushing the logs (as far as we can tell): dropping the logger or AsyncGuard instance.  This sounds like something that would happen automatically on a panic, but there's a catch.  Rust writes panics to stderr as part of the ["panic hook"](https://doc.rust-lang.org/std/panic/fn.set_hook.html).  The panic hook is invoked _prior_ to unwinding, and therefore prior to the logger being dropped and flushed.  So technically, the problem here isn't about the logger not being flushed, it's about the logger not being flushed early enough.

Leo and I were looking at using `catch_unwind` to catch panics and do custom logging, but that won't work. Unwinding happens after the panic hook is called, so it's too late, stderr has already been written to.  The only solution I can think of here would be to write a custom panic hook that writes to our logger instead of stderr.  I don't want to jump into doing that without more discussion and consideration for priorities...

I'm thinking I'll use Leo's early suggestion of `std::mem::drop(logger)`, that'll solve this particular case at least.

PR #284, opened 2018-08-20
Fix and un-ignore test, install IPFS on Travis

leodasvacas:
This test depends on IPFS so it was ignored, install IPFS on Travis so we can run it on CI.


Jannis:
Alright, tests have passed. This is good to go.

leodasvacas:
@Jannis Since Travis was getting stuck specifically on this test, I'm worried that it might be flaky, I'll try to improve it and see if Travis succeeds consistently.

leodasvacas:
Removed uses of `wait` which is always suspicious, all green, hopefully it wasn't just chance and the flakiness is gone.

PR #285, opened 2018-08-21
Add subscriptions to the Diesel store and rework QueryRunner

### LINKED ISSUE #273: Add entity change triggers / notification channel to Postgres
This includes:

1. Installing tiggers for when entities are added/changed/removed in the entities table.
2. Having the triggers emit Postgres notifications with subgraph ID, entity type, entity ID, operation type and potentially the entity data itself.

### LINKED ISSUE #274: Add subscription functionality to the Store
This includes:

1. Reading entity change notifications from Postgres.
2. Providing a `subscribe` method that returns a stream of entity change events.
3. Managing these subscriptions internally and allow to terminate them at any point.

Jannis:
Resolves graphprotocol/graph-node#273 and resolves #274.

It took longer than planned but this is approx. the first half of work towards GraphQL subscriptions.

1. Postgres now emits entity change notifications for any change to an entity in the entities table.
2. The Store provides a new subscription API for other components to subscribe to entity change events for a given subgraph and set of entities.
3. The QueryRunner is now called directly to execute queries. We will add a way to execute subscriptions to it in the same way. This is less convoluted than the one-shot result sender we were using before.


store/postgres/src/store.rs
```diff
@@ -257,15 +312,36 @@ impl StoreTrait for Store {
...
         }
+        subscriptions.insert(id.clone(), subscription);
+
+        // Return the subscription ID and entity change stream
+        (id, Box::new(receiver))
+    }
+
+    fn unsubscribe(&mut self, id: String) {
+        let subscriptions = self.subscriptions.clone();
+        subscriptions.write().unwrap().remove(&id);
```
leodasvacas: Is this sophisticated cancellation system necessary? Wouldn't it be sufficient to drop the receiver, then if a `send` fails in `handle_entity_changes` that sender is removed from the list?

store/postgres/tests/store.rs
```diff
@@ -43,12 +44,22 @@ where
...
-        let result = panic::catch_unwind(|| test());
-        remove_test_data();
-        sender.send(result).map_err(|_| ())
-    }));
+
+    let mut runtime = tokio::runtime::Runtime::new().unwrap();
+    tokio_executor::with_default(
+        &mut runtime.executor(),
+        &mut tokio_executor::enter().unwrap(),
+        |_| {
```
leodasvacas: I think this whole `tokio_executor::with_default` stuff is actually not necessary, and if you just remove it and leave only the inner `runtime.block_on` it will still work.

store/postgres/src/store.rs
```diff
@@ -5,17 +5,29 @@ use diesel::prelude::*;
 use diesel::sql_types::Text;
 use diesel::{debug_query, delete, insert_into, result, select};
 use filter::store_filter;
+use futures::stream;
```
leodasvacas: This `use` is in `graph::prelude`.

leodasvacas:
Also this has some dead code warnings.

graph/src/components/store.rs
```diff
@@ -92,6 +92,15 @@ pub enum StoreEvent {
...
 }
 
+/// Entity change events emitted by [Store](trait.Store.html) implementations.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct EntityChange {
+    pub subgraph: String,
+    pub entity: String,
+    pub id: String,
+    pub data: Entity,
+}
```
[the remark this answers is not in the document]
Jannis: Yep, agreed, will add doc strings.

core/src/query/runner.rs
```diff
@@ -1,4 +1,5 @@
-use futures::sync::mpsc::{channel, Receiver, Sender};
+use futures::future;
+use futures::prelude::*;
```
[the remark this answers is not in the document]
Jannis: Noted, I'll update the files I changed accordingly.

store/postgres/src/entity_changes.rs
```diff
@@ -0,0 +1,73 @@
...
+            let iter = notifications.blocking_iter();
+
+            conn.execute("LISTEN entity_changes", &[])
+                .expect("failed to listen to entity changes in Postgres");
+
+            iter.iterator()
+                .filter_map(|notification| match notification {
+                    Ok(notification) => Some(notification),
+                    Err(_) => None,
+                })
```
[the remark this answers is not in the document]
Jannis: Oh, nice, indeed. That's very elegant.

store/postgres/src/store.rs
Jannis: I'm ok doing it that way. It's definitely cleaner than this extra logic.

Jannis: Also: sophisticated? 🤣

Jannis:
@leodasvacas Reworked how subscriptions are removed and addressed the other comments. Mind taking another look?

I wonder what you think about https://github.com/graphprotocol/graph-node/pull/285/commits/f430ae536eb5e99c38c302b12421e4d0edd1daa2#diff-3b473a972b5cc133476d080697a5f36bR155 in particular — I bet the extra `for` loop may not be necessary; there must be a more elegant way.

Jannis:
I'm going to rebase this one and add a Postgres store test that verifies that Postgres notifications are sent and forwarded to store subscriptions. I won't change anything else though, so this can still be reviewed fine while I'm working on that.

leodasvacas:
Heads up: The `SchemaProvider` removal caused conflicts.

server/http/src/server.rs
```diff
@@ -101,41 +96,18 @@ impl GraphQLServer {
...
-            info!(logger, "Received store event"; "event" => format!("{:?}",  event));
-            Ok(())
-        }));
-    }
 }
 
-impl GraphQLServerTrait for GraphQLServer {
+impl<Q> GraphQLServerTrait for GraphQLServer<Q>
+where
+    Q: QueryRunner + Sized + 'static,
```
leodasvacas: The `Sized` bound is implicit, you don't need to include it.

server/http/src/service.rs
```diff
@@ -143,8 +167,9 @@ mod tests {
                  ",
             ).unwrap(),
         })));
-        let (query_sink, _) = channel(1);
-        let mut service = GraphQLService::new(schema, query_sink);
+
+        let query_runner = Arc::new(TestQueryRunner::default());
```
leodasvacas: You don't need the `Default` derive, you can write simply `TestQueryRunner` and that will make one for you.

server/http/tests/server.rs
```diff
@@ -41,8 +44,8 @@ fn rejects_empty_json() {
         .block_on(futures::lazy(|| {
             let logger = slog::Logger::root(slog::Discard, o!());
 
-            let mut server = HyperGraphQLServer::new(&logger);
-            let query_stream = server.query_stream().unwrap();
+            let query_runner = Arc::new(TestQueryRunner::default());
```
leodasvacas: Same here and other occurrences in this file.

store/postgres/Cargo.toml
```diff
@@ -8,10 +8,14 @@ diesel = { version = "1.3.2", features = ["postgres", "serde_json", "numeric"] }
...
-slog = "2.2.3"
+postgres = "0.15.2"
+serde = "1.0"
 serde_json = "1.0"
+slog = { version = "2.2.3", features = ["release_max_level_debug"] }
+uuid = "0.6"
 
 [dev-dependencies]
-lazy_static = "1.1"
+lazy_static = "1.1"
```
leodasvacas: newline

store/postgres/Cargo.toml
```diff
@@ -8,10 +8,14 @@ diesel = { version = "1.3.2", features = ["postgres", "serde_json", "numeric"] }
...
 ethereum-types = "0.3"
 diesel-dynamic-schema = { git = "https://github.com/diesel-rs/diesel-dynamic-schema" }
+fallible-iterator = "0.1.4"
 futures = "0.1.21"
 graph = { path = "../../graph" }
-slog = "2.2.3"
+postgres = "0.15.2"
+serde = "1.0"
 serde_json = "1.0"
+slog = { version = "2.2.3", features = ["release_max_level_debug"] }
```
leodasvacas: `slog` should only be in the `Cargo.toml` of the `graph` crate, where we control the logging level. To import the slog stuff you need to `use graph::prelude::*`.

store/postgres/src/entity_changes.rs
```diff
@@ -0,0 +1,72 @@
...
+                            value
+                        ).as_str(),
+                    )
+                });
+
+            for change in changes {
+                // We'll assume here that if sending fails, this means that the
+                // entity change listener has already been dropped, the receiving
+                // is gone and we should terminate the listener loop
+                if let Err(_) = sender.clone().send(change).wait() {
```
leodasvacas: This is elegant. It'd be good to note that `for change in changes` will loop forever. You can use `wait().is_err()` instead of `if let Err(_) =`.

store/postgres/src/store.rs
```diff
@@ -54,19 +66,105 @@ impl Store {
...
+                        .collect::<Vec<_>>();
+
+                    // Remove all stale subscriptions
+                    for id in stale_ids {
+                        subscriptions.remove(&id);
+                    }
+
+                    Ok(())
+                })
+                .map_err(|_| ()),
```
leodasvacas: This can never error so you can replace `()` with `unreachable!()`

store/postgres/src/store.rs
```diff
@@ -232,15 +330,31 @@ impl BasicStore for Store {
...
+            sender,
+        };
+
+        // Add the new subscription (log an error if somehow the UUID is not unique)
+        let subscriptions = self.subscriptions.clone();
+        let mut subscriptions = subscriptions.write().unwrap();
+        if subscriptions.contains_key(&id) {
+            error!(self.logger, "Duplicate Store subscription ID generated; \
+                                 subscriptions will not work as expected";
+                   "id" => &id);
```
leodasvacas: This is virtually impossible to happen, we can just remove this check.

leoyvens (changes requested):
Overall looking good, left a few nits. Someone who knows their SQL would have to look at the DB functions, @fordN?

store/postgres/src/store.rs
Jannis: I'm not sure I want to bet on that (see https://softwareengineering.stackexchange.com/a/130298), even if it's unlikely to ever happen.

leodasvacas: If we think this can happen then panicking would be more correct than logging and leaving the state inconsistent. Or looping until we generate a different one, unless we're concerned about generating duplicate UUIDs forever.

Jannis: Hey, the looping idea is good. An infinite sequence of colliding UUIDs is _VERY_ unlikely. 😁

server/http/src/service.rs
```diff
@@ -118,20 +122,40 @@ impl Service for GraphQLService {
...
+    use std::iter::FromIterator;
     use std::sync::Mutex;
 
     use graph::prelude::*;
 
     use super::GraphQLService;
     use test_utils;
 
+    /// A simple stupid query runner for testing.
+    #[derive(Default)]
```
leodasvacas: This derive can be removed.

server/http/tests/server.rs
```diff
@@ -6,32 +6,35 @@ extern crate http;
...
-                map.insert("name".to_string(), Value::String("Jordi".to_string()));
-                let data = Value::Object(map);
-                let result = QueryResult::new(Some(data));
-                query.result_sender.send(result).unwrap();
-                Ok(())
-            })
-            .fuse(),
-    );
+/// A simple stupid query runner for testing.
+#[derive(Default)]
```
leodasvacas: Can also be removed.

leoyvens (changes requested):

Jannis:
I'm not sure why Travis shows this as failing, as the tests passed fine: https://travis-ci.org/graphprotocol/graph-node/builds/420161088

leoyvens (approved):

Jannis:
@graphprotocol/developers Mind doing another review of this one? I made a few more changes to get this in shape for merging. It'll need rebasing against master but I want to wait with that until the tests have passed.

store/postgres/src/entity_changes.rs
```diff
@@ -0,0 +1,156 @@
...
+        url: String,
+    ) -> (
+        Receiver<EntityChange>,
+        thread::JoinHandle<()>,
+        Arc<RwLock<bool>>,
+        Arc<Barrier>,
+    ) {
+        // Create two ends of a boolean variable for signalling when the worker
+        // thread should be terminated
+        let terminate = Arc::new(RwLock::new(false));
```
leodasvacas: This can be an `AtomicBool` (we probably lost that improvement in some branch).

store/postgres/src/entity_changes.rs
```diff
@@ -0,0 +1,156 @@
...
+                // HACK: Travis seems to have serious problems with running the
+                // integration tests for the Diesel store. Unless we log frequently
+                // from this thread, it all store tests either take a long time
+                // or never finish (before Travis' 10mins timeout).
+                // As soon as you log something from this thread, suddenly all
+                // tests run relatively fast and reliable. This _could_ be a
+                // timing-sensitive synchronization issue on our side we've tried
+                // almost everything and this hack was the only thing that made
+                // the tests work in Travis.
+                if let Ok(_) = env::var("TRAVIS") {
```
leodasvacas: You can use `.is_some()` here.

store/postgres/src/entity_changes.rs
```diff
@@ -0,0 +1,156 @@
...
+
+                // Terminate the thread if desired
+                let terminate_now = terminate.read().unwrap();
+                if *terminate_now {
+                    return;
+                }
+
+                if let Some(Ok(notification)) = notifications.next() {
+                    // Only handle notifications from the "entity_changes" channel.
+                    if notification.channel != String::from("entity_changes") {
```
leodasvacas: I don't think `String::from` is necessary here.

store/postgres/src/entity_changes.rs
```diff
@@ -0,0 +1,156 @@
...
+                    io::stdout().flush().unwrap();
+                }
+
+                // Terminate the thread if desired
+                let terminate_now = terminate.read().unwrap();
+                if *terminate_now {
+                    return;
+                }
+
+                if let Some(Ok(notification)) = notifications.next() {
```
leodasvacas: This flow would be clearer if where we now have `loop` we had `for notification_result in notifications`. With a `.filter` for the channel name check. Also we should check errors and log them before exiting the loop, right now errors are silent.

store/postgres/src/entity_changes.rs
```diff
@@ -0,0 +1,156 @@
...
+
+        (receiver, worker_handle, terminate_worker, worker_barrier)
+    }
+}
+
+impl Drop for EntityChangeListener {
+    fn drop(&mut self) {
+        // When dropping the change listener, also make sure we signal termination
+        // to the worker and wait for it to shut down
+        if let Some(worker_handle) = self.worker_handle.take() {
```
leodasvacas: This should be unwrapped as it never fails.

store/postgres/tests/store.rs
```diff
@@ -1544,5 +1632,127 @@ fn revert_block_with_partial_update() {
...
+        let logger = Logger::root(slog::Discard, o!());
+        let url = postgres_test_url();
+        let mut store =
+            DieselStore::new_without_processing_entity_changes(StoreConfig { url }, &logger);
+
+        // Create a store subscription
+        let subscription = store.subscribe(String::from("subgraph-id"), vec![String::from("User")]);
+
+        // Start processing entity changes from the DB now that we're subscribed
+        store.start_processing_entity_changes();
```
leodasvacas: Thinking again about this, it no longer makes sense to me. It shouldn't matter that the listener is already started, it should only matter that the store changes are done after the subscription.

leoyvens (changes requested):

store/postgres/tests/store.rs
Jannis: That is... true. Let me try without this.

.travis.yml
```diff
@@ -57,4 +57,4 @@ script:
       cargo fmt --all -- --check;
     fi
   # Run unit and sccenario tests
-  - RUST_BACKTRACE=1 cargo test --verbose --all
```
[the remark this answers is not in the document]
Jannis: Yes, very much so.

mock/src/server.rs
```diff
@@ -85,85 +80,47 @@ impl MockGraphQLServer {
...
+                Query {
+                    schema: schema.clone().unwrap(),
+                    document: graphql_parser::parse_query("{ allUsers { name }}").unwrap(),
+                    variables: None,
+                }
             })
-            .collect::<Vec<(oneshot::Receiver<QueryResult>, Query)>>();
+            .collect::<Vec<Query>>();
 
         println!("Requests: {:?}", requests);
```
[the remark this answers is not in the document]
Jannis: Nope!

store/postgres/src/entity_changes.rs
Jannis: Indeed.

Jannis: `.is_ok()` you mean.

.travis.yml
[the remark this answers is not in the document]
Jannis: All other points have been addressed. Care to take another look?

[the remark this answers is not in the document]
leodasvacas: Could you add a comment on why it's necessary on Travis? Since we don't use it locally anyone that takes a look at this might try to remove it.

leoyvens (approved):
Looks good, could be nice to add a comment.

Jannis:
@leodasvacas Better: I've made CI run tests in multiple threads again. It no longer seems to be a problem.

Merging this once the tests have passed.

leodasvacas:
:shipit:

Issue #282, opened 2018-08-20
Add deploy and remove admin API commands


yanivtal:

leodasvacas:
See comments on #227 for specs.

Issue #227, opened 2018-08-01
Support hot deploying a subgraph to a running node


yanivtal:
If we do deploys through an admin RPC endpoint then the Graph CLI can just be a client over RPC.

leodasvacas:
This + multiple subgraphs have a lot of surface area so I have a bunch of questions on the requirements, from a discussion with @Jannis:

What happens to current CLI support:
- Do we still want to receive subgraphs through the CLI?
- If so, will the CLI support multiple subgraphs?

HTTP endpoint:
- What CLI options do we need for it?
- What will the endpoint be?
- What operations will it support, add/remove/update subgraph?
- Will we use a protocol such as JSON-RPC? 
- What's the payload, simply the subgraph hash?
- What will we do for auth?

Deploy through graph-cli:
- Will we add a subcommand like `graph deploy --node IP_ADDR subgraph.yaml`?
- What parameters and options will it take?

yanivtal:
What happens to current CLI support:
- Yes, let's continue to accept the currently supported CLI arguments
- No, it could be a nice enhancement to allow passing multiple subgraphs on the CLI but it's not necessary for now

HTTP endpoint:
- I think the only CLI option we need for now is the port. We can see if there are more once we get further into implementation
- Operations: add, remove, update, pause, list
   - Right now subgraphs don't have names in the YAML. The naming is handled at the GNS layer. We might need to add support for names to make `update` more ergonomic. It should be possible to just pass a new subgraph hash and have the node stop the old version and start the new one. This can be done naively for now and a separate ticket can track making upgrades more efficient.
- I would recommend [grpc](https://grpc.io/)
- Payload: subgraph hash
- No auth for this first pass. Let's create a separate ticket for this.

Deploy through graph-cli:
- Yes this looks good `graph deploy SUBGRAPH_HASH --node HOST_AND_PORT`

Jannis:
For the admin endpoint: How about GraphQL mutations rather than GRPC? It would require a small amount of work to support mutations in our execution logic but it would align well with our vision.

We could then have an admin schema and could implement the operations through a dedicated resolver.

Jannis:
Regarding the subgraph name: we could take in `--subgraph [OPTIONAL NAME:]IPFS_HASH` and use `graph deploy --node HOST_AND_PORT [OPTIONAL_NAME:]IPFS_HASH` for deploying.

leodasvacas:
[...]

PR #291, opened 2018-08-23
Initial multiple subgraph support

### LINKED ISSUE #226: Support hosting multiple subgraphs

leodasvacas:
This is a first implementation of support for hosting multiple subgraphs, according to [this plan](https://github.com/graphprotocol/graph-node/issues/227#issuecomment-413149755).

Only the `add` operation is supported. Subgraph must have a name when added, and the GraphQL server keeps a map from name to schema. Each subgraph is mounted relative to it's name, so graphiql for subgraph "foo" is mounted at `/foo/`. Because we don't have an index page yet and to not disrupt the current workflow, if only one subgraph is hosted then the root will redirect to it.

At the store level we're already selecting by subgraph id as @fordN informs me so no changes seem necessary there.

Notably remaining work is to support more operations #282 and to have an index page for subgraphs #303.

resolves #226

### DIFF
--- a/server/http/src/server.rs
+++ b/server/http/src/server.rs
@@ -95,10 +95,15 @@ impl GraphQLServer {
                     },
                     Err(e) => return Ok(error!(logger, "error deriving schema {}", e)),
                 };
-                schemas
+                if schemas
                     .write()
                     .unwrap()
-                    .insert(derived_schema.name.clone(), derived_schema);
+                    .insert(derived_schema.name.clone(), derived_schema)
+                    .is_some()
+                {
+                    // We need to support clean shutdown of a subgraph before update support.
+                    panic!("updating a subgraph is not supported yet")
+                }
             } else {
                 panic!("schema removal is yet not supported")
             }
--- a/server/http/src/server.rs
+++ b/server/http/src/server.rs
@@ -2,10 +2,11 @@ use futures::sync::mpsc::{channel, Receiver, Sender};
 use hyper;
 use hyper::Server;
 
+use std::collections::BTreeMap;
 use std::error::Error;
 use std::fmt;
 use std::net::{Ipv4Addr, SocketAddrV4};
-use std::sync::Mutex;
+use std::sync::RwLock;
 
 use graph::components::store::StoreEvent;
 use graph::data::query::Query;
@@ -49,7 +50,8 @@ pub struct GraphQLServer {
     query_sink: Option<Sender<Query>>,
     schema_event_sink: Sender<SchemaEvent>,
     store_event_sink: Sender<StoreEvent>,
-    schema: Arc<Mutex<Option<Schema>>>,
+    // Maps a graph name to it's schema.
+    schemas: Arc<RwLock<BTreeMap<String, Schema>>>,
 }
 
 impl GraphQLServer {
@@ -65,7 +67,7 @@ impl GraphQLServer {
             query_sink: None,
             schema_event_sink,
             store_event_sink: store_sink,
-            schema: Arc::new(Mutex::new(None)),
+            schemas: Arc::new(RwLock::new(BTreeMap::new())),
         };
 
         // Spawn tasks to handle incoming schema and store events.
@@ -79,21 +81,24 @@ impl GraphQLServer {
     /// Handle incoming schema events.
     fn handle_schema_events(&mut self, stream: Receiver<SchemaEvent>) {
         let logger = self.logger.clone();
-        let schema = self.schema.clone();
+        let schemas = self.schemas.clone();
 
         tokio::spawn(stream.for_each(move |event| {
             info!(logger, "Received schema event");
 
             if let SchemaEvent::SchemaAdded(new_schema) = event {
-                let mut schema = schema.lock().unwrap();
                 let derived_schema = match api_schema(&new_schema.document) {
                     Ok(document) => Schema {
+                        name: new_schema.name.clone(),
                         id: new_schema.id.clone(),
                         document,
                     },
                     Err(e) => return Ok(error!(logger, "error deriving schema {}", e)),
                 };
-                *schema = Some(derived_schema);
+                schemas
+                    .write()
+                    .unwrap()
+                    .insert(derived_schema.name.clone(), derived_schema);
             } else {
                 panic!("schema removal is yet not supported")
             }
@@ -153,9 +158,9 @@ impl GraphQLServerTrait for GraphQLServer {
         // On every incoming request, launch a new GraphQL service that writes
         // incoming queries to the query sink.
         let query_sink = query_sink.clone();
-        let schema = self.schema.clone();
+        let schemas = self.schemas.clone();
         let new_service = move || {
-            let service = GraphQLService::new(schema.clone(), query_sink.clone());
+            let service = GraphQLService::new(schemas.clone(), query_sink.clone());
             future::ok::<GraphQLService, hyper::Error>(service)
         };
 
@@ -171,7 +176,6 @@ impl GraphQLServerTrait for GraphQLServer {
 #[test]
 fn emits_an_api_schema_after_one_schema_is_added() {
     use graph_graphql::schema::ast;
-    use std::ops::Deref;
     use std::time::{Duration, Instant};
 
     let mut runtime = tokio::runtime::Runtime::new().unwrap();
@@ -187,6 +191,7 @@ fn emits_an_api_schema_after_one_schema_is_added() {
                 let input_doc =
                     ::graphql_parser::parse_schema("type User { name: String! }").unwrap();
                 let input_schema = Schema {
+                    name: "input-schema".to_string(),
                     id: "input-schema".to_string(),
                     document: input_doc,
                 };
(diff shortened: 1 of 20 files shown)


server/http/src/service.rs
```diff
@@ -16,14 +19,36 @@ pub type GraphQLServiceResponse =
...
+    fn index(&self) -> GraphQLServiceResponse {
+        let len = self.schemas.read().unwrap().len();
+        match self.schemas.read().unwrap().values().next().clone() {
+            // If there's only 1 schema, redirect to it.
+            Some(schema) if len == 1 => Box::new(future::ok(
+                Response::builder()
+                    .status(StatusCode::SEE_OTHER)
+                    .header(
+                        header::LOCATION,
+                        header::HeaderValue::from_str(&format!("/{}", schema.name))
```
leodasvacas: @timmclean there are URL injection risks here, because the graph name is inserted directly in the redirect the URL and there are no restrictions on a name and no percent encoding, so it could redirect to any route. Do you have an opinion on a good solution here?

timmclean: Ah, interesting. Is `schema.name` just any UTF-8 string the user provides? That's tough. Usual solutions here are: (1) use an ID instead (randomly generated with only friendly characters), (2) generate a "sanitized" version of the name that uses only a whitelist of characters, (3) both but the ID takes precedence.

I think I'd suggest (1), unless you want the name in there for better UX. StackOverflow and many news sites do (3) for SEO reasons. (2) can work, but then you need to make sure that no two schemas have the same sanitized name.
Also, seems like a good idea to follow Jannis's earlier suggestion with using a URL path library to get a path.join function

leodasvacas: Thanks for the advice here! For now this is a naive naming system, we will build a Graph Naming System on ethereum contracts to solve things like name collisions. Allowing the user to choose a name makes a more friendly UI. Would percent encoding/decoding the name be a decent implementation of (2)?

timmclean: So, straight-up percent encoding doesn't entirely solve the problem, because `..` percent-encodes to `..` (and even if you force it to be `%2e%2e`, that still has the "parent directory" semantics). The sanitization approach wouldn't actually use encoding: it involves completely removing any characters that are not in the whitelist.

I _think_ you can probably get away with using [the method Jannis suggested earlier](https://docs.rs/url/1.7.1/url/struct.PathSegmentsMut.html#method.push), if you check afterwards that a component was actually pushed (that method treats `.`, `..`, and their encoded equivalents as no-ops -- it should treat them as errors in my opinion).  It's just a bit riskier, because it depends on implementation details of the server. E.g. imagine a server that ignores NUL bytes. It looks like the string `%2e%2e%00` would be accepted by the `push` method as a valid path component, but once that URL reaches the server, the server would ignore `%00` and see only `%2e%2e`, which decodes to `..` again.  I think that behaviour would be against spec, but, you know how that often goes...

server/http/src/server.rs
```diff
@@ -49,7 +50,8 @@ pub struct GraphQLServer {
     query_sink: Option<Sender<Query>>,
     schema_event_sink: Sender<SchemaEvent>,
     store_event_sink: Sender<StoreEvent>,
-    schema: Arc<Mutex<Option<Schema>>>,
+    // Maps a graph name to it's schema.
```
Jannis: I think you mean `subgraph name` here. Can we make this `Maps subgraph names and IDs to their schemas`? I'll describe why further down.

server/http/src/server.rs
```diff
@@ -79,21 +81,29 @@ impl GraphQLServer {
...
         let logger = self.logger.clone();
-        let schema = self.schema.clone();
+        let schemas = self.schemas.clone();
 
         tokio::spawn(stream.for_each(move |event| {
             info!(logger, "Received schema event");
 
             if let SchemaEvent::SchemaAdded(new_schema) = event {
-                let mut schema = schema.lock().unwrap();
                 let derived_schema = match api_schema(&new_schema.document) {
```
Jannis: I would like us to serve a subgraph with name `xyz` and ID `abc` via two URLs: `/xyz` and `/abc`. The simplest way to achieve that would probably be to insert the schema in the `schemas` map twice (once under the name, once under the subgraph ID) and then resolve both in the same way. Thoughts?

server/http/src/service.rs
```diff
@@ -94,24 +122,41 @@ impl Service for GraphQLService {
...
                 self.serve_file(include_str!("../assets/graphiql.min.js"))
             }
 
-            // POST / receives GraphQL queries
-            (&Method::POST, "/graphql") => self.handle_graphql_query(req),
+            // Request is relative to a subgraph.
+            (method, path) => {
+                let mut path = path.split('/');
+                let _empty = path.next();
+                let name = path.next();
```
Jannis: As a consequence of handling both, names and IDs here, perhaps `name` here should be `subgraph` or `name_or_id`. But I think `subgraph` is ok.

server/http/src/service.rs
```diff
@@ -94,24 +122,41 @@ impl Service for GraphQLService {
...
+                match (method, name, rest.as_str()) {
+                    // GraphiQL
+                    (Method::GET, Some(_), "") => {
+                        self.serve_file(include_str!("../assets/index.html"))
+                    }
 
-            // OPTIONS / allows to check for GraphQL HTTP features
-            (&Method::OPTIONS, "/graphql") => self.handle_graphql_options(req),
+                    // POST / receives GraphQL queries
+                    (Method::POST, Some(name), "graphql") => self.handle_graphql_query(&name, req),
```
Jannis: Same here.

server/http/src/service.rs
```diff
@@ -16,14 +19,36 @@ pub type GraphQLServiceResponse =
...
+        let len = self.schemas.read().unwrap().len();
+        match self.schemas.read().unwrap().values().next().clone() {
+            // If there's only 1 schema, redirect to it.
+            Some(schema) if len == 1 => Box::new(future::ok(
+                Response::builder()
+                    .status(StatusCode::SEE_OTHER)
+                    .header(
+                        header::LOCATION,
+                        header::HeaderValue::from_str(&format!("/{}", schema.name))
+                            .expect("invalid subgraph name"),
```
Jannis: `subgraph name` -> `subgraph name or ID`

server/http/src/service.rs
```diff
@@ -37,19 +62,22 @@ impl GraphQLService {
     }
 
     /// Handles GraphQL queries received via POST /.
-    fn handle_graphql_query(&self, request: Request<Body>) -> GraphQLServiceResponse {
+    fn handle_graphql_query(&self, name: &str, request: Request<Body>) -> GraphQLServiceResponse {
```
Jannis: `name` -> `subgraph`

server/http/src/service.rs
timmclean: Sounds good!

leodasvacas:
@timmclean I've followed the solution we discussed of restricting subgraph names. Now only ascii alphanumerics, `-` and `_` are allowed.

server/json-rpc/src/lib.rs
```diff
@@ -45,30 +43,18 @@ impl JsonRpcServerTrait for JsonRpcServer {
...
             info!(add_logger, "Received subgraph_add request"; "params" => params.to_string());
             provider
-                .add(format!("/ipfs/{}", params.ipfs_hash))
+                .add(params.name, format!("/ipfs/{}", params.ipfs_hash))
                 .map_err(|e| json_rpc_error(0, e.to_string()))
                 .map(|_| Ok(Value::Null))
                 .flatten()
         };
         handler.add_method("subgraph_add", move |params: Params| {
             let handler = add_handler.clone();
```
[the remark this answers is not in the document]
leodasvacas: The clone is necessary because we can't move `add_handler` through the outer closure straight into the inner one. But I tried to clean up this code by moving the handler logic into a function.

server/json-rpc/src/lib.rs
```diff
@@ -45,30 +43,18 @@ impl JsonRpcServerTrait for JsonRpcServer {
...
-                Err(json_rpc_error(
-                    1,
-                    "adding multiple subgraphs is not yet supported".to_owned(),
-                ))
-            } else {
-                Ok(())
-            }.into_future()
-                .and_then(|_| params.parse())
-                .and_then(handler)
+            params.parse().into_future().and_then(handler)
```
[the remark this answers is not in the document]
leodasvacas: Doing that would mean cloning the provider and logger twice, that's why I avoided it. This code is not very pretty but I can't think of anything strictly better right now.

leodasvacas:
@Jannis All points addressed I believe

server/http/src/service.rs
```diff
@@ -37,19 +70,31 @@ impl GraphQLService {
...
         Box::new(
             request
                 .into_body()
                 .concat2()
                 .map_err(|_| GraphQLServerError::from("Failed to read request body"))
-                .and_then(move |body| {
-                    let schema = schema.lock().unwrap();
-                    GraphQLRequest::new(body, schema.clone())
-                })
+                .and_then(move |body| GraphQLRequest::new(body, schema.clone()))
```
Jannis: I wonder: since we're already cloning the schema in line 86, is this clone necessary?

leodasvacas: Yes, because the closure owns `schema` and if it moves it into the `GraphQLRequest` when called then it'd be a `FnOnce` not an `Fn`.

leodasvacas:
Oops, tests fixed.

server/http/src/service.rs
Jannis: Got it!

PR #307, opened 2018-08-27
Improve IPFS startup errors and implement custom panic hook with guarded logger

### LINKED ISSUE #268: Improve IPFS unavailable error

fordN:
Resolves #268 

Return clear error messaging to the user if the IPFS connection fails on startup.  

An attempt to add a file to the IPFS node is used to test the connection before moving on to the rest of the node setup. To facilitate the improved error messaging a few general changes were also made: 
- implement custom panic hook 
- use guarded async slog in panic logging to ensure logs are returned to the user

### DIFF
--- a/graph/src/util/log.rs
+++ b/graph/src/util/log.rs
@@ -1,6 +1,7 @@
 use slog;
 use slog_async;
 use slog_term;
+use std::panic;
 
 use slog::Drain;
 
@@ -16,4 +17,19 @@ pub fn guarded_logger() -> (slog::Logger, slog_async::AsyncGuard) {
     let drain = slog_term::CompactFormat::new(decorator).build().fuse();
     let (drain, guard) = slog_async::Async::new(drain).build_with_guard();
     (slog::Logger::root(drain.fuse(), o!()), guard)
+}
+
+pub fn register_panic_hook(panic_logger: slog::Logger) {
+    panic::set_hook(Box::new(move |panic_info| {
+        let panic_payload = panic_info.payload().downcast_ref::<String>();
+        let panic_location = if let Some(location) = panic_info.location() {
+            format!("{}:{}", location.file(), location.line().to_string())
+        } else {
+            "NA".to_string()
+        };
+        error!(panic_logger, "Node error";
+            "error" => panic_payload.clone(),
+            "location" => panic_location.clone()
+           );
+    }));
 }
--- a/graph/src/util/log.rs
+++ b/graph/src/util/log.rs
@@ -28,20 +28,16 @@ pub fn register_panic_hook(panic_logger: slog::Logger) {
         } else {
             "NA".to_string()
         };
-        error!(panic_logger, "Node error";
+        crit!(panic_logger, "Node error";
             "error" => panic_payload.clone(),
             "location" => panic_location.clone()
            );
         match env::var_os("RUST_BACKTRACE") {
-            Some(val) => {
-                if &val == "0" {
-                    ()
-                } else {
-                    error!(panic_logger, "Backtrace";
+            Some(ref val) if val != "0" => {
+                crit!(panic_logger, "Backtrace";
                     "trace" => format!("{:?}", Backtrace::new()));
-                }
             }
-            None => (),
+            _ => (),
         }
     }));
 }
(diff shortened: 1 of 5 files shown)


graph/src/util/log.rs
```diff
@@ -10,3 +11,25 @@ pub fn logger() -> slog::Logger {
...
+        let panic_location = if let Some(location) = panic_info.location() {
+            format!("{}:{}", location.file(), location.line().to_string())
+        } else {
+            "NA".to_string()
+        };
+        error!(panic_logger, "Node error";
+            "error" => panic_payload.clone(),
+            "location" => panic_location.clone()
+           );
+    }));
```
timmclean: How about adding backtrace support too? Looks like it'd be easy enough to steal that code from the default hook: https://doc.rust-lang.org/src/std/panicking.rs.html#210

fordN:
I've updated the panic_hook to include the backtrace if the user sets `RUST_BACKTRACE` to a value greater than zero.

fordN:
This PR successfully logs the IPFS error, but not all tokio workers shutdown. 
I created another branch for my work on shutting down the entire tokio event loop, [link to other branch](graphprotocol/graph-node/tree/ford/panic-shuts-down-event-loop).  
In its current state, a panic causes the runtime to shut down successfully after logging the error, but a second panic fires in the process.  :(

graph/src/util/log.rs
```diff
@@ -10,3 +12,36 @@ pub fn logger() -> slog::Logger {
...
+
+pub fn register_panic_hook(panic_logger: slog::Logger) {
+    panic::set_hook(Box::new(move |panic_info| {
+        let panic_payload = panic_info.payload().downcast_ref::<String>();
+        let panic_location = if let Some(location) = panic_info.location() {
+            format!("{}:{}", location.file(), location.line().to_string())
+        } else {
+            "NA".to_string()
+        };
+        error!(panic_logger, "Node error";
```
leodasvacas: slog has a `crit!` level which seems suitable here.

graph/src/util/log.rs
```diff
@@ -10,3 +12,36 @@ pub fn logger() -> slog::Logger {
...
+            "error" => panic_payload.clone(),
+            "location" => panic_location.clone()
+           );
+        match env::var_os("RUST_BACKTRACE") {
+            Some(val) => {
+                if &val == "0" {
+                    ()
+                } else {
+                    error!(panic_logger, "Backtrace";
```
leodasvacas: I guess this should also be `crit!` to match the error message.

graph/src/util/log.rs
```diff
@@ -10,3 +12,36 @@ pub fn logger() -> slog::Logger {
...
+        } else {
+            "NA".to_string()
+        };
+        error!(panic_logger, "Node error";
+            "error" => panic_payload.clone(),
+            "location" => panic_location.clone()
+           );
+        match env::var_os("RUST_BACKTRACE") {
+            Some(val) => {
+                if &val == "0" {
```
leodasvacas: This can be written as `Some(val) if val == "0" {` and then the second branch would be `_ => ()`

graph/src/util/log.rs
```diff
@@ -10,3 +12,33 @@ pub fn logger() -> slog::Logger {
...
+        } else {
+            "NA".to_string()
+        };
+        crit!(panic_logger, "Node error";
+            "error" => panic_payload.clone(),
+            "location" => panic_location.clone()
+           );
+        match env::var_os("RUST_BACKTRACE") {
+            Some(ref val) if val == "0" => (),
+            Some(_val) => {
```
leodasvacas: Sorry, I got it the wrong way around. This should be `Some(val) if val != "0"` and then you can remove the first match arm.

leodasvacas:
@fordN You need to rustfmt, after that this is good to go.

PR #308, opened 2018-08-27
Detailed logging of ethereum event handling

fordN:
Resolve graphprotocol/graph-cli#96

Add detail to logging of Ethereum event subscriptions, event receipt, event handling, and contract calls. 
- Prepare subscription  --> now logging event name and handler name
- Subscribe to event --> now logging subscription id
- Ethereum event received --> now logging event signature
- Contract call --> now logging block hash and function name
- Call event handler --> now logging event signature

### DIFF
--- a/runtime/wasm/src/host.rs
+++ b/runtime/wasm/src/host.rs
@@ -154,6 +154,10 @@ impl RuntimeHost {
                     .event_handlers
                     .iter()
                     .map(|event_handler| {
+                        debug!(logger, "Prepare subscription";
+                                "event" => &event_handler.event,
+                                "handler" => &event_handler.handler,
+                            );
                         let subscription_id = Uuid::new_v4().simple().to_string();
                         let event = util::ethereum::contract_event_with_signature(
                             &contract,
@@ -178,8 +182,9 @@ impl RuntimeHost {
             // Merge all event streams.
             let mut event_stream: Box<Stream<Item = _, Error = _>> = Box::new(stream::empty());
             for subscription in subscription_results {
-                info!(logger, "Subscribe to event"; "name" => &subscription.event.name);
-
+                info!(logger, "Subscribe to event";
+                      "name" => &subscription.event.name,
+                      "subscription_id" => &subscription.subscription_id);
                 event_stream = Box::new(
                     event_stream.select(
                         ethereum_adapter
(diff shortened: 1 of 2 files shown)


runtime/wasm/src/host.rs
```diff
@@ -154,6 +154,10 @@ impl RuntimeHost {
                     .event_handlers
                     .iter()
                     .map(|event_handler| {
+                        debug!(logger, "Prepare subscription";
+                            "event" => &event_handler.event,
```
Jannis: Small misalignment here.

runtime/wasm/src/host.rs
```diff
@@ -178,8 +182,9 @@ impl RuntimeHost {
             // Merge all event streams.
             let mut event_stream: Box<Stream<Item = _, Error = _>> = Box::new(stream::empty());
             for subscription in subscription_results {
-                info!(logger, "Subscribe to event"; "name" => &subscription.event.name);
-
+                info!(logger, "Subscribe to event";
+                    "name" => &subscription.event.name,
```
Jannis: And here.

runtime/wasm/src/host.rs
```diff
@@ -197,7 +202,9 @@ impl RuntimeHost {
 
         event_stream
             .map(move |event| {
-                info!(event_logger, "Ethereum event received");
+                info!(event_logger, "Ethereum event received";
+                    "signature" => event.event_signature.to_string(),
```
Jannis: And here.

runtime/wasm/src/host.rs
```diff
@@ -215,7 +222,8 @@ impl RuntimeHost {
                         .to_owned();
 
                     debug!(event_logger, "  Call event handler";
-                               "name" => &event_handler.handler);
+                               "name" => &event_handler.handler,
```
Jannis: And here.

runtime/wasm/src/module.rs
```diff
@@ -313,8 +313,10 @@ where
         let unresolved_call: UnresolvedContractCall = self.heap.asc_get(call_ptr);
 
         info!(self.logger, "Call smart contract";
+              "function" => &unresolved_call.function_name,
+              "address" => &unresolved_call.block_hash.to_string(),
```
Jannis: This isn't correct. You're passing in the block hash but call it the contract address? Perhaps you want both?

runtime/wasm/src/host.rs
fordN: I'm not clear on the proper alignment for the logging macros; I looked at our codebase for reference and see the second line indented differently in a few spots.

Jannis: Yeah, `rustfmt` doesn't handle them properly. The goal should always be this:
```
debug!(logger, "Prepare subscription";
       "event" => &event_handler.event,
       ...)
```
So: align the `"key" => value` pairs with the other arguments to the macro.

Issue #281, opened 2018-08-20
Plan implementation for block and event streaming


Jannis:

timmclean:
Ford and I had a great discussion about this today, and a few realizations came out of it.

First, the store updates that occur when a mapping is executed should be done atomically within a single transaction. Otherwise, a query might end up seeing entities that are only partially updated to reflect a particular event. It doesn't look like this is handled currently.

Second, we suspect it makes the most sense to make store updates atomic at the block level. So, if one transaction triggers multiple events, then either none of those events or all of those events will be visible to queries, which is less confusing for users.

So, based on the above, I'm thinking that the `EventStream` should group events by block, since this will match what happens in the store.

```rust
// Each EventsUpdate is about a single block
enum EventsUpdate {
    ProcessEvents(Vec<Event>, BlockHash),
    RollbackEvents(BlockHash),
}

// EventStream is a EventProducer<EventsUpdate>
// Note: should we use "Log" to avoid overloading the term "Event"?
```

Another piece that came up is that, currently, the Ethereum adapter pulls all events matching a filter since the beginning of time, so that the mappings have a chance to see every event that might be relevant to them. I don't want to have to produce an `EventsUpdate` for every single block back to the genesis, so I'm thinking that `EventStream`s will be allowed (but not required) to omit updates for blocks that contain no relevant events. This also saves a bunch of processing when getting caught up after a node has been disconnected for a while.

I made an unfortunate discovery today: [`eth_getLogs`](https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_getlogs) allows filtering with `fromBlock` and `toBlock`, but it requires those values to be block numbers, not block hashes. This seems like a major oversight, since the meaning of "block #N" can change over time. I think this RPC call is only safe to use for block numbers much below that of the latest block -- in other words, blocks that are old enough such that we can assume there will be no reorgs -- otherwise, we can't know what blocks we're actually accessing. For new blocks, we'll have to use the slower, block-by-block approach, checking each block's bloom filter.

---

I'm going to start the implementation tomorrow with a `BlockIngestor` component. It will be a `tokio::timer::Interval` used to poll for latest block via web3. Each poll result will be used following the algorithm from #235:
- upsert into DB
- check for gaps
- update chain.head
- garbage collect
- updates to chain head will trigger `NOTIFY chain_head_update, ${chain_id}`

timmclean:
Some useful posts on how to choose a value for T:
[...]

PR #309, opened 2018-08-28
Initial implementation of Block Ingestor & BlockStore

### LINKED ISSUE #286: Implement block ingestor component
Based on designs in #235 and #281.

This task includes polling for blocks, storing them in a database, and tracking a head block for each blockchain.

timmclean:
This implementation is based on the old design and will need to be extended (see #281).

Closes #286.


datasource/ethereum/src/block_ingestor.rs
```diff
@@ -0,0 +1,136 @@
...
+        };
+        tokio::spawn(block_ingestor.into_polling_stream());
+
+        Ok(())
+    }
+
+    fn into_polling_stream(self) -> impl Future<Item = (), Error = ()> {
+        let err_logger = self.logger.clone();
+
+        tokio::timer::Interval::new(Instant::now(), Duration::from_secs(5))
```
yanivtal: Where does this 5s come from?

timmclean: Ha, that's the polling interval. Shall I make that a command line arg?

yanivtal: 5s is kind of an eternity. Maybe 1s or less.. Ideally we'd be able to process a block and push any changes in <500ms.

timmclean: Oh I see!
I just pushed a new commit to add `polling_interval` to the `BlockIngestorConfig` and reduce it to 0.5s. Also created #310 for config file.

store/postgres/migrations/2018-08-22-140000_create_head_update_procedure/up.sql
```diff
@@ -0,0 +1,76 @@
...
+
+    -- Set head block pointer to candidate head block
+    UPDATE ethereum_networks
+    SET
+        head_block_hash = new_head_hash,
+        head_block_number = new_head_number
+    WHERE name = net_name;
+
+    -- Fire head block update event
+    NOTIFY head_block_update;
```
Jannis: This works with a single network but it would be better to include the network name and the new head block hash and number in the notification here (e.g. with `pg_notify` + `json_build_object`).

store/postgres/src/store.rs
```diff
@@ -231,6 +234,76 @@ impl BasicStore for Store {
...
+
+        Ok(())
+    }
+
+    fn attempt_head_update(
+        &self,
+        network_name: &str,
+        ancestor_count: u64,
+    ) -> Result<Vec<H256>, Error> {
+        select(attempt_head_update(network_name, ancestor_count as i64))
```
Jannis: Where does the `select` here come from?

graph/src/components/store.rs
```diff
@@ -123,8 +126,40 @@ pub trait BasicStore {
...
+    ///
+    /// If the candidate new head block had one or more missing ancestors, returns
+    /// `Ok(missing_blocks)`, where `missing_blocks` is a nonexhaustive list of missing blocks.
+    fn attempt_head_update(
+        &self,
+        network_name: &str,
+        ancestor_count: u64,
+    ) -> Result<Vec<H256>, Error>;
+
+    /// Delete blocks with block number less than `head_block_number - ancestor_count`.
```
Jannis: Since this isn't used anywhere, I'd leave it out of the PR for now.

graph/src/components/store.rs
```diff
@@ -123,8 +126,40 @@ pub trait BasicStore {
...
+    fn delete_old_blocks(&self, network_name: &str, ancestor_count: u64) -> Result<(), Error>;
+
+    // TODO will receive from postgres "LISTEN" command
```
Jannis: Same with this commented out code, I'd also leave that out.

datasource/ethereum/src/block_ingestor.rs
```diff
@@ -0,0 +1,140 @@
...
+            })
+    }
+
+    fn do_poll(&self) -> Result<(), Error> {
+        let latest_block = self.get_latest_block()?;
+
+        let mut new_blocks = vec![latest_block];
+        while new_blocks.len() > 0 {
+            let missing_block_hashes = self.ingest_blocks(&new_blocks)?;
+            new_blocks = self.get_blocks(&missing_block_hashes)?;
```
leodasvacas: Generally you did a great job of documenting the algorithm, but this section of the code is scarce on comments and I'm finding it a bit hard to follow. What I imagine is happening here in case of a reorg:

- `latest_block` is ingested, we get back a list of up to `ancestor_count` blocks.
- The ancestors on the list are fetched.
- The ancestors are ingested and `attempt_head_update` is a no-op because `latest_block` remains the head.

If this is the algorithm then this while loop doesn't represent it well. If it's not then I'm confused and we could use some comments explaining when this would iterate more than once and when it would terminate.

leodasvacas:
On the implementation of `BlockIngestor` you took a polling approach, is that a design choice or a workaround web3 API limitations? I see they have a [new blocks filter](https://tomusdrw.github.io/rust-web3/web3/api/struct.Eth.html#method.new_block_filter), would that not be preferable?

Calling `wait` from within a tokio task is not ok, to remove them you need to return futures and streams then use their combinator methods to glue them into a single future which you will spawn in tokio. If you want help with the future fun we can pair on it.

datasource/ethereum/src/block_ingestor.rs
```diff
@@ -0,0 +1,140 @@
...
+}
+
+pub struct BlockIngestor<S: BlockStore, T: BatchTransport + Debug + Clone> {
+    store: Arc<Mutex<S>>,
+    network_name: String,
+    web3_transport: T,
+    ancestor_count: u64,
+    logger: slog::Logger,
+    polling_interval: Duration,
+}
```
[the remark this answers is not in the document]
Jannis: I like the idea. The separation doesn't add much here.

datasource/ethereum/Cargo.toml
```diff
@@ -3,8 +3,10 @@ name = "graph-datasource-ethereum"
 version = "0.1.0"
 
 [dependencies]
+diesel = { version = "1.3.2", features = ["postgres", "serde_json", "numeric"] }
```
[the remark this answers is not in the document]
timmclean: Agreed! Done.

datasource/ethereum/src/block_ingestor.rs
```diff
@@ -0,0 +1,140 @@
...
+
+impl<S: BlockStore + Send + 'static, T: BatchTransport + Debug + Clone> BlockIngestor<S, T> {
+    pub fn spawn(config: BlockIngestorConfig<S, T>) -> Result<(), Error>
```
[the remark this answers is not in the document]
timmclean: I think you're right about this. Changed so that `async_main` calls `new()` and `into_polling_stream()` and then calls `tokio::spawn` itself

datasource/ethereum/src/block_ingestor.rs
```diff
@@ -0,0 +1,140 @@
...
+
+        Ok(())
+    }
+
+    fn into_polling_stream(self) -> impl Future<Item = (), Error = ()> {
+        let err_logger = self.logger.clone();
+
+        tokio::timer::Interval::new(Instant::now(), self.polling_interval)
+            .map_err(Error::from)
+            .map(move |_| {
```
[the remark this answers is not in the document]
timmclean: Ahh that's much better

datasource/ethereum/src/block_ingestor.rs
```diff
@@ -0,0 +1,140 @@
...
+                self.do_poll().unwrap_or_else(|e| {
+                    error!(self.logger, "failed to poll for latest block: {:?}", e);
```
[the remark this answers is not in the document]
timmclean: I like it!

datasource/ethereum/src/block_ingestor.rs
```diff
@@ -0,0 +1,140 @@
...
+        // Add requests to batch
+        let blocks = block_hashes
+            .into_iter()
+            .map(|block_hash| web3.eth().block_with_txs(BlockId::from(*block_hash)))
+            .collect::<Vec<_>>();
+
+        // Submit all requests in batch
+        web3.transport()
+            .submit_batch()
+            .wait()
```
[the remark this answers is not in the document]
timmclean: I'm trying to remember why I did this originally... I think I was stuck in Send/Sync hell for a bit. Anyway, I changed everything to Futures/Streams today! No more `wait()`.
I used `loop_fn` to emulate a while loop.

node/src/main.rs
```diff
@@ -90,6 +91,14 @@ fn async_main() -> impl Future<Item = (), Error = ()> + Send + 'static {
                 .value_name("FILE")
                 .help("Ethereum IPC pipe"),
         )
+        .arg(
```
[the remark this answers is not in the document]
timmclean: I think that's a lot better. Done!

store/postgres/migrations/2018-08-22-140000_create_head_update_procedure/up.sql
timmclean: I was looking to do that originally, but surprisingly the payload has to be a constant string in postgres: https://www.postgresql.org/docs/9.6/static/sql-notify.html

I could workaround that, but I think it's OK -- checking if the head block has changed is quick, and these extra checks would only happen if processing is already caught up

timmclean: (maybe worth putting in a separate issue to investigate as an optimization?)

store/postgres/src/store.rs
timmclean: Added some comments on this. It's the way to run a stored SQL function in diesel (`attempt_head_update` is the stored function).

graph/src/components/store.rs
timmclean: Done!

graph/src/components/store.rs
timmclean: Done!

datasource/ethereum/src/block_ingestor.rs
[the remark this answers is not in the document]
timmclean: Done!

datasource/ethereum/src/block_ingestor.rs
timmclean: Yeah it is hard to follow. I added some comments here. Let me know if it makes more sense now

timmclean:
@leodasvacas 
`wait()`s have been removed :)

We went with the polling approach because we didn't want to rely too heavily on individual Infura nodes. We were worried about the subscription connection dying out without us noticing, etc

store/postgres/migrations/2018-08-22-140000_create_head_update_procedure/up.sql
Jannis: It does take a constant string, but that's only for the channel name. It has a second, optional `payload` parameter. See here ;)

https://github.com/graphprotocol/graph-node/blob/master/store/postgres/migrations/2018-08-16-143800_create_entity_change_triggers/up.sql#L13

datasource/ethereum/src/block_ingestor.rs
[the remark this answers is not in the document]
Jannis: Sweet, what a nice find! (`loop_fn` I mean).

mock/src/store.rs
```diff
@@ -75,6 +78,32 @@ impl BasicStore for MockStore {
     }
 }
 
+impl BlockStore for MockStore {
```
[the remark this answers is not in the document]
Jannis: [Witness the fitness](https://www.youtube.com/watch?v=NDWgtB_MD24) 😀

store/postgres/src/store.rs
Jannis: Ah, cool, thanks.

store/postgres/src/store.rs
```diff
@@ -57,7 +60,7 @@ pub struct Store {
     logger: slog::Logger,
     subscriptions: Arc<RwLock<HashMap<String, Subscription>>>,
     change_listener: EntityChangeListener,
-    pub conn: PgConnection,
+    pub conn: Arc<Mutex<PgConnection>>,
```
Jannis: Curious: Why is this needed?

Jannis: Oh, nevermind. I see you need the collection in the async parts of the `BlockStore` implementation.

leodasvacas:
You need to rustfmt.

mock/src/store.rs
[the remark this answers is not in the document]
leodasvacas: lol what have I just watched

datasource/ethereum/src/block_ingestor.rs
```diff
@@ -0,0 +1,178 @@
...
+        let static_self: &'static _ = Box::leak(Box::new(self));
+
+        // Create stream that emits at polling interval
+        tokio::timer::Interval::new(Instant::now(), static_self.polling_interval)
+            .map_err(move |e| {
+                error!(static_self.logger, "timer::Interval failed: {:?}", e);
+            })
+            .for_each(move |_| {
+                // Attempt to poll
+                static_self.do_poll().then(move |result| {
```
leodasvacas: This `then` can be refactored as a `map_err`.

datasource/ethereum/src/block_ingestor.rs
```diff
@@ -0,0 +1,178 @@
...
+                    .map_err(|e| format_err!("could not get block from Ethereum: {}", e))
+            })
+            .collect::<Vec<_>>();
+
+        // Submit all requests in batch
+        Box::new(
+            web3.transport()
+                .submit_batch()
+                .map_err(|e| format_err!("could not get blocks from Ethereum: {}", e))
+                .map(|_| stream::futures_unordered(block_futures))
```
leodasvacas: Nice way to turn futures into a stream 👍 .

datasource/ethereum/src/block_ingestor.rs
```diff
@@ -0,0 +1,178 @@
...
+
+        // Add requests to batch
+        let block_futures = block_hashes
+            .into_iter()
+            .map(|block_hash| {
+                web3.eth()
+                    .block_with_txs(BlockId::from(*block_hash))
+                    .map_err(|e| format_err!("could not get block from Ethereum: {}", e))
+            })
+            .collect::<Vec<_>>();
```
leodasvacas: This `collect` can just be removed I think.

datasource/ethereum/src/block_ingestor.rs
```diff
@@ -0,0 +1,178 @@
...
+            })
+    }
+
+    fn do_poll<'a>(&'a self) -> impl Future<Item = (), Error = Error> + 'a {
+        // Ask for latest block from Ethereum node
+        self.get_latest_block()
+            .and_then(move |latest_block: Block<Transaction>| {
+                // Store latest block in block store.
+                // Might be a no-op if latest block is one that we have seen.
+                // ingest_blocks will return a (potentially incomplete) list of blocks that are
```
leodasvacas: When would it be incomplete? Why would blocks be missing? It'd good to give some background for readers who have limited knowledge of how the ethereum chain works (such as myself).

datasource/ethereum/src/block_ingestor.rs
```diff
@@ -0,0 +1,178 @@
...
+                    move |missing_block_hashes| -> Box<Future<Item = _, Error = _> + Send> {
+                        if missing_block_hashes.is_empty() {
+                            // If no blocks were missing, then the block head pointer was updated
+                            // successfully, and this poll has completed.
+                            Box::new(future::ok(future::Loop::Break(())))
+                        } else {
+                            // Some blocks are missing: load them, ingest them, and repeat.
+                            let missing_blocks = self.get_blocks(&missing_block_hashes);
+                            Box::new(self.ingest_blocks(missing_blocks).map(
+                                |missing_block_hashes| future::Loop::Continue(missing_block_hashes),
```
leodasvacas: These new comments are great at describing what is being done but I still don't get the why. My previous doubt still stands, in what situation would this call to `ingest_blocks` update the head pointer? Can this loop forever or is there a maximum number of iterations?

datasource/ethereum/src/block_ingestor.rs
```diff
@@ -0,0 +1,178 @@
...
+                        );
+                    }
+
+                    // Continue polling even if polling failed
+                    future::ok(())
+                })
+            })
+    }
+
+    fn do_poll<'a>(&'a self) -> impl Future<Item = (), Error = Error> + 'a {
```
leodasvacas: > We went with the polling approach because we didn't want to rely too heavily on individual Infura nodes.

Ah nice, could you put a note about that in a comment (here would probably be a good spot) so that nobody is tempted to change that in the future?

datasource/ethereum/src/block_ingestor.rs
```diff
@@ -0,0 +1,178 @@
...
+            web3_transport,
+            ancestor_count,
+            logger: logger.new(o!("component" => "BlockIngestor")),
+            polling_interval,
+        })
+    }
+
+    pub fn into_polling_stream(self) -> impl Future<Item = (), Error = ()> {
+        // Currently, there is no way to stop block ingestion, so just leak self
+        let static_self: &'static _ = Box::leak(Box::new(self));
```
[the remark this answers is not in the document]
timmclean: mem::forget does something a little different. It takes ownership and gets rid of the value. Here I'm trying to keep the value around so I can use it

I tried and failed to come up with a solution involving move... the issue is that the lifetime of a reference to self needs to outlive the lifetimes of the futures that are created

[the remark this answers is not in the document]
leodasvacas: Can't the `for_each` closure below move in `self`?

datasource/ethereum/src/block_ingestor.rs
leodasvacas: Ah, weird API, if that's the case then could add a note explaining that need for `collect`?

datasource/ethereum/src/block_ingestor.rs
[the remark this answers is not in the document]
timmclean: I got an error about conflicting lifetime requirements with that. If I move self into the closure, then I can only borrow it into the closure function body ends, but I'm returning a future whose lifetime depends on &self, so the borrow needs to last longer

node/src/main.rs
```diff
@@ -253,3 +274,35 @@ fn async_main() -> impl Future<Item = (), Error = ()> + Send + 'static {
...
+    C: FnOnce(&str) -> (EventLoopHandle, Transport),
+{
+    // Check for common mistakes
+    if s.starts_with("wss://") || s.starts_with("http://") || s.starts_with("https://") {
+        warn!(logger, "Is your Ethereum connection string missing a network name? Try 'mainnet:' + the connection URL.");
+    }
+
+    // Parse string (format is "network_name:url_or_path")
+    let split_at = s.find(':').expect(
+        "A network name must be provided alongside the Ethereum node location. Try 'mainnet:URL'.",
```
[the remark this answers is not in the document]
Jannis: Oh, yeah, fair enough.

datasource/ethereum/src/block_ingestor.rs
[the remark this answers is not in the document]
leodasvacas: Ah, yhea, complicated situation. I don't mind the current solution.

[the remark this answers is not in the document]
Jannis: Me neither. Thanks for explaining the situation. 😄

node/src/main.rs
```diff
@@ -71,26 +72,34 @@ fn async_main() -> impl Future<Item = (), Error = ()> + Send + 'static {
...
                 .required_unless_one(&["ethereum-rpc", "ethereum-ws"])
                 .conflicts_with_all(&["ethereum-rpc", "ethereum-ws"])
                 .long("ethereum-ipc")
-                .value_name("FILE")
-                .help("Ethereum IPC pipe"),
+                .value_name("NETWORK_NAME:FILE")
+                .help("Ethereum network name (e.g. 'mainnet') and Ethereum IPC pipe path, separated by a ':'"),
+        )
+        .arg(
+            Arg::with_name("ethereum-network-name")
```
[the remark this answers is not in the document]
timmclean: Done!

datasource/ethereum/src/block_ingestor.rs
timmclean: Hmm does that work? I want to change Err into Ok here to prevent polling from stopping

leodasvacas: Ah yes, nevermind then.

mock/src/store.rs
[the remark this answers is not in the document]
timmclean: yessssssss

store/postgres/migrations/2018-08-22-140000_create_head_update_procedure/up.sql
timmclean: `pg_notify` did the trick. Done in the tim/block-processing branch: https://github.com/graphprotocol/graph-node/commit/9a7fc2f1df8815a017fda5f7a104732e0b78b590#diff-e69c79fa22704b08685b3412c3ba9eabR75

timmclean:
Closing this PR in favour of a new PR that includes all block ingestion & processing work at once

PR #311, opened 2018-08-29
Test Ethereum endpoint with get current block_number call

### LINKED ISSUE #269: Improve Ethereum unavailable error
When the infura url is invalid we get the following error which does not make it obvious that it is a simple setup fix:
```
web3::transports::ws: WebSocketError: Error(Transport("ResponseError(\"Status code must be Switching Protocols\")"), State { next_error: None, backtrace: None})
```

fordN:
Resolves #269 

In the case of an invalid Ethereum endpoint being supplied the application previously did not fail quickly, instead providing RPC errors on each event subscription call.  The continued operation and indirect errors made for inobvious debugging of a simple programmer error. 

Now a test call is made immediately after creating the Ethereum Adapter that gets the current block_number or upon an error stops the application and returns a helpful error.  

Unfortunately Web3 also logs a `WebSocketError` which I could not keep from firing.

### DIFF
--- a/node/src/main.rs
+++ b/node/src/main.rs
@@ -172,6 +172,16 @@ fn async_main() -> impl Future<Item = (), Error = ()> + Send + 'static {
     let ethereum_watcher = graph_datasource_ethereum::EthereumAdapter::new(
         graph_datasource_ethereum::EthereumAdapterConfig { transport },
     );
+
+    match ethereum_watcher.block_number().wait() {
+        Ok(number) => info!(logger, "Connected to Ethereum node";
+                            "most_recent_block" => &number.to_string()),
+        Err(e) => {
+            error!(logger, "Was a valid Ethereum node endpoint provided?");
+            panic!("Failed to connect to Ethereum node: {}", e);
+        }
+    }
+
     let runtime_host_builder =
         WASMRuntimeHostBuilder::new(&logger, Arc::new(Mutex::new(ethereum_watcher)), resolver);
     let runtime_manager =


node/src/main.rs
```diff
@@ -172,6 +172,17 @@ fn async_main() -> impl Future<Item = (), Error = ()> + Send + 'static {
     let ethereum_watcher = graph_datasource_ethereum::EthereumAdapter::new(
         graph_datasource_ethereum::EthereumAdapterConfig { transport },
     );
+
+    match ethereum_watcher.block_number().wait() {
+        Ok(number) => {
+            info!(logger, "Connected to Ethereum endpoint"; "block_number" => &number.as_u64())
```
leodasvacas: It's not clear to the user (and also not to me)  what `block_number` is. Are you sure that it always fits in an `u64`? Probably `to_string()` would be better than `as_u64()`.

fordN: Simply logging "Connected to Ethereum node" does ensure clarity.  Given the perceived issues we have been having with the infura node not providing the latest/highest blocks I thought I would provide some context of the current state of the provided Ethereum node.

fordN: Currently the highest block number (as displayed on etherscan.io) is 6236205, I can't imagine it will be hitting 18446744073709551615 anytime soon.

fordN: We should either make the block_number logging more clear or remove it altogether:
`Connected to Ethereum node, current_block_number: {}`,
`Connected to Ethereum node, highest_block_number: {}`,
or simply 
`Connected to Ethereum node`

Votes?

Jannis: Ok, I get that. How about `"most_recent_block" => block_number.to_string()`, as a k/v pair?

Jannis:
I'm merging this since Ford is on a plane right now.

PR #315, opened 2018-08-30
Support updating subgraphs

leodasvacas:
If the admin endpoint `subgraph_add` is called twice with the same subgraph name, it will shutdown the previous subgraph and host the new one.

Open question: Do we want to separate the `update` endpoint or keep it unified with `add`? If they are unified, should it be renamed to `set` or some other name? Opinions welcome, I'd like to settle this before merging. Ping @html5cat you might have an opinion on API design.

The implementation consists of:
- Properly shutting down the host thread when the host is dropped.
- Keeping track of added sugraphs in `SubgraphProvider` and sending a message to the `RuntimeManager` signaling the removal of the old subgraph when a subgraph is overwritten.

The "update" part of #282

### DIFF
--- a/graph/src/components/subgraph/provider.rs
+++ b/graph/src/components/subgraph/provider.rs
@@ -8,8 +8,8 @@ use tokio::prelude::*;
 pub enum SubgraphProviderEvent {
     /// A subgraph was added to the provider.
     SubgraphAdded(SubgraphManifest),
-    /// A subgraph was removed from the provider.
-    SubgraphRemoved(SubgraphManifest),
+    /// A subgraph with the given id was removed from the provider.
+    SubgraphRemoved(String),
 }
 
 /// Schema-only events emitted by a [SubgraphProvider](trait.SubgraphProvider.html).
--- a/graph/src/components/subgraph/provider.rs
+++ b/graph/src/components/subgraph/provider.rs
@@ -8,7 +8,7 @@ use tokio::prelude::*;
 pub enum SubgraphProviderEvent {
     /// A subgraph was added to the provider.
     SubgraphAdded(SubgraphManifest),
-    /// A subgraph with the given id was removed from the provider.
+    /// A subgraph with the given ID was removed from the provider.
     SubgraphRemoved(String),
 }
 
--- a/runtime/wasm/src/host.rs
+++ b/runtime/wasm/src/host.rs
@@ -1,5 +1,6 @@
 use ethereum_types::Address;
 use futures::sync::mpsc::{channel, Receiver};
+use futures::sync::oneshot;
 use std::str::FromStr;
 use std::sync::Mutex;
 use std::thread;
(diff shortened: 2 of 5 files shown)


html5cat:
@leodasvacas what's the current API for `update`/`add`/`set`? I can't find the relevant docs.

leodasvacas:
@html5cat we're lacking docs, but the api is, if you want to add (and now update) a subgraph either POST with no body to:
`http://host:8020/subgraph_add/name/hash`

Or POST to:
`http://host:8020/`
With body:
```json
{
"jsonrpc":"2.0",
"method":"subgraph_add",
"params": {"name":"name","ipfs_hash":"hash"},
"id":"something"
}
```
The `8020` port is the default but can be configured. In both cases the header `Content-type: application/json` is mandatory. My question is whether `add` should be called something else in this context.

graph/src/components/subgraph/provider.rs
```diff
@@ -8,8 +8,8 @@ use tokio::prelude::*;
 pub enum SubgraphProviderEvent {
     /// A subgraph was added to the provider.
     SubgraphAdded(SubgraphManifest),
-    /// A subgraph was removed from the provider.
-    SubgraphRemoved(SubgraphManifest),
+    /// A subgraph with the given id was removed from the provider.
```
Jannis: Nitpick: In documentation comments, I'd always capitalize `ID`.

runtime/wasm/src/host.rs
```diff
@@ -94,7 +97,11 @@ impl RuntimeHost {
             link_resolver: link_resolver.clone(),
         };
 
-        info!(logger, "Loading WASM runtime"; "data_source" => &config.data_source.name);
+        let name = config.data_source.name.clone();
+        info!(logger, "Loading WASM runtime"; "data_source" => &name);
```
Jannis: Is the `clone()` above necessary? Can we not pass `&config.data_source.name` to `info!`?

runtime/wasm/src/host.rs
```diff
@@ -106,15 +113,25 @@ impl RuntimeHost {
             let module = WasmiModule::new(&logger, wasmi_config);
 
             // Process one event at a time, blocking the thread when waiting for
-            // the next event.
+            // the next event. Also check for a cancelation signal.
             Self::subscribe_to_events(&logger, data_source, module, ethereum_adapter)
+                .select(
```
Jannis: Yes, our first use (I believe) of `select`. Select is beautiful.

leodasvacas: There's another use in this file, when we select among all ethereum event subscriptions. Beautiful it is.

runtime/wasm/src/host.rs
leodasvacas: It's not clear but it's because we move it into the thread so we can also log the name on shutdown.

Jannis: Ah, good.

PR #320, opened 2018-09-03
Add GraphQL subscription support

### LINKED ISSUE #275: Add subscription support to GraphQL (execution and resolvers)
This includes:

1. Adding an `execute_subscription` function that implements the [subscribe algorithm from the GraphQL spec](http://facebook.github.io/graphql/draft/#Subscribe()).
2. Adding stream resolution to `Resolver` and `StoreResolver`.

### LINKED ISSUE #276: Rename QueryRunner to GraphQLRunner and add support for subscriptions
The title says it all: we'll want to also use the `QueryRunner` for handling subscriptions, so it should have a more general name. Also, we'll need it to execute subscriptions and return the streams back to the caller (in this case, the subscription server).

### LINKED ISSUE #277: Add GraphQLSubscriptionServer trait and WS-based implementation
This includes:

1. Adding the trait.
2. Adding a new `server/ws-subscriptions` (or similar) crate.
3. Implementing the [GraphQL over WebSocket protocol](https://github.com/apollographql/subscriptions-transport-ws/blob/master/PROTOCOL.md).
4. Handling the lifetime of subscriptions, requesting result streams and terminating the streams cleanly when connections are dropped.

### LINKED ISSUE #287: Generate root Subscription type in the GraphQL API schema

Jannis:
Resolves #275, resolves #276, resolves #277, resolves #287.

This adds support for GraphQL subscriptions (tested in GraphiQL all evening) by

1. generating a root `Subscription` type as part of the GraphQL API schema,
2. adding an `graph_graphql::execute_subscription` function as the entry point to executing subscriptions and obtaining an asynchronous result stream for them,
3. adding a `GraphQLRunner::run_subscription` method for other components to run subscriptions,
4. adding a `SubscriptionServer` component trait to the `graph` crate,
5. adding a `graph-server-websocket` crate with an implementation of `SubscriptionServer` based on [tokio-tungstenite](https://github.com/snapview/tokio-tungstenite/).

There are a couple of open questions:

1. <s>How do we best terminate running subscriptions and entire connections (maybe with a `oneshot::channel` and `select`) when receiving `stop` or `connection_terminate` messages.</s> Connections and running subscriptions are now terminated cleanly on either `GQL_STOP` or `GQL_CONNECTION_TERMINATE`.
2. Can we break up the pretty complicated `SubscriptionServer::into_future` and `GraphQlConnection::handle_incoming_messages` functions, or maybe even the entire operation management in `GraphQlConnection`?

Suggestions welcome!


Jannis:
The oneshot channel shutdown trick is also employed in the [tk-listen docs](https://docs.rs/tk-listen/0.2.0/tk_listen/#example-with-listener-shutdown).

server/websocket/src/server.rs
```diff
@@ -0,0 +1,174 @@
...
+        path.iter()
+            .nth(1)
+            .and_then(|os| os.to_str())
+            .map(|s| s.into())
+    }
+}
+
+impl<Q> SubscriptionServerTrait for SubscriptionServer<Q>
+where
+    Q: GraphQLRunner + Sized + 'static,
```
leodasvacas: Remove `Sized`.

server/websocket/src/server.rs
```diff
@@ -0,0 +1,174 @@
...
+                })
+            });
+
+        Ok(Box::new(task))
+    }
+}
+
+impl<Q> EventConsumer<SchemaEvent> for SubscriptionServer<Q>
+where
+    Q: GraphQLRunner + Sized + 'static,
```
leodasvacas: Remove `Sized`.

store/postgres/src/entity_changes.rs
```diff
@@ -65,53 +62,46 @@ impl EntityChangeListener {
...
+                // Read notifications until there hasn't been one for 500ms
+                for notification in notifications
+                    .timeout_iter(Duration::from_millis(500))
+                    .iterator()
+                    .filter_map(Result::ok)
+                    .filter(|notification| notification.channel == "entity_changes")
+                {
+                    // Terminate the thread if desired
+                    if terminate.load(Ordering::SeqCst) {
+                        return;
```
leodasvacas: This can be just a `break` I think.

leoyvens (changes requested):
That's a ton of good stuff, amazing that you got it all done over a weekend!

.travis.yml
```diff
@@ -57,4 +57,4 @@ script:
       cargo fmt --all -- --check;
     fi
   # Run unit and sccenario tests
-  - RUST_BACKTRACE=1 cargo test --verbose --all -- --nocapture
+  - RUST_BACKTRACE=1 cargo test --verbose --all -- --nocapture --test-threads=1
```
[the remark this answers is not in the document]
Jannis: We can try with that mutex again. Let me push a version to a separate branch and see what Travis says.

server/websocket/src/connection.rs
```diff
@@ -0,0 +1,340 @@
...
+}
+
+impl From<OutgoingMessage> for WsMessage {
+    fn from(msg: OutgoingMessage) -> Self {
+        WsMessage::text(serde_json::to_string(&msg).expect("invalid GraphQL/WebSocket message"))
+    }
+}
+
+/// A WebSocket connection implementing the GraphQL over WebSocket protocol.
+pub struct GraphQlConnection<Q, S> {
```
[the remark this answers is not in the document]
Jannis: Yep, I'm slowly adjusting to idiomatic Rust naming conventions...

I like your suggestion and have tracked it here: https://github.com/graphprotocol/graph-node/issues/324.

graphql/src/store/query.rs
```diff
@@ -141,15 +138,67 @@ pub fn build_subgraph_id(entity: &schema::ObjectType) -> Option<String> {
...
+        // Check if the field exists on the object type
+        if let Some(field_type) = sast::get_field_type(object_type, &field.name) {
+            // Check if the field type corresponds to a type definition (in a valid schema,
+            // this should always be the case)
+            if let Some(type_definition) =
+                sast::get_type_definition_from_field_type(schema, field_type)
+            {
+                // Check if the field's type is that of an entity
+                //
+                // FIXME: We must check for the `@entity` directive here once we have that,
```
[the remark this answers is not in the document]
Jannis: The FIXME is resolved in the follow-up PR (#322).

graphql/src/store/query.rs
```diff
@@ -141,15 +138,67 @@ pub fn build_subgraph_id(entity: &schema::ObjectType) -> Option<String> {
...
+                // otherwise the check below will also be true for the root Query and
+                // Subscription objects as well as all filter object types
+                if let s::TypeDefinition::Object(object_type) = type_definition {
+                    // Obtain the subgraph ID from the object type
+                    if let Some(subgraph_id) = parse_subgraph_id(object_type) {
```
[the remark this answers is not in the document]
Jannis: That, again, is something that should be checked earlier. It means our way of adding subgraph IDs to the schema types in schema is broken.

graphql/src/subscription/mod.rs
```diff
@@ -0,0 +1,178 @@
...
+        None,
+    );
+
+    if grouped_field_set.len() != 1 {
+        return Err(SubscriptionError::from(
+            QueryExecutionError::MultipleSubscriptionFields,
+        ));
+    }
+
+    let fields = grouped_field_set.get_index(0).unwrap();
```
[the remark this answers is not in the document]
Jannis: I've improved the check and am returning two different errors (empty query vs. multiple subscription query fields) now. This is no longer being reached if the selection set is empty.

graphql/src/subscription/mod.rs
```diff
@@ -0,0 +1,178 @@
...
+    let field = fields.1[0];
+    let argument_values = coerce_argument_values(ctx.clone(), subscription_type, field)
+        .map_err(SubscriptionError::from)?;
+
+    resolve_field_stream(ctx, subscription_type, field, argument_values)
+}
+
+fn resolve_field_stream<'a, R1, R2>(
```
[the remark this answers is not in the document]
Jannis: I'd rather not because it corresponds 1:1 to https://facebook.github.io/graphql/draft/#ResolveFieldEventStream().

server/websocket/src/connection.rs
```diff
@@ -0,0 +1,340 @@
...
+    id: String,
+    logger: Logger,
+    graphql_runner: Arc<Q>,
+    stream: WebSocketStream<S>,
+    subgraphs: Arc<RwLock<SubgraphRegistry<Schema>>>,
```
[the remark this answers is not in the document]
Jannis: Pushed that. Turned out nicely, thanks for the suggestion!

server/websocket/src/connection.rs
```diff
@@ -0,0 +1,340 @@
...
+            .inspect(move |msg| {
+                debug!(logger, "Sending message";
+                       "connection" => &id,
+                       "msg" => format!("{}", msg).as_str());
+            })
+            .forward(ws_sink.sink_map_err(|_| ()));
+
+        // Silently swallow internal send results and errors
+        let ws_writer = ws_writer.map(|_| ()).map_err(|_| ());
+        let ws_reader = ws_reader.map(|_| ()).map_err(|_| ());
```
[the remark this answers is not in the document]
Jannis: I'd say no, unless these errors carry any meaningful information for us to debug?

Errors here will cause the connection to close and typically, the client would initiate a new connection (that's what GraphiQL does; I'm almost sure Apollo does the same). So from the client's perspective this isn't too bad.

store/postgres/src/entity_changes.rs
Jannis: Does `break` terminate outer loops as well? Notice we have
```rust
while ... {
  for ... {
    if terminate.load(Ordering::SeqCst) {
      return;
    }
  }
}
```
here. If `break` would terminate both then that would be ok. If not, I'd say the other `break` further down should also be changed to a `return`.

graph/src/data/query/result.rs
```diff
@@ -4,10 +4,19 @@ use serde::ser::*;
...
+    S: Serializer,
+{
+    SerializableValue(data.as_ref().unwrap_or(&q::Value::Null)).serialize(serializer)
+}
+
 /// The result of running a query, if successful.
-#[derive(Debug)]
+#[derive(Debug, Serialize)]
 pub struct QueryResult {
+    #[serde(skip_serializing_if = "Option::is_none", serialize_with = "serialize_data")]
```
leodasvacas: Ah cool, serde has an attribute for everything it seems.

store/postgres/src/entity_changes.rs
leodasvacas: The `break` will terminate the inner `for` and then the outer `while` will terminate because `terminate == true`.

store/postgres/src/entity_changes.rs
```diff
@@ -65,53 +62,46 @@ impl EntityChangeListener {
...
+                            "Invalid entity change received from the database: {:?}",
+                            value
+                        ).as_str(),
+                    );
+
+                    // We'll assume here that if sending fails, this means that the
+                    // entity change listener has already been dropped, the receiving
+                    // is gone and we should terminate the listener loop
+                    if sender.clone().send(change).wait().is_err() {
+                        break;
```
leodasvacas: But you're right, this break looks strange. Since we now set `terminate` upon dropping the `EntityChangeListener`, then this is redundant since we'll break out anyways at the `while` check.

server/websocket/src/connection.rs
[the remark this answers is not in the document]
leodasvacas: Alright, then you can note in a comment what an error there means and why we don't care.

store/postgres/src/entity_changes.rs
Jannis: Oh. Yes, indeed. The other `break` should still be a return though.

graph/src/data/query/result.rs
Jannis: Yup 😄

store/postgres/src/entity_changes.rs
Jannis: I think we still want to `break` here to avoid reading any more notifications. But `break`ing is enough after all, since the terminate check will kick in right away.

store/postgres/src/entity_changes.rs
Jannis: Or not. See the other comment. Marking this one as resolved.

leoyvens (approved):

Issue #279, opened 2018-08-17
GraphQL subscriptions


Jannis:
- [x] Add entity change triggers / notification channel to Postgres https://github.com/graphprotocol/graph-node/issues/273
- [x] Generate root Subscription type in the GraphQL API schema https://github.com/graphprotocol/graph-node/issues/287
- [x] Add subscription functionality to the Store https://github.com/graphprotocol/graph-node/issues/274
- [x] Rename QueryRunner to GraphQLRunner and add support for subscriptions https://github.com/graphprotocol/graph-node/issues/276
- [x] Add subscription support to GraphQL (execution and resolvers) https://github.com/graphprotocol/graph-node/issues/275 
- [x] Add GraphQLSubscriptionServer trait and WS-based implementation https://github.com/graphprotocol/graph-node/issues/277
- [x] Implement `@entity` directive for entity types in GraphQL schema https://github.com/graphprotocol/graph-node/issues/236

PR #322, opened 2018-09-03
Add schema validation, require @entity in schema types

### LINKED ISSUE #236: Implement @entity directive for entity types in GraphQL schema
We currently don't parse the subgraph GraphQL schema's according to what's in the spec here: https://github.com/graphprotocol/specs/tree/master/graphql-api.

There should be an `@entity` directive required for entity types. I'm labeling this for the DX milestone, since this will be a breaking API change and we should make it soon.

Jannis:
Resolves #236. Depends on #320. Last puzzle piece in finishing #279.

This makes `@entity` mandatory on all `type` definitions in subgraph schemas. If we later want to distinguish between embeddable/inline types that are not represented as entities, we can relax this requirement again. (In that case we should e.g. verify that the inline types are actually referenced anywhere.)

Subgraph schemas are now validated (with the above being checked) when subgraphs are deployed. If a subgraph schema fails to validate, the subgraph is rejected.

Now that it's mandator, we can use the `@entity` directive to properly collect entity types involved in subscription queries, and only subscribe to entity change events from the store for these entities. For example, the subscription
```graphql
subscription { memes { id } }
```
would result in a store subscription for `[(<subgraph ID>, Meme)]`. The subscription
```graphql
subscription { memes { id regEntry_creator { id } } }
```
would result in a store subscription for `[(<subgraph ID>, Meme), (<subgraph ID>, User)]`.

### DIFF
--- a/core/src/subgraph/provider.rs
+++ b/core/src/subgraph/provider.rs
@@ -62,10 +62,8 @@ impl<L: LinkResolver> SubgraphProviderTrait for SubgraphProvider<L> {
                 .and_then(
                     // Validate the subgraph schema before deploying the subgraph
                     |subgraph| match validate_schema(&subgraph.schema.document) {
-                        Err(e) => future::err::<_, SubgraphProviderError>(
-                            SubgraphProviderError::SchemaValidationError(e),
-                        ),
-                        _ => future::ok::<_, _>(subgraph),
+                        Err(e) => Err(SubgraphProviderError::SchemaValidationError(e)),
+                        _ => Ok(subgraph),
                     },
                 )
                 .and_then(move |mut subgraph| {
--- a/core/src/subgraph/provider.rs
+++ b/core/src/subgraph/provider.rs
@@ -5,6 +5,7 @@ use futures::sync::mpsc::{channel, Receiver, Sender};
 
 use graph::data::subgraph::SubgraphProviderError;
 use graph::prelude::{SubgraphProvider as SubgraphProviderTrait, *};
+use graph_graphql::prelude::validate_schema;
 
 pub struct SubgraphProvider<L> {
     _logger: slog::Logger,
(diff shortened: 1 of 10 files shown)


Jannis:
Note: We should update the docs as part of this, to make sure people are aware of the introduction of `@entity`.

leodasvacas:
In bad need of a rebase.

Jannis:
@leodasvacas Done. Care to take a look?

core/src/subgraph/provider.rs
```diff
@@ -58,6 +59,15 @@ impl<L: LinkResolver> SubgraphProviderTrait for SubgraphProvider<L> {
         Box::new(
             SubgraphManifest::resolve(name.clone(), Link { link }, self.resolver.clone())
                 .map_err(SubgraphProviderError::ResolveError)
+                .and_then(
+                    // Validate the subgraph schema before deploying the subgraph
+                    |subgraph| match validate_schema(&subgraph.schema.document) {
+                        Err(e) => future::err::<_, SubgraphProviderError>(
+                            SubgraphProviderError::SchemaValidationError(e),
+                        ),
+                        _ => future::ok::<_, _>(subgraph),
```
leodasvacas: These turbofishes seems uncessary, also I wonder If we could just use `Err` and `Ok` instead of `err` and `ok`.

leoyvens (approved):
Looks good apart from minor comment. It'd be good to give the team a heads up and also update our adchain and decentraland subgraphs, so nobody gets a bad surprise when trying to demo something at ETHBerlin.

Jannis:
@leodasvacas Totally agree. I'll get PRs ready for all public subgraphs + docs before we merge this.

core/src/subgraph/provider.rs
Jannis: You were right. Updated!

PR #323, opened 2018-09-03
Readme: Update postgres flag in cargo

nenadjaja:


README.md
```diff
@@ -36,19 +36,22 @@ For Ethereum network data you can either run a local node or use Infura.io:
...
 cargo run -p graph-node --release -- \
-  --postgres-url postgresql://localhost:5432/adchain-subgraph \
+  --postgres-url postgresql://<USERNAME>:<PASSWORD>@localhost:5432/adchain-subgraph \
   --ethereum-ws wss://mainnet.infura.io/_ws \
   --ipfs 127.0.0.1:5001 \
-  --subgraph IPFS_HASH
+  --subgraph <IPFS_HASH>
 ```
 
+Try your OS username as `USERNAME` and any password for `PASSWORD`.
```
html5cat: On macOS the default seems to be `postgres` if you install with Homebrew. Not sure about other methods/os

Jannis:
The password is optional at least in my setup (macOS + Postgres installed via Homebrew), we should make it `USERNAME[:PASSWORD]@localhost...`

nenadjaja:
Alright, I'll create another PR, given that this one is merged

README.md
nenadjaja: I see. What worked for you @html5cat ?

leodasvacas: `postgres` did not work for me. I had an `admin` user but I can't remember if it's default or if I created it at some point.

nenadjaja: I connected to postgres and listed all the databases. The one I created was under `nevena` (I just ran `createdb example-subgraph`), so I assume it's a default option. But maybe it depends on how you install Postgres

html5cat: Username `postgres` worked for me

PR #334, opened 2018-09-07
Simple auth

### LINKED ISSUE #304: Authorization for subgraph deployment
A design @yanivtal proposed is to have a token required to deploy subgraphs, and upon adding a subgraph return the token that authorizes update and removal of the subgraph.

leodasvacas:
If the `GRAPH_MASTER_TOKEN` is not set, no auth will be checked. If it is set, then first to authorize the deployment of a subgraph the `subgraph_authorize` endpoint must be called with a `name` parameter which will return a fresh token that must be sent along any deploy or remove requests for a subgraph with that name. All auth tokens must be provided with the header `Authorization: Bearer TOKEN`.

Right now removal by id doesn't make sense since we need the name to authorize, but it's still supported by the rest of the codebase since we don't yet have a final design for auth.

Resolves #304 at least for now.

### DIFF
--- a/server/json-rpc/src/lib.rs
+++ b/server/json-rpc/src/lib.rs
@@ -73,7 +73,7 @@ pub struct JsonRpcServer<T> {
     provider: Arc<T>,
     logger: Logger,
     // Maps auth tokens to authorized subgraph name.
-    authorized: Arc<RwLock<BTreeMap<String, String>>>,
+    subgraph_api_keys: Arc<RwLock<BTreeMap<String, String>>>,
 }
 
 impl<T: SubgraphProvider> JsonRpcServer<T> {
@@ -86,11 +86,11 @@ impl<T: SubgraphProvider> JsonRpcServer<T> {
         info!(self.logger, "Received subgraph_deploy request"; "params" => params.to_string());
 
         if should_check_auth()
-            && Some(&auth.bearer_token) != self.authorized.read().unwrap().get(&params.name)
+            && Some(&auth.bearer_token) != self.subgraph_api_keys.read().unwrap().get(&params.name)
         {
             return Box::new(future::err(json_rpc_error(
                 JSON_RPC_UNAUTHORIZED_ERROR,
-                "Auth token is invalid".to_owned(),
+                "API key is invalid".to_owned(),
             )));
         }
 
@@ -111,18 +111,21 @@ impl<T: SubgraphProvider> JsonRpcServer<T> {
     ) -> Box<Future<Item = Value, Error = jsonrpc_core::Error> + Send> {
         info!(self.logger, "Received subgraph_remove request"; "params" => params.to_string());
 
+        let name_or_id = params.name_or_id;
+        // We need a name for auth so `name_or_id` being an id is not supported
+        // if we're checking auth.
         if should_check_auth()
-            && Some(&auth.bearer_token) != self.authorized.read().unwrap().get(&params.name_or_id)
+            && Some(&auth.bearer_token) != self.subgraph_api_keys.read().unwrap().get(&name_or_id)
         {
             return Box::new(future::err(json_rpc_error(
                 JSON_RPC_UNAUTHORIZED_ERROR,
-                "Auth token is invalid".to_owned(),
+                "API key is invalid".to_owned(),
             )));
         }
 
         Box::new(
             self.provider
-                .remove(params.name_or_id)
+                .remove(name_or_id)
                 .map_err(|e| json_rpc_error(JSON_RPC_REMOVE_ERROR, e.to_string()))
                 .map(|_| Ok(Value::Null))
                 .flatten(),
@@ -131,7 +134,7 @@ impl<T: SubgraphProvider> JsonRpcServer<T> {
 
     /// Handler for the `subgraph_authorize` endpoint.
     ///
-    /// Takens subgraph name and returns an auth token that can be used to
+    /// Taken subgraph name and returns an API key that can be used to
     /// add/remove subgraphs under the input name.
     ///
     /// Requires bearer authorization with the master token.
@@ -148,7 +151,7 @@ impl<T: SubgraphProvider> JsonRpcServer<T> {
             Ok(_) => {
                 return Err(json_rpc_error(
                     JSON_RPC_UNAUTHORIZED_ERROR,
-                    "auth token is invalid".to_owned(),
+                    "authorization token is invalid".to_owned(),
                 ))
             }
             Err(_) => {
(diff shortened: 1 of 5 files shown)


server/json-rpc/src/lib.rs
```diff
@@ -37,86 +51,221 @@ impl fmt::Display for SubgraphRemoveParams {
...
+    bearer_token: String,
+}
+
+impl Metadata for AuthorizationHeader {}
+
+pub struct JsonRpcServer<T> {
+    provider: Arc<T>,
+    logger: Logger,
+    // Maps auth tokens to authorized subgraph name.
+    authorized: Arc<RwLock<BTreeMap<String, String>>>,
```
Jannis: Can we call this `subgraph_api_keys` to make its purpose a bit more obvious?

server/json-rpc/src/lib.rs
```diff
@@ -37,86 +51,221 @@ impl fmt::Display for SubgraphRemoveParams {
...
+        auth: AuthorizationHeader,
+    ) -> Box<Future<Item = Value, Error = jsonrpc_core::Error> + Send> {
+        info!(self.logger, "Received subgraph_deploy request"; "params" => params.to_string());
+
+        if should_check_auth()
+            && Some(&auth.bearer_token) != self.authorized.read().unwrap().get(&params.name)
+        {
+            return Box::new(future::err(json_rpc_error(
+                JSON_RPC_UNAUTHORIZED_ERROR,
+                "Auth token is invalid".to_owned(),
```
Jannis: Let's use "API key is invalid" here.

server/json-rpc/src/lib.rs
```diff
@@ -37,86 +51,221 @@ impl fmt::Display for SubgraphRemoveParams {
...
+        auth: AuthorizationHeader,
+    ) -> Box<Future<Item = Value, Error = jsonrpc_core::Error> + Send> {
+        info!(self.logger, "Received subgraph_remove request"; "params" => params.to_string());
+
+        if should_check_auth()
+            && Some(&auth.bearer_token) != self.authorized.read().unwrap().get(&params.name_or_id)
+        {
+            return Box::new(future::err(json_rpc_error(
+                JSON_RPC_UNAUTHORIZED_ERROR,
+                "Auth token is invalid".to_owned(),
```
Jannis: Same here.

server/json-rpc/src/lib.rs
```diff
@@ -37,86 +51,221 @@ impl fmt::Display for SubgraphRemoveParams {
...
+                .remove(params.name_or_id)
+                .map_err(|e| json_rpc_error(JSON_RPC_REMOVE_ERROR, e.to_string()))
+                .map(|_| Ok(Value::Null))
+                .flatten(),
+        )
+    }
+
+    /// Handler for the `subgraph_authorize` endpoint.
+    ///
+    /// Takens subgraph name and returns an auth token that can be used to
```
Jannis: `Takens` -> `Takes`

Jannis: And `auth token` -> `API key`

server/json-rpc/src/lib.rs
```diff
@@ -37,86 +51,221 @@ impl fmt::Display for SubgraphRemoveParams {
...
+    ) -> Result<Value, jsonrpc_core::Error> {
+        info!(self.logger, "Received subgraph_authorize request"; "params" => params.to_string());
+
+        let master_token = env::var(GRAPH_MASTER_TOKEN_VAR);
+        match &master_token {
+            Ok(master_token) if *master_token == auth.bearer_token => (), // Authorized.
+            Ok(_) => {
+                return Err(json_rpc_error(
+                    JSON_RPC_UNAUTHORIZED_ERROR,
+                    "auth token is invalid".to_owned(),
```
Jannis: I guess in this case "auth token" is the right term, although I'd make it `Authorization token is invalid`.

server/json-rpc/src/lib.rs
```diff
@@ -37,86 +51,221 @@ impl fmt::Display for SubgraphRemoveParams {
...
+                ))
+            }
+            Err(_) => {
+                return Err(json_rpc_error(
+                    JSON_RPC_UNAUTHORIZED_ERROR,
+                    "internal error".to_owned(),
+                ))
+            }
+        }
+        // Generate a random token of 50 ASCII alhpanumerics.
```
Jannis: Could also generate UUIDs (v4) here but this is simple enough.

Typo here though: `alhpanumerics` -> `alphanumerics`.

server/json-rpc/src/lib.rs
```diff
@@ -37,86 +51,221 @@ impl fmt::Display for SubgraphRemoveParams {
...
+        });
         // `subgraph_deploy` handler.
-        let deploy_provider = provider.clone();
-        let deploy_logger = logger.clone();
-        let deploy_handler = move |params| {
-            Self::deploy_handler(params, deploy_provider.clone(), deploy_logger.clone())
-        };
-        handler.add_method("subgraph_deploy", move |params: Params| {
-            let deploy_handler = deploy_handler.clone();
+        let me = arc_self.clone();
```
Jannis: `me` -> `deploy_self` maybe?

Jannis: Or `self_deploy`?

server/json-rpc/src/lib.rs
```diff
@@ -37,86 +51,221 @@ impl fmt::Display for SubgraphRemoveParams {
...
 
         // `subgraph_remove` handler.
-        let remove_provider = provider.clone();
-        let remove_logger = logger.clone();
-        let remove_handler = move |params| {
-            Self::remove_handler(params, remove_provider.clone(), remove_logger.clone())
-        };
-        handler.add_method("subgraph_remove", move |params: Params| {
-            let remove_handler = remove_handler.clone();
+        let me = arc_self.clone();
```
Jannis: Same here?

server/json-rpc/src/lib.rs
```diff
@@ -37,86 +51,221 @@ impl fmt::Display for SubgraphRemoveParams {
...
             params
                 .parse()
                 .into_future()
-                .and_then(move |params| remove_handler(params))
+                .and_then(move |params| me.remove_handler(params, auth))
         });
 
-        ServerBuilder::new(handler)
+        // `subgraph_authorize` handler.
+        let me = arc_self.clone();
```
Jannis: etc.

server/json-rpc/src/lib.rs
leodasvacas: I just went with a name that could be used on all handlers so that their code looked as similar as possible. I think it looks nice and symmetrical, but if you think different names is better then your suggestion is good.

Jannis: Ok, yeah, I get that. 👍

leodasvacas:
@Jannis review adressed.

PR #359, opened 2018-09-12
travis: Don't cache

leodasvacas:
Rust build outputs are very large, uploading and downloading them is taking 3 minutes on each build, and for nothing because our cache uploading always times out. We could increase the timeout, but the tradeoff probably weights against caching, at least until Travis can do it outside of the actual build.


leodasvacas:
The builds turned out slightly slower than usual, so lets just leave things as they are.

Issue #336, opened 2018-09-07
Document block ingestor algorithm in comments/rustdocs


timmclean:
I need to make sure the code is fully documented so that the design and algorithms used can be understood (without referring to designs in the original GitHub issues).

Leo had some good feedback here: https://github.com/graphprotocol/graph-node/pull/309#pullrequestreview-152036427

timmclean:
I walked through the block ingestor code with @leodasvacas so he has a good understanding now.

Based on that conversation, here are a few points for further documentation:
- `upsert_blocks` uses "on conflict do update": add a comment about what conflict means, when it would update, and why update is not always a no-op
- To be investigated: will exposing `block.uncles` to mappings cause non-determinism?
- Clarify on BlockStore::attempt_head_update docs the decision to not update head block when chain length is no bigger. We don't always move to what the Ethereum node considers to be the "latest block".
- Why/when the `attempt_head_update` `SELECT LEFT OUTER JOIN` query is useful: common case is processing one missing block at a time, but node restarts can cause parallelism (add these notes to the BlockStore trait docs to keep documentation in one place, mention that this is the "current implementation").
- Add comment justifying loop termination for BlockIngestor::do_poll loop_fn

I'm holding off on completing this documentation work until the block processing work has stabilized to avoid merge conflicts

PR #363, opened 2018-09-13
Add Ethereum block ingestion algorithm

### LINKED ISSUE #286: Implement block ingestor component

Jannis:
This resolves #286.


datasource/ethereum/src/block_ingestor.rs
```diff
@@ -0,0 +1,179 @@
...
+                // Might be a no-op if latest block is one that we have seen.
+                // ingest_blocks will return a (potentially incomplete) list of blocks that are
+                // missing.
+                self.ingest_blocks(stream::once(Ok(latest_block)))
+            })
+            .and_then(move |missing_block_hashes| {
+                // Repeatedly fetch missing blocks, and ingest them.
+                // ingest_blocks will continue to tell us about more missing blocks until we have
+                // filled in all missing pieces of the blockchain (that we care about).
+                future::loop_fn(
```
leodasvacas: I still don't understand how this can loop more than twice, but that's something I can ask Tim later if you are also not sure of what's going on.

leoyvens (approved):

datasource/ethereum/src/block_ingestor.rs
leodasvacas: [This comment](https://github.com/graphprotocol/graph-node/pull/309#discussion_r214336772) is the understanding I had from reading the code, if it's correct this should probably be refactored for clarity. I'll catch @timmclean later to discuss this so that more than one person has good grip of what's going on here.

Jannis: Ok, please do. In the meantime, we'll merge the PR to get this in.

timmclean: Yup, Jannis' description is correct. Leo's totally right, this needs to be clearer. I filed #336 to track documenting this. Leo: let's connect when I'm back next week and find a way to make this clearer :)

PR #369, opened 2018-09-14
Bump and re-export web3 from the graph crate, clean up Ethereum type usage

Jannis:
Previously, we were using the same types from `ethereum-types`, `ethabi` and `web3` in various places. Most of our crates would depend on `web3` separately. Altogether, this made it hard to be consistent and bump any of these dependencies, because this would introduce version conflicts.

This PR makes it so that only the `graph` crate depends on `web3`. It re-exports it for all other crates to use. The dependency on `ethereum-types` is dropped altogether and `ethabi` is only used for those types that are not re-exported by `web3`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -461,32 +461,10 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "error-chain"
-version = "0.11.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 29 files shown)


leoyvens (approved):
Lgtm, hooray for simplifying our dependency configuration.

timmclean:
Ohhhhhhh I am **so** excited to see this :tada: :tada: :tada:
I kept running into this

PR #373, opened 2018-09-15
docs: Add info about store.get

html5cat:


docs/getting-started.md
```diff
@@ -181,18 +181,40 @@ token.setAddress('owner', event.params.to)
 token.setU256('amount', event.params.tokens)
 ```
 
-There is also a global `Store` class which has a `set` method for setting the value(s) of a particular entity's attribute(s) in the store.
+There is also a global `Store` class which has a `set` and `get` methods for setting ang getting the value(s) of a particular entity's attribute(s) in the store.
```
Jannis: Grammar: "a set and get methods" -> "set and get methods"
Typo: "ang getting" -> "and getting"

docs/getting-started.md
```diff
@@ -181,18 +181,40 @@ token.setAddress('owner', event.params.to)
...
 ```
 
-The eventHandlers functions return `void`. The only way that entities may be added to the The Graph is by calling `Store.set()`. `Store.set()` may be called multiple times in an event handler.
+The eventHandlers functions return `void`. The only way that entities may be added to the The Graph is by calling `store.set()`. `store.set()` may be called multiple times in an event handler.
+
+**Note** `store.set()` will only set the entity attributes that have explicitly been set on the `Entity`. Attributes which are not explicitly set, or unset by calling `Entity.unset(<attribute>)`, will not be overwritten.
+
+#### `store.get(entity: string, id: string)`
+
+You can use `store.get` to retreive information previously added with `store.get`.
```
Jannis: Typo: "previously added with store.get" -> "previously added with store.set"

docs/getting-started.md
```diff
@@ -181,18 +181,40 @@ token.setAddress('owner', event.params.to)
...
+`store.get` expects entity type and the id of the entity.
+
+##### Example
+
+```javascript
+  store.set('Challenge', challengeId.toHex(), challenge)
+  let challenge2 = store.get('Challenge', challengeId.toHex())
+  let id = challenge2.getString("application")
+  let challenge2entity = new Entity()
+  challenge2entity.setString('id', id)
```
Jannis: It should be possible to just call `setString` on `challenge2`, rather than creating a new, empty entity.

Jannis: The current code example is confusing due to that extra entity.

docs/getting-started.md
leodasvacas: `Store` is not a class just namespace, and it should be in small caps `store`.

html5cat:
@Jannis could you review updated doc?

html5cat:
@Jannis updated as per feedback – could you review and merge? Thanks!

docs/getting-started.md
```diff
@@ -181,18 +181,40 @@ token.setAddress('owner', event.params.to)
 token.setU256('amount', event.params.tokens)
 ```
 
-There is also a global `Store` class which has a `set` method for setting the value(s) of a particular entity's attribute(s) in the store.
+There is also a global `store` class which has `set` and `get` methods for setting and getting the value(s) of a particular entity's attribute(s) in the store.
```
Jannis: `store` is a module, not a class.

docs/getting-started.md
```diff
@@ -181,18 +181,40 @@ token.setAddress('owner', event.params.to)
...
+
+`store.set` expects the name of an entity type, the id of the entity and the `Entity` itself.
 
 ##### Example
 ```typescript
   store.set('Token', tokenId, token)
 ```
 
-The eventHandlers functions return `void`. The only way that entities may be added to the The Graph is by calling `Store.set()`. `Store.set()` may be called multiple times in an event handler.
+The eventHandlers functions return `void`. The only way that entities may be added to the The Graph is by calling `store.set()`. `store.set()` may be called multiple times in an event handler.
```
Jannis: Double "the" in "to the The Graph".

docs/getting-started.md
```diff
@@ -181,18 +181,40 @@ token.setAddress('owner', event.params.to)
...
+
+**Note** `store.set()` will only set the entity attributes that have explicitly been set on the `Entity`. Attributes which are not explicitly set, or unset by calling `Entity.unset(<attribute>)`, will not be overwritten.
+
+#### `store.get(entity: string, id: string)`
+
+You can use `store.get` to retreive information previously added with `store.set`.
+`store.get` expects entity type and the id of the entity.
```
Jannis: Grammar: "expects the entity type and ID of the entity" instead of "expects entity type and the ID of the entity".

docs/getting-started.md
```diff
@@ -181,18 +181,40 @@ token.setAddress('owner', event.params.to)
...
+
+##### Example
+
+```javascript
+  store.set('Challenge', challengeId.toHex(), challenge)
+  let challenge2 = store.get('Challenge', challengeId.toHex())
+  let id = challenge2.getString("application")
+  let challenge2entity = new Entity()
```
Jannis: I find this example to be a little artificial and confusing — it would be better to use two completely separate entity types here (e.g. `Challenge` and `Registration` or something).

PR #376, opened 2018-09-17
Postgres order_by attribute cast to correct data type before performing comparisons

### LINKED ISSUE #331: Fix BigInt comparisons in GraphQL filters
This is clearly not the right order:
image

fordN:
Resolves #331 

This PR fixes the filter query bug where all `order_by` comparisons were done in alphabetical order because the `jsonb` objects are returned as `text`; when adding an `order_by` filter the `jsonb` attribute must be cast from text to the correct data type before performing comparisons 

A new enum was created, `ValueType`, with a variant for each supported data type.  `ValueType` is now passed into the `order_by` query fragment builder and used to decide which data cast to perform before ordering.  The `order_by` queries have all been updated to now include `NULLS LAST`, so null values will not appear in the beginning of the filtered set.

Data casting from `jsonb::text` before comparing:
  - BigInt -> cast to `numeric`
  - Int -> cast to `bigint`
  - String -> leave as `text`
  - Float -> cast to `double precision`
  - Id -> leave as `text`
  - Bytes -> leave as `text`
  - Boolean -> cast to `boolean`

### DIFF
--- a/graph/src/components/store.rs
+++ b/graph/src/components/store.rs
@@ -1,7 +1,7 @@
 use failure::Error;
 use futures::Future;
 use futures::Stream;
-use graphql_parser::{schema as s};
+use graphql_parser::schema as s;
 use web3::types::{Block, Transaction, H256};
 
 use data::store::*;
--- a/graph/src/components/store.rs
+++ b/graph/src/components/store.rs
@@ -6,6 +6,7 @@ use web3::types::{Block, Transaction, H256};
 
 use data::store::*;
 use std::fmt;
+use std::str::FromStr;
 use web3::types::H256;
 
 use prelude::*;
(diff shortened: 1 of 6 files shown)


graph/src/components/store.rs
```diff
@@ -57,6 +59,40 @@ pub struct StoreRange {
...
+                _ => panic!("Found unexpected NamedType variant when converting to ValueType"),
+            },
+            s::Type::NamedType(ref name) if name == "Boolean" => ValueType::Boolean,
+            s::Type::NamedType(ref name) if name == "BigInt" => ValueType::BigInt,
+            s::Type::NamedType(ref name) if name == "Bytes" => ValueType::Bytes,
+            s::Type::NamedType(ref name) if name == "Float" => ValueType::Float,
+            s::Type::NamedType(ref name) if name == "ID" => ValueType::ID,
+            s::Type::NamedType(ref name) if name == "String" => ValueType::String,
+            _ => panic!("Found unexpected NamedType variant when converting to ValueType"),
+        }
```
leodasvacas: It looks like `ValueType` could use a [FromStr](https://doc.rust-lang.org/std/str/trait.FromStr.html) impl.

graphql/src/store/query.rs
```diff
@@ -100,11 +100,18 @@ fn list_values(value: Value) -> Vec<Value> {
...
+    arguments: &HashMap<&q::Name, q::Value>,
+) -> Option<(String, ValueType)> {
     arguments
         .get(&"orderBy".to_string())
         .and_then(|value| match value {
-            q::Value::Enum(name) => Some(name.to_owned()),
+            q::Value::Enum(name) => {
+                let field = sast::get_field_type(entity, &name)
+                    .expect("order by attribute does not belong to entity");
+                Some((name.to_owned(), ValueType::from(field.clone().field_type)))
```
leodasvacas: `field.clone().field_type` is probably better written as `field.field_type.clone()`

store/postgres/src/store.rs
```diff
@@ -290,19 +290,50 @@ impl BasicStore for Store {
...
+                    sql::<Text>("(data ->>")
+                        .bind::<Text, _>(order_attribute)
+                        .sql(&format!(") {} NULLS LAST", direction)),
+                ),
+                ValueType::Boolean => diesel_query.order(
+                    sql::<Text>("(data ->>")
+                        .bind::<Text, _>(order_attribute)
+                        .sql(&format!(")::boolean {} NULLS LAST", direction)),
+                ),
+            };
```
leodasvacas: This code duplication is unfortunate, could we try to put the cast string `::bigint`, `::boolean`, etc.. (or empty if it's just a string) in a variable and then pass it as an argument to `format!`?

graph/src/components/store.rs
```diff
@@ -57,6 +59,40 @@ pub struct StoreRange {
...
+    Float,
+    String,
+    Bytes,
+    BigInt,
+}
+
+impl From<s::Type> for ValueType {
+    fn from(schema_type: s::Type) -> ValueType {
+        match schema_type {
+            s::Type::NonNullType(ref inner) => match inner.deref() {
```
leodasvacas: You can remove the `ref` and the `.deref()` here.

Jannis:
I have to look at the code still but one thing stood out from the start: our `Float` attribute value type is `f32`, so that should not be cast to `numeric`. `BigInt` has to be cast to `numeric`, since Postgres' `bigint` type is only 8 bytes (= 64 bits).

graph/src/components/store.rs
fordN: That would indeed cleanup the From<s::Type> impl.  I'll add it in.

store/postgres/src/store.rs
```diff
@@ -290,19 +290,27 @@ impl BasicStore for Store {
...
-
+            let cast_type = match value_type {
+                ValueType::BigInt => "::numeric".to_string(),
+                ValueType::Boolean => "::boolean".to_string(),
+                ValueType::Bytes => "".to_string(),
+                ValueType::Float => "::float".to_string(),
+                ValueType::ID => "".to_string(),
+                ValueType::Int => "::bigint".to_string(),
+                ValueType::String => "".to_string(),
+            };
```
leodasvacas: Here you can move the `to_string()` to after the match, so this line will be `}.to_string();`

graph/src/components/store.rs
```diff
@@ -57,6 +59,58 @@ pub struct StoreRange {
...
+            "ID" => Ok(ValueType::ID),
+            "Int" => Ok(ValueType::Int),
+            "String" => Ok(ValueType::String),
+            e => Err(ValueTypeError::UnexpectedTypeName {
+                name: e.to_string(),
+            }),
+        }
+    }
+}
+impl From<s::Type> for ValueType {
```
Jannis: We can't panic when executing queries. So `From` is not an option.

graphql/src/store/query.rs
```diff
@@ -100,11 +100,18 @@ fn list_values(value: Value) -> Vec<Value> {
 }
 
 /// Parses GraphQL arguments into an attribute name to order by, if present.
-fn build_order_by(arguments: &HashMap<&q::Name, q::Value>) -> Option<String> {
+fn build_order_by(
```
Jannis: This function needs to return a result and so should `build_query`. Because if the query is somehow not quite right, we want to send an error back to the client.

graphql/src/store/query.rs
```diff
@@ -100,11 +100,18 @@ fn list_values(value: Value) -> Vec<Value> {
...
+    entity: &s::ObjectType,
+    arguments: &HashMap<&q::Name, q::Value>,
+) -> Option<(String, ValueType)> {
     arguments
         .get(&"orderBy".to_string())
         .and_then(|value| match value {
-            q::Value::Enum(name) => Some(name.to_owned()),
+            q::Value::Enum(name) => {
+                let field = sast::get_field_type(entity, &name)
+                    .expect("order by attribute does not belong to entity");
```
Jannis: Same here, this error needs to be returned. It also needs to be more specific and mention the types/names involved. Like:
```
Value "foo" is not valid for MyEntity_OrderBy. Valid values are: baz, ruux.
```

Our `Strings` type can be used to represent the valid values.

graphql/src/schema/ast.rs
```diff
@@ -104,6 +106,21 @@ pub fn get_field_type<'a>(object_type: &'a ObjectType, name: &Name) -> Option<&'
...
 
+
+/// Returns the ValueType of a schema field
+pub fn get_value_type(schema_type: Type) -> Result<ValueType, ValueTypeError> {
+    match schema_type {
+        Type::NamedType(ref name) => Ok(ValueType::from_str(&name).unwrap()),
+        Type::NonNullType(inner) => match *inner {
+            Type::NamedType(ref name) => Ok(ValueType::from_str(&name).unwrap()),
+            Type::NonNullType(_) => Err(ValueTypeError::NestedNonNullType),
+            Type::ListType(_) => Err(ValueTypeError::CannotConvertFromListType),
```
leodasvacas: Nested non-null types [are not valid graphql](https://facebook.github.io/graphql/draft/#sec-Type-References) and I've tested that it's a parse error, so you can remove that error and replace this whole match with a recursive call to `get_value_type`.

Jannis:
Obsoleted by #494.

PR #379, opened 2018-09-18
Add store.get() to runtime for use by subgraph mapping handlers

### LINKED ISSUE #357: Implement and export store.get
When building a subgraph, developers will often want to access existing entity data-e.g. a token owner or an item being sold in an auction-and update it. We already support partial updates via `store.set()` but there are many situations, where it is necessary to first obtain the existing data with `store.get()` before performing calculations and writing updated data back with `store.set()`.

fordN:
Resolves #357 

Developers may want to access existing data to build the next `Entity`; this PR adds a hosted function, `Store.get()`, to the runtime to give subgraph developers a method to access existing data in the mappings. 
 - Pass the `Store` from `RuntimeManager` down to the `HostedExternals` of the `WasmiModule`
 - Implement the hosted function `Store.get()` for use by the mappings. 
 - Implement `From<Entity>` trait for `Hashmap<Attribute, Value>`. 
 - Implement `ToAscObj<AscEntity>` for `HashMap<String, store::Value>`.
 - Use mock store in runtime module tests.

### DIFF
--- a/runtime/wasm/src/module.rs
+++ b/runtime/wasm/src/module.rs
@@ -318,17 +318,12 @@ where
         };
 
         // Retrieve an Entity from the store
-        let get_result = self
-            .store
+        self.store
             .lock()
             .unwrap()
             .get(store_key)
-            .map(|entity| entity);
-        let get_result: HashMap<_, _> = match get_result {
-            Ok(entity) => entity.into(),
-            Err(_e) => return Err(host_error("Error getting entity".to_string())),
-        };
-        Ok(Some(RuntimeValue::from(self.heap.asc_new(&get_result))))
+            .map_err(|_| host_error("Error getting entity".to_string()))
+            .and_then(|result| Ok(Some(RuntimeValue::from(self.heap.asc_new(&result)))))
     }
 
     /// function ethereum.call(call: SmartContractCall): Array<Token>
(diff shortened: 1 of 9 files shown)


runtime/wasm/src/module.rs
```diff
@@ -299,6 +305,32 @@ where
...
+
+        // Retrieve an Entity from the store
+        let get_result = self
+            .store
+            .lock()
+            .unwrap()
+            .get(store_key)
+            .map(|entity| entity);
+        let get_result: HashMap<_, _> = match get_result {
+            Ok(entity) => entity.into(),
```
fordN: Type ascription is still experimental; other ideas for supplying into() with the type without using an intermediate variable?

runtime/wasm/src/module.rs
```diff
@@ -299,6 +305,32 @@ where
...
+            Err(_e) => return Err(host_error("Error getting entity".to_string())),
```
leodasvacas: It's usually nice to format the error `e` into the message.

runtime/wasm/src/module.rs
leodasvacas: I prefer `from` over `into` to avoid these problems, so you can do `HashMap::from(entity)`

runtime/wasm/src/module.rs
fordN: Store.get() currently returns `Result<Entity, ()>`, though that will be changing soon.  I've made a note to update this error handling in my store errors branch.

Jannis:
@fordN Great, I'd say this is ready.

PR #380, opened 2018-09-18
Subgraph list

### LINKED ISSUE #282: Add deploy and remove admin API commands

### LINKED ISSUE #361: Make JSON-RPC server log using a component context

leoyvens:
Adds the `subgraph_list` json-rpc command. It returns an object with `name: id` pairs. No auth is required to list subgraphs.

Adds tests for subgraph deploying, removing and listing. Fixes bug caught by test, where the `SchemaRemoval` event was not being sent when updating a subgraph. Also refactors `SubgraphProvider::deploy` to take `Arc<Self>`.

Fixes #361. Resolves #282.


core/src/subgraph/provider.rs
```diff
@@ -36,11 +35,30 @@ impl<L: LinkResolver> SubgraphProvider<L> {
...
+        let host_remove_sink = self.event_sink.clone();
+        graphql_remove_sink
+            .send(SchemaEvent::SchemaRemoved(name, id.clone()))
+            .map_err(|e| panic!("Failed to forward schema removal: {}", e))
+            .map(|_| ())
+            .join(
+                host_remove_sink
+                    .send(SubgraphProviderEvent::SubgraphRemoved(id))
+                    .map(|_| ())
+                    .map_err(|e| panic!("Failed to forward subgraph removal: {}", e)),
```
Jannis: Same here

core/src/subgraph/provider.rs
```diff
@@ -36,11 +35,30 @@ impl<L: LinkResolver> SubgraphProvider<L> {
...
+            ).map(|_| ())
+    }
 }
 
 impl<L: LinkResolver> SubgraphProviderTrait for SubgraphProvider<L> {
     fn deploy(
-        &self,
+        arc_self: &Arc<Self>,
```
Jannis: IMHO, this should not be a caller concern. I'd rather clone the subgraphs Arc, the resolver and the senders etc. internally than require callers to do `SubgraphProvider::deploy(&provider_arc, ...)`.

core/src/subgraph/provider.rs
```diff
@@ -166,9 +186,9 @@ fn rejects_name_bad_for_urls() {
         }
     }
     let logger = slog::Logger::root(slog::Discard, o!());
-    let provider = SubgraphProvider::new(logger, Arc::new(FakeLinkResolver));
+    let provider = Arc::new(SubgraphProvider::new(logger, Arc::new(FakeLinkResolver)));
     let bad = "/../funky%2F:9001".to_owned();
-    let result = provider.deploy(bad.clone(), "".to_owned());
+    let result = SubgraphProvider::deploy(&provider, bad.clone(), "".to_owned());
```
Jannis: I'm not a fan :P

Jannis (changes requested):

core/src/subgraph/provider.rs
leodasvacas: This is a common pattern in our code and I wanted to try something different here, it's good to discuss whether we like it or not.

I agree call sites are nasty, but we only have one non-test call of this function. I like it because besides cutting 8 lines of `.clone()`s and not having to come up with a good name for each cloned variable, it's easy to break long compositions of futures into methods because `self` can be moved into the closures, as is done here with `send_remove_events`.

There is an RFC for writing `self: Arc<Self>` and another for async/await which would really solve things here but they are far off.

leodasvacas: On further thought, I could probably get the same benefits by cloning `self` entirely. I'll try that.

leodasvacas:
@Jannis I think I got a nice solution for `SubgraphProvider::deploy` on the latest commit.

Jannis (dismissed):
Yup, I like it.

fordN (approved):
👍

PR #384, opened 2018-09-19
Add network identifier check on startup

### LINKED ISSUE #377: Show helpful error when node is run on a different network but with same network name
When the same network name (e.g. `--ethereum-rpc ganache:...`) is used with two different chains (e.g. two separate instances of Ganache; or one Ethereum mainnet and one testnet endpoint), the following warning will be logged:

```
Sep 17 22:17:18.482 WARN failed to poll for latest block: ErrorMessage {
  msg: "could not get block from Ethereum: RPC error: Error {
    code: ServerError(-32000),
    message: \"Key not found in database\",
    data: Some(Object({\"stack\": String(\"NotFoundError: Key not found in database\\n    at /usr/local/lib/node_modules/ganache-cli/build/cli.node.js:2:773948\\n    at /usr/local/lib/node_modules/ganache-cli/build/cli.node.js:2:776500\\n    at /usr/local/lib/node_modules/ganache-cli/build/cli.node.js:2:748581\\n    at ReadFileContext.callback (/usr/local/lib/node_modules/ganache-cli/build/cli.node.js:2:750214)\\n    at FSReqWrap.readFileAfterOpen [as oncomplete] (fs.js:352:13)\"), \"name\": String(\"NotFoundError\")})) 
  }" 
}
```

Doing this results in the Graph Node having an invalid latest block for this network. It will then get an error from Ethereum when calling `eth_getBlockByHash` with the hash of this block, which belongs to the previous chain but not the one in use at the time of the error.

Discovered in https://github.com/graphprotocol/graph-node/issues/375#issuecomment-422159481

Users shouldn't do this, but we should detect this mistake and show a more helpful error (and panic).

timmclean:
Check that we are connected to the same Ethereum network as the last time this network name was used. Panic block ingestor if, for example, the user switches from mainnet to kovan but does not use a new network name.

`add_network_if_missing` will be refactored away when the new block processing work gets merged in, but I'm sticking with that design for now

Maybe @nenadjaja could test this?

Closes #377

### DIFF
--- a/store/postgres/src/store.rs
+++ b/store/postgres/src/store.rs
@@ -3,7 +3,7 @@ use diesel::pg::Pg;
 use diesel::pg::PgConnection;
 use diesel::prelude::*;
 use diesel::sql_types::Text;
-use diesel::{debug_query, delete, insert_into, result, select};
+use diesel::{debug_query, delete, insert_into, result, select, update};
 use filter::store_filter;
 use futures::sync::mpsc::{channel, Sender};
 use std::collections::HashMap;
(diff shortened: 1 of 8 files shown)


store/postgres/src/store.rs
```diff
@@ -330,17 +330,58 @@ impl BasicStore for Store {
...
+                            .eq::<Option<String>>(Some(format!("{:x}", new_genesis_block_hash))),
+                    )).on_conflict(name)
+                    .do_nothing()
+                    .execute(&*self.conn.lock().unwrap())?;
+            }
+
+            // Network is in database and has identifiers
+            Some((Some(last_net_version), Some(last_genesis_block_hash))) => {
+                if last_net_version != new_net_version {
+                    panic!("Ethereum node provided net_version {}, but we expected {}. Did you change networks without changing the network name?", new_net_version, last_net_version);
```
Jannis: The two panic lines are a little too long. Can we break them up roughly like this?
```rust
panic!("Ethereum node provided net_version {}, \
        but we expected {}. Did you change networks \
        ...);
```

Jannis (changes requested):

store/postgres/src/store.rs
timmclean: Oh, good catch. I guess cargo fmt doesn't remove all of my formatting obligations :)

Jannis (approved):
Looks good now. I'll update Travis tomorrow to ignore failed builds with Rust beta.

PR #387, opened 2018-09-19
Bump all dependencies

Jannis:
The main reason for this is to get rid of the dozens of compiler warnings logged for Diesel since we switched to Rust 1.29. But I figured it can't hurt to get everything else updated as well.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,7 +4,7 @@ version = "0.11.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "lazy_static 1.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
- "regex 1.0.2 (registry+https://github.com/rust-lang/crates.io-index)",
+ "regex 1.0.5 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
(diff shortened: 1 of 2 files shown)


leodasvacas:
Great! This broke a bunch of store tests, not sure why.

Jannis:
Ah, should've run those tests! I pinned jsonrpc-http-server to a specific commit because the latest commit failed to compile. (I'd have to try again to see what the error was exactly.)

leodasvacas:
@Jannis Makes sense, no need to check why exactly their master is broken.

Jannis:
I investigated why the store tests fail. The error is introduced between `serde_json = "1.0.25"` and `serde_json = "1.0.26"`. Between those releases, they switched from [dtoa](https://github.com/dtolnay/dtoa) to [ryu](https://github.com/dtolnay/ryu) for floating point (de)serialization.

I think what may be happening here is that we write entities to the store with `f32` values and then what comes back when deserializing is not `f32` and so `serde_json` cannot match float values to our `Value::Float(f32)` variant.

Would have to investigate further how we can fix this.

Jannis:
Closing for now, deps to be updated later.

PR #393, opened 2018-09-19
store/postgres: Include subgraph id as input to block revert functions

timmclean:
Let's get this in master and then we can bring it up to feature/block-stream


timmclean:
@fordN if this still looks good go ahead and merge :)

PR #395, opened 2018-09-19
Add EthereumBlockPointer type

timmclean:
Ported from tim/block-processing


graph/src/components/ethereum/adapter.rs
```diff
@@ -112,6 +112,57 @@ pub struct EthereumEvent {
...
+        EthereumBlockPointer {
+            hash: b.hash.unwrap(),
+            number: b.number.unwrap().as_u64(),
+        }
+    }
+}
+
+impl From<(H256, u64)> for EthereumBlockPointer {
+    fn from((hash, number): (H256, u64)) -> EthereumBlockPointer {
+        if number >= (1 << 63) {
```
Jannis: Can that even happen?

Jannis: Ah, nevermind, I just saw the comment above.

PR #398, opened 2018-09-20
Allow failures on Rust beta

leodasvacas:
ipfs-api is regressed on 1.30 beta, ignore that until it's un-regressed.


leodasvacas:
The CI failure is very strange but obviously unrelated to this PR, opened #401 to track investigating it.

PR #409, opened 2018-09-24
Handle IPFS hostname in addition to IP address

### LINKED ISSUE #339: Use to_socket_addrs() to allow `--ipfs` flag to take hostnames in addition to IP addresses
Right now the node only takes IP addresses because it constructs the `SocketAddr` manually rather than converting the `--ipfs` value using `to_socket_addrs()`.

A nice little example can be found here: https://stackoverflow.com/a/28255903

fordN:
Resolves #339 

The IPFS node location may be defined by a hostname rather than specific IP address.  This PR used to_socket_addr() to resolve a hostname to one or more IP addresses, then loops through the set of addresses attempting to create an IPFS client for each.  If none of the resolved addresses are valid a panic fires with the collection or errors and invalid urls.


Jannis:
@fordN  I started reviewing the IPFS hostname PR — I didn’t like the `mut failed_sockets` and the iterator `peek`ing. Too much mutable state and custom iteration.

I felt I could explain it better in code, so I reworked the IPFS address resolution and pushed a new commit to the PR. Instead of two things I mentioned it now uses `Itertools::fold_while` to resolve the addresses to a `Result<(IpfsClient, SockAddr), Vec<InvalidUrl>>`. S either to the first successfully created client (using short circuiting) or to collected errors.

Let me know what you think. IMHO it always pays off to think about how a problem can be solved in a functional way (think: filter, map, reduce/fold) rather than stateful loops. 😄

leoyvens (approved):
@Ford so we were looking for `fold_while` which is in `itertools`, nice find @Jannis!

PR #416, opened 2018-09-26
Subscription removal and cancelation helpers

### LINKED ISSUE #362: Fix GraphQL subscriptions to not panic when removing subgraphs

### LINKED ISSUE #392: Generic solution for cancelling streams/futures
We often need to be able to shutdown event streams or cancel long-running operations. Right now, we're doing this by using an async `oneshot::Sender` to trigger an error on the stream. There are a few problems with this:

- It's verbose and cumbersome for something that we want so often.
- Sometimes there are streams used in the production of streams. We usually want to cancel the innermost stream when the outermost stream is cancelled, so that way cancellation takes effect as early as possible.
- `oneshot::Sender` isn't clone, and we sometimes want to be able to cancel from more than one place.
- It's a bit hard to read. Dropping a Sender isn't very intuitive.

https://github.com/jonhoo/stream-cancel is close to what we want. "Valves" are used to shut off streams. Internally, they are implemented similarly with a oneshot channel, but valves are Clone, so they can be passed into closures.

Unfortunately, valves do not have the semantics that we want. Valves shutdown a stream gracefully, ending the flow of items without an error. This is fine for event streams, but can be an issue elsewhere. What we actually want is for the stream to terminate with an error. Suppose we are reading a stream of logs from a block. If the stream ends early without an error, it makes it seem like we successfully read all logs from the block when, actually, the operation was cancelled.

Proposal for a generic solution:
```rust
let (canceller, cancellable) = Cancellable::new();

// Stream extension trait added to graph prelude
let new_stream = stream.cancellable(cancellable.clone());
// new_stream has same Item type, but Error type is wrapped in CancellationOrError

// Future extension trait added to graph prelude
let new_future = future.cancellable(cancellable.clone());
// new_future has same Item type, but Error type is wrapped in CancellationOrError

// Reads from an AtomicBool
let is_cancelled: bool = cancellable.is_cancelled();

let canceller2 = canceller.clone();
```

When `canceller` (or one of its clones) is called, then the future and stream immediately resolve with an error value `CancellationOrError::Cancelled`. The `is_cancelled` method is also available for more fine-grained control.

We'll need a type like this (name open to bikeshedding):

```rust
enum CancellationOrError<E> {
  Cancelled,
  Error(E),
}
```

It would be nice if we had a way of collapsing `CancellationOrError<CancellationOrError<E>>` automatically into `CancellationOrError<E>`. Whether the inner operation or the outer operation was cancelled is, in my opinion, an implementation detail that generally shouldn't be exposed to callers.

Tentatively, I think that dropping the canceller should not cause cancellation. If the reason for possibly cancelling goes away, that seems like a reason to keep going.

leodasvacas:
This shutdown subscriptions cleanly when the subgraph is removed, by responding to all started operations on all connections with GQL_COMPLETE. Resolves #362.

It also extends Stream and Future with a `cancelable` method that returns a cancelation guard. Part of #392.


graph/src/extension.rs
```diff
@@ -0,0 +1,120 @@
+use futures::sync::oneshot;
```
Jannis: The filename `extension.rs` could be an extension of anything. Can we rename it to e.g. `graph/src/ext/futures.rs` or `graph/src/util/futures.rs`?

graph/src/components/subgraph/registry.rs
```diff
@@ -15,33 +15,57 @@ impl<T> Clone for SubgraphRegistry<T> {
...
+impl<T> SubgraphRegistry<T> {
     pub fn new() -> Self {
         SubgraphRegistry {
             names_to_ids: Default::default(),
             ids_to_values: Default::default(),
         }
     }
 
-    pub fn resolve(&self, id_or_name: &String) -> Option<T> {
+    pub fn resolve_by<U, F: FnOnce(&T) -> U>(&self, id_or_name: &String, by: F) -> Option<U> {
```
Jannis: Does this extra `map(by)` call at the end really warrant a dedicated method? It would seem trivial to just add it to via `subgraphs.resolve(&subgraph).map(|s| s.schema.clone())`.

Jannis: I see that the regular `resolve` clones at the end. How about we make that return `Option<&T>` and then we can clone/map where we need to outside the method?

Jannis: _If_ we decide to keep this, I'd prefer `resolve_map` (like `filter_map`).

graph/src/extension.rs
```diff
@@ -0,0 +1,120 @@
...
+use std::sync::{Arc, Mutex};
+use tokio::prelude::{Future, Poll, Stream};
+
+pub struct Cancelable<T, C> {
```
Jannis: This type could use some docs about how it works and what you can do with it.

graph/src/extension.rs
```diff
@@ -0,0 +1,120 @@
...
+}
+
+impl SharedCancelGuard {
+    /// Cancels the stream, a noop if already canceled.
+    pub fn cancel(&self) {
+        *self.guard.lock().unwrap() = None
+    }
+
+    pub fn is_canceled(&self) -> bool {
+        self.guard.lock().unwrap().is_some()
```
Jannis: Unless I'm reading this wrong, I think this check should be `.is_none()`.

graph/src/extension.rs
```diff
@@ -0,0 +1,120 @@
...
+    }
+}
+
+pub trait StreamExtension: Stream + Sized {
+    /// When `cancel` is called on a `CancelGuard` or it is dropped,
+    /// `Cancelable` receives an error.
+    ///
+    /// `on_cancel` is called to make an error value upon cancelation.
+    fn cancelable<C: Fn() -> Self::Error>(self, on_cancel: C)
```
Jannis: @timmclean How do you feel about the `on_cancel` callback? Is that something that would work well with cascading streams?

I'm not sure if I'd prefer an equivalent to `map_sink_err` (e.g. `map_cancelable_error`) instead of passing the callback in to this function. It feels slightly more idiomatic but also less obvious. @leodasvacas What do you think?

server/websocket/src/connection.rs
```diff
@@ -69,13 +71,106 @@ impl From<OutgoingMessage> for WsMessage {
...
+    msg: OutgoingMessage,
+) -> Result<(), WsError> {
+    sink.unbounded_send(msg.into())
+        .map_err(|_| WsError::Http(500))
+}
+
+/// Helper function to send error messages.
+fn send_error_string(
+    sink: &mpsc::UnboundedSender<WsMessage>,
+    id: String,
```
Jannis: Let's be more obvious here and call the two parameters `operation_id` and `err_str` or something.

server/websocket/src/connection.rs
```diff
@@ -69,13 +71,106 @@ impl From<OutgoingMessage> for WsMessage {
...
+        }
+    }
+
+    fn contains(&self, id: &str) -> bool {
+        self.operations.contains_key(id)
+    }
+
+    fn insert(&mut self, id: String) -> impl Future<Item = (), Error = ()> {
+        // Create a oneshot channel to stop the subscription later.
+        let (stopper, stopped) = oneshot::channel();
```
Jannis: Am I mistaken or would this be a great application for `Cancelable`? 😉

server/websocket/src/server.rs
```diff
@@ -137,19 +158,30 @@ where
...
                                 ws_stream,
                                 graphql_runner.clone(),
                             );
-                            tokio::spawn(service.into_future()).into_future()
+
+                            let cancel_subgraph = subgraph.clone();
+                            let (service, guard) = service.into_future().cancelable(move || {
+                                debug!(
+                                    logger,
+                                    "Canceling subscriptions for subgraph `{}`", cancel_subgraph
```
Jannis: We've been _fairly_ consistent in using structured logging, so it would be nice to use that here as well:
```
debug!(
    logger,
    "Canceling subscriptions for subgraph";
    "subgraph_name" => ...
    "subgraph_id" => ...,
);
```

server/websocket/src/server.rs
```diff
@@ -137,19 +158,30 @@ where
...
                                 subgraphs.clone(),
-                                subgraph,
+                                subgraph.clone(),
                                 ws_stream,
                                 graphql_runner.clone(),
                             );
-                            tokio::spawn(service.into_future()).into_future()
+
+                            let cancel_subgraph = subgraph.clone();
+                            let (service, guard) = service.into_future().cancelable(move || {
```
Jannis: I know it was called `service` before (my bad), but can we rename it to `conn` or `gql_conn`?

graph/src/components/subgraph/registry.rs
leodasvacas: I tried making it return a reference but couldn't get it to work because the reference is limited by the lifetime of the lock, even if I also returned the lock it still doesn't work. The rename sounds good.

graph/src/extension.rs
leodasvacas: I had a 50% chance of getting it right.

Jannis: 🤣

graph/src/extension.rs
leodasvacas: The error handling is tricky here. `Cancelable` supports the `map_err` methods since it's a stream or future, the problem is determining what should be the error type of `Cancelable`, which must be convertible from `Self::Error`.

@timmclean suggested an enum approach in the issue, but I think overly nested errors can get hard to manage. My first idea was requiring `Self::Error: From<Canceled>` but it was hard to use.

The closure I think is pretty flexible and ergonomic, for example it allows `Self::Error` to be an enum with a `Canceled` variant. Though it doesn't resemble the standard futures and stream APIs. It's my best idea so far, let's see how it fares as we start to use this more.

graph/src/ext/futures.rs
```diff
@@ -0,0 +1,123 @@
...
+        let (canceler, canceled) = oneshot::channel();
+        (
+            Cancelable {
+                cancelable: self,
+                canceled,
+                on_cancel,
+            },
+            CancelGuard { canceler },
+        )
+    }
```
timmclean: I think we need a way to cancel a bunch of futures/streams with one canceler. Is that possible with this design?

graph/src/ext/futures.rs
```diff
@@ -0,0 +1,123 @@
...
+}
+
+/// It's not viable to use `select` directly, so we do a custom implementation.
+impl<S: Stream, C: Fn() -> S::Error> Stream for Cancelable<S, C> {
+    type Item = S::Item;
+    type Error = S::Error;
+
+    fn poll(&mut self) -> Poll<Option<Self::Item>, Self::Error> {
+        // Error if the stream was canceled, otherwise poll it.
+        if self.canceled.poll().is_err() {
```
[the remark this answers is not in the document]
leodasvacas: exactly

graph/src/ext/futures.rs
```diff
@@ -0,0 +1,123 @@
...
+}
+
+#[derive(Debug)]
+pub struct CancelGuard {
+    canceler: oneshot::Sender<()>,
+}
+
+impl CancelGuard {
+    /// A more readable `drop`.
+    pub fn cancel(self) {}
```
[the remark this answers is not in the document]
leodasvacas: Do you have a use case where a cancelable stream becomes uncancelabe? Seems unusual. A guard that does not cancel on drop would be harder to implement I think, so perhaps we can delay that effort until a need arises.

graph/src/ext/futures.rs
leodasvacas: It's not currently possible, this design can be extended to allow that, I can do it in a follow-up PR.

timmclean: Yeah, I think it'd be worth iterating on this design -- we're going to be using this everywhere

leodasvacas: I'm thinking it'd be a matter of adding two methods, one to create the a `SharedCancelGuard` along with a shared receiver, then a method `cancelable_with` that takes a shared receiver.

graph/src/ext/futures.rs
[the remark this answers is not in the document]
timmclean: Cool, that'd be worth some comments I think, either here or in the follow-up PR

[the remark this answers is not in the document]
leodasvacas: I've improved the comment here.

graph/src/ext/futures.rs
```diff
@@ -0,0 +1,123 @@
...
+    }
+}
+
+/// A version of `CancelGuard` that can be cloned.
+///
+/// To cancel the stream, call `cancel` on one of the guards or drop all guards.
+#[derive(Clone, Debug)]
+pub struct SharedCancelGuard {
+    guard: Arc<Mutex<Option<oneshot::Sender<()>>>>,
+}
```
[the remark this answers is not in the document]
timmclean: fair enough!

Jannis:
@timmclean So this is good to go? If so, could you approve it?

Jannis:
I've approved it but it would be nice if you did too, since you had more comments.

leodasvacas:
I want to try changes to cancelation helpers but I see no problem in changing that in a follow-up. Thanks for the reviews, merging!

PR #422, opened 2018-09-27
Return `null` from store.get if the entity wasn't found

Jannis:
This allows to conditionally create entities if they don't exist
yet, via e.g.

```typescript
let entity = store.get('MyEntity', 'some-id')
if (entity === null) {
  entity = new Entity()
  entity.setString('id', 'some-id')
}
entity.setString('someAttr', 'someValue')
store.set('MyEntity', entity.getString('id'), entity as Entity)
```

### DIFF
--- a/runtime/wasm/src/module/mod.rs
+++ b/runtime/wasm/src/module/mod.rs
@@ -333,8 +333,8 @@ where
             .lock()
             .unwrap()
             .get(store_key)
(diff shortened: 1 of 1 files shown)


davekay100 (approved):

leodasvacas:
This will return null on any database error. `get` should return an `Option`.

PR #428, opened 2018-09-28
Ford/store errors to interface

### LINKED ISSUE #48: Report store errors back to client instead of panicking
Example: We won't support all filters for all types, so we should

* Try to only expose filters that we support in the GraphQL schema
* Allow the Store to return useful errors when filters are not supported for certain value types.

fordN:
Resolves #48 

This PR solves the issue query building and execution errors causing an unrecoverable error without  providing context to the end user.  Errors no longer poison the thread and are now caught and displayed in the interface to help troubleshoot issues with the query. 

  - All store methods return a `Result<_, QueryExecutionError>`.
  - Resolver and StoreResolver return `Result<_, QueryExecutionError>`.
  - All GraphQL execution functions that call the resolver return `Result<q::Value, Vec<GraphQlError>>`.
  - Execute_query and execute_subscription build `QueryResult` using the vector of `QueryExecutionErrors`. 
  - All expect() and unwrap() have been removed. `?` is your friend

### DIFF
--- a/graph/src/data/query/error.rs
+++ b/graph/src/data/query/error.rs
@@ -14,12 +14,17 @@ pub enum QueryExecutionError {
     NoRootQueryObjectType,
     NoRootSubscriptionObjectType,
     ResolveEntityError(Pos, String),
+    ResolveListError,
     NonNullError(Pos, String),
     ListValueError(Pos, String),
     NamedTypeError(String),
+    ObjectFieldError,
     AbstractTypeError(String),
     InvalidArgumentError(Pos, String, q::Value),
     MissingArgumentError(Pos, String),
+    StoreQueryError(String),
+    FilterNotSupportedError(String, String),
+    //    ExecutionErrorList(Vec<QueryExecutionError>),
     UnknownField(Pos, String, String),
     EmptyQuery,
     MultipleSubscriptionFields,
@@ -61,6 +66,7 @@ impl fmt::Display for QueryExecutionError {
             QueryExecutionError::NamedTypeError(s) => {
                 write!(f, "Failed to resolve named type: {}", s)
             }
+            QueryExecutionError::ObjectFieldError => write!(f, "Object field contains 0 types"),
             QueryExecutionError::AbstractTypeError(s) => {
                 write!(f, "Failed to resolve abstract type: {}", s)
             }
@@ -69,6 +75,17 @@ impl fmt::Display for QueryExecutionError {
             }
             QueryExecutionError::MissingArgumentError(_, s) => {
                 write!(f, "No value provided for required argument: {}", s)
+            }
+            QueryExecutionError::StoreQueryError(s) => {
+                write!(f, "Failed to resolve store query, error: {}", s)
+            }
+            QueryExecutionError::FilterNotSupportedError(value, filter) => write!(
+                f,
+                "Value does not support this filter, value: {}, filter: {}",
+                value, filter
+            ),
+            QueryExecutionError::ResolveListError => {
+                write!(f, "Failed to resolve named type within list type")
             }
             QueryExecutionError::UnknownField(_, t, s) => {
                 write!(f, "Type \"{}\" has no field \"{}\"", t, s)
@@ -79,6 +96,12 @@ impl fmt::Display for QueryExecutionError {
                 "Only a single top-level field is allowed in subscriptions"
             ),
         }
+    }
+}
+
+impl From<QueryExecutionError> for Vec<QueryExecutionError> {
+    fn from(e: QueryExecutionError) -> Self {
+        vec![e]
     }
 }
 
--- a/graph/src/data/query/error.rs
+++ b/graph/src/data/query/error.rs
@@ -83,7 +83,7 @@ impl fmt::Display for QueryExecutionError {
                 write!(f, "Failed to get entity from store: {}", s)
             }
             QueryExecutionError::FilterNotSupportedError(value, filter) => {
-                write!(f, "Filter not supported by value {}:{}", value, filter)
+                write!(f, "Filter not supported by value {} : {}", value, filter)
             }
             QueryExecutionError::UnknownField(_, t, s) => {
                 write!(f, "Type \"{}\" has no field \"{}\"", t, s)
@@ -117,7 +117,7 @@ impl fmt::Display for QueryExecutionError {
                 write!(f, "Failed to decode {} value: {}", t, e)
             }
             QueryExecutionError::AttributeTypeError(value, ty) => {
-                write!(f, "Query contains value with invalid type {}:{}", ty, value)
+                write!(f, "Query contains value with invalid type {} : {}", ty, value)
             }
             QueryExecutionError::EntityParseError(s) => {
                 write!(f, "Broken entity found in store: {}", s)
--- a/graph/src/data/query/error.rs
+++ b/graph/src/data/query/error.rs
@@ -1,8 +1,10 @@
+use data::store::Value;
 use graphql_parser::{query as q, Pos};
 use serde::ser::*;
 use std::collections::HashMap;
 use std::error::Error;
 use std::fmt;
+use std::mem::Discriminant;
 use std::string::FromUtf8Error;
 
 /// Error caused while executing a [Query](struct.Query.html).
@@ -31,6 +33,9 @@ pub enum QueryExecutionError {
     SubgraphParseError(String),
     BuildRangeTypeError,
     BuildFilterError,
+    EntityAttributeError,
+    ListTypesError(Discriminant<Value>, Discriminant<Value>),
+    ListFilterError,
 }
 
 impl Error for QueryExecutionError {
@@ -98,9 +103,24 @@ impl fmt::Display for QueryExecutionError {
                 f,
                 "Only a single top-level field is allowed in subscriptions"
             ),
-            QueryExecutionError::SubgraphParseError(s) => write!(f, "Failed to get subgraph ID from type: {}", s),
-            QueryExecutionError::BuildRangeTypeError => write!(f, "Range inputs must be an integer"),
-            QueryExecutionError::BuildFilterError=> write!(f, "Filter must by an object"),
+            QueryExecutionError::SubgraphParseError(s) => {
+                write!(f, "Failed to get subgraph ID from type: {}", s)
+            }
+            QueryExecutionError::BuildRangeTypeError => {
+                write!(f, "Range inputs must be an integer")
+            }
+            QueryExecutionError::BuildFilterError => write!(f, "Filter must by an object"),
+            QueryExecutionError::EntityAttributeError => {
+                write!(f, "Attribute does not belong to entity")
+            }
+            QueryExecutionError::ListTypesError(t, d) => {
+                write!(f, "List field contains inconsistent Value types")
+            }
+            QueryExecutionError::ListFilterError => write!(
+                f,
+                "Non-list value resolved for list filter  \
+                 *Hint*: IN and NOT_IN filters take an array of values"
+            ),
         }
     }
 }
--- a/graph/src/data/query/error.rs
+++ b/graph/src/data/query/error.rs
@@ -7,8 +7,8 @@ use std::collections::HashMap;
 use std::error::Error;
 use std::fmt;
 use std::mem::Discriminant;
-use std::string::{FromUtf8Error};
 use std::str::FromStr;
+use std::string::FromUtf8Error;
 
 /// Error caused while executing a [Query](struct.Query.html).
 #[derive(Debug)]
--- a/graph/src/data/query/error.rs
+++ b/graph/src/data/query/error.rs
@@ -29,8 +29,8 @@ pub enum QueryExecutionError {
     SupgraphIdError(String),
     RangeArgumentsError(Vec<String>),
     InvalidFilterError,
-    EntityAttributeError(String, String),
-    ListTypesError(String),
+    EntityFieldError(String, String),
+    ListTypesError(String, Vec<String>),
     ListFilterError(String),
     ValueParseError(String, String),
     AttributeTypeError(String, String),
@@ -83,7 +83,7 @@ impl fmt::Display for QueryExecutionError {
                 write!(f, "Failed to get entity from store: {}", s)
             }
             QueryExecutionError::FilterNotSupportedError(value, filter) => {
-                write!(f, "Filter not supported by value, {}:{}", value, filter)
+                write!(f, "Filter not supported by value {}:{}", value, filter)
             }
             QueryExecutionError::UnknownField(_, t, s) => {
                 write!(f, "Type \"{}\" has no field \"{}\"", t, s)
@@ -96,31 +96,29 @@ impl fmt::Display for QueryExecutionError {
             QueryExecutionError::SupgraphIdError(s) => {
                 write!(f, "Failed to get subgraph ID from type: {}", s)
             }
-            QueryExecutionError::RangeArgumentsError(s) => write!(
-                f,
-                "Range inputs must be an integer, {:}",
-                s.iter().fold(String::new(), |acc, value| acc + &value)
-            ),
+            QueryExecutionError::RangeArgumentsError(s) => {
+                write!(f, "Range arguments must be properly formed integer: {:}", s.join(", "))
+            }
             QueryExecutionError::InvalidFilterError => write!(f, "Filter must by an object"),
-            QueryExecutionError::EntityAttributeError(e, a) => {
+            QueryExecutionError::EntityFieldError(e, a) => {
                 write!(f, "Entity {} has no attribute {}", e, a)
             }
-            QueryExecutionError::ListTypesError(s) => write!(
+
+            QueryExecutionError::ListTypesError(s, v) => write!(
                 f,
-                "Values passed to filter {} must be of the same type but are of different types",
-                s
+                "Values passed to filter {} must be of the same type but are of different types: {}",
+                s,
+                v.join(", ")
             ),
             QueryExecutionError::ListFilterError(s) => {
                 write!(f, "Non-list value passed to {} filter", s)
             }
             QueryExecutionError::ValueParseError(t, e) => {
-                write!(f, "Failed to decode {} value, {}", t, e)
+                write!(f, "Failed to decode {} value: {}", t, e)
             }
-            QueryExecutionError::AttributeTypeError(value, ty) => write!(
-                f,
-                "Query contains attribute with invalid type, {}:{}",
-                value, ty
-            ),
+            QueryExecutionError::AttributeTypeError(value, ty) => {
+                write!(f, "Query contains value with invalid type {}:{}", ty, value)
+            }
             QueryExecutionError::EntityParseError(s) => {
                 write!(f, "Broken entity found in store: {}", s)
             }
--- a/graph/src/data/query/error.rs
+++ b/graph/src/data/query/error.rs
@@ -34,6 +34,7 @@ pub enum QueryExecutionError {
     ListFilterError(String),
     ValueParseError(String, String),
     AttributeTypeError(String, String),
+    EntityParseError(String),
 }
 
 impl Error for QueryExecutionError {
@@ -79,7 +80,7 @@ impl fmt::Display for QueryExecutionError {
                 write!(f, "No value provided for required argument: {}", s)
             }
             QueryExecutionError::ResolveEntitiesError(s) => {
-                write!(f, "Failed to resolve entities: {}", s)
+                write!(f, "Failed to get entity from store: {}", s)
             }
             QueryExecutionError::FilterNotSupportedError(value, filter) => {
                 write!(f, "Filter not supported by value, {}:{}", value, filter)
@@ -120,6 +121,9 @@ impl fmt::Display for QueryExecutionError {
                 "Query contains attribute with invalid type, {}:{}",
                 value, ty
             ),
+            QueryExecutionError::EntityParseError(s) => {
+                write!(f, "Broken entity found in store: {}", s)
+            }
         }
     }
 }
--- a/graph/src/data/query/error.rs
+++ b/graph/src/data/query/error.rs
@@ -36,7 +36,7 @@ pub enum QueryExecutionError {
     BuildRangeTypeError,
     BuildFilterError,
     EntityAttributeError,
-    ListTypesError(Discriminant<Value>, Discriminant<Value>),
+    ListTypesError(),
     ListFilterError,
 }
 
@@ -115,7 +115,7 @@ impl fmt::Display for QueryExecutionError {
             QueryExecutionError::EntityAttributeError => {
                 write!(f, "Attribute does not belong to entity")
             }
-            QueryExecutionError::ListTypesError(t, d) => {
+            QueryExecutionError::ListTypesError() => {
                 write!(f, "List field contains inconsistent Value types")
             }
             QueryExecutionError::ListFilterError => write!(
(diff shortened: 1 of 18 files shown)


graph/src/data/store/mod.rs
```diff
@@ -31,51 +32,58 @@ pub enum Value {
...
-                    .expect("Unable to parse graphql_parser::query::Number into i64")
+                    .ok_or(QueryExecutionError::NamedTypeError("Int".to_string()))?
                     as i32,
             ),
             (query::Value::Float(f), _) => Value::Float(f.to_owned() as f32),
             (query::Value::Boolean(b), _) => Value::Bool(b.to_owned()),
             (query::Value::Null, _) => Value::Null,
-            _ => unimplemented!(),
-        }
+            _ => Value::Null,
```
leodasvacas: Is `Null` really what we want here? Shouldn't we have a "type mismatch" error?

graph/src/data/subscription/error.rs
```diff
@@ -7,6 +7,8 @@ use prelude::QueryExecutionError;
 pub enum SubscriptionError {
     #[fail(display = "GraphQL error: {}", _0)]
     GraphQLError(QueryExecutionError),
+    #[fail(display = "GraphQL errors: {:?}", _0)]
+    GraphQLErrorList(Vec<QueryExecutionError>),
```
leodasvacas: This can be refactored into a single variant `GraphQLError(Vec<QueryExecutionError>)`.

graphql/src/execution/execution.rs
```diff
@@ -104,19 +103,18 @@ where
...
+                    errors.append(&mut e);
                 }
             };
         }
     }
 
-    // If we have result data, wrap it in an output object
-    if !result_map.is_empty() {
-        result.data = Some(q::Value::Object(result_map));
+    if errors.is_empty() & !result_map.is_empty() {
```
leodasvacas: `&` -> `&&`

graphql/src/execution/execution.rs
```diff
@@ -356,27 +354,29 @@ where
             &ctx.schema.document
         },
         type_name,
-    ).ok_or(QueryExecutionError::NamedTypeError(type_name.to_string()))?;
+    ).ok_or(vec![QueryExecutionError::NamedTypeError(
```
leodasvacas: There is `impl From<QueryExecutionError> for Vec<QueryExecutionError>` so you can remove the surrounding `vec!`

graphql/src/execution/execution.rs
```diff
@@ -410,6 +410,14 @@ where
...
 
         _ => unimplemented!(),
+    };
+
+    match value {
+        Ok(v) => Ok(v),
+        Err(e) => Err(vec![
+            e,
+            QueryExecutionError::NamedTypeError(type_name.to_string()),
+        ]),
```
leodasvacas: this is `value.map_err`

graphql/src/execution/execution.rs
```diff
@@ -451,33 +459,39 @@ where
...
+                            &field.name,
+                            field_definition,
+                            &t,
+                            argument_values,
+                        )
+                    };
+                    match object {
+                        Ok(v) => Ok(v),
+                        Err(e) => Err(vec![e]),
+                    }
```
leodasvacas: again `.map_err`

graph/src/data/query/error.rs
```diff
@@ -14,15 +19,25 @@ pub enum QueryExecutionError {
...
+    StoreQueryError(String),
+    FilterNotSupportedError(String, String),
     UnknownField(Pos, String, String),
     EmptyQuery,
     MultipleSubscriptionFields,
+    SubgraphParseError(String),
+    BuildRangeTypeError,
+    BuildFilterError,
+    EntityAttributeError,
+    ListTypesError(),
```
leodasvacas: You probably don't want this `()`.

graphql/src/store/query.rs
```diff
@@ -2,63 +2,85 @@ use graph::prelude::*;
...
+fn build_range(
+    arguments: &HashMap<&q::Name, q::Value>,
+) -> Result<Option<StoreRange>, QueryExecutionError> {
+    let first = match arguments.get(&"first".to_string()) {
+        Some(value) => match value {
+            q::Value::Int(n) => Ok(n.as_i64()),
+            _ => Err(QueryExecutionError::BuildRangeTypeError),
+        },
+        None => Ok(None),
+    }.and_then(|n| match n {
```
leodasvacas: This `and_then` should be just a `map`.

graphql/src/store/query.rs
```diff
@@ -2,63 +2,85 @@ use graph::prelude::*;
...
+        None => Ok(None),
+    });
+
+    let skip = match arguments.get(&"skip".to_string()) {
+        Some(value) => match value {
+            q::Value::Int(n) => Ok(n.as_i64()),
+            _ => Err(QueryExecutionError::BuildRangeTypeError),
+        },
+        None => Ok(None),
+    }.and_then(|n| match n {
```
leodasvacas: again just a `map`.

graphql/src/store/query.rs
```diff
@@ -67,61 +89,79 @@ fn build_filter_from_object(
...
+    })))
 }
 
 /// Parses a list of GraphQL values into a vector of entity attribute values.
-fn list_values(value: Value) -> Vec<Value> {
+fn list_values(value: Value) -> Result<Vec<Value>, QueryExecutionError> {
     match value {
-        Value::List(values) => values,
-        _ => panic!("value is not a list"),
+        Value::List(ref values) if !values.is_empty() => {
```
leodasvacas: Why is an empty list an error?

graph/src/data/query/error.rs
```diff
@@ -78,10 +105,46 @@ impl fmt::Display for QueryExecutionError {
...
+            QueryExecutionError::EntityAttributeError => {
+                write!(f, "Attribute does not belong to entity")
+            }
+            QueryExecutionError::ListTypesError() => {
+                write!(f, "List field contains inconsistent Value types")
+            }
+            QueryExecutionError::ListFilterError => write!(
+                f,
+                "Non-list value resolved for list filter  \
+                 *Hint*: IN and NOT_IN filters take an array of values"
```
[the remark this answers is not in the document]
Jannis: I'm not a fan of the formatting here. Let's do
```
Non-list value passed to list filter. Hint: _in and _not_in filters take a list of values.
```

graph/src/data/query/error.rs
```diff
@@ -70,6 +86,17 @@ impl fmt::Display for QueryExecutionError {
             QueryExecutionError::MissingArgumentError(_, s) => {
                 write!(f, "No value provided for required argument: {}", s)
             }
+            QueryExecutionError::StoreQueryError(s) => {
+                write!(f, "Failed to resolve store query, error: {}", s)
```
Jannis: As far as the user/client is concerned, the store is an implementation detail. Let's make this 
```rust
ResolveEntitiesError(s) => {
  write!(f, "Failed to resolve entities: {}", s)
}
```

store/postgres/src/store.rs
```diff
@@ -283,9 +283,7 @@ impl BasicStore for Store {
         // Add specified filter to query
         if let Some(filter) = query.filter {
             diesel_query = store_filter(diesel_query, filter).map_err(|e| {
-                error!(self.logger, "value does not support this filter";
-                                    "value" => format!("{:?}", e.value),
-                                    "filter" => e.filter)
+                QueryExecutionError::FilterNotSupportedError(format!("{:?}", e.value), e.filter)
```
Jannis: We're not going to send debug-formatted values to clients. We have to come up with something better.

graph/src/data/query/error.rs
```diff
@@ -70,6 +86,17 @@ impl fmt::Display for QueryExecutionError {
...
+            }
+            QueryExecutionError::FilterNotSupportedError(value, filter) => write!(
+                f,
+                "Value does not support this filter, value: {}, filter: {}",
```
Jannis: Let's make this:
```rust
"Filter not supported by value '{}': {}"
```

graph/src/data/query/error.rs
```diff
@@ -14,15 +19,25 @@ pub enum QueryExecutionError {
...
+    ObjectFieldError,
     AbstractTypeError(String),
     InvalidArgumentError(Pos, String, q::Value),
     MissingArgumentError(Pos, String),
+    StoreQueryError(String),
+    FilterNotSupportedError(String, String),
     UnknownField(Pos, String, String),
     EmptyQuery,
     MultipleSubscriptionFields,
+    SubgraphParseError(String),
```
Jannis: The name of the variant is misleading. How about `SubgraphIdError`?

graph/src/data/query/error.rs
```diff
@@ -14,15 +19,25 @@ pub enum QueryExecutionError {
     NoRootQueryObjectType,
     NoRootSubscriptionObjectType,
     ResolveEntityError(Pos, String),
+    ResolveListError,
```
Jannis: This appears to be unused.

graph/src/data/query/error.rs
```diff
@@ -14,15 +19,25 @@ pub enum QueryExecutionError {
...
     NonNullError(Pos, String),
     ListValueError(Pos, String),
     NamedTypeError(String),
+    ObjectFieldError,
```
Jannis: This appears to be unused.

graph/src/data/query/error.rs
```diff
@@ -14,15 +19,25 @@ pub enum QueryExecutionError {
...
     AbstractTypeError(String),
     InvalidArgumentError(Pos, String, q::Value),
     MissingArgumentError(Pos, String),
+    StoreQueryError(String),
+    FilterNotSupportedError(String, String),
     UnknownField(Pos, String, String),
     EmptyQuery,
     MultipleSubscriptionFields,
+    SubgraphParseError(String),
+    BuildRangeTypeError,
```
Jannis: Let's do better here. Let's make this `RangeArgumentsError(Vec<String>)` and add the names of all range parameters that are not integers.

graph/src/data/query/error.rs
```diff
@@ -14,15 +19,25 @@ pub enum QueryExecutionError {
...
+    BuildFilterError,
```
Jannis: Make this `InvalidFilterError`.

graph/src/data/query/error.rs
```diff
@@ -14,15 +19,25 @@ pub enum QueryExecutionError {
...
+    EntityAttributeError,
```
Jannis: Let's make this report the entity type and attribute name like so:
```rust
EntityAttributeError(String, String)
```
and use this for the error message:
```rust
"Entity '{}' has no attribute '{}'"
```

graph/src/data/query/error.rs
```diff
@@ -78,10 +105,46 @@ impl fmt::Display for QueryExecutionError {
...
+            }
+            QueryExecutionError::BuildRangeTypeError => {
+                write!(f, "Range inputs must be an integer")
+            }
+            QueryExecutionError::BuildFilterError => write!(f, "Filter must by an object"),
+            QueryExecutionError::EntityAttributeError => {
+                write!(f, "Attribute does not belong to entity")
+            }
+            QueryExecutionError::ListTypesError() => {
+                write!(f, "List field contains inconsistent Value types")
```
Jannis: Users/clients don't know about `Value`, so we shouldn't mention the type here. Secondly, this is not about a `list field`, it's about the values in a list passed to a filter. Thirdly, we should report which filter this is about.

So:
```rust
ListTypesError(String)
```
Then, make the error message:
```rust
"Values passed to filter '{}' must be of the same type but are of different types"
```

graph/src/data/query/error.rs
```diff
@@ -78,10 +105,46 @@ impl fmt::Display for QueryExecutionError {
...
 
+impl From<QueryExecutionError> for Vec<QueryExecutionError> {
+    fn from(e: QueryExecutionError) -> Self {
+        vec![e]
+    }
+}
+
+impl From<FromHexError> for QueryExecutionError {
+    fn from(_e: FromHexError) -> Self {
+        QueryExecutionError::NamedTypeError("Bytes".to_string())
```
Jannis: This is not correct. `NamedTypeError` is used for when a type cannot be resolved. This is a case of an invalid value provided for type `Bytes`.

graph/src/data/query/error.rs
```diff
@@ -78,10 +105,46 @@ impl fmt::Display for QueryExecutionError {
...
+    }
+}
+
+impl From<num_bigint::ParseBigIntError> for QueryExecutionError {
+    fn from(_e: num_bigint::ParseBigIntError) -> Self {
+        QueryExecutionError::NamedTypeError("BigInt".to_string())
```
Jannis: This is also not correct (see the remark about `Bytes`).

graph/src/data/query/error.rs
```diff
@@ -70,6 +79,12 @@ impl fmt::Display for QueryExecutionError {
             QueryExecutionError::MissingArgumentError(_, s) => {
                 write!(f, "No value provided for required argument: {}", s)
             }
+            QueryExecutionError::ResolveEntitiesError(s) => {
+                write!(f, "Failed to get entity from store: {}", s)
+            }
+            QueryExecutionError::FilterNotSupportedError(value, filter) => {
+                write!(f, "Filter not supported by value, {}:{}", value, filter)
```
Jannis: Formatting: `by value, {}:{}` -> `by value {}: {}`.

graph/src/data/query/error.rs
```diff
@@ -78,10 +93,59 @@ impl fmt::Display for QueryExecutionError {
                 f,
                 "Only a single top-level field is allowed in subscriptions"
             ),
+            QueryExecutionError::SupgraphIdError(s) => {
+                write!(f, "Failed to get subgraph ID from type: {}", s)
+            }
+            QueryExecutionError::RangeArgumentsError(s) => write!(
+                f,
+                "Range inputs must be an integer, {:}",
```
Jannis: Three improvements here:

* Use a `:` instead of a `,` to separate the higher-level message from the list of arguments.
* Call them `arguments`, not `inputs`.
* Use `graph_graphql::schema::validation::Strings` to obtain a CSV-formatted list of strings instead of using `fold`: https://github.com/graphprotocol/graph-node/blob/4e0478648230c32735d4c02ca8208a9421b67f25/graphql/src/schema/validation.rs#L7-L15 (we should move this into `graph::util` but for now pulling it from `validation` is ok).

graph/src/data/query/error.rs
```diff
@@ -78,10 +93,59 @@ impl fmt::Display for QueryExecutionError {
...
+                s.iter().fold(String::new(), |acc, value| acc + &value)
+            ),
+            QueryExecutionError::InvalidFilterError => write!(f, "Filter must by an object"),
+            QueryExecutionError::EntityAttributeError(e, a) => {
+                write!(f, "Entity {} has no attribute {}", e, a)
+            }
+            QueryExecutionError::ListTypesError(s) => write!(
+                f,
+                "Values passed to filter {} must be of the same type but are of different types",
```
Jannis: Add a `: {}` at the end here, use `Strings` for the values.

graph/src/data/query/error.rs
```diff
@@ -78,10 +93,59 @@ impl fmt::Display for QueryExecutionError {
...
+                s
+            ),
+            QueryExecutionError::ListFilterError(s) => {
+                write!(f, "Non-list value passed to {} filter", s)
+            }
+            QueryExecutionError::ValueParseError(t, e) => {
+                write!(f, "Failed to decode {} value, {}", t, e)
```
Jannis: `:` instead of `,`

graph/src/data/query/error.rs
```diff
@@ -78,10 +93,59 @@ impl fmt::Display for QueryExecutionError {
...
+            }
+            QueryExecutionError::AttributeTypeError(value, ty) => write!(
+                f,
+                "Query contains attribute with invalid type, {}:{}",
```
Jannis: `Attributes` are not a thing here. Make this `Query contains value with invalid type {}: {}`. The first is the type, the second is the value, so swap the other arguments to `write!`.

graph/src/data/store/mod.rs
```diff
@@ -31,51 +33,82 @@ pub enum Value {
...
+                    ty.to_string(),
+                ))
+            }
+        })
+    }
+}
+
+impl fmt::Display for Value {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let printable = match self {
```
Jannis: Wherever possible, it's good to not establish new names/variables. This one is a pretty simple case, so I'm ok with it. You could avoid it with
```
write!(f, "{}", match self {
  ...
})
```

graphql/src/store/query.rs
```diff
@@ -2,126 +2,182 @@ use graph::prelude::*;
...
-    let skip = arguments
-        .get(&"skip".to_string())
-        .and_then(|value| match value {
-            q::Value::Int(n) => n.as_i64(),
-            _ => None,
-        }).and_then(|n| if n >= 0 { Some(n as usize) } else { None });
+fn build_range(
+    arguments: &HashMap<&q::Name, q::Value>,
+) -> Result<Option<StoreRange>, QueryExecutionError> {
+    let first = match arguments.get(&"first".to_string()) {
```
Jannis: Use `map` here instead of `match`.

graphql/src/store/query.rs
```diff
@@ -2,126 +2,182 @@ use graph::prelude::*;
...
+        Some(value) => match value {
+            q::Value::Int(n) => Ok(n.as_i64()),
+            _ => Err("first".to_string()),
+        },
+        None => Ok(None),
+    }.map(|n| match n {
```
Jannis: Use `and_then` here instead of `map`.

graphql/src/store/query.rs
```diff
@@ -2,126 +2,182 @@ use graph::prelude::*;
...
+        Some(n) => if n > 0 {
+            Some(n as usize)
+        } else {
+            None
+        },
+        None => None,
+    });
+
+    let skip = match arguments.get(&"skip".to_string()) {
+        Some(value) => match value {
```
Jannis: Same here about `map`/`match` and `and_then`/`map`.

graphql/src/store/query.rs
```diff
@@ -2,126 +2,182 @@ use graph::prelude::*;
...
             .iter()
             .map(|(key, value)| {
                 use self::sast::FilterOp::*;
 
                 let (attribute, op) = sast::parse_field_as_filter(key);
 
-                let field = sast::get_field_type(entity, &attribute)
-                    .expect("attribute does not belong to entity");
+                let field = sast::get_field_type(entity, &attribute).ok_or(
+                    QueryExecutionError::EntityAttributeError(
```
Jannis: Can we call this `EntityFieldError` instead of `EntityAttributeError` and also rename the variable `field_name` instead of `attribute`?

graphql/src/store/query.rs
```diff
@@ -2,126 +2,182 @@ use graph::prelude::*;
...
+fn build_order_by(
+    arguments: &HashMap<&q::Name, q::Value>,
+) -> Result<Option<String>, QueryExecutionError> {
+    Ok(arguments
         .get(&"orderBy".to_string())
         .and_then(|value| match value {
             q::Value::Enum(name) => Some(name.to_owned()),
             _ => None,
-        })
+        }))
```
Jannis: The formatting here strikes me as being off: shouldn't the closing `)` be on its own line, intended like `Ok`? I mean if `rustfmt` does it this way, I'll shut up 😉

graphql/src/store/query.rs
```diff
@@ -2,126 +2,182 @@ use graph::prelude::*;
...
+    arguments: &HashMap<&q::Name, q::Value>,
+) -> Result<Option<StoreOrder>, QueryExecutionError> {
+    Ok(arguments
         .get(&"orderDirection".to_string())
         .and_then(|value| match value {
             q::Value::Enum(name) if name == "asc" => Some(StoreOrder::Ascending),
             q::Value::Enum(name) if name == "desc" => Some(StoreOrder::Descending),
             _ => None,
-        })
+        }))
```
Jannis: Same here

graphql/src/store/query.rs
```diff
@@ -135,7 +191,7 @@ pub fn parse_subgraph_id(entity: &s::ObjectType) -> Option<String> {
         }).and_then(|(_, value)| match value {
             s::Value::String(id) => Some(id),
             _ => None,
-        })
+        }).ok_or(QueryExecutionError::SupgraphIdError(entity.clone().name))
```
Jannis: Do we need to clone the entire entity object type here? Do we have to clone at all here? How about this?
```rust
let entity_name = entity.name.clone();

entity
  .directives()
  ...
  .and_then(|(_, value)| match value {
    s::Value::String(id) => Some(id.clone()),
    _ => None,
  })
  .ok_or(QueryExecutionError::SubgraphIdError(entity_name))
```

store/postgres/src/filter.rs
```diff
@@ -272,6 +272,9 @@ fn store_filter_by_mode<'a>(
         // Is `attribute` equal to some `v` in `query_values`?
         StoreFilter::In(attribute, query_values) => {
             let op = " = ANY (";
+            if query_values.is_empty() {
+                return Ok(add_filter(query, filter_mode, sql("true")));
```
Jannis: I'm curious: what does this do?

graph/src/data/query/error.rs
Jannis: You could also just do `strings.join(", ")` here if you want. It's probably not worth pulling in `Strings`.

store/postgres/src/filter.rs
leodasvacas: This seems like it makes `x_in: []` true, which wouldn't be correct.

graphql/src/store/query.rs
leodasvacas: @fordN Ping so this comment doesn't fall off your radar.

store/postgres/src/filter.rs
fordN: This is similar to writing "1=1", it will always evaluate to true, so it will not filter the results when used in a where clause on its own or when combined with other conditions using `AND`.   
The Postgres query planner should ignore these clauses in the execution plan. 

I was thinking I would use 'false' in the case of filter_mode=Or, but there are some cases where adding `true` or `false` would change the filtering result... so it looks like I need a better solution.

graphql/src/store/query.rs
fordN: I agree, it makes sense to place the closing `)` at the same indent level as the opening `Ok(`...but this is how `rustfmt` likes it these days. 🤷

graph/src/data/query/error.rs
```diff
@@ -83,7 +83,7 @@ impl fmt::Display for QueryExecutionError {
                 write!(f, "Failed to get entity from store: {}", s)
             }
             QueryExecutionError::FilterNotSupportedError(value, filter) => {
-                write!(f, "Filter not supported by value, {}:{}", value, filter)
+                write!(f, "Filter not supported by value {}:{}", value, filter)
```
Jannis: Add space after the `:` here.

graph/src/data/query/error.rs
```diff
@@ -96,31 +96,29 @@ impl fmt::Display for QueryExecutionError {
...
             ),
             QueryExecutionError::ListFilterError(s) => {
                 write!(f, "Non-list value passed to {} filter", s)
             }
             QueryExecutionError::ValueParseError(t, e) => {
-                write!(f, "Failed to decode {} value, {}", t, e)
+                write!(f, "Failed to decode {} value: {}", t, e)
+            }
+            QueryExecutionError::AttributeTypeError(value, ty) => {
+                write!(f, "Query contains value with invalid type {}:{}", ty, value)
```
Jannis: Add a space after the `:` here.

PR #429, opened 2018-09-28
Improve cancellation helpers

### LINKED ISSUE #392: Generic solution for cancelling streams/futures

leodasvacas:
`cancelabe` takes the guard as an argument, so cancelling multiple things with one guard is supported. It's also nice that we no longer return tuples allowing `cancelable` to be chained with other methods.

Resolves #392.

### DIFF
--- a/graph/src/ext/futures.rs
+++ b/graph/src/ext/futures.rs
@@ -57,7 +57,7 @@ pub trait Canceler {
 pub struct CancelGuard {
     /// This is the only non-temporary strong reference to this `Arc`, therefore
     /// the `Vec` should be dropped shortly after `self` is dropped.
-    cancelers: Arc<Mutex<Vec<oneshot::Sender<()>>>>,
+    cancel_senders: Arc<Mutex<Vec<oneshot::Sender<()>>>>,
 }
 
 impl CancelGuard {
@@ -71,14 +71,14 @@ impl CancelGuard {
 
     pub fn handle(&self) -> CancelHandle {
         CancelHandle {
-            guard: Arc::downgrade(&self.cancelers),
+            guard: Arc::downgrade(&self.cancel_senders),
         }
     }
 }
 
 impl Canceler for CancelGuard {
     fn add_canceler(&self, cancel_sender: oneshot::Sender<()>) {
-        self.cancelers.lock().unwrap().push(cancel_sender);
+        self.cancel_senders.lock().unwrap().push(cancel_sender);
     }
 }
 
--- a/graph/src/ext/futures.rs
+++ b/graph/src/ext/futures.rs
@@ -42,20 +42,38 @@ impl<F: Future, C: Fn() -> F::Error> Future for Cancelable<F, C> {
     }
 }
 
-#[derive(Debug)]
+/// A `CancelGuard` or `SharedCancelGuard`.
+pub trait CancelGuardTrait {
+    /// Adds `canceler` to the set being guarded.
+    fn add_canceler(&mut self, canceler: oneshot::Sender<()>);
+}
+
+/// Cancels any guarded futures and streams when dropped.
+#[derive(Debug, Default)]
 pub struct CancelGuard {
-    canceler: oneshot::Sender<()>,
+    cancelers: Vec<oneshot::Sender<()>>,
 }
 
 impl CancelGuard {
+    /// Creates a guard that initially guards nothing.
+    pub fn new() -> Self {
+        Self::default()
+    }
+
     /// A more readable `drop`.
     pub fn cancel(self) {}
 
     /// Convert into a `SharedGuard`.
     pub fn shared(self) -> SharedCancelGuard {
         SharedCancelGuard {
-            guard: Arc::new(Mutex::new(Some(self.canceler))),
+            guard: Arc::new(Mutex::new(Some(self))),
         }
+    }
+}
+
+impl CancelGuardTrait for CancelGuard {
+    fn add_canceler(&mut self, canceler: oneshot::Sender<()>) {
+        self.cancelers.push(canceler);
     }
 }
 
@@ -64,7 +82,7 @@ impl CancelGuard {
 /// To cancel the stream, call `cancel` on one of the guards or drop all guards.
 #[derive(Clone, Debug)]
 pub struct SharedCancelGuard {
-    guard: Arc<Mutex<Option<oneshot::Sender<()>>>>,
+    guard: Arc<Mutex<Option<CancelGuard>>>,
 }
 
 impl SharedCancelGuard {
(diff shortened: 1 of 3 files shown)


graph/src/ext/futures.rs
```diff
@@ -42,29 +42,47 @@ impl<F: Future, C: Fn() -> F::Error> Future for Cancelable<F, C> {
     }
 }
 
-#[derive(Debug)]
+/// A `CancelGuard` or `SharedCancelGuard`.
+pub trait CancelGuardTrait {
+    /// Adds `canceler` to the set being guarded.
+    fn add_canceler(&mut self, canceler: oneshot::Sender<()>);
+}
```
timmclean: It looks like CancelGuard is missing `is_canceled` -- should that go in the trait?

timmclean:
I like it! I'm coming around to the SharedCancelGuard vs CancelGuard separation too.

I want to work through the usage a bit to make sure I understand what this would look like. A primary use case here is being able to stop an operation composed of a number of sub-steps, which means cancelling each of the inner steps. So, I imagine that looking like this:

```rust
fn do_long_operation(cancel_guard: &mut impl CancelGuard) -> impl Future<...> {
  get_items()
    .cancelable(cancel_guard, || ())
    .and_then(|item| {
      load_item(item, "some URL", cancel_guard)
    }).collect()
    .and_then(|all_items| whatever(all_items))
    .cancelable(cancel_guard, || ())
}

fn load_item(item: ItemId, url: &str, cancel_guard: &mut impl CancelGuard) -> impl Future <...> {
  lookup_location(item, url)
    .cancelable(cancel_guard, || ())
    .and_then(|item_location| download(item_location))
    .cancelable(cancel_guard, || ())
}
```

Now, `cancel_guard` won't live long enough to be passed to `load_item`, so that's a problem. So I guess the first function should take `Arc` or `SharedCancelGuard` instead of a `&mut`.

Let's say we replace `&mut impl CancelGuard` with `SharedCancelGuard` in the first function signature. Now we can clone the guard, move it into closures, etc. One interesting thing that I notice here is that the caller of `do_long_operation` cannot rely on cancel-on-drop semantics. A clone of the guard will be kept alive in the closure until partway through the operation.

I don't think it's really a big problem, just an oddity that sticks out to me, so I thought I should bring it up. I think we should plan not to rely on the cancel-on-drop semantics of SharedCancelGuard because it won't work properly with complex operations -- always use `.cancel()` instead.

This leads me to wonder: what happens if a SharedCancelGuard is cancelled just before a new sub-operation is started and made cancellable, using the same cancel guard? Is there a race condition where in SharedCancelGuard the guard Option could be None when `add_canceler` is called?

Last thought on usage. Here's how it looks to call a cancellable function if you don't care that it's cancellable:

```rust
let cancel_guard = CancelGuard::new();
do_long_operation(&mut cancel_guard)
  .and_then(move |_| {
    // Need to move cancel guard into closure to keep it alive
    let _ = cancel_guard;
  })
```

I suppose `mem::forget` is also an option? It feels a bit unergonomic either way, but I can live with it if it's saving implementation complexity.

OK, I'll stop rambling now :)

graph/src/ext/futures.rs
leodasvacas: `CancelGuard` is never canceled, it only cancels on drop.

leodasvacas:
@timmclean Thanks for walking through some use cases! The `CancelGuardTrait` is not meant to be a general abstraction over `CancelGuard` and `SharedCancelGuard`, but I agree with your observation that `CancelGuard` coulnd't be used on the first one, I've added `CancelHandle` which works as a weak reference to a `CancelGuard` and can be cloned into closures.

I also agree that cancel-on-drop for the shared guard isn't very reliable and `.cancel()` should be used.

> Is there a race condition where in SharedCancelGuard the guard Option could be None when add_canceler is called?

I've clarified with an explicit `drop` that the cancelation will be immediate in that case.

The use case for not caring about cancelation is very legitimate, I've added `DummyCancelGuard` to function as a guard that never cancels. I had ideas for dynamically making guards uncancelable, but it seems like an unnecessary complication for now.

Jannis:
@timmclean Can we prioritize this work (and the work towards using cancellable to terminate subgraphs) over any performance tuning? Canceling subgraphs when they are removed or redeploy is crucial.

graph/src/ext/futures.rs
timmclean: Makes sense!

graph/src/ext/futures.rs
```diff
@@ -1,13 +1,13 @@
...
 
 /// A cancelable stream or future.
 ///
 /// It can be canceled through the corresponding `CancelGuard`.
 pub struct Cancelable<T, C> {
     cancelable: T,
-    canceled: oneshot::Receiver<()>,
+    canceled: Fuse<oneshot::Receiver<()>>,
     on_cancel: C,
 }
```
timmclean: How about something like `cancel_receiver` and `cancel_sender`?

graph/src/ext/futures.rs
```diff
@@ -1,13 +1,13 @@
 use futures::sync::oneshot;
-use std::sync::{Arc, Mutex};
-use tokio::prelude::{Future, Poll, Stream};
+use std::sync::{Arc, Mutex, Weak};
+use tokio::prelude::{future::Fuse, Future, Poll, Stream};
 
 /// A cancelable stream or future.
 ///
 /// It can be canceled through the corresponding `CancelGuard`.
```
timmclean: A couple other details would be useful here, how about "This type wraps a stream or future to make it cancelable." and a hint about creating these using `.cancelable` in the FutureExtension/StreamExtension?

graph/src/ext/futures.rs
```diff
@@ -1,13 +1,13 @@
...
 pub struct Cancelable<T, C> {
     cancelable: T,
```
timmclean: `cancelable` -> `inner` or `original`, maybe?

graph/src/ext/futures.rs
```diff
@@ -42,32 +44,79 @@ impl<F: Future, C: Fn() -> F::Error> Future for Cancelable<F, C> {
     }
 }
 
-#[derive(Debug)]
+/// A `CancelGuard` or `SharedCancelGuard`.
+pub trait CancelGuardTrait {
+    /// Adds `canceler` to the set being guarded.
+    fn add_canceler(&self, canceler: oneshot::Sender<()>);
+}
```
timmclean: Similar thought here on `canceler` -> `cancel_sender`.

This method is only used internally by the helpers, right? Maybe add a comment so that new API users don't get confused and try to use it directly.

It seems like this could also be implemented by types that are not scope guards. How about `CancelGuardTrait` -> `Canceler`?

leodasvacas:
@timmclean Not sure if I got all your suggestions, but I tried addressing them in the latest commit.

graph/src/ext/futures.rs
timmclean: one more: `add_canceler` -> `add_cancel_sender`

graph/src/ext/futures.rs
```diff
@@ -33,41 +35,90 @@ impl<F: Future, C: Fn() -> F::Error> Future for Cancelable<F, C> {
...
+    fn add_canceler(&self, cancel_sender: oneshot::Sender<()>);
+}
+
+/// Cancels any guarded futures and streams when dropped.
+#[derive(Debug, Default)]
 pub struct CancelGuard {
-    canceler: oneshot::Sender<()>,
+    /// This is the only non-temporary strong reference to this `Arc`, therefore
+    /// the `Vec` should be dropped shortly after `self` is dropped.
+    cancelers: Arc<Mutex<Vec<oneshot::Sender<()>>>>,
```
timmclean: how about cancel_senders?

leodasvacas: @timmclean done

PR #440, opened 2018-10-02
Less logging in block ingestor

timmclean:
Infura often doesn't have transaction receipts right away for new blocks, and that results in lots of warnings getting logged during normal operation

### DIFF
--- a/graph/src/util/futures.rs
+++ b/graph/src/util/futures.rs
@@ -52,7 +52,7 @@ where
 
         try_it().map_err(move |e| {
             if attempt_count >= log_after {
-                warn!(
+                debug!(
                     logger,
                     "Trying again in ~{} seconds after {} failed (attempt #{})",
                     (delay_ms as f64 / 1000.0).round(),
--- a/graph/src/util/futures.rs
+++ b/graph/src/util/futures.rs
@@ -43,3 +43,18 @@ where
         RetryError::TimerError(e) => panic!("tokio timer error: {}", e),
     })
 }
+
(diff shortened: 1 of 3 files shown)


graph/src/util/futures.rs
```diff
@@ -43,3 +43,18 @@ where
         RetryError::TimerError(e) => panic!("tokio timer error: {}", e),
     })
 }
+
+pub fn with_retry_no_logging<F, R, I, E>(try_it: F) -> impl Future<Item = I, Error = E> + Send
```
Jannis: Can we maybe log every 10 or 20 retries so that we can see that it is still trying?

Jannis (changes requested):

graph/src/util/futures.rs
```diff
@@ -28,13 +51,15 @@ where
...
         try_it().map_err(move |e| {
-            warn!(
-                logger,
-                "Trying again in ~{} seconds after {} failed (attempt #{})",
-                (delay_ms as f64 / 1000.0).round(),
-                &operation_name,
-                attempt_count + 1,
-            );
+            if attempt_count >= log_after {
+                warn!(
```
Jannis: Can we make this a `debug!` instead of `warn!`? See my earlier comment on the PR.

Jannis (changes requested):
My only request is a super-small logging change. The rest looks great.

graph/src/util/futures.rs
timmclean: I like it! Done.

leoyvens (dismissed):
lgtm

Jannis (dismissed):

Jannis (approved):

PR #444, opened 2018-10-02
Fix bytes filters

### LINKED ISSUE #415: GraphQL Filters are broken for Bytes fields

Bug

Filtering on a `Bytes` field with e.g.
```graphql
query {
  auctions(where: { owner_contains: "db" }) {
    id
    owner
  }
}
```

should return auctions that have an owner value that is any of `"0x.*db.*"`. However, queries like the above always return no matches.

The reason:

Internally, the `"db"` string is converted to a `Bytes` value (because that's the type of the `owner_contains` filter field). We then convert that to a string again for Postgres (`data->>owner LIKE "..."`). Due to the `Display` implementation for `Bytes`, the `Bytes([219])` value is converted to `"0xdb"`. The resulting SQL query has `data->>owner LIKE "0xdb"`, which is incorrect.

Interestingly, this also doesn't match auctions with owners that start with `0xdb`.

Auctions whose owner contains `db` anywhere in its hex string representation are returned from the query.

leodasvacas:
Fixes #415. I suspected that list equality was also affected but it's not.

I added a test but store tests are currently borked.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -814,6 +814,7 @@ dependencies = [
  "fallible-iterator 0.1.5 (registry+https://github.com/rust-lang/crates.io-index)",
  "futures 0.1.23 (registry+https://github.com/rust-lang/crates.io-index)",
  "graph 0.1.0",
+ "hex 0.3.2 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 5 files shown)


Jannis:
I can confirm that it works:
image

PR #445, opened 2018-10-02
Faster block loading

### LINKED ISSUE #425: Investigate loading BlockStream blocks asynchronously
Instead of pausing to load a set of blocks, it might be faster to stream blocks one at a time, while trying to overload processing of a block with loading of the next block.

Also investigate: batching web3 requests for blocks

timmclean:
Overlay block loading with block processing to keep as many Ethereum RPC calls in flight as possible. Also, increase Web3 parallel request limit from 64 to 2048.

Resolves #425

### DIFF
--- a/datasource/ethereum/src/block_stream.rs
+++ b/datasource/ethereum/src/block_stream.rs
@@ -58,7 +58,7 @@ enum ReconciliationStep {
     /// Move forwards, processing one or more blocks.
     ProcessDescendantBlocks {
         from: EthereumBlockPointer,
-        descendant_blocks: Vec<EthereumBlock>,
+        descendant_blocks: Box<Stream<Item = EthereumBlock, Error = Error> + Send>,
     },
 
     /// Skip forwards from `from` to `to`, with no processing needed.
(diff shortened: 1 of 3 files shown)


datasource/ethereum/src/ethereum_adapter.rs
```diff
@@ -366,11 +366,6 @@ where
                         1 => {
                             let mut block_ptrs = vec![];
                             for log in chunks[0].iter() {
-                                if block_ptrs.len() >= 100 {
-                                    // That's enough to process in one iteration
-                                    break;
-                                }
-
```
timmclean: This chunking now happens in `load_blocks` in `block_stream.rs`

datasource/ethereum/src/block_stream.rs
```diff
@@ -355,22 +355,17 @@ where
...
-                                                descendant_blocks,
-                                            }
-                                        })
-                                    )
+                                    let descendant_hashes = descendant_ptrs.into_iter().map(|ptr| ptr.hash).collect();
+                                    Box::new(future::ok(
+                                        // Proceed to those blocks
+                                        ReconciliationStep::ProcessDescendantBlocks {
+                                            from: subgraph_ptr,
+                                            descendant_blocks: Box::new(ctx.load_blocks(descendant_hashes)),
```
Jannis: Good idea to do this instead of individual `load_block`s!

datasource/ethereum/src/block_stream.rs
```diff
@@ -499,59 +494,78 @@ where
...
+    ) -> impl Stream<Item = EthereumBlock, Error = Error> + Send {
+        let ctx = self.clone();
+
+        let block_hashes_batches = block_hashes
+            .chunks(200) // max batch size
+            .map(|chunk| chunk.to_vec())
+            .collect::<Vec<_>>();
+
+        // Return a stream that lazily loads batches of blocks
+        stream::iter_ok::<_, Error>(block_hashes_batches)
```
Jannis: The batching + chunking logic here is pretty darn elegant.

datasource/ethereum/src/ethereum_adapter.rs
Jannis: Noted!

datasource/ethereum/src/block_stream.rs
timmclean: Thanks :blush:

PR #447, opened 2018-10-03
travis: Upload binaries to github releases

### LINKED ISSUE #403: Implement process for building binaries against GitHub release tags

leodasvacas:
This sets things up so that we make a github release whenever a tag is pushed and attach a linux binary to it, built in release mode. We'll also want osx but I'd rather leave that to a follow up PR since we don't currently test on osx.

[This test](https://github.com/graphprotocol/graph-node/releases/tag/TEST_TAG) shows that it works.

Part of #403

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,14 +27,17 @@ before_install:
 
 matrix:
   fast_finish: true
-# Allow nightly to fail
+  # Allow nightly to fail
   allow_failures:
     - rust: nightly
     - rust: beta # until ipfs-api is un-regressed
   # Only check formatting when running against stable
   include:
     - env: CHECK_FORMATTING=true
-      if: group = stable
+  # Build tagged commits in release mode
+  include:
+    - env: RELEASE=true
+      if: tag IS present
 
 # Cache dependencies
 cache: cargo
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -27,14 +27,17 @@ before_install:
...
   fast_finish: true
-# Allow nightly to fail
+  # Allow nightly to fail
   allow_failures:
     - rust: nightly
     - rust: beta # until ipfs-api is un-regressed
   # Only check formatting when running against stable
   include:
     - env: CHECK_FORMATTING=true
-      if: group = stable
```
Jannis: If we do this in other releases, won't the format checks fail because other releases have other formatting rules?

.travis.yml
```diff
@@ -57,5 +60,20 @@ script:
...
     fi
-  # Run unit and sccenario tests
+  # Run tests
   - RUST_BACKTRACE=1 cargo test --verbose --all -- --nocapture
+  - if [ "$RELEASE" = "true" ]; then
+      cargo build -p graph-node --release;
+      mv target/release/graph-node target/release/graph-node-$TRAVIS_OS_NAME;
+    fi
+
+
```
Jannis: One empty line here should be enough.

.travis.yml
```diff
@@ -57,5 +60,20 @@ script:
...
+deploy:
+  provider: releases
+  api_key:
+    secure: ygpZedRG+/Qg/lPhifyNQ+4rExjZ4nGyJjB4DYT1fuePMyKXfiCPGicaWRGR3ZnZGNRjdKaIkF97vBsZ0aHwW+AykwOxlXrkAFvCKA0Tb82vaYqCLrBs/Y5AEhuCWLFDz5cXDPMkptf+uLX/s3JCF0Mxo5EBN2JfBQ8vS6ScKEwqn2TiLLBQKTQ4658TFM4H5KiXktpyVVdlRvpoS3pRIPMqNU/QpGPQigaiKyYD5+azCrAXeaKT9bBS1njVbxI69Go4nraWZn7wIhZCrwJ+MxGNTOxwasypsWm/u1umhRVLM1rL2i7RRqkIvzwn22YMaU7FZKCx8huXcj0cB8NtHZSw7GhJDDDv3e7puZxl3m/c/7ks76UF95syLzoM/9FWEFew8Ti+5MApzKQj5YWHOCIEzBWPeqAcA8Y+Az7w2h1ZgNbjDgSvjGAFSpE8m+SM0A2TOOZ1g/t/yfbEl8CWO6Y8v2x1EONkp7X0CqJgASMp+h8kzKCbuYyRnghlToY+5wYuh4M9Qg9UeJCt9dOblRBVJwW5CFr62kgE/gso8F9tXXHkRTv3hfk5madZR1Vn5A7KadEO8epfV4IQNsd+VHfoxoJSprx5f77Q2bLMBD1GT/qMqECgSznoTkU5ajkKJRqUw4AwLTohrYir76j61eQfxOhXExY/EM8xvlxpd1w=
```
Jannis: How "secure" is this? Should we inject this key via a Travis environment variable?

.travis.yml
leodasvacas: That option is not mandatory, by default it will be set to the first config of the list which is stable. I was having trouble doing `stable && tagged` logic and just removing the stable condition fixed it. I can try putting it back, this one in particular is trivial to put back, if you think it's more readable.

.travis.yml
leodasvacas: This is an [encrypted variable](https://docs.travis-ci.com/user/environment-variables/#defining-encrypted-variables-in-travisyml) which is how travis recommends doing this because the encrypted value is tied to this repo, so we're good.

.travis.yml
Jannis: We don't need to if it works.

.travis.yml
Jannis: Ok!

PR #470, opened 2018-10-07
Store.set() merges id into the entity before pushing the `EntityOperation`

### LINKED ISSUE #386: Entity IDs have to set twice and forgetting that results in non-obvious errors

Bug

This has come up a few times. When writing an entity to the store with `store.set`, developers have to provide the entity ID as the second argument, e.g.
```js
store.set('MyType', id, entityData)
```

However, if they forget to also do
```
entityData.setString('id', ...)
```
before that, the entity won't be valid. It will even break queries and throw a
```
Null value resolved for non-null field: myTypes
```
GraphQL error. This is very confusing and non-obvious.

First of all, `setString('id', ...)` should not be necessary. Since we _aways_ get the ID via `store.set`, we should add it to the entity behind the scenes. Our host-exported implementation of `store.set` should merge this ID into the entity before it writes it to the store.

This way developers can't write broken entities to the store without it being their own "fault".

fordN:
Resolves #386

`Store.set()` uses input to add an `id` attribute to the `entity` or if already exists ensures they  are equal.

### DIFF
--- a/runtime/wasm/src/module/mod.rs
+++ b/runtime/wasm/src/module/mod.rs
@@ -293,7 +293,17 @@ where
     ) -> Result<Option<RuntimeValue>, Trap> {
         let entity: String = self.heap.asc_get(entity_ptr);
         let id: String = self.heap.asc_get(id_ptr);
-        let data: HashMap<String, Value> = self.heap.asc_get(data_ptr);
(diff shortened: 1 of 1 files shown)


Jannis:
@fordN This is ready to merge.

PR #476, opened 2018-10-11
datasource, graph: Add retry deadline

timmclean:
It looks like the tokio retry crate doesn't do as much as I thought it did, so this adds a 60-second `deadline` to the `with_retry` util, which should help with Eth RPC calls getting stuck sometimes.

### DIFF
--- a/datasource/ethereum/src/ethereum_adapter.rs
+++ b/datasource/ethereum/src/ethereum_adapter.rs
@@ -46,6 +46,7 @@ where
         event_signatures: Vec<H256>,
     ) -> impl Future<Item = Vec<Log>, Error = Error> {
         let eth_adapter = self.clone();
+        let event_sig_count = event_signatures.len();
         with_retry(
             self.logger.clone(),
             "eth_getLogs RPC call".to_owned(),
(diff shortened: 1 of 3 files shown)


datasource/ethereum/src/ethereum_adapter.rs
```diff
@@ -69,7 +70,11 @@ where
                     }).map_err(SyncFailure::new)
                     .from_err()
             },
-        )
+        ).map_err(move |e| {
+            e.into_inner().unwrap_or_else(move || {
+                format_err!("Ethereum node took too long to respond to eth_getLogs (from = {}, to = {}, # of event sigs = {})", from, to, event_sig_count)
```
Jannis: This line is a _bit_ too long. Wrap the string around with e.g.
```rust
"Ethereum node took too long to respond to eth_getLogs \
 (from block {}, to block {}, {} event signatures)"
```

Also, note: I've changed the error message a little. I think it is a little easier to read this way.

datasource/ethereum/src/ethereum_adapter.rs
```diff
@@ -301,7 +322,14 @@ where
                                         })
                                     })
                             },
-                        )
+                        ).map_err(move |e| {
+                            e.into_inner().unwrap_or_else(move || {
+                                format_err!(
+                                    "Ethereum node took too long to return transaction receipt {}",
```
Jannis: This line also may be too long. Try to wrap the string into two lines.

Jannis:
I've changed the deadline to 30s and am merging this into `ethsf` so others can try this asap.

PR #477, opened 2018-10-11
datasource, graph: Add retry deadline

timmclean:
Same as #476 but for master branch.

It looks like the tokio retry crate doesn't do as much as I thought it did, so this adds a 60-second `deadline` to the `with_retry` util, which should help with Eth RPC calls getting stuck sometimes.

### DIFF
--- a/graph/src/util/futures.rs
+++ b/graph/src/util/futures.rs
@@ -1,6 +1,7 @@
 use slog::Logger;
-use std::time::Duration;
+use std::time::{Duration, Instant};
 use tokio::prelude::*;
+use tokio::timer::DeadlineError;
 use tokio_retry::strategy::{jitter, ExponentialBackoff};
 use tokio_retry::Error as RetryError;
 use tokio_retry::Retry;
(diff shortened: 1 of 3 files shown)


graph/src/util/futures.rs
```diff
@@ -52,21 +53,34 @@ where
...
-                debug!(
-                    logger,
-                    "Trying again in ~{} seconds after {} failed (attempt #{})",
-                    (delay_ms as f64 / 1000.0).round(),
-                    &operation_name,
-                    attempt_count + 1,
-                );
-            }
+        try_it()
+            .deadline(Instant::now() + Duration::from_secs(60))
```
Jannis: 60 seconds is twice as long as most network timeouts in other systems (which are usually 30s). I'd even make it less in the hopes of speeding things up. 10s? 20s? But at most 30s.

timmclean: Hmmm I did some testing with 30s and it actually ends up preventing some calls from completing. In particular, I saw eth_getLogs consistently take over 40 seconds... it's long but I think 60s would be safer

Jannis: Ok, that's fair enough. Better than nothing!

timmclean:
Brought over the suggested line wrapping from the other PR :)

PR #478, opened 2018-10-11
datasource, store: Fix block ingestion on short chains

timmclean:
Same as https://github.com/graphprotocol/graph-node/pull/469 but for master branch


Jannis:
@timmclean Something's up with Travis though.

PR #480, opened 2018-10-11
Store::get and Store::get_entity return Result<Option<Entity>, Error>

### LINKED ISSUE #423: Return Result<Option<Entity>, Error> from Store::get
This allows is to improve `store.get` to return the entity or `null` and an error if there really is one.

fordN:
Resolves #423 

`Store::get` and `Store::get_entity` now return `Result<Option<Entity>, Error>` allowing handling of query errors and returning `None` if no data is found.


Jannis:
Obsoleted by #494.

PR #484, opened 2018-10-11
datasource: Add BLOCK_BATCH_SIZE & MAX_PARALLEL_HTTP env var (master)

timmclean:


leodasvacas:
This looks like "too many knobs" territory, will we want to keep this configs around or are they temporary? If we do want to keep them around a CLI param would probably be preferable over an env var.

timmclean:
Yeah, it's looking like for performance we're going to need to be able to tweak things like this based on what the Ethereum node can handle. E.g. with infura, we can send a ton of requests at once, which they'll load balance internally, but with a single node we have to be more gentle

Jannis:
@leodasvacas Long-term, I'd say we want to follow what Parity does: support configuration through a file like
```toml
[ethereum]
block-chunk-size = 50

[ethereum.rpc]
max-parallel-requests = 2048
```
and have corresponding CLI flags like
```
--ethereum-block-chunk-size 50
--ethereum-rpc-max-parallel-requests 2048
```
I've created #486 to track this.

For the moment, we don't want to expose these knobs to everyone, that's why they are only accepted through env vars.

PR #492, opened 2018-10-12
datasource, graph, mock, store: Fix block ingestor stuck on missing tx

### LINKED ISSUE #491: Block ingestor gets stuck on transactions from uncled blocks
Sometimes a transaction appears in a block, but that block is removed due to a chain reorganization. Usually the transaction will be included in a later block, but not always. In the latter case, the block ingestor might get stuck trying to load a receipt for this transaction.

See also #490

timmclean:
Resolves #491

### DIFF
--- a/datasource/ethereum/src/block_ingestor.rs
+++ b/datasource/ethereum/src/block_ingestor.rs
@@ -303,13 +303,18 @@ where
     }
 }
 
-#[derive(Debug)]
+#[derive(Debug, Fail)]
 enum BlockIngestorError {
     /// The Ethereum node does not know about this block for some reason, probably because it
     /// disappeared in a chain reorg.
+    #[fail(
+        display = "block data unavailable, block was likely uncled (block hash = {:?})",
+        _0
+    )]
     BlockUnavailable(H256),
 
     /// An unexpected error occurred.
+    #[fail(display = "error in block ingestor: {}", _0)]
     Unknown(Error),
 }
 
--- a/datasource/ethereum/src/block_ingestor.rs
+++ b/datasource/ethereum/src/block_ingestor.rs
@@ -60,10 +60,22 @@ where
                 static_self.do_poll().then(move |result| {
                     if let Err(e) = result {
                         // Some polls will fail due to transient issues
-                        warn!(
-                            static_self.logger,
-                            "Trying again after block polling failed: {:?}", e
-                        );
+                        match e {
+                            BlockIngestorError::BlockUnavailable(block_hash) => {
+                                trace!(
+                                    static_self.logger,
+                                    "Trying again after block polling failed: \
+                                    block data unavailable, block was likely uncled";
+                                    "block_hash" => format!("{:?}", block_hash)
+                                );
+                            }
+                            BlockIngestorError::Unknown(e) => {
+                                warn!(
+                                    static_self.logger,
+                                    "Trying again after block polling failed: {:?}", e
+                                );
+                            }
+                        }
                     }
 
                     // Continue polling even if polling failed
@@ -72,11 +84,12 @@ where
             })
     }
 
-    fn do_poll<'a>(&'a self) -> impl Future<Item = (), Error = Error> + 'a {
+    fn do_poll<'a>(&'a self) -> impl Future<Item = (), Error = BlockIngestorError> + 'a {
         trace!(self.logger, "BlockIngestor::do_poll");
 
         // Get chain head ptr from store
         future::result(self.chain_store.chain_head_ptr())
+            .from_err()
             .and_then(move |head_block_ptr_opt| {
                 // Ask for latest block from Ethereum node
                 self.get_latest_block()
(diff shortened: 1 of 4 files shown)


Jannis (dismissed):
Looks good!

datasource/ethereum/src/block_ingestor.rs
```diff
@@ -60,10 +60,22 @@ where
...
-                            static_self.logger,
-                            "Trying again after block polling failed: {:?}", e
-                        );
+                        match e {
+                            BlockIngestorError::BlockUnavailable(block_hash) => {
+                                trace!(
+                                    static_self.logger,
+                                    "Trying again after block polling failed: \
+                                    block data unavailable, block was likely uncled";
+                                    "block_hash" => format!("{:?}", block_hash)
```
leodasvacas: This should print with `{}` rather than `{:?}`.

datasource/ethereum/src/block_ingestor.rs
```diff
@@ -60,10 +60,22 @@ where
...
+                                );
+                            }
+                            BlockIngestorError::Unknown(e) => {
+                                warn!(
+                                    static_self.logger,
+                                    "Trying again after block polling failed: {:?}", e
```
leodasvacas: Same here.

datasource/ethereum/src/block_ingestor.rs
```diff
@@ -262,9 +296,25 @@ where
...
-                .map_err(|e| format_err!("could not get blocks from Ethereum: {}", e))
+                .map_err(|e| format_err!("could not get blocks from Ethereum: {}", e).into())
                 .map(|_| stream::futures_unordered(block_futures))
                 .flatten_stream(),
         )
     }
 }
+
+#[derive(Debug)]
+enum BlockIngestorError {
```
leodasvacas: Even though you chose to convert from `Error` to this rather than the other way around, it's still good to derive `Fail` so we get `Display`.

Jannis (dismissed):

datasource/ethereum/src/block_ingestor.rs
```diff
@@ -153,28 +153,66 @@ where
...
+                        if receipt.block_hash != block_hash {
+                            // This block is no longer on the main chain according to the Ethereum
+                            // node.  Nothing we can do from here except give up trying to ingest
+                            // this block.
+                            Err(BlockIngestorError::BlockUnavailable(block_hash))
+                        } else {
+                            Ok(receipt)
+                        }
+                    })
+            }).collect::<Vec<_>>();
```
leodasvacas: is this `collect` necessary?

leoyvens (changes requested):
Looks good, a few minor comments.

datasource/ethereum/src/block_ingestor.rs
timmclean: I used to have that, but H256's Display doesn't print the entire hash :(

datasource/ethereum/src/block_ingestor.rs
timmclean: I think so, otherwise they'll load serially instead of parallel

datasource/ethereum/src/block_ingestor.rs
timmclean: Oh good idea. Will do

datasource/ethereum/src/block_ingestor.rs
leodasvacas: I don't follow, `futures_unordered` would execute in parallel.

datasource/ethereum/src/block_ingestor.rs
leodasvacas: Ah, what I meant to say is that if we have `Fail` for BlockIngestError then maybe we should not be matching here and instead should just log `e`.

timmclean:
Oh yeah, I see what you're getting at. This match statement is actually why
I made the BlockIngestorError type though -- because I needed to make some
errors `trace` and some errors `warn` -- so I can't really do too much there

datasource/ethereum/src/block_ingestor.rs
timmclean: Oh good point, those methods collect implicitly don't they. 
So I just tried to remove the collect and the compiler error now reminds me that I collected here to save having to clone. I need `block` later on to move it into the EthereumBlock, so to avoid collecting I would have to clone a bunch anyway, which seemed like a waste...

leodasvacas: Weird, well let's not spend too much time on this.

datasource/ethereum/src/block_ingestor.rs
leodasvacas: This message now duplicates the `Fail` derivation, you can just print `e` instead.

leoyvens (changes requested):

Jannis:
@leodasvacas How's this looking now?

leodasvacas:
@Jannis my only remaining comment is https://github.com/graphprotocol/graph-node/pull/492#discussion_r226123682

Jannis:
Ah, yes, sorry, that is still valid.

timmclean:
@leodasvacas Fixed!

leoyvens (approved):

PR #494, opened 2018-10-15
Disambiguate entity field deserialization, fix order_by, return option from Store::get

### LINKED ISSUE #331: Fix BigInt comparisons in GraphQL filters

Jannis:
Obsoletes #376, obsoletes #480.

Fixes #331, making `orderBy`, `_lt`, `_lte` and other filters work for all supported entity attribute types.

## Fix ambiguous (and broken) entity deserialization

Previously, we were deserializing entity attributes of type `Value::Bytes` incorrectly as `Value::String`. This wasn't noticeable in the GraphQL API, since we're representing `Bytes` as strings there. However, it made using a `Bytes` value returned by `store.get` in mappings impossible (handlers would crash).

This PR changes the way entities are stored in Postgres: Instead of storing
```json
{"id": "...",
 "owner": "0x..."}
```
we now store tagged versions of the attribute values:
```json
`{"id": {"type": "String", "data": "..."},
  "owner": {"type": "Bytes", "data": "0x..."}}
```
This is certainly more verbose but we will change how entities are stored soon anyway (e.g. one row per attribute). Until then, this change removes all ambiguity from loading/deserializing entities from Postgres and should make mappings that use attributes returned by `store.get` work again.

## `Store::get` now returns a `Result<Option<Entity>, QueryExecutionError>`

From the original PR (#480):

> `Store::get` and `Store::get_entity` now return `Result<Option<Entity>, ...>`, allowing handling of query errors and returning `None` if no data is found.

This also allows mappings to distinguish between no entity being returned from `store.get` and `store.get` failing.

## `orderBy` fix

From the original PR (#376):

> This PR fixes the filter query bug where all order_by comparisons were done in alphabetical order because the jsonb objects are returned as text; when adding an order_by filter the jsonb attribute must be cast from text to the correct data type before performing comparisons
>
> A new enum was created, ValueType, with a variant for each supported data type. ValueType is now passed into the order_by query fragment builder and used to decide which data cast to perform before ordering. The order_by queries have all been updated to now include NULLS LAST, so null values will not appear in the beginning of the filtered set.
>
> Data casting from jsonb::text before comparing:
>
> BigInt -> cast to `numeric`
> Int -> cast to `bigint`
> String -> leave as `text`
> Float -> cast to `double precision`
> Id -> leave as `text`
> Bytes -> leave as `text`
> Boolean -> cast to `boolean`

### DIFF
--- a/graph/src/data/store/mod.rs
+++ b/graph/src/data/store/mod.rs
@@ -1,3 +1,4 @@
+use failure::Error;
 use graphql_parser::query;
 use graphql_parser::schema;
 use prelude::QueryExecutionError;
@@ -28,18 +29,9 @@ pub enum ValueType {
     String,
 }
 
-#[derive(Debug, Fail)]
-pub enum ValueTypeError {
-    #[fail(display = "Found unexpected type name: {}", name)]
-    UnexpectedTypeName { name: String },
-    #[fail(display = "Cannot convert from ListType to ValueType")]
-    CannotConvertFromListType,
-    #[fail(display = "Found invalid schema type: nested NonNull type")]
-    NestedNonNullType,
-}
-
 impl FromStr for ValueType {
-    type Err = ValueTypeError;
+    type Err = Error;
+
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         match s {
             "Boolean" => Ok(ValueType::Boolean),
@@ -49,9 +41,7 @@ impl FromStr for ValueType {
             "ID" => Ok(ValueType::ID),
             "Int" => Ok(ValueType::Int),
             "String" => Ok(ValueType::String),
-            e => Err(ValueTypeError::UnexpectedTypeName {
-                name: e.to_string(),
-            }),
+            s => Err(format_err!("Type not available in this context: {}", s)),
         }
     }
 }
--- a/graph/src/data/store/mod.rs
+++ b/graph/src/data/store/mod.rs
@@ -1,7 +1,6 @@
 use graphql_parser::query;
 use graphql_parser::schema;
 use prelude::QueryExecutionError;
-
 use std::collections::{BTreeMap, HashMap};
 use std::fmt;
 use std::iter::FromIterator;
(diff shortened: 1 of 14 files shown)


graph/src/data/store/mod.rs
```diff
@@ -20,7 +20,7 @@ pub const BIG_INT_SCALAR: &str = "BigInt";
 
 /// An attribute value is represented as an enum with variants for all supported value types.
 #[derive(Clone, Debug, Deserialize, Serialize, PartialEq)]
-#[serde(untagged)]
+#[serde(tag = "type", content = "data")]
```
fordN: thank you `serde`!

fordN (dismissed):
Looking good!

leoyvens (changes requested):
I'm hitting the following error trying to deploy the decentraland subgraph to this branch:

```
 Oct 17 19:57:44.999 ERRO Subgraph instance failed to run:
Error while processing block stream for a subgraph:
Failed to get Parcel entity with ID "0xa1ffffffffffffffffffffffffffffff83ffffffffffffffffffffffffffffff" from store:
Invalid entity: invalid type: string "-125", expected a borrowed string, id: QmaMUYb8GBZBrK6664P822xYyfN2nqR5VVfpkWFyvtwyD9
```

leodasvacas:
I'm also getting different behaviour on the ens subgraph. If I try using a pre-existing database I get:
```
ERRO Subgraph instance failed to run: Error while processing block stream for a subgraph:
Failed to get Transfer entity with ID "0xbdbebabdf34e6c4edd540fc75d71450ae06a79b7fae284926bdd7a04a1d86945" from store: 
Invalid entity: invalid type: string "0xbdbebabdf34e6c4edd540fc75d71450ae06a79b7fae284926bdd7a04a1d86945",
expected adjacently tagged enum Value, id: QmZwmo649caQf4s8CmYvSb2geE1LAZy6QLEsNogUGqrwiB
```
Which seems to be fixed by re-creating the db. I guess for now we should just communicate in release notes that the database should be re-created between releases of the node, though it would be nice to get some simple db versioning that could detect this and give a better error.

More importantly the ens subgraph is giving me this on a clean database:
```
ERRO Subgraph instance failed to run: Failed to process event:
Failed to handle Ethereum event with handler "transfer":
Trap: Trap { kind: UnexpectedSignature }, id: QmZwmo649caQf4s8CmYvSb2geE1LAZy6QLEsNogUGqrwiB
```
On a first look I could not find any hint on the change that caused this, but it does not happen on master.

Jannis:
Yep, this is a breaking change, so the first error is expected. I'll see about the second one.

Jannis:
Should be fixed now after I pushed another commit:
>  runtime/wasm: Return null from store.get if entity is not found
>
> We expect the entity to be null if it wasn't found. Returning `Ok(None)`
is not the equivalent of `null` - `Ok(Some(RuntimeValue::from(0)))` is.

leoyvens (dismissed):
All bugs squashed! Could you also update the comment above `fn store_get` to reflect the correct signature?

leoyvens (approved):

PR #495, opened 2018-10-15
Return null from store.get if entity does not exist

Jannis:
This is a follow-up to #494. It fixes an "Unexpected signature" error when calling `store.get` and `Ok(None)` was returned, which can happen after #494 is merged.

Instead of returning `Ok(None)`, we return `Ok(RuntimeValue::from(0))`, which AssemblyScript understands as `null`.


leodasvacas:
Needs a rebase.

Jannis:
Obsoleted by a commit pushed to #494.

PR #501, opened 2018-10-19
Add Docker and Compose files from the graph-node-docker repo

### LINKED ISSUE #499: Move Docker and Compose files into main repo
This helps guarantee that the docker image is always in sync with the latest code.

Jannis:
The problem with the separate repo is that it's always out of sync with the main codebase. By moving the files into this repo, a new Docker image will always be built as we push features into master. We can also selectively add Docker tags for other branches and they too will remain in sync.

Resolves #499.


Jannis:
It's built fine from this repo on Docker Hub, see https://hub.docker.com/r/graphprotocol/graph-node/builds/.

Jannis:
I'm going to go ahead and merge this, as it doesn't affect the main code base and would be nice to clean up asap.

PR #502, opened 2018-10-19
docker: use alpine as base image

### LINKED ISSUE #489: Use Alpine as the base image for our docker image

bohendo:
closes #489

We need to specify the `alpine/edge/community` repo to make sure we download a compatible version of rust (1.29).

Cargo is needed to build but we can delete it once we're done to save space.

clang-dev and postgresql-dev are both required to build successfully, they seem to be dynamically linked because deleting them afterwards doesn't work.

The commit we want to build is specified explicitly, update this commit hash (or, preferably, specify a release tag) to update which version is built.

The image this Dockerfile builds is 430 MB (compared to 2.2 GB for the version based on rust)


bohendo:
It's fairly common for projects to have a heavier image that prioritizes good features and a lighter image that prioritizes small size.

I'm doing custom stuff in my node's entrypoint so I don't need the `start-node` script or `wait-for-it`; this simple alpine image that I've submitted is exactly what I'm going to be using.

That said, I'd be happy to put together a best of both worlds image that tries to keep the features you've added while also cutting down some size. Should be able to get around to it in the next day or two.

Jannis:
@bohendo We'd prefer to only maintain a single image for now, ideally a small like you've suggested. For custom cmds / entry points, it's easy enough to build on top of the official image and just override CMD / ENTRYPOINT.

So if you could make `docker/Dockerfile` small, that would be awesome!

docker/Dockerfile
```diff
@@ -1,20 +1,21 @@
-FROM rust:latest
+FROM alpine:edge
```
bohendo: edge is needed to get rust/cargo v1.3

bohendo:
When I try to build in Alpine I get the following error:

```
error: failed to run custom build command for `openssl v0.9.24`
process didn't exit successfully: `/graph-node/target/release/build/openssl-d71d668c76f38adb/build-script-build` (exit code: 101)
--- stderr
thread 'main' panicked at 'Unable to detect OpenSSL version', /root/.cargo/registry/src/github.com-1ecc6299db9ec823/openssl-0.9.24/build.rs:16:14
note: Run with `RUST_BACKTRACE=1` for a backtrace.

warning: build failed, waiting for other jobs to finish...
error: failed to compile `graph-node v0.4.1 (/graph-node/node)`, intermediate artifacts can be found at `/graph-node/target`
```

I've seen this same error in other projects (eg [kryptco](https://github.com/kryptco/kr/issues/237) and [parity](https://github.com/paritytech/substrate/issues/886)) and it looks like it's due to a dependency on an outdated version of openssl. [Here's](https://github.com/paritytech/substrate/pull/913/files) how parity fixed it. Looks like they edited the `Cargo.lock` file to depend on newer versions of this lib.

I'm not super familiar w rust & even less familiar w how this project uses ssl so will hold off on trying to solve this.

Side note: this build did work on release 0.4.0 (commit a0da5a4) so something between then & now is likely the culprit here.

bohendo:
Actually, nevermind. The old (commit a0da5a4) version of this build process is failing too now & producing the following error while trying to install cargo:

```
ERROR: unsatisfiable constraints:
  so:libcrypto.so.1.1 (missing):
    required by:
                 cargo-1.30.0-r1[so:libcrypto.so.1.1]
                 cargo-1.30.0-r1[so:libcrypto.so.1.1]
                 cargo-1.30.0-r1[so:libcrypto.so.1.1]
  so:libssl.so.1.1 (missing):
    required by:
                 cargo-1.30.0-r1[so:libssl.so.1.1]
                 cargo-1.30.0-r1[so:libssl.so.1.1]
                 cargo-1.30.0-r1[so:libssl.so.1.1]
```

This probably arose out of nowhere because we're using the alpine edge upstream package list which, I assume, is being actively developed & a breaking change was pushed recently..

leodasvacas:
@bohendo If you'd like to give this another shot, we've since updated our openssl dependency.

bohendo:
No bandwidth for this atm. Plus, this branch is so far behind master that it would be just as easy to start over than to wrestle w conflicts. Let's close this for now and re-open a similar one later once someone has time to push this feature to completion.

PR #506, opened 2018-10-19
Fix array deserialization

leodasvacas:
In asc, arrays are backed by an ArrayBuffer, and this PR fixes how we deal with those.

The first commit changes the way we read from an ArrayBuffer, by intially reading it as just bytes and then interpreting only the requested range in `ArrayBuffer::get`, respecting length and offset. See the third commit for how this was useful to fix how we deserialize slices of TypedArray.

The first commit also removes all To/FromAsc conversions for ArrayBuffer and updates tests accordingly, we don't use them and probably should continue not to, preferring the higher level array types.

The second commit fixes the issue where pushing to an array in a mapping would sometimes break things, by giving Asc the extra capacity it expects to be present at the end of an ArrayBuffer. Without that it would go out of bounds and overwrite other data.

@davekaj please test if this solves the issues you were seeing with arrays when building subgraphs.


runtime/wasm/src/asc_abi/class.rs
```diff
@@ -87,7 +111,9 @@ impl<T: AscValue> AscType for ArrayBuffer<T> {
 #[repr(C)]
 pub(crate) struct TypedArray<T> {
     pub buffer: AscPtr<ArrayBuffer<T>>,
+    /// Byte position in `buffer` of the array start.
     byte_offset: u32,
+    /// Byte position in `buffer` of the array end.
     byte_length: u32,
```
timmclean: The field names here make it sound like the array end would be at `byte_offset + byte_length`. Are these comments right?

leodasvacas: The comment is correct, the field name is really misleading, but I chose to keep the same name [asc uses](https://github.com/AssemblyScript/assemblyscript/wiki/Memory-Layout-&-Management#arrays).

timmclean: Ah I see what you mean, that's weird. I'd vote for renaming the fields `byte_start` and `byte_end` or similar and leaving a comment noting the difference from the spec. Must be a mistake on the asc side.
(also: how about adding a comment mentioning that the 2nd ptr points to one byte past the end of the array?)

runtime/wasm/src/asc_abi/class.rs
```diff
@@ -17,29 +18,52 @@ pub(crate) struct ArrayBuffer<T> {
     // elements in `content` will be aligned for any element type.
     _padding: [u8; 4],
```
timmclean: Unrelated to this PR, but do you know if this padding is actually needed? If we're transmuting ArrayBuffer, I don't think this is safe because Rust's memory layout for structs is undefined IIRC (except with `repr`)

davekaj:
arrays are working properly for dharma subgraph on this branch

runtime/wasm/src/asc_abi/class.rs
leodasvacas: I've renamed it, good point about it being one byte past, hadn't realized that.

runtime/wasm/src/asc_abi/class.rs
leodasvacas: It's not needed, it's only there to illustrate the layout. We don't transmute it because ArrayBuffer has a custom impl of AscType. I've refactored so we use that field in `to_asc_bytes`.

runtime/wasm/src/asc_abi/class.rs
```diff
@@ -15,61 +16,83 @@ pub(crate) struct ArrayBuffer<T> {
...
-            _padding: [0; 4],
+            padding: [0; 4],
             content: content.into(),
+            ty: PhantomData,
         }
     }
+
+    /// Read `length` elements of type `T` starting at `byte_offset`.
+    ///
+    /// Panics if that tries to read beyond the lenght of `self.content`.
```
timmclean: typo 'lenght'

PR #512, opened 2018-10-23
Remove only by name, lookup subgraph in url by name first

leodasvacas:
First commit changes subgraph removal to work only by name, removal by id doesn't have much of a use case and it didn't with auth.

Second commit prevents a subgraph from impersonating another by looking up subgraphs identifiers in urls by id first and then by name.

### DIFF
--- a/server/http/src/service.rs
+++ b/server/http/src/service.rs
@@ -79,15 +79,21 @@ where
         let graphql_runner = self.graphql_runner.clone();
         let schemas = self.schemas.read().unwrap();
 
-        // First try `name_or_id` as a name, if that fails try it as an id.
-        let schema = schemas
-            .get(name_or_id)
-            .or_else(|| schemas.get(self.names.read().unwrap().get(name_or_id)?));
+        // First try `name_or_id` as an id, if that fails try it as a name.
+        // This is so that a subgraph cannot impersonate another by making its
+        // name equal to the other's id.
+        let schema = schemas.get(
+            self.names
+                .read()
+                .unwrap()
+                .get(name_or_id)
+                .map(String::as_str)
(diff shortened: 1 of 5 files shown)


server/http/src/service.rs
```diff
@@ -79,15 +79,21 @@ where
...
-        // First try `name_or_id` as a name, if that fails try it as an id.
-        let schema = schemas
-            .get(name_or_id)
-            .or_else(|| schemas.get(self.names.read().unwrap().get(name_or_id)?));
-
-        let schema = if let Some(schema) = schema {
-            schema.clone()
-        } else {
-            return self.handle_not_found();
+        // First try `name_or_id` as an id, if that fails try it as a name.
```
Jannis: Is this comment correct? "First try... as an id, if that fails try it as a name"?

leodasvacas: Yes, that's the intended change. Name first we're vulnerable to the following attack: If subgraph A has id `Qm1234` then subgraph B can easily impersonate it by setting its name to `Qm1234`. This is virtually impossible if we do id first.

leodasvacas: It's all a bit subtle, but I'm not sure how to improve the comment. Suggestions?

Jannis:
```rust
// If `name_or_id` is the ID for a name, use that name. Otherwise,
// use `name_or_id` as is. This protects us from requests that try
// to remove a subgraph using a name that equals the ID of another
// subgraph.
```
Maybe?

leodasvacas:
@Jannis I've tried an improvement to the comment

Jannis:
Thanks @leodasvacas, much appreciated 😄

PR #513, opened 2018-10-23
Persistent subgraphs

### LINKED ISSUE #404: Make subgraph deployments persistent across restarts
For this, nodes that we deploy for hackathons need to have a persistent Postgres database and persistent IPFS node. Any subgraph that is deployed over the JSON-RPC admin API should be resumed after restarting a node.

So basically: Need a persistent IPFS node. Need to remember subgraphs that are registered in Postgres.

timmclean:
I believe with this PR we can leave `ethsf` behind?

Resolves #404 

---

Major changes:
- Schemas, SubgraphInstances, BlockStreams, etc no longer have a subgraph name associated with them. Just a SubgraphId.
- `Store` has 7 new methods related to reading/writing/deleting subgraph name->ID mappings, access tokens, etc. This information was previously stored in-memory in graph-node.
- Components that used to track subgraph name->ID mappings in-memory now have an `Arc<Store>` and have been rewritten to use that (e.g. JSON RPC service, SubgraphProvider deploy/remove)
- `SubgraphProvider::new` is now `SubgraphProvider::init`. It retrieves a list of subgraph names from the Store and starts those subgraphs.
- `MockStore` has been extended with mock implementations for some of the new methods. `MockStore` is used in a bunch of new places now that more components use a Store.

SubgraphProvider `deploy` and `remove` are hopefully pretty readable. `deploy` is essentially:
- Resolve the subgraph manifest
- Call `SubgraphProvider::remove` to undeploy the subgraph name if it already exists
- Add subgraph ID directives
- Write this subgraph name->ID mapping to the Store
- Look for other names in the Store pointing at this ID. If none, start processing on this subgraph, otherwise, done (subgraph is already be running).

`remove` is essentially:
- Read the subgraph name->ID mapping from the Store
- Delete the name->ID mapping in the Store
- Look for other names in the Store pointing at this ID. If none, shut down processing on this subgraph, otherwise, done (leave the subgraph running).


timmclean:
Opening this up for code review now since there's a lot here, despite the fact that it's not done. Most of the work from the last few days has been trying to remove the assumption that each subgraph has one and only one name... would be interested in some feedback on whether that's the right approach. @Jannis

leodasvacas:
In an architecture with a dedicated name service the node would not be managing names at all, but we still want built-in name management for standalone nodes. To make this mode switching easier, what do you think of extracting the name logic in `SubgraphProvider` to a `SubgraphNames` component (or whatever name you prefer). `SubgraphNames` would have methods analogous to `deploy` and `remove`, that would return whether or not the subgraph should be deployed or removed. When running with a dummy `SubgraphNames`, the name is completely ignored and the subgraph is always deployed or removed. @timmclean makes sense? What do you think?

timmclean:
Ah interesting. That could work. What would this look like externally? Right now I believe the JSON RPC service has deploy/remove/authorize/list. Should we use the existing verbs, revise them to make them more flexible, or add a separate JSON RPC service with a new set?

leodasvacas:
@timmclean I wouldn't touch that layer at all yet, we're gonna need to figure that out later (hopefully it will be graphql mutations rather than json-rpc) but for now I just wanna make sure it will be clean and easy to turn off the code that writes and removes names.

timmclean:
So how would `deploy` and `remove` work in SubgraphProvider in that case? They both have `name` arguments... would we pass an empty string? None?

timmclean:
Planned refactor for separating name functionality from SubgraphProvider:
- In SubgraphProvider:
  - Does not depend on Store.
  - Track a HashSet of SubgraphIds that are currently running in this graph node instance.
  - `deploy(name, link)` will become `start(link)`. Error if subgraph ID already in the HashSet. Add ID to HashSet.
  - `remove(name)` will become `stop(id)`. Error if subgraph ID not in the HashSet. Remove ID from HashSet.
  - Emits events just like current SubgraphProvider.
- New component: SubgraphProviderWithNames
  - Does depend on Store.
  - Used by JSON RPC service.
  - `deploy(name, link)`: same functionality as current `SubgraphProvider::deploy(name, link)`
  - `remove(name)`: same functionality as current `SubgraphProvider::remove(name)`

@leodasvacas

leodasvacas:
@timmclean thanks from writing out the plan, looks solid!

timmclean:
Refactor done!

server/http/src/service.rs
```diff
@@ -76,38 +71,46 @@ where
...
+                .and_then(move |id_opt_opt| -> GraphQLServiceResponse {
+                    let service = service.clone();
+                    let schemas_rwlock = service.schemas.clone();
+                    let schemas = schemas_rwlock.read().unwrap();
+
+                    let id = match id_opt_opt {
+                        // If name_or_id matched a name with non-null ID
+                        Some(Some(id)) => id,
+
+                        // Otherwise, treat name_or_id as an ID
```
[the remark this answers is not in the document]
timmclean: I think it's right, `name_or_id` gets put into `id`, which is used in `schemas.get(&id)`, a `BTreeMap<SubgraphId, Schema>`

core/src/subgraph/provider_with_names.rs
```diff
@@ -0,0 +1,213 @@
...
+                let self_clone = self_clone.clone();
+
+                let subgraph_ids = subgraph_names_and_ids
+                    .into_iter()
+                    .filter_map(|(_name, id_opt)| id_opt)
+                    .collect::<HashSet<SubgraphId>>();
+
+                stream::iter_ok(subgraph_ids).for_each(move |id| {
+                    let link = format!("/ipfs/{}", id);
+                    self_clone.provider.start(link).from_err()
```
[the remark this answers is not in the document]
timmclean: don't worry, subgraph_ids is deduped in the HashSet

timmclean:
@leodasvacas pushed some commits, how does this look?

server/http/src/service.rs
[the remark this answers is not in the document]
leodasvacas: Got it. But the behaviour should be the opposite here, trying `name_or_id` first as an id and then as a name, for the reason that's in the comment that got lost here. We should bring back the comment and the previous behaviour.

timmclean:
Latest commit adds separate endpoints to `server/http` and `server/websocket` for accessing subgraphs by ID or name. E.g. `http://localhost:8000/by-name/adchain` or `http://localhost:8000/by-id/QmaMUYb8GBZBrK6664P822xYyfN2nqR5VVfpkWFyvtwyD9`

server/http/src/service.rs
```diff
@@ -52,7 +52,27 @@ where
...
+                            if let Some(_) = subgraph_id_opt {
+                                return service
+                                    .handle_temp_redirect(&format!("/by-name/{}", subgraph_name));
+                            }
+                        }
+
+                        service.handle_not_found()
+                    },
+                ),
+        )
```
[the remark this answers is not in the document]
timmclean: Oh true, yeah, I don't feel too strongly on having a redirect or not! I was tempted to put in a quicky plaintext list of subgraph names...

Comment added though

server/websocket/src/server.rs
```diff
@@ -92,17 +95,38 @@ where
         }));
     }
 
-    fn subgraph_from_url_path(path: PathBuf) -> Option<String> {
-        path.iter()
-            .nth(1)
-            .and_then(|os| os.to_str())
-            .map(|s| s.into())
+    fn subgraph_id_from_url_path(store: Arc<S>, path: PathBuf) -> Result<String, ()> {
```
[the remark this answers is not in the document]
timmclean: Hmmm true. I think it's OK because the namespaces are separate and similar mostly by coincidence. I could change this to use a matching style similar to server/http though if you like

node/src/main.rs
```diff
@@ -318,10 +328,23 @@ fn async_main() -> impl Future<Item = (), Error = ()> + Send + 'static {
             )).and_then(|_| Ok(())),
     );
 
+    // Create named subgraph provider for resolving subgraph name->ID mappings
+    let named_subgraph_provider = Arc::new(
+        IpfsSubgraphProviderWithNames::init(
+            logger.clone(),
+            Arc::new(subgraph_provider),
+            store.clone(),
+        ).wait()
```
Jannis: This `wait`... can it not cause freezing (e.g. if something inside `init` were to spawn something)?

server/http/src/service.rs
```diff
@@ -123,6 +183,20 @@ where
...
 
+    /// Handles 302 redirects
+    fn handle_temp_redirect(&self, destination: &str) -> GraphQLServiceResponse {
+        Box::new(future::ok(
+            Response::builder()
+                .status(StatusCode::FOUND)
+                .header(
+                    header::LOCATION,
+                    header::HeaderValue::from_str(destination)
+                        .expect("invalid redirect destination"),
```
Jannis: Is this something that should make the node panic?

server/json-rpc/src/lib.rs
```diff
@@ -96,29 +99,59 @@ impl<T: SubgraphProvider> JsonRpcServer<T> {
...
+                json_rpc_error(JSON_RPC_INTERNAL_ERROR, "database error".to_owned())
+            })?;
+
+        if is_authd {
+            return Ok(());
+        }
+
+        return Err(json_rpc_error(
+            JSON_RPC_UNAUTHORIZED_ERROR,
+            "API key is invalid".to_owned(),
```
Jannis: This should probably be `Invalid access token`, since `API key` is not what we call this any more.

server/json-rpc/src/lib.rs
```diff
@@ -160,36 +196,62 @@ impl<T: SubgraphProvider> JsonRpcServer<T> {
     ) -> Result<Value, jsonrpc_core::Error> {
         info!(self.logger, "Received subgraph_authorize request"; "params" => params.to_string());
         Self::require_master_token(auth)?;
-        *self.subgraph_api_keys.write().unwrap() = params.subgraph_api_keys;
+
+        for (subgraph_name, access_token) in params.subgraph_api_keys {
```
Jannis: Let's rename `subgraph_api_keys` to `subgraph_access_tokens` as well.

server/websocket/src/server.rs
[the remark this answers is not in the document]
Jannis: I think it's ok. This logic here is simple compared to that in the GraphQL/HTTP server.

store/postgres/src/store.rs
```diff
@@ -434,6 +434,104 @@ impl Store {
 }
 
 impl StoreTrait for Store {
+    fn authorize_subgraph_name(&self, name: String, new_access_token: String) -> Result<(), Error> {
```
Jannis: We're putting a bit much into the `Store`. At some point it will be worth thinking about splitting this up into separate traits and separate implementations. But not right now.

store/postgres/src/store.rs
```diff
@@ -434,6 +434,104 @@ impl Store {
...
+        let real_access_token_opt = subgraph_names
+            .select(access_token)
+            .filter(subgraph_name.eq(&name))
+            .first::<Option<String>>(&*self.conn.lock().unwrap())?;
+
+        match real_access_token_opt {
+            None => {
+                debug!(
+                    self.logger,
+                    "Subgraph name {:?} has no associated access token. Access denied by default.",
```
Jannis: This should use structured logging:
```rust
debug!(
  self.logger,
  "Subgraph has no associated access token. Access denied by default.";
  "subgraph_name" => &subgraph_name
);
```

Also: you're passing in `access_token` instead of `subgraph_name` anyway. 😉

node/src/main.rs
timmclean: Based on my current level of tokio understanding, I think it's fine... async_main is already running inside the context of a tokio runtime, and tokio can use other threads

@leodasvacas ?

server/json-rpc/src/lib.rs
timmclean: makes sense!

store/postgres/src/store.rs
timmclean: Agreed. We should have a separate store component for subgraph name info

server/json-rpc/src/lib.rs
Jannis: This is a breaking change in the JSON-RPC API but that's fine IMHO.

timmclean:
I think I got everything!

PR #516, opened 2018-10-23
BlockStream: Change log from `error` to `info`

leodasvacas:
This is not an error but expected behaviour when a subgraph is removed.

### DIFF
--- a/datasource/ethereum/src/block_stream.rs
+++ b/datasource/ethereum/src/block_stream.rs
@@ -803,8 +803,8 @@ where
     fn event_sink(&self) -> Box<Sink<SinkItem = ChainHeadUpdate, SinkError = ()> + Send> {
         let logger = self.ctx.logger.clone();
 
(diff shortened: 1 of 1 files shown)


datasource/ethereum/src/block_stream.rs
```diff
@@ -803,8 +803,8 @@ where
     fn event_sink(&self) -> Box<Sink<SinkItem = ChainHeadUpdate, SinkError = ()> + Send> {
         let logger = self.ctx.logger.clone();
 
-        Box::new(self.chain_head_update_sink.clone().sink_map_err(move |e| {
-            error!(logger, "Component was dropped: {}", e);
+        Box::new(self.chain_head_update_sink.clone().sink_map_err(move |_| {
+            info!(logger, "terminating chain head updates");
```
Jannis: Our convention for log messages is to start with a capital letter.

PR #518, opened 2018-10-24
Deprecation notice for the `--subgraph` option

leodasvacas:
Now that `graph deploy` exists this has mostly lost its purpose and we'd rather not maintain this code through all the incoming changes for the hosted service.

### DIFF
--- a/server/http/src/server.rs
+++ b/server/http/src/server.rs
@@ -197,7 +197,7 @@ mod tests {
                         } else if Instant::now().duration_since(start_time) > max_wait {
                             panic!("Timed out, schema not received")
                         }
-                        ::std::thread::yield_now();
+                        ::std::thread::sleep(Duration::from_millis(10));
                     };
 
                     assert_eq!(output_schema.id, input_schema.id);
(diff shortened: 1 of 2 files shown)


server/http/src/server.rs
```diff
@@ -188,9 +188,9 @@ mod tests {
                     schema_sink.send(input_event).wait().unwrap();
 
                     // Wait for the schema to be received and extract it.
-                    // Wait for thirty seconds for that to happen, otherwise fail the test.
+                    // Wait a while for that to happen, otherwise fail the test.
                     let start_time = Instant::now();
-                    let max_wait = Duration::from_secs(30);
+                    let max_wait = Duration::from_secs(60);
```
timmclean: Huh why does this take so long? If it's taking that long, how about thread sleep 1ms instead of yield_now?

leodasvacas: It was just travis being slow, it's usually fast.

timmclean: I'm wondering if sleeping could speed that up

leodasvacas: ah got it, done

Jannis:
Closed because we've decided to keep the `--subgraph` option.

PR #520, opened 2018-10-24
travis: Fix and refactor style check and release building

leodasvacas:
Using tricks I learned from the [diesel travis](https://github.com/diesel-rs/diesel/blob/master/.travis.yml#L40) to fix and cleanup our included builds. Also pin the rustfmt version so we're not forced to upgrade as soon as Rust upgrades.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,7 @@ matrix:
     # Some env var is always necessary to differentiate included builds
     # Check coding style
     - env: CHECK_FORMATTING=true
-      rust: 1.29.2
+      rust: stable
       script:
         - rustup component add rustfmt-preview
         - cargo fmt --all -- --check
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,16 +27,22 @@ before_install:
 
 matrix:
   fast_finish: true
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -27,16 +27,22 @@ before_install:
...
   fast_finish: true
-  # Allow nightly to fail
   allow_failures:
     - rust: nightly
-  # Only check formatting when running against stable
   include:
+    # Some env var is always necessary to differentiate included builds
+    # Check coding style
     - env: CHECK_FORMATTING=true
+      rust: 1.29.2
```
Jannis: Could we make this 'stable', so we're always using the current stable formatting rules?

leodasvacas: Using `stable` is what I was trying to avoid, because it tends to break our builds as soon as a new stable is released. Pinning this allows us to upgrade our Rust in our own pace.

Jannis: That's why I made it `stable` initially — so we'd be incentivized to stay on the latest stable.

leodasvacas:
@Jannis using stable is also a valid choice, I put it back on stable.

PR #524, opened 2018-10-25
enable debug mode with the docker image

tsuberim:


docker/start-node
```diff
@@ -6,4 +6,5 @@ set -x
 graph-node \
   --postgres-url "postgresql://$postgres_user:$postgres_pass@$postgres_host/$postgres_db" \
   --ethereum-rpc "$ethereum" \
-  --ipfs "$ipfs"
+  --ipfs "$ipfs" \
+  ${debug:+--debug}
```
Jannis: The `:+` substitution is clever but (at least to me) it's not obvious how to _disable_ debug mode. What I'd rather expose is `env GRAPH_LOG "debug"`. `graph-node` will then pick that up and it's obvious that it can be changed to e.g. `"info"` or `"warn"` from the outside.

tsuberim: Does `graph-node` support those log levels? The only documented one is `debug`.

Jannis: Yep, `GRAPH_LOG` is not documented yet but supports `debug`, `info`, `warn` etc., so `env GRAPH_LOG "debug"` would be ideal.

PR #527, opened 2018-10-25
Use address field in eth_getLogs

### LINKED ISSUE #497: Scanning blocks for common event signatures (like Transfer) takes too long

Bug

Scanning blocks for very common event signatures takes a long time. For very uncommon event signatures, scanning up to the first block with events can be as fast as 15-30 seconds. For common event signatures (like `Transfer(address,address,uint256)`) it can take hours.

Create a subgraph that triggers on a common event like `Transfer(address,address,uint256)` for any contract, deploy it to a graph node and observe its progress.

Graph Node should find the first block with events for such a subgraph quickly. Less than one minute would be a start, 20-30 seconds would be better.

timmclean:
Nice performance improvement for some mappings by supplying a list of contract addresses to `eth_getLogs`

Resolves #497

### DIFF
--- a/datasource/ethereum/src/ethereum_adapter.rs
+++ b/datasource/ethereum/src/ethereum_adapter.rs
@@ -42,6 +42,7 @@ where
         &self,
         from: u64,
         to: u64,
+        addresses: Vec<H160>,
         event_signatures: Vec<H256>,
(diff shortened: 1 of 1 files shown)


leoyvens (approved):
How were we not doing this? Awesome.

PR #535, opened 2018-10-29
Shutdown tokio runtime gracefully when panic occurs in spawned future

### LINKED ISSUE #314: Replace tokio::run with runtime::block_on
When using block_on, the tokio runtime successfully shuts down on a panic which is an upgrade. Previously once additional tokio threads were spawned a panic in one did not cause all of them to shutdown. However, an additional panic message is sent during panic and shutdown.

fordN:
Resolves #314

This PR replaces `tokio::run()` with a more custom build of a multi-threaded, Tokio runtime in order to  allow panics in spawned futures to shutdown the entire runtime gracefully. 

A runtime is instantiated and provided the necessary components to setup a context for running our main function. Default components are setup for the context using the respective `<>::with_default()` method. Included with the runtime: reactor handle, executor context, and timer instance. 
The PR also updates the futures library and replaces its now deprecated `deadline()` function with `timeout()`.

Issues created in tokio:
Tough to track down issue setting up the Timer: https://github.com/tokio-rs/tokio/issues/727
Errors resulting from the fact that block_on does not set default runtime context components: https://github.com/tokio-rs/tokio/issues/728


leodasvacas:
@fordN is this still WIP? I imagine having direct access to the runtime and executor will be crucial to solving #314 but this PR isn't there yet, the only change in behaviour I could observe when testing is that we shutdown the runtime when idle, which never happens in practice.

fordN:
@leodasvacas, I tested this by panicking within a future spawned on the runtime.
  
My experience testing:
  - Before the changes: the panic log would be written out (but not always) and the runtime would continue to hobble along.  
  - After the changes: the panic is logged and the application shuts down immediately.

timmclean:
Looks like you need to rebase this onto the new retry helper

fordN:
@leodasvacas I updated the shutdown process to send a shutdown signal to a thread that calls `runtime.shutdown_now` in order to stop all futures, and shutdown the threadpool and reactor.  This update solves several of my test cases, but there are still situations where a future continues to run.  I believe this is because tokio workers do not maintain a reference to the future in order to shutdown.   I've found several related issues. 

Issue with futures holding references: https://github.com/rust-lang-nursery/futures-rs/issues/853
Brainstorm about tokio implementing a dedicated worker thread to run a reactor instance which among other things allows a worker to track all futures that it owns and drop incomplete futures when it shuts down: https://github.com/tokio-rs/tokio/issues/424
Open PR for dropping incomplete tasks on shutdown: https://github.com/tokio-rs/tokio/pull/722

graph/src/util/log.rs
```diff
@@ -35,7 +37,10 @@ pub fn guarded_logger() -> (slog::Logger, slog_async::AsyncGuard) {
     (slog::Logger::root(drain.fuse(), o!()), guard)
 }
 
-pub fn register_panic_hook(panic_logger: slog::Logger) {
+pub fn register_panic_hook(
+    panic_logger: slog::Logger,
+    shutdown_sender: Mutex<Option<oneshot::Sender<()>>>,
```
leodasvacas: I think you can remove the `Mutex` and the `Option` around this, and then construct the mutex inside this function, to simplify the signature.

graph/src/util/log.rs
```diff
@@ -69,5 +74,15 @@ pub fn register_panic_hook(panic_logger: slog::Logger) {
                 );
             }
         };
+
+        if let Ok(ref mut mutex) = shutdown_sender.lock() {
```
leodasvacas: The mutex may be unwrapped, since it will only error if a panic happens while holding the lock, but in this case that would be a double panic which will abort.

node/src/main.rs
```diff
@@ -44,14 +50,48 @@ use graph_server_websocket::SubscriptionServer as GraphQLSubscriptionServer;
...
+
+    std::thread::spawn(|| {
+        shutdown_receiver
+            .wait()
+            .map(|_| {
+                let _ = runtime
+                    .shutdown_now()
+                    .wait()
+                    .expect("Failed to shutdown Tokio Runtime");
+                thread::sleep(Duration::from_millis(3000));
```
leodasvacas: Don't know what you observed in your tests, but the idea here is that we'd have a timeout for `shutdown_now`. If `shutdown_now` succeeds, then we should not need to call `process_exit(1)` since the process should naturally terminate with the termination of the main thread.

So we should have a timeout instead of calling `wait` directly on `shutdown_now`, and we should call `exit` only if the timeout elapses.

graph/src/util/log.rs
leodasvacas: Another thing, we should probably set the exit code to 1 inside the panic hook, since a panic means we'll terminate the process abnormally.

node/src/main.rs
fordN: I had planned on going the route you mention.  Using a `timeout` instead of `wait`, but I found that in one of my test cases (a panic in `instance_manager` around line 76) `shutdown_now` completes successfully but there is still a process running.  So instead of calling 'exit' only if it doesn't complete in a given time I allow it to complete and sleep the shutdown thread for 3 seconds before exiting.

fordN: Revisiting the issue this week, I was not able to repeat the case of `shutdown_now()` hanging, so the timeout didn't have an effect on my test cases.

leodasvacas: Ah I see. So based on that experiment, having the unconditional `exit` is good. But I don't see why we need to sleep before doing that. Also it still seems good to have a timeout around `shutdown_now()` since at some point could observe it hanging.

fordN: I added the sleep in because when testing with multiple panics (in separate threads), the sleep gives time for both to be logged before exiting.  I could go either way on this; it's not necessarily a problem that we don't see issues that occur after the original panic, but is nice to see them both.    Agreed, I'll add a timeout around `shutdown_now` (it's a little tricky using tokio_timer around for an operation that is shutting down tokio, I may create a new timer instance or use `futures-timer`.)

leodasvacas:
#620 caused conflicts. The `Cargo.lock` you'll want to just get the version on master.

graph/src/util/log.rs
fordN: I moved the sleep and exit into the panic_hook, so that now it sends the shutdown signal to main, sleeps, and then exits the process.

fordN: The compiler didn't like constructing the `Mutex<Option<>>` inside of `register_panic_hook()`., so I am still passing the it in as an input.

leodasvacas: Could be because you did it inside `panic::set_hook`. But it's fine, this is not important.

graph/src/util/log.rs
```diff
@@ -67,5 +73,20 @@ pub fn register_panic_hook(panic_logger: Logger) {
...
         };
+
+        // Send a shutdown signal to main which will attempt to cleanly shutdown the runtime
+        // After sending shutdown, the thread sleeps for 3 seconds then forces the process to
+        // exit because the shutdown is not always able to cleanly exit all workers
+        shutdown_sender
+            .lock()
+            .unwrap()
+            .take()
+            .expect("Shutdown signal already sent")
```
leodasvacas: So, the mutex can be unwrapped, but I don't think it's the best idea to do `expect` on the `take()`, I'm not sure but it may be the case that multiple panics may call this hook simultaneously or in sequence.

node/src/main.rs
```diff
@@ -39,14 +45,47 @@ use graph_server_websocket::SubscriptionServer as GraphQLSubscriptionServer;
...
+                    .shutdown_now()
+                    .wait()
+                    .expect("Failed to shutdown Tokio Runtime");
+            }).expect("Runtime shutdown process did not finish");
+    });
+
+    // Setup runtime context with defaults and run the main application
+    tokio_reactor::with_default(&reactor, &mut enter, |enter| {
+        tokio_executor::with_default(&mut executor, enter, |enter| {
+            tokio_timer::with_default(&timer_handle, enter, |enter| {
```
leodasvacas: Is it necessary to start all this defaults? I think in tests we don't do this and just call `runtime.block_on` directly and things still work.

graph/src/util/log.rs
```diff
@@ -67,5 +73,20 @@ pub fn register_panic_hook(panic_logger: Logger) {
...
+            .send(())
+            .map(|_| ())
+            .map_err(|_| ())
+            .expect("Failed to signal shutdown");
+        thread::sleep(Duration::from_millis(3000));
```
leodasvacas: Moving the sleep here was clever!

graph/src/util/log.rs
```diff
@@ -67,5 +73,20 @@ pub fn register_panic_hook(panic_logger: Logger) {
...
+        // exit because the shutdown is not always able to cleanly exit all workers
+        shutdown_sender
+            .lock()
+            .unwrap()
+            .take()
+            .expect("Shutdown signal already sent")
+            .send(())
+            .map(|_| ())
+            .map_err(|_| ())
+            .expect("Failed to signal shutdown");
```
leodasvacas: No particular reason to panic here, maybe there are two panics running and one of them already signaled the shutdown. In `map_err` it might be worth it to log something.

node/src/main.rs
```diff
@@ -39,14 +45,47 @@ use graph_server_websocket::SubscriptionServer as GraphQLSubscriptionServer;
...
+
+    // Shutdown the runtime after a panic
+    std::thread::spawn(|| {
+        shutdown_receiver
+            .wait()
+            .map(|_| {
+                let _ = runtime
+                    .shutdown_now()
+                    .wait()
+                    .expect("Failed to shutdown Tokio Runtime");
```
leodasvacas: When this shutdown suceeds, it would be nice getting a log saying that the node shutdown cleanly.

node/src/main.rs
fordN: I've removed the reactor default.  Removing either of the other defaults results in errors such as below.

```
`CRIT called `Result::unwrap()` on an `Err` value: SpawnError { is_shutdown: true }, location: libcore/result.rs:1009` 
```

graph/src/util/log.rs
fordN: Yep, updated.

timmclean:
Adding this to the Hosted Service milestone, as we need this so that nodes don't get stuck in a broken state due to a transient panic

leodasvacas:
Needs to be fmt'ed.

node/src/main.rs
```diff
@@ -39,14 +45,47 @@ use graph_server_websocket::SubscriptionServer as GraphQLSubscriptionServer;
...
+    // Register guarded panic logger which ensures logs flush on shutdown
     let (panic_logger, _panic_guard) = guarded_logger();
-    register_panic_hook(panic_logger);
-    tokio::run(future::lazy(async_main))
+    register_panic_hook(panic_logger, shutdown_sender);
+
+    // Create components for tokio context: multi-threaded runtime, reactor reference,
+    // executor context on the runtime, and Timer handle.
+    let runtime = tokio::runtime::Runtime::new().expect("Failed to create runtime");
+    let reactor = tokio_reactor::Handle::current();
```
leodasvacas: Is this variable unused? Looks like the whole `tokio_reactor` dependency is unused now.

fordN: yep 👍, I've now removed the entire `tokio_reactor` dependency.

graph/src/lib.rs
```diff
@@ -21,7 +21,10 @@ extern crate slog_envlogger;
 extern crate slog_term;
 extern crate tiny_keccak;
 pub extern crate tokio;
+pub extern crate tokio_executor;
+pub extern crate tokio_reactor;
```
leodasvacas: This is dead code.

graph/Cargo.toml
```diff
@@ -30,5 +30,8 @@ slog-envlogger = "2.1.0"
 slog-term = "2.4.0"
 tiny-keccak = "1.0"
 tokio = "0.1.11"
+tokio-executor = "0.1.5"
+tokio-reactor = "0.1.6"
```
leodasvacas: This dependency is dead.

Jannis:
🎉

PR #538, opened 2018-10-31
fix various clippy warnings

llogiq:
There are many, but each is fairly trivial. There are some removed clones and some closured-out code paths which should improve perf a small bit, plus some readability improvements. I've erred on the safe side choosing not to implement API changes for now, where e.g. clippy reports needless ref for copy types.

### DIFF
--- a/core/src/subgraph/instance_manager.rs
+++ b/core/src/subgraph/instance_manager.rs
@@ -175,7 +175,7 @@ impl SubgraphInstanceManager {
                             let transaction = block
                                 .transaction_for_log(&log)
                                 .map(Arc::new)
-                                .ok_or(format_err!("Found no transaction for event"));
+                                .ok_or_else(|| format_err!("Found no transaction for event"));
 
                             future::result(transaction).and_then(move |transaction| {
                                 instance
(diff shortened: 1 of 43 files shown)


server/websocket/src/connection.rs
```diff
@@ -208,7 +208,7 @@ where
 
             match msg {
                 // Always accept connection init requests
-                ConnectionInit { payload: _ } => send_message(&msg_sink, ConnectionAck),
+                ConnectionInit { .. } => send_message(&msg_sink, ConnectionAck),
```
leodasvacas: I don't like this one, I prefer to know what is being ignored. And in case the struct changes, to explicitly ignore (or not) any new fields.

llogiq: Ok, I'll revert it. Please note that the possibility of change (and subsequent requirement for extension of the pattern) is the exact reason we have this lint. The rationale is that as we don't need to know the struct layout to decide here, we can omit it to concentrate on the enum discriminant.

leodasvacas: That makes sense, I guess you can switch it back if you agree with this lint. Btw you still need to rustfmt with Rust 1.30 so this will pass CI.

llogiq:
@leodasvacas thank you for the review. I ran rustfmt on it and reverted that one change you were unhappy with, we can reapply it later.

PR #548, opened 2018-11-02
runtime: `big_int_to_hex` prints in big endian and trims zeros

### LINKED ISSUE #481: `toHex` for number types should be in big endian
Calling `toHex` in a mapping will print to a string in little-endian representation, which reads backwards and is probably not what users expect. Conversions for I/U256 and I/U128 should print in big-endian, and so should BigInt (which is a bit different because it's a byte array).

leodasvacas:
Resolves #481

### DIFF
--- a/graph/src/data/store/scalar.rs
+++ b/graph/src/data/store/scalar.rs
@@ -27,6 +27,10 @@ impl BigInt {
         self.0.to_bytes_le()
     }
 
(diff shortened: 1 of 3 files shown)


leodasvacas:
@Jannis I believe we store BigInts as decimal strings, so I don't think this would apply.

PR #552, opened 2018-11-05
Include log data in EthereumEvent, also include additional transaction fields

### LINKED ISSUE #539: Include log_index in EthereumEventData to help uniquely identify events
@tsuberim brought to our attention the potential utility of the `log_index` field from transaction logs https://docs.rs/web3/0.4.0/web3/types/struct.Log.html#structfield.log_index.  
Using `log_index` in combination with `transaction_hash` would provide a unique event id. 

While we are at it let's do an audit to ensure we pass down a complete data structure that includes all reasonable `block`, `transaction`, and `transaction_log` data.

fordN:
Resolves #539 

This PR includes additional data with the `EthereumEvent` that is supplied to the mapping functions.  Using the added data in a subgraph mapping requires graphprotocol/graph-ts/pull/19.
  - The `EthereumTransactionData` struct now includes a more complete set of [Transaction](https://tomusdrw.github.io/rust-web3/web3/types/struct.Transaction.html) data, added: `index`, `from`, `to`, `value`, `gas_used`.  
  - Additional fields from [Log](https://tomusdrw.github.io/rust-web3/web3/types/struct.Log.html) have been added directly to `EthereumEventData`.


graph/src/components/ethereum/types.rs
```diff
@@ -57,18 +57,44 @@ impl<'a, T> From<&'a Block<T>> for EthereumBlockData {
...
+            to: tx.to.unwrap_or(H160::from(0)),
+            value: tx.value,
             gas_used: tx.gas,
+            gas_price: tx.gas_price,
+        }
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct EthereumLogData {
```
Jannis: Let's merge these fields into `EthereumEventData`. There's no need for duplicating the thing that represents the event.

graph/src/components/ethereum/types.rs
```diff
@@ -57,18 +57,44 @@ impl<'a, T> From<&'a Block<T>> for EthereumBlockData {
...
 
 impl<'a> From<&'a Transaction> for EthereumTransactionData {
     fn from(tx: &'a Transaction) -> EthereumTransactionData {
         EthereumTransactionData {
             hash: tx.hash,
-            block_hash: tx.block_hash.unwrap(),
-            block_number: tx.block_number.unwrap(),
+            index: tx.transaction_index.unwrap(),
+            from: tx.from,
+            to: tx.to.unwrap_or(H160::from(0)),
```
Jannis: @leodasvacas Would it be realistic to support `ToAscObject<AscPtr<AscOption<U>>> for Option<T>` where `AscOption<U>` corresponds to `U | null` in AssemblyScript?

Jannis: Because I don't really like putting incorrect data like `0` in here.

fordN: Agreed, I meant to present this as an open question in the description. :)

leodasvacas: This is straightforward to do. We'll want a `AscPtr::null()` constructor (just like the `default` impl). `EthereumTransactionData` should use `Option` for the optional fields. Then in `impl ToAscObj<AscEthereumTransaction> for EthereumTransactionData`, we should check for `None` and in that case use an `AscPtr::null()`, otherwise we call `heap.asc_new` on the `Some` payload.

Jannis: Great, let's do it. Thanks @leodasvacas!

Jannis: Done (pushed to this PR).

fordN:
I've made `EthereumEventDat.log_type` optional as well.

fordN:
Passing now!  I used a modified version of the graph-cli example [(graph-cli)](https://github.com/graphprotocol/graph-cli/tree/ford/setup-test-wasm) to generate a new wasm test file that uses the latest versions of `graph-cli` and `graph-ts`.  

It could be nice to align an example in `graph-cli` master with the tests here to facilitate compatibility between the repos.

Jannis:
Good point, graph-cli needs to be updated as well.

PR #554, opened 2018-11-06
Add __typename fields to the API schema, set __typename in store.set()

### LINKED ISSUE #431: Support __typename meta field on entities

Bug

```graphql
{
  parcels {
    id
    __typename
  }
}
```

Jannis:
Resolves #431, allowing everyone to use Graph Node with Apollo, without having to disable the `addTypename` flag of `InMemoryCache` (see https://www.apollographql.com/docs/react/advanced/caching.html#configuration).

### DIFF
--- a/graphql/src/schema/ast.rs
+++ b/graphql/src/schema/ast.rs
@@ -103,9 +103,33 @@ pub fn get_interface_type_definitions<'a>(schema: &'a Document) -> Vec<&'a Inter
         }).collect()
 }
 
+/// Returns all interface definitions in the schema.
+pub fn get_interface_type_definitions_mut<'a>(
+    schema: &'a mut Document,
+) -> Vec<&'a mut InterfaceType> {
+    schema
+        .definitions
+        .iter_mut()
+        .filter_map(|d| match d {
+            Definition::TypeDefinition(TypeDefinition::Interface(t)) => Some(t),
+            _ => None,
+        }).collect()
+}
+
 /// Returns the type of a field of an object type.
 pub fn get_field_type<'a>(object_type: &'a ObjectType, name: &Name) -> Option<&'a Field> {
     object_type.fields.iter().find(|field| &field.name == name)
+}
+
+/// Returns the type of a field of an interface type.
+pub fn get_interface_field_type<'a>(
+    interface_type: &'a InterfaceType,
+    name: &Name,
+) -> Option<&'a Field> {
+    interface_type
+        .fields
+        .iter()
+        .find(|field| &field.name == name)
 }
 
 /// Returns the value type for a GraphQL field type.
--- a/graphql/src/schema/ast.rs
+++ b/graphql/src/schema/ast.rs
@@ -92,6 +92,17 @@ pub fn get_object_type_definitions<'a>(schema: &'a Document) -> Vec<&'a ObjectTy
         }).collect()
 }
 
+/// Returns all object type definitions in the schema.
+pub fn get_object_type_definitions_mut<'a>(schema: &'a mut Document) -> Vec<&'a mut ObjectType> {
+    schema
+        .definitions
+        .iter_mut()
+        .filter_map(|d| match d {
+            Definition::TypeDefinition(TypeDefinition::Object(t)) => Some(t),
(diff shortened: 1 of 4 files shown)


graphql/src/schema/ast.rs
```diff
@@ -92,6 +92,17 @@ pub fn get_object_type_definitions<'a>(schema: &'a Document) -> Vec<&'a ObjectTy
         }).collect()
 }
 
+/// Returns all object type definitions in the schema.
+pub fn get_object_type_definitions_mut<'a>(schema: &'a mut Document) -> Vec<&'a mut ObjectType> {
```
leodasvacas: The lifetime annotation here can be elided.

graphql/src/schema/ast.rs
```diff
@@ -92,6 +92,17 @@ pub fn get_object_type_definitions<'a>(schema: &'a Document) -> Vec<&'a ObjectTy
...
+        .definitions
+        .iter_mut()
+        .filter_map(|d| match d {
+            Definition::TypeDefinition(TypeDefinition::Object(t)) => Some(t),
+            _ => None,
+        }).collect()
+}
+
 /// Returns all interface definitions in the schema.
 pub fn get_interface_type_definitions<'a>(schema: &'a Document) -> Vec<&'a InterfaceType> {
```
leodasvacas: same here.

graphql/src/schema/ast.rs
```diff
@@ -103,11 +114,35 @@ pub fn get_interface_type_definitions<'a>(schema: &'a Document) -> Vec<&'a Inter
         }).collect()
 }
 
+/// Returns all interface definitions in the schema.
+pub fn get_interface_type_definitions_mut<'a>(
```
leodasvacas: Same.

server/http/src/server.rs
```diff
@@ -204,8 +204,8 @@ mod tests {
 
                     // The output schema must include the input schema types
                     assert_eq!(
-                        ast::get_named_type(&input_schema.document, &"User".to_string()),
-                        ast::get_named_type(&output_schema.document, &"User".to_string())
+                        ast::get_named_type(&input_schema.document, &"User".to_string()).is_some(),
+                        ast::get_named_type(&output_schema.document, &"User".to_string()).is_some(),
```
leodasvacas: You could improve this test to be also be a test for `__typename` by asserting the full output schema.

leoyvens (changes requested):
Comments about style and testing, otherwise this is great.

server/http/src/server.rs
Jannis: Yeah, I'll add a test, but `graphql/src/schema/api.rs` feels like the better place to add it — since that's where all the API schema tests live.

Jannis: I've added a better test for `__typename` being present in `graphql/src/schema/api.rs`.

Jannis:
@leodasvacas Everything should be addressed now.

leoyvens (approved):

adamsoffer:
@Jannis So this is strange. After pulling down this update, I get this error now: `Unhandled Rejection (Error): Name "__typename" must not begin with "__", which is reserved by GraphQL introspection.`

Setting `addTypename` to false doesn't make it go away. I guess apollo prefers to only use that name internally. Are you able to reproduce this?

leodasvacas:
It's likely the case that Apollo expects that we can respond to the `__typename` introspection query, but that field should not be present in the schema (since it should be implicitly availible for all entities).

This would agree with [the spec](https://facebook.github.io/graphql/draft/#sec-Type-Name-Introspection) that says:

> GraphQL supports type name introspection at any point within a query by the meta‐field __typename: String! when querying against any Object, Interface, or Union. It returns the name of the object type currently being queried.
This field is implicit and does not appear in the fields list in any defined type.

Could we find a more general way to implement this that doesn't involve injecting a `__typename` field into every entity?

adamsoffer:
@leodasvacas Gotcha. Yeah that makes sense.

leodasvacas:
I've reopened #431 to keep tracking this.

PR #556, opened 2018-11-06
runtime: Implement env.abort

### LINKED ISSUE #551: Implement the "abort" host export
AssemblyScript requires an "abort" export, which I believe is used in asserts and throws. In our module we have it's signature pointing at index 0, but we never implemented it, so we don't use the error information and log an unhelpful "Unimplemented function at 0" when abort is called in a mapping. We should implement it to log the error information and return a trap.

leodasvacas:
Resolves #551


runtime/wasm/src/host_exports.rs
```diff
@@ -65,6 +65,34 @@ where
...
 
+    pub(crate) fn abort(
+        &self,
+        message: Option<String>,
+        file_name: Option<String>,
+        line_number: Option<u32>,
+        column_number: Option<u32>,
+    ) -> Result<(), HostExportError<impl ExportError>> {
+        let message = message
+            .map(|message| format!("message: `{}`", message))
```
Jannis: Can we format the message without backticks around it?

leodasvacas: Hmm maybe backticks aren't right here but I do like to mark it somehow, Rust backtraces use single quotes `'message'` for the panic message, would that be better?

Jannis: I feel the : already marks it clear enough.

leodasvacas: Done

PR #557, opened 2018-11-06
runtime: Make error message more user friendly

leodasvacas:
Saw a user hit this so maybe we can improve it.

### DIFF
--- a/runtime/wasm/src/host_exports.rs
+++ b/runtime/wasm/src/host_exports.rs
@@ -206,7 +206,7 @@ where
             .find(|abi| abi.name == unresolved_call.contract_name)
             .ok_or_else(|| {
                 HostExportError(format!(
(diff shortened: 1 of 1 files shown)


runtime/wasm/src/host_exports.rs
```diff
@@ -206,7 +206,7 @@ where
             .find(|abi| abi.name == unresolved_call.contract_name)
             .ok_or_else(|| {
                 HostExportError(format!(
-                    "Unknown contract \"{}\" called from WASM runtime",
+                    "Could not find ABI for contract \"{}\", try adding it to the 'abis' section of the subgraph manifest",
```
Jannis: Let's wrap this string around so the line isn't so long. Other than that, I like this very much. 👍

leodasvacas: Done

PR #560, opened 2018-11-08
Add optional Elasticsearch logging per subgraph

### LINKED ISSUE #530: Add an optional setup flag for enabling per-subgraph Elasticsearch logging
Depends on https://github.com/graphprotocol/tasks/issues/420.

Jannis:
Resolves #530.

This introduces the following changes:

1. Add a new `ElasticDrain` for slog that writes logs to Elasticsearch.
2. Add an `elastic_logger(config)` helper that creates an async logger based on `ElasticDrain`.
3. Add a new `SplitDrain<D1, D2>` for slog and a `split_logger(drain1, drain2)` helper that creates an async logger that forwards logs to two drains in parallel.
4. Add `--elasticsearch-endpoint`, `--elasticsearch-user` and `--elasticsearch-password` options to `graph-node`. If these are set, the `SubgraphInstanceManager` creates a split logger to log to both the terminal (default) AND Elasticsearch.


core/src/log/elastic.rs
```diff
@@ -0,0 +1,188 @@
...
+            level: record.level(),
+            meta,
+        };
+
+        // Build the document URL
+        let mut document_url = reqwest::Url::parse(self.config.general.endpoint.as_str())
+            .expect("invalid Elasticsearch URL");
+        document_url
+            .path_segments_mut()
+            .expect("failed to set the Elasticsearch documenet path")
```
leodasvacas: typo, though 'documenet' would be a cool word.

core/src/log/elastic.rs
```diff
@@ -0,0 +1,188 @@
...
+        let response = client
+            .put(document_url)
+            .basic_auth(
+                self.config.general.username.clone(),
+                self.config.general.password.clone(),
+            ).json(&log)
+            .send()?;
+
+        // Return an error if the server returned an error response
+        response.error_for_status().map(|_| ())
```
leodasvacas: Is the response completely worthless here? Not worth logging (to a logger that doesn't go to ES, of course)?

core/src/log/split.rs
```diff
@@ -0,0 +1,74 @@
...
+}
+
+/// An slog `Drain` that forwards log messages to two other drains.
+struct SplitDrain<D1, D2>
+where
+    D1: Drain,
+    D2: Drain,
+{
+    drain1: Box<D1>,
+    drain2: Box<D2>,
```
leodasvacas: This `SplitDrain` is really elegant, but why the `Box`es?

core/src/subgraph/instance_manager.rs
```diff
@@ -68,7 +75,21 @@ impl SubgraphInstanceManager {
...
                 SubgraphStart(manifest) => {
-                    let logger = logger.new(o!("subgraph_id" => manifest.id.clone()));
+                    let term_logger = logger.new(o!("subgraph_id" => manifest.id.clone()));
+                    let logger = elastic_config
+                        .clone()
+                        .map(|elastic_config| {
+                            split_logger(
+                                term_logger.clone(),
+                                elastic_logger(ElasticDrainConfig {
+                                    general: elastic_config.clone(),
```
leodasvacas: This `clone()` seems unnecessary.

core/src/log/elastic.rs
```diff
@@ -0,0 +1,188 @@
...
+
+/// General configuration parameters for Elasticsearch logging.
+#[derive(Clone, Debug)]
+pub struct ElasticLoggingConfig {
+    /// The Elasticsearch service to log to.
+    pub endpoint: String,
+    /// The Elasticsearch username.
+    pub username: String,
+    /// The Elasticsearch password (optional).
+    pub password: Option<String>,
```
leodasvacas: In the ES API, what does an empty user or password mean? If the username is really optional, then it should be an `Option` here, you can do `unwrap_or("")` when calling `basic_auth`.

leoyvens (changes requested):
Nice, a few comments. From reading the slog code it's difficult to visualize the logs, could you perhaps document in a comment what a log ends up looking like?

core/src/log/split.rs
Jannis: Don't I need those because `D1` and `D2` have unknown sizes at compile time?

leodasvacas: The `Sized` bound, which guarantees the type (whatever type it is) has a known size at compile time, is implicit.

leodasvacas:
> failed sends will cause the logger to panic due to it being fused.

Seems bad, I don't think we want ES failures to cause a panic. Also what do you think of this comment:
> From reading the slog code it's difficult to visualize the logs, could you perhaps document in a comment what a log ends up looking like?

Jannis:
@leodasvacas I've added a comment with an example log document here: https://github.com/graphprotocol/graph-node/pull/560/files#diff-3baf393cfb430e71fed9c2dce1311e35R102.

About the ES errors: indeed, that would be bad. We have a couple of options how to handle this:

1. Just ignore errors.
2. Dump errors to the standard error output with `eprintln!`.
3. Retry sending the log messages (in an async loop if necessary).

2 could lead to garbled terminal output, because the terminal logger would still log asynchronously.
3 could lead to weird UI behavior if a UI assumes logs are append-only. Could be mitigated with an async queue / channel.

So go with 3 using our a channel, Tokio task and `retry`?

leodasvacas:
@Jannis could the split logger still log to the terminal even if ES fails? I'd be fine with skipping ES failures for now, as long as both the log and the failure itself are still logged to the terminal.

Jannis:
@leodasvacas Not with the current logic, but I can change `SplitDrain::log` to ignore errors from the two sub-drains.

leodasvacas:
@Jannis sounds good.

leodasvacas:
So, I think it's fine to proceed on failure, but we need to log the failure somewhere so we know what's going on.

Jannis:
Ah, yeah. We do.

leodasvacas:
@Jannis where exactly?

Jannis:
@leodasvacas I have an idea. Should have that working in a few minutes.

leodasvacas:
@Jannis Ah, I understood "Yes, we already do" rather than "Yes, we should do". Great!

leoyvens (dismissed):

leodasvacas:
@Jannis rustdoc is trying to run your comment as a doctest. One way to fix this is to use ```ignore

leoyvens (approved):

PR #562, opened 2018-11-09
Rewrite tests based on new Store, SubgraphInstanceManager, RuntimeHost, etc interfaces

### LINKED ISSUE #401: Investigate non deterministic test failure
[This travis build](https://travis-ci.org/graphprotocol/graph-node/jobs/431062769#L1149) failed the store test `find_string_less` on this assertion:
https://github.com/graphprotocol/graph-node/blob/daba3e88a5fef85f919a872650ed752c88cf9335/store/postgres/tests/store.rs#L491-L495
The returned name was in fact "Cindini".

There is clearly some non-determinism here as this test has always passed before, but this failure is highly suspicious and likely indicates some bug. Perhaps the non-determinism is in the order of the elements in the vector? Assigning @fordN to investigate.

### LINKED ISSUE #434: Add tests disabled after block stream work back

timmclean:
Also fixes all (?) compiler warnings except for the internal Diesel warnings, renames a few types and field names for clarity (StoreKey -> EntityKey, StoreOrder -> EntityOrder, etc), and removes a bunch of dead code.

A few tests have been rewritten but are marked `#[ignore]`. Those tests are broken because of known bugs.

Resolves #434 
Resolves #401

### DIFF
--- a/store/postgres/tests/store.rs
+++ b/store/postgres/tests/store.rs
@@ -9,14 +9,11 @@ extern crate hex;
 use diesel::pg::PgConnection;
 use diesel::*;
 use std::fmt::Debug;
-use std::mem;
 use std::panic;
 use std::str::FromStr;
 use std::sync::Mutex;
 
-use graph::components::store::{
-    EventSource, StoreFilter, StoreKey, StoreOrder, StoreQuery, StoreRange,
-};
+use graph::components::store::{StoreFilter, StoreKey, StoreOrder, StoreQuery, StoreRange};
 use graph::data::store::scalar;
 use graph::prelude::*;
 use graph::web3::types::H256;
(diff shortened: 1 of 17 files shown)


store/postgres/tests/store.rs
```diff
@@ -1478,65 +1120,56 @@ fn revert_block() {
 }
 
 #[test]
-#[cfg(any())]
+#[ignore]
```
leodasvacas: Could you comment above this test with the issue number?

leoyvens (changes requested):
Lovely to get the tests back, thanks for the effort! Just a couple of comments.

timmclean:
@leodasvacas done!

leoyvens (approved):

PR #564, opened 2018-11-09
Improve Elasticsearch logging by sending logs in batches

Jannis:
This makes `WithErrorDrain` obsolete.


leodasvacas:
You're using `use` to import macros which was introduced in the latest stable. You can either use `#[macro_use]` to keep supporting the older Rust versions or remove the older versions from the test matrix.

core/src/log/elastic.rs
```diff
@@ -125,28 +130,123 @@ pub struct ElasticDrainConfig {
 /// ```
 pub struct ElasticDrain {
     config: ElasticDrainConfig,
+    error_logger: Logger,
+    logs: Arc<RwLock<Vec<ElasticLog>>>,
```
leodasvacas: We never call `.read` on this, so it can just be a mutex.

leoyvens (changes requested):

core/src/log/elastic.rs
Jannis: Good point!

Jannis:
@leodasvacas Addressed the macro importing issue and switched over to a `Mutex`. Care to take another look?

leoyvens (approved):

PR #566, opened 2018-11-09
Add --block-polling-interval flag

### LINKED ISSUE #563: Make block ingestor interval configurable
This will help speed up running the graph-node in an automated integration test workflow with ganache, since the polling interval is the main source of latency for block processing (when using ganache).

timmclean:
Allow users to specify how often the block ingestor polls for new blocks. Requested by a user.

@Jannis are these good flag and env var names?

Resolves #563

### DIFF
--- a/node/src/main.rs
+++ b/node/src/main.rs
@@ -141,11 +141,11 @@ fn async_main() -> impl Future<Item = (), Error = ()> + Send + 'static {
                 .hide_env_values(true)
                 .help("Password to use for Elasticsearch logging"),
         ).arg(
-            Arg::with_name("block-polling-interval")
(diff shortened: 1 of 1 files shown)


Jannis (changes requested):
If prefix them with ETHEREUM / --ethereum. That'll make them pretty long but provided a better basis for future option/conf partitioning by subsystem.

If you want them to be shorter, you could drop the 'block', so it's just ETHEREUM_POLLING_INTERVAL.

The code itself looks good.

timmclean:
Yeah that sounds pretty good!

timmclean:
@Jannis updated with better flag name

Jannis (approved):

PR #568, opened 2018-11-09
Fix boolean queries

timmclean:
Queries involving a Not, NotIn, or Equal filter on a boolean value were causing database errors. It seems that Postgres doesn't allow boolean jsonb values to be cast to actual booleans. Converting the jsonb value to a string before casting to boolean fixes the problem.


Jannis (approved):
Weird but ok...

Issue #532, opened 2018-10-29
When a subgraph is deployed, parse it, write it to Postgres as entities using `subgraphs` as the subgraph ID


Jannis:
Depends on https://github.com/graphprotocol/graph-node/issues/531.

Includes deploying this implicit subgraph of subgraphs at startup.

<s>Includes listening to the store for subgraph updates from the GraphQL servers to keep the available subgraphs and their schemas up to date.</s>

timmclean:
If we have all query nodes storing schemas from all subgraphs, will that be a scalability issue?

leodasvacas:
I didn't quite get the requirements for this one.

> keep the available subgraphs and their schemas up to date

Why is this necessary? Isn't it just a matter of putting the subgraph in the store upon deployment?

Jannis:
There are two aspects here:

1. We'll want to be able to query which subgraphs are deployed, what data sources they have etc. We'll want to do this through GraphQL, so when a subgraph is deployed, its manifest needs to
     1. either be stored either as entities (so we can reuse the existing query logic), or
     2. be queryable through a dedicated GraphQL resolver much like `StoreResolver`.

For this, putting the subgraphs into the store is enough.

2. When a subgraph `foo` is queried via e.g. `http://somehost.com/foo/graphql` or subscribed to via `ws://somehost.com/foo`, the GraphQL/HTTP server or GraphQL/WebSocket server need to be able to load the schema of `foo` in order to pass it to the GraphQL runner.

That's what the `keep the available subgraphs and their schemas up to date` is about. Now, it may be fine to resolve the subgraph's schema using the store when the request comes in. In that case we can lazily resolve it in the GraphQL/HTTP or GraphQL/WebSocket server and then pass it to the GraphQL runner. Perhaps that's what we should start with.

@leodasvacas @timmclean What do you think?

timmclean:
Sounds good to me. I like the lazy schema resolution :+1:
My sense is that if all query nodes can handle any query, then we shouldn't have any subgraph-specific state in the query nodes (caching or otherwise). But schemas are small, so I might be overestimating the issue

Jannis:
[...]

Issue #531, opened 2018-10-29
Define a GraphQL schema for a built-in "subgraph of subgraphs"


Jannis:
Could be based on the schema that we have in graph-cli for validating subgraph manifests.
https://github.com/graphprotocol/graph-cli/blob/master/manifest-schema.graphql

fordN:
```graphql
type Subgraph @entity {
    id: ID!
    name: String!
    manifest: SubgraphManifest!
    description: String!
    numberEntities: Int!
    createdAt: BigInt!
    updatedAt: BigInt!
    repository: GithubRepository
}

type GithubRepository @entity {
    id: ID!
    url: String!
    tags: [GithubTag]
    contributors: [GithubContributor!]
}

type GithubContributor @entity {
    id: ID!
    userName: String!
    avatarUrl: String!
    Repos: [GithubRepository!] @derivedFrom(name: "contributors")
}

type GithubTag @entity {
    id: ID!
    name: String!
}

type SubgraphManifest @entity {
    id: ID!
    specVersion: String!
    schema: Schema!
    dataSources: [DataSource!]!
}

union DataSource = EthereumContractDataSource

type EthereumContractDataSource @entity {
    id: ID!
    kind: String!
    name: String!
    source: EthereumContractSource!
    mapping: EthereumContractMapping!
}

type EthereumContractSource @entity {
    id: ID!
    address: String!
    abi: String!
}

type EthereumContractMapping @entity {
    id: ID!
    kind: String!
    apiVersion: String!
    language: String!
    file: File!
    entities: [String!]!
    abis: [EthereumContractAbi!]!
    eventHandlers: [EthereumContractEventHandlers]!
}

type EthereumContractAbi @entity {
    id: ID!
    name: String!
    file: File!
}

type EthereumContractEventHandler @entity {
    id: ID!
    event: String!
    handler: String!
}
```

nenadjaja:
Looks good @fordN ! 
A few things:
1. What's a `GithubTag` and what do we need it for?
2. Do we have the information about what's the network that a contract is deployed to?

nenadjaja:
I am a little confused about why do we use `__subgraphs`. In my understanding, introspection is predefined in GraphQL, and there are 3 options for introspection queries: `__type`, `__schema` and `__typename`. Are we defining our own introspection query?

Jannis:
@fordN A couple of remarks:

- Move `description` from `Subgraph` to `SubgraphManifest` and make it optional.
[...]

PR #575, opened 2018-11-14
Subgraph of subgraphs

### LINKED ISSUE #102: Make components that receive or emit events use `EventProducer` and `EventConsumer`
These recently introduced abstractions are not used by the components that pre-date it, we should make sure all components that can use it do so.

### LINKED ISSUE #533: Add `/subgraphs` handler to the GraphQL/HTTP and GraphQL/WebSocket servers
Depends on https://github.com/graphprotocol/graph-node/issues/532.

leoyvens:
Yo dawg, we heard you like subgraphs, so we put your subgraph in a subgraph, so you can query your queries.

This makes subgraph metadata queriable like any other subgraph, under `/by-id/__subgraphs` or its alias `/subgraphs`. The schema is based on the one defined [here](https://github.com/graphprotocol/graph-node/issues/531#issuecomment-434935048) but with some changes:

- No `name` or `updatedAt`.
- No `numberEntities` for now, need to figure that out as a follow-up.
- The type `Link` is changed to just `String`.

The key parts are:`SubgraphProvider` that was modified to insert the built-in subgraph on init and inserts subgraph entities when they are deployed; `graph/src/data/subgraph/schema.rs` that contains the Rust representation of the schema, it's conversion from our Rust representation of manifests and conversions to write the schema to the store as an entity.

Random UUIDs are used as the id for sub-entities. I could check for duplicates, but I think that's silly since:

> the probability to find a duplicate within 103 trillion version 4 UUIDs is one in a billion.

I did a few refactorings along the way. `Mapping` is no longer cloned anywhere, since that's expensive. Schema validation was moved from `graphql` to the `graph` crate, this required duplicating ~10 lines of helpers between the crates but it allows `Schema::parse` to both parse and validate.

Part of #532 (entity count still missing)
Resolves #533
Resolves #102 

For testing, this is the query that queries all subfields:
```graphql
query {
 subgraphs {
    id
    createdAt
    manifest {
      id    
      specVersion
      description
      schema
      repository
      dataSources {
         id
        kind
        name
        source {
          id
          address
          abi
        }
        mapping {
          id
          kind
          apiVersion
          language
          file
          entities
          abis {
            id
            name
            file
          }
          eventHandlers {
            id
            event
            handler
          }
        }
      }
    }
  }
}
```


core/src/subgraph/provider.rs
```diff
@@ -32,11 +33,26 @@ where
             schema_event_sink,
             resolver,
             subgraphs_running: Arc::new(Mutex::new(HashSet::new())),
+            store,
         };
 
+        provider.send_builtin_schema(&include_str!("subgraph.graphql"), SUBGRAPHS_ID.to_owned());
```
Jannis: Can we rename the file to `subgraphs.graphql` to match the `/subgraphs` URL?

core/src/subgraph/provider.rs
```diff
@@ -105,35 +123,36 @@ where
...
-                        // Send events to trigger subgraph processing
-                        Box::new(self_clone.send_add_events(subgraph).from_err())
-                    },
-                ),
+                            .as_secs(),
+                    ).write_to_store(&*self_clone.store)
+                    .map_err(|err| {
+                        error!(
+                            self_clone.logger,
+                            "failed to write subgraph entity to store: {}", err
```
Jannis: Capital `F`

Jannis: Plus I think I'd just make this `Failed to write subgraph to store: {}`.

graph/src/data/subgraph/schema.rs
```diff
@@ -0,0 +1,343 @@
...
+
+use super::SubgraphId;
+use components::store::{EntityKey, EntityOperation, Store};
+use data::store::Value;
+use failure::Error;
+use std::collections::HashMap;
+use uuid::Uuid;
+
+/// ID of the subgraph of subgraphs.
+pub const SUBGRAPHS_ID: &str = "__subgraphs";
```
Jannis: Let's make this `subgraphs` instead of `__subgraphs`. I thought the `__` made sense initially but I don't think it's necessary.

graph/src/data/subgraph/schema.rs
```diff
@@ -0,0 +1,343 @@
...
+const EVENT_SOURCE: &str = "SubgraphAdded";
```
Jannis: And I think this one we should just leave empty for now.

graph/src/data/subgraph/schema.rs
```diff
@@ -0,0 +1,343 @@
...
+impl SubgraphEntity {
+    pub fn new(source_manifest: &super::SubgraphManifest, created_at: u64) -> Self {
+        Self {
+            id: source_manifest.id.clone(),
+            manifest: SubgraphManifest::from(source_manifest),
+            created_at,
+        }
+    }
+
+    pub fn write_to_store(self, store: &impl Store) -> Result<(), Error> {
```
Jannis: I deployed the same subgraph a few times (I think) and somehow ended up with duplicate entities in the store:

image

Jannis: One thing I would do (may not be related) is used deterministic instead of random IDs for all entities. I.e.

* `<subgraph ID>-schema`
* `<subgraph ID>-data-source-0`
* `<subgraph ID>-data-source-0-abi-0`
* `<subgraph ID>-data-source-1`

etc. Unless we're already protecting against writing the same subgraph to the DB twice?

runtime/wasm/src/host.rs
```diff
@@ -139,14 +143,30 @@ impl RuntimeHost {
 
         // Start the WASMI module runtime
         let module_logger = logger.clone();
-        let module_config = config.clone();
+        let data_source_name = config.data_source.name.clone();
```
Jannis: Any particular reason not to keep `config` and `data_source` around?

store/postgres/src/store.rs
```diff
@@ -294,15 +294,27 @@ impl Store {
...
+        self.apply_set_operation_with_conn(conn, operation, &block_ptr_to.hash_hex())
+    }
+
+    /// Applies a set operation in Postgres.
+    fn apply_set_operation_with_conn(
+        &self,
+        conn: &PgConnection,
+        operation: EntityOperation,
+        op_event_source: &str,
+    ) -> Result<(), Error> {
```
Jannis: Any particular reason (other than that we're not currently using this) not to return the number of rows here?

Jannis (changes requested):
This is looking great! I have a few minor remarks and questions (and one potential bug) but that's all. 😀

runtime/wasm/src/host.rs
Jannis: I should've read the PR description more thoroughly.

store/postgres/src/store.rs
leodasvacas: We don't use it anywhere, and it's not something we want to return in the public interface since we try to keep the `Store` trait backend-agnostic while the number of affected rows is specific to relational dbs.

graph/src/data/subgraph/schema.rs
leodasvacas: Good catch. We have a check for duplicate deployments, but not for writing the subgraph entity twice, which is something we want to be able to do, for example after upgrading `subgraphs.graphql`. Deterministic IDs for all subentities seems like a robust fix, I'll try that.

core/src/subgraph/instance.rs
```diff
@@ -24,22 +21,41 @@ impl<T> SubgraphInstanceTrait<T> for SubgraphInstance<T>
...
-            hosts: hosts,
+        if !errors.is_empty() {
+            let joined_errors = errors
+                .into_iter()
+                .map(Result::unwrap_err)
+                .map(|e| e.to_string())
+                .collect::<Vec<_>>()
+                .join(", ");
+            return Err(format_err!(
+                "errors starting data sources: {}",
```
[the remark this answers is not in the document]
Jannis: Not with the term by itself but I _think_ (may be wrong) we're not talking about data sources as something that we `start`, `run` or so, unlike with mappings. I'm undecided here though.

graph/src/data/subgraph/schema.rs
Jannis: Ok, I'm coming around to that (it does make sense 😉). `subgraph-added` perhaps? (I'm a fan of separating things with dashes, blame Clojure.)

store/postgres/src/store.rs
Jannis: Yep, ok, let's do it this way then. 😄

core/src/subgraph/instance.rs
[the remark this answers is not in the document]
leodasvacas: The only error this can currently return is "No ABI entry found for the main contract of data source", which is about the data source itself, not about the mapping. So if a user is trying to debug this, "data source" is the more helpful term imo.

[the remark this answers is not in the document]
Jannis: How about `Errors loading data sources` then?

graph/src/data/subgraph/schema.rs
leodasvacas: Clojure style it is.

core/src/subgraph/instance.rs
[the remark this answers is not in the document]
leodasvacas: I like it.

leodasvacas:
@Jannis I've changed all ids to be deterministic. This avoids duplicating the sub-entities when you, for example, deploy, remove and then re-deploy the same subgraph.

Jannis:
Great. I've filed https://github.com/graphprotocol/graph-node/issues/577 to track the incorrect query results for references such as `dataSources` and `eventHandlers` after identifying that the issue is not caused by this PR.

Jannis (approved):

PR #576, opened 2018-11-15
Fix orderBy for all types of attribute values

Jannis:
Changing the way we store entities (serializing them as tagged values) broke `orderBy`. This PR fixes it by adding the missing `->> 'data'` traversal before converting the attribute value for comparisons with e.g. `::numeric`.


leodasvacas:
Could we get a store test for this? There are similar tests.

Jannis:
Yes, sir! 🤦‍♂️

leoyvens (approved):

PR #578, opened 2018-11-16
Rewrite EntityFilter -> SQL translation to support nested filters

### LINKED ISSUE #577: Non-derived 1:n references are broken

Bug

Querying a 1:n reference that was explicitly set (as opposed to derived with `@derivedFrom`) currently returns all entities of the referenced type, rather than just the ones that are actually referenced.

Only the referenced entities are returned. This issue surfaced in #575 but is really caused by our `store_filter()` function generating a flat/broken sequence of `AND`/`OR` expressions for Postgres.

Jannis:
This fixes #577 by correctly generating the correct SQL to query just the entities referenced in a list of references.

It also fixes the SQL generated for any other nested `where: { ..., or: { ... }, and: { ... } }` filters.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -664,7 +664,7 @@ dependencies = [
 
 [[package]]
 name = "graph"
-version = "0.4.0"
+version = "0.4.1"
 dependencies = [
  "backtrace 0.3.9 (registry+https://github.com/rust-lang/crates.io-index)",
  "ethabi 6.0.1 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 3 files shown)


Jannis:
I've added a store test (`find_nested_and_or`) that fails on master and passes on this branch. 

The test queries users with an `And([Or([Equal("id", "1), Equal("id", "2"))])])` filter. On master, this incorrectly returns all three test users, which is exactly the problem described in #577.
```
test find_where_nested_and_or ... thread 'tokio-runtime-worker-0' panicked at 'assertion failed: `(left == right)`
  left: `["1", "2", "3"]`,
 right: `["1", "2"]`', store/postgres/tests/store.rs:443:9
```

leoyvens (approved):
I was never too confident on this code, good to see it revamped! Good to merge after travis.

PR #596, opened 2018-11-22
Fix GraphiQL subscription URL and use the configured WS port

### LINKED ISSUE #332: Identify and fix GraphQL subscriptions sometimes not working
No entity changes seem to be logged.

1. Verify that there are any notifications in Postgres.
2. Verify whether the entity change listener thread is somehow broken.
3. Verify why change events are not being logged.
and so on.

Jannis:
This may well fix #332. Without this fix, GraphiQL isn't even able to establish the WebSocket connection because it's using the wrong URL.

### DIFF
--- a/graph/src/components/server/query.rs
+++ b/graph/src/components/server/query.rs
@@ -90,5 +90,6 @@ pub trait GraphQLServer: EventConsumer<SchemaEvent> {
     fn serve(
         &mut self,
         port: u16,
+        ws_port: u16,
     ) -> Result<Box<Future<Item = (), Error = ()> + Send>, Self::ServeError>;
 }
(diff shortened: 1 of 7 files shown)


Jannis:
I realize this won't fix. I'll put that test back in and make it pass too.

leoyvens (approved):
I'm sure I've seen subscriptions working in the past, so I doubt this is what #332 referred to. But then if we don't know what it referred to there is no sense in keeping the issue around...

Jannis:
Since this currently breaks _everyone_ trying to use subscriptions from GraphiQL _all the time_, let's merge it and leave #332 open until we've checked the failing entity changes test as well.

Issue #583, opened 2018-11-19
Add more fields to subgraphs and their manifests


Jannis:
These fields are:

* `numberOfEntities` (only in the database, not in manifest)
* `imageUrl` (only in the database, separate table)
* `description` (manifest and database)
* `repository` (manifest and database)
* `network` (field on the data source)

PR #597, opened 2018-11-22
Add `description`, `repository` and `network` to manifest

leoyvens:
All as optional strings, and add them to the schema for the subgraph of subgraphs. Part of #583 (will do `numberOfEntites` separately). Example query to test this:
```
query {
 subgraphs {
    id
    createdAt
    manifest {
      id    
  	description
  	repository
      dataSources {
         id
        kind
        name
        network
      }
    }
  }
}
```
Requires https://github.com/graphprotocol/graph-cli/pull/176 for the `network` field to validate on the cli.

### DIFF
--- a/docs/subgraph-manifest.md
+++ b/docs/subgraph-manifest.md
@@ -31,7 +31,7 @@ Any data format which has a well-defined 1:1 mapping with [IPLD Canonical Format
 | --- | --- | --- |
 | **kind** | *String | The type of data source. Possible values: *ethereum/contract*|
 | **name** | *String* | The name of the source data. Will be used to generate APIs in mapping, and also for self-documentation purposes |
-| **network** | *String* | For blockhains this describes which network the subgraph targets. For Ethereum this could be, for example, "mainnet" or "rinkeby". |
+| **network** | *String* | For blockchains this describes which network the subgraph targets. For Ethereum this could be, for example, "mainnet" or "rinkeby". |
 | **source** | [*EthereumContractSource*](#151-ethereumcontractsource) | The source data on a blockchain such as Ethereum |
 | **mapping** | [*Mapping*](#152-mapping) | The transformation logic applied to the data prior to being indexed |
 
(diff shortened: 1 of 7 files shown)


docs/subgraph-manifest.md
```diff
@@ -29,6 +31,7 @@ Any data format which has a well-defined 1:1 mapping with [IPLD Canonical Format
 | --- | --- | --- |
 | **kind** | *String | The type of data source. Possible values: *ethereum/contract*|
 | **name** | *String* | The name of the source data. Will be used to generate APIs in mapping, and also for self-documentation purposes |
+| **network** | *String* | For blockhains this describes which network the subgraph targets. For Ethereum this could be, for example, "mainnet" or "rinkeby". |
```
Jannis: Blockhains may be the next big thing but for now we're focusing on blockchains. 😉

graph/src/data/subgraph/schema.rs
```diff
@@ -66,7 +66,10 @@ impl SubgraphManifest {
         let mut entity = HashMap::new();
         entity.insert("id".to_owned(), id.clone().into());
         entity.insert("specVersion".to_owned(), self.spec_version.into());
-        entity.insert("description".to_owned(), self.description.into());
+        let description = self.description.unwrap_or_default();
+        entity.insert("description".to_owned(), description.into());
```
Jannis: This will set the description attribute to an empty string, right? We can work with that but it would be cleaner to set it to `Value::Null` instead. Same with the other optional fields.

Jannis (changes requested):
Great! Two minor remarks only. Thanks for updating the docs as well!

graph/src/data/subgraph/schema.rs
leodasvacas: That's much better.

leodasvacas:
@Jannis addressed review.

graph/src/data/store/mod.rs
```diff
@@ -203,6 +203,18 @@ impl From<Address> for Value {
     }
 }
 
+impl<T> From<Option<T>> for Value
```
Jannis: Nice!

Jannis (approved):

PR #602, opened 2018-11-26
graph, store, graphql: Add `entityCount` field to subgraph entities

### LINKED ISSUE #583: Add more fields to subgraphs and their manifests

leodasvacas:
Resolves #583 (the url should be tracked in a different issue).

This adds the `entityCount` field. Instead of storing it with the data and incrementing on every entity insert, which would be error-prone, this introduces the concept of a "calculated field", which is a field of an entity that is not in the store but rather calculated on the fly. We can probably also use this for `__typename`.

### DIFF
--- a/graphql/src/store/resolver.rs
+++ b/graphql/src/store/resolver.rs
@@ -5,6 +5,7 @@ use std::result;
 use std::sync::Arc;
 
 use graph::components::store::*;
+use graph::data::subgraph::schema::{SUBGRAPHS_ID, SUBGRAPH_ENTITY_TYPENAME};
 use graph::prelude::*;
 
 use prelude::*;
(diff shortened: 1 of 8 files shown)


graphql/src/store/resolver.rs
```diff
@@ -182,6 +183,30 @@ where
                 _ => true,
             }).unwrap_or(true)
     }
+
+    /// Compute special fields that are not stored such as as `entityCount`.
+    fn add_computed_fields(
```
Jannis: Overall, this is great, but I would do this part differently. Adding computed fields like this will mean that we'll always count entities for all subgraphs, even if `entityCount` is not part of the query.

What I would do instead is resolve fields like this lazily. Here's what I'd recommend:

1. Change the signature of `Resolver::resolve_scalar_value` so that it takes the object type and the field name:
   ```rust
   fn resolve_scalar_value(
       &self,
       object_type: &s::ObjectType,
       field_name: &q::Name,
       scalar_type: &s::ScalarType,
       value: Option<&q::Value>,
   ) -> q::Value {
   ```
2. Implement `resolve_scalar_value` in `StoreResolver`. Something like this should do:
   ```rust
    fn resolve_scalar_value(
        &self,	
        object_type: &s::ObjectType,
        field_name: &q::Name,
        scalar_type: &s::ScalarType,
        value: Option<&q::Value>,
    ) -> q::Value {
        match field_name.as_str() {
          "entityCount" => {
            ... parse subgraph ID, if possible, fall back to value ...
            ... obtain entity count ...
          },
          _ => value
        }.and_then(|value| value.coerce(scalar_type)).unwrap_or(q::Value::Null)
    }
   ```

Jannis: But I'm not completely convinced of my approach because it makes adding support for GraphQL interfaces harder (which don't have `ObjectType`). Perhaps let's go with your approach first.

PR #621, opened 2018-12-02
Fix subgraph_deployments migration not running

Jannis:
This ensures that the `subgraph_deployments` table migration is always being run. Previously, it was added as part of the old `subgraphs` table migration and so Diesel wouldn't run it if the `subgraphs` table already existed (or rather, if its migration had already been run once).

Giving the `subgraph_deployments` migration its own ID ensures it is being performed.


leodasvacas:
This duplicates the existing `up` migration. Shouldn't we be moving rather than duplicating the migration?

Jannis:
@leodasvacas Oh, yes, I forgot to commit the removal from the other file.

timmclean (approved):

PR #624, opened 2018-12-03
Don't include __typename in schema introspection

### LINKED ISSUE #431: Support __typename meta field on entities

Jannis:
I was considering to drop generating the __typename field at all but that has other implications (like query execution not returning values for the field at all because it wouldn't exist on the types).

Fixes #431 (or rather, the bug caused by the initial implementation of `__typename`).


Jannis:
Obsoleted by #628.

PR #627, opened 2018-12-04
Add cost column to subgraph_deployments table

### LINKED ISSUE #619: Add cost column to subgraph deployments
Use `sum(cost)` instead of `count(*)` to measure load on a graph-node

timmclean:
Resolves #619


Jannis (changes requested):
Shouldn't we also write a cost of `1` when adding a deployment? Otherwise the costs of all nodes will be `0` unless we change it manually.

timmclean:
@Jannis happens automatically :)
The `DEFAULT 1` column constraint applies to future insertions too, not just the initial column creation

Jannis (approved):

PR #628, opened 2018-12-04
Insert `__typename` at query time

### LINKED ISSUE #431: Support __typename meta field on entities

leodasvacas:
This makes it so we no longer add `__typename` to the schema (which does not conform to the spec) neither store it in the store (which is inefficient) but instead insert it at query time.

Fixes #431 for good. Alternative to #624.


adamsoffer:
Nice one! Thanks for this.

PR #630, opened 2018-12-05
Asc bump

leoyvens:
This is a companion to https://github.com/graphprotocol/graph-cli/pull/183 and https://github.com/graphprotocol/graph-ts/pull/29 in the adventure to upgrade our AS commit.

The most relevant changes were how imports are organized (they now go all in one module) and a small change to array deserialization. Most tests were updated, some tests that were hard to update and are not so relevant I removed.


runtime/wasm/src/module/test.rs
```diff
@@ -595,13 +455,3 @@ fn big_int_arithmetic() {
     let result: BigInt = module.heap.asc_get(result_ptr);
     assert_eq!(result, BigInt::from(1));
 }
-
-#[test]
-fn abort() {
-    let mut module = test_module(mock_data_source("wasm_test/abort.wasm"));
```
Jannis: How come we're no longer able to test handling errors thrown by the module?

runtime/wasm/src/module/mod.rs
```diff
@@ -603,196 +610,62 @@ impl ModuleImportResolver for EnvModuleResolver {
...
+pub struct ModuleResolver;
 
-impl ModuleImportResolver for StoreModuleResolver {
-    fn resolve_func(&self, field_name: &str, _signature: &Signature) -> Result<FuncRef, Error> {
+impl ModuleImportResolver for ModuleResolver {
+    fn resolve_func(&self, field_name: &str, signature: &Signature) -> Result<FuncRef, Error> {
+        let signature = signature.clone();
         Ok(match field_name {
-            "set" => FuncInstance::alloc_host(
-                Signature::new(&[ValueType::I32, ValueType::I32, ValueType::I32][..], None),
```
Jannis: With these signatures gone (which saves a ton of boilerplate, so that's nice), is there still something that asserts that the call signature matches the function implementation signature?

Jannis (dismissed):
I only have two questions, everything looks spot on.

runtime/wasm/src/module/mod.rs
leodasvacas: Yes, wasmi itself will error if you return the wrong signature (which makes me wonder what's the point if it knows the signature already...)

runtime/wasm/src/module/test.rs
leodasvacas: hmm I'll try something with this one.

leodasvacas: Thanks for pointing this out, came up with a test that is better than the original.

Jannis (approved):

runtime/wasm/src/module/test.rs
Jannis: Great!

leodasvacas:
ah, new release today, need to upgrade the formatting. Let's merge this without doing that. @Jannis feel free to merge whenever it's best for you to coordinate with the other PRs.

Jannis:
👍

PR #632, opened 2018-12-06
Refactor subgraph entity logic 

timmclean:
In preparation for addition of subgraph status fields.

Part of #584

### DIFF
--- a/graph/src/data/subgraph/schema.rs
+++ b/graph/src/data/subgraph/schema.rs
@@ -1,23 +1,22 @@
 //! See `core/src/subgraph/subgraphs.graphql` for corresponding graphql schema.
 
 use super::SubgraphId;
-use components::store::{EntityKey, EntityOperation, Store};
-use data::store::Value;
-use failure::Error;
+use components::store::{EntityKey, EntityOperation};
+use data::store::{Entity, Value};
 use std::collections::HashMap;
 
 /// ID of the subgraph of subgraphs.
-pub const SUBGRAPHS_ID: &str = "subgraphs";
+lazy_static! {
+    pub static ref SUBGRAPHS_ID: SubgraphId = SubgraphId::new("subgraphs").unwrap();
+}
 
 /// Type name of the root entity in the subgraph of subgraphs.
 pub const SUBGRAPH_ENTITY_TYPENAME: &str = "Subgraph";
 
-const EVENT_SOURCE: &str = "subgraph-added";
-
 #[derive(Debug)]
 pub struct SubgraphEntity {
     id: SubgraphId,
-    manifest: SubgraphManifest,
+    manifest: SubgraphManifestEntity,
     created_at: u64,
 }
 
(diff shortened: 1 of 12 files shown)


core/src/subgraph/definitions.rs
```diff
@@ -0,0 +1,169 @@
...
+    pub fn new(store: Arc<S>) -> Self {
+        Self { store }
+    }
+
+    fn write(
+        &self,
+        entity_typename: impl Into<String>,
+        entity_id: impl Into<String>,
+        data: impl Into<Entity>,
+    ) -> Result<(), Error> {
```
leodasvacas: This is nice to reduce duplication.

graph/src/data/subgraph/schema.rs
```diff
@@ -1,352 +1,9 @@
...
-use components::store::{EntityKey, EntityOperation, Store};
-use data::store::Value;
-use failure::Error;
-use std::collections::HashMap;
 
 /// ID of the subgraph of subgraphs.
-pub const SUBGRAPHS_ID: &str = "subgraphs";
+lazy_static! {
+    pub static ref SUBGRAPHS_ID: SubgraphId = SubgraphId::new("subgraphs").unwrap();
+}
```
leodasvacas: Fancy!

graph/src/data/subgraph/mod.rs
```diff
@@ -198,7 +198,7 @@ pub struct UnresolvedMapping {
 }
 
 // Avoid deriving `Clone` because cloning a `Module` is expensive.
-#[derive(Debug)]
+#[derive(Clone, Debug)]
```
leodasvacas: As the comment notes, having `Clone` here is not good. Are you using this anywhere?

leodasvacas:
Having written this originally, I'm biased for having a Rust struct that directly corresponded to each entity, following the old quote:
> Show me your flowchart and conceal your tables, and I shall continue to be mystified. Show me your tables, and I won't usually need your flowchart; it'll be obvious

Given the loc reduction in this PR, it might have been overkill, but I still prefer it because checking the schema against the structs is easier than checking the schema against the code.

A third opinion might be good here.

Jannis:
I can see benefits in both approaches. @timmclean Perhaps it would help if you described your thinking here?

The main advantage I see in Tim's approach is that all write logic lives in a single place and that you can extend it naturally without having to touch the existing subgraph structs.

Personally, I am a fan of "stupid data", data that doesn't carry functionality around with it, apart from conversion helpers, getters, setters etc.

graph/src/data/subgraph/schema.rs
timmclean: :sunglasses:

graph/src/data/subgraph/mod.rs
timmclean: Ah shoot, I should've noticed that comment!

timmclean:
So, the main reason for this refactor was to rework the code in a way that would let me add methods to modify parts of an entity without having to rewrite the whole entity (e.g. setting the subgraph status field). That doesn't preclude having the extra layer of structs though.

I think it's important that we have some sort of schema definition somewhere for any data we're keeping in a database (I've dealt with undocumented MongoDB schema-less messes before, it gets hard to follow). In this case though, we have that schema already: https://github.com/graphprotocol/graph-node/blob/master/core/src/subgraph/subgraphs.graphql
So I think my first argument would be that the structs are redundant, at least as far as documenting the schema goes.

Proceeding on that line of thought, there are types like `EthereumContractSource`, which has the same fields as `Source` in `graph/src/data/subgraph/mod.rs`. For schema documentation, I get why it's desirable to have the separate type layer. I think given that we have the GraphQL schema already though, the extra translation layer just adds complexity.

That said, in cases where the entity in the store does not have an exact correspondence to an existing type, we may want to add additional types (especially if we need to read data out of subgraph entities -- something we don't do yet). I think I want to argue in favour of trying to make those types generally useful instead of specific to the storage mechanism, so that way code that uses this storage mechanism can interoperate better with other parts of the codebase that may use similar types, but stored or serialized in a different way.

graph/src/data/subgraph/mod.rs
timmclean: I'm going to rework the write methods to take the values by ref and clone internally, instead of cloning before passing the subgraph manifest to the write method

leodasvacas:
The argument that the previous code was excessively self-documenting can go either way. I agree that it's best when types are not specific to the storage mechanism, and they aren't, they are specific to the data model.

An amusing observation on the current approach: Reading the `write_to_store` methods, they are so symmetrical in their signature and implementation that's it's not hard to imagine the existence of a `#[derive(WriteToStore)]`. That macro would be complex, but if it becomes a common pattern the possibility for abstraction is there.

I guess what you see as painfully redundant I see as beautifully symmetrical 😄

Jannis:
It seems the perhaps best option would be to implement `Into<Entity>` for these structs AND have a dedicated store interface to add extra fields.

That way the structs contain the serialization like but they don't require knowledge of the store.

leodasvacas:
@Jannis by extra fields do you mean the relationships?

leodasvacas:
Sounds good to me, so `write_ops` folds through the `write_ops` of the subtypes? I like it, cleaner than taking the store as a parameter.

timmclean:
Yup! Some nice purity there. I think I'll move things back to graph/src/data again now that it's not keeping a Store handle around

store/postgres/src/store.rs
```diff
@@ -615,25 +603,32 @@ impl StoreTrait for Store {
...
+    fn apply_entity_operations(
         &self,
-        operation: EntityOperation,
-        op_event_source: String,
+        operations: Vec<EntityOperation>,
+        event_source: EventSource,
     ) -> Result<(), Error> {
         let conn = self.conn.lock().unwrap();
-        self.apply_set_operation_with_conn(&conn, operation, &op_event_source)
+        conn.transaction(|| self.apply_entity_operations_with_conn(&conn, operations, event_source))
```
leodasvacas: What's the effect of this extra transaction?

store/postgres/src/store.rs
```diff
@@ -615,25 +603,32 @@ impl StoreTrait for Store {
             panic!("transact_block_operations must transact a single block only");
         }
 
+        // All operations should apply only to entities in this subgraph
```
Jannis: 👍

graph/src/data/subgraph/schema.rs
```diff
@@ -1,102 +1,89 @@
...
         Self {
             id: source_manifest.id.clone(),
-            manifest: SubgraphManifest::from(source_manifest),
+            manifest: SubgraphManifestEntity::from(source_manifest),
             created_at,
         }
     }
 
-    pub fn write_to_store(self, store: &impl Store) -> Result<(), Error> {
+    pub fn write_ops(self) -> Vec<EntityOperation> {
```
Jannis: I'm ok calling variables `something_ops` but could we spell this one out and make it `write_operations`?

timmclean: Done!

store/postgres/src/store.rs
leodasvacas: Makes sense, were we failing to do this before or was this in another part of the code?

timmclean: Wasn't being done before, although less of an issue back then because this function used to only commit one entity change

Jannis:
@timmclean Approved. Let's ignore the Travis error, I think it's one of those "test timed out" issues that we see sometimes but that are not persistent.

timmclean:
Hmmm that's unfortunate. Confirmed that `cargo test` still passes locally though after merging to master

leodasvacas:
@timmclean Some of our tests will randomly hang. I think they were always racy, but after the tokio update it started actually happenning. I tried putting some delays around to improve things, but it wasn't sufficient. I'll just restart the specific build in travis when that happens.

PR #640, opened 2018-12-11
Allow to query any subgraph present in the store

### LINKED ISSUE #534: Allow to query any subgraph that is in Postgres, not just the ones currently deployed

leodasvacas:
Resolves #534. The core logic is in the new store method `schema_of` which fetches schemas by id, with a cache layer to improve performance. This turns out to simplify a lot of code. We're also careful to check in the store if the subgraph is deployed before querying or subscribing to it with the new method `is_deployed`.

In `server/http/src/service.rs` we were doing some validations when serving graphiql that were redundant (and a bit inconsistent) with those done when receiving queries. So now we unconditionally serve graphiql.

In `server/websocket`, we had logic to cancel outstanding subscriptions when a subgraph was removed. However the `SchemaEvent::Removed` that triggered that no longer exist. I chose to remove the logic entirely, since in light of the changes done in this PR we'd need to implement it differently. It was certainly not sufficient to prevent the server from being overloaded with subscriptions anyways, so we're likely to revisit that soon.


server/websocket/src/server.rs
```diff
@@ -189,30 +116,38 @@ where
                             // Obtain the subgraph ID or name that we resolved the request to
                             let subgraph_id = subgraph_id.lock().unwrap().clone().unwrap();
 
+                            // Check if the subgraph is deployed
+                            match store2.is_deployed(&subgraph_id) {
+                                Err(_) |
+                                Ok(false) => {
+                                    error!(logger, "Failed to establish WS connection, subgraph not deployed";
+                                                "subgraph" => subgraph_id.to_string(),
```
Jannis: Indentation is slightly off here (macros, I know...).

server/websocket/src/server.rs
```diff
@@ -189,30 +116,38 @@ where
...
+                                }
+                                Ok(true) => (),
+                            }
+
+                            // Get the subgraph schema
+                            let schema = match store2.schema_of(subgraph_id.clone()) {
+                                Ok(schema) => schema,
+                                Err(e) => {
+                                    error!(logger, "Failed to establish WS connection, could not find schema";
+                                                "subgraph" => subgraph_id.to_string(),
```
Jannis: Indentation

server/websocket/src/server.rs
```diff
@@ -189,30 +116,38 @@ where
...
                             let subgraph_id = subgraph_id.lock().unwrap().clone().unwrap();
 
+                            // Check if the subgraph is deployed
+                            match store2.is_deployed(&subgraph_id) {
+                                Err(_) |
+                                Ok(false) => {
+                                    error!(logger, "Failed to establish WS connection, subgraph not deployed";
+                                                "subgraph" => subgraph_id.to_string(),
+                                    );
+                                    return Ok(());
```
Jannis: Will returning here close the WS connection? I think that's desirable. And if we can it would be good to report a 'subgraph not found' type error back to client.

server/websocket/src/server.rs
```diff
@@ -189,30 +116,38 @@ where
...
                                 &logger,
-                                subgraphs.clone(),
-                                subgraph_id.clone(),
+                                schema,
                                 ws_stream,
                                 graphql_runner.clone(),
                             );
-
-                            // Setup cancelation.
-                            let guard = CancelGuard::new();
```
Jannis: Can't we still do this, but based on subgraph stopped events?

store/postgres/src/store.rs
```diff
@@ -6,16 +6,19 @@ use diesel::sql_types::Text;
 use diesel::{delete, insert_into, select, update};
 use filter::store_filter;
 use futures::sync::mpsc::{channel, Sender};
+use lru_time_cache::LruCache;
```
Jannis: I was thinking last night: I wonder what cache Leo will pick for this, LRU would make sense. And here we go 😍

store/postgres/src/store.rs
```diff
@@ -823,6 +828,66 @@ impl SubgraphDeploymentStore for Store {
...
+            .filter(subgraph_id.eq(id.to_string()))
+            .execute(&*self.conn.lock().unwrap())?;
+        Ok(match deployment_count {
+            0 => false,
+            1 => true,
+            _ => unreachable!(), // The `subgraph_id` column is `UNIQUE`
+        })
+    }
+
+    fn schema_of(&self, subgraph_id: SubgraphId) -> Result<Schema, Error> {
```
Jannis: I'd prefer `subgraph_schema` here, just because you don't have to wonder what the schema is `of` when reading code that uses this function.

leodasvacas: Sounds good.

store/postgres/src/store.rs
leodasvacas: LRU does seem good.

Mental notes I made here: A built-in cache is of course the fastest possible, but the cache size will not really scale with the number of query nodes. Since queries have no affinity to particular query nodes (and we probably want to avoid that), then it's likely that all query nodes will have very similar contents in their caches. In fact, having more query nodes means it will take more initial queries to start getting consistent cache hits across all nodes.

Having it still better than not having it, but we should be aware of the limitations.

server/websocket/src/server.rs
leodasvacas: I considered this, but it feels like a big mixup of index nodes and query nodes. An approach I thought might be best, is that the store resolver interrupts the stream of responses with an error when the subgraph is no longer deployed. What do you think, would that be an abuse of the resolver?

Jannis: Yep, let's create an issue for it and move on. The resolver approach might work. We do return a stream there that we can decide to cancel when it makes sense.

store/postgres/src/store.rs
Jannis: True. I think we can live with that.

server/websocket/src/server.rs
Jannis: https://github.com/graphprotocol/graph-node/issues/642

server/websocket/src/server.rs
leodasvacas: Moved this up to reject the connection earlier with a 404.

leodasvacas:
@Jannis All comments addressed.

server/http/src/service.rs
```diff
@@ -193,28 +142,44 @@ where
...
+                )))
+            }
+            Ok(false) => {
+                return Box::new(future::err(GraphQLServerError::ClientError(format!(
+                    "Subgraph {} not deployed",
+                    id
+                ))))
+            }
+            Ok(true) => (),
         }
```
timmclean: Is it possible here to still allow querying of subgraphs without a deployment? I'm hoping to use "no deployment" to mean paused, and it would be nice to still allow queries in that case

store/postgres/src/store.rs
```diff
@@ -823,6 +828,66 @@ impl SubgraphDeploymentStore for Store {
...
+
+        let raw_schema = if subgraph_id == SubgraphId::new(SUBGRAPHS_ID).unwrap() {
+            // The subgraph of subgraphs schema is built-in.
+            include_str!("subgraphs.graphql").to_owned()
+        } else {
+            let manifest_entity = self
+                .get(EntityKey {
+                    subgraph_id: SubgraphId::new(SUBGRAPHS_ID).unwrap(),
+                    entity_type: MANIFEST_ENTITY_TYPENAME.to_owned(),
+                    entity_id: format!("{}{}", subgraph_id, MANIFEST_SUFFIX),
```
timmclean: How about extracting these to functions in schema.rs that either create the entity ID, or create the whole EntityKey? `SubgraphManifestEntity::id(subgraph_id)`, `SubgraphManifestEntity::entiyt_key(subgraph_id)` or something

store/postgres/src/store.rs
```diff
@@ -823,6 +828,66 @@ impl SubgraphDeploymentStore for Store {
...
+                        subgraph_id
+                    ))
+                }
+            }
+        };
+        let mut schema = Schema::parse(&raw_schema, subgraph_id.clone())?;
+        schema.document = api_schema(&schema.document)?;
+
+        if !self.schema_cache.lock().unwrap().contains_key(&subgraph_id) {
+            trace!(self.logger, "schema cache miss"; "id" => subgraph_id.to_string());
```
timmclean: It looks like this should be on line 855, maybe a separate trace here for updating the cache

timmclean:
nice, it's pretty great that we don't need the schema events anymore!

server/http/src/service.rs
leodasvacas: When we have subgraph status, we can change `is_deployed` to `is_queriable` and have that mean whatever we want it to mean.

graph/src/data/subgraph/schema.rs
```diff
@@ -71,6 +68,10 @@ struct SubgraphManifest {
 }
 
 impl SubgraphManifest {
+    pub fn id(subgraph_id: &SubgraphId) -> String {
+        format!("{}-manifest", subgraph_id.to_string())
```
timmclean: SubgraphId is Display

leodasvacas:
@Jannis @timmclean Ready for re-review.

PR #646, opened 2018-12-13
Add status and block counts to subgraph entity

### LINKED ISSUE #584: Write per-subgraph status to Postgres (“block N of M”)
Write these values to the subgraph entity when we're updating the subgraph block pointer and applying entity operations.

timmclean:
Add status and progress fields to subgraph of subgraphs, and update them during block stream processing.

Resolves #584


Jannis (approved):
No objections!

PR #648, opened 2018-12-13
Fix subscription stream termination issue

### LINKED ISSUE #332: Identify and fix GraphQL subscriptions sometimes not working

timmclean:
Resolves #332 

`Store::subscribe` returns a stream of entity changes. This stream is actually a `mpsc` `Receiver`. Each time `Store::subscribe` is invoked, a new `Sender`/`Receiver` pair is created, and the `Sender` is saved into the `Store` instance.

When the `Receiver` is dropped, the corresponding `Sender` will return an error the next time the store tries to send it an entity change, at which point that `Sender` (and information about the subscription it represents) is removed from the `Store`. However, the `Sender` error was not being suppressed properly, so it brought down the store's global entity change stream. As soon as any subscription was removed, all subscriptions would silently stop receiving entity changes.

The bug is really just a case of accidentally using `.and_then(|_| Ok(()))` instead of `.then(|_| Ok(()))`, which would have correctly suppressed the error. (I wonder if a Clippy lint could catch this mistake?)

### DIFF
--- a/store/postgres/src/store.rs
+++ b/store/postgres/src/store.rs
@@ -226,13 +226,17 @@ impl Store {
             stream::iter_ok::<_, ()>(matches).for_each(move |(id, sender)| {
                 let logger = logger.clone();
                 let subscriptions = subscriptions.clone();
-                sender
-                    .send(change.clone())
-                    .map_err(move |_| {
-                        debug!(logger, "Unsubscribe"; "id" => &id);
-                        subscriptions.write().unwrap().remove(&id);
-                    })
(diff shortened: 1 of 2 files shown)


Jannis:
Oh wow, what a subtle mistake. Good catch!

Jannis (approved):

PR #649, opened 2018-12-14
Fix passing positive BigInt starting with `0x80` to AS

### LINKED ISSUE #626: Adding BigInts does not work
bug

See how I am setting `userStats.totalHolders`:
```
  if (userStats == null) {
    userStats = new UserStats()
    userStats.totalHolders = BigInt.fromI32(0)

  }
```
Everytime the mapping runs, I add to it like so:
```
userStats.totalHolders = userStats.totalHolders.plus(BigInt.fromI32(1))
```
But, when I query the value, it will actually change the crazy, changing from positive to negative values.

So when I watched it closely, I noticed it actually randomly will flip the sign. What typically happens is it will count up to say, `50`, and then all of a sudden another mapping runs and it is at `-50` , and then it starts counting up again, gets positive, and then will continually flip to negative on repeat. So it usually never gets higher than `150` or so because by then it will have flipped, and it will start climbing back up. 

Interestingly, when I preset the values to a large number like `100000` instead of `0`, it works perfectly. I am watching the count go up to `103500` without flipping the sign. 

But also interestingly, when I am adding `BigInts` like `23124513` plus `1234412` it will flip signs too. So I don't really know what is happening.

leoyvens:
Turns out it was an upstream bug with `to_signed_bytes_le`, it flips the sign of specific numbers such as `128`. Nice catch and investigation @davekaj!

Fixes #626.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -693,7 +693,7 @@ dependencies = [
  "hex 0.3.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "ipfs-api 0.5.0-alpha2 (registry+https://github.com/rust-lang/crates.io-index)",
  "lazy_static 1.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
- "num-bigint 0.2.1 (git+http://github.com/leodasvacas/num-bigint?branch=fix-to-signed-bytes)",
+ "num-bigint 0.2.2 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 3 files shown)


Jannis (dismissed):
Oops. Good catch you two!

leodasvacas:
Upstream was super-quick to fix and relase this, so I've switched back to it, on version `0.2.2`. Needs to be re-approved.

Jannis (approved):
Awesome!

PR #650, opened 2018-12-17
Use a DB connection pool

### LINKED ISSUE #396: Remove Mutex on Store and use a connection pool
Currently, we only allow one thread at a time to access the store, which limits the number of queries that can run in parallel to 1. There's no reason we have to have this limitation though. We could use a crate like [r2d2](https://docs.diesel.rs/r2d2/) to manage a connection pool so that we could run many SQL queries in parallel. Databases are like a giant `RefCell` conceptually, and they provide their own primitives for managing concurrency and preventing race conditions, so we can safely make the store methods take `&self` instead of `&mut self`.

A nice side effect of fixing this is that we would no longer need `store.lock().unwrap()` at every store method call site.

leodasvacas:
Resolves #396. I didn't tweak the pool configurations, in the spirit of "any pooling is better than no pooling". But it'd be good if those who know more about DBs could take a look at the [builder defaults](https://docs.diesel.rs/r2d2/struct.Builder.html) and check if they're ok for our use. We can fine tune later.

### DIFF
--- a/store/postgres/src/store.rs
+++ b/store/postgres/src/store.rs
@@ -2,6 +2,7 @@ use diesel::dsl::sql;
 use diesel::pg::Pg;
 use diesel::pg::PgConnection;
 use diesel::prelude::*;
+use diesel::r2d2::{self, ConnectionManager, Pool};
 use diesel::sql_types::Text;
 use diesel::{delete, insert_into, select, update};
 use filter::store_filter;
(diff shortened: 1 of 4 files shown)


store/postgres/src/store.rs
```diff
@@ -81,14 +82,25 @@ impl Store {
...
         let logger = logger.new(o!("component" => "Store"));
 
-        // Connect to Postgres
-        let conn = PgConnection::establish(config.postgres_url.as_str())
-            .expect("failed to connect to Postgres");
+        #[derive(Debug)]
+        struct ErrorHandler(Logger);
+        impl r2d2::HandleError<r2d2::Error> for ErrorHandler {
+            fn handle_error(&self, error: r2d2::Error) {
+                error!(self.0, "postgres connection error"; "error" => error.to_string())
```
Jannis: Our log messages start with a capital letter. 😉

leodasvacas:
@Jannis Thanks for the review! Corrected the log message.

Issue #154, opened 2018-07-09
Add support for GraphQL interfaces


Jannis:
Plan:

1. Add "__typename" meta field to all entities.
2. When resolving interface objects, create an `Or(Equals(__typename, ...), Equals(__typename, ...))` filter for all types that implement the interface in the schema.

leodasvacas:
This will require a way to disambiguate IDs. Suppose there are two entities that implement the interface `Pet`, one of type `Dog`, the other of type `Cat`, and they both have instances with `id: "Fred"`. If an entity `PetOwner` has `pets: [Pet]` then with the value `pets: ["Fred"]` there's currently no way to disambiguate if that's Fred the dog, Fred the cat or both. This is because IDs don't carry type information.

A possible solution is to introduce a new variant `TypedId` to `Value` which stores both type and ID. For backwards compatibility and to keep the common case simple we probably want objects that are not coming from an interface field to keep simple string IDs. The question then is how do we expose that in the mappings so that it's not a pain to work with `TypedId`s. Some ideas by @Jannis:

Could generate types like `class Cat { class Reference extends Pet.Reference { id: string } }`, so you can do `owner.pets = [new Cat.Reference("some-id")]`

Another alternative would be to just require `Array<Pet>`. So you’d do `owner.pets = [new Cat("some-id"), new Dog("some-id"), Dog.load("other-id") as Dog]`. We can also have methods for conveniently downcasting `Pet` to the class of each of the implementing entities, such as `as_dog(): Dog | null`.

We'll also need to go back to somehow storing `__typename`, so that we return the concrete typename when querying an interface, which we stopped doing in #628. We could store it along with user data in the `data` column, but that duplicates the `entity` column so we might want to at least not store that information twice.

PR #652, opened 2018-12-18
Fix graphql introspection bug, cache map of interfaces to types

leoyvens:
This fixes a bug that prevented doing the `__type(name: String!)` introspection query, adds a test for `possibleTypes` introspection in interfaces and adds a map of interfaces to implementing types to our processed `Schema`, which is some groundwork for #154. I removed `MockSubgraphProvider`, it was getting in the way a bit and was just dead code.


graph/src/data/query/error.rs
```diff
@@ -151,6 +152,9 @@ impl fmt::Display for QueryExecutionError {
             StoreError(e) => {
                 write!(f, "Store error: {}", e)
             }
+            MissingNameInTypeIntrospection => {
```
Jannis: This is `MissingArgumentError`, we don't have to invent a more specific one here IMHO.

graph/src/data/schema.rs
```diff
@@ -1,28 +1,61 @@
...
         schema.add_subgraph_id_directives(id);
+
         Ok(schema)
     }
 
+    pub fn type_for_interface(&self, interface_name: &query::Name) -> Option<&Vec<ObjectType>> {
+        self.types_for_interface.get(interface_name)
+    }
+
+    // #[cfg(test)]
```
Jannis: Should we remove this is if it's commented out anyway?

graphql/src/execution/execution.rs
```diff
@@ -717,9 +723,9 @@ fn merge_selection_sets(fields: Vec<&q::Field>) -> q::SelectionSet {
 /// Coerces argument values into GraphQL values.
 pub fn coerce_argument_values<'a, R1, R2>(
     ctx: ExecutionContext<'a, R1, R2>,
-    object_type: &'a s::ObjectType,
-    field: &'a q::Field,
-) -> Result<HashMap<&'a q::Name, q::Value>, Vec<QueryExecutionError>>
+    object_type: &s::ObjectType,
+    field: &q::Field,
+) -> Result<HashMap<q::Name, q::Value>, Vec<QueryExecutionError>>
```
Jannis: What's the reason for cloning the names here rather than referencing them? Since this is only about arguments (and there won't be that many arguments in any given query) this is probably ok from a performance perspective, although I generally tried to clone as little as possible in the execution.

Jannis (changes requested):
Small remarks and questions. This is sweet!

graph/src/data/schema.rs
leodasvacas: Ouch forgot that.

graphql/src/execution/execution.rs
leodasvacas: Yes this is a small pessimization. Thinking about it again, I can avoid this by using`lazy_static`, I'll try that.

leodasvacas:
@Jannis Thanks for the review, all points addressed. I've managed to not pessimize the arguments data structure by using `lazy_static`, the diff is also smaller now.

Jannis (approved):

PR #657, opened 2018-12-21
Create partial Postgres indexes for attributes during subgraph startup 

fordN:
This PR adds a process on subgraph startup that builds an index on each entity attribute in the subgraph. 
- Created Postgres function `build_attribute_index()` for adding subgraph entity specific indexes to the entities table. 
- Parse Schema on startup for Entity attribute values and create entity indexing operations for each.

To do: Create indexing cleanup process which removes all indexes when a subgraph is deprecated, and pass all index operations to the Postgres function as a single array instead of calling it once per index.

### DIFF
--- a/core/src/subgraph/provider.rs
+++ b/core/src/subgraph/provider.rs
@@ -4,6 +4,8 @@ use std::sync::Mutex;
 
 use graph::prelude::{SubgraphAssignmentProvider as SubgraphAssignmentProviderTrait, *};
 use graph_graphql::schema::ast as sast;
+use graph_graphql::prelude::schema::{Definition, Document, TypeDefinition};
+
 
 pub struct SubgraphAssignmentProvider<L, S> {
     logger: Logger,
@@ -75,73 +77,25 @@ where
                         ));
                     }
 
-                    // Build indexes for each entity attribute pair in the Subgraph
-                    let mut indexing_ops = vec![];
-                    for (entity_number, schema_type) in subgraph
-                        .schema
-                        .document
-                        .definitions
-                        .clone()
-                        .into_iter()
-                        .enumerate()
-                    {
-                        match schema_type {
-                            schema::Definition::TypeDefinition(definition) => match definition {
-                                schema::TypeDefinition::Object(schema_object) => {
-                                    for (attribute_number, entity_field) in
-                                        schema_object.fields.into_iter().enumerate()
-                                    {
-                                        let subgraph_id = subgraph.id.clone();
-                                        let index_name: String = format!(
-                                            "{}_{}_{}_idx",
-                                            subgraph_id, entity_number, attribute_number,
-                                        );
-                                        let field_value_type = match sast::get_valid_value_type(
-                                            &entity_field.field_type,
-                                        ) {
-                                            Ok(value_type) => value_type,
-                                            Err(_) => continue,
-                                        };
-
-                                        let (index_type, index_operator) = match field_value_type {
-                                            ValueType::Boolean
-                                            | ValueType::BigInt
-                                            | ValueType::Bytes
-                                            | ValueType::Float
-                                            | ValueType::ID
-                                            | ValueType::Int => {
-                                                (String::from("btree"), String::from(""))
-                                            }
-                                            ValueType::String => {
-                                                (String::from("gin"), String::from("gin_trgm_ops"))
-                                            }
-                                        };
-                                        let attribute_name = entity_field.name;
-                                        let entity_name = schema_object.name.clone();
-
-                                        indexing_ops.push(AttributeIndexOperation {
-                                            subgraph_id,
-                                            index_name,
-                                            index_type,
-                                            index_operator,
-                                            attribute_name,
-                                            entity_name,
-                                        });
-                                    }
-                                }
-                                _ => (),
-                            },
-                            _ => (),
-                        }
-                    }
+                    // Build indexes for each entity attribute in the Subgraph
+                    let indexing_ops = attribute_indexing_operations(
+                        subgraph.id.clone(),
+                        subgraph.schema.document.clone(),
+                    );
                     self_clone
                         .store
                         .clone()
                         .build_entity_attribute_indexes(indexing_ops)
                         .map_err(|err| {
-                            error!(
+                            debug!(
                                 self_clone.logger,
-                                "Failed to create indexes on subgraph entities: {}", err
+                                "No subgraph entity indexes created: {}", err
+                            )
+                        })
+                        .map(|_| {
+                            info!(
+                                self_clone.logger,
+                                "Successfully created indexes for all attributes"
                             )
                         })
                         .ok();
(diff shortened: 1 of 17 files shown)


core/src/subgraph/provider.rs
```diff
@@ -137,3 +163,48 @@ impl<L, S> EventProducer<SubgraphProviderEvent> for SubgraphProvider<L, S> {
             .map(|s| Box::new(s) as Box<Stream<Item = SubgraphProviderEvent, Error = ()> + Send>)
     }
 }
+
+fn attribute_indexing_operations(
```
fordN: I'm open to suggestions for a better place to put this function, I was very close to putting it in /graph/src/data/subgraph/schema.rs, but that would have required importing `graph_graphql` into graph.

leodasvacas: A method on `Schema` seems better, the only real depedency on `graph_graphql` I see is `get_valid_value_type` which you can put somewhere in `graph::data::graphql`, the rest you can get directly from `graphql_parser`.

As a general note, I've wanted to move `graph_graphql::schema` into `graph` before, to be able to do more in `Schema` methods.

core/src/subgraph/provider.rs
```diff
@@ -137,3 +163,48 @@ impl<L, S> EventProducer<SubgraphProviderEvent> for SubgraphProvider<L, S> {
...
+                        };
+                    let (index_type, index_operator) = match field_value_type {
+                        ValueType::Boolean
+                        | ValueType::BigInt
+                        | ValueType::Bytes
+                        | ValueType::Float
+                        | ValueType::ID
+                        | ValueType::Int => (String::from("btree"), String::from("")),
+                        ValueType::String => (String::from("gin"), String::from("gin_trgm_ops")),
+                    };
```
leodasvacas: The choice of index type is tied to the storage backend so it belongs in the store implementation. How about replacing `index_type` and `index_operator` with `field_value_type` in  `AttributeIndexOperation`, and moving this match into the store?

leodasvacas:
Code looks good to me, but I don't have much experience with DB indexes so it'd be good to have a second review from @Jannis or @timmclean.

This does not handle list types in any special way, so on a field `foo: [Int]` are we building indexes for the `foo_contains` filter?

fordN:
@leodasvacas good reminder, I held off on `List` field indexes because it warranted further investigation.  I have finished that inquiry, landing on GIN index, so I'll add that in today.

graph/src/data/graphql/validation.rs
```diff
@@ -60,3 +63,12 @@ pub fn get_object_type_directive(object_type: &ObjectType, name: Name) -> Option
         .iter()
         .find(|directive| directive.name == name)
 }
+
+/// Returns the value type for a GraphQL field type.
+pub fn get_valid_value_type(field_type: &Type) -> Result<ValueType, Error> {
```
Jannis: I'm not sure this has anything to do with validation, so I'd put it next to the `attribute_index_operations()` function for now. And I'd call it `inner_type_name` instead of `get_valid_value_type`.

graph/src/data/subgraph/schema.rs
```diff
@@ -362,3 +364,35 @@ fn set_entity_operation(
         data: data.into(),
     }
 }
+
+pub fn attribute_indexing_operations(
```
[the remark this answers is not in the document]
Jannis: Or rather `attribute_index_definitions` if we rename the type.

graph/src/data/subgraph/schema.rs
```diff
@@ -362,3 +364,35 @@ fn set_entity_operation(
...
+    let mut indexing_ops = vec![];
+    for (entity_number, schema_type) in document.definitions.clone().into_iter().enumerate() {
+        if let Definition::TypeDefinition(definition) = schema_type {
+            if let TypeDefinition::Object(schema_object) = definition {
+                for (attribute_number, entity_field) in schema_object.fields.into_iter().enumerate()
+                {
+                    indexing_ops.push(AttributeIndexOperation {
+                        subgraph_id: subgraph_id.clone(),
+                        index_name: format!(
+                            "{}_{}_{}_idx",
```
Jannis: Do we even need the `_idx` suffix?

graphql/src/lib.rs
```diff
@@ -37,4 +37,6 @@ pub mod prelude {
     pub use super::store::{build_query, StoreResolver};
     pub use super::subscription::{execute_subscription, SubscriptionExecutionOptions};
     pub use super::values::{object_value, MaybeCoercible};
+
+    pub use graphql_parser::schema as parser_schema;
```
Jannis: Unused, let's remove.

store/postgres/migrations/2018-12-21-003727_create_attribute_indexing_procedure/down.sql
```diff
@@ -0,0 +1,2 @@
+--  DROP ATTRIBUTE INDEXING FUNCTION
+DROP FUNCTION BuildSubgraphEntityAttributeIndexes(attribute_array entity_attribute_type[]);
```
Jannis: The name `BuildSubgraphEntityAttributeIndexes` doesn't appear anywhere else. Is this supposed to be `build_attribute_index` maybe?

store/postgres/src/store.rs
```diff
@@ -459,6 +462,54 @@ impl Store {
...
+        .execute(conn)
+        .map_err(Error::from)
+        .and_then(move |row_count| match row_count {
+            1 => Ok(()),
+            _ => Err(format_err!(
+                "failed to create indexes for subgraph:{:?}",
+                operation.clone().subgraph_id,
+            )),
+        })
+        //
```
Jannis: Remove this empty comment.

store/postgres/src/store.rs
```diff
@@ -459,6 +462,54 @@ impl Store {
...
+            index_operator,
+            operation.attribute_name.clone(),
+            operation.entity_name.clone(),
+        ))
+        .execute(conn)
+        .map_err(Error::from)
+        .and_then(move |row_count| match row_count {
+            1 => Ok(()),
+            _ => Err(format_err!(
+                "failed to create indexes for subgraph:{:?}",
```
Jannis: Our convention is for non-panic errors to start with a capital letter. Also, there's a space missing after the `subgraph:`. I'd make it `{}` for the subgraph ID formatting too, if possible.

core/src/subgraph/provider.rs
```diff
@@ -95,6 +96,29 @@ where
...
+                        subgraph.schema.document.clone(),
+                    );
+                    self_clone
+                        .store
+                        .clone()
+                        .build_entity_attribute_indexes(indexing_ops)
+                        .map_err(|err| {
+                            debug!(
+                                self_clone.logger,
+                                "No subgraph entity indexes created: {}", err
```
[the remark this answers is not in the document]
fordN: I've updated the sql function to catch `relation already exists` errors, so this is no longer an expected result.  Accordingly,  I've added `SubgraphProviderError::IndexesBuildError` and used it here.

graph/src/data/subgraph/schema.rs
fordN: I included this to follow Postgres naming conventions as referenced here, https://gist.github.com/popravich/d6816ef1653329fb1745.
It is not completely necessary; when looking at the DB it should be clear to Graph developers that these relations are indexes, but it's always nice to be specific just in case. :)

fordN:
@Jannis I've updated the build indexes process to add indexes for List type fields and changed the json operator used in the list contains filter to ensure the new indexes are utilized.

graph/src/data/subgraph/schema.rs
Jannis: Ok, agreed.

store/postgres/src/store.rs
```diff
@@ -365,6 +368,12 @@ impl Store {
                 ValueType::ID => "",
                 ValueType::Int => "::bigint",
                 ValueType::String => "",
+                ValueType::List => {
+                    return Err(QueryExecutionError::OrderByNotSupportedError(
+                        "List".to_string(),
+                        "All".to_string(),
```
Jannis: `All` doesn't make sense to me. Perhaps we should add a different `QueryExecutionError` variant here, like `OrderByNotSupportedForType("List".to_string())`.

graph/src/data/subgraph/mod.rs
```diff
@@ -241,6 +241,8 @@ pub enum SubgraphAssignmentProviderError {
     /// Occurs when a subgraph's GraphQL schema is invalid.
     #[fail(display = "GraphQL schema error: {}", _0)]
     SchemaValidationError(failure::Error),
+    #[fail(display = "subgraph attribute indexes build error: {}", _0)]
```
Jannis: Error messages start with a capital letter. Also, this is a bit cryptic. Could we make it `"Error building index for subgraph \"{}\", entity \"{}\" and attribute \"{}\": {}"`?

And add the subgraph deployment ID, entity name and attribute name to the error, if possible.

Jannis:
@fordN Did another review pass here. Also, please note that the test suite is failing.

PR #662, opened 2019-01-04
Store row level operation type in history, use in revert

### LINKED ISSUE #660: Error reverting block: null value in column "data" violates non-null constraint
Seen on 0x subgraph:
```
Chain head pointer, number: 6927382, hash: 0x7fbc4c8a9aff2d4c285005a1b22007f7ebd866a27c170dcef2f9ea384dcdb897, component: BlockStream
Subgraph pointer, number: 6927352, hash: 0x8cc052f3c841987e7a7f8c63e7f9354228e94d94eb8794979a99f3ca89475c84, component: BlockStream
Reverting block to get back to main chain, block_hash: 0x8cc0…5c84, block_number: 6927352, component: BlockStream
Trying again after error in block stream reconcile: Error reverting block: null value in column "data" violates not-null constraint, component: BlockStream
Identify next step, component: BlockStream
Subgraph pointer, number: 6927352, hash: 0x8cc052f3c841987e7a7f8c63e7f9354228e94d94eb8794979a99f3ca89475c84, component: BlockStream
Chain head pointer, number: 6927382, hash: 0x7fbc4c8a9aff2d4c285005a1b22007f7ebd866a27c170dcef2f9ea384dcdb897, component: BlockStream
Reverting block to get back to main chain, block_hash: 0x8cc0…5c84, block_number: 6927352, component: BlockStream
Trying again after error in block stream reconcile: Error reverting block: null value in column "data" violates not-null constraint, component: BlockStream
Identify next step, component: BlockStream
Subgraph pointer, number: 6927352, hash: 0x8cc052f3c841987e7a7f8c63e7f9354228e94d94eb8794979a99f3ca89475c84, component: BlockStream
Chain head pointer, number: 6927383, hash: 0x44af4c5de55b58cda6ac8d67a34628c1444e1efff7ae444d0b047ba120b3308a, component: BlockStream
Reverting block to get back to main chain, block_hash: 0x8cc0…5c84, block_number: 6927352, component: BlockStream
Trying again after error in block stream reconcile: Error reverting block: null value in column "data" violates not-null constraint, component: BlockStream
Identify next step, component: BlockStream
Chain head pointer, number: 6927384, hash: 0xd41d142ebed3bff2a224f6fc13051e6bbce1b4d2102b5a9fb18106856b7852b1, component: BlockStream
Subgraph pointer, number: 6927352, hash: 0x8cc052f3c841987e7a7f8c63e7f9354228e94d94eb8794979a99f3ca89475c84, component: BlockStream
Reverting block to get back to main chain, block_hash: 0x8cc0…5c84, block_number: 6927352, component: BlockStream
Subgraph instance failed to run: Error reverting block: null value in column "data" violates not-null constraint, id: QmdecCGpn5FMQQRFzQCNuNB4wLjb9Qtph3BtfRzvpKAz5T
Terminating chain head updates, component: BlockStream
```

fordN:
Resolves #660 

Entity change triggers were tracking operation type only at a transaction level which is sometimes (~2%) inconsistent with the row level operation type. Revert operations were failing on the `entity_history` rows which were not correctly described by the transaction level operation type. 

To fix the entity change historical tracking and reversion process this PR will:
  - Store `op_id` in the `entity_history` table. 
  - Consolidate trigger functions down to just one function with logic that detects the row level operation type using `TG_OP`.
  - Use `entity_history.op_id` in revert operations and in the process reduce the number of times `entity_history` and `event_meta_data` (our largest tables) need to be joined together.


fordN:
@timmclean great point, we can move `source` to the `entity_history` table and no longer reference the `event_meta_data` table in our block revert procedures.  In that case the value of `event_meta_data` would be solely for analytics which could be replaced by Postgres logging of each transaction. The question becomes where and how is the best way to store transaction history?  I'm looking into processes to store and effectively use the Postgres logs now (Elastic Search maybe?)

fordN:
@timmclean  I'm leaning towards keeping the `event_meta_data` table since moving its data to `entity_history` will lead to data amplification (redundant rows) and the revert_block function is keying off the event source already anyways.   Thinking through potentially applying transactions from multiple blocks in one transaction warrants more time and discussion and the migration won't be intrusive to add in the future.

timmclean:
OK, that might be fine. Just so I understand better, a couple questions: What would the redundant rows look like in `entity_history`? Also, does this work with how block pointer updates happen, now that the block pointer is in an entity as well? See [here](https://github.com/graphprotocol/graph-node/blob/6825cfb6897b8646c37f566cbbed02f44a0d85fd/store/postgres/src/store.rs#L772) for how this happens -- it's just updated using EntityOperations. When reverting a block, will that block pointer change be reverted as well? (I think ideally it wouldn't, but that might be manageable. Note that the pointer is [currently reverted separately](https://github.com/graphprotocol/graph-node/blob/6825cfb6897b8646c37f566cbbed02f44a0d85fd/store/postgres/src/store.rs#L814))

timmclean:
How about bumping the migration date to today? There have been other migrations since that date

PR #670, opened 2019-01-07
Remove dead code in resolver

leoyvens:
Resolving a single object with no parent and no supplied `id` dosen't make much sense when you think about it. Unless it is some sort of unique global object? Which is not something that exists in our graphql API, so I'd rather just remove this confusing code, we can rethink this if we ever get a top level query that returns a single object but doesn't take an `id` parameter.

### DIFF
--- a/graphql/src/store/resolver.rs
+++ b/graphql/src/store/resolver.rs
@@ -261,7 +261,6 @@ where
         &self,
         parent: &Option<q::Value>,
         field: &q::Name,
-        field_definition: &s::Field,
         object_type: &s::ObjectType,
         arguments: &HashMap<&q::Name, q::Value>,
     ) -> Result<q::Value, QueryExecutionError> {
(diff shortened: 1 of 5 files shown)


graphql/src/store/resolver.rs
```diff
@@ -288,21 +287,7 @@ where
...
-                        parent,
-                        field_definition,
-                        object_type,
-                    );
-
-                    query.range = Some(EntityRange { first: 1, skip: 0 });
-
-                    self.store.find(query)?.into_iter().next()
-                }
+                _ => panic!("top level queries must either take an `id` or return a list"),
```
Jannis: If this can't be hit by a user query then I'm ok with the `panic!`, otherwise I'd prefer returning a `QueryExecutionError`.

Jannis (approved):

graphql/src/store/resolver.rs
leodasvacas: It's impossible for a user to hit this because the `Query` type is given by us, and all generated fields in it either take an `id` argument for querying a single object or return a list when querying multiple objects. Does that sound right to you?

Jannis: Yep!

PR #671, opened 2019-01-07
Add subgraph versioning

### LINKED ISSUE #643: Update subgraph entities to include subgraphs, versions and deployments

### LINKED ISSUE #644: Add subgraph_create JSON-RPC method

### LINKED ISSUE #645: Update subgraph_deploy command to incorporate subgraph versions and deployments

### LINKED ISSUE #658: Add support for subgraph names with more than one path segment (e.g. `foo/bar`)
We'll want this in order to namespace subgraphs by e.g. account / organization / user in hosted environments.

There are two parts to this:

1. Allowing subgraph names with slashes.
2. Updating the request handlers in the GraphQL HTTP and WebSocket servers to support multi-segment subgraph names.

timmclean:
Track a history of what versions have been deployed under a subgraph name.

This PR is primarily about four entity types:
- Subgraph
- SubgraphVersion
- SubgraphState: replaces the `subgraphs` table and the previous `Subgraph` entity type
- SubgraphDeployment: replaces the `subgraph_deployments` table

Using entities for persisting system state has some nice advantages. It allows graph-node users to query the state externally using GraphQL. It also allows us to reuse entity subscriptions instead of having to reimplement similar logic in each case with postgres pub/sub. This PR pushes forward on that idea, moving some key components into entities.

Change list:
- SubgraphProviderWithNames is replaced by SubgraphRegistrar, which is responsible for creating/updating/removing Subgraph, SubgraphVersion, and related entities.
- New JSON RPC method: `subgraph_create`, corresponds to `SubgraphRegistrar::create_subgraph`
- JSON RPC methods `subgraph_deploy` and `subgraph_remove` correspond to `SubgraphRegistrar::create_subgraph_version` and `SubgraphRegistrar::remove_subgraph`, respectively
- SubgraphProvider is replaced by SubgraphDeploymentProvider, which is responsible for starting/stopping a subgraph deployment (i.e. block stream, etc) on this graph-node instance.
- SubgraphRegistrar watches for changes to SubgraphDeployment entities and calls the appropriate methods on SubgraphDeploymentProvider. (Previously: this was implemented with a custom SQL `LISTEN` operation; now, this uses the same codepath as entity subscriptions)
- `graph/src/data/subgraph/schema.rs` provides statically typed methods for creating and updating entities in the subgraph of subgraphs. There is only minimal static typing for reading entity data, though. Should we add that, or is that overkill?
- `MockStore` now implements Set/Remove/AbortUnless entity operations, and some basic querying functionality. This was needed to test the functionality that is now based on entities instead of database tables.
- To help with creating EntityQuery instances by hand, I added some shorthand syntax for creating an EntityQuery. E.g. `SubgraphStateEntity::query().filter( ... )`.
- SubgraphDeploymentStore's methods are mostly removed now in favour of working directly with `SubgraphDeploymentEntity`s.
- `SubgraphDeploymentStore::is_deployed` is now called `is_queryable`. This reflects the fact that subgraph versions that are not current are in what I've been thinking of as a "paused" state -- they have a SubgraphStateEntity and entity data that can be queried, but do not have a deployment or a block stream. These subgraph versions are still queryable (by subgraph hash, just not by subgraph name).
- Fixes bug in queries with `EntityFilter::Equal(foo, Value::Null)` and adds unit tests for those cases
- Adds some migrations to support the new changes. Unfortunately, they are not sufficient to avoid needing `dropdb`. Should I implement proper migrations here? It's non-trivial.

There is a new EntityOperation type called `AbortUnless`.  This makes it possible to detect race conditions and abort/retry an operation when another operation has made conflicting changes in parallel. Some patterns for working with entities safely:
- Whenever an entity change is conditional on the result of an entity query, AbortUnless must be used to verify that the result of that query has not changed. **Example**: Suppose that the store is queried and no Subgraph entity exists yet with the name "foo". An AbortUnless must be added to confirm that there is still no Subgraph named "foo" when the transaction actually goes through.
- Whenever some entity changes are created based on information from a store query, AbortUnless must be used to verify that the result of that store query has not changed in any way that would alter what the entity changes should have been.
- Pitfall: When building up a list of EntityOperations, the store will not reflect those changes until the operations are actually sent to the store. This may seem obvious, but it can be confusing at times. In a SQL transaction, changes made earlier in the transaction are visible in queries made later in the transaction, whereas here, that is not the case.

As a rule of thumb, generally each `store.find` or `store.get` leading up to a `store.apply_entity_operations` should have a corresponding `AbortUnless`.

It is generally fine to retry a transaction if it aborted. Note though that a retry must include rerunning the store queries to get the latest data -- rerunning the same set of entity operations will abort again for the exact same reason. I didn't prioritize implementing retry logic here because I suspect aborts will be infrequent, and all situations where they can happen at the moment can be retried by the user.

Resolves #643, resolves #644, and resolves #645 
Resolves #658


timmclean:
Copying over a comment on the subgraph entities schema from https://github.com/graphprotocol/graph-node/issues/643#issuecomment-451520383

---

```
type Subgraph @entity {
    id: ID! 
    name: String!
    versions: [SubgraphVersion!]! @derivedFrom(field: "subgraph")
    currentVersion: SubgraphVersion
    createdAt: BigInt!
}

type SubgraphVersion @entity {
    id: ID! 
    subgraph: Subgraph!
    state: SubgraphState!
    createdAt: BigInt!
}

type SubgraphState @entity {
    id: ID! # Subgraph IPFS hash
    manifest: SubgraphManifest!
    failed: Boolean!
    deployments: [SubgraphDeployment!]! @derivedFrom(field: "id")
    latestEthereumBlockHash: String!
    latestEthereumBlockNumber: BigInt!
    totalEthereumBlocksCount: BigInt!
    entityCount: BigInt! # Computed field, not stored.
}

type SubgraphDeployment @entity {
    id: ID! # Subgraph IPFS hash
    nodeId: String!
    synced: Boolean!
    cost: BigInt!
}
```

There are a few deviations from what we had discussed pre-holidays:
- To replace the `subgraphs` database table (which currently houses the subgraph block pointer), I added a `SubgraphState` entity. This entity is distinct from `SubgraphDeployment` in that it continues to exist even when the subgraph is not deployed (e.g. is not in a "current" subgraph version, or is in a paused subgraph). Conceivably these two entities could be combined however.
- I replaced the `status` field with 2 flags (`SubgraphState.failed` and `SubgraphDeployment.synced`) and the existence of an entity (`SubgraphDeployment`). A subgraph can be considered to be `PAUSED` if it does not have a deployment (in this sense, explicitly setting the status to `PAUSED` was redundant information). The `synced` flag is on `SubgraphDeployment` because the question "is this subgraph synced" does not make sense unless the subgraph is deployed.
- `latestEthereumBlockHash` was added in order to move the full block pointer into `SubgraphState`.

timmclean:
Thanks Leo!

`subgraph_deploy` is the same, but you do need to use `subgraph_create` first, is that what tripped you up? There is an error message for this on the client side when you run `graph deploy`, but graph-node side logging isn't so clear, I'll clean that up.

I like the idea of telling the user about the db update. Maybe we should add a version field to the database that we can check on startup (and only increment with backwards incompatible changes).

There's a bit on SubgraphState vs SubgraphDeployment in my previous comment (first bullet point). Basically the distinction is that a SubgraphState exists if the subgraph hash is queryable, while a SubgraphDeployment only exists if an index node has been assigned to work on this subgraph. There can be a SubgraphState without a SubgraphDeployment any time the subgraph hash is not the "current version" of any subgraph, since we don't actively process blocks for non-current subgraph versions. The inverse is different though: any subgraph hash with a SubgraphDeployment will always have a SubgraphState.

I had a nullable `deployment` field before in SubgraphState, but it doesn't look like we support `@derivedFrom` for nullable fields?

leodasvacas:
So this also needs changes to graph-cli so that we call create before deploy. Or we could create the name if a deploy is done to a non-existing name.

> I had a nullable deployment field before in SubgraphState, but it doesn't look like we support @derivedFrom for nullable fields?

My point was that `deployments` shouldn't be an array at all, since there can only be one deployment.

timmclean:
I think there was a product decision that we would add a `graph create` command to graph-cli, but I can't swear to it. Agreed, we need some kind of change there.
(for testing try `curl http://localhost:8020 -H 'content-type:application/json' -d '{"jsonrpc":"2.0","id":"1","method":"subgraph_create","params":{"name":"dharma"}}'`)

> My point was that deployments shouldn't be an array at all, since there can only be one deployment.

Right, I agree. I don't think `@derivedFrom` works with non-arrays though

graph/src/data/subgraph/schema.rs
```diff
@@ -362,3 +609,16 @@ fn set_entity_operation(
...
+
+pub fn generate_entity_id() -> String {
+    // Fast crypto RNG from operating system
+    let mut rng = OsRng::new().unwrap();
+
+    // 128 random bits
+    let id_bytes: [u8; 16] = rng.gen();
+
+    // 32 hex chars
+    // Comparable to uuidv4, but without the hyphens,
```
Jannis: Is the likelihood (or the lack of it) comparable to uuidv4?

server/http/src/service.rs
```diff
@@ -115,31 +123,37 @@ where
...
         let service = self.clone();
 
         Box::new(
-            future::result(SubgraphDeploymentName::new(name.clone()))
-                .map_err(move |()| {
-                    GraphQLServerError::ClientError(format!("invalid subgraph name {:?}", name))
+            SubgraphName::new(subgraph_name)
+                .map_err(|()| {
+                    GraphQLServerError::ClientError(format!(
+                        "invalid subgraph name {:?}",
```
Jannis: Capital first letter

server/http/src/service.rs
```diff
@@ -115,31 +123,37 @@ where
...
-                            })
-                        })
-                        .and_then(move |(subgraph_id, _node_id)| {
-                            service.handle_graphql_query(subgraph_id, request.into_body())
+                .and_then(|subgraph_name| {
+                    self.store
+                        .resolve_subgraph_name_to_id(subgraph_name)
+                        .map_err(|e| {
+                            GraphQLServerError::InternalError(format!(
+                                "error resolving subgraph name: {}",
```
Jannis: Capital first letter

server/http/src/service.rs
```diff
@@ -115,31 +123,37 @@ where
...
+                                e
+                            ))
                         })
+                })
+                .into_future()
+                .and_then(|subgraph_id_opt| {
+                    subgraph_id_opt.ok_or(GraphQLServerError::ClientError(
+                        "subgraph name not found".to_owned(),
```
Jannis: Capital first letter

server/http/src/service.rs
```diff
@@ -176,9 +190,10 @@ where
         let schema = match self.store.subgraph_schema(id) {
             Ok(schema) => schema,
             Err(e) => {
+                println!("handle_graphql_query: schema error: {}", e);
```
Jannis: Remove the println or turn it into a log.

store/postgres/migrations/2018-07-10-062730_create_revert_procedures/up.sql
```diff
@@ -252,24 +252,7 @@ $$
 DECLARE
     event_row RECORD;
     entity_row RECORD;
-    subgraph_ptr_hash VARCHAR;
```
Jannis: Can't we easily update this revert procedure by removing it and adding the new version in a new migration?

store/postgres/src/store.rs
```diff
@@ -323,126 +321,264 @@ impl Store {
...
+                    key.subgraph_id,
+                    key.entity_type,
+                    key.entity_id,
                     e
                 )
             })?;
 
         // Either add or update the entity in Postgres
-        insert_into(entities)
+        insert_into(entities::table)
```
Jannis: Isn't the `entities` DSL available any more?

Jannis (changes requested):
Wow, ok. First of all: YES! Second of all: I've tried to review everything as thoroughly as possible. One thing I need to ponder on for a moment longer is the subgraph state / subgraph deployments split. I'll get back to you on that one. 😉

Jannis:
After some more thinking, I think the only thing I don't like about the state / deployment split is the names. State feels more "local" compared to deployment.

Jannis:
@timmclean How about we call `SubgraphState` -> `SubgraphDeployment` and `SubgraphDeployment` -> `SubgraphDeploymentInstance` (or `SubgraphDeploymentAllocation` something similar)?

core/src/subgraph/provider.rs
```diff
@@ -71,36 +69,17 @@ where
...
+                        ));
                     }
 
-                    // Place subgraph info into store
-                    let created_at = SystemTime::now()
-                        .duration_since(UNIX_EPOCH)
-                        .unwrap()
-                        .as_secs();
-                    let entity_ops =
-                        SubgraphEntity::new(&subgraph, SubgraphStatus::Syncing, 0, 0, created_at)
```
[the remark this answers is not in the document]
timmclean: I think you're reading removed code :)

core/src/subgraph/registrar.rs
```diff
@@ -0,0 +1,776 @@
...
+        name.to_string().into(),
+    )))?;
+    let subgraph_entity = match subgraph_entities.len() {
+        0 => Err(SubgraphRegistrarError::NameNotFound(name.to_string())),
+        1 => {
+            let mut subgraph_entities = subgraph_entities;
+            Ok(subgraph_entities.pop().unwrap())
+        }
+        _ => panic!(
+            "multiple subgraph entities with name {:?}",
```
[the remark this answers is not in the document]
timmclean: Ah, this is just a trick to get the quotes but also to get the escaping of special characters. I've found it helpful for debugging to have special characters escaped in a predictable way. Although, I guess that doesn't matter here because of subgraph name validation...

timmclean:
@Jannis It's long, but I like `SubgraphDeployment`/`SubgraphDeploymentAllocation`! With that name, I can change `is_queryable` back to `is_deployed`, since deployed and queryable are equivalent. I've been using the term "paused subgraph" for when it is queryable but not actively indexing; a new term for that could be "unallocated deployment", which seems more clear to me. Also, maybe this means that `SubgraphId` could be renamed `SubgraphDeploymentId`, which would resolve the awkward confusion between IPFS hashes and the subgraph entity IDs.

I just thought of `SubgraphDeploymentAssignment` as an alternative, do you like that better or worse than `Allocation`? I'm good with either :)

Jannis:
I like `SubgraphDeploymentId`. No preference regarding assignment / allocation / instance. Assignment is maybe a _little_ more specific.

store/postgres/src/store.rs
timmclean: Yeah, I just got frustrated with having column names conflict with local variable names -- they're in the same namespace when importing the dsl. I can revert if you like

store/postgres/migrations/2018-07-10-062730_create_revert_procedures/up.sql
timmclean: Yup! Should I write a proper migration to avoid needing to dropdb? (This part is pretty easy... the hard part is translating data from the `subgraphs` and `subgraph_deployments` tables into subgraph entities to keep the subgraph block pointer intact, etc)

graph/src/data/subgraph/schema.rs
timmclean: Yup, likelihood of collisions is low enough to be ignored. With UUID v4, 122 bits are random, so the first duplicate ID would be expected after about 2^61 IDs have been generated. Here, we don't waste 6 bits on a version identifier, so 32 hex chars = 16 bytes = 128 bits are fully random, and the first duplicate ID is expected after about 2^64 IDs, slightly better than UUID v4.

store/postgres/src/store.rs
Jannis: Nah, it's fine.

core/src/subgraph/registrar.rs
```diff
@@ -0,0 +1,776 @@
...
+    store: Arc<impl Store>,
+    name: SubgraphName,
+) -> Result<(), SubgraphRegistrarError> {
+    debug!(logger, "Removing subgraph: {:?}", name.to_string());
+
+    let mut ops = vec![];
+
+    // Find the subgraph entity
+    let matching_subgraph_entities = store
+        .find(SubgraphEntity::query().filter(EntityFilter::Equal(
```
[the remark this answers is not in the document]
timmclean: Makes sense. I just pushed a commit to add `store.find_one`, which sets a range and returns `Option<Entity>` instead of `Vec<Entity>`, and refactored these `find` calls to use `find_one` instead

graph/src/data/store/mod.rs
```diff
@@ -231,6 +188,66 @@ impl Value {
             }
         })
     }
+
+    pub fn as_string(self) -> Result<String, ()> {
```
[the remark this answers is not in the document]
timmclean: I like it!

timmclean:
Latest update on this:

All the entity renaming ("deployments" and "assignments") is done. I removed the `deployments` field in the schema that Leo was asking about, so we don't have to worry about deriving a nullable field. I added a Store method `find_one` to make it easy to get a single entity when only one is expected. Also, I changed `Value::as_string` and friends to return an Option type. Jannis has a PR on graph-cli for `graph create`.

Currently outstanding:
- I have a few code comments left from Jannis that I'll get to tomorrow
- Migrations: the decision was to not make a proper migration for this PR. Leo suggested adding a friendly startup message telling the user that they need to dropdb (and why). Any suggestions on how to implement this?

Jannis:
About the migrations: If we can make the migrations fail when the db needs to be dropped, then we could use the error handling logic in https://github.com/graphprotocol/graph-node/blob/master/store/postgres/src/store.rs#L39-L56 to print that suggestion.

graph/src/components/store.rs
```diff
@@ -173,35 +231,21 @@ impl EntityOperation {
...
-                    entity.merge(data.clone());
-                    entity
-                },
-            },
-        }
-    }
+        // Only continue if all operations are Set/Remove.
+        ops.iter().for_each(|op| match op {
+            Set { .. } | Remove { .. } => {}
+            AbortUnless { .. } => panic!("Cannot apply {:?} to an Entity", op),
```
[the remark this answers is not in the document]
timmclean: Done!

graph/src/data/subgraph/mod.rs
```diff
@@ -77,50 +86,106 @@ impl<'de> de::Deserialize<'de> for SubgraphId {
...
+
+impl SubgraphName {
+    pub fn new(s: impl Into<String>) -> Result<Self, ()> {
+        let s = s.into();
+
+        // Note: these validation rules must be kept consistent with the validation rules
+        // implemented in any other components that rely on subgraph names.
+
+        // Enforce length limits
+        if s.len() < 1 || s.len() > 255 {
```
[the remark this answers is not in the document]
timmclean: I added these constraints and some test cases in the latest commit

timmclean:
@Jannis @leodasvacas All ready for ya :)

Jannis (changes requested):
I noticed there is no link from `SubgraphDeployment`s to `SubgraphDeploymentAssignments` in the GraphQL schema. Could we add that?

Jannis (approved):

PR #673, opened 2019-01-09
Return subgraph ID from `subgraph_create`

Jannis:
This is so users of `subgraph_create` can then use the new ID (e.g. store them somewhere).

### DIFF
--- a/server/json-rpc/src/lib.rs
+++ b/server/json-rpc/src/lib.rs
@@ -67,7 +67,7 @@ where
                         json_rpc_error(JSON_RPC_CREATE_ERROR, e.to_string())
                     }
                 })
-                .map(move |_| Value::Null),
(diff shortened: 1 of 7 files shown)


server/json-rpc/src/lib.rs
```diff
@@ -67,7 +67,7 @@ where
                         json_rpc_error(JSON_RPC_CREATE_ERROR, e.to_string())
                     }
                 })
-                .map(move |_| Value::Null),
+                .map(move |id| Value::String(id)),
```
timmclean: How about returning `{"id": ...}` here so we can add more properties without breaking compatibility? In case external graph-node users decide to use it

Jannis:
@timmclean Updated to return `{"id": ...}`.

PR #675, opened 2019-01-11
Add support for variables in subscription queries

Jannis:
This adds missing support for variables in subscriptions, allowing subscriptions such as the following to be executed:
```graphql
subscription foo($id: ID!) {
  subgraph(id: $id) {
    id
    versions { id }
  }
}
```

### DIFF
--- a/server/websocket/src/connection.rs
+++ b/server/websocket/src/connection.rs
@@ -240,14 +240,36 @@ where
                         }
                     };
 
-                    // TODO Parse query variables and operation name
+                    // Parse the query variables, if present
+                    let variables = match payload.variables {
+                        None | Some(serde_json::Value::Null) => None,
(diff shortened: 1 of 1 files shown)


server/websocket/src/connection.rs
```diff
@@ -240,14 +240,36 @@ where
...
+                        }
+                    };
 
                     // Construct a subscription
                     let subscription = Subscription {
                         query: Query {
                             schema: schema.clone(),
                             document: query,
-                            variables: None,
+                            variables: variables,
```
timmclean: can this be shortened?

timmclean:
one style nit, otherwise looks good

server/websocket/src/connection.rs
Jannis: Done

timmclean (approved):

PR #677, opened 2019-01-11
runtime: Inject timeout checks in mappings, log execution time

### LINKED ISSUE #665: Watchdog to kill WASM handlers that take too long to run
Part of epic #132. 

If an event handler doesn't finish in a reasonable amount of time we should detect that and kill the subgraph. We should have a dedicated tokio task tracking the timers. The timeout should be configurable through an env var.

Note that the time taken for a mapping to run is sensitive to a bunch of external conditions, for example if an indexing node gets generally slow for some reason that may cause the watchdog to kill all running subgraphs, so the timeout should not be too tight.

### LINKED ISSUE #668: Measure and log handler execution time

leoyvens:
Use `pwasm_utils::inject_gas_counter` to inject calls to a callback that checks if the handler has timed out based on the maximum processing time given by an env var. If the variable is not defined mappings never timeout. Trap in case of a timeout. The mechanism for injecting the callbacks is exactly the same parity uses for metering, so it should be easy to use it for our own metering in the future.

The heap and the exports became mutually dependent, so the structures are unified with `WasmiModule`.

` runtime/wasm/src/asc_abi/test.rs` is removed, but no tests were harmed in the making of this PR, they've been moved to `runtime/wasm/src/module/test/abi.rs`. It's just that those were tested with a mock `TestModule`, now everything is tested through `WasmiModule`.

Resolves #665, resolves #668.


runtime/wasm/src/host_exports.rs
```diff
@@ -404,6 +407,22 @@ where
...
     }
+
+    pub(crate) fn check_timeout(
+        &self,
+        start_time: Instant,
+    ) -> Result<(), HostExportError<impl ExportError>> {
+        let event_handler_timeout = std::env::var(TIMEOUT_ENV_VAR)
+            .ok()
+            .and_then(|s| u64::from_str(&s).ok())
+            .map(Duration::from_secs);
```
Jannis: Would it make sense to parse the duration once instead of doing it for every gas metering call? I'm not sure how expensive it is, but it could slow execution down.

Jannis (approved):
Approved, with one potential improvement commented on.

runtime/wasm/src/host_exports.rs
leodasvacas: I don't think reading and parsing the env var is extraordinarily expensive. Improving performance here is still a good thought, but we'd be trading off the ability to change this variable during the execution of the mapping.

Jannis: It may be a premature optimization. Let's roll with the current approach.

leodasvacas: @lutter Sorry for missing this. WASM modules are sandboxed and don't have access to env vars, and generally don't have access to the host environment memory or syscalls.

PR #681, opened 2019-01-14
runtime, graph: Allow panics in mapping runtime threads

### LINKED ISSUE #663: Recover from panics in runtime threads
Part of epic #132.

A panic anywhere in the node usually means we want to crash, but there are many ways a mapping could do non-conformant things and panic the WASM runtime thread. so instead of trying to make them error on all the places they might panic which would complicate the code and not guaranteed to be 100%, we'll start with catching any panic in a runtime thread and recovering from it.

leoyvens:
Threads are already a point where panics are naturaly catched. But by design we exit the process on
panics, so give the mapping threads special names which the handler checks for and then doesn't exit the process.

The subgraph will go into failed status as soon as it needs calls the mapping and finds that the channel was dropped.

Resolves #663.

### DIFF
--- a/graph/src/util/log.rs
+++ b/graph/src/util/log.rs
@@ -8,6 +8,8 @@ use std::sync::Mutex;
 use std::time::Duration;
 use std::{env, panic, process, thread};
 
+pub const MAPPING_THREAD_PREFIX: &str = "mapping-thread";
+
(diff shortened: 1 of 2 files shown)


Jannis (approved):
Looks good!

PR #686, opened 2019-01-14
Add environment variable to disable block ingestor

### LINKED ISSUE #682: Add environment variable to disable block ingestion

timmclean:
Resolves #682

### DIFF
--- a/node/src/main.rs
+++ b/node/src/main.rs
@@ -399,18 +399,20 @@ fn async_main() -> impl Future<Item = (), Error = ()> + Send + 'static {
     let mut subscription_server =
         GraphQLSubscriptionServer::new(&logger, graphql_runner.clone(), store.clone());
 
-    // Create Ethereum block ingestor
-    let block_ingestor = graph_datasource_ethereum::BlockIngestor::new(
-        store.clone(),
-        transport.clone(),
-        50, // ancestor count, which we could make configuable
-        logger.clone(),
-        block_polling_interval,
-    )
-    .expect("failed to create Ethereum block ingestor");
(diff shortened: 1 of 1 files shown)


node/src/main.rs
```diff
@@ -399,18 +399,20 @@ fn async_main() -> impl Future<Item = (), Error = ()> + Send + 'static {
...
-    // Create Ethereum block ingestor
-    let block_ingestor = graph_datasource_ethereum::BlockIngestor::new(
-        store.clone(),
-        transport.clone(),
-        50, // ancestor count, which we could make configuable
-        logger.clone(),
-        block_polling_interval,
-    )
-    .expect("failed to create Ethereum block ingestor");
+    if env::var_os("DISABLE_BLOCK_INGESTOR").unwrap_or("".into()) != "true" {
```
leodasvacas: We can just test if the var is present at all, what do you think?

leoyvens (commented):
I have a suggestion. Nice to see that the ingestor is 100% self contained.

node/src/main.rs
timmclean: Ah, I did it this way in case someone tries to do `DISABLE_BLOCK_INGESTOR=false` actually

leodasvacas: Makes sense, though there's also the case of trying `DISABLE_BLOCK_INGESTOR=1` or `DISABLE_BLOCK_INGESTOR=yes` or something. But either way works.

leoyvens (approved):

node/src/main.rs
timmclean: Yeah, that's true... :smile:

PR #689, opened 2019-01-16
Allow manual creation of pg_trgm extension in Postgres database

fordN:
The Graph node may be connecting to Postgres with a user that has limited permissions, so it is unable to create extensions.
This small PR updates the `create_attribute_indexing_procedure` migration to allow a superuser to add the extension manually before starting the node and running the migration.


Jannis (approved):

Jannis:
@fordN Looks good. I'll merge right away, the Travis hiccup is normal.

Issue #678, opened 2019-01-11
Issues with GraphQL List field contains filter


fordN:
- [x] Contains filters are not showing up for List field types in the GraphiQL docs.  
- [x] When added as a filter to a query, a List contains filter triggers an error message. 
Example: `Argument "where" has invalid value {owners_contains: ["QUERY_VALUE"]}. In field "owners.contains": Unknown field. 
- [x] Using a "ListAttribute_contains" filter in a query leads to an error: `Failed to get entities from store: operator does not exist: text @> text`

heri16:
@fordN "_contains" is available for List fields, isn't it weird why "_in" is not available for List fields?

PR #692, opened 2019-01-16
Fix issues with filters for List field types

### LINKED ISSUE #684: Contains filters are not showing up for List field types in the GraphiQL docs

fordN:
Resolves #684 and #678 

This PR fixes the following issues with filters for List field types:
  -  Filters were not showing up for List field types in the GraphiQL docs. 
  -  Generated PostgreSQL filters were incorrect and causing errors because a jsonb operator was being used with text data.

### DIFF
--- a/graphql/src/schema/api.rs
+++ b/graphql/src/schema/api.rs
@@ -483,7 +483,7 @@ mod tests {
 
     #[test]
     fn api_schema_contains_object_type_filter_enum() {
-        let input_schema = parse_schema("type User { id: ID!, name: String!}")
+        let input_schema = parse_schema("type User { id: ID!, name: String!, pets: [String!]}")
             .expect("Failed to parse input schema");
         let schema = api_schema(&input_schema).expect("Failed to derived API schema");
 
@@ -524,7 +524,11 @@ mod tests {
                 "name_starts_with",
                 "name_not_starts_with",
                 "name_ends_with",
-                "name_not_ends_with"
+                "name_not_ends_with",
+                "pets",
+                "pets_not",
+                "pets_contains",
+                "pets_not_contains"
             ]
             .iter()
             .map(|name| name.to_string())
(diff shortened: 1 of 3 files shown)


graphql/src/schema/api.rs
```diff
@@ -191,7 +188,9 @@ fn field_filter_input_values(
                 _ => vec![],
             })
         }
-        Type::ListType(ref t) => Ok(field_list_filter_input_values(schema, field, t)),
+        Type::ListType(ref t) => {
+            Ok(field_list_filter_input_values(schema, field, t).map_or(vec![], |values| values))
```
Jannis: `map_or` -> `unwrap_or`. Then you don't need the extra closure.

graphql/src/schema/api.rs
```diff
@@ -266,10 +265,22 @@ fn field_enum_filter_input_values(
...
 fn field_list_filter_input_values(
     _schema: &Document,
-    _field: &Field,
-    _field_type: &Type,
-) -> Vec<InputValue> {
-    vec![]
+    field: &Field,
+    field_type: &Type,
+) -> Option<Vec<InputValue>> {
+    let _value_type = ast::get_field_value_type(field_type).ok()?;
```
Jannis: Is the `let _value_type =` assignment necessary at all (it may be)?

graphql/src/schema/api.rs
```diff
@@ -266,10 +265,22 @@ fn field_enum_filter_input_values(
...
+    Some(
+        vec!["", "not", "contains", "not_contains"]
+            .into_iter()
+            .map(|filter_type| {
+                input_value(
```
Jannis: This isn't quite right. For `contains` and `not_contains`, we'd want the input value to be a `T`, not a `[T]`.

This will affect the SQL for the two filter types as well.

graphql/src/schema/api.rs
fordN: I'm using `get_field_value_type()` here to check that `field_type` is a ValueType variant (supports filters.)

graphql/src/schema/api.rs
Jannis: I changed my mind on this, this is fine for the moment.

heri16:
@fordN "_contains" is available for List fields, isn't it weird why "_in" is not available for List fields?

Issue #685, opened 2019-01-14
Only update subscriptions once per block


Jannis:

timmclean:
I wonder if the subscriptions updating could be implemented in [apply_entity_operations](https://github.com/graphprotocol/graph-node/blob/master/store/postgres/src/store.rs#L591) so that it doesn't need special casing for the subgraph of subgraphs?

Jannis:
True, that could save the overhead of Postgres notifications for all entity changes as well.

lutter:
What I've been shooting for is to add a `block_ticks` method to `Store` that produces a `BlockTickStream`. The GraphQL subscriptions would then use that instead of `subscribe` to trigger updates. Internally, `handle_entity_changes` would receive both kinds of events (entity changes and block ticks) and feed either kind of subscriber with the appropriate data.

That sidesteps the issue of special-casing metadata updates, but does have the disadvantage of generating entity changes that might not be needed, though I have a hard time seeing how we could turn off entity change notifications without pushing some knowledge about current subscriptions into the DB.

The reason why I am putting both entity changes and block ticks into the same notification channel is ordering: I will generate the block tick in `transact_block_operations` (via `pg_notify`) after all the entity change events have been generated. That way, a block tick also means "you've already seen all relevant entity changes" in case we need to deliver something like that with a block tick. For now, `BlockTick` will just be a very simple struct, maybe with the block id/number in it, but mostly for logging purposes.

Jannis:
Just saw this, thanks for the great write-up @lutter!

Trying do to this in `apply_entity_operations` could be problematic for any setup where querying and indexing is handled by different nodes. So I think going through the database for this is the right thing to do.

lutter:
Rebased onto latest master

PR #693, opened 2019-01-17
Update subscriptions at most once per block

lutter:
This addresses Issue #685. GraphQL subscriptions now receive `BlockTicks` rather than `EventChanges`, and therefore are updated at most once per block.

This PR passes `cargo test`, but I need to test some more before this could be merged. One difficulty in writing tests is that I am not sure how to check that a stream does _not_ send anything. Ideally, the tests could close the sending end of a stream and pull out everything that was sent on the stream, but I am not sure how to do that.

The PR tries to minimize how much it mucks with existing stuff. There's a simpler, but somewhat more invasive way to achieve the same result: we could bundle up all the changes that were made in a block and send them as part of the `BlockTick` in `apply_entity_operations` rather than continuing to send them as `EventChange` events and piecing the `BlockTick` together in `handle_store_events`. That would require changes to the triggers on `entities` to stop them from sending `EventChange` for anything but the fake `subgraphs` subgraph and remove the possibility to listen to individual changes for 'real' subgraphs. I didn't go that route since it seemed a little too invasive, but it would greatly simplify `handle_store_events`.

On a more pedestrian level, the way `subscriptions` and `dirty_blocks` interact in the Postgres `Store` right now isn't ideal, since removing a subscription requires clearing it out from `dirty_blocks`, too; otherwise, we could build up entries in `dirty_blocks` if subscriptions are removed while they are dirty but before the corresponding `BlockTick` is sent. This should be bundled up in a `Subscriptions` struct that encapsulates that logic.


store/postgres/tests/store.rs
```diff
@@ -1533,6 +1540,24 @@ fn entity_changes_are_fired_and_forwarded_to_subscriptions() {
             )
             .unwrap();
 
+        // FIXME: If we produce fewer than 2 block ticks, the test hangs
```
Jannis: @timmclean @leodasvacas Any clever ideas here?

store/postgres/tests/store.rs
```diff
@@ -92,7 +94,11 @@ where
     runtime
         .block_on(future::lazy(move || {
             // Set up Store
-            let logger = Logger::root(slog::Discard, o!());
+            let logger = match env::var_os("GRAPH_LOG") {
```
Jannis: I'm going to do a proper review later, but this is a great idea. I've often been missing logs from tests due to the `slog::Discard` everywhere.

store/postgres/tests/store.rs
leodasvacas: I don't think failure by timeout is so bad, we can put a fixed [timeout](https://docs.rs/tokio/0.1.14/tokio/prelude/trait.StreamExt.html#method.timeout) around it.

leodasvacas:
Travis is failing because you need to run `cargo fmt`.

lutter:
@leodasvacas thanks for the heads up; I just did `cargo fmt` on each patch. Let's hope that that will make Travis happy.

lutter:
`cargo fmt` still fails because Travis pulls in Rust 1.32.0, and that has different ideas on how to format master. PR #694 addresses that. Once that is merged, I'll rebase this branch.

lutter:
Rebased PR to pick up general `cargo fmt` fixes from PR #694

store/postgres/src/store.rs
```diff
@@ -23,18 +23,204 @@ use graph::{tokio, tokio::timer::Interval};
...
+impl Subscription {
+    fn send(
+        self,
+        e: StoreEvent,
+        id: &String,
+        dirty_blocks: &DirtyBlocks,
+    ) -> impl Future<Item = (), Error = ()> {
+        match (self, e) {
+            (Subscription::Changes { sender, .. }, StoreEvent::Change(change)) => {
+                EitherFuture::new_left(sender.send(change))
```
leodasvacas: The custom `EitherFuture` doesn't seem very necessary. I'd just box the futures and return a boxed trait object, if you'd rather avoid that another option is to use a nested built-in [Either](https://docs.rs/futures/0.1.11/futures/future/enum.Either.html) so that you have three variants with `Either<T, Either<U, V>>`. The `UnitFuture` is just `.map(|_| ()).map_err(|_| ())` which doesn't seem to bad to write out.

lutter:
Just updated the PR. Changes I made:

- used @leodasvacas's tip on using `Either` and got rid of my homegrown cruft around that
- updated the test to put a timeout on getting BlockTicks
- rebased on top of latest master

lutter:
I just realized a problem with all this when I looked at why subscriptions to the `subgraphs` subgraph get updates (they should, and they do, but I didn't understand why)

When we see a `BlockTick` in `handle_store_events`, we update all subscriptions for block ticks that had an entity change. Since a block is processed for each subgraph separately, we right now have cross-talk between subgraphs: a block tick for subgraph A might also cause a block tick for subgraph B (if that had interesting entity changes)

To address that, I need to amend these patches with two things:
* `BlockTick` needs to carry the `subgraph_id` for which it was generated (this might be easiest if we add `subgraph_id` to `EventSource::EthereumBlockPointer`, I did that already, and this seems simplest)
* special-case block tick subscriptions for the `subgraphs` subgraph so that entity changes cause updates to such subscriptions right away (I'll leave limiting frequency of updates for these for another PR)
* restrict block tick updates to subscriptions for the block tick's subgraph

store/postgres/src/store.rs
```diff
@@ -23,18 +24,134 @@ use graph::{tokio, tokio::timer::Interval};
...
+    },
+}
+
+/// Track which subscriptions have had changes to their blocks. The entries
+/// in the set are the id's of subscriptions for which we've seen an
+/// EntityChange matching their interesting entities. When the corresponding
+/// BlockTick comes by, we pass a BlockTick to the receiver and remove
+/// the id from the set
+#[derive(Clone)]
+struct DirtyBlocks {
```
Jannis: I find the name `DirtyBlocks` non-obvious because what it really represents are IDs of subscriptions that have had changes to their blocks. I know it's long but how about `SubscriptionsWithBlockChanges`?

store/postgres/src/store.rs
```diff
@@ -23,18 +24,134 @@ use graph::{tokio, tokio::timer::Interval};
...
+                StoreEvent::Tick(_) => false,
+            },
+            Subscription::Ticks { .. } => match event {
+                StoreEvent::Change(_) => false,
+                StoreEvent::Tick(_) => dirty_blocks.contains(id),
+            },
+        }
+    }
+
+    fn is_dirty(&self, event: &StoreEvent) -> bool {
```
Jannis: Reading this without comments,I'm asking myself : is`is_dirty` the right name here? I feel `event_matches_entities` would be a more intuitive choice.

store/postgres/src/store.rs
```diff
@@ -23,18 +24,134 @@ use graph::{tokio, tokio::timer::Interval};
...
+    }
+
+    fn poll_ready(&mut self) -> Poll<(), SendError<()>> {
+        match self {
+            Subscription::Changes { sender, .. } => sender.poll_ready(),
+            Subscription::Ticks { sender, .. } => sender.poll_ready(),
+        }
+    }
+
+    fn matches(&self, event: &StoreEvent, id: &String, dirty_blocks: &DirtyBlocks) -> bool {
```
Jannis: The semantics of this method are not entirely clear to me. Could you add a comment here, including a description of why this is needed in addition to the other matching method below?

store/postgres/src/store.rs
```diff
@@ -209,59 +328,74 @@ impl Store {
...
+                }
+                StoreEvent::Tick(tick) => {
+                    trace!(logger, "Received block tick"; "block" => tick.block.to_string());
+                }
+            }
+
+            // Record which BlockTick subscriptions have had 'interesting'
+            // changes and are therefore dirty
+            for (id, subscription) in subscriptions.read().unwrap().iter() {
+                if subscription.is_dirty(&event) {
```
Jannis: What confuses me a little here is that non-BlockTick subscriptions are also added to subscriptions with changes yet. So the block tick concept isn't separate. Another reason why is rename `is_dirty` and `DirtyBlocks`.

store/postgres/src/store.rs
```diff
@@ -209,59 +328,74 @@ impl Store {
...
                 .read()
                 .unwrap()
                 .iter()
-                .filter(|(_, subscription)| {
-                    subscription
-                        .entities
-                        .contains(&(change.subgraph_id.clone(), change.entity_type.clone()))
-                })
-                .map(|(id, subscription)| (id.clone(), subscription.sender.clone()))
+                .filter(|(id, subscription)| subscription.matches(&event, id, &dirty_blocks))
```
Jannis: A comment here would be nice too: What does it mean if a subscription matches here?

store/postgres/src/store.rs
```diff
@@ -209,59 +328,74 @@ impl Store {
...
             stream::iter_ok::<_, ()>(matches).for_each(move |(id, sender)| {
                 let logger = logger.clone();
                 let subscriptions = subscriptions.clone();
-                sender.send(change.clone()).then(move |result| {
-                    match result {
-                        Err(_send_error) => {
+                let dirty_blocks = dirty_blocks.clone();
+
+                trace!(logger, "Sending event {:?}", event);
+                sender
```
Jannis: Let's rename `sender` to `subscription`, so we're not tempted to confuse subscriptions and their internal senders.

store/postgres/src/store.rs
```diff
@@ -657,6 +790,47 @@ impl Store {
...
+        subscriptions.insert(id, subscription);
+    }
+
+    fn emit_block_tick(
+        &self,
+        conn: &PgConnection,
+        event_source: EventSource,
+    ) -> Result<(), StoreError> {
+        // Only emit block ticks for actual blocks
+        if event_source != EventSource::None {
```
Jannis: Since we might add more event sources later, a more robust check would be 
```
if let event_source = EventSource::EthereumBlock { ..} {
```

store/postgres/src/store.rs
```diff
@@ -657,6 +790,47 @@ impl Store {
...
+            trace!(self.logger, "Emit block tick"; "block" => event_source.to_string());
+            let v = serde_json::to_value(BlockTick {
+                block: event_source,
+            })?;
+            select(pg_notify("entity_changes", v.to_string())).execute(conn)?;
+        }
+        return Ok(());
```
Jannis: Drop the return and the semicolon. The last expression becomes the return value (similar to Ruby) unless you add a semicolon at the end of it.

store/postgres/src/store_events.rs
```diff
@@ -0,0 +1,65 @@
...
+pub struct StoreEventListener {
+    notification_listener: NotificationListener,
+}
+
+impl StoreEventListener {
+    pub fn new(postgres_url: String) -> Self {
+        StoreEventListener {
+            notification_listener: NotificationListener::new(
+                postgres_url,
+                SafeChannelName::i_promise_this_is_safe("entity_changes"),
```
Jannis: I was going to suggest changing the channel name but it's not worth it. The migration required to drop existing legs fun procedures and replace them just to call this "store_events" would be to complex.

store/postgres/src/store_events.rs
```diff
@@ -0,0 +1,65 @@
...
+    fn take_event_stream(&mut self) -> Option<Box<Stream<Item = StoreEvent, Error = ()> + Send>> {
+        self.notification_listener.take_event_stream().map(
+            |stream| -> Box<Stream<Item = _, Error = _> + Send> {
+                Box::new(stream.map(|notification| {
+                    // Parse notification as JSON
+                    let value: serde_json::Value = serde_json::from_str(&notification.payload)
+                        .expect("invalid JSON entity change received from database");
+                    // The JSON has a 'block' field for a BlockTick. If it doesn't,
+                    // it must be an EntityChange.
+                    // FIXME: We should really send an entire StoreEvent rather
```
Jannis: Agreed. This is something we can look into in a couple of weeks. Could you create a task for this that we put in "Ready" in Waffle?

store/postgres/src/store_events.rs
```diff
@@ -0,0 +1,65 @@
...
+}
+
+impl EventProducer<StoreEvent> for StoreEventListener {
+    fn take_event_stream(&mut self) -> Option<Box<Stream<Item = StoreEvent, Error = ()> + Send>> {
+        self.notification_listener.take_event_stream().map(
+            |stream| -> Box<Stream<Item = _, Error = _> + Send> {
+                Box::new(stream.map(|notification| {
+                    // Parse notification as JSON
+                    let value: serde_json::Value = serde_json::from_str(&notification.payload)
+                        .expect("invalid JSON entity change received from database");
```
Jannis: I guess this is now store event instead of entity change,could you rename it in the log  message?

store/postgres/src/store_events.rs
```diff
@@ -0,0 +1,65 @@
...
+                    // it must be an EntityChange.
+                    // FIXME: We should really send an entire StoreEvent rather
+                    // than its two components, but that requires changing the
+                    // triggers on the entities table.
+                    match value.get("block") {
+                        Some(_) => {
+                            let block: BlockTick = serde_json::from_value(value.clone())
+                                .unwrap_or_else(|_| {
+                                    panic!(
+                                        "invalid block change received from database: {:?}",
```
Jannis: Block change -> block tick

Jannis (changes requested):
Looking good! I have a few suggestions, let me know what you think about them.

Jannis:
@lutter I have thought about this design a bit more after reviewing it. I've come to the conclusion that I like the block ticks but I find the logic implemented in this PR convoluted. I feel it would be rather hard to extend this later (e.g. adding different kinds of filtering or transformations to store event streams).

I think what would be ideal is an API on top of `StoreEvent` streams that is very similar to regular iterator APIs. I think the following is desirable and very doable:

```rust
// A subscription that contains _all_ store events (entity changes, block ticks)
// are emitted by the db.
let subscription = store.subscribe();

// A subscription that only contains block ticks and entity changes for
particular entity types.
let subscription = store
    .subscribe()
    .filter_by_entity_types(vec![
        (SubgraphDeploymentId::new("some-id"), String::from("SomeEntity")
    ]);

// A subscription that groups events by blocks.
let subscription = store
    .subscribe()
    .events_by_block();

// A subscription that throttles emissions while subgraphs are syncing.
let subscription = store
    .subscribe()
    .events_by_block()
    .throttled_while_syncing(subgraphDeploymentStore.clone(), Duration::from_secs(5));

// The combination of all: a subscription that
// - filters by entity types,
// - groups events by block,
// - throttles while a subgraph is syncing
let subscription = store
    .subscribe()
    .filter_by_entity_types(vec![
        (SubgraphDeploymentId::new("some-id"), String::from("SomeEntity")
    ])
    .events_by_block()
    .throttled_while_syncing(subgraphDeploymentStore.clone(), Duration::from_secs(5));
```

Implementing this would require the following:

1. A combined `StoreEvent` enum that has `EntityChange` and `BlockFinished` variants.
2. Two stream types: `StoreEventStream` and `GroupedStoreEventStream`.
3. Two stream extension traits:
    - `StoreEventStreamExt` (including `filter_by_entity_types` and `events_by_block`),
    - `GroupedStoreEventStreamExt` (including `throttled_while_syncing`).
4. Implementations of the extension traits for `StoreEventStream` and `GroupedStoreEventStream`.

This design would give us the tools we need to add further transformation / grouping / filtering logic to optimize subscriptions over time. My feeling is the design implemented in the PR right now makes the filtering logic that it adds a little hard to follow/understand, plus it doesn't provide this extensibility.

What do you think?

Jannis:
To validate the design I'm proposing, I've written down potential implementations. While I haven't even tried to compile them, these could serve as an inspiration. I'd love some input from @leodasvacas on the approach, as there may be a more idiomatic and simpler way to do this by reusing more of the existing `Stream` and `StreamExt` traits in the `futures` crate.

Anyway, here's the code:
```rust
/// Store event
enum StoreEvent {
    EntityChange { change: EntityChange },
    BlockFinished { subgraph_id: SubgraphDeploymentId, block: BlockPointer }
}

impl SubgraphDeploymentStoreEvent {
    pub event_source(&self) -> EventSource {
        match self {
            EntityChange { change } => change.event_source.clone(),
            BlockFinished { block } => EventSource::EthereumBlock(block),
        }
    }

    pub subgraph_id(&self) -> &SubgraphDeploymentId {
        match self {
            EntityChange { change } => change.subgraph_id,
            BlockFinished { subgraph_id } => subgraph_id,
        }
    }
}

/// Base stream
pub type StoreEventStream = Box<Stream<Item = StoreEvent, Error = ()> + Send>;

/// Grouped stream
pub type GroupedStoreEventsStream = Box<Stream<Item = Vec<StoreEvent>, Error = ()> + Send>;

/// Trait for transforming store event streams
trait StoreEventStreamExt {
    /// Filters the store events such that events for unrelated entities are dropped.
    fn filter_by_entity_types(self, types: Vec<SubgraphEntityPair>) -> StoreEventStream;

    /// Groups events by Ethereum block and returns a stream of grouped events.
    fn events_by_block(self) -> GroupedStoreEventsStream;
}

/// Trait for transforming entity events by block streams
trait GroupedStoreEventsStreamExt {
    fn throttled_while_syncing(
        self,
        store: Arc<impl SubgraphDeploymentStore>,
        interval: Duration
    ) -> GroupedStoreEventsStream;
}

///
/// Implementations
///

impl SubgraphDeploymentStoreEventStreamExt for StoreEventStream {
    fn filter_by_entity_types(self, types: Vec<SubgraphEntityPair>) -> StoreEventStream {
        Box::new(self.0.filter(|event| {
            // This function call stands for any logic we might want to put here
            event_matches_entity_types(event, types)
        }))
    }

    pub fn events_by_block(self) -> GroupedStoreEventsStream {
        let events_by_subgraph = HashMap<SubgraphDeploymentId, Vec<StoreEvents>>::new();

        Box::new(self.0.filter_map(move |event| {
            match event {
                StoreEvent::EntityChange { change } => {
                    match change.event_source() {
                        // Only handle entity changes that are related to Ethereum blocks
                        EventSource::EthereumBlock(block) => {
                            // Queue the entity change so we can emit it once the block has
                            // finished processing
                            events_by_subgraph.entry(&change.subgraph_id)
                                .or_insert(vec![])
                                .push(event);
                        },

                        // Ignore entity changes not related to blocks
                        None,
                    }
                },
                StoreEvent::BlockFinished { subgraph_id, block } => {
                    // Queue this block finished event
                    events_by_subgraph.entry(&subgraph_id).or_insert(vec![]).push(event);

                    // Start a new event queue; this will return the events queued for the
                    // block we just finished
                    events_by_subgraph.insert(&subgraph_id, vec![])
                },
            }
        }))
    }
}

impl GroupedStoreEventsStreamExt for GroupedStoreEventsStream {
    pub fn throttled_while_syncing(
        self,
        store: Arc<impl SubgraphDeploymentStore>,
        interval: Duration
    ) -> GroupedStoreEventsStream {
        let last_emission_by_subgraph = HashMap<SubgraphDeploymentId, Instant>::new();

        Box::new(self.0.filter_map(|block_events| {
            let subgraph_id = block_events.iter().first().subgraph_id();

            // Look up the deployment by its ID (NOTE: the real API may be different)
            let deployment = store
                .find_deployment(&subgraph_id)
                .expect("received store events for a subgraph that isn't deployed");

            if (deployment.get("synced") == Value::Boolean(true)) {
                // The subgraph is already synced, don't throttle, forward events as they come in
                Some(block_events)
            } else {
                let last_emission = last_emission_by_subgraph
                    .get(&subgraph_id)
                    .unwrap_or(Instant::new(0));

                let now = Instant::now();

                if (now >= last_emission + interval) {
                    // Time to emit: reset the last emission time to now and forward the events
                    last_emission_by_subgraph.insert(subgraph_id.clone(), now);
                    Some(block_events)
                } else {
                    // Not yet time to emit, simply drop the received events here
                    None
                }
            }
        }))
    }
}
```

Jannis:
Once we e.g. identify subscriptions that are only for particular entities (e.g. one entity with a specific ID), I imagine it would be almost trivial to add another stream filter to the `StoreStreamExt` trait (e.g. `filter_by_entities(vec![(SubgraphDeploymentId::new("some-id"), entity_id)])`).

lutter:
I like that design a lot, and totally agree that the logic I implemented makes it much harder to follow what's happening, and is sort of a one-off implementation.

One thing I am wondering is if we should even distinguish between an EntityChange and a BlockChange event - the main difference between the two is that one carries a single change, while the other carries a vector of changes. Maybe we should simplify things by always dealing in a vector of changes, where the single change is just a vector with one change in it.

That also seems to mesh well with the more general implementation you've outlined above where a StoreEventStream with filters applied to it basically means "you'll get a bunch of changes when we deem that they are ready/you should react to them"

What do you think of a data structure like this for the events:

```
struct StoreEvent {
  source: EventSource,
  changes: Vec<EntityChange>
}
```

There's still a bit of a problem in that `source` doesn't have the subgraph id in it, even though it probably should. `EntityChange` would duplicate that - since everything is scoped to a subgraph, the subgraph id is a bit redundant then, but I am not sure it's worth trying to change that.

Jannis:
I like the idea of grouping events into vectors by default and only having a stream variant for that, not for single changes.

I do think we'll need the explicit block finished event. There are only two ways to identify whether a block has finished processing: an explicit finished event or receiving the first event for the next block. Sadly, the latter would require to wait for around 10-29s (Ethreum block time) and that's not acceptable. 

bdf3564e-1355-49d6-bb69-ee5407584ff3

Therefor we will need an explicit hint / event.

leodasvacas:
@Jannis First, no need to put all extension methods on the existing `StreamExt`, we can make new extension traits.

I'm a bit confused between the goals of this PR and the API you propose. For `events_by_block` to be useful, we'd need the assumption that before each `BlockFinished` all entity changes related to that subgraph id have been emitted as `EntityChange`s (right?), but the purpose of this PR is to stop emitting so many `EntityChange`s and only emit `BlockFinished` (right?), so they seem to be going in opposite directions. I'm probably getting something wrong.

Jannis:
@leodasvacas `StoreEventStreamExt` and `GroupedStoreEventsStreamExt` are new traits, no? 😉 

To clarify: What we are aiming for with this work is to only update subscriptions once per block. But we also need the entity change events in each block to decide whether they affect a subscription or not.

What I'm proposing is an extensible API to incrementally filter and transform entity changes:

- `filter_by_entity_types` takes entity types relevant for a description (e.g. `User` and `Team`) and only lets entity changes through that are for these entity types. And block finished events. This is essentially what we do right now.

- `events_by_block` collects entity change events until a block finished event arrives, then sends them out together. A GraphQL subscription can then consume this to only fire once per block.

- `throttled_while_syncing` controls how frequently groups of events are sent out using a time window. We can decide whether we want to drop events that come in or whether we want to collect them and flush them out together (probably better).

Now, we could say `filter_by_entity_types` + `only_block_finished` (would only allow block finished events through) would be enough to achieve the original goal of only updating subscriptions once per block but use the same "event to subscription matching" technique as before. We could do that as a first pass.

The `events_by_block` feature is useful because it would allow us to do more fine-grained checks down the road (like: "this subscription only cares for entities with ID `a` and `b`, so filter by those").

leodasvacas:
@Jannis Ah yes, I now get that we need the events to know whether the query for a subscription needs to be re-run or not. That API does look like a nice way to do this.

In terms of style, I might using a struct such as:
```rust
struct StoreEventStream<S>(S) where  S: Stream<Item = StoreEvent, Error = ()> + Send;
```
And then use inherent methods instead of extension traits.

lutter:
@Jannis @leodasvacas after working on this for a bit, I think I found a fly in this particular ointment: with this design, we'd have one stream for each subscription. `handle_store_events` would go from consuming DB events and sending them to subscriptions that it reasonably knows are interested in those events to sending them to all subscriptions where they'd get filtered along the way. That means we have to clone each event for each subscription (since `send` takes ownership of the event), even though most of these clones will be discarded pretty quickly. That will stress the allocator quite a bit, and the stress increases with the number of subscriptions. I can send `Arc<StoreEvent>` rather than `StoreEvent`, but we will still allocate quite a lot more than we do now. What do you think?

lutter:
Actually, after looking at the `Arc` implementation, this might not be that bad. It's essentially a pointer. I think I'll continue with implementing this based on `StoreEvent`, and we can then switch it to `Arc<StoreEvent>` when it all works.

Jannis:
Agreed!

leodasvacas:
@lutter I think we don't have a limit on how long an entity id might be, so cloning those around could become problematic, good thought. Arc seems like a good solution.

lutter:
The feedback for this PR is now addressed in a new PR #735

lutter:
> I like the idea of grouping events into vectors by default and only having a stream variant for that, not for single changes.
> 
> I do think we'll need the explicit block finished event.

@Jannis  I just realized I never replied to your comment above. My understanding from talking to Tim is that, for a given subgraph, we collect all the entity operations for a given block, and then pass all that to `apply_entity_operations`. We therefore don't need an explicit 'block finished' event, as we will send the event containing all the changes for a block right after the block is finished.

PR #694, opened 2019-01-17
Incorporate `cargo fmt` changes for Rust 1.32.0

lutter:
Rust 1.32.0 was released this morning, and `cargo fmt` has a few different
ideas on how things should be formatted. Since Travis pulls in the latest
stable automatically, we need to make sure that `cargo fmt` doesn't find
any changes on master.


Jannis (approved):
Looks good!

PR #695, opened 2019-01-18
Replace unsafe transmutes with safe equivalents

timmclean:
We can use these now that they've been stabilized in Rust v1.32.0

### DIFF
--- a/runtime/wasm/src/asc_abi/class.rs
+++ b/runtime/wasm/src/asc_abi/class.rs
@@ -67,7 +67,6 @@ impl<T> AscType for ArrayBuffer<T> {
     fn to_asc_bytes(&self) -> Vec<u8> {
         let mut asc_layout: Vec<u8> = Vec::new();
 
-        // This is just `self.byte_length.to_bytes()` which is unstable.
         let byte_length: [u8; 4] = self.byte_length.to_le_bytes();
         asc_layout.extend(&byte_length);
         asc_layout.extend(&self.padding);
--- a/runtime/wasm/src/asc_abi/class.rs
+++ b/runtime/wasm/src/asc_abi/class.rs
@@ -3,7 +3,7 @@ use ethabi;
 use graph::data::store;
 use graph::serde_json;
 use std::marker::PhantomData;
-use std::mem::{self, size_of, size_of_val};
+use std::mem::{size_of, size_of_val};
 
 ///! Rust types that have with a direct correspondence to an Asc class,
 ///! with their `AscType` implementations.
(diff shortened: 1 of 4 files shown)


runtime/wasm/src/asc_abi/class.rs
```diff
@@ -68,7 +68,7 @@ impl<T> AscType for ArrayBuffer<T> {
         let mut asc_layout: Vec<u8> = Vec::new();
 
         // This is just `self.byte_length.to_bytes()` which is unstable.
```
leodasvacas: Comment left behind.

timmclean: Ah thanks!

timmclean:
@leodasvacas Good point, do I need to tweak Travis for that?

leodasvacas:
@timmclean No because we haven't put 1.31 on travis, it's just an observation.

timmclean:
:+1:

runtime/wasm/src/asc_abi/class.rs
```diff
@@ -241,8 +238,7 @@ impl AscValue for EnumPayload {}
 
 impl From<EnumPayload> for i32 {
     fn from(payload: EnumPayload) -> i32 {
-        // This is just `i32::from_bytes` which is unstable.
-        unsafe { ::std::mem::transmute::<u32, i32>(payload.0 as u32) }
+        payload.0 as i32
```
leodasvacas: These ones had a reason to be the way they were. A cast has overflow semantics which under some build configurations, such as debug, may panic. Here we just want to transmute the bytes with no arithmetic consequence, so transmute is more correct than a cast. `from_bytes` is a safe way of doing that transmute that we can use now.

runtime/wasm/src/asc_abi/class.rs
```diff
@@ -260,8 +256,7 @@ impl From<EnumPayload> for bool {
 
 impl From<i32> for EnumPayload {
     fn from(x: i32) -> EnumPayload {
-        // This is just `i32::from_bytes` which is unstable.
-        EnumPayload(unsafe { ::std::mem::transmute::<i32, u32>(x) } as u64)
+        EnumPayload(x as u64)
```
leodasvacas: same

runtime/wasm/src/asc_abi/class.rs
```diff
@@ -279,8 +274,7 @@ impl From<bool> for EnumPayload {
 
 impl From<i64> for EnumPayload {
     fn from(x: i64) -> EnumPayload {
-        // This is just `u64::from_bytes` which is unstable.
-        EnumPayload(unsafe { ::std::mem::transmute::<i64, u64>(x) })
+        EnumPayload(x as u64)
```
leodasvacas: same

PR #696, opened 2019-01-18
Implement delayed version switching feature

### LINKED ISSUE #622: Send queries to older subgraph version until latest version has synced
Uploading a new version of a subgraph results in a potentially lengthy rebuild, during which the graph can't be queried. It'd be great if we could upload a new version of an existing subgraph, and have the node build the new subgraph in parallel with serving the old one, only switching over once the new graph is fully up to date.

timmclean:
Add a new environment variable, `EXPERIMENTAL_SUBGRAPH_VERSION_SWITCHING_MODE`, that supports two values: `instant` (default) and `synced`. Instant switching preserves the old behaviour: queries to a subgraph name start hitting the new version immediately after a `subgraph_deploy` request, even though that new version may not be synced yet, and the old version is immediately unassigned from the index node. Synced switching continues to direct queries to the existing synced version until the new version has synced. When a new version is deployed, it is set as the pending version for that subgraph. Both the pending version and the current version remain assigned (to an indexing node) until the pending version is synced, at which point the pending version is promoted and replaces the current version, and the old current version is unassigned.

Some details of synced switching:
- When a version is deployed to a subgraph that has not been deployed to yet, that version will immediately become current.
- If a subgraph already has a pending version, then deploying will replace that version with the new pending version and unassign the old pending version, without affecting the current version at all (it can continue to receive queries).

Additionally, this PR fixes some issues with how null values are handled and queried in the entity store. If an entity has a property set to `null`, that is now equivalent to the property being absent entirely.

Finally, this PR changes `subgraph_remove` to keep SubgraphDeployment entities even when they are no longer referenced by any subgraph version. This is better than the old behaviour where the deployment entity was kept, but the entity data was not cleaned up, which could lead to confusion if the deployment was redeployed.

Resolves #622


graph/src/components/store.rs
```diff
@@ -393,19 +415,255 @@ pub trait Store: Send + Sync + 'static {
 
     /// Counts the total number of entities in a subgraph.
     fn count_entities(&self, subgraph: SubgraphDeploymentId) -> Result<u64, Error>;
-}
 
-pub trait SubgraphDeploymentStore: Send + Sync + 'static {
     fn resolve_subgraph_name_to_id(
```
timmclean: Note: this method is not new and has not actually changed -- I just moved the body to here from the diesel/postgres store

graph/src/components/store.rs
```diff
@@ -393,19 +415,255 @@ pub trait Store: Send + Sync + 'static {
...
+        }));
+
+        ops
+    }
 
     /// Check if the store is accepting queries for the specified subgraph.
     /// May return true even if the specified subgraph is not currently assigned to an indexing
     /// node, as the store will still accept queries.
-    fn is_deployed(&self, id: &SubgraphDeploymentId) -> Result<bool, Error>;
+    fn is_deployed(&self, id: &SubgraphDeploymentId) -> Result<bool, Error> {
```
timmclean: Note: this method is not new and has not actually changed -- I just moved the body to here from the diesel/postgres store

graph/src/components/store.rs
```diff
@@ -393,19 +415,255 @@ pub trait Store: Send + Sync + 'static {
...
+        SubgraphDeploymentId::new(subgraph_id_str)
+            .map_err(|()| {
+                format_err!("SubgraphVersion entity has invalid subgraph ID in `deployment`")
+            })
+            .map(Some)
+    }
+
+    /// Read all version entities pointing to the specified deployment IDs and determine whether
+    /// they are current or pending in order to produce `SubgraphVersionSummary`s.
+    fn read_subgraph_version_summaries(
```
timmclean: I found the previous assignment creation/deletion logic hard to follow, so I refactored to create this method and the following method (`reconcile_assignments`). This method finds what versions currently exist. `reconcile_assignments` takes "before" and "after" lists of what versions existed vs what versions will exist, diffs them, and handles creating/removing assignments.

These two methods are used anywhere assignments are created and/or removed, which is currently three places:

- `create_subgraph_version` (i.e. subgraph_deploy)
- `remove_subgraph`
-  BlockStream when setting the `synced` flag

timmclean:
I just added some comments to help guide review -- it might help to skim through those first :slightly_smiling_face:

leodasvacas:
> Finally, this PR changes subgraph_remove to keep SubgraphDeployment entities even when they are no longer referenced by any subgraph version. 

To clarify, now if I redeploy subgraph after removing it, it will pickup from where it left?

store/postgres/src/filter.rs
```diff
@@ -198,13 +198,27 @@ fn build_filter(filter: EntityFilter) -> Result<FilterExpression, UnsupportedFil
...
+                } else {
+                    Box::new(
+                        sql("data -> ")
+                            .bind::<Text, _>(attribute.clone())
+                            .sql(" IS NULL ")
+                            .or(sql("data -> ")
+                                .bind::<Text, _>(attribute)
+                                .sql(" ->> 'type' = 'Null' ")),
+                    )
+                }),
```
leodasvacas: It would be nice to get a comment explaining the new semantics.

core/tests/tests.rs
```diff
@@ -200,6 +200,7 @@ fn added_subgraph_id_eq(
 }
 
 #[test]
+#[ignore]
```
leodasvacas: Why did this need to be ignored?

timmclean: Thank you for catching this! I did mean to do more here.

timmclean: Fixed

store/postgres/src/filter.rs
timmclean: As requested!

graph/src/components/store.rs
```diff
@@ -41,6 +43,26 @@ pub enum EntityFilter {
     NotEndsWith(Attribute, Value),
 }
 
+// Define some convenience methods
+impl EntityFilter {
+    pub fn new_equal(
```
Jannis: I think these could both be just `in` and `equal`. What do you think?

graph/src/components/store.rs
```diff
@@ -397,19 +415,255 @@ pub trait Store: Send + Sync + 'static {
...
+            .as_string()
+            .ok_or_else(|| format_err!("SubgraphVersion entity has wrong type in `deployment`"))?;
+        SubgraphDeploymentId::new(subgraph_id_str)
+            .map_err(|()| {
+                format_err!("SubgraphVersion entity has invalid subgraph ID in `deployment`")
+            })
+            .map(Some)
+    }
+
+    /// Read all version entities pointing to the specified deployment IDs and determine whether
```
Jannis: The comment lines look a little long...

Jannis: It would be nice to document what the returned operations mean.

graph/src/components/store.rs
```diff
@@ -397,19 +415,255 @@ pub trait Store: Send + Sync + 'static {
...
+        });
+
+        // Find subgraphs with one of these versions as current or pending
+        let subgraphs_with_version_as_current_or_pending =
+            self.find(SubgraphEntity::query().filter(EntityFilter::Or(vec![
+                EntityFilter::new_in("currentVersion", version_ids.clone()),
+                EntityFilter::new_in("pendingVersion", version_ids.clone()),
+            ])))?;
+        let subgraph_ids_with_version_as_current_or_pending =
+            subgraphs_with_version_as_current_or_pending
```
Jannis: It would be nice to have a helper for all these transformations. Something that takes an `IntoIterator<T>` and a closure from T to U, maps over the iterator using the closure, calls collect and returns a `Coll<U>` that can be inferred from the variable type the result gets assigned to. Something like:
```
let ids: HasSet<_> = map(versions, |version| version.id().to_string()); 
```

But that's just a general thought, not important for this PR.

graph/src/components/store.rs
```diff
@@ -397,19 +415,255 @@ pub trait Store: Send + Sync + 'static {
...
+                .into_iter()
+                .collect(),
+        });
+
+        // Produce summaries, deriving flags from information in subgraph entities
+        let version_summaries =
+            versions
+                .into_iter()
+                .map(|version_entity| {
+                    let is_current = subgraphs_with_version_as_current_or_pending.iter().any(
```
Jannis: Is there anything that guarantees that a subgraph version always belongs to only one subgraph? I know the updating logic probably prevents that but just looking at this code you could think that a version could be considered  current or pending if it is current or pending for one subgraph but not for others.

It makes sense to batch the processing  here to keep it simple and I know this logic is correct but I wonder if we can make it more obvious here.

graph/src/components/store.rs
```diff
@@ -397,19 +415,255 @@ pub trait Store: Send + Sync + 'static {
...
+                        current: is_current,
+                        pending: is_pending,
+                    }
+                })
+                .collect();
+
+        Ok((version_summaries, ops))
+    }
+
+    /// Produce the EntityOperations needed to create/remove SubgraphDeploymentAssignments to
```
Jannis: Comment lines are too long, I think.

Jannis (changes requested):

graph/src/components/store.rs
timmclean: Unfortunately `in` is a reserved keyword :(

graph/src/components/store.rs
timmclean: Oh what's our preferred max width? I've been wrapping to 100

graph/src/components/store.rs
timmclean: I wonder if @leodasvacas knows any stdlib tricks for shortening this up that I'm not using? It does get annoying having to `iter()` and `collect()` so much...

graph/src/components/store.rs
Jannis: Whatever rustfmt does to the code. 90, I think?

graph/src/components/store.rs
Jannis: Ok, nevermind!

core/src/subgraph/registrar.rs
```diff
@@ -428,14 +499,87 @@ fn create_subgraph_version(
             .write_operations(&version_entity_id),
     );
 
+    // Simulate the creation of the new version and updating of Subgraph.pending/current
```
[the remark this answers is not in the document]
timmclean: There's the "real" changes to the SubgraphVersion and Subgraph entities -- those are done through EntityOperations. But this is just about producing the new list of SubgraphVersionSummary's, "simulating" how things will look after the EntityOperations are applied (the entity ops haven't been applied yet, so we can't just read from the store).

core/src/subgraph/registrar.rs
```diff
@@ -401,22 +411,83 @@ fn create_subgraph_version(
...
+                        .as_string()
+                        .unwrap(),
+                )
+                .unwrap(),
+            )
+        }
+        None => None,
+    };
+
+    // Find all subgraph version entities that point to this hash or a hash
```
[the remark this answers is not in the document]
timmclean: I'm working on a write-up for this, will post in a few minutes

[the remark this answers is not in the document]
timmclean: So, there are three relevant entity types here: Subgraph, SubgraphVersion, and SubgraphDeploymentAssignment. Subgraph and SubgraphVersion are straightforward to add/update/remove. The challenge is to make sure that SubgraphDeploymentAssignments are created/removed properly.

For the most part, you can think of SubgraphDeploymentAssignments as being derived from Subgraphs and SubgraphVersions. In other words, you could write a SQL query that looks only at Subgraphs and SubgraphVersions and tells you exactly what SubgraphDeploymentAssignments should exist right now. Specifically, a SubgraphDeploymentAssignment should exist for a hash if and only if there exists a SubgraphVersion that points to that hash **and** is either current or pending.

What makes this difficult is that there are many cases where SubgraphVersions can be added/removed, or change from being pending or current to not being pending or current. Before this PR, I wrote separate logic to handle each case:
- When a SubgraphVersion was removed, I would check to see if there are any other versions pointing to that hash, filter them down to the versions that are "current" (or "pending" now with this PR), and if no, I would remove the assignment as well.
- When a SubgraphVersion is created, I would check to see if the previously current version's assignment needs to be removed. Similar check: look for other versions pointing to that hash, filter them down to versions that are "current", exclude the version that is about to become non-current, and if none found, remove the assignment.

This PR is adding more cases, so more logic like this would be needed. It struck me that this was getting redundant. I mentioned earlier the rule for when an assignment should exist: "a SubgraphDeploymentAssignment should exist for a hash if and only if there exists a SubgraphVersion that points to that hash **and** is either current or pending". In each case, I was essentially re-encoding this rule for each set of circumstances, and thinking through the logic for each set of circumstances was getting rather difficult, particularly in edge cases (e.g. a new subgraph version is being created, but it has the same hash as the previous current version -- therefore the assignment shouldn't be removed, and no new one should be created).

So instead, I refactored to add two common methods:
- `read_subgraph_version_summaries`: Reads information from the store about Subgraph and SubgraphVersion entities, produces summary objects.
- `reconcile_assignments`: Looks at two sets of information about Subgraph and SubgraphVersion entities (the summary objects from above) and decides what assignments should be removed or created. The first set is a "before" view (how things are now), and the second set is an "after" view (what the Subgraph and SubgraphVersion entities are being changed to).

The rule for determining whether an assignment should exist is now implemented only in one place: `reconcile_assignments`. The "before" view is easy to get using `read_subgraph_version_summaries`. The difficult part is computing the "after" view. In each place that Subgraph and SubgraphVersion entities are added/updated/removed, these three steps must be followed in order to ensure the right SubgraphDeploymentAssignments exist:
- Call `read_subgraph_version_summaries` to get the "before" state.
- Figure out what the "after" state will look like by applying some changes to the summary objects. I called this "simulating" in the comments.
- Call `reconcile_assignments` with the before and after states.

core/src/subgraph/registrar.rs
```diff
@@ -401,22 +411,83 @@ fn create_subgraph_version(
...
-                    .unwrap_or(Value::Null),
-            ),
+            EntityFilter::new_equal("name", name.to_string()),
+            EntityFilter::new_equal("currentVersion", current_version_id_opt.clone()),
+            EntityFilter::new_equal("pendingVersion", pending_version_id_opt.clone()),
         ])),
         entity_ids: vec![subgraph_entity_id.clone()],
     });
 
+    // Look up current version's deployment hash
```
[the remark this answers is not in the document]
timmclean: Done!

graph/src/components/store.rs
timmclean: Ah, I'll add some asserts! It is guaranteed that a subgraph version points to a single subgraph because the subgraph version entity points to only one subgraph, but it's only implicit that Subgraph.currentVersion.subgraph_id matches Subgraph.id

graph/src/components/store.rs
timmclean: Sure. I tightened these up, and maybe next week we can turn on auto comment wrapping in rustfmt

timmclean:
@Jannis ready for next round

core/src/subgraph/registrar.rs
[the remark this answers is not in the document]
Jannis: Ok, got it.

Jannis:
Looks good to me!

Jannis (approved):

timmclean:
Yayyy!

PR #697, opened 2019-01-19
docs: add information about environment variables

lutter:
This came mostly out of my curiosity about the env variables that graph-node uses. I think it only makes sense to have this doc if we can commit to keeping it up-to-date as the code evolves. Otherwise, I am fine with just closing this PR without merging.


docs/environment-variables.md
```diff
@@ -0,0 +1,35 @@
...
+  parallel (defaults to 50)
+
+## Running mapping handlers
+* `GRAPH_EVENT_HANDLER_TIMEOUT`: amount of time an event handler is allowed
+  to take (in seconds, default is unlimited)
+
+## Miscellaneous
+* `GRAPH_LOG`: control log levels, the same way that `RUST_LOG` is
+described [here](https://docs.rs/env_logger/0.6.0/env_logger/)
+* `THEGRAPH_SENTRY_URL`:
```
Jannis: This line allows to set a Sentry endpoint to log panics to, which can be nice in hosted versions.

lutter:
I added a note at the top. Lmk if that's what you had in mind.

For the descriptions, I'd love to have a little more of an explanation of why you would change the defaults (e.g., what's the effect of changing `ETHEREUM_BLOCK_BATCH_SIZE` on the overall system) but I don't know that level of detail, and it's something that can be added over time.

Jannis:
@lutter I'm happy for this to go in. Feel free to merge. 👍

docs/environment-variables.md
```diff
@@ -0,0 +1,38 @@
...
+* `ETHEREUM_RPC_MAX_PARALLEL_REQUESTS`: how many RPC connections to start
+in parallel for block retrieval (defaults to 64)
+* `ETHEREUM_FAST_SCAN_END`: `graph-node` locates
+  blocks with events for a particular subgraph. Most subgraphs do not have
+  events in the first few million blocks, so `graph-node` optimizes for that
+  case by inquiring about a large range. The value of this variable
+  is the block number at which we switch from probing large ranges to
+  smaller ranges (defaults to 4000000)
+* `DISABLE_BLOCK_INGESTOR`: set to `true` to disable block ingestion. Any
+  other value enables block ingestion.
```
timmclean: How about rephrasing this to "Leave variable unset or set to `false` to leave block ingestion enabled" or similar? I don't think we should commit to the "any other value" piece in case that changes

lutter: Ok. I made that change and rebased on the latest master.

PR #699, opened 2019-01-19
Add optional environment variables EXTERNAL_{HTTP,WS}_BASE_URL

timmclean:
Adds two optional environment variables to tell graph-node to send back full URLs in response to `subgraph_deploy` requests.

See also: https://github.com/graphprotocol/graph-cli/pull/202

### DIFF
--- a/server/json-rpc/src/lib.rs
+++ b/server/json-rpc/src/lib.rs
@@ -1,5 +1,6 @@
 extern crate graph;
 extern crate jsonrpc_http_server;
+extern crate lazy_static;
 extern crate serde;
 
 use graph::prelude::{JsonRpcServer as JsonRpcServerTrait, *};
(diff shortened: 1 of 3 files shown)


server/json-rpc/src/lib.rs
```diff
@@ -244,15 +254,25 @@ pub fn parse_response(response: Value) -> Result<(), jsonrpc_core::Error> {
 }
 
 fn subgraph_routes(name: &SubgraphName, http_port: u16, ws_port: u16) -> Value {
+    let http_base_url = EXTERNAL_HTTP_BASE_URL
+        .clone()
+        .unwrap_or_else(|| format!(":{}", http_port));
```
Jannis: Could we omit the port from the output if it is port 80? That would be nice.

server/json-rpc/src/lib.rs
```diff
@@ -244,15 +254,25 @@ pub fn parse_response(response: Value) -> Result<(), jsonrpc_core::Error> {
...
+    let ws_base_url = EXTERNAL_WS_BASE_URL
+        .clone()
+        .unwrap_or_else(|| format!(":{}", ws_port));
```
Jannis: Actually, looking at this, I'd like the env vars to override the ports altogether, so you have full control  over the base url and potentially base paths. If you serve the Graph node from behind a proxy, both http and ws might be served from the same port and different paths. This solution wouldn't work with that kind of setup.

Jannis (changes requested):

server/json-rpc/src/lib.rs
timmclean: Yup, this allows for that! `http_base_url` is either something like `:8000`, in which case graph-cli fills in the protocol and host, or something like `http://example.com` or `http://example.com:8000`, in which case graph-cli will use the URL directly.

server/json-rpc/src/lib.rs
timmclean: That's accounted for here. EXTERNAL_*_BASE_URL can be something like `http://www.example.com:8080/some/path` if needed

Jannis (approved):

timmclean:
@lutter That's great, here are some descriptions for those new vars for when that PR's ready:

* `EXTERNAL_HTTP_BASE_URL`: A base for the externally accessible URL that graph-node clients will use to query this graph-node instance. If, for example, the base URL `https://www.example.com/foo` is provided, then graph-node will assume that requests to `https://www.example.com/foo/subgraphs/name/mysubgraph` will be routed to it. Note: the base URL should not end with a `/`.
* `EXTERNAL_WS_BASE_URL`: A base for the externally accessible URL that graph-node clients will use for GraphQL WebSocket subscriptions with this graph-node instance. If, for example, the base URL `wss://www.example.com/bar` is provided, then graph-node will assume that WebSocket connections to `wss://www.example.com/bar/subgraphs/name/mysubgraph` will be routed to it. Note: the base URL should not end with a `/`.

PR #701, opened 2019-01-21
Serialize warnings to Elasticsearch as 'warning', not 'warn'

Jannis:
Other software we have written expects this. It's also generally a little odd to have `warning` be cut off but all other levels being written out.


leoyvens (dismissed):
This might need a rebase on master to pass travis, but we can merge without that.

Jannis:
@leodasvacas Good point. I've rebased and will wait for the tests to pass.

leoyvens (approved):

PR #705, opened 2019-01-21
server/websocket: Make URL path parsing more robust

### LINKED ISSUE #638: Mixing of path types in server/websocket
```
Dec 11 01:25:02.513 CRIT assertion failed: `(left == right)`
  left: `Some("\\")`,
 right: `Some("/")`, location: server\websocket\src\server.rs:108
```

### LINKED ISSUE #704: CRIT assertion failed: `(left == right)`
🐛

After making some updates to the Livepeer Subgraph and running it on the latest version of Graph Node, after some time it eventually errors out with the following:
```
CRIT assertion failed: `(left == right)`
  left: `Some("http:")`,
 right: `Some("/")`, location: server/websocket/src/server.rs:38
```

I'm not sure what's causing this error, so I'm not sure what the steps are to reproduce it, but it seems to happen consistently after running the subgraph for ~24 hours. A link to the subgraph branch I'm working on can be found [here](https://github.com/adamsoffer/livepeerjs/tree/delegator/packages/subgraph).

timmclean:
An HTTP request line usually looks like this:

```
GET /subgraphs HTTP/1.1
```

The part in the middle called the "request target" (`/subgraphs` in this case) is normally just a path, but can optionally be a full URL. This is unusual, but RFC7230 says that HTTP servers must support it.

In `server/http`, hyper handles this transparently for us, but in `server/websocket` we use tungstenite, which does not seem to preprocess the request target, and graph-node would panic if the path did not start with `/`.

Resolves #638 and resolves #704

### DIFF
--- a/server/websocket/src/server.rs
+++ b/server/websocket/src/server.rs
@@ -2,8 +2,8 @@ use futures::prelude::*;
 use graph::data::subgraph::schema::SUBGRAPHS_ID;
 use graph::prelude::{SubscriptionServer as SubscriptionServerTrait, *};
 use graph::tokio::net::TcpListener;
+use hyper::Uri;
 use std::net::{IpAddr, Ipv4Addr, SocketAddr};
-use std::path::Path;
 use std::sync::Mutex;
 use tokio_tungstenite::accept_hdr_async;
 use tokio_tungstenite::tungstenite::{handshake::server::Request, Error as WsError};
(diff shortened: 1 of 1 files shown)


Jannis (approved):
Looks good, I've double-checked that parsing the URI and getting the path is a no-op for regular paths.

timmclean:
Oh great, thanks!

Issue #664, opened 2019-01-07
Fix panics in GraphQL query execution


Jannis:

timmclean:
@Jannis did you want me to take this one?

Jannis:
@timmclean I didn't. 😉

leodasvacas:
Also assinging @Jannis since I improved what I could find here, but we need to double check what this was originally about.

PR #707, opened 2019-01-22
Fix panics related to invalid schemas

leoyvens:
Part of #664, though I'm not sure if these are what that issue refers to. I grepped for panics, fixed two which could really cause a crash on an invalid schema: One when an undefined interface was implemented, another when an undefined type appeared inside a list. I also refactored away some unwraps.


Jannis (approved):
These are good changes. I'll have to look closer at the execution to assess whether they catch all panics we had in mind but we should definitely merge this because it makes the code cleaner and more robust.

PR #708, opened 2019-01-23
Limit max ipfs file size through env var

### LINKED ISSUE #667: Limit maximum file size that can be fetched from IPFS
Part of epic #132.

To prevent mappings from abusing IPFS capabilities. This should be configurable through an env var.

leoyvens:
Resolves #667.

### DIFF
--- a/graph/src/components/link_resolver.rs
+++ b/graph/src/components/link_resolver.rs
@@ -19,9 +19,11 @@ impl LinkResolver for ipfs_api::IpfsClient {
     fn cat(&self, link: &Link) -> Box<Future<Item = Vec<u8>, Error = failure::Error> + Send> {
         // Discard the `/ipfs/` prefix (if present) to get the hash.
         let path = link.link.trim_left_matches("/ipfs/").to_owned();
-        let max_file_bytes = env::var(MAX_IPFS_FILE_BYTES_ENV_VAR)
-            .ok()
-            .and_then(|s| u64::from_str(&s).ok());
+        let max_file_bytes = env::var(MAX_IPFS_FILE_BYTES_ENV_VAR).ok().map(|s| {
+            u64::from_str(&s).unwrap_or_else(|_| {
+                panic!("failed to parse env var {}", MAX_IPFS_FILE_BYTES_ENV_VAR)
+            })
+        });
 
         let cat = self
             .cat(&path)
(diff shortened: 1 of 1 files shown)


graph/src/components/link_resolver.rs
```diff
@@ -1,9 +1,12 @@
 use data::subgraph::Link;
 use failure;
 use ipfs_api;
+use std::env;
+use std::str::FromStr;
+use std::time::Duration;
 use tokio::prelude::*;
 
-use std::time::Duration;
+const MAX_IPFS_FILE_BYTES_ENV_VAR: &str = "GRAPH_MAX_IPFS_FILE_BYTES";
```
timmclean: How about making this a lazy_static with the parsed value? That way a parsing error will panic on start up, instead of going unnoticed

timmclean: Ah, might need to make it a parameter on IpfsClient then for testing

timmclean (changes requested):

graph/src/components/link_resolver.rs
```diff
@@ -12,18 +15,48 @@ pub trait LinkResolver: Send + Sync + 'static {
...
-                .map(|x| x.to_vec())
-                .map_err(|e| failure::err_msg(e.to_string())),
-        )
+        match max_file_bytes {
+            Some(max_bytes) => Box::new(
+                self.object_stat(&path)
+                    .map_err(|e| failure::err_msg(e.to_string()))
+                    .and_then(move |stat| match stat.cumulative_size > max_bytes {
+                        false => Ok(()),
+                        true => Err(format_err!("Ipfs file {} is too large", path)),
```
Jannis: Ipfs -> IPFS

graph/src/components/link_resolver.rs
```diff
@@ -12,18 +15,48 @@ pub trait LinkResolver: Send + Sync + 'static {
...
+    let file: &[u8] = &[0u8; 201];
+    let client = ipfs_api::IpfsClient::default();
+
+    let mut runtime = tokio::runtime::Runtime::new().unwrap();
+    let link = runtime.block_on(client.add(file)).unwrap().hash;
+    let err = runtime
+        .block_on(LinkResolver::cat(&client, &Link { link: link.clone() }))
+        .unwrap_err();
+    env::remove_var(MAX_IPFS_FILE_BYTES_ENV_VAR);
+    assert_eq!(err.to_string(), format!("Ipfs file {} is too large", link));
```
Jannis: Ipfs -> IPFS

graph/src/components/link_resolver.rs
leodasvacas: lazy_static is lazy so it wouldn't help with failing earlier. Though you make a good point that an env var failing to parse should be a fatal error, I've changed that to panic.

leodasvacas:
@timmclean @Jannis Thanks for the review, I've addressed the comments.

Jannis (approved):

timmclean (approved):

Jannis:
I'll merge this for you @leodasvacas, hope that's alright.

PR #711, opened 2019-01-24
Limit log stream range size

### LINKED ISSUE #676: Update subgraph block pointer while searching for relevant blocks
A significant amount of time can be spent in `find_first_blocks_with_logs`. It would be nice to update the block pointer to reflect the results of each `eth_getLogs` call.

timmclean:
Not perfect, but it's good enough for increasing how often the subgraph block pointer is updated during the first few million blocks 

Resolves #676


Jannis (approved):

Jannis:
Thanks @timmclean! I'm merging this immediately.

PR #721, opened 2019-01-30
Add subgraph names to Elasticsearch logs

### LINKED ISSUE #703: Add subgraph name to Elasticsearch logs

Jannis:
Resolves #703.

This PR includes the following changes:

* `split_logger` is moved from `core::log` to `graph::ext::slog`. Feels more consistent with how we've added extensions for futures in `graph::ext::futures` as well. But that may just be me.

* A new `TryFromEntity` type is added as a temporary alternative to the experimental `TryFrom` trait. It allows to convert from a generic `Entity` to more specific types like `SubgraphEntity` or `SubgraphVersionEntity`. The trait includes a few convenience methods for deserializing entity attributes and handling different errors with that.

* A new `Store::get_typed` method is added. It is essentially `Store::get` with added entity deserialization using `TryFromEntity`. If there are objections, I'm happy to drop this one, as it doesn't do a lot anyway.

* All events and methods related to assignments being added and subgraphs being started (such as `AssignmentEvent::Add` or `SubgraphProvider::start` now take the subgraph name. This is then passed to loggers in the `SubgraphInstanceManager`.

* A dedicated `subgraphName` field is now logged to Elasticsearch so we can easily identify subgraph logs for subgraphs by name.

### DIFF
--- a/core/src/subgraph/registrar.rs
+++ b/core/src/subgraph/registrar.rs
@@ -136,8 +136,21 @@ where
                                     if let Some(entity) = entity_opt {
                                         if entity.get("nodeId") == Some(&node_id.to_string().into())
                                         {
+                                            // Look up the subgraph for this deployment
+                                            let subgraph = get_subgraph_for_deployment_id(
+                                                store.clone(),
+                                                &subgraph_hash,
+                                            );
+
+                                            // Fail if we can't load the subgraph
+                                            let subgraph = match subgraph {
+                                                Ok(subgraph) => subgraph,
+                                                Err(e) => return Box::new(stream::once(Err(e))),
+                                            };
+
                                             // Start subgraph on this node
                                             Box::new(stream::once(Ok(AssignmentEvent::Add {
+                                                subgraph_name: subgraph.name,
                                                 subgraph_id: subgraph_hash,
                                                 node_id: node_id.clone(),
                                             })))
@@ -173,6 +186,7 @@ where
     fn start_assigned_subgraphs(&self) -> impl Future<Item = (), Error = Error> {
         let provider = self.provider.clone();
         let logger = self.logger.clone();
+        let store = self.store.clone();
 
         // Create a query to find all assignments with this node ID
         let assignment_query = SubgraphDeploymentAssignmentEntity::query()
@@ -195,8 +209,15 @@ where
             })
             .and_then(move |subgraph_ids| {
                 let provider = provider.clone();
+                let store = store.clone();
+
                 stream::iter_ok(subgraph_ids).for_each(move |id| {
-                    start_subgraph(id, &*provider, logger.clone()).map_err(|()| unreachable!())
+                    // Look up the subgraph for this deployment (we need its name)
+                    let subgraph = get_subgraph_for_deployment_id(store.clone(), &id)
+                        .expect("failed to resolve subgraph for deployment");
+
+                    start_subgraph(subgraph.name, id, &*provider, logger.clone())
+                        .map_err(|()| unreachable!())
                 })
             })
     }
@@ -301,9 +322,13 @@ where
 
     match event {
         AssignmentEvent::Add {
+            subgraph_name,
             subgraph_id,
             node_id: _,
-        } => Box::new(start_subgraph(subgraph_id, &*provider, logger).map_err(|()| unreachable!())),
+        } => Box::new(
+            start_subgraph(subgraph_name, subgraph_id, &*provider, logger)
+                .map_err(|()| unreachable!()),
+        ),
         AssignmentEvent::Remove {
             subgraph_id,
             node_id: _,
@@ -320,14 +345,15 @@ where
     }
 }
 
-// Never errors.
+/// Never errors.
 fn start_subgraph<P: SubgraphAssignmentProviderTrait>(
+    subgraph_name: SubgraphName,
     subgraph_id: SubgraphDeploymentId,
     provider: &P,
     logger: Logger,
 ) -> impl Future<Item = (), Error = ()> + 'static {
     provider
-        .start(subgraph_id.clone())
+        .start(subgraph_name.clone(), subgraph_id.clone())
         .then(move |result| -> Result<(), _> {
             match result {
                 Ok(()) => Ok(()),
@@ -340,6 +366,7 @@ fn start_subgraph<P: SubgraphAssignmentProviderTrait>(
                         logger,
                         "Subgraph instance failed to start";
                         "error" => e.to_string(),
+                        "subgraph_name" => subgraph_name.to_string(),
                         "subgraph_id" => subgraph_id.to_string()
                     );
                     Ok(())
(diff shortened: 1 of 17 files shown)


graph/src/data/store/mod.rs
```diff
@@ -439,6 +440,75 @@ impl<'a> From<Vec<(&'a str, Value)>> for Entity {
     }
 }
 
+pub trait TryFromEntity: Sized {
```
Jannis: @leodasvacas I'm particularly curious if you have better suggestions for designing the helper methods here.

Jannis:
@timmclean Could you let me know if this is roughly how you'd have done it? 😉

graph/src/data/subgraph/schema.rs
```diff
@@ -58,17 +58,48 @@ pub trait TypedEntity {
...
+    pub created_at: u64,
 }
 
 impl TypedEntity for SubgraphEntity {
     const TYPENAME: &'static str = "Subgraph";
     type IdType = String;
 }
 
+impl TryFromEntity for SubgraphEntity {
+    const ENTITY_TYPE: &'static str = "Subgraph";
```
timmclean: How about merging `TryFromEntity` with `TypedEntity` to avoid the duplication here? Then `get_typed<T>` would be generic over `TypedEntity`, which makes sense

graph/src/data/store/mod.rs
```diff
@@ -439,6 +440,75 @@ impl<'a> From<Vec<(&'a str, Value)>> for Entity {
...
+            },
+            None => Ok(None),
+        }
+    }
+
+    fn extract<T, F>(
+        entity: &mut Entity,
+        name: impl Into<String>,
+        type_name: impl Into<String>,
+        f: F,
```
timmclean: We have a `ValueType` enum in the graph crate -- how about using that in place of `type_name`? Then this method could also take care of calling `.as_string()`, etc

graph/src/data/subgraph/schema.rs
```diff
@@ -133,6 +164,14 @@ impl SubgraphEntity {
...
         ops
     }
+
+    pub fn key_from_version(version: &SubgraphVersionEntity) -> EntityKey {
+        EntityKey {
+            subgraph_id: SUBGRAPHS_ID.to_owned(),
+            entity_type: Self::TYPENAME.to_owned(),
+            entity_id: version.subgraph_id.clone(),
+        }
+    }
```
timmclean: This can be just

```rust
Self::key(version.subgraph_id.clone())
```

graph/src/data/subgraph/schema.rs
```diff
@@ -164,6 +203,35 @@ impl SubgraphVersionEntity {
         entity.set("createdAt", self.created_at);
         vec![set_entity_operation(Self::TYPENAME, id, entity)]
     }
+
+    pub fn query_from_deployment(id: &SubgraphDeploymentId) -> EntityQuery {
+        EntityQuery::new(SUBGRAPHS_ID.to_owned(), Self::TYPENAME)
+            .filter(EntityFilter::new_equal("deployment", id.to_string()))
+    }
```
timmclean: `EntityQuery::new(SUBGRAPHS_ID.to_owned(), Self::TYPENAME)` can be simplified to `Self::query()`

graph/src/data/subgraph/mod.rs
```diff
@@ -266,7 +266,7 @@ impl From<::diesel::result::Error> for SubgraphAssignmentProviderError {
 #[derive(Debug, PartialEq)]
 pub enum SubgraphAssignmentProviderEvent {
     /// A subgraph with the given manifest should start processing.
-    SubgraphStart(SubgraphManifest),
+    SubgraphStart(SubgraphName, SubgraphManifest),
```
timmclean: I'm a bit nervous about having the name back in this event. It used to be here and it ended up being used in many places, which caused problems for deduplication. Can we add some comments here to make clear that the name is only for internal debugging and may not be correct? In particular:

- Subgraph deployments can be associated with more than one name, so there may be other names
- The name is only looked up at start, so this subgraph deployment might no longer be associated with this name
- graph-node will frequently see multiple subgraph deployments with the same "name" here, because of the last bullet point (and also because of pending + current), so it can't be used as an identifier

Maybe in places where this name is used (e.g. `SubgraphAssignmentProvider::start`), it should be called something different to make clear its limitations. "logging_name"? "initial_name"?

timmclean: Alternatively the name resolution could be done closer to the elastic logger so that the risk of accidentally relying on the name for other things is reduced

core/src/subgraph/registrar.rs
```diff
@@ -657,6 +684,30 @@ fn get_subgraph_version_deployment_id(
...
 
+/// Performs a reverse lookup from a deployment to a subgraph.
+fn get_subgraph_for_deployment_id(
+    store: Arc<impl Store>,
+    deployment_id: &SubgraphDeploymentId,
+) -> Result<SubgraphEntity, Error> {
+    // Resolve the deployment into the version it belongs to
+    let version = SubgraphVersionEntity::try_from_entity(
+        store
+            .find_one(SubgraphVersionEntity::query_from_deployment(deployment_id))?
```
timmclean: How about `store.find_one_typed`?

timmclean (changes requested):
Yayyyyy statically typed entity reading!! I wanted this

timmclean:
`Store::get_typed` is great. `Store::find_typed` and `Store::find_one_typed` seem like natural next steps

leodasvacas:
As I was originally concerned, reading from the store into these structs can lead to inefficiency, even though it's an elegant abstraction. All we require is to get the name for an id, for that `get_subgraph_for_deployment_id` does two queries to the db and fetches more data than necessary, I assume that we could construct a single query to do this more efficiently.

I might prefer a simple `name_for_deployment_id` method on the `Store` that can be called directly when `ElasticDrainConfig` is constructed. Then we could choose to either merge these abstractions now since they're likely to prove useful in the future, or just keep them in mind and withhold from merging until they prove necessary.

graph/src/data/subgraph/schema.rs
Jannis: Good idea, will do that.

graph/src/data/store/mod.rs
Jannis: Yes. I didn't like the type name as a string. Let's do it.

graph/src/data/subgraph/mod.rs
Jannis: The more I think about it again, the more I think we don't want this at all. It just doesn't make sense to show the subgraph name in the logs if the deployment could be running for different subgraphs with different names at the same time.

Jannis:
@leodasvacas I'm not concerned with that so much, as this only happens once at startup / deploy time.

_But_, I think I'm going to close this PR. It would be great to have subgraph names in logs but I don't think we can due to the subgraph deployment deduplication. It doesn't make sense.

Let's keep the `TryFromEntity` in mind in case we need it again though.

PR #722, opened 2019-01-30
Clear entity history between test runs

### LINKED ISSUE #709: Spurious failure on `revert_block_with_partial_update`
While testing something unrelated, I got this test failure:
```
thread 'tokio-runtime-worker-0' panicked at 'assertion failed: `(left == right)`
  left: `Entity({"age": Int(67), "coffee": Bool(false), "weight": Float(184.4), "bin_name": Bytes(Bytes([74, 111, 104, 110, 116, 111, 110])), "favorite_color": Null, "id": String("1"), "name": String("Johnton"), "email": String("tonofjohn@email.com")})`,
 right: `Entity({"bin_name": Bytes(Bytes([74, 111, 104, 110, 116, 111, 110])), "email": String("tonofjohn@email.com"), "weight": Float(184.4), "name": String("Johnton"), "id": String("1"), "age": Int(67), "coffee": Bool(false)})`', store/postgres/tests/store.rs:1468:9
note: Run with `RUST_BACKTRACE=1` for a backtrace.
test revert_block_with_partial_update ... FAILED
```
Note that the first entity has `"favorite_color": Null` which the second doesn't.

This suggests a race condition, could be something specific to the test or could be an actual bug. Ping @timmclean and @fordN.

timmclean:
Not clearing the history would cause confusing test failures occasionally

Bonus: I reworked `run_test` to reuse the Store instance so that the Postgres connection can be reused, which speeds up the store tests significantly

Resolves #709

### DIFF
--- a/store/postgres/tests/store.rs
+++ b/store/postgres/tests/store.rs
@@ -28,7 +28,24 @@ fn postgres_test_url() -> String {
 }
 
 lazy_static! {
-    static ref TEST_MUTEX: Mutex<()> = Mutex::new(());
+    static ref STORE_MUTEX: Mutex<Arc<DieselStore>> = {
+        let logger = Logger::root(slog::Discard, o!());
+        let postgres_url = postgres_test_url();
+        let net_identifiers = EthereumNetworkIdentifier {
+            net_version: "graph test suite".to_owned(),
+            genesis_block_hash: TEST_BLOCK_0_PTR.hash,
+        };
+        let network_name = "fake_network".to_owned();
+
+        Mutex::new(Arc::new(DieselStore::new(
+            StoreConfig {
+                postgres_url,
+                network_name,
+            },
+            &logger,
+            net_identifiers,
+        )))
+    };
     static ref TEST_SUBGRAPH_ID: SubgraphDeploymentId =
         SubgraphDeploymentId::new("testsubgraph").unwrap();
     static ref TEST_BLOCK_0_PTR: EthereumBlockPointer = (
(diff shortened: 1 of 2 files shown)


leoyvens (commented):
I have one comment on the code. This solved the issue and the performance is great. However, `entity_changes_are_fired_and_forwarded_to_subscriptions` is consistently hanging for me (though not when run in isolation). It also hanged one of the travis builds. Not sure if it's related or should block this PR, but worth taking a look.

store/postgres/tests/store.rs
```diff
@@ -1570,6 +1586,7 @@ fn entity_changes_are_fired_and_forwarded_to_subscriptions() {
         subscription
             .take(4)
             .collect()
+            .timeout(Duration::from_secs(3))
```
timmclean: This will help the test die more quickly if this comes up again instead of forcing Travis to time out (which takes 10 minutes)

timmclean:
Fixed!

store/postgres/tests/store.rs
leodasvacas: Alright, this is good enough for now.

leoyvens (approved):

store/postgres/tests/store.rs
timmclean: Oh, I also fixed the race condition, don't worry :)

timmclean: :joy: The store mutex guard lifetime wasn't long enough -- I had to move it outside of the `block_on` closure

PR #723, opened 2019-01-31
Replace current version with new version if current is not fully synced yet

lutter:
I believe this does what it should do. The policy for replacing current or pending with a new version now is: if there is no current version, or the current version is not synced, replace it with the new version. Otherwise, replace the pending version. I think the code could be simplified a little if it expressed these rules more clearly.

In particular, I'd subsume the logic around `current_version_id_opt.is_some()` and `current_is_synced` into a variable `replace_current` and structure the code according to that. But it might be better to do that as a separate change as it will be a bit more invasive than this PR.

### DIFF
--- a/core/src/subgraph/registrar.rs
+++ b/core/src/subgraph/registrar.rs
@@ -455,6 +455,12 @@ fn create_subgraph_version(
         None => None,
     };
 
+    // See if current version is fully synced
+    let current_is_synced = match &current_version_hash_opt {
+        Some(hash) => store.is_deployment_synced(hash.to_owned())?,
+        None => false,
+    };
+
     // Find all subgraph version entities that point to this hash or a hash
     let (version_summaries_before, read_summaries_ops) = store.read_subgraph_version_summaries(
         iter::once(manifest.id.clone())
@@ -507,9 +513,12 @@ fn create_subgraph_version(
             version_summaries_after
         }
         SubgraphVersionSwitchingMode::Synced => {
-            // Just set the new version is pending, unless there is no current versions yet
+            // There is a current version. Depending on whether it is synced
+            // or not, make the new version the pending or the current version
             if current_version_id_opt.is_some() {
                 // Previous pending version (if there was one) is no longer pending
+                // Previous current version if it's not fully synced is no
+                // longer the current version
                 let mut version_summaries_after = version_summaries_before
                     .clone()
                     .into_iter()
@@ -517,18 +526,35 @@ fn create_subgraph_version(
                         if Some(&version_summary.id) == pending_version_id_opt.as_ref() {
                             version_summary.pending = false;
                         }
+                        if !current_is_synced
+                            && Some(&version_summary.id) == current_version_id_opt.as_ref()
+                        {
+                            // We will make the new version the current version
+                            version_summary.current = false;
+                        }
 
                         version_summary
                     })
                     .collect::<Vec<_>>();
 
-                // Add new version, which will be pending but not current
+                // Determine if this new version should be the pending or
+                // current version. When we add a version to a subgraph that
+                // already has a current version, the new version becomes the
+                // pending version if the current version is synced,
+                // and replaces the current version if the current version is still syncing
+                let (pending, current) = if current_is_synced {
+                    (true, false)
+                } else {
+                    (false, true)
+                };
+
+                // Add new version, setting pending and current appropriately
                 version_summaries_after.push(SubgraphVersionSummary {
                     id: version_entity_id.clone(),
                     subgraph_id: subgraph_entity_id.clone(),
                     deployment_id: manifest.id.clone(),
-                    pending: true,
-                    current: false,
+                    pending,
+                    current,
                 });
 
                 version_summaries_after
@@ -605,8 +631,9 @@ fn create_subgraph_version(
             ));
         }
         SubgraphVersionSwitchingMode::Synced => {
-            // Just set the new version is pending, unless there is no current versions yet
-            if current_version_id_opt.is_some() {
+            // Set the new version as pending, unless there is no fully synced
+            // current versions
+            if current_is_synced {
                 ops.extend(SubgraphEntity::update_pending_version_operations(
                     &subgraph_entity_id,
                     Some(version_entity_id),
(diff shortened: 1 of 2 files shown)


graph/src/data/subgraph/schema.rs
```diff
@@ -199,6 +200,22 @@ impl SubgraphDeploymentEntity {
...
+        let filter = EntityFilter::new_equal("id", id.0.clone());
+        let entity = store.find_one(SubgraphDeploymentEntity::query().filter(filter))?;
+        entity
+            .map(|entity| match entity.get("synced") {
+                Some(Value::Bool(true)) => Ok(true),
+                _ => Ok(false),
+            })
+            .unwrap_or(Ok(false))
+    }
+
```
timmclean: We had an unwritten rule here of avoiding using references to the `Store` -- no side effects, pure functions only in this file

core/src/subgraph/registrar.rs
```diff
@@ -507,28 +513,48 @@ fn create_subgraph_version(
...
+
+                // Add new version, setting pending and current appropriately
                 version_summaries_after.push(SubgraphVersionSummary {
                     id: version_entity_id.clone(),
                     subgraph_id: subgraph_entity_id.clone(),
                     deployment_id: manifest.id.clone(),
-                    pending: true,
-                    current: false,
+                    pending: pending,
+                    current: current,
```
timmclean: These can tidy up to just `pending,` and `current,`

core/src/subgraph/registrar.rs
```diff
@@ -606,7 +632,7 @@ fn create_subgraph_version(
         }
         SubgraphVersionSwitchingMode::Synced => {
             // Just set the new version is pending, unless there is no current versions yet
-            if current_version_id_opt.is_some() {
+            if current_version_id_opt.is_some() && current_is_synced {
```
timmclean: Can this be simplified to just the 2nd condition? I think `current_is_synced` implies `current_version_id_opt.is_some()`

core/src/subgraph/registrar.rs
```diff
@@ -606,7 +632,7 @@ fn create_subgraph_version(
         }
         SubgraphVersionSwitchingMode::Synced => {
             // Just set the new version is pending, unless there is no current versions yet
```
timmclean: Looks like I typo'd here, can you `s/is pending/as pending/` while you're in here :joy:

timmclean: Actually, it would help to clarify this comment too with the new behaviour

graph/src/data/subgraph/schema.rs
lutter: Is there a better place to put something like that? I am a little hesitant to just stick that into `create_subgraph_version` since it is less clear when it's just inlined.

timmclean: There's some similar stuff as methods on the Store trait, see https://github.com/graphprotocol/graph-node/blob/master/graph/src/components/store.rs#L419
Could also just use a function in this file.

The Store trait is getting a bit heavy, I'm not sure what the best solution is there. We may need another component or something

graph/src/data/subgraph/schema.rs
```diff
@@ -199,6 +200,22 @@ impl SubgraphDeploymentEntity {
...
     }
 
+    /// Return true if the deployment with the given id is fully synced,
+    /// and return false otherwise. Errors from the store are passed back up
+    pub fn is_synced(
+        store: &Store,
+        id: &SubgraphDeploymentId,
+    ) -> Result<bool, QueryExecutionError> {
+        let filter = EntityFilter::new_equal("id", id.0.clone());
+        let entity = store.find_one(SubgraphDeploymentEntity::query().filter(filter))?;
```
timmclean: Ah, one more thing. There is a store query here, so it needs to have a corresponding `AbortUnless` op

graph/src/data/subgraph/schema.rs
lutter: I moved this to the Store trait.

core/src/subgraph/registrar.rs
lutter: Done

core/src/subgraph/registrar.rs
lutter: Yep, good catch.

core/src/subgraph/registrar.rs
lutter: Ok, I rewrote the comment.

lutter:
> Some style/organizational nitpicks, but otherwise looks good! Were you able to test the new functionality by the way?

Yes, from my testing, it does replace a non-synced current version with the new version

lutter:
Addressed comments and rebased to latest master.

graph/src/components/store.rs
```diff
@@ -683,6 +683,19 @@ pub trait Store: Send + Sync + 'static {
             .map_err(|e| format_err!("Failed to query SubgraphDeployment entities: {}", e))
             .map(|entity_opt| entity_opt.is_some())
     }
+
+    /// Return true if the deployment with the given id is fully synced,
+    /// and return false otherwise. Errors from the store are passed back up
+    fn is_deployment_synced(&self, id: &SubgraphDeploymentId) -> Result<bool, Error> {
+        let filter = EntityFilter::new_equal("id", id.to_string());
+        let entity = self.find_one(SubgraphDeploymentEntity::query().filter(filter))?;
```
Jannis: If we already have the ID it would be cheaper to use store.get instead of find_one.

timmclean: Try `self.get(SubgraphDeploymentEntity::key(id))`

graph/src/data/subgraph/schema.rs
timmclean: Yup you're right!

graph/src/components/store.rs
lutter: Good catch! I ended up having the function take ownership so that callers can decide whether `id` needs to be cloned or not.

lutter:
Ugh, I rebased this to latest master, wanting to merge it after the rebase. But it seems the force-push removed the approval. If one of you could approve it again, I'll merge it.

PR #727, opened 2019-02-01
Validate block ingestor ancestor count

timmclean:
Pull out `REORG_THRESHOLD` to node/src/main, and then validate it against ancestor count. I just wanted to make clear that if ancestor count is configured to be smaller than the reorg threshold, the block stream will run into trouble

### DIFF
--- a/node/src/main.rs
+++ b/node/src/main.rs
@@ -43,6 +43,9 @@ use graph_server_json_rpc::JsonRpcServer;
 use graph_server_websocket::SubscriptionServer as GraphQLSubscriptionServer;
 use graph_store_postgres::{Store as DieselStore, StoreConfig};
 
+const REORG_THRESHOLD: u64 = 50;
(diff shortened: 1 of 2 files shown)


node/src/main.rs
```diff
@@ -43,6 +43,9 @@ use graph_server_json_rpc::JsonRpcServer;
 use graph_server_websocket::SubscriptionServer as GraphQLSubscriptionServer;
 use graph_store_postgres::{Store as DieselStore, StoreConfig};
 
+const REORG_THRESHOLD: u64 = 50;
+const ANCESTOR_COUNT: u64 = 50;
+
```
timmclean: We could make these environmental variables... but I'll leave that to someone else ;)

leoyvens (approved):

Jannis (approved):

Jannis:
I'll merge it on @timmclean's behalf. Thanks!

PR #728, opened 2019-02-01
Add multicolumn index to entities table on the subgraph and entity columns

### LINKED ISSUE #724: Create (subgraph) and (subgraph, entity) indexes on entities table if they don't exist yet
These would greatly improve queries against the "subgraph of subgraphs". Let's add them to a new migration in graph-node.

So basically these two indexes (correct me if they could be improved):
```sql
CREATE INDEX subgraph_idx ON entities (subgraph);
CREATE INDEX subgraph_entity_idx ON entities (subgraph, entity);
```

fordN:
Resolves #724

### DIFF
--- a/store/postgres/migrations/2019-02-01-045927_add_entities_indexes/up.sql
+++ b/store/postgres/migrations/2019-02-01-045927_add_entities_indexes/up.sql
@@ -1,0 +1,4 @@
+-- Change the order of columns in the pk index for entities
+ALTER TABLE entities DROP CONSTRAINT entities_pkey;
+ALTER TABLE entities ADD CONSTRAINT
+  entities_pkey primary key (subgraph, entity, id);
--- a/store/postgres/migrations/2019-02-01-045927_add_entities_indexes/down.sql
+++ b/store/postgres/migrations/2019-02-01-045927_add_entities_indexes/down.sql
@@ -1,0 +1,3 @@
+-- no work necessary on a down migration; we can live happily
+-- with either version of the pk index, but the up one provides
+-- better performance


leodasvacas:
Needing this is a bit surprising giving that those fields are part of the primary key. [This answer](https://stackoverflow.com/questions/11352056/postgresql-composite-primary-key) explains that this exact type of index would be created for composite primary keys, by order of declaration. So it seems we got the order of the primary key wrong, we have `PRIMARY KEY (id, subgraph, entity)` when `PRIMARY KEY (subgraph, entity, id)` would be better suit our use case.

So should we keep the primary key as-is and add this extra index, or should we do a migration for the primary so we don't need to add an index?

store/postgres/migrations/2019-02-01-045927_add_entities_indexes/up.sql
```diff
@@ -0,0 +1,3 @@
+CREATE INDEX IF NOT EXISTS entities_subgraph_entity_idx
+ON entities
+USING btree (subgraph, entity);
```
Jannis: Should we add `id` here at the end? As https://github.com/graphprotocol/graph-node/issues/724#issuecomment-459462836 explains, `(subgraph, entity, id)` covers `(subgraph, entity)` but on top of that also speeds up queries that also include the ID.

lutter: I think because we already have the pk index covering those three columns, we don't really need to add the `id` in here. I would only add it if we run into query plans that would benefit from it (though I am not sure what they would be right now)

Jannis:
True, we are now. IIRC though, the current plan for supporting interfaces will involve querying entities by `id` + `subgraph`, because at the time of the query you don't know which exact entity type the results we have. At that point, we'll need the `(id, subgraph, entity)` primary key index again. (@leodasvacas, please correct me if I'm wrong).

fordN:
I agree that if we are always including `subraph` and `entity` in the query filter it makes sense to simply alter the primary key index to be (subgraph, entity, id) rather than add a new index.  
However, I am not completely sold on the base assumption that the query performance will be the same (just because we haven't benchmarked it yet.)  I'll do some quick performance testing.

leodasvacas:
@Jannis I haven't implemented that yet, but I believe the query will look like:

`SELECT 1 FROM entities WHERE entites.subgraph = 'subgraph_id' AND entities.id = 'id' AND (entites.entity = 'interface_implementor1' OR entites.entity = 'interface_implementor2' ...)`

So it filters by `subgraph` and `id`, but also by multiple `entities`. I don't know what would be the query plan here, so don't know if an index by `(subgraph, id)` or by `(subgraph, entity, id)` would be prefered by postgres, but I believe either would work fine.

Jannis:
To understand the query plans used better, I've executed a few queries on the production database.

## Individual entity query

```sql
SELECT 1 FROM entities WHERE subgraph = 'subgraph_id' AND entity = 'foo' AND id = 'bar';
```

uses the `subgraph_entity_idx btree (subgraph, entity)` index.

## Collection query

```sql
SELECT * FROM entities WHERE subgraph = 'subgraph_id' AND entity = 'foo';
```

also uses this index.

## Interface query

```sql
SELECT 1 FROM entities 
WHERE subgraph = 'subgraph_id'
  AND id = 'id' 
  AND entity IN ('interface_implementor1', 'interface_implementor2' ...);
```

also uses this index.

My conclusion: I think we indeed only need the `(subgraph, entity, id)` index. @fordN @lutter Can we alter the primary key index with a migration or would it require a fresh `entities` table?

lutter:
Updated the migration to rebuild the primary key index in the desired column order. Note that this will have to reindex the entities table and might therefore take some time in production.

lutter:
Rebased on lastest master and merged.

lutter:
In case anybody wonders about the failed tests: I was a little too quick merging this, and hte branches that Travis pulls to run its tests don't exist anymore, causing the failures.

PR #733, opened 2019-02-04
Fix UB in runtime by removing all unsafe code

### LINKED ISSUE #607: Mark AscType trait as unsafe
There are some obligations that implementors of the `AscType` trait have in order to maintain memory safety, so we should mark `AscType` as unsafe. This is because `AscType`'s default implementations of `to_asc_bytes`, `from_asc_bytes`, and `asc_size` use unsafe Rust features in a way that is only safe if the type implementing `AscType` has certain properties. We should mark the `AscType` trait as `unsafe` so that implementors are made aware of these requirements.

(From the [nomicon](https://doc.rust-lang.org/nomicon/safe-unsafe-meaning.html):
> On trait declarations, `unsafe` means that implementors of the trait must check the trait documentation to ensure their implementation maintains the contracts the trait requires.

)

Any type implementing `AscType` must follow these rules:

1. The type cannot be a Rust pointer or reference. `from_asc_bytes` allows the assembly script mapping to control the memory address of the pointer or reference. `to_asc_bytes` leaks a memory address to the assembly script mappings, which will allow nondeterminism in the mappings and also could be used to help defeat ASLR. The `AscPtr` type is fine, however, because internally it is just a u32 offset into wasm memory and does not affect safety on the Rust side.
2. The type must never be in a state where it has uninitialized memory. In particular, for structs, the compiler may leave gaps between fields for alignment. The uninitialized bytes in these gaps will leak to the mappings via `to_asc_bytes`.
3. All possible byte sequences that match the type's size must be a valid state for the type. This seems to suggest that most enums are unsafe.
4. Any struct or enum fields must be of types that also follow these rules (the struct must not contain any pointers or references, etc).

Any type implementing `AscType` can alternatively override the built-in methods, removing the need to follow these rules (e.g. `AscString` and `ArrayBuffer` do this).

In addition, we also care about these properties are needed for correctness (but not safety):

5. The type must have a portable, defined memory layout. E.g. structs must be labelled `repr(C)`, `repr(u*)` or `repr(i*)`.
6. For determinism, all possible values for a type must be represented in memory in a canonical form. E.g. we can't have a fraction type `struct Fraction { top: i64, bottom: u64 }` unless we are careful to store the fraction only in reduced form (in which case we would need to be careful to check that invariant on deserialization).

I think we can cover most of these rules by using structs with `repr(C)` and being careful about what is in those structs. We need to make sure that the C representation of all structs have no padding. I'm tempted to say that we should use `repr(C, packed)` to solve this problem, but apparently [`repr(packed)` has soundness issues](https://github.com/rust-lang/rust/issues/27060) :( Enums don't seem to be safe as implemented in the `from_asc_bytes` direction currently.

leoyvens:
Fixes #607.

Instead of abusing `transmute` to get a one size fits all impl that turned out to really be unsafe, we derive `AscType` impls with a custom derive macro (couldn't resist playing with a new toy), or implement manually. Rust requires the macro to be in a separate crate. This allows us to remove the last lines of unsafe code in the repo.

@timmclean you might be curious to take a look at this.


runtime/derive/src/lib.rs
```diff
@@ -0,0 +1,176 @@
...
+                discriminant.to_asc_bytes()
+            }
+
+            fn from_asc_bytes(asc_obj: &[u8]) -> Self {
+                let mut u32_bytes: [u8; size_of::<u32>()] = [0; size_of::<u32>()];
+                u32_bytes.copy_from_slice(&asc_obj);
+                let discr = u32::from_le_bytes(u32_bytes);
+                match discr {
+                    #(#variant_discriminant2 => #enum_name_iter2::#variant_paths2,)*
+                    _ => panic!("value {} is out of range for {}", discr, stringify!(#enum_name))
```
leodasvacas: This is the critical difference from what we did before, we now catch if the value is outside the range of valid values for an enum.

timmclean:
Ohhh a custom derive, that's even better than I was picturing.

The `unsafe`s are gone, so that easily gets a :+1:

runtime/derive/src/lib.rs
```diff
@@ -0,0 +1,176 @@
...
+    };
+
+    TokenStream::from(quote! {
+        impl#impl_generics AscType for #struct_name#ty_generics #where_clause {
+            fn to_asc_bytes(&self) -> Vec<u8> {
+               let mut bytes = Vec::new();
+                #(bytes.extend_from_slice(&self.#field_names.to_asc_bytes());)*
+
+                // Assert that the struct has no padding.
+                assert_eq!(bytes.len(), size_of::<Self>());
```
timmclean: I like this check. I think it's technically not needed for safety now because it's all safe Rust code, but, assuming AssemblyScript's memory layout rules are similar to Rust's, this will help catch cases where (due to alignment or whatever) padding is needed on the ASc side

runtime/derive/src/lib.rs
```diff
@@ -0,0 +1,176 @@
...
+//             1u32 => JsonValueKind::Bool,
+//             2u32 => JsonValueKind::Number,
+//             3u32 => JsonValueKind::String,
+//             4u32 => JsonValueKind::Array,
+//             5u32 => JsonValueKind::Object,
+//             _ => panic!("value {} is out of range for {}", discr, "JsonValueKind"),
+//         }
+//     }
+// }
+fn asc_type_derive_enum(item_enum: ItemEnum) -> TokenStream {
```
timmclean: What happens if one of the enum variants has data fields? Can that be checked with an assert?

runtime/derive/src/lib.rs
timmclean: Ah, now there's a good use of `stringify!`

runtime/wasm/src/asc_abi/mod.rs
```diff
@@ -111,20 +97,131 @@ pub trait AscType: Sized {
 /// be equal to alignment.
```
timmclean: Does this comment need any updates?

runtime/derive/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "graph-runtime-derive"
+version = "0.5.0"
+edition = "2018"
+
+[lib]
+proc-macro = true
+
+[dependencies]
+syn = { version = "0.15", features =["full"] }
```
timmclean: missing space I think after `=`

timmclean (changes requested):

runtime/wasm/src/module/test/abi.rs
```diff
@@ -365,3 +365,23 @@ fn abi_big_int() {
     let new_uint_from_u256 = BigInt::from_signed_u256(&new_uint.to_signed_u256());
     assert_eq!(new_uint, new_uint_from_u256);
 }
+
+// This should panic rather than exhibiting UB. It's hard to test for UB, but
+// when reproducing a SIGILL was observed which would be caught by this.
```
Jannis: Pardon my ignorance: what's a SIGILL?

Jannis (commented):
Very nice! I'll leave it to you and Tim to work out the details in this PR.

runtime/derive/src/lib.rs
leodasvacas: The assert won't check that, but the match won't make sense and will not compile.

runtime/derive/src/lib.rs
leodasvacas: Yhea, it actually did catch one case where I needed to make implicit padding explicit.

runtime/wasm/src/module/test/abi.rs
leodasvacas: It's kinda like a segfault. It's a signal that's fired when the CPU tries to execute an invalid instruction, for example if you try to execute data.

runtime/derive/src/lib.rs
leodasvacas: Ha true, I found one!

runtime/wasm/src/asc_abi/mod.rs
leodasvacas: Good catch, it should not be mentioning a default implementation of `AscType`.

leodasvacas:
Adressed comments and made necessary updates.

runtime/wasm/src/module/test/abi.rs
Jannis: Wow, in all my days of writing C and other low-level languages, I've never come across it. Thanks for clarifying!

leodasvacas:
@timmclean This is gonna need your approval.

timmclean (approved):
Looks good!

leodasvacas:
@timmclean You're right, but those reprs represent the semantics of the custom derive, so if we didn't mark them we'd still think about those structs _as if_ they had those reprs, so we might as well just continue to mark them.

PR #735, opened 2019-02-04
Update subscriptions at most once per block (v2)

lutter:
This is a rewrite of PR #693 based on the feedback in that PR. This PR revamps how events for entity changes are generated and delivered to subscribers so that entity changes are batched per block for block-based data. Since events are generated in the application, we do not need the database triggers that previously generated change events any longer.

The biggest part of what was discussed in PR #693 and still remains to be done is limiting how often subsriptions are updated for subgraphs that are still syncing. I will do that in a separate PR.


lutter:
There was one wrinkle with how I had things initially: if `StoreEvent` has a separate `subgraph_id` field, that suggests that all changes in that event relate to the same subgraph. That is not true in general, and leads to trouble if we want time-based throttling without losing changes (where we would combine multiple events into one event) Rather than rework the whole PR I just tacked a commit onto the end to remove that field.

Curious to hear what yall think about this.

lutter:
> We should discuss how we could maintain events for revert operations caused by block reorgs. If such a reorg happens, clients need to receive updates.

When we revert, isn't it guaranteed that the very next thing we do is process the blocks from the 'new' main chain, which would then trigger updates? If there's a worry that revert affects a subgraph, but that subgraph is not changed when we (almost immediately) go forward again, we could add emitting store events into the pgsql code that does the reversion.

store/postgres/migrations/2019-02-04-114900_remove_entity_change_triggers/down.sql
```diff
@@ -0,0 +1,87 @@
+/**************************************************************
+* NOTIFY ENTITY ADDED
+*
+* Emits an entity added notification over the "entity_changes"
+* notification channel.
+**************************************************************/
+
+CREATE OR REPLACE FUNCTION notify_entity_added()
```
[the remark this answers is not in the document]
lutter: I was following the pattern laid down by all the other migrations where the `down` migration gets you back to where the DB was before the `up` migration. If we are fine with never going back, I can remove this. One consequence would be that there is no way to ever go back once code is deployed in production.

lutter:
I am looking at the `revert_block` stored proc; I think we can come up with a query that gives us all operations when we revert a block. We could then send those on the `stoer_events` channel.

lutter:
Here's [a gist](https://gist.github.com/lutter/e4540d90db1c918103ba2ce10927b326) showing how to query the changes that `revert_block` will make.

leodasvacas:
@fordN You might have some input here since you're the author of the revert procedure.

Jannis:
@lutter I took some time to understand the query that would collect all entity changes for a given block hash and subgraph ID, and played with it too. I believe it's correct. 😄 And together with `entity_history.data_before` it should give us everything we need to send a store event with `pg_notify`. 

You are right that reverting blocks is always followed by processing the blocks that replace them on the main chain. However, these new blocks may not include any data for the subgraphs that were affected by the reversion. If we don't send a store event out for each reverted block and there are no changes relevant to a subgraph on the new blocks following the reversion, then clients will never get an update (until there _is_ activity relevant to the subgraph again). This could leave clients in an inconsistent state for quite some time.

I think this is what @leodasvacas was saying as well.

My conclusion: When we revert, let's send out a store event as well. If you agree it makes sense, could you add this to the PR?

lutter:
Yes, that's the conclusion I came to, too. I'll update the PR to generate store events during reverts with the query above and send them out.

store/postgres/migrations/2019-02-04-114900_remove_entity_change_triggers/down.sql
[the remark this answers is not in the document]
Jannis: You are right actually. These migrations are, according to the Diesel documentation supposed to be, "bi-directional", so their `down.sql` should restore the previous state. I'll withdraw my concerns here 👍

graph/src/components/store.rs
```diff
@@ -166,13 +164,91 @@ impl EntityChange {
...
+}
+
+/// A boxed `StoreEventStream`
+pub type StoreEventStreamBox = Box<Stream<Item = StoreEvent, Error = ()> + Send>;
+
+/// A `StoreEventStream` produces the `StoreEvents`. Various filters can be applied
+/// to it to reduce which and how many events are delivered by the stream.
+pub struct StoreEventStream<S>
+where
+    S: Stream<Item = StoreEvent, Error = ()> + Send,
```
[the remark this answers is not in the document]
lutter: Good to know!

graph/src/components/store.rs
```diff
@@ -166,13 +164,91 @@ impl EntityChange {
...
+    /// at least one change to one of the given (subgraph, entity) combinations
+    /// will be delivered by the filtered stream.
+    pub fn filter_by_entities(
+        self,
+        entities: Vec<SubgraphEntityPair>,
+    ) -> StoreEventStream<impl Stream<Item = StoreEvent, Error = ()> + Send> {
+        let source = self.source.filter(move |event| {
+            event.changes.iter().any({
+                |change| {
+                    entities.contains(&(change.subgraph_id.clone(), change.entity_type.clone()))
```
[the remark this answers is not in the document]
lutter: Yes, that's a good point. I wound up using `any` there instead of `find`, but it does save us a lot of unnecessary allocation.

store/postgres/src/store.rs
```diff
@@ -209,42 +204,30 @@ impl Store {
...
-                .iter()
-                .filter(|(_, subscription)| {
-                    subscription
-                        .entities
-                        .contains(&(change.subgraph_id.clone(), change.entity_type.clone()))
-                })
-                .map(|(id, subscription)| (id.clone(), subscription.sender.clone()))
-                .collect::<Vec<_>>();
+        tokio::spawn(store_events.for_each(move |event| {
+            trace!(logger, "Received store event";
```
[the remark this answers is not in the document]
lutter: Ok. Fixed.

lutter:
I addressed all the comments and added a commit that generates a StoreEvent for block reversion. That was complicated by the fact that diesel makes it somewhat hard to run random queries. There is advice out there to create a view in these situations and declare a `table!` with diesel to ease this, but that doesn't work in our case as the block and subgraph for the reversion would have to be parameters to the view (which is not possible)

One thing I realized thinking about the query (and the `revert_block` logic it is based on): to figure out what to revert, we find all events for the block/subgraph combinations, and then revert based on those events. That means, that in theory, reversion for a block/subgraph combination could revert changes in other subgraphs. That does in fact happen, as the reversion affects the target subgraph and the `subgraphs` subgraph. Other subgraphs won't be affected because of how ingestion happens and history events are generated, but it's a somewhat surprising subtlety.

store/postgres/src/store.rs
```diff
@@ -845,11 +873,12 @@ impl StoreTrait for Store {
...
         // Prepare the new subscription by creating a channel and a subscription object
         let (sender, receiver) = channel(100);
-        let subscription = Subscription { entities, sender };
 
         // Add the new subscription
         let mut subscriptions = subscriptions.write().unwrap();
-        subscriptions.insert(id, subscription);
+        subscriptions.insert(id, sender);
+
+        let receiver = StoreEventStream::new(receiver).filter_by_entities(entities);
```
Jannis: I'd probably make this the caller's responsibility and drop the `entities` parameter from the method, but to maintain the behavior this makes complete sense. Let's roll with this and move the filtering out later if we deem it appropriate then.

store/postgres/src/store_events.rs
```diff
@@ -0,0 +1,141 @@
...
+
+impl QueryableByName<Pg> for EntityChangeQBN {
+    fn build<R: diesel::row::NamedRow<diesel::pg::Pg>>(
+        row: &R,
+    ) -> diesel::deserialize::Result<Self> {
+        let subgraph_id: String = row.get("subgraph")?;
+        let subgraph_id = SubgraphDeploymentId::new(subgraph_id).map_err(|_| {
+            format!(
+                "Illegal subgraph_id {}",
+                row.get("subgraph").unwrap_or("<unknown>".to_string())
```
Jannis: Since we've already obtained the subgraph ID in line 65, let's just use that here. May have to clone it to make it available in the closure but that's ok.

Jannis: An alternative would be to do `row.get("subgraph").unwrap()` here, since we would've returned an error in line 65 already if anything was wrong with the `"subgraph"` entry in the row.

graph/src/components/store.rs
```diff
@@ -135,16 +135,14 @@ impl EntityQuery {
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
 #[serde(rename_all = "lowercase")]
 pub enum EntityChangeOperation {
-    /// A new entity was added.
-    Added,
-    /// An existing entity was updated.
-    Updated,
+    /// An entity was added or updated
```
Jannis: The commit message states `The distinction between [Added and Updated] was meaningless in the code, and with the way we now generate store events, we can no longer say whether a change adds or updates an entity`.

I have no objections to merging the two variants into one but I'd disagree with the second half of that statement: through the `op_id` in the history table, we _can_ tell whether a `EntityOperation::Set` was an addition or an update. In fact, we rely on this to generate entity changes of type `'removed'` when building store events for reverted blocks.

store/postgres/src/store_events.rs
```diff
@@ -0,0 +1,141 @@
...
+// must be one less than block_ptr_from. The source of the event will be
+// block_ptr_to as we will be at that block after the reversion
+pub fn get_revert_event(
+    conn: &PgConnection,
+    subgraph_id: &SubgraphDeploymentId,
+    block_ptr_from: &EthereumBlockPointer,
+    block_ptr_to: EthereumBlockPointer,
+) -> Result<StoreEvent, StoreError> {
+    // Sanity check on block numbers
+    if block_ptr_from.number != block_ptr_to.number + 1 {
```
Jannis: I'd use `assert_eq!` here. Or, if we don't care about this in production builds (I think we _do_), `debug_assert_eq!`.

store/postgres/src/store_events.rs
```diff
@@ -0,0 +1,141 @@
...
+     WHEN h.op_id = 0 THEN 'removed'
+     WHEN h.op_id in (1,2) THEN 'set'
+   END) as change
+FROM entity_history h
+WHERE h.event_id in (select event_id from events_for_block_and_subgraph)
+ORDER BY h.event_id desc";
+    let query = diesel::sql_query(query)
+        .bind::<Text, _>(block_ptr_from.hash_hex())
+        .bind::<Text, _>(subgraph_id.to_string());
+    //dbg!(diesel::debug_query::<Pg, _>(&query));
```
Jannis: Remove

lutter:
> Worth noting also: the test suite is failing at the moment.

Yeah, I saw that last night. Restarting the jobs fixed some of them; I just bumped the timeouts in the tests to 30s, let's hope that takes care of it. To make this more robust, we'd need a way to close subscriptions, signal when all work has been done, and then take everything on the subscription stream and check that for correctness.

store/postgres/src/store.rs
```diff
@@ -818,6 +844,8 @@ impl StoreTrait for Store {
             );
             self.apply_entity_operations_with_conn(&conn, ops, EventSource::None)?;
 
+            self.emit_revert_event(&conn, &subgraph_id, &block_ptr_from, block_ptr_to)?;
```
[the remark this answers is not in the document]
lutter: It doesn't really matter: notifications are transactional, so other DB connections won't see the notifications until the current txn commits. I put it there initially because I wasn't sure if the revert proc would also remove the history entries it's rolling back (though I believe that's not the case)

store/postgres/src/store.rs
lutter: Yep, that was exactly what I was thinking there ;)

store/postgres/src/store_events.rs
lutter: I'd rather go with the second alternative, since using the existing `subgraph_id` requires that we clone for every result row, even though this whole error should never happen. (Though, to be sure, we'd need to put a check constraint on that column, so that we can just unwrap the `SubgraphDeploymentId`)

leodasvacas: You can't clone inside the closure so it's only cloned if necessary?

lutter: No, the `SubgraphDeploymentId::new(subgraph_id)` takes ownership of the value.

leodasvacas: Ah I see. Well subgraph ids have a fixed lenght, so you can reliably clone them in nanoseconds, I'd only worry about cloning variable length things.

graph/src/components/store.rs
lutter: I was referring to how we now generate EntityChanges in the app: at that point we do not know whether an operation will lead to an insert or update. You are of course right that we know from the history table. I'll clarify the commit message.

lutter:
Tracking down the test failures was a lot less fun than initially thought. What was happening was that on my machine the subscription received events that had been generated _before_ the subscription was established, and I erroneously called `take(..)` on the stream to get them plus the event from the revert operation, and then checked the last operation on the stream. In CI, the spurious event(s) did not appear on the stream, causing the `take(..)` to look for too many events, and then hang.

I've addressed that by generating a fake event once the subscription has been established, and skipping everything on the stream up to and including that fake event. That seems to eliminate the race between event production and consumption.

I've also addressed all the other review comments. This should be good to go now.

store/postgres/src/store.rs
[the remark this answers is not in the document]
Jannis: Got it, makes sense

lutter:
Rebased onto latest master

PR #737, opened 2019-02-06
Use ethabi that allows event overloading

leoyvens:
@davekaj can you double check that this works for you?


davekaj:
I can confirm that this works, I approve this pr

davekay100 (approved):
approved

PR #744, opened 2019-02-12
Point graph node to graphprotocol/ethabi/graph-patches to fix errors

davekaj:
Decentraland does not work unless ethabi supports event overloading and being able to parse event fields that aren't bytes32, such as this: https://etherscan.io/tx/0xc760d94f4512aac26123ad4345b824cdf271d0b3c9529778a42d3329b88b35bc#eventlog

The branch we are pointing to, created by @leodasvacas are a combination of https://github.com/paritytech/ethabi/pull/145 and https://github.com/paritytech/ethabi/pull/140

Hopefully those get merged one day into parities upstream branch, but for now this is needed to make Decentraland work.

### DIFF
--- a/graph/Cargo.toml
+++ b/graph/Cargo.toml
@@ -6,7 +6,11 @@ version = "0.5.0"
 backtrace = "0.3.9"
 diesel = { version = "1.3.2", features = ["postgres", "serde_json", "numeric", "r2d2"] }
 
-# Switch to upstream master once https://github.com/paritytech/ethabi/pull/145 is merged.
+# We would like to switch to upstream master once https://github.com/paritytech/ethabi/pull/145
+# is merged.
+# But, graph-patches also contains Jannis' https://github.com/paritytech/ethabi/pull/140 PR
+# to ethabi. Which we believe will not get merged. For now, we shall deviate from ethabi,
+# but long term we want to avoid forking off ethabi if possible
 ethabi = { git = "https://github.com/graphprotocol/ethabi.git", branch = "graph-patches" }
 hex = "0.3.2"
 futures = "0.1.21"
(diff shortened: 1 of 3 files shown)


graph/Cargo.toml
```diff
@@ -7,7 +7,7 @@ backtrace = "0.3.9"
 diesel = { version = "1.3.2", features = ["postgres", "serde_json", "numeric", "r2d2"] }
 
 # Switch to upstream master once https://github.com/paritytech/ethabi/pull/145 is merged.
```
leodasvacas: Could you update this comment to also mention jannis' PR?

runtime/wasm/Cargo.toml
```diff
@@ -4,7 +4,7 @@ version = "0.5.0"
 
 [dependencies]
 # Switch to upstream master once https://github.com/paritytech/ethabi/pull/145 is merged.
```
leodasvacas: same

Jannis:
Hmm. We should not include and depend on the changes in my ethabi PR. They are correct that this was a bug in a previous Solidity compiler version. If Decentraland's contracts were deployed with that bug then we may have to deal with the contracts being broken. Unless we can convince the ethabi team to take the PR in anwyay to be more resilient against bugs like that.

davekaj:
I think we should probably then deal with the bad solidity compiler versions of contracts ourselves. I think it is fair for the `ethabi` team to reject the issue since their code meets what the standard is. 

As you stated previously, Livepeer was having this issue, and now I found Decentraland to have this issue too. There are probably a good amount of other contracts that we haven't seen. 

So, I guess I will use this solution for the mean time, but not merge it into master. but we should have a plan to deal with this within the graph node

Jannis:
As we've just discussed, we're moving forward with this, taking in both fixes.

Jannis:
As we've just discussed, we're moving forward with this, taking in both fixes.

@davekaj Could you add the link to the other bug to the comments like @leodasvacas suggested?

davekaj:
@leodasvacas comments have been updated !

PR #747, opened 2019-02-13
Provide transaction input data to mappings and introduce runtime data structure versioning

### LINKED ISSUE #734: Make transaction input data available to mappings
This will also require small changes in https://github.com/graphprotocol/graph-ts.

fordN:
Resolves #734 

This PR updates the `EthereumTranasction` structures on the Rust side of the runtime module, adding an `input` property.
In order to maintain backward compatibility, the corresponding AssemblyScript structure is now versioned. The data version is chosen based on the subgraph version specified in the manifest (`manifest.SpecVersion`.)

### DIFF
--- a/core/src/subgraph/instance.rs
+++ b/core/src/subgraph/instance.rs
@@ -1,6 +1,5 @@
 use graph::prelude::{SubgraphInstance as SubgraphInstanceTrait, *};
 use graph::web3::types::{Log, Transaction};
-use semver::{Version, VersionReq};
 
 pub struct SubgraphInstance<T>
 where
@@ -27,21 +26,11 @@ where
         // we use the same order here as in the subgraph manifest to make the
         // event processing behavior predictable
         let manifest_id = manifest.id;
-        let spec_version = Version::parse(&manifest.spec_version)?;
-        if !VersionReq::parse("<= 0.0.2")
-            .unwrap()
-            .matches(&spec_version)
-        {
-            return Err(format_err!(
-                "Subgraph spec version {} not supported by this Graph Node (v0.0.2)",
-                spec_version
-            ));
-        }
 
         let (hosts, errors): (_, Vec<_>) = manifest
             .data_sources
             .into_iter()
-            .map(|d| host_builder.build(&logger, manifest_id.clone(), spec_version.clone(), d))
+            .map(|d| host_builder.build(&logger, manifest_id.clone(), d))
             .partition(|res| res.is_ok());
 
         if !errors.is_empty() {
--- a/core/src/subgraph/instance.rs
+++ b/core/src/subgraph/instance.rs
@@ -1,6 +1,6 @@
 use graph::prelude::{SubgraphInstance as SubgraphInstanceTrait, *};
 use graph::web3::types::{Log, Transaction};
-use semver::Version;
+use semver::{Version, VersionReq};
 
 pub struct SubgraphInstance<T>
 where
@@ -28,6 +28,13 @@ where
         // event processing behavior predictable
         let manifest_id = manifest.id;
         let spec_version = Version::parse(&manifest.spec_version)?;
+        if let false = VersionReq::parse("<= 0.0.2")?.matches(&spec_version) {
+            return Err(format_err!(
+                "Subgraph spec version {} not supported by this Graph Node (v0.0.2)",
+                spec_version
+            ));
+        }
+
         let (hosts, errors): (_, Vec<_>) = manifest
             .data_sources
             .into_iter()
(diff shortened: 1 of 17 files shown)


leodasvacas:
You need to rustfmt for CI.

runtime/wasm/src/module/mod.rs
```diff
@@ -207,23 +210,43 @@ where
...
-            block: EthereumBlockData::from(&self.ctx.block.block),
-            transaction: EthereumTransactionData::from(self.ctx.transaction.deref()),
-            address: log.address,
-            log_index: log.log_index.unwrap_or(U256::zero()),
-            transaction_log_index: log.transaction_log_index.unwrap_or(U256::zero()),
-            log_type: log.log_type.clone(),
-            params,
+        // Decide on the destination type using the version provided
+        // in the subgraph manifest (SpecVersion)
+        let event = if self.host_exports.spec_version > Version::parse("0.0.1").unwrap() {
```
leodasvacas: imo it's more meaningful to check for `>= 0.0.2` than `> 0.0.1`, since then we know the minimum supported version.

To avoid the unwrap, you can use `Version::new(0, 0, 1)`.

runtime/wasm/src/to_from/external.rs
```diff
@@ -290,8 +290,45 @@ impl ToAscObj<AscEthereumTransaction> for EthereumTransactionData {
...
+                .unwrap_or_else(|| AscPtr::null()),
+            value: heap.asc_new(&BigInt::from_unsigned_u256(&self.value)),
+            gas_used: heap.asc_new(&BigInt::from_unsigned_u256(&self.gas_used)),
+            gas_price: heap.asc_new(&BigInt::from_unsigned_u256(&self.gas_price)),
+            input: heap.asc_new(&*self.input.0),
+        }
+    }
+}
+
+impl ToAscObj<AscEthereumEvent<AscEthereumTransactionV1>> for EthereumEventData {
```
leodasvacas: I'd think that we can merge this impl and the one below into one that's generic over the transaction type version.

core/src/subgraph/instance.rs
```diff
@@ -26,10 +27,11 @@ where
         // we use the same order here as in the subgraph manifest to make the
         // event processing behavior predictable
         let manifest_id = manifest.id;
+        let spec_version = Version::parse(&manifest.spec_version)?;
```
leodasvacas: Could we have a constant with the maximum supported version for this node, and check it here? So that in the future we don't have old nodes trying to run newer spec versions.

fordN: I've included a version requirement that ensures the subgraph being deployed is using a spec version equal to or below the current Graph-Node spec version. 

However, I don't love having it buried in `core/src/subgraph/instance`.  It could be added as part of the runtime host builder to be passed into `from_manifest()` and eventually be used to version other aspects of the subgraph instance.

core/src/subgraph/instance.rs
```diff
@@ -28,6 +28,13 @@ where
         // event processing behavior predictable
         let manifest_id = manifest.id;
         let spec_version = Version::parse(&manifest.spec_version)?;
+        if let false = VersionReq::parse("<= 0.0.2")?.matches(&spec_version) {
```
leodasvacas: You can unwrap this, it will never fail. Also `if let false = expr` is just `if !expr` 😄

runtime/wasm/src/asc_abi/class.rs
```diff
@@ -429,13 +429,29 @@ pub(crate) struct AscEthereumTransaction {
 
 #[repr(C)]
 #[derive(AscType)]
-pub(crate) struct AscEthereumEvent {
+pub(crate) struct AscEthereumTransactionV2 {
```
Jannis: I wonder if we should make this `AscEthereumTransaction_0_0_2` or something to reflect the spec version that this data structure was introduced in. (The original can remain `AscEthereumTransaction` since it was introduced at the very beginning.) Thoughts?

leodasvacas: Sounds like a good idea.

leodasvacas:
@Jannis true. I kind of wanted to get away with using the manifest spec version on this one just so that we could land it faster. But the correct thing to do is to introduce mapping spec versions, and then use that here.

runtime/wasm/src/host.rs
```diff
@@ -127,6 +128,15 @@ impl RuntimeHost {
             "data_source" => config.data_source.name.clone(),
         ));
 
+        let api_version = Version::parse(&config.data_source.mapping.api_version)?;
+        if !VersionReq::parse("<= 0.0.2").unwrap().matches(&api_version) {
+            return Err(format_err!(
+                "This Graph Node only supports subgraph manifest specs <= 0.0.2, but subgraph `{}` uses `{}`",
```
Jannis: Now this message isn't correct anymore, it should `mapping API versions` instead of `subgraph manifest specs` 😉

Issue #748, opened 2019-02-13
IPFS Pinning - Designing a Robust and live handling of ipfs `cat`


davekaj:
The current behaviour of `ipfs.cat()` is that it will search for the file for 10 seconds, and then fail, and the graph node crashes. The first and most important thing to deal with here is liveness:
1. We shall allow the graph node to continue operation when `ipfs.cat()` cannot find the file in the allotted time. This is a quirk of IPFS we cannot fix ourselves, sometimes this will inevitably happen (quite often actually in my experience). 
2. The allotted time is up for debate. 10 seconds is too short from my experience building subgraphs. I would say either 30 seconds or 1 minute is appropriate. I'm inclined to go with 30 seconds, since we should really encourage subgraph developers to pre-pin (more on that later).
3. There should be clear logs that this is happening, that there is a 30 second wait, that the file was found, or that it wasnt and it has been skipped. 
4. Pretty sure LEO is working on this now, so that is good. 

Making it robust:
1. I believe Jannis is planning on having an IPFS stream, and block stream. So the IPFS stream should go through the blocks along with the block stream, and then once it is fully synced, it should go back and start searching for the documents that it missed on IPFS. For an extended time. Maybe 2 minutes, and then it should continuously loop through them all and keep trying to get the files until it is satisfied. 

Instructing developers how to create their subgraph, with IPFS's network in mind:
1.   First and foremost, it should be stressed to any subgraph developer that their IPFS files should be pinned to their own node, and then shared on as many public gateways as possible. This is how  they can ensure their files are as easily accessible as possible. 
2. If for some reason this isn't a possibility, then they should write the mappings in mind with the fact that some IPFS files will not be found right away, or may never be found. This means that mapping operations should not fail when data is missing. (i.e. creating an entity based on an IPFS file being found, and then loading that entity with full expectations that the fields will exist and won't be null. This would crash the graph node

leoyvens:
`ipfs.cat` is going to be replaced by IPFS data sources.

PR #749, opened 2019-02-13
Expose ipfs failure to mappings

### LINKED ISSUE #732: Read from IPFS crash the subgraph if hash is invalid/ doesn’t exist
Currently when calling `ipfs.cat` the function will behave as follows:
If the hash is a valid reference to an existing data on IPFS it will return the data.
If the hash queried is invalid or if it points to data which is not stored in the IPFS node it will cause the event listener to crash.

This behavior makes accessing to IPFS reliably practically impossible, as it’s impossible to verify the data exist before sending a query, and in case it indeed doesn’t exist it causes the system to crash.

A possible fix would be to return an empty response in case the query failed.
This would allow to gracefully handle errors and continue the operation of the subgraph.

leoyvens:
Return null if anything goes wrong when fetching a file from IPFS. Also increase the default timeout to 30 seconds and make it configurable through an env var. @davekaj do you want to take this for a spin on a subgraph that needs it as part of the review?

Resolves #732. Part of #748.


davekaj:
Yes, I'll try it out now

davekaj:
It worked! I’m actually thinking differently now though about the default wait time . It should probably be back to 10 seconds. 30 seconds means we can only get through 120 ipfs cats an hour which is going to make this subgraph sync suuuper slow.

what i proposed in my issue was to have the ipfs stream circle back to the missed ipfs cats, and I think at that point it would be good to let it search for a minute , and cycle through them all, and then search again for two minutes on the next cycle, and keep doing that. maybe up to 10 minutes

because right now for all the 30 second waits, nothing is being found :disappointed:

Are you able to change the default back to 10? Then I think we can merge this

leodasvacas:
@davekaj The default doesn't matter much, we can configure this with an env var.

runtime/wasm/src/module/mod.rs
```diff
@@ -447,9 +447,22 @@ where
...
+        let ipfs_res = self.host_exports.ipfs_cat(link);
+        match ipfs_res {
+            Ok(bytes) => {
+                let bytes_obj: AscPtr<Uint8Array> = self.asc_new(&*bytes);
+                Ok(Some(RuntimeValue::from(bytes_obj)))
+            }
+
+            // Return null in case of error.
+            Err(e) => {
+                info!(self.logger, "Failed ipfs.cat, returning `null`";
```
Jannis: I'd be in favor of making this a `warn!`.

Jannis (approved):

runtime/wasm/src/module/mod.rs
leodasvacas: My thinking is that for many subgraphs ipfs hashes that cannot be found are just a fact of life, and not something the developer should be concerned about. It might be worth promoting to warn in case of timeouts, which are highly non-deterministic and you can't tell whether the file is accessible.

leodasvacas: But I'd have to look into IPFS behaviour to know if we can really differentiate between a file that's not found and a file that took too long to fetch. Lets leave this as a follow up after we observe what these logs look like for real subgraphs.

PR #750, opened 2019-02-14
Update Docker Compose setup and docs for an Ethereum node on the host

Jannis:
This removes Parity from the Docker Compose setup and instead defaults to connecting to an Ethereum node running on the host (`host.docker.internal:8545`). This allows developers to spin up Parity or Ganache on the host independently of the Graph Node / IPFS and Postgres.

We can work on adding a default Ganache image to the setup, but I think leaving that to developers keeps things most flexible; i.e., we're not making any assumptions about _how_ users set up Ganache, Parity or geth.


leodasvacas:
I don't have much experience with running docker. @nenadjaja do you wanna take a look?

docker/README.md
```diff
@@ -28,41 +28,32 @@ docker run -it \
 
 ## Docker Compose
 
-To use the `docker-compose.yml` configuration in this repository, simply clone
-this repo and run
+The Docker Compose setup requires an Ethereum network name and node
+to connect to. By default, it will use `dev:http://docker.host.internal:8545`
+in order to connect to an Ethereum node running on your host machine.
```
nenadjaja: I would add "on port 8545" even though it's obvious, so people know they need to use that port (which is default for Ganache, but some ppl change it)

docker/docker-compose.yml
```diff
@@ -14,34 +14,22 @@ services:
       postgres_user: graph-node
       postgres_pass: let-me-in
       postgres_db: graph-node
-      ipfs: "ipfs:5001"
-      ethereum: "dev:http://parity:8545"
+      ipfs: 'ipfs:5001'
+      ethereum: 'dev:http://docker.host.internal:8545'
```
nenadjaja: Just wanted to make sure that we are aware of this issue: https://github.com/docker/for-linux/issues/264

Also check this comment: https://github.com/docker/for-linux/issues/264#issuecomment-391364563

nenadjaja (commented):
I left 2 comments :)

docker/docker-compose.yml
Jannis: Ouch... thanks for pointing that out, I was hoping support for it on Linux had been added by now.

lionseal: There is a fix for this now, just tested it https://github.com/docker/for-linux/issues/264#issuecomment-965465879

I created the following PR https://github.com/graphprotocol/graph-node/pull/3031

PR #752, opened 2019-02-14
Configurable timeout for graphql queries

### LINKED ISSUE #742: Make long quries queries (and subscription queries) time out

leoyvens:
Timeout is checked when a new field is executed. I'll open an issue for returning partial results in case of an error, it's not difficult but lets land this first. Resolves #742.

### DIFF
--- a/graphql/src/execution/execution.rs
+++ b/graphql/src/execution/execution.rs
@@ -4,6 +4,7 @@ use indexmap::IndexMap;
 use std::cmp;
 use std::collections::{BTreeMap, HashMap, HashSet};
 use std::ops::Deref;
+use std::time::Instant;
 
 use graph::prelude::*;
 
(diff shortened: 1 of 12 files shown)


graphql/src/execution/execution.rs
```diff
@@ -20,22 +21,33 @@ where
 {
     /// The logger to use.
     pub logger: Logger,
+
```
Jannis: Hehe, about time someone adds these empty lines here!

graphql/src/subscription/mod.rs
```diff
@@ -182,6 +183,7 @@ where
         document: &document,
         fields: vec![],
         variable_values,
+        deadline: None,
```
Jannis: Let's initialize the deadline here as well, because this is where the subscription queries get re-executed on changes. Would be good to have them time out and send errors back if subscription queries take too long.

graphql/tests/query.rs
```diff
@@ -691,3 +692,26 @@ fn include_directive_works_with_query_variables() {
...
+
+    let options = QueryExecutionOptions {
+        logger: logger,
+        resolver: store_resolver,
+        deadline: Some(Instant::now()),
+    };
+
+    match execute_query(&query, options).errors.unwrap()[0] {
+        QueryError::ExecutionError(QueryExecutionError::Timeout) => (), // Expected
+        _ => panic!("did not timeout"),
```
Jannis: The verb is `time out` 😉

Jannis (changes requested):
Great! Can we use this for subscription queries as well? That would be perfect.

Jannis (approved):
Great!

PR #754, opened 2019-02-16
Throttle subscription updates

lutter:
This PR throttles subscription updates so that updates are generated at most every 500ms; at the same time, no `StoreEvent` is held for longer than 500ms.

There are two things that need improvement still:

* `throttle_while_syncing` now queries the `synced` status of a deployment every time a `StoreEvent` comes through. One way to avoid that would be to get the current chain head pointer when throttling is started, and compare it to the source for any event coming through. We would only requery the head pointer once we see event sources newer than the head pointer. We would then switch to querying the deployment status once we see the head pointer advancing very little when we requery (say by 100 blocks) and then switch to querying the deployment status.
* the throttle time for subscriptions is hardcoded to 500ms. We could expose that as an env var and make it configurable that way (Which brings up the tough issue of naming that env var)

I'd appreciate comments on these two things (and on the PR in general, of course)

### DIFF
--- a/graph/src/components/store.rs
+++ b/graph/src/components/store.rs
@@ -4,7 +4,9 @@ use futures::Future;
 use futures::Stream;
 use futures::{Async, Poll};
 use std::collections::HashSet;
+use std::env;
 use std::fmt;
+use std::str::FromStr;
 use std::sync::atomic::{AtomicUsize, Ordering};
 use std::time::{Duration, Instant};
 use web3::types::H256;
@@ -12,6 +14,17 @@ use web3::types::H256;
 use data::store::*;
 use data::subgraph::schema::*;
 use prelude::*;
+
+lazy_static! {
+    pub static ref SUBSCRIPTION_THROTTLE_INTERVAL: Duration =
+        env::var("SUBSCRIPTION_THROTTLE_INTERVAL")
+            .ok()
+            .map(|s| u64::from_str(&s).unwrap_or_else(|_| panic!(
+                "failed to parse env var SUBSCRIPTION_THROTTLE_INTERVAL"
+            )))
+            .map(|millis| Duration::from_millis(millis))
+            .unwrap_or(Duration::from_millis(1000));
+}
 
 /// Key by which an individual entity in the store can be accessed.
 #[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
--- a/graph/src/components/store.rs
+++ b/graph/src/components/store.rs
@@ -149,7 +149,7 @@ impl EntityQuery {
 }
 
 /// Operation types that lead to entity changes.
-#[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq, Hash)]
 #[serde(rename_all = "lowercase")]
 pub enum EntityChangeOperation {
     /// An entity was added or updated
@@ -159,7 +159,7 @@ pub enum EntityChangeOperation {
 }
 
 /// Entity change events emitted by [Store](trait.Store.html) implementations.
-#[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq, Hash)]
 pub struct EntityChange {
     /// ID of the subgraph the changed entity belongs to.
     pub subgraph_id: SubgraphDeploymentId,
@@ -214,7 +214,7 @@ pub struct StoreEvent {
     // The tag is only there to make it easier to track StoreEvents in the
     // logs as they flow through the system
     pub tag: usize,
-    pub changes: Vec<EntityChange>,
+    pub changes: HashSet<EntityChange>,
 }
 
 impl StoreEvent {
@@ -222,6 +222,7 @@ impl StoreEvent {
         static NEXT_TAG: AtomicUsize = AtomicUsize::new(0);
 
         let tag = NEXT_TAG.fetch_add(1, Ordering::Relaxed);
+        let changes = changes.into_iter().collect();
         StoreEvent { tag, changes }
     }
 
--- a/graph/src/components/store.rs
+++ b/graph/src/components/store.rs
@@ -3,6 +3,7 @@ use futures::Future;
 use futures::Stream;
 use std::collections::HashSet;
 use std::fmt;
+use std::sync::atomic::{AtomicUsize, Ordering};
 use web3::types::H256;
 
 use data::store::*;
@@ -184,7 +185,7 @@ impl EntityChange {
     }
 }
 
-#[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
+#[derive(Clone, Debug, Serialize, Deserialize)]
 /// The store emits `StoreEvents` to indicate that some entities have changed.
 /// For block-related data, at most one `StoreEvent` is emitted for each block
 /// that is processed. The `changes` vector contains the details of what changes
@@ -194,8 +195,37 @@ impl EntityChange {
 /// any specific blocks, `StoreEvents` for it are generated as soon as they are
 /// written to the store.
 pub struct StoreEvent {
-    pub source: EventSource,
+    // The tag is only there to make it easier to track StoreEvents in the
+    // logs as they flow through the system
+    pub tag: usize,
     pub changes: Vec<EntityChange>,
+}
+
+impl StoreEvent {
+    pub fn new(changes: Vec<EntityChange>) -> StoreEvent {
+        static NEXT_TAG: AtomicUsize = AtomicUsize::new(0);
+
+        let tag = NEXT_TAG.fetch_add(1, Ordering::Relaxed);
+        StoreEvent { tag, changes }
+    }
+}
+
+impl fmt::Display for StoreEvent {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(
+            f,
+            "StoreEvent[{}](changes: {})",
+            self.tag,
+            self.changes.len()
+        )
+    }
+}
+
+impl PartialEq for StoreEvent {
+    fn eq(&self, other: &StoreEvent) -> bool {
+        // Ignore tag for equality
+        self.changes == other.changes
+    }
 }
 
 /// A boxed `StoreEventStream`
--- a/graph/src/components/store.rs
+++ b/graph/src/components/store.rs
@@ -1,9 +1,12 @@
 use failure::Error;
+use futures::stream::poll_fn;
 use futures::Future;
 use futures::Stream;
+use futures::{Async, Poll};
 use std::collections::HashSet;
 use std::fmt;
 use std::sync::atomic::{AtomicUsize, Ordering};
+use std::time::{Duration, Instant};
 use web3::types::H256;
 
 use data::store::*;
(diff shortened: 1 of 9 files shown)


lutter:
I changed things so that the `synced` flag in `throttle_while_syncing` is cached for `4  * SUBSCRIPTION_THROTTLE_INTERVAL` since it otherwise requires a DB query every time we receive a `StoreEvent`.

The throttle interval is configurable through the `SUBSCRIPTION_THROTTLE_INTERVAL` env var (in ms), and defaults to 1000 ms.

Also rebased to latest master.

leodasvacas:
Reviewing this.

graph/src/components/store.rs
```diff
@@ -245,7 +276,111 @@ where
...
+            }
+
+            if synced {
+                return source.poll();
+            }
+
+            // Check if interval has passed since the last time we sent something.
+            // If it has, start a new delay timer
+            let ready = match delay.poll() {
+                Ok(Async::NotReady) => false,
```
leodasvacas: Why have the `ready` value instead of immediately returning `NotReady` in this case?

graph/src/components/store.rs
```diff
@@ -245,7 +276,111 @@ where
...
+                Ok(Async::Ready(())) => {
+                    delay = tokio_timer::Delay::new(Instant::now() + interval);
+                    true
+                }
+                Err(_) => {
+                    // If there was an error, return whatever we have immediately. If we don't
+                    // have anything, pretend the timer did expire
+                    if pending_event.is_some() {
+                        had_err = true;
```
leodasvacas: We should not propagate timer errors since those are harmless. I'd just make this identical to the case where the timer is ready.

graph/src/components/store.rs
```diff
@@ -245,7 +276,111 @@ where
...
+
+        let source = Box::new(poll_fn(move || -> Poll<Option<StoreEvent>, ()> {
+            if had_err {
+                // We had an error the last time through, but returned the pending
+                // event first. Indicate the error now
+                had_err = false;
+                return Err(());
+            }
+
+            if !synced && synced_last_refreshed + synced_check_interval < Instant::now() {
```
leodasvacas: `synced_last_refreshed + synced_check_interval < Instant::now()` can be rewritten as `synced_last_refreshed.elapsed() < synced_check_interval`.

graph/src/components/store.rs
```diff
@@ -245,7 +276,111 @@ where
...
+        // avoid hitting the database too often to see if the subgraph has
+        // been synced in the meantime. The only downside of this approach is
+        // that we might continue throttling subscription updates for a little
+        // bit longer than we really should
+        static SYNC_REFRESH_FREQ: u32 = 4;
+
+        let mut synced = false;
+        let synced_check_interval = interval.checked_mul(SYNC_REFRESH_FREQ).unwrap();
+        let mut synced_last_refreshed =
+            Instant::now() - synced_check_interval - synced_check_interval;
```
leodasvacas: If we instantiate `synced` above with the store value, then this can be just `Instant::now()`.

graph/src/components/store.rs
```diff
@@ -245,7 +276,111 @@ where
...
+    ) -> StoreEventStreamBox {
+        // We refresh the synced flag every SYNC_REFRESH_FREQ*interval to
+        // avoid hitting the database too often to see if the subgraph has
+        // been synced in the meantime. The only downside of this approach is
+        // that we might continue throttling subscription updates for a little
+        // bit longer than we really should
+        static SYNC_REFRESH_FREQ: u32 = 4;
+
+        let mut synced = false;
+        let synced_check_interval = interval.checked_mul(SYNC_REFRESH_FREQ).unwrap();
```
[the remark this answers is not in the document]
lutter: Do you mean the `unwrap` should go away? It's highly unlikely that we'd get to an overflow here since that would require setting `interval` to something insanely large.

core/src/subgraph/registrar.rs
```diff
@@ -117,6 +117,11 @@ where
             .subscribe(vec![
                 SubgraphDeploymentAssignmentEntity::subgraph_entity_pair(),
             ])
+            .throttle_while_syncing(
```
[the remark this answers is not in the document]
lutter: Ah, yes, you are right.

graph/src/components/store.rs
```diff
@@ -198,15 +214,33 @@ pub struct StoreEvent {
...
+    fn accumulate(ev1: &mut Option<StoreEvent>, ev2: StoreEvent) {
+        if let Some(e) = ev1 {
+            e.extend(ev2);
+        } else {
+            ev1.replace(ev2);
+        }
+    }
+
+    fn extend(&mut self, other: StoreEvent) {
+        if other.source > self.source {
```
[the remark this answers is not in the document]
lutter: Initially I wanted to use `source` to convey to subscribers "we've made it this far in the chain" so they could detect things like block reversion. But as you said, since it's not used right now, I'll take it out.

graph/src/components/store.rs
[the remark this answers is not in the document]
leodasvacas: Nothing to change, it's great that you're being mindful of this.

graph/src/components/store.rs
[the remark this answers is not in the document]
lutter: The one thing for which source is useful is tracing events as they go through the system in logs. I added a comment to the declaration of `StoreEvent` to that effect.

graph/src/components/ethereum/types.rs
```diff
@@ -147,10 +147,10 @@ impl Clone for EthereumEventData {
 /// A block hash and block number from a specific Ethereum block.
 ///
 /// Maximum block number supported: 2^63 - 1
-#[derive(Clone, Copy, Debug, Serialize, Deserialize, PartialEq, Eq)]
+#[derive(Clone, Copy, Debug, Serialize, Deserialize, PartialEq, Eq, PartialOrd)]
 pub struct EthereumBlockPointer {
-    pub hash: H256,
     pub number: u64,
+    pub hash: H256,
 }
```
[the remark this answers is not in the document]
lutter: See my comment below on why `StoreEvent` has a `source`.

graph/src/components/store.rs
```diff
@@ -1,14 +1,30 @@
...
 use prelude::*;
 
+lazy_static! {
+    pub static ref SUBSCRIPTION_THROTTLE_INTERVAL: Duration =
+        env::var("SUBSCRIPTION_THROTTLE_INTERVAL")
+            .ok()
+            .map(|s| u64::from_str(&s).unwrap_or_else(|_| panic!(
+                "failed to parse env var SUBSCRIPTION_THROTTLE_INTERVAL"
+            )))
+            .map(|millis| Duration::from_micros(millis))
```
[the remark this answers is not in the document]
lutter: D'oh

graph/src/components/store.rs
```diff
@@ -245,7 +276,111 @@ where
...
+        interval: Duration,
+    ) -> StoreEventStreamBox {
+        // We refresh the synced flag every SYNC_REFRESH_FREQ*interval to
+        // avoid hitting the database too often to see if the subgraph has
+        // been synced in the meantime. The only downside of this approach is
+        // that we might continue throttling subscription updates for a little
+        // bit longer than we really should
+        static SYNC_REFRESH_FREQ: u32 = 4;
+
+        let mut synced = false;
```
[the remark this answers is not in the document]
lutter: Good point. Doing that.

graph/src/components/store.rs
lutter: Because I also want to make progress on getting `StoreEvents` so they don't pile up in the queue for the entire interval. Basically, this is so that `poll` makes as much progress as it possibly can before returning `Async::NotReady`.

graph/src/components/store.rs
lutter: Ok.

graph/src/components/store.rs
lutter: Ok.

graph/src/components/store.rs
lutter: Nice. But I think that should be a `>` then ;)

lutter:
Addressed all the review comments.

graph/src/components/store.rs
Jannis: That part is the only one that I feel is not obvious. Perhaps a longer comment about what's going on there would be good. Otherwise: nice job!

Jannis: And maybe `ready` is not the ideal name for the variable. `throttling_interval_elapsed` maybe?

Jannis: I see there actually is a comment at the top of the function, so only consider the variable naming comment. 😄

graph/src/components/store.rs
```diff
@@ -194,19 +210,39 @@ impl EntityChange {
...
+    fn accumulate(ev1: &mut Option<StoreEvent>, ev2: StoreEvent) {
+        if let Some(e) = ev1 {
+            e.extend(ev2);
+        } else {
+            ev1.replace(ev2);
+        }
+    }
+
+    fn extend(&mut self, other: StoreEvent) {
+        if other.source > self.source {
```
leodasvacas: Considering this is just for logging, can we not have this comparison and just set `self.source = other.source` unconditionally? Then we can avoid having to define `PartialOrd` at all for block pointers.

lutter: As discussed in Discord, I ripped out the `PartialOrd` stuff for `EventSource`, and removed `source` from `StoreEvent`; it now has a `tag: usize` to help distinguish events in the logs to make it possible to trace where events that were sent are received.

store/postgres/tests/store.rs
```diff
@@ -32,15 +32,16 @@ fn postgres_test_url() -> String {
 }
 
 lazy_static! {
+    static ref LOGGER:Logger = {match env::var_os("GRAPH_LOG") {
```
Jannis: Something's up with the formatting here. It's only tests, so it may not affect the `rustfmt` check on Travis?

lutter: Yeah, that's pretty strange, and neither `cargo fmt` nor `rustfmt` on the file itself seem to have an opinion on it. I fixed it manually (and rebased to latest master)

Jannis: 👍

graph/src/components/store.rs
```diff
@@ -194,19 +211,60 @@ impl EntityChange {
...
+    }
+
+    /// Extend `ev1` with `ev2`. If `ev1` is `None`, just set it to `ev2`
+    fn accumulate(logger: &Logger, ev1: &mut Option<StoreEvent>, ev2: StoreEvent) {
+        if let Some(e) = ev1 {
+            trace!(logger, "Adding changes to event";
+                           "from" => ev2.tag, "to" => e.tag);
+            e.changes.extend(ev2.changes);
+        } else {
+            ev1.replace(ev2);
```
leodasvacas: This is just `ev1 = Some(ev2)`.

lutter: I can't find a way to do that: just `ev1 = Some(ev2)` makes rustc complain because `ev1` is `&mut`, and `ev1 = &mut Some(ev2)` makes it complain that it can't assign twice to `ev1` and that the reference doesn't live long enough. This really needs to change the contents of `ev1` in the caller as that's what we are accumulating into. And returning the combination of `ev1` and `ev2` to the caller isn't desirable since `changes` can become a long list that we'd need to copy every time we combine events.

lutter: Fixed it with `*ev1 = Some(ev2)` as you suggested in Discord.

graph/src/components/store.rs
```diff
@@ -194,19 +211,60 @@ impl EntityChange {
...
+    // logs as they flow through the system
+    pub tag: usize,
+    pub changes: HashSet<EntityChange>,
 }
 
-/// A boxed `StoreEventStream`
-pub type StoreEventStreamBox = Box<Stream<Item = StoreEvent, Error = ()> + Send>;
+impl StoreEvent {
+    pub fn new(changes: Vec<EntityChange>) -> StoreEvent {
+        static NEXT_TAG: AtomicUsize = AtomicUsize::new(0);
```
leodasvacas: Nice way of getting a simple counter.

PR #761, opened 2019-02-23
Fix 'Invalid subgraph hash in assignment entity' error on initial deployments

Jannis:
This was caused by handling all entity changes in the store events emitted for the `SubgraphDeploymentAssignment` subscription. Prior to store events, this subscription would only include changes to `SubgraphDeploymentAssignment` entities. Store events however may contain entity changes for multiple entities of different types, not just the ones subscribed to.

In order to handle assignment changes correctly, we need to filter changes to match `SubgraphDeploymentAssignment` entities.

Without this fix, the first deployment for any subgraph fails with the following kind of error:
```rust
 Feb 23 22:05:15.555 CRIT Assignment event stream failed: subgraph provider error: Invalid subgraph hash in assignment entity: EntityChange {
    subgraph_id: SubgraphDeploymentId(
        "subgraphs"
    ),
    entity_type: "EthereumContractSource",
    entity_id: "QmZwDxfPKJQsFTtXBiDmdjqbX4i4moB9xV92o9je1canvf-manifest-data-source-0-source",
    operation: Set
}
```

### DIFF
--- a/core/src/subgraph/registrar.rs
+++ b/core/src/subgraph/registrar.rs
@@ -99,8 +99,8 @@ where
                     .map_err(move |e| match e {
                         CancelableError::Cancel => {}
                         CancelableError::Error(e) => {
-                            error!(logger_clone2, "assignment event stream failed: {}", e);
-                            panic!("assignment event stream error: {}", e);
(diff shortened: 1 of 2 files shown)


lutter (approved):
Looks good to me!

PR #775, opened 2019-02-26
Cleanup indexes: remove subgraph entity indexes for unassigned subgraphs

fordN:
Many of the subgraphs stored in the database are not the current version of the deployment, so their subgraph entity indexes are an unnecessary load on the `entities` table.  The shear number of indexes has lead to shared memory issues during select statements that create a lock for each index on the `entities` table.  

This PR creates and uses a subgraph indexes cleanup function that:
  - finds all subgraph deployments in the database without an assignment, 
  - gets a list of their subgraph attribute indexes on the entities table, 
  - and drops each of the indexes.

### DIFF
--- a/store/postgres/migrations/2019-02-26-182914_remove_unassigned_deployment_indexes_procedure/up.sql
+++ b/store/postgres/migrations/2019-02-26-182914_remove_unassigned_deployment_indexes_procedure/up.sql
@@ -1,0 +1,43 @@
+-- Find all subgraphs without deployment assignments and remove their indexes
+CREATE OR REPLACE FUNCTION remove_unassigned_deployment_indexes()
+    RETURNS VOID AS
+$$
+DECLARE
+    index_to_drop RECORD;
+    counter INT := 0;
+BEGIN
+    -- Loop through each index
+    FOR index_to_drop IN
+        -- Get all indexes on subgraphs without deployment assignments
+        SELECT
+          indexname as name
+        from pg_indexes
+        where
+          -- Extract the subgraph hash from the index name (subgraph hashes are 46 characters long)
+          left(indexname,46) IN
+          (
+            -- Format the subgraph id for comparison with the index names
+            select left(lower(deployments.id),46)
+            from entities as deployments
+            left join
+              (
+                select
+                  id, entity
+                from entities
+                where
+                  subgraph='subgraphs' and
+                  entity='SubgraphDeploymentAssignment'
+              ) assigments
+            on deployments.id=assigments.id
+            where
+              deployments.subgraph='subgraphs' and
+              deployments.entity='SubgraphDeployment' and
+              assigments.id is null)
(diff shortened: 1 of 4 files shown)


store/postgres/migrations/2019-02-26-182914_remove_unassigned_deployment_indexes_procedure/up.sql
```diff
@@ -0,0 +1,41 @@
...
+    counter INT := 0;
+BEGIN
+    -- Loop through each index
+    FOR index_to_drop IN
+        -- Get all indexes on subgraphs without deployment assignments
+        SELECT
+          indexname as name
+        from pg_indexes
+        where
+          left(indexname,46) IN
```
lutter: It would be good to have a comment where the magic `46` comes from. I tripped over that reading through it.

store/postgres/migrations/2019-02-26-182914_remove_unassigned_deployment_indexes_procedure/up.sql
```diff
@@ -0,0 +1,41 @@
...
+                from entities
+                where
+                  subgraph='subgraphs' and
+                  entity='SubgraphDeploymentAssignment'
+              ) assigments
+            on deployments.id=assigments.id
+            where
+              deployments.subgraph='subgraphs' and
+              deployments.entity='SubgraphDeployment' and
+              assigments.id is null)
```
lutter: I am not clear on what can go wrong if assignments are being modified concurrently, like is it possible that we are in the process of setting up an assignment, have already created the indexes, but not updated the assignments. Do we need to lock the assignments rows `for update` while we do this ?

lutter: Scratch the thing about 'assignments rows', there's nothing to lock there. If assignments was its own table, I'd take an exclusive lock on the table to bar any concurrent modifications from it; but taking an exclusive lock on entities might be a bit too much.

fordN: Good point, I'll do some research into the `SubgraphDeploymentAssignment`operations to ensure that we don't delete indexes for a subgraph that is in the process of being deployed.

fordN: @lutter It looks like there is no risk of deleting a subgraph's indexes before it is given a deployment assignment because the assignment happens before the subgraph is started and its indexes are created.  The indexes are built in `provider.start()` which is called using an existing `SubgraphDeploymentAssignment` (if the query fails to get an existing assignment an error is triggered.)  https://github.com/graphprotocol/graph-node/blob/master/core/src/subgraph/registrar.rs#L194-L215

Issue #759, opened 2019-02-23
Subgraph remains in 'failed' state after restart, should go back to 'syncing'


Jannis:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current and desired behavior?**

When a subgraph fails due to a node error such as
```
Subgraph instance failed to run: Error while processing block stream
for a subgraph: store error: Failed to set entity (subgraphs, SubgraphDeployment,
QmUqmv9NGSewTLVwbaj1ikGU6dS4gJpvkxYMhXLrjgvxj3): out of shared memory, id:
QmUqmv9NGSewTLVwbaj1ikGU6dS4gJpvkxYMhXLrjgvxj3
```
it should pick back up after restarting the node. This actually happens but the `failed` state of the subgraph isn't reset to `syncing` properly.

SuperDev123-hub:
Subgraph syncing failed on mumbai network.
What should I do now?

PR #777, opened 2019-02-26
core/subgraph: mark subgraphs as not failed when starting them

lutter:
Fixes https://github.com/graphprotocol/graph-node/issues/759

### DIFF
--- a/core/src/subgraph/instance_manager.rs
+++ b/core/src/subgraph/instance_manager.rs
@@ -156,6 +156,13 @@ impl SubgraphInstanceManager {
         let block_logger = logger.clone();
         let error_logger = logger.clone();
 
+        // Clear the 'failed' state of the subgraph. We were told explicitly
+        // to start, which implies we assume the subgraph has not failed (yet)
+        // If we can't even clear the 'failed' flag, don't try to start
+        // the subgraph.
+        let status_ops = SubgraphDeploymentEntity::update_failed_operations(&id_for_err, false);
+        store_for_errors.apply_entity_operations(status_ops, EventSource::None)?;
+
         // Forward block stream events to the subgraph for processing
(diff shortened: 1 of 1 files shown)


Jannis:
@lutter Have we tested that this works? I just want to make sure. It looks good!

core/src/subgraph/instance_manager.rs
```diff
@@ -156,6 +156,17 @@ impl SubgraphInstanceManager {
...
+        // Clear the 'failed' state of the subgraph. We were told explicitly
+        // to start, which implies we assume the subgraph has not failed (yet)
+        let status_ops = SubgraphDeploymentEntity::update_failed_operations(&id_for_err, false);
+        if let Err(e) = store_for_errors.apply_entity_operations(status_ops, EventSource::None) {
+            error!(
+                error_logger,
+                "Failed to clear 'failed' status of subgraph: {}", e;
+                "id" => id_for_err.to_string()
+            );
+        }
```
Jannis: One thing I wonder is if we should even continue starting the subgraph if clearing the failed status fails. I'm thinking not.

lutter:
@Jannis I tested it locally by manually flipping a subgraph to failed, and then watching what it does. The only difference with before seems to be that the failed flag is cleared. In both cases, the subgraph was happily ingesting blocks, regardless of the failed flag.

core/src/subgraph/instance_manager.rs
lutter: I agree; I changed things so that the error is just returned back to the caller in `SubgraphInstanceManager::handle_subgraph_events` which will log it.

Jannis:
@lutter I'll go ahead and merge this for you, so we can get it out there quickly. Thanks!

PR #780, opened 2019-02-27
Pass large Postgres notifications through a dedicated table

lutter:
We create a large_notifications table. Notification payloads on the
store_events channel bigger than 8000 bytes are put into this table, and
the id of the entry in that table is sent instead.

The NotificationListener is responsible for transparently unpacking
payloads that go through this table.

### DIFF
--- a/store/postgres/migrations/2019-02-27-035443_create_large_notifications/down.sql
+++ b/store/postgres/migrations/2019-02-27-035443_create_large_notifications/down.sql
@@ -1,0 +1,1 @@
+drop table large_notifications;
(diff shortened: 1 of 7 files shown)


lutter:
@Jannis this is what I have for the notification issue. I have only written the code, it compiles, but I haven't run tests or tried it at all.

I realized that it is easier to do the unpacking of large notifications in the NotificationListener, rather than farther up the stack, as the listener already has a DB connection, while that is harder to get in other places, like the StoreEventListener.

It's a tiny bit wonky that a payload starting with `<` means "unpack that from the large_notifications table", but since we generally send JSON payloads, this should be safe.

Jannis:
@lutter Cool, I've continued working on this a bit. Happy to sync up later to get this ready.

Jannis (approved):

lutter:
Rebased to latest master, and slightly reorganized the patches; no functional changes.

PR #781, opened 2019-02-28
Enable building against OpenSSL 1.1.1

lutter:
So far, it was not possible to build graph-node against openssl-1.1.1. Unfortunately, my recent upgrade to Fedora 29, and blowing away `target/debug` necessitated fixing this.

Ultimately, the problem with updating our openssl crate was that it also required upgrading serde_json, which ran into [this problem](https://github.com/serde-rs/json/issues/520)

Changing the value for `graph::data::store::Value::Float` from `f32` to `f64` gets around the serde_json regression, but it unfortunately touches quite a few places. On the upside, we now use `f64` for floating point numbers in more, if not all, places.

@leodasvacas since there are changes to WASM whose repercussions I don't fully understand, can you have a look at that and make sure it's not going to break anything?


runtime/wasm/wasm_test/Makefile
```diff
@@ -0,0 +1,7 @@
+TS_FILES=$(wildcard *.ts)
+WASM_FILES=$(patsubst %.ts,%.wasm,$(TS_FILES))
+
+all: $(WASM_FILES)
+
+%.wasm: %.ts
+	@asc $< -b $@ --validate
```
leodasvacas: Nice! I can't read this but I also can't read the previous script which I wrote myself, and this uses make so it must be much smarter.

leodasvacas:
serde_json was super quick to fix the issue and they've already released 1.0.39 with the fix, which makes these changes unecessary, but still a good refactoring I guess

runtime/wasm/wasm_test/Makefile
Jannis: Haha

Jannis: I don't like using `make` here though. It requires that you have it installed in the first place, whereas the previous script just required `bash` and was super obvious.

Also, aren't there different variants of make that may or may not work with the Makefile features we're using here? I'd rather keep it simple.

leodasvacas: I do find it convenient to rebuild only for changed inputs, compiling these files is fast but not instant.

lutter: That (and a bit of reflex) was the reason why I changed it to a Makefile. I can go either way on this, but this would only require make on developer machines, and only if people change this test data, at which point you're pretty far down the rabbit hole already.

lutter:
> The `f32` -> `f64` change makes sense and bumping dependencies is also good, especially if it makes building the node on some versions of Linux work again.

The big thing here (and the reason why I looked into this) was that it allows us to update to a newer openssl version; those will hit other distros/OS's sooner or later anyway. And not being able to update openssl makes me super nervous as it is a bit of a CVE magnet.

runtime/wasm/wasm_test/Makefile
Jannis: Ok, I get that. Carry on :)

lutter: > I forgot how crude the Makefile syntax is.
> 
> `all: $(WASM_FILES)` says "when running `make`, build all matched `.wasm` files".
> 
> `%.wasm: %.ts` says "when building a `.wasm` target, use the `.ts` file with the same name as the input.
> 
> `@asc` says "run `asc` command in a shell.
> 
> `$<` is a placeholder for the target name, e.g. `foo.wasm`.
> `$@` is a placeholder for the input name, e.g. `foo.ts`.

Almost ;) `@asc` says "run `asc` but do not print the command you're running" (without the `@` it would print)

`$<` are the dependencies of the target (everything to the right of the `:`) and `$@` is the actual target (to the right of `:`)

Issue #785, opened 2019-02-28
Investigate and improve slow parent_hash query


lutter:
The stored procedure `attempt_chain_head_update` contains a query that does two full table scans over `ethereum_blocks`. That query currently takes ~ 4s in production and needs to be sped up.

Details about the problem in [this gist](https://gist.github.com/lutter/5ebc0689b9369361ec29bd8591fd4785)

lutter:
I've looked a little more into what is happening in `attempt_chain_head_update`, and there's actually two slow queries in  there: the one on line 25 that tries to find a new chain head candidate (~
 780ms in production),  and the one on line 51 that checks if we have the parents of all blocks within `ancestor_count` of that head candidate (~ 2400ms in production).

Both of them can be improved by realizing that we only need to deal with the first  ~`ancestor_count` blocks, restricting the queries appropriately, and supporting that restriction by creating an index on `ethereum_blocks(network_name, number)`.

The query on line 25 (and including the query from line 19) then becomes
```sql
SELECT
   hash,
   number
FROM ethereum_blocks b, ethereum_networks n
WHERE b.network_name = net_name
  and n.name = net_name
  and b.number > n.head_block_number
ORDER BY
   number DESC,
   hash ASC
LIMIT 1;
```

The query on line 51 becomes
```sql
with head as
(select hash,
        parent_hash,
        (number = (new_head_number - ancestor_count)) as last
   from ethereum_blocks
  where network_name = net_name
    and number >= GREATEST(new_head_number - ancestor_count, 0))
select array_agg(h1.parent_hash)
  from head h1
 where not h1.last
   and not exists (select 1 from head h2 where h1.parent_hash = h2.hash);
```

The first query now takes 0.1ms and the second query takes 4ms on a clone of the production database, where they previously took 700ms and 1600ms.

PR #786, opened 2019-02-28
store/postgres: optimize stored proc attempt_chain_head_update

lutter:
The queries in that stored proc caused several full table scans on ethereum_blocks, which becomes very costly when the database contains a lot of blocks.

The new queries use an index on `ethereum_blocks(network_name, number)` to make sure we only ever access the last `ancestor_count` rows in `ethereum_blocks`, yielding a very significant speedup.

Fixes https://github.com/graphprotocol/graph-node/issues/785

### DIFF
--- a/store/postgres/migrations/2019-02-28-190800_update_chain_head_update_procedure/up.sql
+++ b/store/postgres/migrations/2019-02-28-190800_update_chain_head_update_procedure/up.sql
@@ -1,0 +1,83 @@
+CREATE INDEX ethereum_blocks_name_number ON ethereum_blocks(network_name, number);
+
+/**************************************************************
+* Replace this function with a version whose queries are
+* a lot faster.
+**************************************************************/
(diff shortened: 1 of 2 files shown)


store/postgres/migrations/2019-02-28-190800_update_chain_head_update_procedure/up.sql
```diff
@@ -0,0 +1,83 @@
...
+        (SELECT hash,
+                parent_hash,
+                (number = (new_head_number - ancestor_count)) as last
+           FROM ethereum_blocks
+          WHERE network_name = net_name
+            AND number >= GREATEST(new_head_number - ancestor_count, 0))
+    SELECT array_agg(h1.parent_hash)
+      INTO STRICT missing_parents
+      FROM head h1
+     WHERE not h1.last
```
fordN: Clever to not use the > filter here.

PR #792, opened 2019-03-03
travis: reduce the amount of build data we cache

lutter:
The target/debug/ directory is roughly 3.3G big; by removing the build
artifacts for graph-node itself, and especially the test binaries, we can
reduce the size of the cache to 1.6G, and therefore speed up how long
downloading and updating the build cache takes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,6 +46,8 @@ matrix:
 cache:
   cargo: true
   timeout: 1000
+before_cache:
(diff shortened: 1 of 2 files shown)


lutter:
Good point. I'll make that `rm -f deps/*graph[_-]*` which seems just as effective, but avoids deleting the `graphql_parser`.

PR #800, opened 2019-03-04
attempt_chain_head_update SQL function gets stuck in infinite loop

olibearo:
Infinite loop happens only when starting from a fresh graph node, fix the function to default the
number column to 0 when NULL


store/postgres/migrations/2019-03-04-235349_empty-ethereum-network-chain-head-update-fix/up.sql
```diff
@@ -0,0 +1,80 @@
+/**********************************************************************
+* Handle the case where ethereum_networks has a NULL head_block_number
```
lutter: This comment would be better as a comment on the query that does that on line 14. The proc does more than what this comment implies (e.g., check for missing parents)

Issue #807, opened 2019-03-07
Improve how we maintain entity_history


lutter:
I just had a look at the production dump I have, trying to understand in more detail how `entity_history` works, and came across a bunch of things that could be greatly improved.

The dump has 10M rows in the `entities` table, and 39M rows in the `entity_history` table. The latter takes up 41GB of storage. For comparison, `ethereum_blocks` is 104 GB.

* there are 22M history entries for the `subgraphs` subgraph, even though we explicitly exclude it, and never use these entries during reversion
* there are 5M entries for other subgraphs where `data_before = data_after`
* there are 10M entries for other subgraphs where `data_before` is null, i.e. they just record the creation of the entity that's also in the `entities` table.
* we never use `data_after` when we revert, so roughly half the size of the table is never used

If we remove the 27M entries from the first two points, the `entity_history` table would shrink to about 13GB. If we then also remove the unused `data_after` column, the table would shrink to about 7GB. (These are guesses, not actual measurements)

Finally, if we used a scheme that doesn't force us to maintain identical copies of entities in `entities` and `entity_history` for entities that are never updated, we'd be left with 2.2M 'old' entity versions, which would roughly take 1.2GB to store. (That also means that only about 20% of entities are ever changed again after creation)

There are two aspects to this: one is just the sheer amount of storage we need for history, and the other is the number of rows in `entity_history`, which slows down any operation involving history. Aggravating this is that most entities are relatively small, and Postgres stores them in the actual table rather than in toast (large object) storage so that things like full table scans actually have to wade through the entire ~ 40GB of storage. Contrast that with `ethereum_blocks`, where the actual table only takes up about 500MB.

fordN:
I think it's a good idea to remove the `data_after` column to get the significant data volume reduction you noted.   The information is still available by simply scanning to the next `data_before`.  The only thing we really lose out on (which hasn't been used much anyways) is the ability to easily query for the changes that were effected by a specific block: saved queries and views could be created to facilitate peering into an entities history of changes.    

For the case where `data_before`=`data_after` I wonder how much we can reduce the load on the DB by going further upstream and not even apply these transactions to the `entities` table in the first place. We could add a `BEFORE` trigger `FOR EACH ROW` that skips the operation if `data_before`=`data_after`.

lutter:
@fordN agreed on dropping `data_after`; and yes, I also realized that I should rework the triggers to have `WHEN` conditions to avoid going into plsql at all for those rows.

[...]

PR #808, opened 2019-03-08
Improve how we record and handle entity_history

lutter:
The main focus of this change is to stop recording entity_history for the
'subgraphs' subgraph, and to avoid recording history for changes that
actually do not change the entity. The history we recorded for these made
up about 75% of the storage needed in the hosted offering, and we never
need the data thus recorded.

This change also simplifies how the revert_block stored procedure works by
basing it off a single history query, and removes unused input parameters
for it.

The change also gets rid of a few stored procedures that were previously
used, but are not any longer.

More background on these changes can be found at
https://github.com/graphprotocol/graph-node/issues/807

### DIFF
--- a/store/postgres/migrations/2019-03-07-171355_update_history_storage/up.sql
+++ b/store/postgres/migrations/2019-03-07-171355_update_history_storage/up.sql
@@ -1,0 +1,107 @@
+-- Change log_entity_event such that we do not record history entries for
+-- the 'subgraphs' subgraph and for operations that don't actually change
+-- the data of an entity
+drop trigger if exists after_entity_change_trigger on entities;
+
+create trigger entity_change_insert_trigger
+  after insert on entities
+  for each row
+  when (new.subgraph != 'subgraphs')
+  execute procedure log_entity_event();
+
+create trigger entity_change_update_trigger
+  after update on entities
+  for each row
+  when (old.subgraph != 'subgraphs' and old.data != new.data)
+  execute procedure log_entity_event();
+
+create trigger entity_change_delete_trigger
+  after delete on entities
+  for each row
+  when (old.subgraph != 'subgraphs')
+  execute procedure log_entity_event();
+
+-- Create a view to get entity_history by source ordered by event_id desc
+-- We use this both in revert_block and in the Rust code in
+-- store_events.get_revert_event
+create or replace view entity_history_with_source as
+select
+  h.subgraph,
+  h.entity,
+  h.entity_id,
+  h.data_before,
+  h.op_id,
+  m.source
+from entity_history h, event_meta_data m
+where h.event_id = m.id
+order by h.event_id desc;
+
+-- Rewrite revert_block to be selfcontained and to get all the information
+-- needed for reversion in one query
+create or replace function
+  revert_block(block_to_revert_hash VARCHAR,
+               subgraph_id VARCHAR)
+    returns void as
+$$
+declare
+    history record;
+begin
+    -- Revert all relevant events
+    for history in
+        -- Get all history entries associated with the given block. Note
+        -- that the view imposes the correct order
+    select
+        h.entity,
+        h.entity_id,
+        h.data_before,
+        h.op_id
+    from entity_history_with_source h
+    where h.source = block_to_revert_hash
+      and h.subgraph = subgraph_id
+    loop
+        case
+            -- insert case
+            when history.op_id = 0 then
+                -- Delete inserted row
+            begin
+                perform set_config('vars.current_event_source',
+                                   'REVERSION', false);
+                delete from entities
+                 where subgraph = subgraph_id
+                   and entity = history.entity
+                   and id = history.entity_id;
+                -- Row was already updated
+            exception when no_data_found then
+              -- do nothing, the entity was gone already
+            end;
+            -- update or delete case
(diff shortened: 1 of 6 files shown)


lutter:
This PR addresses a few of the bigger points from issue #807. 

Note that the migration that comes with this PR runs for a very long time (~10 minutes on my machine, caused by the delete statements at the end of the migration). It should not affect normal operation of the service as Postgres will only lock the rows that we are deleting, and we never touch them, but it will have an impact on CPU/memory.

To make this even scarier, we will need to issue a `VACUUM (FULL, ANALYZE) entity_history` and a `VACUUM (FULL, ANALYZE) event_meta_data` to reclaim the space from all the deleted rows at some point. But that will make the service inoperable as it will prevent any read or write activity to these tables while it is running. We could also do just a 'normal' vacuum, not a full one, which will still allow normal operation, and only prohibit schema changes while it is running. It will not shrink the storage on disk, but should otherwise be just as good.

store/postgres/migrations/2019-03-07-171355_update_history_storage/up.sql
```diff
@@ -0,0 +1,178 @@
...
+        (COALESCE(new_event_id, event_id), entity_id, subgraph, entity,
+         data_before, data_after, is_reversion, operation_type);
+    RETURN NULL;
+END;
+$$ LANGUAGE plpgsql;
+
+-- Create a view to get entity_history by source ordered by event_id desc
+-- We use this both in revert_block and in the Rust code in
+-- store_events.get_revert_event
+create or replace view entity_history_with_source as
```
Jannis: Nit: Should the SQL keywords such as `CREATE OR REPLACE VIEW` be all-uppercase here, to be consistent with the rest of the file?

store/postgres/migrations/2019-03-07-171355_update_history_storage/up.sql
```diff
@@ -0,0 +1,178 @@
...
+        is_reversion := TRUE;
+    END IF;
+
+    SELECT id INTO event_id
+    FROM event_meta_data
+    WHERE db_transaction_id = txid_current();
+
+    new_event_id := null;
+
+    IF event_id IS NULL THEN
```
Jannis: I know this was here before, but now that I look at it, I don't understand what this (and the `SELECT` query just above) does. Can we add a comment?

store/postgres/migrations/2019-03-07-171355_update_history_storage/up.sql
```diff
@@ -0,0 +1,178 @@
...
+        -- Get all history entries associated with the given block
+    SELECT
+        h.entity,
+        h.entity_id,
+        h.data_before,
+        h.op_id
+    FROM entity_history_with_source h
+    WHERE h.source = block_to_revert_hash
+      AND h.subgraph = subgraph_id
+    LOOP
```
Jannis: I was going to say "these were ordered before" and then I realized `entity_history_with_source` _is_ pre-ordered. It would be nice to have a comment here that expresses that _or_ to rename `entity_history_with_source to `ordered_entity_history_with_source` to make this obvious.

store/postgres/migrations/2019-03-07-171355_update_history_storage/up.sql
lutter: Yeah, a comment is in order. It's making sure that we have a `meta_data_event` for the current transaction, but that we only create one per transaction (so that all `entity_history` entries that get created in the txn are tied to the same event)

store/postgres/migrations/2019-03-07-171355_update_history_storage/up.sql
lutter: Heh. I keep going back and forth on this, and find the all-caps style incredibly shouty (it's not the 80's anymore, we do have terminals with lower and uppercase) but I'll change that to match the rest of the file.

store/postgres/migrations/2019-03-07-171355_update_history_storage/up.sql
lutter: Ok. I'll add a comment, both here and where the view is defined.

lutter:
There's one more thing that I need to address: since we are deleting a ton of data, we will likely run into a scenario where the WAL log becomes way too big while the `delete` is running. I think I need to batch the delete (say 1,000,000 rows at a time) and commit those explicitly to avoid WAL issues.

lutter:
I've decided to change strategy a little bit: rather than bundle everything into one PR (and one migration), I am making this PR only about not recording history that we shouldn't be recording in the first place, and will open another PR for the migration that will remove the needlessly recorded data. These two operations carry hugely different risks.

As it is now, this PR is a fairly routine code change, and can be roled out without putting much stress on the database. I believe I have addressed all review comments so far, but would hugely appreciate another thorough review of the changes.

store/postgres/migrations/2019-03-07-171355_update_history_storage/up.sql
lutter: I solved that by downcasing everything in the `up` migration. The `down` migration is just copy pasta of stuff from previous migrations.

lutter:
Rebased to latest master.

Issue #770, opened 2019-02-25
Limit total number of entities to return


Jannis:
Calculate the number of possible entities returned from a query (similar to GitHub's GraphQL explorer) and fail if the number exceeds a limit defined by an environment variable.

leodasvacas:
This check should be done in the `graphql` crate. The best place to do this is probably in `fn execute_root_selection_set`, in `graphql/src/execution/execution.rs`. That function receives a [SelectionSet](https://docs.rs/graphql-parser/0.2.2/graphql_parser/query/struct.SelectionSet.html). Since we don't currently support fragments, that simplifies the logic, we can recursively dive into the selection set fields (which may also contain selection sets) and calculate the number of entities the query might return. The calculation should be done as described in the [github API docs](https://developer.github.com/v4/guides/resource-limitations/).

nenadjaja:
Assigning it back to @leodasvacas since I got stuck with other stuff.. Sorry Leo. I'll be studying Rust on the side for sure, and will be able to help with future (smaller) tasks

leodasvacas:
### Requirements

- Each query should have a maximum complexity, calculated based on the amount of entities returned, as described in the [github API docs](https://developer.github.com/v4/guides/resource-limitations/).
- If a query exceeds the complexity, it should not be executed and instead return an error.

### Proposed solution
- Check the complexity in `execute_root_selection_set`, before executing the query. Walk the AST selection set tree to calculate the complexity, using the `first` argument.
- Set the complexity limit as an env var, defaulting to no limit.

### Tasks

- [ ] Implement the solution in graph node.

Jannis:
@leodasvacas A couple comments:

* Doing it in `execute_root_selection_set` feels ok for queries but for subscriptions, it would be better if we caught this at the time the subscription is set up and then reject the subscription from the start. So I'd prefer for this to be a standalone function that we run either in `GraphQlRunner` or in `execute_query` / `execute_subscription` before executing a query and/or subscription.

* About the first requirement: <q>amount of entities returned</q> is almost accurate. It's more <q>potential amount of entities returned<q>, which is roughly estimated using the `first` arguments + one-to-one relationships.

PR #809, opened 2019-03-08
Make `first` a required argument

### LINKED ISSUE #783: Feature: Have our GraphQL server enforce "first:" in the query
Currently our graphQL endpoint limits the queries to 100 entities returned, which is good! 

But we haven't made this clear to users. It can get weird, because a user might think that there are only 100 ish entities in existence.

We should enforce `first: x` to be in the query. Then if they try to go over 100, we indicate with an error that 100 is the max, and that paginated queries should be used if the user needs more data returned.

leoyvens:
The implementation is just changing the type of first from `Int` to `Int!`. This also refactors code around `first` and `skip`, the default for `skip` is now set in the schema.

There are also improvements to coercion, so we check that `Int` is an i32, and also accept Ints for `ID` and `BigInt` values.

I did some updates to `docs/graphql-api.md`, but that file is being simultaneously moved to the graph explorer docs, can I remove it so it's not duplicated?

Resolves #783. Required for #770.


Jannis (approved):
No... objections!

Jannis:
Ah, and yes, feel free to delete the GraphQL API docs file. Adding docs for this on thegraph.com/docs/ has higher value I feel.

leodasvacas:
@Jannis great! I've deleted the docs file. Since you've already reviewed this, I'll merge this and follow up with the fix to coercion of input objects.

graph/src/components/store.rs
```diff
@@ -128,7 +128,10 @@ impl EntityQuery {
             filter: None,
             order_by: None,
             order_direction: None,
-            range: None,
+            range: EntityRange {
+                first: 100,
+                skip: 0,
+            },
```
lutter: Sorry i didn't catch this earlier, but this change is too intrusive: we also use `EntityQuery` for our subgraph metadata, and all such queries are now also restricted to 100 entries, even if we mean for them to be unrestricted. For example, we list all subgraphs [here](https://github.com/graphprotocol/graph-node/blob/master/core/src/subgraph/registrar.rs#L284) and if there are more than 100 subgraphs, they will be invisible.

We should go back to the way things were with the internals of `EntityQuery` and how the store handles them, and only restrict queries that were supplied by the user to the first 100 entities.

leodasvacas: Thanks for catching this, this would be awful! I'll fix it.

PR #811, opened 2019-03-08
graph, store: Fix EntityRange

leodasvacas:
By accident, #809 made the internal queries for subgraph of subgraph entities be limited to 100 results.

Thanks @lutter for catching this in review.

### DIFF
--- a/graph/src/components/store.rs
+++ b/graph/src/components/store.rs
@@ -90,11 +90,21 @@ pub enum EntityOrder {
 /// How many entities to return, how many to skip etc.
 #[derive(Clone, Debug, PartialEq)]
 pub struct EntityRange {
-    /// How many entities to return.
-    pub first: u32,
+    /// Limit on how many entities to return.
+    pub first: Option<u32>,
 
     /// How many entities to skip.
     pub skip: u32,
+}
+
+impl EntityRange {
+    /// Query for the first `n` entities.
+    pub fn first(n: u32) -> Self {
+        Self {
+            first: Some(n),
+            skip: 0,
+        }
+    }
 }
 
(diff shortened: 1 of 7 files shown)


graph/src/components/store.rs
```diff
@@ -121,17 +131,18 @@ pub struct EntityQuery {
 }
 
 impl EntityQuery {
-    pub fn new(subgraph_id: SubgraphDeploymentId, entity_types: Vec<String>) -> Self {
+    pub fn new(
+        subgraph_id: SubgraphDeploymentId,
+        entity_types: Vec<String>,
+        range: EntityRange,
```
lutter: Could we make that `Option<EntityRange>` so callers don't have to create their own all the time?

store/postgres/src/store.rs
```diff
@@ -369,9 +369,10 @@ impl Store {
         }
 
         // Add range filter to query
-        diesel_query = diesel_query
-            .limit(query.range.first as i64)
-            .offset(query.range.skip as i64);
+        if let Some(limit) = query.range.first {
+            diesel_query = diesel_query.limit(limit as i64);
+        }
+        diesel_query = diesel_query.offset(query.range.skip as i64);
```
lutter: Even though pg seems to optimize `skip == 0` away, it would be clearer to just write the `if skip > 0 { ... }` here explicitly.

graph/src/components/store.rs
leodasvacas: This has two callers and each uses a different range.

store/postgres/src/store.rs
leodasvacas: Done, thanks.

Issue #812, opened 2019-03-09
Improve how we change the subgraph block pointer


lutter:
I just ran postgres with `log_statement = 'all'` for a little bit, which helps understand where database load is coming from. On my node which is processing a single subgraph, it's apparent that how we update the subgraph deployment's block pointer is very wasteful: updating the block pointer should be a single update statement along the lines of (jsonb complications left out)

```
update entities
set latestEthereumBlockNumber = $1,
      latestEthereumBlockHash = $2
where subgraph='subgraphs' and entity='SubgraphDeployment' and id = $3
    and latestEthereumBlockHash = $4 -- optimistic versioning
```

What happens instead is
1.  `select data, entity from entities where ...` to make sure the block pointer is still in the database (optimistic versioning)
1. `select data from entities where ...` to read the full entity (and update it in memory)
1. upsert the modified entity (in particular its `data`)

Not only does this use 3 queries to do what could be done with one, each query also involves the full jsonb `data` for the entity, which is ~ 500 bytes.

I am confident that we can greatly improve that and considerably reduce the database load if we add an `EntityOperation` to the store that allows describing an operation like the `update` above.

PR #820, opened 2019-03-11
Reduce the number of db operations for updating a subgraph's block pointer

lutter:
This reduces the db traffic caused by certain subgraph metadata updates from 3 (or more) to 1.

### DIFF
--- a/graph/src/components/store.rs
+++ b/graph/src/components/store.rs
@@ -204,7 +204,9 @@ impl EntityChange {
         use self::EntityOperation::*;
 
         match operation {
-            Set { key, .. } => Some(Self::from_key(key, EntityChangeOperation::Set)),
+            Set { key, .. } | Update { key, .. } => {
+                Some(Self::from_key(key, EntityChangeOperation::Set))
+            }
             Remove { key } => Some(Self::from_key(key, EntityChangeOperation::Removed)),
             // AbortUnless is uninteresting for when we need an EntityChange
             AbortUnless { .. } => None,
@@ -439,6 +441,16 @@ pub enum EntityOperation {
         query: EntityQuery,  // The query to run
         entity_ids: Vec<String>, // What entities the query should return
     },
+
+    /// Update an entity. The `data` should only contain the attributes that
+    /// need to be changed, not the entire entity. The update will only happen
+    /// if the given entity matches `guard` when the update is made. `Update`
+    /// provides a way to atomically do a check-and-set change to an entity.
+    Update {
+        key: EntityKey,
+        data: Entity,
+        guard: Option<EntityFilter>,
+    },
 }
 
 impl EntityOperation {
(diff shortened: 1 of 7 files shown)


lutter:
Ideally, this would not get deployed to prod until after PR #808 has been deployed so that we can watch the effect of each change in isolation (say, for 24h)

mock/src/store.rs
```diff
@@ -191,6 +191,41 @@ impl Store for MockStore {
...
+                                    expected_entity_ids: vec![key.entity_id],
+                                    actual_entity_ids: vec![],
+                                    description:
+                                        "update failed because entity does not match guard"
+                                            .to_owned(),
+                                }
+                                .into());
+                            }
+                        }
+                        // FIXME: Check that existing_entity matches guard
```
Jannis: Is this something that still needs to be done?

store/postgres/src/store.rs
```diff
@@ -509,6 +521,70 @@ impl Store {
             })
     }
 
+    /// Applies an update operation to an existing entity
+    fn apply_update_operation(
```
Jannis: Apart from the added guard, I'm not sure I understand why this method is so different from `apply_set_operation`. Shouldn't this be a basic extension of `apply_set_operation`? If the actual updating and error handling logic here is better and more efficient, perhaps we should move it to `apply_set_operation`.

lutter: The big difference is that `Update` uses the knowledge that (1) we are dealing with an existing entity and (2) which attributes to change to bring better performance. Doing this for `Set` in general would probably be worthwhile, but would require tracking more information in the mapping handlers (was the entity loaded from the db or created from scratch? which attributes were actually changed?) to get the same improvements, and would be a more invasive change.

graph/src/components/store.rs
```diff
@@ -439,6 +441,16 @@ pub enum EntityOperation {
         query: EntityQuery,  // The query to run
         entity_ids: Vec<String>, // What entities the query should return
     },
+
+    /// Update an entity. The `data` should only contain the attributes that
+    /// need to be changed, not the entire entity. The update will only happen
+    /// if the given entity matches `guard` when the update is made. `Update`
+    /// provides a way to atomically do a check-and-set change to an entity.
+    Update {
```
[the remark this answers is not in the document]
lutter: IMHO, `Update` fits better with the other entity operations `Set` and `Remove` and it literally is a SQL update statement, just that the `where` clause is generated from mandatory data (the entity key) and optional data (the guard). I'd prefer to stick with `Update` here.

mock/src/store.rs
lutter: Ugh. No, I added the checking at some point when I was working on it, and forgot to take out the comment. Will do that and update the PR.

Jannis:
I'm ok moving forward with this. The reason `Update` felt bad to me is that we only want to expose `set` operations to mappings — and `EntityOperations` were initially intended to reflect the operations that mappings could perform. But we've already loosened that up with `AbortUnless`, so this is fine.

lutter:
Should I merge this now or wait until we've deployed PR #808 to production? I want to avoid both of them hitting prod at once inadvertently

Jannis:
I'd say we merge this now but pick the commit before the merge for our next production upgrade. Just need to remember that. 😉

Jannis:
Resolves #812.

PR #821, opened 2019-03-13
store/postgres: fix syntax error with Postgres 9.6

lutter:
Postgres 9.6 requires '()' after the name of a function when dropping it.


lutter:
This should be ok as we have not applied this migration in production, nor tagged it for a new release. And it's only a problem with Postgres 9.6, not with 10+

PR #822, opened 2019-03-13
Delete unneeded entity_history and event_meta_data

lutter:
Get rid of the following in entity_history:

- all history for the 'subgraphs' subgraph
- all history where data_before = data_after
- the data_after column in entity_history
- any event_meta_data event that is not used by an entry in entity_history

The migration is a little more complicated than one might think because we
are trying to do this as fast as possible: instead of deleting from
entity_history and event_meta_data, we copy the 'good' data to a new
table. We also defer creating constraints and indexes until after the data
has been copied since bulk creation is faster than doing it on every row.

Fixes https://github.com/graphprotocol/graph-node/issues/807

### DIFF
--- a/store/postgres/migrations/2019-03-11-221621_delete_unneeded_entity_history/up.sql
+++ b/store/postgres/migrations/2019-03-11-221621_delete_unneeded_entity_history/up.sql
@@ -1,0 +1,190 @@
+-- Remove mention of data_after from the log_entity_event function
+CREATE OR REPLACE FUNCTION log_entity_event()
+    RETURNS trigger AS
+$$
+DECLARE
+    event_id INTEGER;
+    new_event_id INTEGER;
+    is_reversion BOOLEAN := FALSE;
+    operation_type INTEGER := 10;
+    event_source  VARCHAR;
+    subgraph VARCHAR;
+    entity VARCHAR;
+    entity_id VARCHAR;
+    data_before JSONB;
+BEGIN
+    -- Get operation type and source
+    IF (TG_OP = 'INSERT') THEN
+        operation_type := 0;
+        event_source := NEW.event_source;
+        subgraph := NEW.subgraph;
+        entity := NEW.entity;
+        entity_id := NEW.id;
+        data_before := NULL;
+    ELSIF (TG_OP = 'UPDATE') THEN
+        operation_type := 1;
+        event_source := NEW.event_source;
+        subgraph := OLD.subgraph;
+        entity := OLD.entity;
+        entity_id := OLD.id;
+        data_before := OLD.data;
+    ELSIF (TG_OP = 'DELETE') THEN
+        operation_type := 2;
+        event_source := current_setting('vars.current_event_source', TRUE);
+        subgraph := OLD.subgraph;
+        entity := OLD.entity;
+        entity_id := OLD.id;
+        data_before := OLD.data;
+    ELSE
+        RAISE EXCEPTION 'unexpected entity row operation type, %', TG_OP;
+    END IF;
+
+    IF event_source = 'REVERSION' THEN
+        is_reversion := TRUE;
+    END IF;
+
+    SELECT id INTO event_id
+    FROM event_meta_data
+    WHERE db_transaction_id = txid_current();
+
+    new_event_id := null;
+
+    IF event_id IS NULL THEN
+        -- Log information on the postgres transaction for later use in revert operations
+        INSERT INTO event_meta_data
+            (db_transaction_id, db_transaction_time, source)
+        VALUES
+            (txid_current(), statement_timestamp(), event_source)
+        RETURNING event_meta_data.id INTO new_event_id;
+    END IF;
+
+    -- Log row metadata and changes, specify whether event was an original ethereum event or a reversion
+    INSERT INTO entity_history
+        (event_id, entity_id, subgraph, entity, data_before, reversion, op_id)
+    VALUES
+        (COALESCE(new_event_id, event_id), entity_id, subgraph, entity, data_before, is_reversion, operation_type);
+    RETURN NULL;
+END;
+$$ LANGUAGE plpgsql;
+
+-- Remove unneeded entity_history data, and event_meta_data that becomes
+-- unnecessary because of that, too. Since we need to get rid of most of
+-- the data, it is faster to copy the 'good' data into a new table and swap
+-- tables around
+
+-- Create a new table that clones the entity_history table. For now,
+-- we only clone the schema. We create indexes and constraints after
+-- we are done copying, for speed.
+create table eh(like entity_history
+                     including defaults
+                     including comments
+                     including storage);
+-- eh and entity_history both use entity_history_id_seq now. Remove its
+-- use from entity_history, so we don't get a conflict later when we
+-- drop that table
+alter table entity_history
+      alter column id drop default;
+alter sequence entity_history_id_seq
+      owned by eh.id;
+alter table eh drop column data_after;
+
+-- Lock any writes to entity_history and event_meta_data. We can not
+-- afford to have any changes happen while we copy, as we would lose
+-- them because we are copying
+lock table entity_history in exclusive mode;
+lock table event_meta_data in exclusive mode;
+
+-- Copy the good data. Do not log to avoid stress on the WAL. Note that
+-- either data_before or data_after can be null, and we need to make sure
+-- we include rows with a null in either column, but that 'anything !=
+-- null' will always be false. We lift SQL NULL to JSON null, since JSON
+-- uses normal, not three-state logic for comparisons with null
+alter table eh set unlogged;
+insert into eh
+  select id, event_id, entity_id, subgraph, entity, data_before,
+         reversion, op_id from entity_history
+  where subgraph != 'subgraphs'
+    and coalesce(data_before, 'null'::jsonb)
+        != coalesce(data_after, 'null'::jsonb);
+alter table eh set logged;
+
+-- Swap old and new tables
+alter table entity_history rename to eh_old;
+alter table eh rename to entity_history;
+
+-- Create indexes
+alter index entity_history_pkey rename to eh_old_pkey;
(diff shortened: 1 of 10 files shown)


lutter:
Note that this PR sits on top of PR #820 and should only be merged after that one has been in production for a bit.

I've tested this locally with a couple of production dumps, and `graph-node` seems to work after this migration ran.

I also tested this by creating a new database in CloudSQL (4 vcpu/ 32GB of RAM) by restoring a backup from 3/11 into that instance and running the `up.sql` script, mostly to see what stress it puts on the database. The migration uses less than 1 vcpu, and maximum memory use was ~ 12GB. It does need extra storage (on the order of 10GB) as the tables are copied, and not changed in place. The whole migration takes 20-30 minutes to run. During that time, it is not possible to make any changes to `entities`, and block/subgraph ingestion will pause (and resume once the migration is done) It is possible though to query the database during that time, so query activity should not be affected too much.

Before running this migration, we absolutely need to take another database backup since that will be the only way to really go back should anything unforeseen happen during the migration.

Postgres will also log a lot of messages `checkpoints are occurring too frequently (9 seconds apart)` Those warnings are harmless and can be ignored. It's to be expected that the (untypical) write activity during the migration is forcing the database to checkpoint a lot.

lutter:
I just realized this has a serious bug: the query that selects the rows to keep from `entity_history` does not take nulls into account in the `data_before != data_after` clause, which is always false if one of the values is `null`, and therefore removes all history with `data_before = null`, even though we should keep it.

lutter:
Fixed the problem. SHould be good to go now.

store/postgres/migrations/2019-03-11-221621_delete_unneeded_entity_history/up.sql
```diff
@@ -0,0 +1,190 @@
...
+    INSERT INTO entity_history
+        (event_id, entity_id, subgraph, entity, data_before, reversion, op_id)
+    VALUES
+        (COALESCE(new_event_id, event_id), entity_id, subgraph, entity, data_before, is_reversion, operation_type);
+    RETURN NULL;
+END;
+$$ LANGUAGE plpgsql;
+
+-- Remove unneeded entity_history data, and event_meta_data that becomes
+-- unnecessary because of that, too. Since we need to get rid of most of
```
fordN: The first sentence in this comment is somewhat unclear, "because of...".

store/postgres/migrations/2019-03-11-221621_delete_unneeded_entity_history/up.sql
```diff
@@ -0,0 +1,190 @@
...
+      alter column id drop default;
+alter sequence entity_history_id_seq
+      owned by eh.id;
+alter table eh drop column data_after;
+
+-- Lock any writes to entity_history and event_meta_data. We can not
+-- afford to have any changes happen while we copy, as we would lose
+-- them because we are copying
+lock table entity_history in exclusive mode;
+lock table event_meta_data in exclusive mode;
```
fordN: Will taking these locks result in a lock on `entities` as well? Can normal operation of services that rely on it continue during the migration?

store/postgres/migrations/2019-03-11-221621_delete_unneeded_entity_history/up.sql
lutter: You are right, rereading this makes me wonder what I was trying to say. I'll fix the comment up.

store/postgres/migrations/2019-03-11-221621_delete_unneeded_entity_history/up.sql
lutter: It will not locking the `entities` table, but have a very similar effect, because any change to the `entities` table requires changing `entity_history`, and that will be blocked while the migration is running. The effect is that any write activity will pause while the migration is running, and resumes when it is done. In my (local) testing, it is still possible to query the `entities` table while the migration is running

store/postgres/migrations/2019-03-11-221621_delete_unneeded_entity_history/up.sql
```diff
@@ -0,0 +1,190 @@
...
+alter table entity_history
+      alter column id drop default;
+alter sequence entity_history_id_seq
+      owned by eh.id;
+alter table eh drop column data_after;
+
+-- Lock any writes to entity_history and event_meta_data. We can not
+-- afford to have any changes happen while we copy, as we would lose
+-- them because we are copying
+lock table entity_history in exclusive mode;
```
Jannis: Is this something we should do _before_ dropping the `id` column? Just curious if there could be any race conditions here.

store/postgres/migrations/2019-03-11-221621_delete_unneeded_entity_history/up.sql
```diff
@@ -0,0 +1,190 @@
...
+        (event_id, entity_id, subgraph, entity, data_before, reversion, op_id)
+    VALUES
+        (COALESCE(new_event_id, event_id), entity_id, subgraph, entity, data_before, is_reversion, operation_type);
+    RETURN NULL;
+END;
+$$ LANGUAGE plpgsql;
+
+-- Remove unneeded entity_history data, and event_meta_data that becomes
+-- unnecessary because of that, too. Since we need to get rid of most of
+-- the data, it is faster to copy the 'good' data into a new table and swap
```
Jannis: Curious: Is it worth doing this copy, recreate indexes, drop table, rename dance just for performance? I prefer simplicity over performance (until performance becomes a problem).

store/postgres/migrations/2019-03-11-221621_delete_unneeded_entity_history/up.sql
```diff
@@ -0,0 +1,190 @@
...
+    m.source
+  from entity_history h, event_meta_data m
+  where h.event_id = m.id
+  order by h.event_id desc;
+
+-- Clean up
+drop table eh_old;
+drop table emd_old;
+
+analyze entity_history;
```
Jannis: What does this do?

store/postgres/migrations/2019-03-11-221621_delete_unneeded_entity_history/up.sql
```diff
@@ -0,0 +1,190 @@
...
+alter table event_meta_data
+  add constraint event_meta_data_db_transaction_id_key
+                 unique(db_transaction_id);
+
+-- Make sure the fk constraint on entity_history points to the new
+-- event_meta_data table
+alter table eh_old
+  drop constraint entity_history_event_id_fkey;
+
+alter table entity_history
```
Jannis: A comment here would be great; "match full on update cascade on delete cascade" is quite a mouthful. 😁

store/postgres/migrations/2019-03-11-221621_delete_unneeded_entity_history/up.sql
lutter: The simple version of all this is
```
delete from entity_history where subgraph = 'subgraphs' or data_before = data_after;
vacuum (full, analyze) entity_history;
delete from event_meta_data d where not exists (select 1 from entity_history h where h.event_id = d.id);
vacuum (full, analyze) event_meta_data;
```

but that takes at least twice as long as this script (and the more data we accumulate that needs to be deleted, the bigger the advantage of this script), and the vacuum will also lock the corresponding tables, effectively preventing any writes to `entities` (because of the `log_entity_event` trigger).

If you want more precise timing numbers, I could do that on a clone of the prod db.

store/postgres/migrations/2019-03-11-221621_delete_unneeded_entity_history/up.sql
lutter: I added the sequence moving stuff after the locking, and so this should move up, even though dropping the default for the `id` column will take an `access exclusive` lock, which is even stricter than the exclusive lock, and also prevents any queries against `entity_history` from happening. I need to check though that we are not going to shoot ourselves in the foot with these two locks on the same table.

store/postgres/migrations/2019-03-11-221621_delete_unneeded_entity_history/up.sql
lutter: That's just copy paste from `\d entity_history`; I'll add a comment.

store/postgres/migrations/2019-03-11-221621_delete_unneeded_entity_history/up.sql
lutter: It updates Postgres' internal statistics about the table (important for the query planner) Since we deleted so much data, these statistics will be all wrong. The execution of the statement itself is pretty fast (~ 100ms)

lutter:
From talking to @yanivtal, we want to do this without any downtime. One option would be to continually delete small numbers of rows from the application, along the lines of
```
delete from entity_history
where id in (select id from entity_history where subgraph='subgraphs' or data_before = data_after limit 1000) returning event_id;

delete from event_meta_data d
where d.id in :event_ids -- the ones that the delete above returned
    and not exists (select 1 from entity_history h where d.id = h.event_id);
```

If we do this, we have to have a look at whether the autovacuum daemon can keep up with the deletions, if not we'd have to kick off a `VACUUM` manually, but that should be ok, since it does not block other read or write activity. (as opposed to `VACUUM FULL` which does block writes)

We can also update `data_after` and set it to `NULL` which will reduce the space taken up by tuples since nulls do not take up space in postgres. It'll be hard to drop the column, as that takes an exclusive lock. But once we've cleaned up the `entity_history` table, we can try out on a db dump how long dropping the empty column would take. If it's not too long, we could do it, otherwise we'd just have `NULL` in there forever.

I would couple the deletion to getting new blocks from ethereum, i.e. run the deletes every time we get a new block. That way they are throttled to about 4/min. With a batch size of 1000 it will take about a week to delete all the unneeded data from `entity_history`.

Jannis:
Performed manually.

PR #826, opened 2019-03-14
Block & call triggers

ghost:
Currently, a subgraph can only subscribe to Ethereum activity represented in logs, but many contracts opt out of emitting logs to reduce gas costs. Additionally, some contracts contain implicit state which they would like to compute and capture into the index as new blocks are appended to the chain. 

This PR adds the ability to trigger a mapping function when:

- A `CALL` is made with a specific contract address and method id
- A new block is produced
- A new block is produced in which a specific contract was `CALL`ed

### Defining call & block handlers
```yaml
#...
dataSources:
  - kind: ethereum/contract
    name: Gravity
    #...
    mapping:
      #...
      eventHandlers:
        - event: NewGravatar(uint256,address,string,string)
          handler: handleNewGravatar
      callHandlers:
        - function: createGravatar(string,string)
          handler: handleCreateGravatarCall
      blockHandlers:
        - handler: handleBlock
        - handler: handleBlockWithCallToContract
           filter:
             kind: call
```

### Implementing call & block handlers in a manifest

```javascript
import { NewGravatar } from './types/Gravity/Gravity'
import { Gravatar, Transaction, Block, CreateGravatarCall } from './types/schema'
import { EthereumBlock } from '@graphprotocol/graph-ts'

export function handleNewGravatar(event: NewGravatar): void {
  let gravatar = new Gravatar(event.params.id.toHex())
  gravatar.owner = event.params.owner
  gravatar.displayName = event.params.displayName
  gravatar.imageUrl = event.params.imageUrl
  gravatar.save()
}

// The `CreateGravatarCall` type is generated for the developer using 
// the contract's ABI json file.
// The type extends the `EthereumCall` object defined in graph-ts by 
// including the typed inputs and outputs for the call.
export function handleCreateGravatar(call: CreateGravatarCall): void {
    let id = call.transaction.hash.toHex()
    let displayName = call.inputs_displayName
    let imageUrl = call.inputs._imageUrl

    let transaction = new Transaction(id)
    transaction.displayName = displayName
    transaction.imageUrl = imageUrl

    transaction.save()
}

export function handleBlock(block: EthereumBlock): void {
    // Call view functions/static methods provided by the contract 
    // to create or update entities
}

export function handleBlockWithCallToContract(block: EthereumBlock): void {
    // Call view functions/static methods provided by the contract 
    // to create or update entities
}
```

### How call triggers are produced:

Every block bundles up a set of transactions, and each transaction has a call chain. A transaction's call chain can include interactions with a number of different contract and wallet accounts. (Note: a call in a call chain is equivalent to an internal transaction)

To produce call triggers, `graph-node` analyzes the call chain for each transaction in every block as they are produced. A subgraph can subscribe to calls directed at a specific combination of contract address and method id. Regardless of the depth at which a call lives within the call chain, `graph-node` will trigger the subgraph provided that the encapsulating transaction was not reverted.

### How block triggers are produced:

A subgraph can subscribe to all blocks, and this will ensure that the sugraph's mapping is triggered every block. 

In addition, a subgraph can subscribe to all blocks which include at least one call to a specified contract address. This trigger is generated using a similar process to the one outlined above for call triggers.

### DIFF
--- a/core/src/subgraph/validation.rs
+++ b/core/src/subgraph/validation.rs
@@ -11,14 +11,14 @@ pub fn validate_manifest(manifest: SubgraphManifest) -> Result<SubgraphManifest,
                 .source
                 .address
                 .is_none();
-            let has_call_handlers = data_source
+            let has_call_handlers = !data_source
                 .mapping
                 .call_handlers
-                .len() > 0;
-            let has_block_handlers = data_source
+                .is_empty();
+            let has_block_handlers = !data_source
                 .mapping
-                .block_handler
-                .is_some();
+                .block_handlers
+                .is_empty();
             no_source_address && (has_call_handlers || has_block_handlers)
         });
     if has_invalid_data_source {
@@ -26,5 +26,35 @@ pub fn validate_manifest(manifest: SubgraphManifest) -> Result<SubgraphManifest,
             SubgraphManifestValidationError::SourceAddressRequired
         ))
     }
+
+    // Validate that there are no more than one of each type of
+    // block_handler in each data source.
+    let has_too_many_block_handlers = manifest
+        .data_sources
+        .iter()
+        .any(|data_source| {
+            if data_source.mapping.block_handlers.is_empty() {
+                return false
+            }
+            let mut non_filtered_block_handler_count = 0;
+            let mut call_filtered_block_handler_count = 0;
+            data_source
+                .mapping
+                .block_handlers
+                .iter()
+                .for_each(|block_handler| {
+                    if block_handler.filter.is_none() {
+                        non_filtered_block_handler_count += 1
+                    }
+                    call_filtered_block_handler_count += 1
+                });
+            return non_filtered_block_handler_count > 1 || call_filtered_block_handler_count > 1
+        });
+    if has_too_many_block_handlers {
+        return Err(SubgraphRegistrarError::ManifestValidationError(
+            SubgraphManifestValidationError::DataSourceBlockHandlerLimitExceeded
+        ))
+    }
+    
     Ok(manifest)
 }
--- a/core/src/subgraph/validation.rs
+++ b/core/src/subgraph/validation.rs
@@ -3,6 +3,8 @@ use graph::prelude::*;
 pub fn validate_manifest(
     manifest: SubgraphManifest,
 ) -> Result<SubgraphManifest, SubgraphRegistrarError> {
+    let mut validation_errs: Vec<SubgraphManifestValidationError> = Vec::new();
+
     // Validate that the manifest has a `source` address in each data source
     // which has call or block handlers
     let has_invalid_data_source = manifest.data_sources.iter().any(|data_source| {
@@ -12,9 +14,7 @@ pub fn validate_manifest(
         no_source_address && (has_call_handlers || has_block_handlers)
     });
     if has_invalid_data_source {
-        return Err(SubgraphRegistrarError::ManifestValidationError(
-            SubgraphManifestValidationError::SourceAddressRequired,
-        ));
+        validation_errs.push(SubgraphManifestValidationError::SourceAddressRequired)
     }
 
     // Validate that there are no more than one of each type of
@@ -38,10 +38,13 @@ pub fn validate_manifest(
         return non_filtered_block_handler_count > 1 || call_filtered_block_handler_count > 1;
     });
     if has_too_many_block_handlers {
-        return Err(SubgraphRegistrarError::ManifestValidationError(
-            SubgraphManifestValidationError::DataSourceBlockHandlerLimitExceeded,
-        ));
+        validation_errs.push(SubgraphManifestValidationError::DataSourceBlockHandlerLimitExceeded)
     }
 
-    Ok(manifest)
+    if validation_errs.is_empty() {
+        return Ok(manifest);
+    }
+    return Err(SubgraphRegistrarError::ManifestValidationError(
+        validation_errs,
+    ));
 }
--- a/core/src/subgraph/validation.rs
+++ b/core/src/subgraph/validation.rs
@@ -1,0 +1,30 @@
+use graph::prelude::*;
+
+pub fn validate_manifest(manifest: SubgraphManifest) -> Result<SubgraphManifest, SubgraphRegistrarError> {
+    // Validate that the manifest has a `source` address in each data source
+    // which has call or block handlers
+    let has_invalid_data_source = manifest
+        .data_sources
+        .iter()
+        .any(|data_source| {
+            let no_source_address = data_source
+                .source
+                .address
+                .is_none();
+            let has_call_handlers = data_source
+                .mapping
+                .call_handlers
+                .len() > 0;
+            let has_block_handlers = data_source
+                .mapping
+                .block_handler
+                .is_some();
+            no_source_address && (has_call_handlers || has_block_handlers)
+        });
+    if has_invalid_data_source {
+        return Err(SubgraphRegistrarError::ManifestValidationError(
+            SubgraphManifestValidationError::SourceAddressRequired
+        ))
+    }
+    Ok(manifest)
+}
--- a/core/src/subgraph/validation.rs
+++ b/core/src/subgraph/validation.rs
@@ -32,8 +32,9 @@ pub fn validate_manifest(
             .for_each(|block_handler| {
                 if block_handler.filter.is_none() {
                     non_filtered_block_handler_count += 1
+                } else {
+                    call_filtered_block_handler_count += 1
                 }
-                call_filtered_block_handler_count += 1
             });
         return non_filtered_block_handler_count > 1 || call_filtered_block_handler_count > 1;
     });
--- a/core/src/subgraph/validation.rs
+++ b/core/src/subgraph/validation.rs
@@ -1,60 +1,47 @@
 use graph::prelude::*;
 
-pub fn validate_manifest(manifest: SubgraphManifest) -> Result<SubgraphManifest, SubgraphRegistrarError> {
+pub fn validate_manifest(
+    manifest: SubgraphManifest,
+) -> Result<SubgraphManifest, SubgraphRegistrarError> {
     // Validate that the manifest has a `source` address in each data source
     // which has call or block handlers
-    let has_invalid_data_source = manifest
-        .data_sources
-        .iter()
-        .any(|data_source| {
-            let no_source_address = data_source
-                .source
-                .address
-                .is_none();
-            let has_call_handlers = !data_source
-                .mapping
-                .call_handlers
-                .is_empty();
-            let has_block_handlers = !data_source
-                .mapping
-                .block_handlers
-                .is_empty();
-            no_source_address && (has_call_handlers || has_block_handlers)
-        });
+    let has_invalid_data_source = manifest.data_sources.iter().any(|data_source| {
+        let no_source_address = data_source.source.address.is_none();
+        let has_call_handlers = !data_source.mapping.call_handlers.is_empty();
+        let has_block_handlers = !data_source.mapping.block_handlers.is_empty();
+        no_source_address && (has_call_handlers || has_block_handlers)
+    });
     if has_invalid_data_source {
         return Err(SubgraphRegistrarError::ManifestValidationError(
-            SubgraphManifestValidationError::SourceAddressRequired
-        ))
+            SubgraphManifestValidationError::SourceAddressRequired,
+        ));
     }
 
     // Validate that there are no more than one of each type of
     // block_handler in each data source.
-    let has_too_many_block_handlers = manifest
-        .data_sources
-        .iter()
-        .any(|data_source| {
-            if data_source.mapping.block_handlers.is_empty() {
-                return false
-            }
-            let mut non_filtered_block_handler_count = 0;
-            let mut call_filtered_block_handler_count = 0;
-            data_source
-                .mapping
-                .block_handlers
-                .iter()
-                .for_each(|block_handler| {
-                    if block_handler.filter.is_none() {
-                        non_filtered_block_handler_count += 1
-                    }
-                    call_filtered_block_handler_count += 1
-                });
-            return non_filtered_block_handler_count > 1 || call_filtered_block_handler_count > 1
-        });
+    let has_too_many_block_handlers = manifest.data_sources.iter().any(|data_source| {
+        if data_source.mapping.block_handlers.is_empty() {
+            return false;
+        }
+        let mut non_filtered_block_handler_count = 0;
+        let mut call_filtered_block_handler_count = 0;
+        data_source
+            .mapping
+            .block_handlers
+            .iter()
+            .for_each(|block_handler| {
+                if block_handler.filter.is_none() {
+                    non_filtered_block_handler_count += 1
+                }
+                call_filtered_block_handler_count += 1
+            });
+        return non_filtered_block_handler_count > 1 || call_filtered_block_handler_count > 1;
+    });
     if has_too_many_block_handlers {
         return Err(SubgraphRegistrarError::ManifestValidationError(
-            SubgraphManifestValidationError::DataSourceBlockHandlerLimitExceeded
-        ))
+            SubgraphManifestValidationError::DataSourceBlockHandlerLimitExceeded,
+        ));
     }
-    
+
     Ok(manifest)
 }
--- a/core/src/subgraph/validation.rs
+++ b/core/src/subgraph/validation.rs
@@ -3,49 +3,63 @@ use graph::prelude::*;
 pub fn validate_manifest(
     manifest: SubgraphManifest,
 ) -> Result<SubgraphManifest, SubgraphRegistrarError> {
-    let mut validation_errs: Vec<SubgraphManifestValidationError> = Vec::new();
+    let mut errors: Vec<SubgraphManifestValidationError> = Vec::new();
 
     // Validate that the manifest has a `source` address in each data source
     // which has call or block handlers
     let has_invalid_data_source = manifest.data_sources.iter().any(|data_source| {
         let no_source_address = data_source.source.address.is_none();
-        let has_call_handlers = !data_source.mapping.call_handlers.is_empty();
-        let has_block_handlers = !data_source.mapping.block_handlers.is_empty();
+        let has_call_handlers = !data_source
+            .mapping
+            .call_handlers
+            .as_ref()
+            .map_or(false, |handlers| !handlers.is_empty());
+        let has_block_handlers = !data_source
+            .mapping
+            .block_handlers
+            .as_ref()
+            .map_or(false, |handlers| handlers.is_empty());
+
         no_source_address && (has_call_handlers || has_block_handlers)
     });
+
     if has_invalid_data_source {
-        validation_errs.push(SubgraphManifestValidationError::SourceAddressRequired)
+        errors.push(SubgraphManifestValidationError::SourceAddressRequired)
     }
 
     // Validate that there are no more than one of each type of
     // block_handler in each data source.
     let has_too_many_block_handlers = manifest.data_sources.iter().any(|data_source| {
-        if data_source.mapping.block_handlers.is_empty() {
-            return false;
-        }
-        let mut non_filtered_block_handler_count = 0;
-        let mut call_filtered_block_handler_count = 0;
-        data_source
+        if data_source
             .mapping
             .block_handlers
-            .iter()
-            .for_each(|block_handler| {
+            .as_ref()
+            .map_or(true, |handlers| handlers.is_empty())
+        {
+            return false;
+        }
+
+        let mut non_filtered_block_handler_count = 0;
+        let mut call_filtered_block_handler_count = 0;
+        if let Some(ref handlers) = data_source.mapping.block_handlers {
+            handlers.iter().for_each(|block_handler| {
                 if block_handler.filter.is_none() {
                     non_filtered_block_handler_count += 1
                 } else {
                     call_filtered_block_handler_count += 1
                 }
             });
+        }
         return non_filtered_block_handler_count > 1 || call_filtered_block_handler_count > 1;
     });
+
     if has_too_many_block_handlers {
-        validation_errs.push(SubgraphManifestValidationError::DataSourceBlockHandlerLimitExceeded)
+        errors.push(SubgraphManifestValidationError::DataSourceBlockHandlerLimitExceeded)
     }
 
-    if validation_errs.is_empty() {
+    if errors.is_empty() {
         return Ok(manifest);
     }
-    return Err(SubgraphRegistrarError::ManifestValidationError(
-        validation_errs,
-    ));
+
+    return Err(SubgraphRegistrarError::ManifestValidationError(errors));
 }
--- a/datasource/ethereum/src/block_stream.rs
+++ b/datasource/ethereum/src/block_stream.rs
@@ -1,10 +1,12 @@
 use futures::prelude::*;
 use futures::sync::mpsc::{channel, Receiver, Sender};
 use std::cmp;
-use std::collections::HashSet;
+use std::collections::{HashSet, HashMap};
 use std::env;
 use std::mem;
 use std::sync::Mutex;
+use std::cmp::Ordering;
+use tiny_keccak::keccak256;
 
 use graph::data::subgraph::schema::{
     SubgraphDeploymentEntity, SubgraphEntity, SubgraphVersionEntity,
@@ -57,9 +59,9 @@ enum ReconciliationStep {
     /// Move forwards, processing one or more blocks.
     ProcessDescendantBlocks {
         from: EthereumBlockPointer,
-        log_filter: Option<EthereumLogFilter>,
-        call_filter: Option<EthereumCallFilter>,
-        block_filter: Option<EthereumBlockFilter>,
+        log_filter_opt: Option<EthereumLogFilter>,
+        call_filter_opt: Option<EthereumCallFilter>,
+        block_filter_opt: Option<EthereumBlockFilter>,
         descendant_blocks: Box<Stream<Item = EthereumBlockWithCalls, Error = Error> + Send>,
     },
 
@@ -141,9 +143,9 @@ where
         chain_head_update_guard: CancelGuard,
         node_id: NodeId,
         subgraph_id: SubgraphDeploymentId,
-        log_filter: Option<EthereumLogFilter>,
-        call_filter: Option<EthereumCallFilter>,
-        block_filter: Option<EthereumBlockFilter>,
+        log_filter: EthereumLogFilter,
+        call_filter: EthereumCallFilter,
+        block_filter: EthereumBlockFilter,
         reorg_threshold: u64,
         logger: Logger,
     ) -> Self {
@@ -401,9 +403,9 @@ where
                                             // Proceed to those blocks
                                             ReconciliationStep::ProcessDescendantBlocks {
                                                 from: subgraph_ptr,
-                                                log_filter: log_filter.clone(),
-                                                call_filter: call_filter.clone(),
-                                                block_filter: block_filter.clone(),
+                                                log_filter_opt: log_filter.clone(),
+                                                call_filter_opt: call_filter.clone(),
+                                                block_filter_opt: block_filter.clone(),
                                                 descendant_blocks: Box::new(ctx.load_blocks(descendant_hashes)),
                                             }
                                         ))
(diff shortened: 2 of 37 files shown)


Jannis:
Looking good so far! I've updated the `blockHandlers` and the `...Handlers` names in https://github.com/graphprotocol/graph-cli/pull/229 to `blockHandler` and `...Handler` by the way, as discussed.

lastmjs:
I just wanted to say that I'm really excited for this, keep it up!

datasource/ethereum/src/block_stream.rs
```diff
@@ -1096,5 +1204,182 @@ fn create_log_filter_from_subgraph(manifest: &SubgraphManifest) -> EthereumLogFi
...
+            data_source.source.address
+        })
+        .collect::<EthereumBlockFilter>();
+    match block_filter.contract_addresses.len() {
+        0 => return None,
+        _ => return Some(block_filter),
+    }
+}
+
+fn parse_triggers(
```
olibearo: This deserves a unit test

graph/src/components/ethereum/types.rs
```diff
@@ -45,6 +65,73 @@ impl Default for EthereumBlock {
...
+    pub block_number: u64,
+    pub block_hash: H256,
+    pub transaction_hash: Option<H256>,
+}
+
+impl From<&Trace> for EthereumCall {
+    fn from(trace: &Trace) -> Self {
+        let (from, to, value, input) = match &trace.action {
+            Action::Call(call) => (call.from, call.to, call.value, call.input.clone()),
+            _ => (Address::zero(), Address::zero(), U256::zero(), Bytes::from(vec![])),
```
olibearo: Find a better way to deal with this

graph/src/util/ethereum.rs
```diff
@@ -25,3 +25,26 @@ pub fn contract_event_with_signature<'a>(
         .events()
         .find(|event| event.signature() == string_to_h256(signature))
 }
+
+pub fn contract_function_with_signature<'a>(
```
olibearo: This deserves a unit test

runtime/wasm/src/host.rs
```diff
@@ -267,13 +319,243 @@ impl RuntimeHost {
...
 
+    fn matches_call(&self, call: &EthereumCall) -> bool {
+        self.matches_call_address(call) && self.matches_call_function(call)
+    }
+
+    fn matches_block(&self, block: &EthereumBlock, call: &EthereumCall) -> bool {
+        self.data_source_block_handler.is_some() && self.matches_call_address(call)
+    }
+
+    fn process_call(
```
olibearo: Should be able to make `process_call`, `process_log`, and `process_block` more DRY

datasource/ethereum/src/block_stream.rs
```diff
@@ -98,6 +104,7 @@ struct BlockStreamContext<S, C, E> {
     subgraph_id: SubgraphDeploymentId,
     reorg_threshold: u64,
     logger: Logger,
+    include_calls_in_blocks: bool,
```
olibearo: Think about making this a static function `include_calls_in_block()` instead of a `bool` property

lutter:
I just looked at this PR, and have a really hard time following it. Is there any way you could structure this PR into a series of commits that each get you closer to the goal, but each contain just one logical step towards that and are individually a little easier to grok? I'd really appreciate that!

olibearo:
Per @lutter request, I'll do some rebasing tomorrow to making the series of commits reflect the structural changes to the code base

olibearo:
Got the feature caught up with master and added validation for the subgraph which fails when a data source has no source address but does include call or block handlers. @Jannis I made validation it's own module in `core::subgraph::validation`

datasource/ethereum/src/ethereum_adapter.rs
```diff
@@ -575,26 +648,160 @@ where
...
+                }
+            })
+            .and_then(move |calls| {
+                // Ensure the `EthereumCall` objects are for the correct Ethereum block
+                // by checking the call's block hash against the desired block hash.
+                // 
+                // DISCLAIMER: If the call stream returns calls for the correct block_number but the
+                // incorrect block_hash we can catch this and error. But if the call streams returns no calls
+                // for the correct block_number but the incorrect block_hash, there is no way to tell.
+                // Since we are asking for all the calls in the block this is not a problem.
```
timmclean: Is this not a problem because a block will always have at least one call? Maybe add an assert `len > 0`. Or is there another reason?

datasource/ethereum/src/block_stream.rs
```diff
@@ -1096,5 +1204,188 @@ fn create_log_filter_from_subgraph(manifest: &SubgraphManifest) -> EthereumLogFi
...
+                return Ordering::Equal
+            }
+            if a_tx_index.is_none() {
+                return Ordering::Less
+            }
+            if b_tx_index.is_none() {
+                return Ordering::Greater
+            }
+            a_tx_index.unwrap().cmp(&b_tx_index.unwrap())
+        });
```
timmclean: It looks like all calls within a given transaction will appear after the logs within that transaction. Is that desirable? Might be useful to document that behaviour since it's visible to users I think :)

datasource/ethereum/src/ethereum_adapter.rs
olibearo: Hey Tim! Yeah, that was the assumption. I added the `assert` you mentioned for now, but thinking through this again, I'm not sure that assumption holds.

TODO: Can there be a block without calls? If so a subgraph will stall at that block with this logic.

datasource/ethereum/src/block_stream.rs
olibearo: Good idea, I'll add some documentation in the code and the user docs

Jannis: Yeah, this is our initial convention until we can properly interleave these things.

Jannis:
@olibearo Could we rebase this branch on top of master to get rid of the "Merge branch 'master' into ..." commits?

datasource/ethereum/src/block_stream.rs
```diff
@@ -510,44 +577,64 @@ where
...
+                                     call_filter_opt.clone(),
+                                     block_filter_opt.clone(),
+                                     &descendant_block,
+                                 ) {
+                                     Ok(triggers) => triggers,
+                                     Err(err) => return future::err(err)
+                                 };
+                                 future::ok(EthereumBlockWithTriggers {
+                                     ethereum_block: descendant_block.ethereum_block,
+                                     triggers: triggers,
```
Jannis: This can just be `triggers` instead of `triggers: triggers`.

I'd also merge the two expressions here into one:

```rust
future::result(
    parse_triggers(
        log_filter_opt.clone(),
        ...
    ).map(move |triggers| EthereumBlockWithTriggers {
        ethereum_block: descendant_block.ethereum_block,
        triggers,
    })
)
```

(I haven't tested this but I think it should work.)

datasource/ethereum/src/block_stream.rs
```diff
@@ -753,49 +841,66 @@ where
...
-                            }),
-                    )
+                            }))
+                    }
                 }
-            },
-        )
+            })
+            .and_then(move |block| -> Box<Future<Item = _, Error = _> + Send>{
+                if !include_calls_in_block {
```
Jannis: This `include calls -> either return the block without calls or fetch the calls` logic is showing up in a few places. Could be worth consolidating into a single helper function/method.

datasource/ethereum/src/block_stream.rs
```diff
@@ -1096,5 +1225,188 @@ fn create_log_filter_from_subgraph(manifest: &SubgraphManifest) -> EthereumLogFi
                     (contract_addr, event_sig)
                 })
         })
-        .collect::<EthereumLogFilter>()
+        .collect::<EthereumLogFilter>();
+    match log_filter.contract_address_and_event_sig_pairs.len() {
```
Jannis: I generally prefer an empty line here and there to separate code blocks and longer assignments or call chains. E.g. I would add one above this line.

datasource/ethereum/src/block_stream.rs
```diff
@@ -1096,5 +1225,188 @@ fn create_log_filter_from_subgraph(manifest: &SubgraphManifest) -> EthereumLogFi
...
+                .mapping
+                .call_handlers
+                .iter()
+                .map(move |call_handler| {
+                    let sig = keccak256(call_handler.function.as_bytes());
+                    (contract_addr.unwrap(), [sig[0], sig[1], sig[2], sig[3]])
+                })
+        })
+        .collect::<EthereumCallFilter>();
+    match call_filter.contract_addresses_function_signatures.len() {
```
Jannis: And this

datasource/ethereum/src/block_stream.rs
```diff
@@ -1096,5 +1225,188 @@ fn create_log_filter_from_subgraph(manifest: &SubgraphManifest) -> EthereumLogFi
...
+                .is_some() && data_source.
+                source.
+                address.
+                is_some()
+        })
+        .map(|data_source| {
+            data_source.source.address.unwrap()
+        })
+        .collect::<EthereumBlockFilter>();
+    match block_filter.contract_addresses.len() {
```
Jannis: And this

datasource/ethereum/src/block_stream.rs
```diff
@@ -1096,5 +1225,188 @@ fn create_log_filter_from_subgraph(manifest: &SubgraphManifest) -> EthereumLogFi
...
+}
+
+fn parse_triggers(
+    log_filter_opt: Option<EthereumLogFilter>,
+    call_filter_opt: Option<EthereumCallFilter>,
+    block_filter_opt: Option<EthereumBlockFilter>,
+    descendant_block: &EthereumBlockWithCalls,
+) -> Result<Vec<EthereumTrigger>, Error> {
+    let mut triggers = Vec::new();
+    triggers.append(&mut parse_log_triggers(
```
Jannis: Isn't this basically
```rust
triggers.extend(parse_log_triggers(...));
```
? I know there's a difference but I'd think `extend` would be slightly more efficient and... prettier.

datasource/ethereum/src/block_stream.rs
```diff
@@ -1096,5 +1225,188 @@ fn create_log_filter_from_subgraph(manifest: &SubgraphManifest) -> EthereumLogFi
...
+            (receipt.transaction_hash, receipt.transaction_index.as_u64())
+        })
+        .collect::<HashMap<H256, u64>>();
+
+    // Ensure all `Call` and `Log` triggers have a transaction index
+    for trigger in triggers.iter() {
+        match trigger {
+            EthereumTrigger::Log(log) => {
+                if !tx_hash_indexes.get(&log.transaction_hash.unwrap()).is_some() {
+                    return Err(format_err!("Unable to determine transaction index for Ethereum log."))
```
Jannis: log -> event

datasource/ethereum/src/block_stream.rs
```diff
@@ -1096,5 +1225,188 @@ fn create_log_filter_from_subgraph(manifest: &SubgraphManifest) -> EthereumLogFi
...
+        });
+    
+    Ok(triggers)
+}
+
+fn parse_log_triggers(
+    log_filter: Option<EthereumLogFilter>,
+    block: &EthereumBlock
+) -> Vec<EthereumTrigger> {
+    if log_filter.is_none() {
```
Jannis: A slightly more elegant solution here would be
```rust
log_filter.map_or(vec![], |log_filter| {
    block
        .transaction_receipts
        ...
})
```

datasource/ethereum/src/block_stream.rs
```diff
@@ -1096,5 +1225,188 @@ fn create_log_filter_from_subgraph(manifest: &SubgraphManifest) -> EthereumLogFi
...
+
+fn parse_block_triggers(
+    block_filter: Option<EthereumBlockFilter>,
+    block: &EthereumBlockWithCalls,
+) -> Vec<EthereumTrigger> {
+    if block_filter.is_none() {
+        return vec![]
+    }
+    let call_filter = EthereumCallFilter::from(block_filter.unwrap());
+    let calls = &block.calls.clone().unwrap();
```
Jannis: We're not checking whether `block.calls` is `None`, like we're doing above in `parse_call_triggers`. What we want to do here seems to be:
```rust
block_filter.map_or(vec![], |block_filter| {
    let call_filter = EthereumCallFilter::from(block_filter);
    block.calls.map_or(vec![], |calls| {
        calls
           .iter()
           ...
    })
})
```

Jannis (changes requested):
I'll continue my review in the morning.

Left to review:

* EthereumAdapter
* EthereumCallFilter, EthereumBlockFilter, etc.
* EthereumBlockWithTriggers, EthereumCall, EthereumTrigger etc.
* RuntimeHost / WasmiModule changes

Everything I have seen so far is excellent and most of the remarks are minor.

graph/Cargo.toml
```diff
@@ -33,9 +33,9 @@ slog = { version = "2.2.3", features = ["release_max_level_debug", "max_level_tr
...
 slog-envlogger = "2.1.0"
 slog-term = "2.4.0"
-tiny-keccak = "1.0"
+tiny-keccak = "1.4.2"
 tokio = "0.1.11"
 tokio-executor = "0.1.5"
 tokio-retry = "0.2"
 tokio-timer = "0.2.7"
-web3 = "0.5.0"
+web3 = { git = "https://github.com/olibearo/rust-web3", branch = "trace-struct" }
```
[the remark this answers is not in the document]
olibearo: Last week I tried to switch to upstream, it generated a lot of compilation errors since upstream made a lot of changes to the `ethereum-type`s. I punted since it was late on Friday, I'll give this another shot today to answer your question

core/src/subgraph/validation.rs
```diff
@@ -0,0 +1,30 @@
+use graph::prelude::*;
+
+pub fn validate_manifest(manifest: SubgraphManifest) -> Result<SubgraphManifest, SubgraphRegistrarError> {
+    // Validate that the manifest has a `source` address in each data source
```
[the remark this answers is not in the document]
olibearo: Great idea!

datasource/ethereum/src/block_stream.rs
olibearo: 👍 I made this a pure function on the `BlockStreamContext`

datasource/ethereum/src/block_stream.rs
olibearo: It worked

datasource/ethereum/src/block_stream.rs
olibearo: My thought was that`append` avoids cloning the elements in the vector argument, will keep as is unless you object

datasource/ethereum/src/block_stream.rs
olibearo: Updated to use more combinators

Jannis:
After a lot of rebasing and restructuring, I feel this is getting close. Outstanding items (from my perspective):

- [ ] Avoid interrupting block fetching when beyond the reorg threshold.

  If a subgraph has an unconditional block handler, we're downloading every block for processing. The current number of blocks being fetched at once is set to 10,000. This takes longer than 15s and constantly gets interrupted and restarted after chain head updates.

  My idea how to fix this: don't let chain head updates interrupt this process if the subgraph is still processing blocks older than the reorg threshold.

- [ ] Test reprocessing the current block with new dynamic data sources. Basically verifying that the interplay of call triggers and dynamic data sources works.

- [ ] Add documentation to our hosted docs on the https://thegraph.com/docs/define-a-subgraph page (above or below dynamic data sources).

olibearo:
Jannis and I discussed over Slack, posting here for the record:

Avoiding interrupting the block fetching is a non issue since there isn't a mechanism for block fetching to to be interrupted.

The documentation updates are here: https://github.com/graphprotocol/docs/pull/67

datasource/ethereum/src/ethereum_adapter.rs
```diff
@@ -47,6 +47,80 @@ where
...
+                        .build(),
+                    _ => TraceFilterBuilder::default()
+                        .from_block(from.into())
+                        .to_block(to.into())
+                        .to_address(addresses.clone())
+                        .build(),
+                };
+
+                let logger = logger.clone();
+                let logger_1 = logger.clone();
```
Jannis: Throughout the code base we've been suffixing clones with their purpose (e.g. `logger_for_err`) instead of enumerating them.

datasource/ethereum/src/ethereum_adapter.rs
```diff
@@ -47,6 +47,80 @@ where
...
+                                    traces.len(),
+                                    from,
+                                    to
+                                );
+                            }
+                        }
+                        traces
+                    })
+                    .from_err::<EthereumContractCallError>()
+                    .from_err()
```
Jannis: I wonder if we need the double `from_err` here. Could be, worth double-checking?

datasource/ethereum/src/ethereum_adapter.rs
```diff
@@ -97,6 +171,40 @@ where
...
+            );
+        }
+
+        let eth = self.clone();
+        let logger = logger.to_owned();
+        stream::unfold(from, move |start| {
+            if start > to {
+                return None;
+            }
+            let end = (start + 200 - 1).min(to);
```
Jannis: This could use a comment — what does it mean that we pick 200 blocks here and why?

datasource/ethereum/src/ethereum_adapter.rs
```diff
@@ -153,103 +264,31 @@ where
...
+
+            let log_filter = log_filter.clone();
+            Some(
+                eth.logs_with_sigs(&logger, start, end, addresses.clone(), event_sigs.clone())
+                    .map(move |logs| {
+                        logs.into_iter()
+                            .filter(move |log| log_filter.matches(log))
+                            .collect()
+                    })
+                    .map(move |logs: Vec<Log>| (logs, new_start)),
```
Jannis: I'd move this `Vec<Log>` up and make it `collect::<Vec<Log>>`. It feels that's more obvious.

datasource/ethereum/src/ethereum_adapter.rs
```diff
@@ -327,7 +366,12 @@ where
         let net_version_future = retry("net_version RPC call", &logger)
             .no_limit()
             .timeout_secs(20)
-            .run(move || web3.net().version().map_err(SyncFailure::new).from_err());
+            .run(move || {
+                web3.net()
+                    .version()
+                    .from_err::<EthereumContractCallError>()
+                    .from_err()
```
Jannis: Again: double `from_err` — needed? (Probably)

datasource/ethereum/src/ethereum_adapter.rs
```diff
@@ -336,7 +380,7 @@ where
             .run(move || {
                 web3.eth()
                     .block(BlockNumber::Earliest.into())
-                    .map_err(SyncFailure::new)
+                    .from_err::<EthereumContractCallError>()
                     .from_err()
```
Jannis: I see more `from_err` x2 here. They are needed, huh?

datasource/ethereum/src/ethereum_adapter.rs
```diff
@@ -533,6 +577,35 @@ where
...
+                    web3.eth()
+                        .block(BlockId::Hash(descendant_block_hash))
+                        .from_err::<EthereumContractCallError>()
+                        .from_err()
+                        .map(|block_opt| block_opt.map(|block| block.parent_hash))
+                })
+                .map_err(move |e| {
+                    e.into_inner().unwrap_or_else(move || {
+                        format_err!(
+                            "Ethereum node took too long to return data for block hash = {}",
```
Jannis: `block hash = {}` -> ``` block hash `{}` ```

datasource/ethereum/src/ethereum_adapter.rs
```diff
@@ -533,6 +577,35 @@ where
         )
     }
 
+    fn block_parent_hash_by_block_hash(
```
Jannis: Should we just call this `parent_block_hash`?

datasource/ethereum/src/ethereum_adapter.rs
```diff
@@ -579,26 +652,199 @@ where
...
+                0 => vec![],
+                _ => trace_chunks.into_iter().flatten().collect(),
+            })
+            .and_then(move |traces| {
+                // `trace_stream` returns all of the traces for the block, and this
+                // includes a trace for the block reward which every block should have.
+                // If there are no traces something has gone wrong.
+                if traces.is_empty() {
+                    return future::err(format_err!(
+                        "Trace stream return no traces for block: number = {}, hash = {}",
```
Jannis: return -> returned

`number = {}, hash = {}` -> ``` number `{}`, hash `{}` ```

datasource/ethereum/src/ethereum_adapter.rs
```diff
@@ -579,26 +652,199 @@ where
...
+                    ));
+                }
+                // Since we can only pull traces by block number and we have
+                // all the traces for the block, we need to ensure that the
+                // block hash for the traces is equal to the desired block hash.
+                for trace in traces.iter() {
+                    if trace.block_hash != block_hash {
+                        return future::err(format_err!(
+                            "Trace stream returned traces for an unexpected block: \
+                             number = {}, hash = {}",
```
Jannis: `number = {}, hash = {}` -> ``` number `{}`, hash `{}` ```

datasource/ethereum/src/ethereum_adapter.rs
```diff
@@ -579,26 +652,199 @@ where
...
+                    return future::err(format_err!(
+                        "Trace stream return no traces for block: number = {}, hash = {}",
+                        block_number,
+                        block_hash,
+                    ));
+                }
+                // Since we can only pull traces by block number and we have
+                // all the traces for the block, we need to ensure that the
+                // block hash for the traces is equal to the desired block hash.
+                for trace in traces.iter() {
```
Jannis: Wouldn't checking the first trace be enough? Even if it's by number, it shouldn't return traces for multiple blocks?

datasource/ethereum/src/ethereum_adapter.rs
```diff
@@ -579,26 +652,199 @@ where
...
+                    to,
+                    call_filter_opt.unwrap(),
+                )));
+            }
+            if block_filter_opt.is_some() {
+                let block_filter = block_filter_opt.unwrap();
+                match block_filter.contract_addresses.len() {
+                    0 => (),
+                    _ => {
+                        let call_filter = EthereumCallFilter::from(block_filter);
```
Jannis: Can we add a comment here about why we are using a call filter for the block here?

datasource/ethereum/src/ethereum_adapter.rs
```diff
@@ -579,26 +652,199 @@ where
...
+                    }
+                }
+            }
+        }
+        Box::new(
+            future::join_all(block_futs).and_then(|block_pointer_chunks| {
+                let mut blocks = block_pointer_chunks
+                    .into_iter()
+                    .flatten()
+                    .collect::<Vec<EthereumBlockPointer>>();
```
Jannis: This could be made more elegant with itertools `unique` and `sorted_by` combinators. Either way, I'd do the deduplication before sorting because sorting is the more expensive operation.

datasource/ethereum/src/ethereum_adapter.rs
```diff
@@ -579,26 +652,199 @@ where
...
+                            }),
+                    )
+                })
+                .collect()
+            })
+            .map(move |mut block_pointers| {
+                block_pointers.reverse();
+                block_pointers
+            });
+        Box::new(block_ptrs)
```
Jannis: I'd drop the `let block_ptrs` assignment and instead do `Box::new(...)` right away. The fewer variables, the fewer things can be mutated. 😉

datasource/ethereum/src/ethereum_adapter.rs
```diff
@@ -617,11 +862,78 @@ where
...
+                    if !block_ptrs.contains(&block_ptr) {
+                        if let Some(prev) = block_ptrs.last() {
+                            assert!(prev.number < number);
+                        }
+                        block_ptrs.push(block_ptr);
+                    }
+                }
+                block_ptrs
+            });
+        Box::new(blocks)
```
Jannis: Same here (about `Box::new` vs `let` assignment).

datasource/ethereum/src/ethereum_adapter.rs
```diff
@@ -617,11 +862,78 @@ where
...
+                            .collect()
+                    }
+                }
+            })
+            .map(|calls| {
+                let mut block_ptrs = vec![];
+                for call in calls.iter() {
+                    let hash = call.block_hash;
+                    let number = call.block_number;
+                    let block_ptr = EthereumBlockPointer::from((hash, number));
```
Jannis: I'd pass `call.block_hash` and `call.block_number` to `EthereumBlockPointer::from` directly.

Jannis (changes requested):
Ok, the `EthereumAdapter` review ended up touching minor aspects only. That's good. 👍

leodasvacas:
The `LOG_STREAM_CHUNK_SIZE_IN_BLOCKS` and `LOG_STREAM_PARALLEL_CHUNKS` env vars are dead in this PR.

I don't think doing parallel requests for logs is a big performance win, and it would make the adaptive log range implementation more complicated, so I'm happy to see `LOG_STREAM_PARALLEL_CHUNKS`. We should remove the `lazy_static` and also remove it from `environment-variables.md`.

I'm not so confident about killing `LOG_STREAM_CHUNK_SIZE_IN_BLOCKS`, even with adaptive ranges. There may be other reasons for tweaking it, though I can't recall.

graph/src/components/ethereum/stream.rs
```diff
@@ -4,8 +4,15 @@ use futures::Stream;
 use crate::prelude::*;
 
 pub trait BlockStream:
-    Stream<Item = EthereumBlock, Error = Error> + EventConsumer<ChainHeadUpdate>
+    Stream<Item = EthereumBlockWithTriggers, Error = Error> + EventConsumer<ChainHeadUpdate>
 {
+    fn parse_triggers(
+        log_filter_opt: Option<EthereumLogFilter>,
```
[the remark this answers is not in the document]
olibearo: The indirection of having `parse_triggers` on the `BlockStream` trait and then implementing it as a method defined on the struct it has as a property feels off.

I see that the `SubgraphInstanceManager` needs access to that method through the `BlockStream` though. 

What if instead we create a struct `EthereumTriggerFilters`?

```rust
struct EthereumTriggerFilters {
    log_filter: Option<EthereumLogFilter>,
    call_filter: Option<EthereumCallFilter>,
    block_filter: Option<EthereumBlockFilter>,
}

impl EthereumTriggerFilters {
    fn parse_triggers(
        include_calls_in_block: bool,
        block: EthereumBlockWithCalls
    ) -> Result<EthereumBlockWithTriggers, Error>;
}
```

That way you can use it anywhere you have all the filters and the args, you won't need a reference to the `BlockStream`.

datasource/ethereum/src/ethereum_adapter.rs
olibearo: We need go from the web3 error -> graph-node error -> failure error here, since graph-node errors implement `Fail`. Let me know if you have other ideas on how to get there

graph/src/components/ethereum/stream.rs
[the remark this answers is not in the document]
Jannis: Sounds good to me!

datasource/ethereum/src/ethereum_adapter.rs
Jannis: I don't, just wanted to check.

datasource/ethereum/src/ethereum_adapter.rs
olibearo: Made this controllable by env var, still defaults to 200 though

datasource/ethereum/src/ethereum_adapter.rs
olibearo: Yeah, that _should_ be enough. I updated to reflect this.

datasource/ethereum/src/ethereum_adapter.rs
olibearo: https://doc.rust-lang.org/std/vec/struct.Vec.html#method.dedup

Dedup requires that the vector is sorted to perform a thorough dedup.

For now I'm okay with leaving this re: itertools methods, let me know if you object.

datasource/ethereum/src/ethereum_adapter.rs
```diff
@@ -678,15 +678,13 @@ where
...
-                for trace in traces.iter() {
-                    if trace.block_hash != block_hash {
-                        return future::err(format_err!(
-                            "Trace stream returned traces for an unexpected block: \
-                             number = `{}`, hash = `{}`",
-                            block_number,
-                            block_hash,
-                        ));
-                    }
+                if traces.iter().any(|trace| trace.block_hash != block_hash) {
```
Jannis: So much better 😄

datasource/ethereum/src/ethereum_adapter.rs
Jannis: I don't, it would be a minor improvement.

datasource/ethereum/src/ethereum_adapter.rs
Jannis: That's alright

Jannis:
@leodasvacas Are you happy with the `LOG_STREAM_CHUNK_SIZE_IN_BLOCKS` changes?

leodasvacas:
I'm happy.

Jannis (approved):

PR #828, opened 2019-03-15
Log error that terminated a mapping

leoyvens:


runtime/wasm/src/host.rs
```diff
@@ -353,11 +353,11 @@ impl RuntimeHostTrait for RuntimeHost {
                     )
                 })
                 .and_then(|_| {
-                    result_receiver.map_err(move |_| {
+                    result_receiver.map_err(move |e| {
                         format_err!(
-                            "Mapping terminated before finishing to handle \
-                             Ethereum event: {}",
+                            "Mapping terminated with error when handling event: `{}`, {}",
```
Jannis: Nit: I'd make it ```...when handling event `{}`: {}```. I think it makes things a _teeeeny_ bit more readable. I'm ok if you disagree though. 😉

Jannis (approved):

runtime/wasm/src/host.rs
leodasvacas: That's better, thanks!

PR #829, opened 2019-03-19
Make `first` optional again (defaulting to 100)

Jannis:
Making it required was a breaking change and should've prepared better.


graphql/src/schema/api.rs
```diff
@@ -411,13 +411,11 @@ fn collection_arguments_for_named_type(
     let mut skip = input_value(&"skip".to_string(), "", Type::NamedType("Int".to_string()));
     skip.default_value = Some(Value::Int(0.into()));
 
+    let mut first = input_value(&"first".to_string(), "", Type::NamedType("Int".to_string()));
```
leodasvacas: `first` can still be non-null since it has a default value.

PR #831, opened 2019-03-19
Recursively coerce input objects in GraphQL object coercion

### LINKED ISSUE #789: `BigInt` variables passed to queries return incorrect results
**This is a bug**

Right now, queries that use `BigInt` variables do not return correctly. Passing `"123"` into the query below (using Apollo) will return no results, but hard-coding `"123"` into the query will return results,

```
query myQuery($variable: BigInt!) {
  entities(where: { field: $variable }) {
    ...
  }
}
```

A minimal reproduction is available at: https://github.com/NoahZinsmeister/bigint-bug

leoyvens:
Fixes #789. That was affecting any variables in `where` because we were not coercing values nested in input objects. The adjusted coercion algorithm departs a bit from what's in the spec, but there are a few reasons for this, such as the spec assuming validation, which we don't have, and also the spec does not have pseudocode for coercing input objects.

This also introduces the `test-store` crate, which provides a static store that can be reused across test in different crates. Tests that had their own static stores were refactored to use it, and the query tests now use instead of the mock store. Testing the real store uncovered a bug with `derivedFrom` which is also fixed.

### DIFF
--- a/graphql/src/execution/execution.rs
+++ b/graphql/src/execution/execution.rs
@@ -305,7 +305,7 @@ where
     R1: Resolver,
     R2: Resolver,
 {
-    coerce_argument_values(ctx.clone(), object_type, field)
+    coerce_argument_values(&ctx, object_type, field)
         .and_then(|argument_values| {
             resolve_field_value(
                 ctx.clone(),
@@ -774,7 +774,7 @@ fn merge_selection_sets(fields: Vec<&q::Field>) -> q::SelectionSet {
 
 /// Coerces argument values into GraphQL values.
 pub fn coerce_argument_values<'a, R1, R2>(
-    ctx: ExecutionContext<'_, R1, R2>,
+    ctx: &ExecutionContext<'_, R1, R2>,
     object_type: &'a s::ObjectType,
     field: &q::Field,
 ) -> Result<HashMap<&'a q::Name, q::Value>, Vec<QueryExecutionError>>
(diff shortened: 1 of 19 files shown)


graphql/src/execution/execution.rs
```diff
@@ -894,59 +844,43 @@ pub fn coerce_variable_values(
...
-                    }
-                }
+        let value = variables
+            .as_ref()
+            .and_then(|vars| vars.get(&variable_def.name));
 
-                // A null value was provided, either use it or fail
-                Some(q::Value::Null) => {
-                    if let s::Type::NonNullType(_) = variable_def.var_type {
-                        errors.push(QueryExecutionError::InvalidVariableError(
```
Jannis: I think this was the only place where we were using `InvalidVariableError`, so we can probably remove this enum variant now.

graphql/src/values/coercion.rs
```diff
@@ -48,80 +51,134 @@ impl MaybeCoercible<ScalarType> for Value {
...
-            // Try to coerce Scalar values
-            (TypeDefinition::Scalar(t), value) => value.coerce(t),
+        // Everything else remains unimplemented
+        _ => None,
+    }
+}
 
-            // Try to coerce InputObject values
-            (TypeDefinition::InputObject(_), v) => match v {
-                Value::Object(_) => Some(self.clone()),
```
Jannis: Oh yeah, I can see how this was a bit too simple...

graphql/tests/query.rs
```diff
@@ -772,3 +733,36 @@ fn first_is_nullable() {
...
+          }
+        }
+    ",
+    )
+    .expect("invalid test query");
+
+    let result = execute_query_document_with_variables(
+        query,
+        Some(QueryVariables::new(HashMap::from_iter(
+            vec![(String::from("name"), q::Value::String("Lisa".to_string()))].into_iter(),
```
Jannis: What kind of value would we translate this to before? I wonder if we should use a `BigInt` here to really verify that this PR fixes the original issue.

Jannis (changes requested):
Just one small change and a question about the `nested_variable` test. This looks great otherwise!

graphql/src/execution/execution.rs
leodasvacas: Funny you commented on this because I also noticed that `InvalidVariableError` was dead, but didn't realize I made it so. The error message for this case got worse, and in general the coercion errors are not specific, but that's not something we want to solve now. I removed the dead variant.

graphql/tests/query.rs
leodasvacas: Before this, `$name` would be `Value::Variable` after coercion. So it really has nothing to do with the specific type. I tested manually for `BigInt`. I'm feeling slightly lazy about it but I can also modify the test data here to include some `BigInt` if you feel it's worth it.

Jannis: It's ok, let's move on :)

Jannis (approved):

Issue #803, opened 2019-03-06
Test `multiple_data_sources_per_subgraph` is not really passing


leoyvens:
It's printing an error despite showing up green. The error: `called Result::unwrap() on an Err value: SpawnError { is_shutdown: true }`. This is most likely just an issue with the test and not an actual bug.

perfectmak:
@leodasvacas can I look into this? I'm new here and I really like what graph-node is doing, so I feel fixing some test case would help me get better know how iit works intenally.

leodasvacas:
@perfectmak If you want to take a look you're more than welcome!

PR #832, opened 2019-03-20
Dynamic data sources

### LINKED ISSUE #719: Dynamic data sources
# Requirements

1. Can be initiated by mappings.
2. Allows to dynamically add data sources to the subgraph at runtime.
3. Dynamic data sources are persisted and started/stopped with subgraph deployments.
4. Number of dynamic data sources is limited per subgraph.

# Proposed solution

1. Add a `templates` field to the subgraph manifest under `dataSource`:
    1. `templates` is a list of templates.
    2. Each template is a data source, with one or more fields left out (e.g. the contract address or IPFS file hash). Each of these data sources has a unique name (e.g. `ExampleDataSource`).
2. Add an API to `graph-ts` to instantiate data sources from templates on demand.
    1. Add a generic, low-level API like `subgraph.createDataSource(name: string, params: string[]): void`
    2. Generate bindings for subgraph templates via `graph codegen`. Usage example:
       ```
       import { ExampleDataSource } from './generated/templates/DataSource'
       ...
             
       ExampleDataSource.create(address)
       ```
       The parameters to `create` depend on the type of the data source template. E.g. an IPFS file data source would take an IPLD link or IPFS hash, an `ethereum/contract` data source would take just the contract address.
3. When a data source template is instantiated, add the data source to the subgraph:
   1. Write the data source to the store in the form of entities, under the corresponding subgraph deployment.
   2. Update the block stream so it
        a) Rescans the current block for events for the new data source
        b) Includes events for the new data source going forward

# Tasks

- [x] Graph CLI: Validate data sources templates in the manifest.
  - [x] Add data source templates to the subgraph manifest schema
  - [x] Validate templates based on the schema
  - [x] Validate event names in templates
  - [x] Validate invalid ABI references in templates
  - [x] Validate absence of handlers in templates
- [x] Graph CLI: Generate classes for instantiating data sources from templates.
- [x] Graph CLI: Compile and upload data source templates
  - [x] Write/upload template ABIs
  - [x] Compile/write/upload template mappings
- [x] Graph TS: Add generic data source instantiation API.
- [x] Graph Node: Parse templates in subgraph manifests.
- [x] Graph Node: Write template entities as part of the subgraph meta data.
- [x] Graph Node: Instantiate data sources by adding the data source to the subgraph entities.
- [x] Graph Node: Update the block stream to include events for the new data source from the current block and from future blocks.
- [x] Graph Node: Add an environment variable to limit the number of dynamic data sources per subgraph. Pick a reasonable default.
- [x] Docs: Add documentation for dynamic data sources

Jannis:
Resolves #719 (or rather, makes the graph-node changes needed for it).

Goes along with the following two PRs for graph-ts and graph-cli:

* https://github.com/graphprotocol/graph-ts/pull/44
* https://github.com/graphprotocol/graph-cli/pull/234

## Defining a dynamic data source

Dynamic data sources are defined as templates in the manifest:

```yaml
# ...
dataSources:
  - kind: ethereum/contract
    name: Factory
    # ...
    templates:
      - name: Exchange
        kind: ethereum/contract
        network: mainnet
        source:
          abi: Exchange
        mapping:
          kind: ethereum/events
          apiVersion: 0.0.1
          language: wasm/assemblyscript
          file: ./src/mappings/exchange.ts
          entities:
            - Exchange
            - Transaction
            - User
            - UserExchangeBalance
          abis:
            - name: Exchange
              file: ./abis/exchange.json
          eventHandlers:
            - event: TokenPurchase(address,uint256,uint256)
              handler: handleTokenPurchase
            - event: EthPurchase(address,uint256,uint256)
              handler: handleEthPurchase
            - event: AddLiquidity(address,uint256,uint256)
              handler: handleAddLiquidity
            - event: RemoveLiquidity(address,uint256,uint256)
              handler: handleRemoveLiquidity
```

## Creating a dynamic data source

Dynamic data sources are imported from the generated code and used as follows:

```typescript
import { Exchange as ExchangeTemplate } from "../../generated/Factory/templates"

...

// Start indexing the exchange
ExchangeTemplate.create(event.params.exchange)
```

## How dynamic data sources are implemented

1. When starting a subgraph, the node now not only loads the manifest, it also loads the dynamic data sources from the store and fetches their files from IPFS. It then injects the data source instances into the subgraph instance before starting the indexing.

2. Whenever we we process a block and mappings request new data sources, these data sources are collected and, after having processed the block, are instantiated from templates. We then process the current block again but only with those new data sources. The entity operations from this are merged into the ones we already have for the block. After that, the dynamic data sources are persisted by adding the data sources to the subgraph instance and by adding entity operations to store them in the db (for 1.).

3. If new data sources have been added in a block, the block stream is restarted to incorporate events that are relevant for the new data sources going forward. We figured this would be easier to do (and it is) than modifying the block stream while it is already running.

4. The dynamic data sources in the database can be queried through the "subgraph of subgraphs":
   ```graphql
   {
     subgraphDeployments {
       manifest {
         dataSources { ... }
       }
       dynamicDataSources {
         ...
       }
     }
   }
   ```

## Todos

- [x] Store block hash and block number of when a dynamic data source was created
- [x] Add an environment variable to limit the number of dynamic data sources per subgraph. Pick a reasonable default.
- [x] Fix the test suite
- [x] Add unit tests for `dataSource.create` host export
- [ ] Add other tests


graph/src/data/subgraph/schema.rs
```diff
@@ -591,6 +616,80 @@ impl From<super::MappingEventHandler> for EthereumContractEventHandlerEntity {
...
+        }
+    }
+}
+
+#[derive(Clone, Debug, Hash, Eq, PartialEq)]
+struct EthereumContractDataSourceTemplateSourceEntity {
+    abi: String,
+}
+
+impl TypedEntity for EthereumContractDataSourceTemplateSourceEntity {
```
lutter: Echoing what I said in Discord earlier, I had to read this chunk three times to convince myself that it was about exactly two new entities, `EthereumContractDataSourceTemplateSourceEntity` and  `EthereumContractDataSourceTemplateEntity`. It's incredibly hard to keep the two apart, and we should figure out some naming scheme that makes these names a little easier to read and tell apart. Maybe something like `EthCDSTemplateEntity` and `EthCDSTemplateSourceEntity`; I am not in love with these either, but at least I can now see easily that they are different length.

graph/src/data/subgraph/schema.rs
```diff
@@ -591,6 +616,80 @@ impl From<super::MappingEventHandler> for EthereumContractEventHandlerEntity {
     }
 }
 
+#[derive(Debug)]
+struct EthereumContractDataSourceTemplateEntity {
```
lutter: It would be good to have a comment here how these entities are used and how they fit into the overall picture of dynamic data sources. Same for `EthereumContractDataSourceTemplateSourceEntity`

graph/src/data/subgraph/mod.rs
```diff
@@ -420,16 +425,17 @@ impl UnresolvedMapping {
...
     pub name: String,
     pub source: Source,
     pub mapping: M,
+    pub templates: Option<Vec<T>>,
 }
 
-pub type UnresolvedDataSource = BaseDataSource<UnresolvedMapping>;
-pub type DataSource = BaseDataSource<Mapping>;
+pub type UnresolvedDataSource = BaseDataSource<UnresolvedMapping, UnresolvedDataSourceTemplate>;
+pub type DataSource = BaseDataSource<Mapping, DataSourceTemplate>;
```
lutter: If `UnresolvedMapping` always goes with `UnresolvedDataSourceTemplate`, and `Mapping` always goes with `DataSourceTemplate`, wouldn't it be better if the `templates` lived on the `Mapping`?

store/postgres/src/subgraphs.graphql
```diff
@@ -77,3 +78,17 @@ type EthereumContractEventHandler @entity {
...
+type EthereumContractDataSourceTemplate @entity {
+    id: ID!
+    kind: String!
+    name: String!
+    network: String
+    source: EthereumContractSourceTemplate!
+    mapping: EthereumContractMapping!
+}
+
+type EthereumContractSourceTemplate @entity {
```
lutter: Shouldn't that be `EthereumContractDataSourceTemplateSource` to follow the same naming as the internal names ? :stuck_out_tongue_winking_eye:

store/postgres/src/subgraphs.graphql
```diff
@@ -47,6 +48,19 @@ type EthereumContractDataSource @entity {
     network: String
     source: EthereumContractSource!
     mapping: EthereumContractMapping!
+    templates: [EthereumContractDataSourceTemplate!]
+}
+
+type DynamicEthereumContractDataSource @entity {
+    # TODO: Add created time and block here
```
Jannis: This is one thing I'd like to add still.

Jannis:
@graphprotocol/developers Due to the number of commits and changes here, I'd almost recommend going through commit by commit. I've spent a few hours structuring these so they kinda tell a story and should be easy to follow.

Jannis:
Not yet, but fixing the test suite is the next thing on my list.

graph/src/data/graphql/values.rs
```diff
@@ -0,0 +1,84 @@
+use failure::Error;
+use graphql_parser::query::Value;
+use std::collections::BTreeMap;
+use std::str::FromStr;
+
+use crate::web3::types::H160;
+
+pub trait TryFromValue: Sized {
```
leodasvacas: We can't replace this with `TryFrom`?

runtime/wasm/src/host_exports.rs
```diff
@@ -549,6 +552,53 @@ where
         BigDecimal::from_str(&s)
             .map_err(|e| HostExportError(format!("failed to parse BigDecimal: {}", e)))
     }
+
+    pub(crate) fn data_source_create(
+        &self,
+        ctx: &mut EventHandlerContext,
+        name: String,
+        params: Vec<String>,
+    ) -> Result<bool, HostExportError<impl ExportError>> {
```
leodasvacas: In what case does this return `Ok(false)`?

runtime/wasm/src/module/mod.rs
```diff
@@ -93,7 +94,8 @@ where
     ) -> Result<Self, FailureError> {
         let logger = logger.new(o!("component" => "WasmiModule"));
 
-        let parsed_module = config.data_source.mapping.runtime;
+        // FIXME: Cloning this may be expensive.
+        let parsed_module = config.data_source.mapping.runtime.clone();
```
leodasvacas: Same note as above about putting this in an `Arc` so we can clone it.

runtime/wasm/src/module/mod.rs
```diff
@@ -793,6 +798,21 @@ where
             .big_decimal_equals(self.asc_get(x_ptr), self.asc_get(y_ptr));
         Ok(Some(RuntimeValue::I32(if equals { 1 } else { 0 })))
     }
+
+    /// function dataSource.create(name: string, params: Array<string>): void
```
leodasvacas: The comment says this returns void but we're returning something in the function.

store/postgres/src/store.rs
```diff
@@ -890,9 +890,12 @@ impl StoreTrait for Store {
             panic!("transact_block_operations must transact a single block only");
         }
 
-        // All operations should apply only to entities in this subgraph
+        // All operations should apply only to entities in this subgraph (or
+        // the subgraph of subgraphs)
         for op in &operations {
-            if op.entity_key().subgraph_id != subgraph_id {
+            if op.entity_key().subgraph_id != subgraph_id
+                && *(op.entity_key().subgraph_id) != "subgraphs"
```
leodasvacas: I don't think you need parenthesis here.

leoyvens (changes requested):
You did it! Comments are about possible refactorings.

core/src/subgraph/instance.rs
```diff
@@ -1,32 +1,47 @@
...
     /// The runtime hosts are created and added in the same order the
     /// data sources appear in the subgraph manifest. Incoming block
     /// stream events are processed by the mappings in this same order.
     hosts: Vec<Arc<T::Host>>,
+
+    /// Builder for runtime hosts.
+    host_builder: T,
+
+    /// Dynamic data sources.
+    ethereum_log_filter: EthereumLogFilter,
```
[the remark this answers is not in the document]
Jannis: Putting it on the builder feels too implicit. And since the builder is cloned for each subgraph, clearing the filter would be too hidden too, IMHO.

But I agree, this shouldn't be on the subgraph instance. I've simplified this now, so there's only _one_ log filter ever and it's on the `SubgraphState`. From there it is passed to the `BlockStreamBuilder` only when it is time to build a new stream. See the new commit in this PR:

https://github.com/graphprotocol/graph-node/commit/5ad7f4db6e9a969ba655e0cdcd864236600540e3

core/src/subgraph/instance.rs
```diff
@@ -1,32 +1,47 @@
...
 {
     fn from_manifest(
         logger: &Logger,
         manifest: SubgraphManifest,
         host_builder: T,
     ) -> Result<Self, Error> {
-        // Create a new runtime host for each data source in the subgraph manifest;
-        // we use the same order here as in the subgraph manifest to make the
-        // event processing behavior predictable
-        let manifest_id = manifest.id;
```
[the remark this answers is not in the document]
Jannis: Oops. No reason at all, I've put it back.

core/src/subgraph/instance.rs
```diff
@@ -82,17 +99,49 @@ where
...
+    }
+
+    fn add_dynamic_data_sources(
+        &mut self,
+        data_sources: Vec<DataSource>,
+        runtime_hosts: Vec<Arc<T::Host>>,
+    ) -> Result<(), Error> {
+        // Protect against creating more than the allowed maximum number of data sources
+        if let Some(max_data_sources) = *MAX_DATA_SOURCES {
+            if self.hosts.len() + runtime_hosts.len() > max_data_sources {
```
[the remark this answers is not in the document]
Jannis: It's a 1:1 mapping right now. I agree it could maybe be more obvious but exposing "max data sources" feels more intuitive / easier to communicate to users.

core/src/subgraph/instance_manager.rs
```diff
@@ -1,17 +1,40 @@
...
 
-type InstanceShutdownMap = Arc<RwLock<HashMap<SubgraphDeploymentId, CancelGuard>>>;
+struct SubgraphState<B, S, T>
+where
+    B: BlockStreamBuilder,
+    S: Store + ChainStore,
+    T: RuntimeHostBuilder,
+{
+    pub deployment_id: SubgraphDeploymentId,
+    pub instance: Arc<RwLock<SubgraphInstance<T>>>,
```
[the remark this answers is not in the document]
Jannis: I'm not sure how that would work. We need to read and write from the subgraph instance (read to call `matches_log` and `process_log`, write to add dynamic runtime hosts). And since there's async processing where we send the subgraph instance to another Tokio task for processing logs, I don't see an easy way around `Arc<RwLock` or `Arc<Mutex<`.

core/src/subgraph/instance_manager.rs
```diff
@@ -120,187 +147,78 @@ impl SubgraphInstanceManager {
...
+        // the form
+        // ```
+        // vec![
+        //   ("DataSource1", "Template1", <template struct>),
+        //   ("DataSource2", "Template1", <template struct>),
+        //   ("DataSource2", "Template2", <template struct>),
+        // ]
+        // ```
+        // for easy filtering later
+        let mut templates: Vec<(String, DataSourceTemplate)> = vec![];
```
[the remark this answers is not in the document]
Jannis: Good point, I'm updating the comment.

core/src/subgraph/instance_manager.rs
```diff
@@ -120,187 +147,78 @@ impl SubgraphInstanceManager {
...
+            logger,
+        };
+
+        // Keep restarting the subgraph until it terminates. The subgraph
+        // will usually only run once, but is restarted whenever a block
+        // creates dynamic data sources. This allows us to recreate the
+        // block stream and include events for the new data sources going
+        // forward; this is easier than updating the existing block stream.
+        tokio::spawn(loop_fn(subgraph_state, move |subgraph_state| {
+            run_subgraph(subgraph_state)
```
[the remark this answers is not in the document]
Jannis: `SubgraphInstance` is mutated when `add_dynamic_data_sources` is called.

I've added a commit to try to make this more obvious, splitting up `SubgraphState` into a context with inputs and mutable state. Let me know what you think: https://github.com/graphprotocol/graph-node/commit/a763a6cb14c44db0db85e30aaa05a968f865ec22

Note: I should be able to take this out again easily if you don't like it.

core/src/subgraph/instance_manager.rs
```diff
@@ -317,3 +235,528 @@ impl EventConsumer<SubgraphAssignmentProviderEvent> for SubgraphInstanceManager
...
+                    .read()
+                    .unwrap()
+                    .process_log(&logger, block, transaction, log, block_state)
+                    .map(|block_state| block_state)
+                    .map_err(|e| format_err!("Failed to process event: {}", e))
+            })
+        })
+}
+
+fn process_logs_in_runtime_hosts<R>(
```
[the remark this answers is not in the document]
Jannis: Good idea. I didn't think it would be easy but it was. First time I've defined a static method on a trait, I think:
```
SubgraphInstance::<T>::process_log_in_runtime_hosts(...)
```
Fun!

Here's the commit that I added: https://github.com/graphprotocol/graph-node/commit/3808bb5082b56982142f4a0761cda51fad2b0506

core/src/subgraph/instance_manager.rs
```diff
@@ -317,3 +235,528 @@ impl EventConsumer<SubgraphAssignmentProviderEvent> for SubgraphInstanceManager
...
+}
+
+fn persist_dynamic_data_sources<B, S, T>(
+    logger: Logger,
+    subgraph_state: SubgraphState<B, S, T>,
+    mut block_state: BlockState,
+    data_sources: Vec<DataSource>,
+    runtime_hosts: Vec<Arc<T::Host>>,
+    block_ptr: EthereumBlockPointer,
+) -> impl Future<Item = (SubgraphState<B, S, T>, BlockState, bool), Error = Error>
```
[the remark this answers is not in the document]
Jannis: Done

core/src/subgraph/loader.rs
```diff
@@ -0,0 +1,247 @@
...
+                    .and_then(move |query| self3.query_dynamic_data_sources(deployment_id2, query))
+                    .and_then(move |query_result| {
+                        self4.parse_data_sources(deployment_id3, query_result)
+                    })
+                    .and_then(move |typed_entities| {
+                        self5.convert_to_unresolved_data_sources(typed_entities)
+                    })
+                    .and_then(move |unresolved_data_sources| {
+                        self6.resolve_data_sources(unresolved_data_sources)
+                    })
```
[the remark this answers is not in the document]
Jannis: I tried that but it only collapses `parse_data_sources` and `convert_to_unresolved_data_sources` and ultimately makes the code less uniform that it looks right now:
image

core/src/subgraph/loader.rs
```diff
@@ -0,0 +1,247 @@
...
+                    .map(move |data_sources| {
+                        if data_sources.is_empty() {
+                            future::Loop::Break(state)
+                        } else {
+                            state.skip += dbg!(data_sources.len() as i32);
```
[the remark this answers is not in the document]
Jannis: That's why it _is_ awesome at all, because you can plug it in anywhere and the value is both printed and passed through.

core/src/subgraph/loader.rs
```diff
@@ -0,0 +1,247 @@
...
+    pub fn new(store: Arc<S>, link_resolver: Arc<L>, graphql_runner: Arc<Q>) -> Arc<Self> {
+        Arc::new(Self {
+            store,
+            link_resolver,
+            graphql_runner,
+        })
+    }
+
+    fn dynamic_data_sources_query(
+        self: Arc<Self>,
```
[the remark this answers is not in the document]
Jannis: Done

core/src/subgraph/loader.rs
```diff
@@ -0,0 +1,247 @@
...
+                deployment_id,
+                e
+            )
+        })
+    }
+
+    fn convert_to_unresolved_data_sources(
+        self: Arc<Self>,
+        entities: Vec<EthereumContractDataSourceEntity>,
+    ) -> impl Future<Item = Vec<UnresolvedDataSource>, Error = Error> + Send {
```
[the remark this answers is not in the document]
Jannis: Done

datasource/ethereum/src/block_stream.rs
```diff
@@ -1005,6 +1002,8 @@ pub struct BlockStreamBuilder<S, C, E> {
     eth_adapter: Arc<E>,
     node_id: NodeId,
     reorg_threshold: u64,
+    manifest_log_filter: Option<EthereumLogFilter>,
+    data_sources_log_filter: Option<EthereumLogFilter>,
```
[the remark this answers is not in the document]
Jannis: They are both gone now. 😛

graph/src/data/graphql/values.rs
Jannis: I didn't do it because having a fixed `Error` type for all `TypedFromEntity` implementations seemed convenient. I'll try `TryFrom` again.

Jannis: As discussed in person, we're going to keep this for now, as it would be a lot of hassle to implement `TryFrom` in situations where both the source and target type are external, like with
```rust
impl TryFrom<&serde_json::Value> for String
```

runtime/wasm/src/host_exports.rs
Jannis: Never really, let's make this return `void`. Any error here is fatal.

runtime/wasm/src/module/mod.rs
Jannis: Done

runtime/wasm/src/host_exports.rs
Jannis: Done

runtime/wasm/src/module/mod.rs
Jannis: Changed to _actually_ return `void` now.

store/postgres/src/store.rs
Jannis: Correct!

Jannis:
@leodasvacas I _think_ I've addressed all your comments.

Jannis:
Don't know what's up with Travis, but that test is known to not be reliable (see #803).

leodasvacas:
@Jannis we can `#[ignore]` that test for now.

core/src/subgraph/instance.rs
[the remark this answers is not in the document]
leodasvacas: I see. I think 1:1 data source to mapping is a good model to keep going forward, so this is fine as it is.

core/src/subgraph/instance_manager.rs
```diff
@@ -317,3 +258,491 @@ impl EventConsumer<SubgraphAssignmentProviderEvent> for SubgraphInstanceManager
...
+        .fold(block_state, move |block_state, log| {
+            let logger = logger.clone();
+            let instance = instance.clone();
+            let block = block.clone();
+
+            let instance = instance.read().unwrap();
+
+            instance
+                .process_log(&logger, block, log, block_state)
+                .map(|block_state| block_state)
```
leodasvacas: This is a noop.

core/src/subgraph/instance_manager.rs
```diff
@@ -317,3 +235,528 @@ impl EventConsumer<SubgraphAssignmentProviderEvent> for SubgraphInstanceManager
...
+                    logger_for_restart_check,
+                    "New data sources added, restart the subgraph",
+                );
+                Ok(false)
+            } else {
+                Ok(true)
+            }
+        })
+        // Process blocks from the stream as long as no restart is needed
+        .fold(subgraph_state, move |subgraph_state, block| {
```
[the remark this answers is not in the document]
Jannis: Yeah, I can see that. I can see if returning `needs_restart` from `process_block` makes things more obvious.

core/src/subgraph/loader.rs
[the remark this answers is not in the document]
leodasvacas: Alright, this is cleaner as it is.

core/src/subgraph/loader.rs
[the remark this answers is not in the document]
leodasvacas: Sure it's awesome for debugging, but do we want to print anything without going through a logger? Afaik this can potentially get garbled with log messages.

core/src/subgraph/instance_manager.rs
Jannis: Hehe, oops.

Jannis: Fixed

core/src/subgraph/loader.rs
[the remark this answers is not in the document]
Jannis: Oh, yeah, I wasn't arguing it should remain there. 😆 In fact, I thought I had removed it already. Done now.

core/src/subgraph/instance_manager.rs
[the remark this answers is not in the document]
leodasvacas: This split is nice for readability, and thanks for the commit links!

core/src/subgraph/instance_manager.rs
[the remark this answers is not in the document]
leodasvacas: Nice!

core/src/subgraph/instance.rs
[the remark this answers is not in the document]
leodasvacas: Nice solution!

core/src/subgraph/instance_manager.rs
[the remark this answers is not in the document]
leodasvacas: This one wasn't addressed but it's fine, it's not clear how to improve it and what we have now is pretty good once you figure it out.

core/src/subgraph/instance_manager.rs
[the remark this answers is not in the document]
leodasvacas: This `Arc<RwLock<` is the one thing I still have a gripe with, I think we can remove it entirely, I'll make a new review.

core/src/subgraph/instance_manager.rs
```diff
@@ -317,3 +258,490 @@ impl EventConsumer<SubgraphAssignmentProviderEvent> for SubgraphInstanceManager
...
+    let logger1 = logger.clone();
+    let logger2 = logger.clone();
+    let logger3 = logger.clone();
+    let logger4 = logger.clone();
+
+    // Process events one after the other, passing in entity operations
+    // collected previously to every new event being processed
+    process_logs(
+        logger.clone(),
+        ctx.state.instance.clone(),
```
leodasvacas: afaict this is the one place that makes us need the `Arc<RwLock` around the instance. If we instead keep the pattern we use everywhere else by having `process_logs` take the ctx by value and then return it, I believe we can mutate the instance without needing synchronization.

leoyvens (changes requested):

core/src/subgraph/instance_manager.rs
Jannis: Aaaaand done 😀

leodasvacas: awesome

core/src/subgraph/instance_manager.rs
```diff
@@ -322,16 +328,28 @@ where
...
+                if needs_restart {
+                    needs_restart_set.store(true, Ordering::SeqCst);
+                }
+                ctx
+            })
         })
         // A restart is needed
         .and_then(move |mut ctx| {
+            // Reset the needs restart flag and increase the restart counter
+            needs_restart_reset.store(false, Ordering::SeqCst);
```
leodasvacas: Must this be reset here? We won't check it again until it's re-instantiated, right?

core/src/subgraph/instance_manager.rs
```diff
@@ -322,16 +328,28 @@ where
         })
         // Process blocks from the stream as long as no restart is needed
         .fold(ctx, move |ctx, block| {
+            let needs_restart_set = needs_restart_set.clone();
```
leodasvacas: This declaration looks dead.

core/src/subgraph/instance_manager.rs
```diff
@@ -379,9 +397,8 @@ fn process_block<B, S, T>(
     logger: Logger,
     ctx: IndexingContext<B, S, T>,
     block_stream_cancel_handle: CancelHandle,
-    needs_restart: Arc<AtomicBool>,
     block: EthereumBlock,
-) -> impl Future<Item = IndexingContext<B, S, T>, Error = CancelableError<Error>>
+) -> impl Future<Item = (IndexingContext<B, S, T>, bool), Error = CancelableError<Error>>
```
leodasvacas: Can we document the meaning of the bool above the function?

leoyvens (changes requested):
So close!

core/src/subgraph/instance_manager.rs
Jannis: Erm, no! I'm removing it.

core/src/subgraph/instance_manager.rs
Jannis: I think it's needed to get the ownership moves right with the two nested closures.

leodasvacas: _maybe_ you can drop the `.clone()`/

core/src/subgraph/instance_manager.rs
Jannis: And it's gone

core/src/subgraph/instance_manager.rs
Jannis: I think I just tried all possible combinations of not having the extra assignment or having it, with and without cloning the outer var or the inner var or both... nope :-/

core/src/subgraph/instance_manager.rs
Jannis: Done

core/src/subgraph/loader.rs
```diff
@@ -0,0 +1,245 @@
...
+        // Obtain the "subgraphs" schema
+        let schema = self.store.subgraph_schema(&SUBGRAPHS_ID)?;
+
+        // Construct a query for the subgraph deployment and all its
+        // dynamic data sources
+        Ok(Query {
+            schema,
+            document: parse_query(
+                r#"
+                query deployment($id: ID!, $skip: Int!) {
```
lutter: I really like the idea of using GraphQL here, makes the code much clearer than if this was implemented using store primitives. I am a little concerned about performance, since it's fairly deeply nested: do we have any data on how long this query will take with production-like amounts of data?

core/src/subgraph/instance_manager.rs
leodasvacas: Then this is it, thanks for trying it out.

core/src/subgraph/loader.rs
Jannis: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=a4802e01eeb5ea5ee699ac90c49d1cfa

leodasvacas: It looks like we should be fine in terms of number of SQL queries, the first two levels are by id, I think the worst path is `dynamicDataSources*templates*eventHandlers` which doesn't look like it will misbehave too much, probably not the dominating factor in starting a subgraph compared to resolving from IPFS.

Jannis: I ran a few manual SQL queries for parts of this query and they were all fast. This is no guarantee but I think I want to try this — on staging first, of course.

core/src/subgraph/loader.rs
```diff
@@ -0,0 +1,245 @@
...
+            .run_query(query)
+            .map_err(move |e| format_err!("Failed to load manifest `{}`: {}", deployment, e))
+    }
+
+    fn parse_data_sources(
+        &self,
+        deployment_id: SubgraphDeploymentId,
+        query_result: QueryResult,
+    ) -> Result<Vec<EthereumContractDataSourceEntity>, Error> {
+        let data = match query_result.data.expect("subgraph deployment not found") {
```
leodasvacas: I was going try this PR out, but I had set the query complexity limit at 2 million, the data sources query is just over that limit, triggering this panic.

I don't think having to set a limit over 2 million is an issue (at least not one we have to deal with now), but it seems we shouldn't be panicking here.

leoyvens (changes requested):

core/src/subgraph/loader.rs
leodasvacas: We should however validate a that query complexity satifies a minimum to make sure it doesn't interfere with this, the complexity of that query is 2030100.

Jannis: True, we should fail the subgraph instead.

Jannis: We may want a flag to bypass the complexity, because there's never a guarantee that, if we put in that number as the complexity minimum, we won't forget to keep it up to date in the future.

Jannis: @leodasvacas How about we add
```
    fn run_query_with_complexity(
        &self,
        query: Query,
        max_complexity: Option<u64>,
    ) -> QueryResultFuture {
        let result = execute_query(
            &query,
            QueryExecutionOptions {
                logger: self.logger.clone(),
                resolver: StoreResolver::new(&self.logger, self.store.clone()),
                deadline: GRAPHQL_QUERY_TIMEOUT.map(|t| Instant::now() + t),
                max_complexity: max_complexity,
                max_depth: *GRAPHQL_MAX_DEPTH,
            },
        );
        Box::new(future::ok(result))
    }
```
to `GraphQlRunner` for internal queries?

Jannis: I've added this in the latest commit: https://github.com/graphprotocol/graph-node/pull/832/commits/5cfc3e101b7f059f06d0eaa3bae80dd57c47d375

Jannis: And the subgraph now fails properly if the dynamic data sources query fails, instead of panicking.

leodasvacas: @Jannis that's smarter, great!

leodasvacas:
@Jannis you need to fix the test suite.

Jannis:
Nooooo, not again! Sure, will do that in the morning.

Jannis:
@lutter All green now 🎉

leoyvens (approved):

PR #833, opened 2019-03-21
Add 'ipfs.map' host export

lutter:
This is still a work in progress (in particular, I haven't tested this yet) but since it does some funky stuff with types, in particular adding `Sync + Send` in various places, I would appreciate comments on the general approach.


leodasvacas:
Gave a look, looking good!

lutter:
This implements the first pass through for an `ipfs.map` host function. The only thing missing to close out the graph-node side of issue #833 is removing the restriction that JSON objects in the input must be separated by newlines.

graph/src/components/link_resolver.rs
```diff
@@ -53,19 +77,158 @@ impl LinkResolver for ipfs_api::IpfsClient {
...
     }
+
+    fn json_stream(
+        &self,
+        link: &Link,
+    ) -> Box<Future<Item = ValueStream, Error = failure::Error> + Send + 'static> {
+        // Discard the `/ipfs/` prefix (if present) to get the hash.
+        let path = link.link.trim_start_matches("/ipfs/").to_owned();
+        let mut stream = self.cat(&path).fuse();
+        let mut buf = BytesMut::with_capacity(1024);
```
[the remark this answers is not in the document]
lutter: I _think_ that's ok, because we are going to create a bunch of objects in memory that represent the entire file anyway, hoping it will be some constant times `GRAPH_MAX_IPFS_MAP_FILE_SIZE`, so adding 1 to that constant doesn't strike me as that bad, especially since a file with one line that's a few hundred MB seems pretty unlikely.

runtime/wasm/src/module/mod.rs
```diff
@@ -258,14 +263,39 @@ where
             )
         })
     }
+
+    pub(crate) fn handle_json_callback(
+        mut self,
+        handler_name: &str,
+        value: &graph::serde_json::Value,
+    ) -> Result<Vec<EntityOperation>, FailureError> {
+        self.start_time = Instant::now();
```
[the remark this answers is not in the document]
lutter: Yes, you are right. Not sure what I was thinking.

runtime/wasm/src/host_exports.rs
```diff
@@ -335,6 +343,48 @@ where
...
+                                ctx.clone(),
+                            )?;
+                            module.handle_json_callback(&*callback, &v)
+                        })
+                        .collect()
+                })
+                .map_err(|e| HostExportError(e.to_string())),
+        )?;
+        // Collect all results into one Vec
+        Ok(results.into_iter().flatten().collect())
```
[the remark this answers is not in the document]
lutter: I am confused by your 'fail fast' comment: I thought the `collect` on line 380 would abort processing the stream as soon as an error happens, and we would abort the whole `block_on` as soon as the `and_then` on line 373 encounters an error. Is that not correct?

graph/src/components/link_resolver.rs
```diff
@@ -53,19 +77,158 @@ impl LinkResolver for ipfs_api::IpfsClient {
...
+        // We need that to adjust the line number in error messages from serde_json
+        // to translate from line numbers in the snippet we are deserializing
+        // to the line number in the overall file
+        let mut count = 0;
+
+        let stream: ValueStream =
+            Box::new(poll_fn(move || -> Poll<Option<Value>, failure::Error> {
+                loop {
+                    if let Some(offset) = buf.iter().position(|b| *b == b'\n') {
+                        let line_bytes = buf.split_to(offset + 1);
```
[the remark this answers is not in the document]
lutter: I think this will lead to another complication: when we are processing a bunch of bytes we read, we need to make sure that we do not process the last bit of that chunk unless it happens to align with a line boundary; i.e., if we read a chunk `"stuff"\n"more `, we must not call the JSON callback with `"more `. With the `BufRead.lines` method, we have to check every time we get a line if whether it is the last line in `buf`, and whether that ended in a newline or not. Or we'd have to keep an ending incomplete line somewhere else before creating the `BufReader`. Both options seem to have at least as much overhead as what's there now.

graph/src/components/link_resolver.rs
```diff
@@ -53,19 +77,158 @@ impl LinkResolver for ipfs_api::IpfsClient {
...
+                    true => Err(format_err!(
+                        "IPFS file {} is too large. It can be at most {} bytes but is {} bytes",
+                        path,
+                        max_file_bytes,
+                        stat.cumulative_size
+                    )),
+                })
+                .and_then(|()| future::ok(stream)),
+        )
+    }
```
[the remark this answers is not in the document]
lutter: Yeah, I was initially a bit scared of getting the types right. I now have something that seems to work for both cases.

runtime/wasm/src/host_exports.rs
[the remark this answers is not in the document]
leodasvacas: You're right of course, I misunderstood. I think what mislead me is that because of the name I thought `results` held was a vector of `Result`, so maybe renaming that variable would be good.

leodasvacas:
Re-thinking the approach a bit, if we're holding a large buffer in memory, potentially the entire file, then it seems arbitrary to require a JSON lines format. If the JSON fails to parse at a newline, we should just read the next lines until we get a valid object.

runtime/wasm/src/host_exports.rs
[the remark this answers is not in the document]
lutter: I renamed it to `operations`

runtime/wasm/src/module/mod.rs
```diff
@@ -489,24 +529,55 @@ where
...
+        let result = match self.host_exports().ipfs_map(&self, link, &*callback, flags) {
+            Ok(ops) => {
+                self.ctx.entity_operations.extend(ops);
+                Ok(None)
+            }
+            Err(e) => {
+                info!(self.logger, "Failed ipfs.map";
+                                    "link" => self.asc_get::<String, _>(link_ptr),
+                                    "callback" => callback,
+                                    "error" => e.to_string());
```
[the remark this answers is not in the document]
lutter: I've changed things so that the error from `HostExports.ipfs_map` contains the link and callback and removed the logging.

lutter:
I've addressed most of the review comments and updated the branch.

lutter:
> Re-thinking the approach a bit, if we're holding a large buffer in memory, potentially the entire file, then it seems arbitrary to require a JSON lines format. If the JSON fails to parse at a newline, we should just read the next lines until we get a valid object.

I don't think that how we process the file internally should have any bearing on the file format that we ask users to give us. And the case where we hold a lot of the file in memory at once would be a pretty pathological case. The input format should really be dictated by how onerous the line restriction if for users.

(Also, a file with one large JSON object is a case where I feel the approach of "keep trying to parse until it works" would be very bad as we'd have to do a lot of parses with increasingly bigger pieces of text until one works out or we reach EOF; bracket counting would be much more performant)

leodasvacas:
@lutter Not sure I follow, you think it's advantegous to keep the JSON lines requirement? Why?

About the issue of re-parsing, I agree it can be slow, I think the solution there is to have a minimum amount of content buffered before trying to parse, to make sure the network remains the dominating factor.

lutter:
@leodasvacas I think JSON lines and how much of the file we need to buffer are orthogonal (either way, we have to buffer at least one entire JSON object, if those turn out to be huge, we have to buffer that) JSON lines saves a bit of implementation effort. Either way, could we just get this in with the JSON lines restriction and decide on whether that needs to be removed or not in a separate ticket? What's there right now is useful at lest for user feedback, so I wouldn't want to put that off over this discussion.

lutter:
I just tried this all out with the setup in the `ipfs-map` branch of [`dice2win`](https://github.com/lutter/dice2win-subgraph/tree/ipfs-map)

The data file `import/bens.json` has 2863 lines. Each run through the file in the `handleCommit` handler takes about 2 minutes, which works out to be about 40ms per line. That seems way too slow to really importing things in bulk. The slowdown should be in cloning the `WasmiModule` for each line. I tried to use one module for all invocations, but ran into the issue that `WasmiModule` is not `Send`, and couldn't figure out a quick way to make it `Send`.

@leodasvacas you said that creating a new `WasmiModule` should be much faster. Is there anything I can do to speed it up? Is there an easy way to reuse the same module in `HostExport.ipfs_map` to at least see how much of a difference that would make?

leodasvacas:
@lutter Between which lines did you measure the 40ms? Did you compile with `--release`?

Jannis:
I'm playing with this and the ENS subgraph now. Let's wait with merging until I can report on how well that works.\

runtime/wasm/src/host_exports.rs
```diff
@@ -335,6 +343,53 @@ where
...
+            "ipfs_map: callback '{}' failed when processing file '{}'",
+            &*callback, &link
+        );
+
+        let operations = self.block_on(
+            self.link_resolver
+                .json_stream(&Link { link })
+                .and_then(|stream| {
+                    stream
+                        .and_then(move |v| {
```
Jannis: I'd use `fold` here instead of `and_then`. That way you don't have to do the `into_iter().flatten().collect()` dance at the end.

Essentially, something like this:
```
stream
    .fold(vec![], |mut operations, v| {
        ...
        let new_operations = module.handle_json_callback(&*callback, &v)?;
        v.extend(&new_operations);
        v
    })
```

Jannis: Maybe `iter().flatten().collect()` is more efficient even? I'm not sure. Just whenever I see a stream or array being turned into a single result value (e.g. a vector), I think of it as a reducing/folding operation.

leodasvacas: I think [concat](https://docs.rs/futures/0.1.17/futures/stream/trait.Stream.html#method.concat2) may work here.

graph/src/components/link_resolver.rs
```diff
@@ -1,71 +1,248 @@
...
+
+// The timeout for IPFS requests in seconds
+fn ipfs_timeout() -> Duration {
+    let timeout = env::var("GRAPH_IPFS_TIMEOUT").ok().map(|s| {
+        u64::from_str(&s).unwrap_or_else(|_| panic!("failed to parse env var GRAPH_IPFS_TIMEOUT"))
+    });
+    Duration::from_secs(timeout.unwrap_or(30))
+}
+
+fn read_env_var(name: &str) -> Option<u64> {
```
Jannis: `read_u64_from_env`? `read_env_u64`? Something that is more specific would be nice.

graph/src/components/link_resolver.rs
```diff
@@ -1,71 +1,248 @@
...
+    env::var(name)
+        .ok()
+        .map(|s| u64::from_str(&s).unwrap_or_else(|_| panic!("failed to parse env var {}", name)))
```
Jannis: It may be worth mentioning `(expected: unsigned 64-bit integer)` or something.

graph/src/components/link_resolver.rs
```diff
@@ -1,71 +1,248 @@
...
-                        true => Err(format_err!("IPFS file {} is too large", path)),
-                    })
-                    .and_then(|()| cat),
-            ),
-            None => Box::new(cat),
-        }
+        restrict_file_size(&self, path, max_file_bytes, Box::new(cat))
+    }
+
+    fn json_stream(
```
Jannis: `json_lines_stream`

runtime/wasm/src/host_exports.rs
Jannis: I have another request here: Could we log some progress here every now and then? Either every 1000th line or so (with an RwLock counter or similar). Maybe the stream could even emit stats along with the value, like bytes read and lines read that we can use.

runtime/wasm/src/module/mod.rs
```diff
@@ -258,14 +263,37 @@ where
             )
         })
     }
+
+    pub(crate) fn handle_json_callback(
```
Jannis: `handle_json_lines_callback`

runtime/wasm/src/module/mod.rs
```diff
@@ -489,24 +527,49 @@ where
...
+        let flags = self.asc_get(flags);
+        let start_time = Instant::now();
+        let result = match self.host_exports().ipfs_map(&self, link, &*callback, flags) {
+            Ok(ops) => {
+                self.ctx.entity_operations.extend(ops);
+                Ok(None)
+            }
+            Err(e) => Err(e.into()),
+        };
+        // Advance this module's start time by the time it took to run the entire
```
Jannis: Nit: I'd put an empty line above this

runtime/wasm/wasm_test/ipfs_map.ts
```diff
@@ -0,0 +1,141 @@
...
+    OBJECT = 5,
+}
+
+type JSONValuePayload = u64
+class JSONValue {
+    kind: JSONValueKind
+    data: JSONValuePayload
+
+    toString(): string {
+        assert(this.kind == JSONValueKind.STRING, 'JSON value is not a string.')
```
Jannis: The indentation in this file is inconsistent (2 spaces vs. 4 spaces). Doesn't matter much, since it's a test file, so I'll leave it up to you to decide if this is worth changing. 😉

Jannis: (2 spaces being more idiomatic / aligned with our graph-ts code for instance)

graph/src/components/link_resolver.rs
```diff
@@ -1,71 +1,248 @@
...
 /// Resolves links to subgraph manifests and resources referenced by them.
 pub trait LinkResolver: Send + Sync + 'static {
     /// Fetches the link contents as bytes.
     fn cat(&self, link: &Link) -> Box<Future<Item = Vec<u8>, Error = failure::Error> + Send>;
+
+    /// Read the contents of `link` and deserialize them into a stream of JSON
+    /// values. The values must each be on a single line; newlines are significant
+    /// as they are used to split the file contents and each line is deserialized
+    /// separately.
+    fn json_stream(
```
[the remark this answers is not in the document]
leodasvacas: I don't think we should have `json_lines` in any of the naming because as we discussed above it would be simple to lift this requirement.

Jannis:
Nick has sent us files for ENS that we can try. Each of them has about 8300000 lines, each of the format `{"t": "victoryceratech"}`.

If each of these lines would take `20ms` to process, it would take `8300000 * 0.02s = 166000s = 2766.7m = 44h` to process _one_ of these files.  At `1ms` per line (which I don't think we can pull off), it would still be around `2.2h`.

graph/src/components/link_resolver.rs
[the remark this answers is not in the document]
Jannis: Hmmm, I can see that. Ok, ignore related comments.

lutter:
> @lutter Between which lines did you measure the 40ms? Did you compile with `--release`?

I had indeed not run this with `--release`. Rerunning it with that drops the time to process the test file to 1.3s. That speedup is amazing.

runtime/wasm/src/host_exports.rs
lutter: If I use `fold` here, i.e., turn the code to
```
        let operations = self.block_on(
            self.link_resolver
                .json_stream(&Link { link })
                .and_then(|stream| {
                    stream.fold(vec![], move |mut ops, v| {
                        let module = WasmiModule::from_valid_module_with_ctx(
                            valid_module.clone(),
                            ctx.clone(),
                        )?;
                        ops.extend(module.handle_json_callback(&*callback, &v)?);
                        Ok(ops)
                    })
                })
                .map_err(move |e| HostExportError(format!("{}: {}", errmsg, e.to_string()))),
        )?;
```

I get an error
```
error[E0283]: type annotations required: cannot resolve `graph::prelude::Error: std::convert::From<_>`
   --> runtime/wasm/src/host_exports.rs:377:28
    |
377 |                     stream.fold(vec![], move |mut ops, v| {
    |                            ^^^^

error: aborting due to previous error
```

and I am not quite sure what rustc is trying to tell me here.

graph/src/components/link_resolver.rs
lutter: Renamed it to `read_u64_from_env`

graph/src/components/link_resolver.rs
lutter: Good point, that was copy-pasta from the existing code. Changed the error message to `expected env var {} to contain a number (unsigned 64-bit integer), but got '{}'`

runtime/wasm/src/module/mod.rs
lutter: Done

runtime/wasm/wasm_test/ipfs_map.ts
lutter: Done

runtime/wasm/src/host_exports.rs
lutter: I've fixed things up so that we log progress every 15s on a logger dedicated to this ipfs_map invocation; the log message is
```
                               debug!(
                                    logger,
                                    "processed {} lines in {}ms so far",
                                    sv.line,
                                    start.elapsed().as_millis()
                                );
```

lutter: Actually, I'll log it in seconds; millis is silly here if we only log every 15s

lutter:
@leodasvacas @Jannis I addressed all the review comments that I could. Can you have a look and also go back to the open review comments and update them?

runtime/wasm/src/host_exports.rs
```diff
@@ -335,6 +343,67 @@ where
...
+                            let module = WasmiModule::from_valid_module_with_ctx(
+                                valid_module.clone(),
+                                ctx.clone(),
+                            )?;
+                            let result = module.handle_json_callback(&*callback, &sv.value);
+                            // Log progress every 15s
+                            if last_log.elapsed() > Duration::from_secs(15) {
+                                debug!(
+                                    logger,
+                                    "processed {} lines in {}s so far",
```
Jannis: Our log messages start with a capital first letter unless it's a symbol like `ipfs_map`.

runtime/wasm/src/host_exports.rs
Jannis: Awesome, I like it!

Jannis: I wouldn't worry about it in this case — I bet the flatten approach that you've picked isn't going to be a problem. @leodasvacas What do you think?

runtime/wasm/src/host_exports.rs
lutter: Fixed

runtime/wasm/src/host_exports.rs
leodasvacas: The error is misterious to me as well, and I also don't think the current approach is an issue.

lutter:
I've updated the PR (and the corresponding `graph-ts` PR) to take a `userData` parameter and pass it through to the callback. I've also updated the [dice2win ipfs-map](https://github.com/lutter/dice2win-subgraph/tree/ipfs-map) branch to use the new call signatures, and tried this out end-to-end with that subgraph.

lutter:
Rebased to latest master

Jannis:
@lutter Great, I've tested this end to end as well. Works great. :)

Jannis:
@lutter I've taken the liberty to add one commit that adds `ms` after the times logged, so the unit is clear.

Jannis:
I'm going to go ahead and merge this, so we can pull it into today's releases.

PR #836, opened 2019-03-25
Resolve entityCount only if queried for

Jannis:
<s>This comes with a small limitation: We only check the entity type name
to be `SubgraphDeployment` now, rather than also checking the ID of the
subgraph this entity originates from. We should improve this.</s>

This now limits the lazy `entityCount` "resolver" to the subgraph of subgraphs.

### DIFF
--- a/graphql/src/execution/resolver.rs
+++ b/graphql/src/execution/resolver.rs
@@ -79,6 +79,7 @@ pub trait Resolver: Clone + Send + Sync {
     /// Resolves a scalar value for a given scalar type.
     fn resolve_scalar_value(
         &self,
+        _parent_object_type: &s::ObjectType,
         _parent: &BTreeMap<String, q::Value>,
         _field: &q::Name,
         scalar_type: &s::ScalarType,
--- a/graphql/src/execution/resolver.rs
+++ b/graphql/src/execution/resolver.rs
@@ -70,10 +70,27 @@ pub trait Resolver: Clone + Send + Sync {
     ) -> Result<q::Value, QueryExecutionError>;
 
     /// Resolves an enum value for a given enum type.
-    fn resolve_enum_value(&self, enum_type: &s::EnumType, value: Option<&q::Value>) -> q::Value {
-        value
-            .and_then(|value| value.coerce(enum_type))
-            .unwrap_or(q::Value::Null)
+    fn resolve_enum_value(
+        &self,
+        field: &q::Field,
+        enum_type: &s::EnumType,
+        value: Option<&q::Value>,
+    ) -> Result<q::Value, QueryExecutionError> {
+        value.map_or(Ok(q::Value::Null), |value| {
+            value.coerce(enum_type).ok_or_else(|| {
+                QueryExecutionError::EnumCoercionError(
+                    field.position.clone(),
+                    field.name.to_owned(),
+                    value.clone(),
+                    enum_type.name.to_owned(),
+                    enum_type
+                        .values
+                        .iter()
+                        .map(|value| value.name.to_owned())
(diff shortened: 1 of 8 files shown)


Jannis:
Updated the tests to include `@derivedFrom` on a non-list field.

graphql/src/execution/resolver.rs
```diff
@@ -79,12 +79,15 @@ pub trait Resolver: Clone + Send + Sync {
...
+        _field: &q::Name,
         scalar_type: &s::ScalarType,
         value: Option<&q::Value>,
-    ) -> q::Value {
-        value
+    ) -> Result<q::Value, QueryExecutionError> {
+        Ok(value
             .and_then(|value| value.coerce(scalar_type))
-            .unwrap_or(q::Value::Null)
+            .unwrap_or(q::Value::Null))
```
leodasvacas: Since the signature now returns a Result, can't we return the error instead of having `unwrap_or`?

graphql/src/store/resolver.rs
```diff
@@ -192,39 +191,66 @@ where
...
+                Ok(Value::from(
+                    self.store
+                        .count_entities(hash)
+                        .map_err(QueryExecutionError::StoreError)?,
+                )
+                .into())
+            }
+            _ => Ok(value
+                .and_then(|value| value.coerce(scalar_type))
+                .unwrap_or(q::Value::Null)),
```
leodasvacas: Same here.

leoyvens (changes requested):
Nice! Just a comment on error handling.

Jannis:
@leodasvacas Updated resolvers so that coercion errors are being returned to the client; in the case of enum with a list of possible values.

graphql/src/introspection/resolver.rs
```diff
@@ -385,7 +385,9 @@ fn input_value(
             input_value
                 .default_value
                 .as_ref()
-                .map_or(q::Value::Null, |value| value.clone()),
+                .map_or(q::Value::Null, |value| {
+                    q::Value::String(format!("{}", value))
+                }),
```
leodasvacas: Could we get test for introspecting the default value of an Int, for example?

graphql/tests/introspection.rs
```diff
@@ -449,7 +449,7 @@ fn expected_mock_schema_introspection() -> q::Value {
             q::Value::List(vec![object_value(vec![
                 ("name", q::Value::String("language".to_string())),
                 ("description", q::Value::Null),
-                ("defaultValue", q::Value::String("English".to_string())),
+                ("defaultValue", q::Value::String("\"English\"".to_string())),
```
leodasvacas: How comes this got the extra quotes? Is this desirable?

leoyvens (changes requested):
Nice error message and correctness fixes, just a couple comments on the introspection fix.

graphql/tests/introspection.rs
Jannis: Yep! E.g. in GraphiQL, you'd want
```graphql
someField(
  someInput = "default string"
)
```
to be shown instead of
```graphql
someField (
  someInput = default string
)
```

graphql/src/introspection/resolver.rs
Jannis: Done!

Jannis:
@leodasvacas Ready for another round 😄

graphql/tests/introspection.rs
leodasvacas: ah interesting

leoyvens (approved):

PR #837, opened 2019-03-25
Implement @derivedFrom for non-list fields

Jannis:
This allows to define types like
```graphql
type Domain @entity {
  id: ID!
  ...
}

type DomainAuction @entity {
  id: ID!
  domain: Domain @derivedFrom(field: "id")
  ...
}
```
and resolve them in queries by virtue of the two sides sharing the same ID.

### DIFF
--- a/core/tests/interfaces.rs
+++ b/core/tests/interfaces.rs
@@ -147,7 +147,7 @@ fn one_interface_multiple_entities() {
         "Furniture",
     );
 
-    let query = "query { leggeds(first: 100) { legs } }";
+    let query = "query { leggeds(first: 100, orderBy: legs) { legs } }";
 
     let res = insert_and_query(subgraph_id, schema, vec![animal, furniture], query).unwrap();
     assert!(res.errors.is_none());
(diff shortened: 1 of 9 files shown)


Jannis:
Updated the tests to include `@derivedFrom` on a non-list field.

leodasvacas:
As we were discussing on Discord, if more than one entity has the matching id, this may return any of the matching entities as the value. We can consider that to be a logic error by the user, but we don't catch it in any way. So I'm not confortable merging until we either come up with a better solution or at least have a wider discussion to check if we are ok with the implications.

Jannis:
Ideas:

* We could query matches with `first: 2` instead of `first: 1` and return a GraphQL error if there is more than one. This is the easiest solution, but makes the user deal with the problem rather than the subgraph author.

* When writing an entity, we could check if there are any non-list `@derivedFrom` fields that involve the entity type. If there are any, check if there already is an entity of the same entity type that matches that `@derivedFrom`. If that's the case fail, similar to how we fail with ID clashes for interfaces.

leodasvacas:
@Jannis I think the first idea is good enough to allow landing this.

Having an integrity check that prevents the situation from arising is good in theory, but it's complicated. Your second idea solves half of the problem, but there is also the case that two entities already exist referencing the same id, and then a deriving entity with that id is added.

Jannis:
@leodasvacas Ok, first approach implemented!

leoyvens (changes requested):
Looking good, can we get a test for the `AmbiguousDerivedResult` error?

Jannis:
@leodasvacas I guess... 😬

leoyvens (approved):

PR #841, opened 2019-03-27
Big decimal

### LINKED ISSUE #720: Big Decimal Support
Feature

Big Ints can't return fractions, so in some cases we lose precision that we need

Reproduction of the problem can be seen on uniswap subgraph, where we try to divide ETH price by MKR price and it returns 0 instead of `~0.270`

We get the fraction returned

I closed this issue https://github.com/graphprotocol/graph-ts/issues/34 in `graph-ts` that is related

leoyvens:
Closes #720.


graph/src/data/store/mod.rs
```diff
@@ -164,21 +165,14 @@ impl Value {
...
-            (query::Value::Enum(e), NamedType(n)) => {
-                // Check if `ty` is a custom scalar type, otherwise assume it's
-                // just a string.
-                match n.as_str() {
-                    BYTES_SCALAR => Value::Bytes(scalar::Bytes::from_str(e)?),
-                    BIG_INT_SCALAR => Value::BigInt(scalar::BigInt::from_str(e)?),
-                    _ => Value::String(e.clone()),
-                }
-            }
+            (query::Value::Enum(e), NamedType(_)) => Value::String(e.clone()),
```
Jannis: Interesting, the above didn't really make sense, huh (an enum value that represents a `Bytes` or `BigInt` value isn't even possible, since enum values don't have quotes around them like strings/bytes/bigints)?

graph/src/data/store/mod.rs
```diff
@@ -188,8 +182,6 @@ impl Value {
                     .ok_or_else(|| QueryExecutionError::NamedTypeError("Int".to_string()))?
                     as i32,
             ),
-            (query::Value::Variable(var), _) => Value::String(var.to_owned()),
```
Jannis: Is this not needed?

graph/src/data/store/scalar.rs
```diff
@@ -9,6 +9,10 @@ use std::str::FromStr;
 
 pub use num_bigint::Sign as BigIntSign;
 
+// Caveat: The expoent is currently an i64 and may overflow.
```
Jannis: Typo: expoent -> exponent

graphql/tests/query.rs
```diff
@@ -325,7 +325,7 @@ fn can_query_one_to_many_relationships_in_both_directions() {
         query {
             musicians(first: 100, orderBy: id) {
                 name
-                writtenSongs(first: 100) {
+                writtenSongs(first: 100, orderBy: id) {
```
Jannis: Ah, great. I think I added that in one of my branches as well.

runtime/wasm/src/host_exports.rs
```diff
@@ -432,6 +446,38 @@ where
...
+    pub(crate) fn big_decimal_minus(&self, x: BigDecimal, y: BigDecimal) -> BigDecimal {
+        x - y
+    }
+
+    pub(crate) fn big_decimal_times(&self, x: BigDecimal, y: BigDecimal) -> BigDecimal {
+        x * y
+    }
+
+    pub(crate) fn big_decimal_divided_by(&self, x: BigDecimal, y: BigDecimal) -> BigDecimal {
+        x / y
```
Jannis: What happens if `y` is essentially `0`?

runtime/wasm/src/host_exports.rs
```diff
@@ -432,6 +446,38 @@ where
...
+    pub(crate) fn big_decimal_to_string(&self, x: BigDecimal) -> String {
+        x.to_string()
+    }
+
+    pub(crate) fn big_decimal_from_string(
+        &self,
+        s: String,
+    ) -> Result<BigDecimal, HostExportError<impl ExportError>> {
+        BigDecimal::from_str(&s)
+            .map_err(|e| HostExportError(format!("Failed to parse BigDecimal: {}", e)))
```
Jannis: Nit: The capitalization between this error and the ```attempted to divide BigInt `{}` by zero``` is inconsistent.

runtime/wasm/src/to_from/external.rs
```diff
@@ -58,6 +58,24 @@ impl FromAscObj<AscBigInt> for BigInt {
     }
 }
 
+impl ToAscObj<AscBigDecimal> for BigDecimal {
+    fn to_asc_obj<H: AscHeap>(&self, heap: &mut H) -> AscBigDecimal {
+        // "exponent" here is the opposite of what you'd expect.
```
Jannis: Could we make this a little more specific? I found `Note that a positive exponent indicates a negative power of 10` in the `BigDecimal` docs, which made it more clear: https://docs.diesel.rs/bigdecimal/struct.BigDecimal.html#method.as_bigint_and_exponent

Jannis: On another note: Is this a normal thing to do, do we know? It's so weird. 😉

Jannis (changes requested):
This is looking great overall!

graph/src/data/store/mod.rs
leodasvacas: Yhea I understand that an enum is an unquoted string, and those can never be numbers, so it doesn't seem to make sense.

graph/src/data/store/mod.rs
leodasvacas: No, we should have resolved variables by now.

runtime/wasm/src/host_exports.rs
leodasvacas: Fixed.

runtime/wasm/src/to_from/external.rs
leodasvacas: Don't know why the Rust library does that, it's also weird to me. I checked the Python library and it uses exponents.

graph/src/data/store/scalar.rs
leodasvacas: Ha, in portuguese we don't have that `n`.

leodasvacas:
@Jannis review addressed.

graph/src/data/store/mod.rs
Jannis: Ah, yes 👍

graph/src/data/store/scalar.rs
Jannis: 😁

Jannis (approved):

PR #842, opened 2019-03-28
Use entities.id rather than entities.data->'id'->>'data' in queries

lutter:
This PR makes sure we do not query `entities.id` via the `data` column, and instead always use `entities.id`. That allows us to get rid of ~ 260 attribute indexes in production, which the migration in this PR does.


Jannis:
@lutter I'm going to merge this on your behalf.

PR #847, opened 2019-03-31
store/postgres: fix bug in attempt_chain_head_update for short chains

lutter:
For short chains, we would pull the genesis block into the query that
checks if we have parents for at least ancestor_count many blocks before
the new head. But the genesis block has no parent, and so this would fail
and we would report that the genesis block has a missing parent and never
make progress

### DIFF
--- a/store/postgres/migrations/2019-03-31-010824_short_chain_head_update/down.sql
+++ b/store/postgres/migrations/2019-03-31-010824_short_chain_head_update/down.sql
@@ -1,0 +1,81 @@
+-- Restore attempt_chain_head_update to the state it had with migration
+-- 2019-03-04-235349_empty-ethereum-network-chain-head-update-fix
+-- (commit c9dbb2b9)
+CREATE OR REPLACE FUNCTION attempt_chain_head_update(net_name VARCHAR, ancestor_count BIGINT)
+    RETURNS VARCHAR[] AS
+$$
+DECLARE
+    current_head_number BIGINT;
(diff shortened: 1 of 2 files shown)


lutter:
I tested this with ganache locally, with the `example-subgraph`, and it is processing events from the exmaple contract, and the chain head pointer is advancing. The database then also contains 8 blocks. But I would still appreciate if somebody could double check that.

Jannis:
I've tested this with the example subgraph + Ganache as well and here only a single block ends up in the db, with the `head_block_hash` and `head_block_number` in `ethereum_networks` being empty. I'll take a closer look again tomorrow.

lutter:
I just tried this again (starting from an empty database), and it worked for me. The one thing I noticed is that diesel did not run the migration the first time I did this; it seems that that's because the change only contains a migration, and cargo does not pick up that it needs to rebuild `graph-store-postgres`. After touching a random `.rs` file in `store/postgres/src`, cargo rebuilt, the migration was run, and everything worked. Without the migration, I saw exactly what you described above.

Longer term, I would really like to pull this and most of the other stored procs out and put them in the Rust code; much easier to manage (like, this fix would have been a two line fix, without all the hassle we just went through)

Jannis:
Ha! I was just coming back to this and was about to say the same: I had to recompile `graph-store-postgres` and then it worked.

Jannis:
Let's merge this now so we can get it out to people.

PR #850, opened 2019-04-02
Make contract addresses in data sources optional

### LINKED ISSUE #616: Support indexing events without specifying a contract address
At present you have to specify the address from which events will come. For some event types (for instance, ERC20 transfer), it's beneficial to be able to listen to all matching events, regardless of the address that emitted them.

Jannis:
Resolves #616 (together with https://github.com/graphprotocol/graph-cli/pull/240).

To not leave out data sources without contract addresses in our `eth_getLogs` requests, we need to exclude contract addresses from the logs filter if there is even a single data source without an address. That makes the scan broader and thus less efficient but doesn't break our processing. We still have `matches_log_address()` checks in place to direct events to the right data sources and filter out false positives from the bloom filter.

### DIFF
--- a/datasource/ethereum/src/ethereum_adapter.rs
+++ b/datasource/ethereum/src/ethereum_adapter.rs
@@ -118,9 +118,15 @@ where
 
         // Collect all contract addresses; if we have a data source without a contract
         // address, we can't add addresses to the filter because it would only match
-        // the contracts for which we _have_ addresses; therefor if we have a data source
+        // the contracts for which we _have_ addresses; therefore if we have a data source
         // without a contract address, we perform a broader logs scan and filter out
-        // irrelevant events ourselves
+        // irrelevant events ourselves.
+        //
+        // Our own filtering is performed later when the events are passed to
+        // subgraphs and runtime hosts for processing:
+        // - At the top level in `SubgraphInstanceManager::start_subgraph`
+        // - At the subgraph level in `SubgraphInstance::matches_log`
+        // - At the data source level in `RuntimeHost::matches_log`
         let addresses = if log_filter
             .contract_address_and_event_sig_pairs
             .iter()
--- a/datasource/ethereum/src/ethereum_adapter.rs
+++ b/datasource/ethereum/src/ethereum_adapter.rs
@@ -116,14 +116,32 @@ where
             .into_iter()
             .collect::<Vec<H256>>();
 
-        // Collect all contract addresses
-        let addresses = log_filter
+        // Collect all contract addresses; if we have a data source without a contract
+        // address, we can't add addresses to the filter because it would only match
+        // the contracts for which we _have_ addresses; therefor if we have a data source
+        // without a contract address, we perform a broader logs scan and filter out
+        // irrelevant events ourselves
+        let addresses = if log_filter
             .contract_address_and_event_sig_pairs
             .iter()
-            .map(|(addr, _sig)| *addr)
-            .collect::<HashSet<H160>>()
-            .into_iter()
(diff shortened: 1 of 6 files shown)


Jannis:
Tested locally using our example subgraph. If you remove the contract address in the main data sources, it still indexes all the same events + 2 more from two other Gravity smart contracts.

For making extra sure that it worked, I added a `contract: Bytes!` field to the `Gravatar` type in the schema, then used `gravatar.contract = event.address` in `handleNewGravatar` to store the contract the gravatar came from. The following screenshot shows that there are two gravatars from different contracts, followed by all the gravatars from the contract that we explicitly pointed to before:

image

datasource/ethereum/src/ethereum_adapter.rs
```diff
@@ -116,14 +116,32 @@ where
             .into_iter()
             .collect::<Vec<H256>>();
 
-        // Collect all contract addresses
-        let addresses = log_filter
+        // Collect all contract addresses; if we have a data source without a contract
+        // address, we can't add addresses to the filter because it would only match
+        // the contracts for which we _have_ addresses; therefor if we have a data source
```
leodasvacas: s/therefor/therefore

graph/src/components/ethereum/adapter.rs
```diff
@@ -112,17 +112,21 @@ impl EthereumLogFilter {
...
-            Some(sig) => self
-                .contract_address_and_event_sig_pairs
-                .contains(&(log.address, *sig)),
+            Some(sig) => {
+                self.contract_address_and_event_sig_pairs
+                    .contains(&(Some(log.address), *sig))
+                    || self
+                        .contract_address_and_event_sig_pairs
+                        .contains(&(None, *sig))
+            }
```
leodasvacas: Instead of `contains` you could use `.iter().any(|(_, s)| s == sig)`.

runtime/wasm/src/host.rs
```diff
@@ -245,7 +245,8 @@ impl RuntimeHost {
     }
 
     fn matches_log_address(&self, log: &Log) -> bool {
-        self.data_source_contract.address == log.address
+        self.data_source_contract.address.is_none()
+            || self.data_source_contract.address.unwrap() == log.address
```
leodasvacas: Maybe nicer as `.address.map(|a| a == log.address).unwrap_or(true)`.

datasource/ethereum/src/ethereum_adapter.rs
```diff
@@ -116,14 +116,32 @@ where
...
+        // without a contract address, we perform a broader logs scan and filter out
+        // irrelevant events ourselves
```
leodasvacas: It would be good to refer to where this filtering is done.

leoyvens (changes requested):
That really was simple, just some minor comments.

Jannis:
@leodasvacas I believe I've addressed all your comments.

runtime/wasm/src/host.rs
```diff
@@ -245,8 +245,12 @@ impl RuntimeHost {
...
 
     fn matches_log_address(&self, log: &Log) -> bool {
-        self.data_source_contract.address.is_none()
-            || self.data_source_contract.address.unwrap() == log.address
+        // The runtime host matches the contract address of the `Log`
+        // if the data source contains the same contract address or
+        // if the data source doesn't have a contract address at all
+        self.data_source_contract
+            .address
+            .map_or(true, |addr| addr == log.address)
```
leodasvacas: `map_or`, nice!

leoyvens (approved):

PR #852, opened 2019-04-02
Log reason for terminating runtime thread

leoyvens:


lutter (approved):
Looks good. Can you amend the commit message to explain why that change was made? There is no context otherwise why that's necessary/desirable.

Jannis (approved):

leodasvacas:
@lutter Usually the thread is canceled from stopping the subgraph so this is not interesting, but if the failure is in starting up the module then wouldn't otherwise see what caused it. Ideally it would be an `error!` in that case, but it's fine as is. I'l just merge so it doesn't have to go through travis and reviews again.

lutter:
@leodasvacas thanks for the explanation. What I meant was that that explanation should have gone into the commit message, so that we have the context of why the change was made.

leodasvacas:
@lutter I wanted to be quick in merging this since it's useful for what I'm doing right now, and GitHub links commits back to PRs, so for as long as GitHub exists I think it's fine to keep some of the context in the PR. Not that this is an excuse for not writing better messages and descriptions, I'll be more mindful in the future.

PR #853, opened 2019-04-02
Don't print passwords from URLs in the logs

### LINKED ISSUE #849: Postgres password shouldn't be printed to the logs
Hi! I love this project.

Bug

I set up a Docker node with `graphprotocol/graph-node:latest`, using the environment variables to provide my postgres username and password. That password that I provided is being logged twice — first by `bash` when the Docker image boots up:

```
+ graph-node --postgres-url postgresql://postgres:[redacted password here]@127.0.0.1:5432/subgraph --ethereum-rpc mainnet:http://ethereum-rpc.default.svc.cluster.local --ipfs localhost:8080
```

And again by subgraph itself when it starts the connection:

```
Apr 01 21:34:54.671 INFO Connecting to Postgres, url: postgresql://postgres:[redacted password here]@127.0.0.1:5432/subgraph
```

The application log happens on every run of the software. The Docker log happens on every run of the software in the Docker container that doesn't include a custom command.

No password logging to stdout or stderr please! Security best practices and all that.

Jannis:
Resolves #849. WIP.


lutter:
Is there any way to do this so that it's secure by default, i.e. in a way that the `Display` trait and `to_string` will return the redacted URL, and you need to call something like `get_unredacted_url()` to get the URL with the password? Otherwise, I feel this is going to turn into a game of whack-a-mole to make sure the password doesn't get logged again somewhere.

Jannis:
@lutter I don't think this is something we can get by default. But this makes me rethink my approach: 

How about we have a `ToSafeValue` trait with a `.to_safe_value()` method (open to better names) that we implement for `SocketAddr`, `String`, `Url` etc.

Perhaps we can even extend the `slog` macros to require `Into<SafeValue>`s to be passed in, with us providing a derive macro for this + derived implementations for various types.

Let me think about this more. 😄

PR #863, opened 2019-04-08
Don't ask if old blocks are on the main chain

leoyvens:
This saves us a few eth_getBlockByNumber calls when scanning old blocks. A small gain from a small change.


timmclean:
This could cause some confusion when support for full block explorer use cases is added, since uncled blocks would be indexed and might (?) end up being provided to this function. How about updating the trait documentation to reflect that `is_on_main_chain` has a new assumption on its inputs? Or, could move this optimization out to the function's caller if it's more clear there that the assumption will hold

leodasvacas:
@timmclean Tim! Thanks for commenting! That's a good point. I'll just close for now since this is a very small optimization, I only realized how small it was after I did it, it's not worth the special case.

PR #866, opened 2019-04-10
Change GraphQL query execution to decide whether introspection is needed upfront

lutter:
With this PR, the decision which fields in a selection set require introspection and which ones require 'normal' data lookup is made based just on the root fields. That simplifies `ExecutionContext` as we can now just pass different contexts into query execution rather than have the context decide when to use an introspection schema and when to use the subgraph schema.

### DIFF
--- a/graphql/src/execution/execution.rs
+++ b/graphql/src/execution/execution.rs
@@ -62,6 +62,145 @@ where
         ctx.fields.push(field);
         ctx
     }
+
+    // Helper functions that decide whether to call underlying functions on the
+    // regular or on the introspection schema. Only these functions should be
+    // making decisions on whether we are introspecting or not
+    fn get_schema(&self) -> &s::Document {
+        if self.introspecting {
+            self.introspection_schema
+        } else {
+            &self.schema.document
+        }
+    }
+
+    fn resolve_object(
+        &self,
+        parent: &Option<q::Value>,
+        field: &q::Field,
+        field_definition: &s::Field,
+        object_type: ObjectOrInterface<'_>,
+        arguments: &HashMap<&q::Name, q::Value>,
+    ) -> Result<q::Value, QueryExecutionError> {
+        if self.introspecting {
+            self.introspection_resolver.resolve_object(
+                parent,
+                field,
+                field_definition,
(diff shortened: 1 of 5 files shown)


graphql/src/execution/execution.rs
```diff
@@ -15,87 +15,149 @@ use crate::values::coercion;
...
+        span: selection_set.span.clone(),
+        items: Vec::new(),
+    };
+    let mut intro_set = q::SelectionSet {
+        span: selection_set.span.clone(),
+        items: Vec::new(),
+    };
+
+    let ictx = ctx.as_introspection_context();
+    let introspection_query_type = sast::get_root_query_type(&ictx.schema.document).unwrap();
```
leodasvacas: By the spec, introspection queries should be executed against the same root type as the regular queries.

Jannis: I skimmed through the spec but I didn't find it to be very specific about this. I don't think it makes a difference here.

leodasvacas: In the spec introspection doesn't have a full schema, but is set of fields and types that are implicitly injected into any root query type. This does seem equivalent as long as the query types are the same wrt to directives and arguments (we currently have none of those anyways), and this change will help with sql based execution, so I'm not saying we shouldn't do this, but just to be mindful of this detail.

lutter: It's also worth mentioning that this ultimately does things the same way we used to do them, just that the switch to the introspection schema is more explicit and at the root level.

graphql/src/execution/execution.rs
```diff
@@ -15,87 +15,149 @@ use crate::values::coercion;
...
+    for (_, fields) in collect_fields(ctx.clone(), query_type, selection_set, None) {
+        let name = fields[0].name.clone();
+        let selections = fields.into_iter().map(|f| q::Selection::Field(f.clone()));
```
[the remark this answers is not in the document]
lutter: Yeah, I didn't like it either when I wrote it, but can't find a way to get around it. On the plus side, this only happens at the root level of the query.

PR #869, opened 2019-04-11
Limit query complexity

### LINKED ISSUE #614: Limit query complexity
At present it's very easy to OOM the process by querying, say, every entity of a given type, if there are a lot of them. In order to make a Graph endpoint publicly accessible, there needs to be some kind of limit on the number of entities a query will execute - either by counting operations and terminating a query early if it's reached, by enforcing a result set size limit on queries, or both.

### LINKED ISSUE #770: Limit total number of entities to return

leoyvens:
Resolves #770.

Complexity is checked in `execute_query` and `execute_subscription`. The logic is in the `root_query_complexity` function. Because not all collection queries have `first`, there is a need to check the type of the field, and introspection adds special cases here.

@nenadjaja This was not easy at all!


core/src/graphql/runner.rs
```diff
@@ -21,6 +21,10 @@ lazy_static! {
             u64::from_str(&s)
                 .unwrap_or_else(|_| panic!("failed to parse env var GRAPH_GRAPHQL_QUERY_TIMEOUT"))
         ));
+    static ref GRAPHQL_MAX_COMPLEXITY: Option<u64> = env::var("GRAPH_GRAPHQL_MAX_COMPLEXITY")
```
Jannis: While we're at it, should we also make `GRAPH_GRAPHQL_MAX_DEPTH` a variable?

docs/environment-variables.md
```diff
@@ -36,14 +36,17 @@ in parallel for block retrieval (defaults to 64)
   is done processing. This setting therefore limits how much memory a call
   to `ipfs.map` may use. (in bytes, defaults to 256MB)
 
+## GraphQL
+* `GRAPH_GRAPHQL_QUERY_TIMEOUT`: maximum execution time for a graphql query, in seconds. Default is unlimited.
+* `SUBSCRIPTION_THROTTLE_INTERVAL`: while a subgraph is syncing,
+  subscriptions to that subgraph get updated at most this often, in
+  ms. Default is 1000ms.
+* `GRAPH_GRAPHQL_MAX_COMPLEXITY`: maximum complexity for a graphql query. See [here](https://developer.github.com/v4/guides/resource-limitations) for what that means. Default is unlimited.
+
```
Jannis: If we make `GRAPH_GRAPHQL_MAX_DEPTH` a thing, this where we'd document it.

graph/src/data/query/error.rs
```diff
@@ -50,6 +50,8 @@ pub enum QueryExecutionError {
     Unimplemented(String),
     EnumCoercionError(Pos, String, q::Value, String, Vec<String>),
     ScalarCoercionError(Pos, String, q::Value, String),
+    TooComplex(u64, u64), // (complexity, max_complexity)
+    TooDeep,
```
Jannis: Could we add the actual and allowed depth as arguments here, like with `TooComplex(u64, u64)`?

graph/src/data/query/error.rs
```diff
@@ -182,6 +184,10 @@ impl fmt::Display for QueryExecutionError {
             ScalarCoercionError(_, field, value, scalar_type) => {
                 write!(f, "Failed to coerce value `{}` of field `{}` to scalar type `{}`", value, field, scalar_type)
             }
+            TooComplex(complexity, max_complexity) => {
+                write!(f, "Query has complexity `{}` which is over the maximum `{}`", complexity, max_complexity)
```
Jannis: Users will no doubt see this and "complexity" won't mean much to them. It would improve the user experience if we described its meaning. GitHub returns the following error message:

> "By the time this query traverses to the organizations connection, it is requesting up to 1,000,000 possible nodes which exceeds the maximum limit of 500,000." 

I don't think the message is great, specifically the `organizations connection` part is confusing because there could be many of them.

How about this?

```
Query potentially returns `{}` entities or more and thereby exceeds
the limit of `{}` entities. Possible solutions are reducing the depth
of the query, querying fewer relationships or using `first` to
return smaller collections.
```

graph/src/data/query/error.rs
```diff
@@ -182,6 +184,10 @@ impl fmt::Display for QueryExecutionError {
...
+            }
+            TooDeep => write!(f, "Query is too deep")
```
Jannis: How about this?
```
Query has a depth of `{}` and thereby exceeds the depth limit of `{}`.
```

graphql/src/execution/execution.rs
```diff
@@ -67,6 +74,151 @@ where
...
+        root_selection_set: &q::SelectionSet,
+    ) -> Result<u64, QueryExecutionError> {
+        const MAX_DEPTH: u32 = 100;
+
+        match self.query_complexity(root_type, root_selection_set, MAX_DEPTH, 0) {
+            Ok(complexity) => Ok(complexity),
+            Err(ComplexityError::Invalid) => Ok(0),
+            Err(ComplexityError::TooDeep) => Err(QueryExecutionError::TooDeep),
+            Err(ComplexityError::Overflow) => {
+                Err(QueryExecutionError::TooComplex(u64::max_value(), 0))
```
Jannis: Is the `0` here the right thing to pass to `TooComplex`? Intuitively I'm thinking this should be the limit?

graphql/src/execution/execution.rs
```diff
@@ -67,6 +74,151 @@ where
...
+        }
+
+        // Helpers to look for types and fields on both the introspection and
+        // regular schemas.
+        fn get_named_type(
+            schema: &s::Document,
+            name: &Name,
+        ) -> Result<s::TypeDefinition, ComplexityError> {
+            if name.starts_with("__") {
+                let doc = introspection_schema(SubgraphDeploymentId::new("").unwrap()).document;
```
Jannis: I may be wrong but creating the introspection schema for every matching named type feels expensive.

graphql/src/execution/execution.rs
```diff
@@ -67,6 +74,151 @@ where
...
+                s::Type::ListType(inner) => get_type_definition_from_type(schema, inner),
+                s::Type::NonNullType(inner) => get_type_definition_from_type(schema, inner),
+            }
+        }
+        fn get_field<'a>(
+            object_type: impl Into<ObjectOrInterface<'a>>,
+            name: &Name,
+        ) -> Option<s::Field> {
+            if name == "__schema" || name == "__type" {
+                let doc = introspection_schema(SubgraphDeploymentId::new("").unwrap()).document;
```
Jannis: Same here (about creating the introspection schema every time).

graphql/src/query/mod.rs
```diff
@@ -22,6 +23,9 @@ where
 
     /// Time at which the query times out.
     pub deadline: Option<Instant>,
+
+    /// Maximum complexity for a query.
+    pub max_complexity: Option<u64>,
```
Jannis: Add `max_depth` here as well, if we decide to not hard-code it.

graphql/src/execution/execution.rs
Jannis: I wonder if this could benefit from @lutter's splitting, where the schema is decided at the top level.

graphql/src/query/mod.rs
```diff
@@ -66,17 +64,33 @@ where
...
+                query_logger,
+                "Execute query";
+                "query" => query.document.format(&Style::default().indent(0)).replace('\n', " "),
+                "complexity" => format!("{:?}", complexity),
+            );
 
+            match (complexity, options.max_complexity) {
+                (Err(e), _) => Err(vec![e]),
+                (Ok(complexity), Some(max_complexity)) if complexity > max_complexity => {
+                    Err(vec![QueryExecutionError::TooComplex(
```
Jannis: I'm realizing it would be worth logging query errors further down where we log `Finished query`. That way we capture them in the node logs and not just on the client (which often, we don't have access to).

graphql/src/subscription/mod.rs
```diff
@@ -57,13 +58,38 @@ where
...
+            info!(
+                ctx.logger,
+                "Execute subscription";
+                "query" => ctx.document.format(&Style::default().indent(0)).replace('\n', " "),
+                "complexity" => complexity,
+            );
+
+            match options.max_complexity {
+                Some(max_complexity) if complexity > max_complexity => {
+                    Err(vec![QueryExecutionError::TooComplex(complexity, max_complexity)].into())
```
Jannis: Same here about logging — if we don't log subscriptions that fail early like this yet, I think we should.

Jannis (changes requested):
Great! I definitely expected this to be easier but it all makes sense.

graph/src/data/query/error.rs
leodasvacas: That's much better!

core/src/graphql/runner.rs
leodasvacas: Sure.

graph/src/data/query/error.rs
leodasvacas: Nice thanks for the suggestion.

graphql/src/execution/execution.rs
leodasvacas: These are abitrary values. A query has to be so huge to hit this case that I don't want to write any additional code to support it. This function doesn't have `max_complexity` as a parameter so that I can easily put it here.

graph/src/data/query/error.rs
leodasvacas: The actual doesn't make sense because we stop if it hits the max. But I added the max to this error.

graphql/src/subscription/mod.rs
leodasvacas: The error is now being logged on both cases.

graphql/src/execution/execution.rs
leodasvacas: Good catch, I've put the introspection document in a `lazy_static`. This is being called before the logic for deciding the schema. Though it would be possible to call it after, then we'd sum the complexity for the introspection and regular sides.

leodasvacas:
@Jannis all comments addressed!

graph/src/data/query/error.rs
Jannis: Oh yeah, of course. Cool!

graphql/src/execution/execution.rs
Jannis: Oh, of course, this is the overflow case, which should rarely be hit. Makes sense.

Jannis (approved):

PR #872, opened 2019-04-15
Note on value of `GRAPH_GRAPHQL_MAX_COMPLEXITY`

leoyvens:
The relationship between introspection queries and query complexity is something to be discussed, but this is a useful note about the current implementation.


lutter (approved):
I am surprised that introspection produces this much data, but the docs update is certainly the right thing to do.

Jannis:
I guess it doesn't _really_ process that much, it's just that the introspection query sent by typical clients is unbounded.

leodasvacas:
GitHub doesn't seem to limit introspection, but their full introspection query returns 56k lines, so maybe they should. Anyways, merging this.

PR #874, opened 2019-04-15
Bump ipfs-api and add support for IPFS over HTTPS

### LINKED ISSUE #854: Add support for https protocol in ipfs connection
The Rust IPFS client library we use only supports host + port,

Starting the graph-node docker container with an `ipfs` url like `https://ipfs.infura.io`  will lead to a timeout error:
```
graph-node_1  | wait-for-it.sh: waiting 30 seconds for https://ipfs.infura.io
graph-node_1  | wait-for-it.sh: timeout occurred after waiting 30 seconds for https://ipfs.infura.io
client_graph-node_1 exited with code 124
```

It should just work :-)

Jannis:
Resolves #854.

HTTPS support has been merged in ipfs-api today (see https://github.com/ferristseng/rust-ipfs-api/commit/55902e98d868dcce047863859caf596a629d10ec). This PR switches to the latest master (for now) and updates our use of `IpfsClient` in `graph-node` to support any of `127.0.0.1:5001`, `localhost:5001`, `http://localhost:5001` and `https://ipfs.infura.io` (or whatever their endpoint is).

### DIFF
--- a/node/src/main.rs
+++ b/node/src/main.rs
@@ -20,11 +20,8 @@ extern crate url;
 use clap::{App, Arg};
 use futures::sync::oneshot;
 use ipfs_api::IpfsClient;
-use itertools::FoldWhile::{Continue, Done};
-use itertools::Itertools;
 use lazy_static::lazy_static;
 use std::env;
-use std::net::ToSocketAddrs;
 use std::str::FromStr;
 use std::sync::Arc;
 use std::time::Duration;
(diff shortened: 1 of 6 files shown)


node/src/main.rs
```diff
@@ -295,56 +292,52 @@ fn async_main() -> impl Future<Item = (), Error = ()> + Send + 'static {
...
+
+    // Try to create an IPFS client for this URL
+    let ipfs_client = match IpfsClient::new_from_uri(ipfs_address.as_ref()) {
+        Ok(ipfs_client) => Arc::new(ipfs_client),
+        Err(e) => {
+            error!(
+                logger,
+                "Failed to create IPFS client for `{}`: {}", &ipfs_address, e
+            );
+            panic!("Cloud not connect to IPFS");
```
leodasvacas: s/Cloud/Could

leoyvens (changes requested):
We depend on ipfs-api in a lot of places, it would be nice to centralize that in `graph` as we do for other common dependencies, but it's fine if you don't want to do that right now. Other than that, just a typo.

node/src/main.rs
Jannis: Good catch, fixed.

leoyvens (approved):

PR #875, opened 2019-04-16
graph: Log the error when retrying

leoyvens:
A subgraph is stuck retrying here and it's hard to tell why, this might help.


lutter (approved):
Nice improvement!

PR #876, opened 2019-04-16
Add env vars for range size & parallel ranges

Jannis:
The variables added are:

* `ETHEREUM_PARALLEL_BLOCK_RANGES` - defines the number of block
  ranges to request from the Ethereum node in parallel.
* `ETHEREUM_BLOCK_RANGE_SIZE` - defines the number of blocks to
  request events for in each request.

Tweaking these may be necessary to get past ranges of very dense blocks, see https://thegraph.com/explorer/subgraph/nicosantangelo/watchtower?selected=logs (the pending version) for example.


fordN (approved):
Looks good 👍

PR #882, opened 2019-04-18
Make subgraph logs less scary by using "updates" instead of "restarts"

Jannis:
Just a super small tweak.


leoyvens (approved):
It really isn't a hard restart, so seems appropriate.

PR #885, opened 2019-04-20
Reduce logging volume by removing unnecessary logs

fordN:
A significant portion of the log volume has been taken up by messaging that is not strictly necessary for monitoring progress of the specific subgraph or tracking down errors.  
Many log lines were simply noting inaction such as: "0 events found in this block for this subgraph" or "Creating 0 dynamic data sources".

Removing these logs serves to cut down on the noise, making the logs easier to parse. Also, the overall volume reduction makes it more efficient to send the logs to a storage engine such as Elasticsearch.

### DIFF
--- a/datasource/ethereum/src/block_stream.rs
+++ b/datasource/ethereum/src/block_stream.rs
@@ -216,8 +216,6 @@ where
         let ctx = self.clone();
         let reorg_threshold = ctx.reorg_threshold;
 
-        debug!(ctx.logger, "Identify next step");
-
         // Get pointers from database for comparison
         let head_ptr_opt = ctx.chain_store.chain_head_ptr().unwrap();
         let subgraph_ptr = ctx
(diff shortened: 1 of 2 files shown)


node/src/main.rs
```diff
@@ -367,7 +365,7 @@ fn async_main() -> impl Future<Item = (), Error = ()> + Send + 'static {
     let eth_adapter = Arc::new(graph_datasource_ethereum::EthereumAdapter::new(transport));
 
     // Ask Ethereum node for network identifiers
-    info!(
+    debug!(
```
Jannis: Can we keep this file as is? It's useful to see the trying IPFS node message and it's useful to have all connection attempt messages as info. There are more like these and they are all info as well.

fordN: Works for me! Especially since the starting connection messages only log on startup, so removing them doesn't have any noticeable effect on overall log volume.

datasource/ethereum/src/block_stream.rs
```diff
@@ -234,17 +232,6 @@ where
 
         let head_ptr = head_ptr_opt.unwrap();
 
-        debug!(
-            ctx.logger, "Chain head pointer";
-            "hash" => format!("{:?}", head_ptr.hash),
-            "number" => &head_ptr.number
-        );
-        debug!(
```
leodasvacas: Can we downgrade these two to `trace!` instead of removing?

PR #887, opened 2019-04-22
Split entities table into one table per subgraph

lutter:
This PR makes it so that newly deployed subgraphs will be stored in their own `entities` (and `entity_histroy`) table. Deploying a new subgraph will create a schema in Postgres, and store the data for the subgraph in that schema.

The PR does not migrate existing data; instead, queries (and changes) against existing data are still done against the existing `public.entities` table. Migrating data will be the subject of another PR.

(Opening this as draft as I haven't tested this other than running unit tests; will remove draft status once I've done that. The code is ready for review, though)

### DIFF
--- a/store/postgres/Cargo.toml
+++ b/store/postgres/Cargo.toml
@@ -5,6 +5,9 @@ edition = "2018"
 
 [dependencies]
 diesel = { version = "1.3.2", features = ["postgres", "serde_json", "numeric", "r2d2"] }
+# We use diesel-dynamic-schema straight from git as the project has not
+# made a release as a crate yet
+diesel-dynamic-schema = { git = "https://github.com/diesel-rs/diesel-dynamic-schema", rev="a8ec4fb1" }
 diesel_migrations = "1.3.0"
 failure = "0.1.2"
 fallible-iterator = "0.1.4"
--- a/store/postgres/Cargo.toml
+++ b/store/postgres/Cargo.toml
@@ -14,6 +14,7 @@ fallible-iterator = "0.1.4"
 futures = "0.1.21"
 graph = { path = "../../graph" }
 graph-graphql = { path = "../../graphql" }
+Inflector = "0.11.3"
 lru_time_cache = "0.8"
 postgres = "0.15.2"
 serde = "1.0"
--- a/store/postgres/migrations/2019-04-19-171709_split_entities/up.sql
+++ b/store/postgres/migrations/2019-04-19-171709_split_entities/up.sql
@@ -1,0 +1,106 @@
+-- metadata to track the DB schemas we create for subgraph deployments
+
+-- The name of the DB schema corresponding to a deployment
+create table deployment_schemas(
+  id         serial primary key,
+  subgraph   varchar unique not null,
+  name       varchar not null default 'sgd' || currval('deployment_schemas_id_seq')
+);
+
+insert into deployment_schemas(id, subgraph, name)
+  values(0, 'subgraphs', 'subgraphs');
+
+-- Schema for the subgraph of subgraphs
+-- Note that this schema does not have a entity_history table
+-- because we do not collect history for the subgraph of subgraphs
+create schema subgraphs;
+
+create table subgraphs.entities (
+  entity       varchar not null,
+  id           varchar not null,
+  data         jsonb,
+  event_source varchar not null,
+  primary key(entity, id)
+);
+
+insert into subgraphs.entities
+  select entity, id, data, event_source
+  from public.entities
+  where subgraph='subgraphs';
+
+delete from public.entities
+  where subgraph='subgraphs';
+
+--
+-- Log entity history into sgdN.entities (without a subgraph field)
+--
+CREATE OR REPLACE FUNCTION public.subgraph_log_entity_event()
+ RETURNS trigger
+ LANGUAGE plpgsql
+AS $function$
+DECLARE
+    event_id INTEGER;
+    new_event_id INTEGER;
+    is_reversion BOOLEAN := FALSE;
+    operation_type INTEGER := 10;
+    event_source  VARCHAR;
+    entity VARCHAR;
+    entity_id VARCHAR;
+    data_before JSONB;
+BEGIN
+    -- Get operation type and source
+    IF (TG_OP = 'INSERT') THEN
+        operation_type := 0;
+        event_source := NEW.event_source;
+        entity := NEW.entity;
+        entity_id := NEW.id;
+        data_before := NULL;
+    ELSIF (TG_OP = 'UPDATE') THEN
+        operation_type := 1;
+        event_source := NEW.event_source;
+        entity := OLD.entity;
+        entity_id := OLD.id;
+        data_before := OLD.data;
+    ELSIF (TG_OP = 'DELETE') THEN
+        operation_type := 2;
+        event_source := current_setting('vars.current_event_source', TRUE);
+        entity := OLD.entity;
+        entity_id := OLD.id;
+        data_before := OLD.data;
+    ELSE
+        RAISE EXCEPTION 'unexpected entity row operation type, %', TG_OP;
+    END IF;
+
+    IF event_source = 'REVERSION' THEN
+        is_reversion := TRUE;
+    END IF;
+
+    SELECT id INTO event_id
+    FROM event_meta_data
+    WHERE db_transaction_id = txid_current();
+
+    new_event_id := null;
+
+    IF event_id IS NULL THEN
+        -- Log information on the postgres transaction for later use in
+        -- revert operations
+        INSERT INTO event_meta_data
+            (db_transaction_id, db_transaction_time, source)
+        VALUES
+            (txid_current(), statement_timestamp(), event_source)
+        RETURNING event_meta_data.id INTO new_event_id;
+    END IF;
+
+    -- Log row metadata and changes, specify whether event was an original
+    -- ethereum event or a reversion
+    EXECUTE format('INSERT INTO %I.entity_history
+        (event_id, entity_id, entity,
+         data_before, reversion, op_id)
+      VALUES
+        ($1, $2, $3, $4, $5, $6)', TG_TABLE_SCHEMA)
+    USING COALESCE(new_event_id, event_id), entity_id, entity,
+          data_before, is_reversion, operation_type;
+    RETURN NULL;
+END;
+$function$
+;
--- a/store/postgres/src/entities.rs
+++ b/store/postgres/src/entities.rs
@@ -1,3 +1,4 @@
+use diesel::connection::SimpleConnection;
 use diesel::deserialize::QueryableByName;
 use diesel::dsl::{any, sql};
 use diesel::pg::{Pg, PgConnection};
@@ -9,10 +10,12 @@ use diesel::{OptionalExtension, QueryDsl, RunQueryDsl};
 use diesel_dynamic_schema::{schema, Column, Table as DynamicTable};
 use graph::data::subgraph::schema::SUBGRAPHS_ID;
 use graph::prelude::{
-    format_err, EntityChange, EntityChangeOperation, EntityFilter, EntityKey, Error, EventSource,
-    QueryExecutionError, StoreError, StoreEvent, SubgraphDeploymentId, TransactionAbortError,
+    format_err, AttributeIndexDefinition, EntityChange, EntityChangeOperation, EntityFilter,
+    EntityKey, Error, EventSource, QueryExecutionError, StoreError, StoreEvent,
+    SubgraphDeploymentId, TransactionAbortError, ValueType,
 };
 use graph::serde_json;
+use inflector::cases::snakecase::to_snake_case;
 
 use crate::filter::{build_filter, store_filter};
 use crate::functions::set_config;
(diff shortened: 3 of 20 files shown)


lutter:
After the small fixes I made, this works for me now, both with existing entities, a manually migrated subgraph, and a freshly deployed one. As always, somebody else checking is hugely appreciated.

store/postgres/src/entities.rs
```diff
@@ -685,6 +688,67 @@ impl Table {
...
+                    entity_name=&index.entity_name);
+                conn.batch_execute(&*query)?;
+                Ok(1)
+            }
+            Table::Split(entities) => {
+                let name = format!(
+                    "entities_{}_{}",
+                    to_snake_case(&index.entity_name),
+                    to_snake_case(&index.attribute_name)
+                );
```
lutter: This creates an index of the form `entities_domain_labelhash`. I think we should give these indexes names that makes it easier to identify them as attribute indexes (and tell them apart from the `entities_pkey` index we need) I suggest calling them `attr_domain_labelhash` - that'll also give users 4 more characters before they run into the 63 char identifier limitation in pg ;)

leodasvacas:
If I may ask a naive question, did you consider the approach of creating a view of the split tables that looks like our current entities table and letting Diesel treat that as single table? I've read that not all views are updatable, presumably such a view would have that issue?

lutter:
> If I may ask a naive question, did you consider the approach of creating a view of the split tables that looks like our current entities table and letting Diesel treat that as single table? I've read that not all views are updatable, presumably such a view would have that issue?

Yes, with a view we would have to put a trigger on the view that shuffles the data into the right tables; that would force us into a situation where every CRUD operation on entities is inherently a single-object operation and preclude optimizations I'd like to do further down the line like using bulk inserts (insert with many entries in the `VALUES` clause or an outright `COPY` statement)

What makes this awkward with Diesel is that Diesel's design is fundamentally based on the idea that you know the entire schema at compile time, which does not apply to our situation.

store/postgres/src/entities.rs
```diff
@@ -0,0 +1,993 @@
...
+struct Schema {
+    id: i32,
+    subgraph: String,
+    name: String,
+}
+
+type EntityColumn<ST> = Column<DynamicTable<String>, String, ST>;
+
+#[derive(Debug, Clone)]
+pub(crate) struct SplitTable {
```
Jannis: What does split table mean here?

Jannis: IMHO splitting is just the action we're taking here, I'd give this a more permanent name. `DeploymentEntitiesTable` or something similar.

Jannis: What confused me here was that I thought the table in itself was split somehow.

store/postgres/src/entities.rs
```diff
@@ -0,0 +1,993 @@
...
+        key: &EntityKey,
+        event_source: EventSource,
+    ) -> Result<usize, StoreError> {
+        diesel::select(set_config(
+            "vars.current_event_source",
+            event_source.to_string(),
+            true,
+        ))
+        .execute(conn)
+        .map_err(|e| format_err!("Failed to save event source for remove operation: {}", e))
```
Jannis: Could we call this 'set' instead of 'save'.

store/postgres/src/entities.rs
```diff
@@ -0,0 +1,993 @@
...
+
+        Ok(StoreEvent::new(changes))
+    }
+
+    fn build_attribute_index(
+        &self,
+        conn: &PgConnection,
+        index: &AttributeIndexDefinition,
+    ) -> Result<usize, StoreError> {
+        println!("BUILD_INDEX {:?}", &index);
```
Jannis: Should this be here?

store/postgres/src/entities.rs
Jannis: Sound good to me!

store/postgres/src/entities.rs
```diff
@@ -0,0 +1,993 @@
...
+                Ok(1)
+            }
+        }
+    }
+}
+
+/// Delete all entities. This function exists solely for integration tests
+/// and should never be called from any other code. Unfortunately, Rust makes
+/// it very hard to export items just for testing
+pub fn delete_all_entities(conn: &PgConnection) -> Result<usize, StoreError> {
```
Jannis: To make it even more obvious that this is dangerous, we could put 'dangerous' or similar into the name of the function.

store/postgres/src/entities.rs
```diff
@@ -0,0 +1,993 @@
...
+    subgraph_id: &SubgraphDeploymentId,
+) -> Result<(), StoreError> {
+    // Create a schema for the deployment
+    let schemas: Vec<String> = diesel::insert_into(sgd_schemas::table)
+        .values(sgd_schemas::subgraph.eq(subgraph_id.to_string()))
+        .returning(sgd_schemas::name)
+        .get_results(conn)?;
+    let schema_name = schemas
+        .first()
+        .ok_or_else(|| format_err!("failed to read schema name for {} back", subgraph_id))?;
```
Jannis: Our error messages start with a capital letter.

Jannis: There's more cases like this in this and other files.

store/postgres/src/entities.rs
```diff
@@ -0,0 +1,993 @@
...
+            id           varchar not null,
+            data         jsonb,
+            event_source varchar not null,
+            primary key(entity, id)
+        )",
+        schema_name
+    );
+    conn.batch_execute(&*query)?;
+
+    // TODO: should we check that event_source is not "none"?
```
Jannis: Is this an open question still?

store/postgres/src/entities.rs
```diff
@@ -0,0 +1,993 @@
...
+}
+
+#[allow(dead_code)]
+pub(crate) fn drop_schema(
+    conn: &diesel::pg::PgConnection,
+    subgraph: &SubgraphDeploymentId,
+) -> Result<usize, StoreError> {
+    let info = find_schema(conn, subgraph)?;
+    if let Some(schema) = info {
+        let query = format!("drop schema if exists {} cascade", schema.name);
```
Jannis: Will cascade delete triggers as well?

store/postgres/src/store.rs
```diff
@@ -316,102 +301,65 @@ impl Store {
...
-                    "{}, query = {:?}",
-                    e, diesel_query_debug_info
-                ))
-            })
+        conn.query(
+            &query.subgraph_id,
+            query.entity_types,
+            query.filter,
+            order,
+            query.range.first.unwrap_or(0),
```
Jannis: The default for first is 100, not 0.

store/postgres/src/entities.rs
leodasvacas: A trick I've used is to stick `#[cfg(debug)]` on top of test-only builds, this way it can never make it into the release builds we use for prod.

store/postgres/src/entities.rs
leodasvacas: Though the api guidelines says they should be like this, imo we should switch our convention to what the guidelines say https://rust-lang-nursery.github.io/api-guidelines/interoperability.html#examples-of-error-messages.

store/postgres/src/store.rs
leodasvacas: Should it be mandatory here? Isn't this that case where it would cause our internal unlimited queries to have a limit of 100?

store/postgres/src/entities.rs
```diff
@@ -0,0 +1,993 @@
...
+
+/// Marker trait for tables that store entities
+pub(crate) trait EntitySource {}
+
+// The entities and related tables in the public schema. We put them in
+// this module to make sure that nobody else gets access to them. All access
+// to these tables must go through functions in this module.
+mod public {
+    table! {
+        entities (id, subgraph, entity) {
```
leodasvacas: For readability, lets put the primary key in the order it actually is.

store/postgres/src/entities.rs
Jannis: OK, I am willing to relax my urge to begin all sentences we present to users with a capital letter. ;)

store/postgres/Cargo.toml
```diff
@@ -5,12 +5,14 @@ edition = "2018"
 
 [dependencies]
 diesel = { version = "1.3.2", features = ["postgres", "serde_json", "numeric", "r2d2"] }
+diesel-dynamic-schema = { git = "https://github.com/diesel-rs/diesel-dynamic-schema", rev="a8ec4fb1" }
```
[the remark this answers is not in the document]
lutter: Done. In a nutshell, they have not made an official release

store/postgres/migrations/2019-04-19-171709_split_entities/up.sql
```diff
@@ -0,0 +1,106 @@
...
+  subgraph   varchar unique not null,
+  name       varchar not null default 'sgd' || currval('sgd_schemas_id_seq')
+);
+
+insert into sgd_schemas(id, subgraph, name)
+  values(0, 'subgraphs', 'subgraphs');
+
+-- Schema for the subgraph of subgraphs
+-- Note that this schema does not have a entity_history table
+-- because we do not collect history for the subgraph of subgraphs
```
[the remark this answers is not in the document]
lutter: It will only freak me out if we collect too much history again ;) As it is right now, we won't collect any history for dynamic data sources, because the triggers are set up to not fire for the `subgraphs` subgraph. I think we should add the `subgraphs.entity_history` table when we address that.

store/postgres/src/store.rs
Jannis: We have internal queries that can have arbitrary complexity but at least in the dynamic data sources loading code, we paginate based on the assumption that the limit is `100` per collection.

store/postgres/migrations/2019-04-19-171709_split_entities/up.sql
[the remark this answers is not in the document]
Jannis: "When we address that" — ideally the "when" would be now. If you're saying this already doesn't work on master, then let's create an issue for it, because we're really going to need that for dynamic data sources to be reliable with regards to block reorgs.

store/postgres/migrations/2019-04-19-171709_split_entities/up.sql
```diff
@@ -0,0 +1,106 @@
+-- metadata to track the DB schemas we create for subgraph deployments
+
+-- The name of the DB schema corresponding to a deployment
+create table sgd_schemas(
```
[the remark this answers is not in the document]
lutter: Yep, I renamed it. We should have a longer convo about code readability at some point since we'll keep going back and forth on these. My basic point is that what makes code hard to read are concepts that are not well explained (and I definitely should have put more comments on the `SplitTable` struct to make it easier to understand what that's for) Anyway, I changed the name of the table.

store/postgres/src/entities.rs
lutter: Yeah, I should have explained in a comment more clearly what that's for; added a bit of an explanation

store/postgres/src/entities.rs
lutter: Yep, done.

store/postgres/src/entities.rs
lutter: Not at all, removing it

store/postgres/src/entities.rs
lutter: I renamed it to `delete_all_entities_for_test_use_only`

store/postgres/src/entities.rs
lutter: Yes, the triggers are attached to the `entities` table, and they will get removed when the table is dropped (of course, it won't get rid of the actual stored procedure that the triggers run)

store/postgres/src/store.rs
lutter: I was a little overzealous in making the `entities` API low-level and barebones. `0` signals 'do not restrict the number of entities'. I'll just pass the original `Option` through to make that more obvious.

store/postgres/src/entities.rs
lutter: Done

lutter: Actually, this was copied verbatim from `db_schema`, and when I change the order of pk columns to `entities (subgraph, entity, id)` I get a test failure:
```
thread 'reference_interface' panicked at 'assertion failed: `(left == right)`
  left: `"Object({\"leggeds\": List([Object({\"leg\": Null})])})"`,
 right: `"Object({\"leggeds\": List([Object({\"leg\": Object({\"id\": String(\"1\")})})])})"`', core/tests/interfaces.rs:180:5
note: Run with `RUST_BACKTRACE=1` environment variable to display a backtrace.


failures:
    reference_interface
``` 

I have no idea why that is; we should probably investigate that, but as part of a separate ticket. I'll leave that as is for now.

store/postgres/migrations/2019-04-19-171709_split_entities/up.sql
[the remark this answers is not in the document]
lutter: Yes, it won't work now because the triggers explicitly exclude the `subgraphs` subgraph. That's also what my `TODO` comment was about: I am not sure if `EventSource::None` is a reliable enough indicator whether history should be collected or not. Actually, looking at [this](https://github.com/graphprotocol/graph-node/blob/master/store/postgres/src/store.rs#L904-L913) makes it clear that if we go by `EventSource::None` we will collect history for each block pointer update (or at least the ones that also have entity changes)

Because the right fix seems a little more involved, I would prefer capturing that in a separate issue and addressing it on its own.

store/postgres/src/entities.rs
leodasvacas: Ah it must expect the columns in that order for `find` queries. Sure lets leave it as a refactor for later.

store/postgres/src/entities.rs
lutter: It is, but I am removing the comment; as the discussion around reverting dynamic data sources shows, there will be a little more work involved with determining the right condition for collecting history, at least for the subgraph of subgraphs. For the other ones, it is fine to always collect history (which is what this comment applied to).

store/postgres/src/entities.rs
lutter: @leodasvacas just saw your comment about `#[cfg(debug)]` now. When I stick that into the code, the store tests fail to compile; when I also annotate where this gets called in `remove_test_data` in the store tests, I get a warning about `conn` not being used in VSCode (but not release builds strangely). I'll leave that out therefore, and just stick with the scary name.

store/postgres/migrations/2019-04-19-171709_split_entities/up.sql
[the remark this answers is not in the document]
lutter: One more wrinkle: our revert procedure fundamentally assumes we only revert for a single subgraph. With reversions in `subgraphs`, we need to make sure we change that, too.

lutter:
Addressed all comments (I believe)

store/postgres/migrations/2019-04-19-171709_split_entities/up.sql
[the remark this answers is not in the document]
Jannis: I agree. There's an amount of ambiguity in the terms we use (like "subgraph" is being used for a bunch of things). I don't feel like introducing acronyms is the answer though. Let's discuss at some point! 😄

store/postgres/migrations/2019-04-19-171709_split_entities/up.sql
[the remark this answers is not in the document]
Jannis: Yep, ok. I've added https://github.com/graphprotocol/graph-node/issues/890 to track this.

store/postgres/src/entities.rs
Jannis: I still find `Split` and `SplitTable` to not be great names. `Dedicated`? `PerSubgraph`, `Subgraph`, `DeploymentEntities` (putting it back on the table 😋)...?

store/postgres/src/entities.rs
lutter: I just changed the code so that the index name is guaranteed to be unique even if the identifier is longer than pg's 63 char limit (by including the entity and attribute numbers)

store/postgres/src/entities.rs
leodasvacas: @lutter sorry my suggestion was wrong, the correct attribute is `#[cfg(debug_assertions)]`.

Issue #756, opened 2019-02-21
Add function signature support for anonymous events 


davekaj:
This feature request is specifically to allow for a Maker subgraph to work. Right now, it is impossible to get any CDP data from events because of how they use anonyomous events. 

See the Solidity code here, for anonymous events : https://github.com/dapphub/ds-note/blob/master/src/note.sol
See solidity docs for anonymous explained: https://solidity.readthedocs.io/en/v0.5.4/contracts.html#events

Basically, the anonymous modifier removes the event signature as topic0, which is what the graph node searches for while indexing. But `contract DSNote` sets up the `LogNote` event to have the signature of the method that was called to be topic0, thus the contract events are easily indexable by what functions are emitting events, rather than just the events themselves. 

This is not a totally common pattern, I have only seen Maker use it. However they are so big, and we have a few teams who want to build this subgraph, so it would be great to add this functionality soon.

Note that the method signature is only four bytes, so the graph node will be searching for `0x1234567800000000......`. see an example here : https://etherscan.io/address/0x448a5065aebb8e423f0896e6c5d525c040f59af3#events (side note, I just noticed these events also emit some event topics that are not bytes32, which link back to decetraland and livepeer, and the ethabi not supporting the broken solidity compiler of the past)

At first I thought the simplest solution would be to do something like the following in the `subgraph.yaml` :
```
- anonymousEvent: CreateCDP(address,uint256,uint256)
- handler: handleCreateCDP
```
Then, when graph node sees that it is anonymousEvent instead of event, it will only take the first 4 bytes of the hash, and append the other 28 bytes with `0`'s, which will allow it to find the anonymous events.

HOWEVER, it makes me think, that this doesn't necessarily prepare the graph node to deal with ALL anoyonmous events. You could easily make an anoymous event, and index the user address as topic 0, which would take the user address instead, and append 12 bytes to the end. 

Maybe a better solution would be to require that the user pass in the hashed value directly, and leave it up the the subgraph creator to predetermine the hash, rather than have the graph node hash it. So it would look like:

Handles method signatures :
```
- anonymousEvent: 0x123456780000000000000000000000000000000000000000000000000000000000000000
- handler: handleUserAddress
```
Handles addresses:
```
- anonymousEvent: 0x1234567890123456789012345678901234567890123456789000000000000000000000000
- handler: handleCreateCDP
```
This is what I would recommend we do

davekaj:
@leodasvacas @Jannis

[...]

PR #888, opened 2019-04-22
Custom topic0

leodasvacas:
Resolves https://github.com/graphprotocol/graph-node/issues/756. I haven't properly tested this yet. @davekaj would you have time to also give this a shot, along with https://github.com/graphprotocol/graph-cli/pull/246?

### DIFF
--- a/docs/subgraph-manifest.md
+++ b/docs/subgraph-manifest.md
@@ -64,6 +64,7 @@ The `mapping` field may be one of the following supported mapping manifests:
 | --- | --- | --- |
 | **event** | *String* | An identifier for an event that will be handled in the mapping script. For Ethereum contracts, this must be the full event signature to distinguish from events that may share the same name. No alias types can be used. For example, uint will not work, uint256 must be used.|
 | **handler** | *String* | The name of an exported function in the mapping script that should handle the specified event. |
+| **topic0** | optional *String* | A `0x` prefixed hex string. If provided, logs whose topic 0 is equal to this value will be processed by the given handler. By default, this is equal to the hash of the event signature. This is useful for processing solidity anonymous events. |
 
 ## 1.6 Path
 A path has one field `path`, which either refers to a path of a file on the local dev machine or an [IPLD link](https://github.com/ipld/specs/).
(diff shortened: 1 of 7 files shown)


davekaj:
yes i will test it tomorrow!

graph/src/data/graphql/values.rs
```diff
@@ -37,6 +37,20 @@ impl TryFromValue for H160 {
     }
 }
 
+impl TryFromValue for H256 {
+    fn try_from_value(value: &Value) -> Result<Self, Error> {
+        match value {
+            Value::String(s) => {
+                // `H160::from_str` takes a hex string with no leading `0x`.
```
Jannis: I believe this should say `H256`

leodasvacas: Thanks

docs/subgraph-manifest.md
```diff
@@ -64,6 +64,7 @@ The `mapping` field may be one of the following supported mapping manifests:
 | --- | --- | --- |
 | **event** | *String* | An identifier for an event that will be handled in the mapping script. For Ethereum contracts, this must be the full event signature to distinguish from events that may share the same name. No alias types can be used. For example, uint will not work, uint256 must be used.|
 | **handler** | *String* | The name of an exported function in the mapping script that should handle the specified event. |
+| **topic0** | optional *String* | A `0x` prefixed hex string. If provided, logs whose topic 0 is equal to this value will be processed by the given handler. By default, this is equal to the hash of the event signature. This is useful for processing solidity anonymous events. |
```
Jannis: solidity anonymous events -> anonymous events in Solidity

Jannis: I'd also mention that if topic0 is provided, the subgraph will index events with matching to topic0 only, _instead_ of all events with the given signature.

davekaj: I added in Jannis' comments, and some of my own for what I think would make sense to a Solidity developer

A `0x` prefixed hex string. If provided, logs whose topic0 is equal to this value will be processed by the given handler. When topic0 is provided, the subgraph will _only_ match the topic0 hash, and it will not match on the hash of the event signature. This is useful for processing anonymous events in Solidity, which can have their topic0 set to anything.  By default, topic0 is equal to the hash of the event signature.

Jannis: 👍 Just one thing: Let's talk about `events` rather than `logs` wherever we can. It makes our language more consistent.

leodasvacas:
@davekaj Took a look, what happens is that `give` is a function, not an event.

docs/subgraph-manifest.md
leodasvacas: Thanks for the suggestions! I've updated these docs.

davekaj:
I can confirm the test worked on the maker subgraph

PR #891, opened 2019-04-24
Use a shared `ChainHeadUpdateListener` to avoid too many Postgres connections

### LINKED ISSUE #889: Dynamic data sources occasionally cause too many Postgres connections
We restart subgraph block streams whenever new data sources were added in a block. Each stream creates a new chain head update listener. We try to shut the previous stream and update listener down ASAP but the connections don't seem to be cleaned up fast enough in some occasions.

This can be fixed by using a shared chain head update listener that all subgraphs on the node consume; this would only require a single extra connection for the entire node.

Jannis:
This fixes #889 by using a single Postgres connection for the chain
head update streams of all subgraph deployments.

### DIFF
--- a/store/postgres/src/chain_head_listener.rs
+++ b/store/postgres/src/chain_head_listener.rs
@@ -1,40 +1,60 @@
-use crate::notification_listener::{NotificationListener, SafeChannelName};
+use futures::sync::mpsc::{channel, Sender};
+use std::collections::HashMap;
+use std::sync::{Arc, RwLock};
+use uuid::Uuid;
+
 use graph::prelude::{ChainHeadUpdateListener as ChainHeadUpdateListenerTrait, *};
 use graph::serde_json;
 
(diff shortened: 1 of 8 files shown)


store/postgres/src/chain_head_listener.rs
```diff
@@ -1,40 +1,60 @@
...
+    subscribers: ChainHeadUpdateSubscribers,
+    _listener: NotificationListener,
 }
 
 impl ChainHeadUpdateListener {
     pub fn new(logger: &Logger, postgres_url: String, network_name: String) -> Self {
+        let logger = logger.new(o!("component" => "ChainHeadUpdateListener"));
+        let subscribers = Arc::new(RwLock::new(HashMap::new()));
+
+        // Create a Postgres notification listene for chain head updates
```
leodasvacas: s/listene/listener

leoyvens (approved):
lgtm, just a typo

PR #906, opened 2019-05-02
Fix writing and loading of dynamically created data sources

Jannis:
This fixes two problems:

1. When writing dynamic data sources entities to the store, the mapping entity was missing. This caused dynamic data sources loading on restarts to fail.

2. There was a typo in the code that parses a mapping entity back from a GraphQL value. This would also cause dynamic data sources to not load correctly.


ghost (approved):

leodasvacas:
We should get a macro for deriving these things.

PR #908, opened 2019-05-03
Denormalize `entityCount`

lutter:
This PR stores the `entityCount` as an attribute of the `SubgraphDeployment`, and keeps it updated as entities are added and removed from the subgraph.

### DIFF
--- a/graph/src/data/subgraph/mod.rs
+++ b/graph/src/data/subgraph/mod.rs
@@ -22,7 +22,7 @@ use crate::data::subgraph::schema::{
     EthereumBlockHandlerEntity, EthereumCallHandlerEntity, EthereumContractAbiEntity,
     EthereumContractDataSourceEntity, EthereumContractDataSourceTemplateEntity,
     EthereumContractDataSourceTemplateSourceEntity, EthereumContractEventHandlerEntity,
-    EthereumContractMappingEntity, EthereumContractSourceEntity,
+    EthereumContractMappingEntity, EthereumContractSourceEntity, SUBGRAPHS_ID,
 };
 use crate::util::ethereum::string_to_h256;
 
@@ -67,6 +67,12 @@ impl SubgraphDeploymentId {
         Link {
             link: format!("/ipfs/{}", self),
         }
+    }
+
+    /// Return true if this is the id of the special
+    /// "subgraph of subgraphs" that contains metadata about everything
+    pub fn is_meta(&self) -> bool {
+        self.0 == *SUBGRAPHS_ID.0
     }
 }
 
--- a/graphql/src/store/resolver.rs
+++ b/graphql/src/store/resolver.rs
@@ -1,6 +1,5 @@
 use graphql_parser::{query as q, schema as s};
 use std::collections::{BTreeMap, HashMap};
-use std::ops::Deref;
 use std::result;
 use std::sync::Arc;
 
@@ -161,67 +160,6 @@ impl<S> Resolver for StoreResolver<S>
 where
     S: Store,
 {
-    /// Resolves a scalar value for a given scalar type.
-    fn resolve_scalar_value(
-        &self,
-        parent_object_type: &s::ObjectType,
-        parent: &BTreeMap<String, q::Value>,
-        field: &q::Field,
-        scalar_type: &s::ScalarType,
-        value: Option<&q::Value>,
-    ) -> Result<q::Value, QueryExecutionError> {
-        let subgraph_id = parse_subgraph_id(parent_object_type).unwrap();
-
-        // Extract __typename from the parent object
-        let typename = parent
-            .get("__typename")
-            .and_then(|typename| match typename {
-                q::Value::String(typename) => Some(typename),
-                _ => None,
-            })
-            .expect("GraphQL object must have `__typename`");
-
-        match (
-            subgraph_id.deref().as_str(),
-            typename.as_str(),
-            field.name.as_str(),
-        ) {
-            // Compute `entityCount` on-demand
-            ("subgraphs", "SubgraphDeployment", "entityCount") => {
-                // Extract the entity ID
-                let id = parent
-                    .get("id")
-                    .and_then(|id| match id {
-                        q::Value::String(id) => Some(id),
-                        _ => None,
-                    })
-                    .expect("GraphQL object must have an `id`");
-
-                // Convert the ID to a subgraph deployment ID
-                let hash =
-                    SubgraphDeploymentId::new(id.clone()).expect("invalid subgraph ID in database");
-
-                // Return the entity count of the deployment as a BigInt
-                Ok(Value::from(
-                    self.store
-                        .count_entities(hash)
-                        .map_err(QueryExecutionError::StoreError)?,
-                )
-                .into())
-            }
-            _ => value.map_or(Ok(q::Value::Null), |value| {
-                value.coerce(scalar_type).ok_or_else(|| {
-                    QueryExecutionError::ScalarCoercionError(
-                        field.position.clone(),
-                        field.name.to_owned(),
-                        value.clone(),
-                        scalar_type.name.to_owned(),
-                    )
-                })
-            }),
-        }
-    }
-
     fn resolve_objects(
         &self,
         parent: &Option<q::Value>,
--- a/store/postgres/migrations/2019-05-03-164052_store_entity_count/up.sql
+++ b/store/postgres/migrations/2019-05-03-164052_store_entity_count/up.sql
@@ -1,0 +1,46 @@
+create function store_entity_count()
+returns void
+language plpgsql
+as $function$
+declare
+  deployment   record;
+  entity_count int8;
+begin
+  -- Initialize the entityCount for all subgraphs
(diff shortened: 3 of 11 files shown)


lutter:
I just realized one problem with the PR: the count should not be set from `count_entities` since that's not on the write path, and the cached count will be incorrect.

store/postgres/src/entities.rs
```diff
@@ -677,20 +697,72 @@ impl Table {
...
             }
             Table::Split(entities) => {
                 let table = entities.table.clone();
                 let count: i64 = table.count().get_result(conn)?;
-                Ok(count as u64)
+                (count as u64, &entities.subgraph)
             }
-        }
+        };
+        let query = format!(
```
Jannis: This is one nasty side effect. I'd rather not do it this way. Instead, I would move the side effect of setting the `entityCount` into the store if there is no `entityCount` yet.

I get the idea of transitioning subgraphs over to have `entityCount` set incrementally, but I'd prefer for the "get and update" logic to live in the store.

Jannis: It could live even higher, in the GraphQL resolver, because at the time an `entityCount` field value exists on the deployment entity, the resolver doesn't even have to go to the store anymore.

So the `StoreResolver` could call `compute_and_return_entity_count` or similar if `entityCount` is absent and otherwise just return its value.

store/postgres/src/entities.rs
```diff
@@ -677,20 +697,72 @@ impl Table {
...
+        // do have an entity count without roundtripping lots of data
+        // to the server. Note that `coalesce` only evaluates the arguments
+        // it needs, so that in the case where we already have an entityCount
+        // we do not needlessly cause a count of entities
+        let query = format!(
+            "
+            update subgraphs.entities
+            set data = data || (format('{{\"entityCount\":
+                                  {{ \"data\": %s, 
+                                    \"type\": \"Int\"}}}}',
```
Jannis: `entityCount` is a `BigInt` in the subgraph of subgraphs schema; and if we want the GraphQL resolver to return it as is (if it exists), it should probably be a `BigInt` here too.

store/postgres/src/entities.rs
lutter: My problem with doing it this way is twofold: (1) it adds writes to a code path that should really only be reads (2) the subgraph writer and this code could race against each other, and if the writer loses the race, we have an incorrect count in there.

I think we have no choice on the query side but to run `select count(*) ...` if the `entityCount` isn't set yet, and only initialize it in the code that writes to the subgraph anyway. The other option with initialization would be to do it in a migration, or on subgraph start, but that would mean that startup would be very slow when this gets rolled out. Though thinking about it, we probably should initialize it in a migration because with ENS, a few queries that kick off a `select count(*) ..` are catastrophic for the server.

store/postgres/src/entities.rs
lutter: Good point. I changed that to `BigInt`

lutter:
I revamped the PR a bit. Changes include:

* a migration to initialize the `entityCount` property of all existing `SubgraphDeployment`
* initialize `entityCount` for new deployments with `0`
* remove special treatment of `entityCount` from GraphQL resolver
* it is possible to force a recount, which has to happen from the subgraph writer to ensure accuracy, by setting `entityCount` to `NULL` or to the string `"00"` (since `BigInt` is stored as a string)
* added some checks to the unit tests

I also tried this out with a production dump, where I let it run for a bit and verified that the entity count after some changes was still accurate.

store/postgres/src/entities.rs
```diff
@@ -676,21 +687,53 @@ impl Table {
...
+        // is `NULL` or `00`, forcing `coalesce` to evaluate its second
+        // argument, the query to count entities. In all other cases,
+        // `coalesce` does not evaluate its second argument
+        let query = format!(
+            "
+            update subgraphs.entities
+            set data = data || (format('{{\"entityCount\":
+                                  {{ \"data\": \"%s\", 
+                                    \"type\": \"BigInt\"}}}}',
+                                  coalesce((nullif(data->'entityCount'->>'data', '00'))::numeric + $1, ({}))))::jsonb
```
fordN: Great idea to use '00'!

fordN:
There's a small typo in the 54f4efcbe9e0f74c0370677a2aa5f4e57112e85c commit message that's worth updating before we merge (revresion -> reversion).

graph/src/data/subgraph/mod.rs
```diff
@@ -68,6 +68,12 @@ impl SubgraphDeploymentId {
             link: format!("/ipfs/{}", self),
         }
     }
+
+    /// Return true if this is the id of the special
+    /// "subgraph of subgraphs" that contains metadata about everything
+    pub fn is_meta(&self) -> bool {
+        self.0 == "subgraphs"
```
Jannis: Not `self == *SUBGRAPHS_ID`? 😉

graphql/src/store/resolver.rs
```diff
@@ -164,62 +163,22 @@ where
...
-                // Convert the ID to a subgraph deployment ID
-                let hash =
-                    SubgraphDeploymentId::new(id.clone()).expect("invalid subgraph ID in database");
-
-                // Return the entity count of the deployment as a BigInt
-                Ok(Value::from(
-                    self.store
-                        .count_entities(hash)
-                        .map_err(QueryExecutionError::StoreError)?,
+        value.map_or(Ok(q::Value::Null), |value| {
```
Jannis: We could drop the entire method now, since the `Resolver` trait has a default implementation that probably does the same. Worth checking.

store/postgres/migrations/2019-05-03-164052_store_entity_count/up.sql
```diff
@@ -0,0 +1,46 @@
...
+  entity_count int8;
+begin
+  -- Initialize the entityCount for all subgraphs
+  update subgraphs.entities e
+  set data = data || '{"entityCount": {"data": "0", "type": "BigInt"}}'
+  where e.entity='SubgraphDeployment';
+
+  -- Handle public.entities
+  with entity_counts as (
+    select subgraph, count(*) as count
```
Jannis: Question: How long do we think this `store_entity_count` function will run for on production?

store/postgres/src/entities.rs
Jannis: This line strikes me as too long. Can we break up the string a little?

store/postgres/src/store.rs
```diff
@@ -607,10 +615,40 @@ impl Store {
...
+                subgraph = operation
+                    .subgraph()
+                    .filter(|s| !s.is_meta())
+                    .map(|s| s.clone());
+            } else {
+                // Verify that we only have one non-meta subgraph in operations
+                if let Some(other) = operation.subgraph() {
+                    if !other.is_meta() && operation.subgraph() != subgraph.as_ref() {
+                        panic!(
+                            "applying entity operations to two non-metadata subgraphs {:?} and {}",
```
Jannis: These could both be `{}` if we do `subgraph.to_string()`, I think.

graph/src/data/subgraph/mod.rs
lutter: Do as I say, not as I do ;) Fixing it

store/postgres/migrations/2019-05-03-164052_store_entity_count/up.sql
lutter: 26.5s :wink:  (I just tried it out on the prod db)

store/postgres/src/entities.rs
lutter: I tried improving it, see if you like that better

store/postgres/src/store.rs
lutter: `subgraph` is an `Option<SubgraphDeploymentId>`, so there's no `to_string()` for it. But this is also something we should never hit, and if we hit it we have a serious bug somewhere; the only thing that users can do with this error message is cut-and-paste it into an issue, there's nothing they can do to address it.

graphql/src/store/resolver.rs
lutter: Yep, you are right - the two were character-for-character identical

lutter:
Updated PR and addressed all the comments so far.

graph/src/data/subgraph/mod.rs
Jannis: I love those cross-review references. 😁

store/postgres/migrations/2019-05-03-164052_store_entity_count/up.sql
Jannis: Ok, great, that's more than acceptable!

store/postgres/src/store.rs
Jannis: Nevermind!

store/postgres/src/entities.rs
Jannis: Not super pretty but I don't have any better ideas. 😁

lutter:
Rebased to latest master

PR #911, opened 2019-05-04
Adaptive log range

### LINKED ISSUE #776: Subgraph gets stuck when eth_getLogs fails because there are too many results

Bug

When catching up with historical data, the block stream asks for `eth_getLogs` on 100,000 blocks at once. It is possible that the eth node (Infura in my case) responds with
```
  Error { code: ServerError(-32005), message: "query returned more than 1000 results", data: None }
```

We blindly retry the request, even though there is no chance retrying it will ever succeed.

This was observed with the Dice2Win subgraph, with the subgraph at block 6290742, while the ethereum head ptr was at 7271924.

The code should realize that it has to reduce the size of the range (e.g., by repeatedly halving it) until this error isn't returned any more. When I manually reduced the range to 100 blocks in the code, the block stream started working again.

As a sidenote, the logs should show more details about what exactly went wrong when `eth_getLogs` fails. Right now it just says something like `Trying again after eth_getLogs RPC call failed (attempt #15)` which does not have enough detail to understand what's happening.

leodasvacas:
Resolves #776. If we start getting this error even when requesting logs for a single block, we should re-evaluate in a new issue.

The first thing this does is have `ETHEREUM_FAST_SCAN_END` affect the block stream as a whole, we now multiply `ETHEREUM_BLOCK_RANGE_SIZE` by 10 if doing a fast scan. The one-size-fits-all nature of these values can be an issue, I also added a special case to go one block at a time if there are unconditional block triggers. We should at some point change the way we control the amount of triggers processed at a time.

The adaptive logic for log ranges first tries requesting the entire given range, if that doesn't work the step is reduced on each failure until we no longer get the 1000 logs error.

And then there are some refactors along the way.

### DIFF
--- a/datasource/ethereum/src/ethereum_adapter.rs
+++ b/datasource/ethereum/src/ethereum_adapter.rs
@@ -18,10 +18,6 @@ pub struct EthereumAdapter<T: web3::Transport> {
 }
 
 lazy_static! {
-    static ref LOG_STREAM_FAST_SCAN_END: u64 = ::std::env::var("ETHEREUM_FAST_SCAN_END")
-        .unwrap_or("4000000".into())
-        .parse::<u64>()
-        .expect("nvalid fast scan end block number");
     static ref TRACE_STREAM_STEP_SIZE: u64 = ::std::env::var("ETHEREUM_TRACE_STREAM_STEP_SIZE")
         .unwrap_or("200".into())
         .parse::<u64>()
@@ -263,16 +259,10 @@ where
 
         let logger = logger.to_owned();
 
-        stream::unfold(from, move |start| {
+        stream::unfold((from, to), move |(start, end)| {
             if start > to {
                 return None;
             }
-            let end = if start < *LOG_STREAM_FAST_SCAN_END {
-                (start + 100_000 - 1).min(to).min(*LOG_STREAM_FAST_SCAN_END)
-            } else {
-                (start + 1_000 - 1).min(to)
-            };
-            let new_start = end + 1;
 
             debug!(logger, "Requesting logs for blocks [{}, {}]", start, end);
 
@@ -284,7 +274,7 @@ where
                             .filter(move |log| log_filter.matches(log))
                             .collect::<Vec<Log>>()
                     })
-                    .map(move |logs| (logs, new_start)),
+                    .map(move |logs| (logs, (end + 1, to))),
             )
         })
         .concat2()
--- a/datasource/ethereum/src/ethereum_adapter.rs
+++ b/datasource/ethereum/src/ethereum_adapter.rs
@@ -114,6 +114,44 @@ where
             })
     }
 
+    fn logs_with_sigs(
+        &self,
+        logger: &Logger,
+        from: u64,
+        to: u64,
+        addresses: Vec<H160>,
+        event_signatures: Vec<H256>,
+        too_many_logs_fingerprint: &'static str,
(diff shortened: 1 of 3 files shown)


leodasvacas:
@yanivtal Yes, it's specific to Infura.

datasource/ethereum/src/ethereum_adapter.rs
```diff
@@ -208,7 +154,41 @@ where
         from: u64,
         to: u64,
         log_filter: EthereumLogFilter,
-    ) -> impl Stream<Item = Vec<Log>, Error = Error> + Send {
+    ) -> impl Future<Item = Vec<Log>, Error = Error> {
+        // Code returned by Infura if a requests returns too many logs.
+        // web3 doesn't seem to offer a better way of checking the error code.
+        const TOO_MANY_LOGS_FINGERPRINT: &str = "ServerError(-32005)";
+        let eth_get_logs = move |eth_adapter: Self,
```
Jannis: I'd prefer if this remained in its own method, taking in the error fingerprint.

datasource/ethereum/src/ethereum_adapter.rs
```diff
@@ -118,56 +114,6 @@ where
...
-                    .logs(log_filter)
-                    .map(move |logs| {
-                        debug!(logger, "Received logs for blocks [{}, {}]", from, to);
-                        logs
-                    })
-                    .from_err::<EthereumContractCallError>()
-                    .from_err()
-            })
-            .map_err(move |e| {
-                e.into_inner().unwrap_or_else(move || {
```
Jannis: It looks like we're losing this error message. Is that deliberate?

datasource/ethereum/src/ethereum_adapter.rs
```diff
@@ -263,31 +243,45 @@ where
...
+                    logs.into_iter()
+                        .filter(move |log| log_filter.matches(log))
+                        .collect::<Vec<Log>>()
+                })
+                .then(move |res| match res {
+                    Err(e) => {
+                        let string_err = e.to_string();
+                        if string_err.contains(TOO_MANY_LOGS_FINGERPRINT) {
+                            let new_step = (step / 10).max(1);
+                            debug!(logger, "Reducing log step"; "new_step" => new_step);
```
Jannis: How about `"Reducing block range size to scan for events"` + `"new_size"`, since that's what the environment variable is called?

datasource/ethereum/src/ethereum_adapter.rs
leodasvacas: Yes, since this is retried indefinitely, this error can never be printed afaict.

leodasvacas:
Review addressed. This will help with slow syncs, but I've seen subgraphs quickly drop to a range of 10 blocks. Bringing parallel log requests back will probably be worth it.

Jannis:
Are we not doing parallel log requests anymore?

Another thought: How soon do we go back to the configured block range size? If we have a block range size of 10,000 and we have to reduce the range size to 5 to get past the first block in this range, does that mean we stick to a range size of 5 for the remaining 9,999 blocks?

leodasvacas:
> Are we not doing parallel log requests anymore?

No, we dropped it in the block triggers PR. I was ok with that since it could complicate making this change which we need quickly, but we should bring it back with adaptive ranges in mind.

Yes, that's right. We can try to increase the range more often, but there is a tradeoff where in one extreme we keep the low range forever and in the other extreme we start from the large range every time.

Issue #912, opened 2019-05-06
Improved log fetching performance


Jannis:
1. Parallelize fetching logs in `EthereumAdapter`
2. Increasing block range size again as soon as possible after getting past a block.

Potentially do the same for traces.

leodasvacas:
Point 1 is done. As I outlined in #914, imo the rest can be considered low priority.

PR #914, opened 2019-05-06
Parallel get logs

leodasvacas:
Part #912, specifically item 1.

When the log range is shortened due to hitting the Infura limit, we start making parallel requests in order to retain performance. The env var `ETHEREUM_PARALLEL_BLOCK_RANGES` limits the amount of parallelism possible, defaulted to 100 (is that a reasonable default?).

Regarding the rest of #912, I think this should be enough to regain the performance of log scanning. At the default values of a block range of 10000 and 100 parallel requests, even when the step drops to 10 we're still simultaneously requesting a total range of 1000, which is what we hard coded before, no performance is lost. So I'd say that trying to proactively re-widen the log range is low-prority, we can leave it as an idea if we still have performance issues after this.

This could be applied to traces if we get a specific response when the limit is hit as we do for logs.

### DIFF
--- a/datasource/ethereum/src/ethereum_adapter.rs
+++ b/datasource/ethereum/src/ethereum_adapter.rs
@@ -261,24 +261,19 @@ where
             // making as many parallel requests of size `step` as necessary,
             // respecting `LOG_STREAM_PARALLEL_CHUNKS`.
             let mut chunk_futures = vec![];
-            let mut offset = start;
+            let mut low = start;
             for _ in 0..*LOG_STREAM_PARALLEL_CHUNKS {
-                if offset == to + 1 {
+                if low == to + 1 {
                     break;
                 }
                 let log_filter = log_filter.clone();
-                let next_offset = (offset + step).min(to) + 1;
-                debug!(
-                    logger,
-                    "Requesting logs for blocks [{}, {}]",
-                    offset,
-                    next_offset - 1
-                );
+                let high = (low + step).min(to);
+                debug!(logger, "Requesting logs for blocks [{}, {}]", low, high);
                 chunk_futures.push(
                     eth.logs_with_sigs(
                         &logger,
-                        offset,
-                        next_offset - 1,
+                        low,
+                        high,
                         addresses.clone(),
                         event_sigs.clone(),
                         TOO_MANY_LOGS_FINGERPRINT,
@@ -289,7 +284,7 @@ where
                             .collect::<Vec<Log>>()
                     }),
                 );
-                offset = next_offset;
+                low = high + 1;
             }
             let logger = logger.clone();
             Some(
@@ -299,7 +294,14 @@ where
                     .then(move |res| match res {
                         Err(e) => {
                             let string_err = e.to_string();
-                            if string_err.contains(TOO_MANY_LOGS_FINGERPRINT) {
+
+                            // If the step is already 0, we're hitting the log
+                            // limit even for a single block. We hope this never
+                            // happens, but if it does, make sure to error.
+                            if string_err.contains(TOO_MANY_LOGS_FINGERPRINT) && step > 0 {
+                                // The range size for a request is `step + 1`.
+                                // So it's ok if the step goes down to 0, in
+                                // that case we'll request one block at a time.
                                 let new_step = step / 10;
                                 debug!(logger, "Reducing block range size to scan for events";
                                                "new_size" => new_step + 1);
@@ -309,7 +311,7 @@ where
                                 Err(err_msg(string_err))
                             }
                         }
-                        Ok(logs) => Ok((logs, (offset + 1, step))),
+                        Ok(logs) => Ok((logs, (low + 1, step))),
                     }),
             )
         })
--- a/datasource/ethereum/src/ethereum_adapter.rs
+++ b/datasource/ethereum/src/ethereum_adapter.rs
@@ -311,7 +311,7 @@ where
                                 Err(err_msg(string_err))
                             }
                         }
-                        Ok(logs) => Ok((logs, (low + 1, step))),
+                        Ok(logs) => Ok((logs, (low, step))),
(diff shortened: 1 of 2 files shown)


datasource/ethereum/src/ethereum_adapter.rs
```diff
@@ -255,39 +256,61 @@ where
...
                 return None;
             }
-            let end = start + step;
-            debug!(logger, "Requesting logs for blocks [{}, {}]", start, end);
+
+            // This will iterate more than once iff `step` has been reduced,
+            // making as many parallel requests of size `step` as necessary,
+            // respecting `LOG_STREAM_PARALLEL_CHUNKS`.
+            let mut chunk_futures = vec![];
+            let mut offset = start;
```
olibearo: `lower` and `upper` would be clearer here instead of `offset` and `next_offset` since they aren't really offsets from `start`

datasource/ethereum/src/ethereum_adapter.rs
```diff
@@ -255,39 +256,61 @@ where
...
+            for _ in 0..*LOG_STREAM_PARALLEL_CHUNKS {
```
olibearo: Curious, is there a good way to accumulate the `chunk_futures` with an iterator over the `0..*LOG_STREAM_PARALLEL_CHUNKS` range?

datasource/ethereum/src/ethereum_adapter.rs
```diff
@@ -255,39 +256,61 @@ where
...
-                            let new_step = step / 10;
-                            debug!(logger, "Reducing block range size to scan for events";
+                stream::futures_ordered(chunk_futures)
+                    .collect()
+                    .map(|chunks| chunks.into_iter().flatten().collect::<Vec<Log>>())
+                    .then(move |res| match res {
+                        Err(e) => {
+                            let string_err = e.to_string();
+                            if string_err.contains(TOO_MANY_LOGS_FINGERPRINT) {
+                                let new_step = step / 10;
```
olibearo: Should we handle the lower bound here differently?

What if the step is already `1`? I think step is going to turn into `0` and then the stream `unfold` will never end.
Maybe we should just log and error in that case.

datasource/ethereum/src/ethereum_adapter.rs
```diff
@@ -255,39 +256,61 @@ where
...
-            let end = start + step;
-            debug!(logger, "Requesting logs for blocks [{}, {}]", start, end);
+
+            // This will iterate more than once iff `step` has been reduced,
+            // making as many parallel requests of size `step` as necessary,
+            // respecting `LOG_STREAM_PARALLEL_CHUNKS`.
+            let mut chunk_futures = vec![];
+            let mut offset = start;
+            for _ in 0..*LOG_STREAM_PARALLEL_CHUNKS {
+                if offset == to + 1 {
```
olibearo: Changing this to `> to` would be simpler

datasource/ethereum/src/ethereum_adapter.rs
```diff
@@ -255,39 +256,61 @@ where
...
+                    break;
+                }
+                let log_filter = log_filter.clone();
+                let next_offset = (offset + step).min(to) + 1;
```
olibearo: Could get rid of this `+1` and the `-1` below if you just adding the `1` when you set the `offset` below.

datasource/ethereum/src/ethereum_adapter.rs
leodasvacas: true, this isn't an offset at all.

datasource/ethereum/src/ethereum_adapter.rs
leodasvacas: I can't think of a very good way of turning this loop into combinators, but maybe there is one.

datasource/ethereum/src/ethereum_adapter.rs
leodasvacas: `0` is fine, the range is inclusive and we always increase the start so it always advances by at least one block. This is worth a comment.

datasource/ethereum/src/ethereum_adapter.rs
leodasvacas: It would behave the same, but I like how this is more precise.

datasource/ethereum/src/ethereum_adapter.rs
leodasvacas: That will be much better, thanks!

datasource/ethereum/src/ethereum_adapter.rs
olibearo: Yeah, if we are coming from a step of `0` or `1` though we already know there are too many logs in that block for Infura to return a response. Would we just get stuck in a loop with a step size of `0` and the `TOO_MANY_LOGS_ERROR` on every fold? 

Might be worth logging (and erroring?) that the smallest step size is still too big.

leodasvacas: If we're coming from a step of `0` this will loop forever. I'll add a check for this.

leodasvacas:
@olibearo I've adressed all your comments, and fixed a bug where we'd skip a block.

datasource/ethereum/src/ethereum_adapter.rs
Jannis: https://docs.rs/itertools/0.3.0/itertools/fn.times.html

Jannis: Basically,
```rust
let chunk_futures = itertools::times(*LOG_STREAM_PARALLEL_CHUNKS)
  .map(|_| { ... })
  .collect();

stream::futures_ordered(chunk_futures)
  ...
```

Jannis:
I'm happy with the changes if @olibearo is. 😄

PR #915, opened 2019-05-06
Add `LogCode` enum for logs

olibearo:
Adds codes to log messages we would like to run analytics on for [health checks](https://github.com/graphprotocol/hosted-service/issues/25)


graph/src/log/mod.rs
```diff
@@ -0,0 +1,31 @@
+use std::fmt::{Display, Error, Formatter};
+
+pub enum LogCode {
```
Jannis: Shall we move this to `graph/src/util/log.rs` instead of creating another `log` module?

datasource/ethereum/src/block_ingestor.rs
```diff
@@ -97,17 +97,27 @@ where
                                 let latest_number = latest_block.number.unwrap().as_u64() as i64;
                                 let head_number = head_block_ptr.number as i64;
                                 let distance = latest_number - head_number;
+                                if distance > 0 {
+                                    info!(
+                                        self.logger,
```
Jannis: Can we combine this with the two messages below? It seems unnecessary to log two messages each time. And let's use more human-readable names for the different values.

`local_block_ptr` -> `current_head_block`
`network_block_ptr` -> `latest_head_block`
`distance` -> `blocks_behind`?

graph/src/log/mod.rs
Jannis: We can still export `LogCode` via `graph::prelude::LogCode` to provide easier access.

graph/src/log/mod.rs
```diff
@@ -0,0 +1,31 @@
...
+    SubgraphExecutionFailure,
+    SubgraphExecutionFailureNotRecorded,
+    BlockSyncStatus,
+}
+
+impl Display for LogCode {
+    fn fmt(&self, f: &mut Formatter) -> Result<(), Error> {
+        let value = match self {
+            LogCode::SubgraphStartFailure => "SubgraphStartFailure",
+            LogCode::SubgraphExecutionFailure => "SubgraphExecutionFailure",
```
Jannis: I'm not a fan of introducing the term `execution` here. We already use `indexing`, `syncing` and `processing` in different places. Let's make `SubgraphSyncing...`.

graph/src/log/mod.rs
```diff
@@ -0,0 +1,31 @@
...
+            LogCode::SubgraphExecutionFailureNotRecorded => "SubgraphExecutionFailureNotRecorded",
+            LogCode::BlockSyncStatus => "BlockSyncStatus",
```
Jannis: `BlockIngestionStatus` to stick to our existing vocabulary?

olibearo:
@Jannis feedback addressed

graph/src/util/log.rs
```diff
@@ -103,3 +104,37 @@ pub fn register_panic_hook(panic_logger: Logger, shutdown_sender: oneshot::Sende
...
+}
+
+impl Display for LogCode {
+    fn fmt(&self, f: &mut Formatter) -> Result<(), Error> {
+        let value = match self {
+            LogCode::SubgraphStartFailure => "SubgraphStartFailure",
+            LogCode::SubgraphSyncingFailure => "SubgraphSyncingFailure",
+            LogCode::SubgraphSyncingFailureNotRecorded => "SubgraphSyncingFailureNotRecorded",
+            LogCode::BlockIngestionStatus => "BlockIngestionStatus",
+            LogCode::GraphQlQuerySuccess => "GraphQLQueryTimer",
```
Jannis: These two are not used yet; is this planned?

olibearo: They are now being used !

server/http/src/service.rs
```diff
@@ -219,7 +225,25 @@ where
                         .run_query(query)
                         .map_err(|e| GraphQLServerError::from(e))
                 })
-                .then(|result| GraphQLResponse::new(result)),
+                .then(move |result| {
+                    let elapsed = start.elapsed().as_millis();
+                    match result {
+                        Ok(_) => info!(
+                            logger,
+                            "GraphQL query served.";
```
Jannis: Can we remove the `.` at the end of the messages?

Also: the log codes are not being used in these two messages.

Jannis: Another thing: I believe we're already logging query times in the `graph-graphql` crate. Those messages should probably be removed then.

server/http/src/service.rs
```diff
@@ -219,7 +225,25 @@ where
...
+                            "subgraph_deployment_id" => sd_id.deref(),
```
Jannis: `subgraph_deployment` is good enough.

server/http/src/service.rs
```diff
@@ -219,7 +225,25 @@ where
...
+                            "query_duration_ms" => elapsed
```
Jannis: `time_ms`?

server/http/src/service.rs
```diff
@@ -219,7 +225,25 @@ where
...
+                        ),
+                        Err(ref e) => error!(
+                            logger,
+                            "GraphQL query failed.";
+                            "subgraph_deployment_id" => sd_id.deref(),
+                            "error" => e.to_string(),
+                            "query_duration_ms" => elapsed
```
Jannis: `time_ms`?

server/http/src/service.rs
```diff
@@ -219,7 +225,25 @@ where
...
+                        Ok(_) => info!(
+                            logger,
+                            "GraphQL query served.";
+                            "subgraph_deployment_id" => sd_id.deref(),
+                            "query_duration_ms" => elapsed
+                        ),
+                        Err(ref e) => error!(
+                            logger,
+                            "GraphQL query failed.";
+                            "subgraph_deployment_id" => sd_id.deref(),
```
Jannis: `subgraph_deployment` is good enough.

server/http/src/service.rs
```diff
@@ -219,7 +225,25 @@ where
...
+                    match result {
+                        Ok(_) => info!(
+                            logger,
+                            "GraphQL query served.";
+                            "subgraph_deployment_id" => sd_id.deref(),
+                            "query_duration_ms" => elapsed
+                        ),
+                        Err(ref e) => error!(
+                            logger,
+                            "GraphQL query failed.";
```
Jannis: No `.` at the end please 😄

Issue #856, opened 2019-04-04
Faster mapping execution


Jannis:

leodasvacas:
I assume this issue is about improving indexing time. First we shoud try to answer questions such as:
- What subgraphs are representative of the use pattern we want to speed up? How complex are their mappings? Do they call contracts or IPFS? How many events are there per block? How many entities are there per block?
- What speedup are we targeting?
- Are we considering a warm database with blocks loaded (probably from indexing a previous version of the subgraph) or cold?
- Do we want to speed up indexing from genesis or indexing the newest block in a subgraph that's already synced?

#### Case studies

I went ahead and did three case studies: Compound, ENS and Uniswap. One case I didn't cover is IPFS, which none of these use.

Compound makes 3-4 contract calls per event. The time to make a contract call is dependent on the latency to the eth node, so I can't measure on my machine, we'd need production timings, but I'd suspect that the contract calls dominate the indexing time.

ENS does no contract or ipfs calls and has very simple mappings. From a cold db, fetching blocks from the eth node dominates the indexing time. From a warm DB, it gets more subtle. The three major steps to process a block (and my local measurments for ENS) are:
1. Get block from the block stream 3-10ms
2. Process events in the runtime ~3ms*n of events in block
3. Apply operations to the DB 10-20ms, proportional to n of operations
4. Repeat.

Steps 1 and 3 take a while because they involve the DB, while step 2 is the interaction with wasm.

Uniswap like ENS doesn't do contract or ipfs calls. However it has more complex mappings that take ~15ms per event, making the runtime the slowest step in many blocks.

#### Possible solutions

From a cold db I don't have a better idea other than making sure there's good networking to the eth node, and then there are also bigger ideas such as local contract execution. From a warm DB there are a few things:

For Compound, caching contract calls might help.

For ENS step 3 is significant, applying the operations to the DB seems to be a bottleneck if the mapping is simple and each block has only a few events. We could pipeline operations to the DB so that we could start processing the next block while the previous one is being applied. This would require architectural changes, currently the block stream cannot proceed while the previous block has not been fully processed. Could be an up to 2x speed up for some blocks.

[...]

PR #917, opened 2019-05-08
Monitor contention in tokio threadpool

leoyvens:
A simple mechanism that periodically checks the threadpool responsiveness. Part of the investigation for #856.

### DIFF
--- a/node/src/main.rs
+++ b/node/src/main.rs
@@ -18,13 +18,12 @@ extern crate lazy_static;
 extern crate url;
 
 use clap::{App, Arg};
-use futures::sync::oneshot;
+use futures::sync::{mpsc, oneshot};
 use git_testament::{git_testament, render_testament};
 use ipfs_api::IpfsClient;
 use lazy_static::lazy_static;
 use std::env;
 use std::str::FromStr;
-use std::sync::Arc;
 use std::time::Duration;
 
 use graph::components::forward;
@@ -64,7 +63,6 @@ lazy_static! {
 git_testament!(TESTAMENT);
 
 fn main() {
-    println!("REORG_THRESHOLD = {:#?}", *REORG_THRESHOLD);
     let (shutdown_sender, shutdown_receiver) = oneshot::channel();
     // Register guarded panic logger which ensures logs flush on shutdown
     let (panic_logger, _panic_guard) = guarded_logger();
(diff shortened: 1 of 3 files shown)


node/src/main.rs
```diff
@@ -64,7 +63,6 @@ lazy_static! {
 git_testament!(TESTAMENT);
 
 fn main() {
-    println!("REORG_THRESHOLD = {:#?}", *REORG_THRESHOLD);
```
olibearo: 👍

node/src/main.rs
```diff
@@ -574,6 +572,29 @@ fn async_main() -> impl Future<Item = (), Error = ()> + Send + 'static {
...
+    // task that simply responds to "ping" requests. Then spawn a separate
+    // thread to periodically ping it and check responsiveness.
+    let (ping_send, ping_receive) = mpsc::channel::<std::sync::mpsc::Sender<()>>(1);
+    tokio::spawn(
+        ping_receive
+            .for_each(move |pong_send| pong_send.clone().send(()).map(|_| ()).map_err(|_| ())),
+    );
+    let contention_logger = logger.clone();
+    std::thread::spawn(move || loop {
+        std::thread::sleep(Duration::from_secs(10));
```
olibearo: My gut is telling me that 10 seconds is too sparse, contention could come and go during that window. 

This seems cheap enough to run every 10ms - 100ms. Is that the case?

node/src/main.rs
```diff
@@ -574,6 +572,29 @@ fn async_main() -> impl Future<Item = (), Error = ()> + Send + 'static {
...
+        let (pong_send, pong_receive) = std::sync::mpsc::channel();
+        ping_send.clone().send(pong_send).wait().unwrap();
+        let mut timeout = Duration::from_millis(1);
+        while pong_receive.recv_timeout(timeout).is_err() {
+            warn!(contention_logger, "Possible contention in tokio threadpool";
```
olibearo: If add a `LogCode` here, that would help in making a Grafana panel + alert for this

node/src/main.rs
leodasvacas: Agreed, I'll lower to 100ms.

node/src/main.rs
leodasvacas: I'll add a code, but we should probably alert on any `warn` level log.

ghost (approved):

PR #919, opened 2019-05-08
Add Travis pass to deny warnings

leoyvens:
And fix the warnings found.


lutter (approved):
Awesome! Thanks for fixing these issues and making sure Travis will keep them from happening again.

PR #920, opened 2019-05-09
Use split subgraph table when testing the store

Jannis:
Ideally, we'd test both the `public` entities table and a subgraph-specific entities table. However, with the tests being defined as `FnOnce` functions, we can't easily clone them and run them twice.

So it's not 100% clear which variant is best to test. This PR switches from `public` to subgraph-specific. @lutter What are your thoughts?


lutter:
This will come as part of migrating `public::entities` into split tables; when I did it, it required changes to a few more tests - the way we'll do per-subgraph migration also requires that subgraphs get started explicitly so the db can clean up intermediate state. So I would just hold off on that until my migration PR is ready.

lutter:
We can close this now; the tests now use a split subgraph, and `public.entities` is not in use anymore since PR #950 was merged.

Jannis:
Great!

PR #925, opened 2019-05-10
Add `ens.nameByHash` host export and supporting database table

lutter:
This PR sets things up so that we can have a function `ens.nameByHash` that maps keccak256 hashes to their preimage, based on a rainbow table in the database. The PR does not fill the rainbow table, as that will be huge.


lutter:
As I was thinking about this, I realized that we could save a lot of storage space for the hashes by storing them as numbers, rather than strings. Postgres doesn't have a native 256 bit integer type, but if I did my math right, we could store it as a `numeric(10,0)` (10-digit decimal integer)

The hashes (if we strip the `0x` prefix) take 68 bytes each to store, whereas `numeric(10,0)` requires 10 bytes to store. That means for 130M hashes, we'd need ~ 8GB to store the hashes as strings, and ~ 1.2GB to store them as numbers, and since the hashes need to be stored in both the table and the primary key index, we'd actually reap the storage savings twice, at the cost of slightly more complicated access and import (need to convert input hex strings to decimal numbers)

lutter:
> As I was thinking about this, I realized that we could save a lot of storage space for the hashes by storing them as numbers, rather than strings. Postgres doesn't have a native 256 bit integer type, but if I did my math right, we could store it as a `numeric(10,0)` (10-digit decimal integer)

_sigh_ I shouldn't do higher math. A 256 bit integer takes of course 80 some decimal digits.

graph/src/components/store.rs
```diff
@@ -660,6 +660,14 @@ pub trait Store: Send + Sync + 'static {
     /// Queries the store for a single entity matching the store query.
     fn find_one(&self, query: EntityQuery) -> Result<Option<Entity>, QueryExecutionError>;
 
+    /// Find the reverse of keccak256 for `hash` through looking it up in the
+    /// rainbow table.
+    fn find_ens_name(&self, _hash: &str) -> Result<Option<String>, QueryExecutionError> {
+        // The default implementation knows nothing about mapping keccak256 hashes
+        // to their preimage
+        Ok(None)
+    }
```
leodasvacas: I'm not a fan of having dummy default implementations, is this necessary?

runtime/wasm/src/module/test.rs
```diff
@@ -246,13 +246,14 @@ where
     }
 
     fn takes_ptr_returns_ptr<P, Q>(&mut self, fn_name: &str, arg: AscPtr<P>) -> AscPtr<Q> {
-        self.module
+        dbg!(self
```
leodasvacas: Remove the `dbg!`.

store/postgres/src/store.rs
```diff
@@ -839,6 +839,27 @@ impl StoreTrait for Store {
...
+    fn find_ens_name(&self, hash: &str) -> Result<Option<String>, QueryExecutionError> {
+        use crate::db_schema::ens_names as dsl;
+
+        let conn = self
+            .get_conn()
+            .map_err(|e| QueryExecutionError::StoreError(e.into()))?;
+
+        dsl::table
+            .select(dsl::name)
+            .filter(dsl::hash.eq(hash))
```
leodasvacas: Since hash is the primary key, I believe we can do `.find(hash)`.

graph/src/components/store.rs
lutter: Neither am I, really. Fixing it.

lutter:

PR #929, opened 2019-05-11
Don't block updates on a full channel

leoyvens:
Previously a syncing subgraph with a full buffer of chain head updates could prevent other subgraph from getting updates. I have not reproduced and tested the fix, but the hypothesis seems solid.

### DIFF
--- a/store/postgres/src/chain_head_listener.rs
+++ b/store/postgres/src/chain_head_listener.rs
@@ -89,7 +89,7 @@ impl ChainHeadUpdateListener {
                         let logger = logger.clone();
                         let subscribers = subscribers.clone();
 
-                        sender.send(update.clone()).then(move |result| {
+                        tokio::spawn(sender.send(update.clone()).then(move |result| {
                             if result.is_err() {
                                 // If sending to a subscriber fails, we'll assume that
                                 // the receiving end has been dropped. In this case we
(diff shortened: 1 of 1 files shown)


store/postgres/src/chain_head_listener.rs
```diff
@@ -89,7 +89,7 @@ impl ChainHeadUpdateListener {
                         let logger = logger.clone();
                         let subscribers = subscribers.clone();
 
-                        sender.send(update.clone()).then(move |result| {
+                        tokio::spawn(sender.send(update.clone()).then(move |result| {
```
Jannis: Shall we add the PR messages here as a comment so we know why we're spawning? We may need this again in other places :)

Jannis (changes requested):

store/postgres/src/chain_head_listener.rs
Jannis: An alternative approach could be `futures::unordered` maybe? But even that could hold progress back if one of the futures blocks. So no, this is probably better.

Jannis: One concern I would have is that we spawn more and more tasks here over time, and still eat up more and more memory until it's too much.

Jannis (approved):

PR #933, opened 2019-05-13
Include arguments to log in ElasticSearch log as a map

olibearo:
Log arguments are sent to ElasticSearch as a comma separated list appended to the log message. To create panels in Grafana from the log arguments they need to be accessible as a property in the log's JSON record.


Jannis:
Nice! I've been wanting this for so long, just didn't know how to do it.

PR #935, opened 2019-05-13
Fix source of log argument kvs for Elasticsearch

olibearo:
The kvs are in the `Record` not the `OwnedKVList`

### DIFF
--- a/graph/src/log/elastic.rs
+++ b/graph/src/log/elastic.rs
@@ -296,15 +296,14 @@ impl Drain for ElasticDrain {
         // Serialize log message arguments
         let mut serializer = SimpleKVSerializer::new();
         values
-            .clone()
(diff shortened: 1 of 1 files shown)


graph/src/log/elastic.rs
```diff
@@ -296,13 +296,15 @@ impl Drain for ElasticDrain {
...
         values
+            .clone()
             .serialize(record, &mut serializer)
             .expect("failed to serialize log message arguments");
         let (n_value_kvs, value_kvs) = serializer.finish();
 
         // Serialize log message arguments into hash map
         let mut serializer = HashMapKVSerializer::new();
         values
+            .clone()
```
Jannis: I think this one can be dropped.

Issue #599, opened 2018-11-23
Add logging ability in mappings


tsuberim:
**Do you want to request a *feature* or report a *bug*?**
feature
**What is the current behavior?**
No ability to print or log stuff inside a mapping.

**What is the expected behavior?**
Something like `console.log(msg)` that users can use to log strings or objects out and view them in `graph-node`'s output.

Jannis:
This has been resolved with the `log` API.

PR #943, opened 2019-05-14
Add a `log.log` host export for basic logging in mappings

Jannis:
Part of resolving #599 and https://github.com/graphprotocol/graph-ts/issues/5.

### DIFF
--- a/runtime/wasm/src/asc_abi/class.rs
+++ b/runtime/wasm/src/asc_abi/class.rs
@@ -1,6 +1,7 @@
 use super::{AscHeap, AscPtr, AscType, AscValue};
 use ethabi;
 use graph::data::store;
+use graph::prelude::slog;
 use graph::serde_json;
 use graph_runtime_derive::AscType;
 use std::marker::PhantomData;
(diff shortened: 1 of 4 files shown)


Jannis:
I'll add a test for `log.log` tomorrow.

Jannis:
I've come to the conclusion that there's no simple/good way to capture and check the `log.log`, so I'd rather not add a test for it.

runtime/wasm/src/asc_abi/class.rs
```diff
@@ -536,3 +537,24 @@ pub(crate) struct AscBigDecimal {
     // Decimal exponent. This is the opposite of `scale` in rust BigDecimal.
     pub exp: AscPtr<AscBigInt>,
 }
+
+#[repr(C)]
```
leodasvacas: This should be `#[repr(u32)]`.

leoyvens (changes requested):
Yay logging! Just a comment.

runtime/wasm/src/asc_abi/class.rs
Jannis: @leodasvacas Updated to `#[repr(u32)]`!

leoyvens (approved):

Issue #916, opened 2019-05-07
Missing attribute indexes for foreign keys


lutter:
For ENS, the GraphQL model of `Domain` has a property `parent` which itself is a `Domain`, i.e., `parent` is a foreign key for a `Domain`. We do not have an attribute index on it, and in some cursory testing, I had some trouble creating an index that Postgres will actually use to look up along the `parent` relationship (I had to create it on `(entity, data->'parent'->>'data')`).

We need such an index to back queries like `domains { id subdomain { id } }`. Without the index, Postgres performs an index scan on the primary key index of the `entities` table, which is ok for smaller tables, but becomes problematic as tables grow in size.

I would argue that fk indexes are the only attribute indexes we need, though that's somewhat outside the realm of this issue.

Issue #927, opened 2019-05-10
Support for inline objects and lists of inline objects


Jannis:
# Motivation / Rationale

We currently don't support inline objects: GraphQL types that are not entities and that be embedded in parent objects instead of referencing them via their IDs.

This feature is necessary to better support Ethereum tuples and arrays of Ethereum tuples in the GraphQL schema and particularly in `graph init --from-contract` scaffolding.

Imagine a contract has an event like the following

```solidity
struct Receiver {
  address account,
  string name,
}

struct Details {
  uint256 amount,
  address token,
}
  
event MultiTransfer(
  address from,
  Receiver[] receivers,
  Details details,
)
```

We currently can't represent `Receiver[]` in the GraphQL schema and we represent `Details` by unrolling the fields into a flat list of fields:

```graphql
type MultiTransfer @entity {
  id: ID!
  from: Bytes!
  // Receivers[] cannot be represented
  details_amount: BigInt!
  details_token: Bytes!
}
```

What we really want is this:

```graphql
type Receiver {
  id: ID!
  account: Bytes!
  name: String!
}

type Details {
  amount: BigInt!
  token: Bytes!
}
  
type MultiTransfer @entity {
  from: Bytes!
  receivers: [Receiver!]!
  details: Details!
}
```

leodasvacas:
@Jannis Have we considered the possibility of having the tuple type be an entity, and generating the scaffolding code to handle it as such?

Inline objects are a possibility, but it introduces a special case in query execution, in the runtime and in the general mental model. This sounds like something we'll want eventually, possibly due to different performance characteristics, but for now having those be entities could be the simplest way to support tuple arrays.

PR #944, opened 2019-05-14
graph/data: Ensure we index foreign keys in the GraphQL schema

lutter:
We only created indexes for fields whose type was a scalar, which left out
fields that reference other entities. This PR makes sure we also create
indexes on such fields.

Fixes https://github.com/graphprotocol/graph-node/issues/916

### DIFF
--- a/graph/src/data/subgraph/schema.rs
+++ b/graph/src/data/subgraph/schema.rs
@@ -13,7 +13,7 @@
 
 use failure::Error;
 use graphql_parser::query as q;
-use graphql_parser::schema::{Definition, Document, Type, TypeDefinition};
+use graphql_parser::schema::{Definition, Document, Field, Name, Type, TypeDefinition};
 use hex;
 use rand::rngs::OsRng;
 use rand::Rng;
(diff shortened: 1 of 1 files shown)


graph/src/data/subgraph/schema.rs
```diff
@@ -1181,11 +1190,35 @@ pub fn attribute_index_definitions(
...
+fn is_derived_field(field: &Field) -> bool {
+    field
+        .directives
+        .iter()
+        .any(|dir| dir.name == Name::from("derivedFrom"))
+}
+
+fn is_entity(type_name: &str, definitions: &[Definition]) -> bool {
+    definitions.iter().any(|defn| {
+        if let Definition::TypeDefinition(TypeDefinition::Object(schema_object)) = defn {
```
Jannis: As soon as we add #927, this check will be wrong. The following check is more robust and would be good to adopt here (even if it's copied in):

https://github.com/graphprotocol/graph-node/blob/master/graphql/src/schema/ast.rs#L377-L394

lutter: Good point. Just fixed that. Longer term, we should think about how to restructure things so we do not have to have this logic in two places, since it's easy right now to make a change in only one and forget the other.

lutter:
Rebased to latest master and addressed review comment.

PR #945, opened 2019-05-14
store/postgres: Add an index on event_meta_data(source)

lutter:
We find event_meta_data during block reversions by the source attribute,
and not having an index on it slows down those queries, and therefore block
reversion, dramatically. In one case, query time for a block reversion goes
from 18s to 5ms after adding the index.


lutter:
Building this index in production will take about 3 minutes.

lutter:
I actually just added the index manually to production so that rolling this out won't slow down startup

Issue #937, opened 2019-05-13
Reduce thread contention in subgraph indexing


Jannis:

leodasvacas:
Plan as outlined in https://github.com/graphprotocol/graph-node/issues/856#issuecomment-491342191:

- [ ] Set the tokio threadpool size to 100.
- [ ] Solve #905 by moving all store operations into a separate, blocking threadpool, and have them return futures. This is not hard but it's laborious and will have a big diff.
- [ ] Try going back to the default threadpool size.

leodasvacas:
Since increasing the size I still see 1ms and 10ms contention warnings, but no 100ms which is an improvement, contention seems to have shifted to the DB connection pool which was expected. Still the situation in tokio is not great, probably due to the issue described in https://github.com/tokio-rs/tokio/issues/426#issuecomment-493710173.

Jannis:
What's the latest here? Didn't we add the blocking threadpool recently?

leodasvacas:
I believe we've sufficiently mitigated this issue that it shouldn't have any practical effect, but the solution is not ideal, so I'd like to keep this open. My explanation for the thread pool contention warnings that can still be observed sometimes is that half of them are due do [this bug](https://github.com/crossbeam-rs/crossbeam/pull/384) (fix isn't released yet) and the other half is simply that there is contention for the CPUs, I haven't confirmed this yet, but it would be an ops concern, so I'm removing the label from this issue.

PR #946, opened 2019-05-14
Update tokio, simplify panic handling, increase threadpool size

leoyvens:
Part of #937.


node/src/main.rs
```diff
@@ -63,38 +63,45 @@ lazy_static! {
...
+                            std::process::exit(1);
+                        });
+                        runtime
+                            .shutdown_now()
+                            .wait()
+                            .expect("Failed to shutdown Tokio Runtime");
+                        println!("Runtime cleaned up and shutdown successfully");
+                    }
+                });
+            })
```
lutter: It looks like the old panic handler tried to log details about the panic, and the new one doesn't. Shouldn't that information still be logged?

leodasvacas: Good point. This removes the custom panic handler and uses the default panic handler, which will print the panic information to stdout. I'm assuming this is good enough, and that we don't need to have the panic information go through `slog`.

lutter: Makes sense to me.

lutter (approved):

PR #947, opened 2019-05-15
Handle block stream errors gracefully in SubgraphInstanceManager

olibearo:
The blockstream should faithfully attempt to produce blocks for a subgraph.

An error in the blockstream is non fatal and temporary, so the `SubgraphInstanceManager` should not fail the subgraph when these errors occur.

### DIFF
--- a/core/src/subgraph/instance_manager.rs
+++ b/core/src/subgraph/instance_manager.rs
@@ -298,6 +298,7 @@ where
     let store_for_err = ctx.inputs.store.clone();
     let logger_for_err = logger.clone();
     let logger_for_restart_check = logger.clone();
+    let logger_for_block_stream_errors = logger.clone();
 
     let block_stream_canceler = CancelGuard::new();
     let block_stream_cancel_handle = block_stream_canceler.handle();
@@ -347,6 +348,19 @@ where
                 Ok(true)
             }
         })
+        // Log and drop the errors from the block_stream
+        // The block stream will continue attempting to produce blocks
+        .then(move |result| match result {
+            Ok(block) => Ok(Some(block)),
+            Err(e) => {
+                error!(
+                    logger_for_block_stream_errors,
+                    "Block stream produced a non fatal error = {}", e,
+                );
+                Ok(None)
+            }
+        })
+        .filter_map(|block_opt| block_opt)
         // Process blocks from the stream as long as no restart is needed
         .fold(ctx, move |ctx, block| {
             let needs_restart_set = needs_restart_set.clone();
--- a/core/src/subgraph/instance_manager.rs
+++ b/core/src/subgraph/instance_manager.rs
@@ -353,9 +353,10 @@ where
         .then(move |result| match result {
             Ok(block) => Ok(Some(block)),
             Err(e) => {
-                error!(
+                debug!(
                     logger_for_block_stream_errors,
-                    "Block stream produced a non fatal error = {}", e,
+                    "Block stream produced a non-fatal error";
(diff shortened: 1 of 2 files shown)


core/src/subgraph/instance_manager.rs
```diff
@@ -347,6 +348,19 @@ where
...
             }
         })
+        // Log and drop the errors from the block_stream
+        // The block stream will continue attempting to produce blocks
+        .then(move |result| match result {
+            Ok(block) => Ok(Some(block)),
+            Err(e) => {
+                error!(
+                    logger_for_block_stream_errors,
+                    "Block stream produced a non fatal error = {}", e,
```
Jannis: How about this?
```rust
debug!(
  logger_for_stream_errors,
  "Block stream produced a non-fatal error";
  "error" => format!("{}", e),
);
```

leodasvacas: The "non-fatal" wording sounds a bit dismissive, there could be errors that the subgraph can't recover from.

leodasvacas:
This PR makes `consecutive_err_count` dead, right? So it can be removed.

core/src/subgraph/instance_manager.rs
olibearo: Which errors are you referring to?

olibearo:
@leodasvacas I decided to add a pause into the `BlockStream` after errors which leverages the the `consecutive_err_count`, will push in a moment

core/src/subgraph/instance_manager.rs
leodasvacas: Don't know what specific errors could happen here, but "non-fatal" sounds like it will go away by itself and is not interrupting progress, when it could be something like the DB or the network that got disconnected and will require action to fix.

leodasvacas: Now I get that this is meant to signal that there's nothing wrong with the subgraph itself, which makes sense.

datasource/ethereum/src/block_stream.rs
```diff
@@ -1090,28 +1099,21 @@ where
...
-                            state = BlockStreamState::Reconciliation(next_blocks_future);
-
-                            // Poll the next_blocks() future
-                            continue;
+                            // Pause before trying again
+                            let secs = (5 * self.consecutive_err_count).min(120) as u64;
+                            let instant = Instant::now() + Duration::from_secs(secs);
+                            state = BlockStreamState::Paused(Box::new(
+                                Delay::new(instant).map_err(|err| {
+                                    format_err!("Paused state's delay future failed = {}", err)
```
Jannis: Our convention is for errors to either be combined with `...: {}`, not `... = {}`, or when logged, to be added using an `error` key/value pair in the log record.

datasource/ethereum/src/block_stream.rs
```diff
@@ -56,6 +59,12 @@ enum BlockStreamState {
     /// Valid next states: Reconciliation
     YieldingBlocks(Box<Stream<Item = EthereumBlockWithTriggers, Error = Error> + Send>),
 
+    /// The BlockStream experienced an error and is pausing before attempting to produce
+    /// blocks again.
+    ///
+    /// Valid next states: Reconciliation
+    Paused(Box<Future<Item = (), Error = Error> + Send>),
```
Jannis: Paused sounds like it could also be paused for a long time, like being put on halt. Maybe `RetryAfterDelay`?

Jannis:
@olibearo Looks good now! Let's see if the tests pass before we merge.

PR #950, opened 2019-05-16
Framework for per-subgraph db migrations and migrate existing subgraphs into their own entities tables

lutter:
This PR puts a framework in place to perform store migrations for subgraphs one at a time, and uses that to migrate existing subgraphs out of the `public.entities` table into dedicated tables for each subgraph.

I tested this by running it against a dump of the production database, and assigning the subgraphs for one of the mainnet indexing nodes to my graph-node instance. When a subgraph is picked for migration, indexing for that subgraph pauses and its data is successfully transferred into the new tables.

It is still possible for indexing of non-migrating subgraphs to stall; in my testing, these stalls didn't last very long, and block ingestion never fell farther than about 15 blocks behind.

Eventually, all assigned and non-failed subgraphs will be migrated. Subgraphs that are not assigned to a node, and subgraphs that fail during their processing will not be migrated by this mechanism.

### DIFF
--- a/store/postgres/migrations/2019-05-15-215022_migrate_entities/up.sql
+++ b/store/postgres/migrations/2019-05-15-215022_migrate_entities/up.sql
@@ -1,0 +1,186 @@
+-- The different storage schemes/versions we support for subgraphs.
+-- Over time, we will add more as our storage scheme for individual
+-- subgraphs evolves
+create type deployment_schema_version as enum ('public', 'split');
+-- How far a currently running migration has gotten. See
+-- store/postgres/src/entities.rs for more details
+create type deployment_schema_state as enum ('ready', 'tables');
+
+-- Modify deployment_schemas to track the progress of any needed migrations
+alter table deployment_schemas
+  add column version   deployment_schema_version,
+  add column migrating bool not null default false,
+  add column state     deployment_schema_state not null default 'ready';
+
+-- Existing subgraphs (subgraph of subgraphs and subgraphs that have been
+-- deployed since commit d2248402) are stored in their own split-out
+-- database schema
+update deployment_schemas
+   set version='split';
+
+alter table deployment_schemas
+  alter column version set not null;
+
+-- Legacy subgraphs (those deployed before commit d2248402) are stored in
+-- the big public.entities table. With the migration machinery we are
+-- putting in place, we need an entry in deployment_schemas for every
+-- deployed subgraph
+insert into deployment_schemas(subgraph, version)
+select id, 'public'::deployment_schema_version
+  from subgraphs.entities e
+ where entity='SubgraphDeployment'
+   and not exists (select 1 from deployment_schemas d
+                           where d.subgraph=e.id);
+
+-- Migrate the entities for one subgraph into their own schema. The
+-- deployment_schemas table must already contain an entry for that
+-- subgraph
+create or replace function migrate_entities_tables(
+  schema_name varchar,
+  schema_version deployment_schema_version,
+  subgraph_id varchar
+) returns void
+language plpgsql
+as $function$
+declare
+  index_row  record;
+begin
+  if schema_version <> 'public' then
+    raise 'Schema for %(%) has version %, but should have version ''public''',
+          schema_name, subgraph_id, schema_version;
+  end if;
+
+  execute format('create schema %I', schema_name);
+
+  -- Dirty trick: modify the search path; unqualified table references
+  -- will be looked up in schema_name first
+  perform set_config('search_path', format('%I, public', schema_name), true);
+
+  -- Create the entities table and all its trimmings
+  create table entities(
+      entity       varchar not null,
+      id           varchar not null,
+      data         jsonb,
+      event_source varchar not null,
+      primary key(entity, id)
+  );
+
+  create table entity_history (
+      id           serial primary key,
+      event_id     integer references event_meta_data(id)
+                             on update cascade on delete cascade,
+      entity       varchar not null,
+      entity_id    varchar not null,
+      data_before  jsonb,
+      reversion    bool not null default false,
+      op_id        int2 NOT NULL
+  );
+
+  create index entity_history_event_id_btree_idx
+      on entity_history(event_id);
+
+  -- Set the search path back to the default so that we do not
+  -- influence other database work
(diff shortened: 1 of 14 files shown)


leodasvacas:
Why do we migrate when advancing the subgraph pointer? Is it not sufficient to do it when starting the subgraph?

lutter:
If we do it on subgraph start, we will try to migrate everything at the same time, and because of the shared `event_meta_data` table, subgraph migrations can lock each other out, so that subgraphs that are fast to migrate can get stuck behind subgraphs that are slow (for large subgraphs, the migration can take 10-30 minutes) By spreading the migrations out, and only doing a small number of them at any given time, we minimize the risk of work piling up behind migrating subgraphs. In addition, it gives other database write work (e.g., other subgraph indexing activity) a chance to catch up in between subgraph migrations.

store/postgres/migrations/2019-05-15-215022_migrate_entities/up.sql
```diff
@@ -0,0 +1,173 @@
+create type deployment_schema_version as enum ('public', 'split');
+create type deployment_schema_state as enum ('ready', 'tables');
+
+alter table deployment_schemas
+  add column version   deployment_schema_version,
+  add column migrating bool not null default false,
+  add column state     deployment_schema_state not null default 'ready';
```
leodasvacas: I'm trying to understand this table and it's columns, but it's not the easiest thing to figure out betwen Rust code and SQL migrations. For example I can't figure out what `deployment_schema_state` does. If there was a place describing what this table does the code would be easier to review. Perhaps inside the diesel `table!` macro for it?

store/postgres/migrations/2019-05-15-215022_migrate_entities/up.sql
```diff
@@ -0,0 +1,173 @@
...
+
+alter table deployment_schemas
+  alter column version set not null;
+
+insert into deployment_schemas(subgraph, version)
+select id, 'public'::deployment_schema_version
+  from subgraphs.entities e
+ where entity='SubgraphDeployment'
+   and not exists (select 1 from deployment_schemas d
+                           where d.subgraph=e.id);
```
leodasvacas: I'm deducting from previous migrations and the state of my local DB that the only column we had in this table was the subgraph of subgraphs, and its `version` is null. Does this `insert` assume those things? If so could that be added in a comment?

lutter:
Updated lots of comments, especially in `entities.rs` and rebased to latest master.

store/postgres/migrations/2019-05-15-215022_migrate_entities/up.sql
lutter: I've added comments in `entities.rs` to try and clarify it. Let me know if it doesn't.

store/postgres/migrations/2019-05-15-215022_migrate_entities/up.sql
lutter: Up to this point, the only entries in `deployment_schemas` were the subgraph of subgraphs, and any subgraphs that were deployed after commit `d2248402`. With the changes in this PR, we need entries for all subgraphs in `deployment_schemas` (I also added comments to the migration to that effect)

store/postgres/src/entities.rs
```diff
@@ -95,12 +121,47 @@ mod public {
     joinable!(entity_history -> event_meta_data (event_id));
     allow_tables_to_appear_in_same_query!(entity_history, event_meta_data);
 
+    /// We support different storage schemes per subgraph. This enum is used
+    /// to track which scheme a given subgraph uses and corresponds to the
+    /// `deployment_schema_version` type in the database.
+    ///
+    /// The column `deployment_schemas.version` stores that information for
+    /// each subgraph. Subgraphs that use the `Public` scheme have their
+    /// entities stored in the monolithic `public.entitites` table, subgraphs
```
leodasvacas: s/entitites/entities

store/postgres/src/entities.rs
```diff
@@ -95,12 +121,47 @@ mod public {
...
+    /// the way in which history is recorded. The second step, moving data,
+    /// only requires relatively weak locks that do not block write activity
+    /// to rows in `public.entities` that are not affected by the migration
+    /// of a different subgraph.
+    ///
+    /// The `Ready` state indicates that the subgraph is ready to use the
+    /// storage scheme indicated by `deployment_schemas.version`. After the
+    /// first step of the migration has been done, the `version` field remains
+    /// unchanged, but we indicate that we have put the new schema in place by
+    /// setting the `version` to `Tables`. At the end of the second migration
```
leodasvacas: This sentence is not parsing, it says that `version` remains unchanged and then that it is set to `Tables`.

store/postgres/src/entities.rs
lutter: Oops. Fixed.

store/postgres/src/entities.rs
lutter: That second 'version' should have been 'state'. Fixed that.

store/postgres/src/entities.rs
```diff
@@ -258,6 +370,18 @@ impl<'a> Connection<'a> {
         }
     }
 
+    /// Do any cleanup to bring the subgraph into a known good state
+    pub(crate) fn start_subgraph(&self, subgraph: &SubgraphDeploymentId) -> Result<(), StoreError> {
+        use public::deployment_schemas as dsl;
+
+        Ok(
+            diesel::update(dsl::table.filter(dsl::subgraph.eq(subgraph.to_string())))
+                .set(dsl::migrating.eq(false))
```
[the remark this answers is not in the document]
lutter: No, `migrating` is set to true in its own transaction, the migration work happens in separate txns. That's necessary because `migrating` is essentially a lock, and taking that lock needs to be visible to all db users _before_ the migration work starts. And if the server crashes/gets killed, we are left with `migrating` set to true without actually doing migration work.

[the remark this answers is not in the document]
leodasvacas: Makes sense, thanks for the explanation. Maybe you can add that example of a server crash as a comment on this line?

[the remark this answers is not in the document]
lutter: Yep, good idea. Just did that.

PR #951, opened 2019-05-17
Add an `ETHEREUM_START_BLOCK` environment variable

Jannis:
Consider this a proposal. We've discussed https://github.com/graphprotocol/graph-cli/issues/218 as an alternative to this, but especially with the new block and call handlers, it can take a long time for a subgraph to reach a certain block.

I'm coming around to @davekaj's point that being able to start indexing at a specific block number when developing or debugging a subgraph locally. I've used this environment variable twice already.

### DIFF
--- a/docs/environment-variables.md
+++ b/docs/environment-variables.md
@@ -1,61 +1,74 @@
 # Environment Variables
 
-**Warning**: the names of some of these environment variables will be changed
-at some point in the near future.
+**Warning**: the names of some of these environment variables will be changed at
+some point in the near future.
 
-This page lists the environment variables used by `graph-node` and what
-effect they have. Some environment variables can be used instead of command
-line flags. Those are not listed here, please consult `graph-node --help`
-for details on those.
+This page lists the environment variables used by `graph-node` and what effect
+they have. Some environment variables can be used instead of command line flags.
+Those are not listed here, please consult `graph-node --help` for details on
+those.
 
 ## Getting blocks from Ethereum
 
-* `ETHEREUM_POLLING_INTERVAL`: how often to poll Ethereum for new blocks
-(in ms, defaults to 500ms)
-* `ETHEREUM_RPC_MAX_PARALLEL_REQUESTS`: how many RPC connections to start
-in parallel for block retrieval (defaults to 64)
-* `ETHEREUM_FAST_SCAN_END`: `graph-node` locates
-  blocks with events for a particular subgraph. Most subgraphs do not have
-  events in the first few million blocks, so `graph-node` optimizes for that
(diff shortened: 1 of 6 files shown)


davekaj:
Yeah this would be great! 

Also longer term what would be great is being able to save the state up to a certain block. Lets say the contract starts at 6,000,000, and then you keep running into a failure at 7,000,000. It would be great to say " store the state up to 6,999,999 and then whenever you restart it rolls back to that number. 

.....But now that I think about it this is way harder than I originally thought. Because if you change the mappings, then the first 999,999 blocks you sourced will likely return different values, and you can't really use them. So nevermind !

docs/environment-variables.md
```diff
@@ -1,61 +1,73 @@
...
-* `ETHEREUM_BLOCK_RANGE_SIZE` - number of blocks to scan for events in
-  each request (defaults to 10000).
-* `ETHEREUM_PARALLEL_BLOCK_RANGES` - Maximum number of parallel `eth_getLogs` calls to make when scanning logs for a subgraph. Defaults to 100.
+* `ETHEREUM_BLOCK_BATCH_SIZE`: number of Ethereum blocks to request in parallel
+  (defaults to 50)
+* `ETHEREUM_BLOCK_RANGE_SIZE`: number of blocks to scan for events in each
+  request (defaults to 10000).
+* `ETHEREUM_PARALLEL_BLOCK_RANGES` - Maximum number of parallel `eth_getLogs`
+  calls to make when scanning logs for a subgraph. Defaults to 100.
+* `ETHEREUM_START_BLOCK`: the block number at which subgraphs should start
```
lutter: It might be good to have a warning along the lines of 'do not use this for production subgraphs as it can cause the subgraph to contain incorrect data' or some such here.

lutter (approved):
LGTM

lutter:
> .....But now that I think about it this is way harder than I originally thought. Because if you change the mappings, then the first 999,999 blocks you sourced will likely return different values, and you can't really use them. So nevermind !

I still think it would make sense for testing/developing subgraphs, e.g. when you change your mappings just to work around a problem in a specific block. Just like with using `ETHEREUM_START_BLOCK`, the result would be a tainted subgraph, but for testing I think it would be great to have a command to say 'deploy this subgraph, and initialize it from that other subgraph at this block number'.

davekaj:
when merged should close https://github.com/graphprotocol/graph-explorer/issues/429

docs/environment-variables.md
Jannis: Agreed, can also log a runtime warning.

PR #954, opened 2019-05-19
ethereum, store: Replace chain head listener channels with `tokio::watch`

### LINKED ISSUE #930: Use tokio watch instead of channels for chain head updates
The use of channels for chain head updates caused trouble in #929, tokio [watch](https://docs.rs/tokio/0.1.19/tokio/sync/watch/index.html) is a:

> A single-producer, multi-consumer channel that only retains the last sent value.

Which is exactly what chain head updates should be.

leoyvens:
Simplifies and improves performance of the communication between the listener and block streams. Resolves #930.


lutter (approved):
Nice simplifications. I like seeing those diffstats.

PR #960, opened 2019-05-22
Gracefully handle event data ABI mismatches by trying all potential event handlers

### LINKED ISSUE #913: Differentiate between indexed and non-indexed dynamic-array event fields
Right now graph-node can properly distinguish between indexed and non indexed dynamic arrays (arrays, strings, bytes) when indexing an event. (I believe this is rooted in parity eth-abi)

The outlier problem is this - when we start to track generic interface events, i.e. not supplying a contract address, there is ambiguity we have to deal with. (see https://github.com/ensdomains/resolvers/issues/13#issuecomment-488829262 )

Example:
` event TextChanged(bytes32 indexed node, string indexed indexedKey, string key); `
` event TextChanged(bytes32 indexed node, string indexedKey, string key); `

The above two events have the exact same event signature, however the data they return is different. The top one returns `string indexed indexedKey` as `bytes32` (because the string is hashed and stored as such), and the second one `string indexedKey` is handled as a string - which makes multiple `bytes32` to represent the variable length. 

The way graph-node is implemented now, it can only deal with data type for the event field. We need to allow it to support both. This is a minor blocker on ENS, and an edge case, but it is worthy of us to support it. 

Suggestions:
- look at the ABI for the event and determine if it is `indexed: true` or `false`. Then , combine that with the event signature, so that we cab distinguish them from each other

Jannis:
Resolves #913, together with https://github.com/graphprotocol/graph-cli/pull/281.

This changes the event processing in subgraphs to try all potential event handlers for an event in turn. The problem is that events that match a signature may have parameter data that was encoded in different ways, e.g. using indexed vs. non-indexed event parameters.

By trying all matching handlers, we can handle this situation gracefully and allow any event handler that can decode the event data to process the event. For those that cannot, we'll log a warning just in case.

### DIFF
--- a/graph/src/data/subgraph/mod.rs
+++ b/graph/src/data/subgraph/mod.rs
@@ -479,7 +479,8 @@ pub struct MappingEventHandler {
 
 impl MappingEventHandler {
     pub fn topic0(&self) -> H256 {
-        self.topic0.unwrap_or_else(|| string_to_h256(&self.event))
+        self.topic0
+            .unwrap_or_else(|| string_to_h256(&self.event.replace("indexed ", "")))
     }
 }
 
--- a/runtime/wasm/src/host.rs
+++ b/runtime/wasm/src/host.rs
@@ -353,14 +353,15 @@ impl RuntimeHost {
         source_address_matches && self.handler_for_block(block_trigger_type).is_ok()
     }
 
-    fn handler_for_log(&self, log: &Arc<Log>) -> Result<MappingEventHandler, Error> {
+    fn handlers_for_log(&self, log: &Arc<Log>) -> Result<Vec<MappingEventHandler>, Error> {
         // Get signature from the log
         let topic0 = match log.topics.iter().next() {
             Some(topic0) => topic0,
             None => return Err(format_err!("Ethereum event has no topics")),
         };
 
-        self.data_source_event_handlers
+        let handlers = self
+            .data_source_event_handlers
             .as_ref()
             .ok_or_else(|| {
                 format_err!(
@@ -369,14 +370,18 @@ impl RuntimeHost {
                 )
             })?
             .iter()
-            .find(|handler| *topic0 == handler.topic0())
+            .filter(|handler| *topic0 == handler.topic0())
             .cloned()
-            .ok_or_else(|| {
-                format_err!(
-                    "No event handler found for event in data source \"{}\"",
-                    self.data_source_name,
-                )
-            })
+            .collect::<Vec<_>>();
+
+        if !handlers.is_empty() {
+            Ok(handlers)
+        } else {
+            Err(format_err!(
+                "No event handler found for event in data source \"{}\"",
+                self.data_source_name,
+            ))
+        }
     }
 
     fn handler_for_call(&self, call: &Arc<EthereumCall>) -> Result<MappingCallHandler, Error> {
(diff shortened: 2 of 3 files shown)


graph/src/data/subgraph/mod.rs
```diff
@@ -467,7 +467,8 @@ pub struct MappingEventHandler {
 
 impl MappingEventHandler {
     pub fn topic0(&self) -> H256 {
-        self.topic0.unwrap_or_else(|| string_to_h256(&self.event))
+        self.topic0
+            .unwrap_or_else(|| string_to_h256(&self.event.replace("indexed ", "")))
```
leodasvacas: We can use [trim_start_matches](https://doc.rust-lang.org/std/primitive.str.html#method.trim_start_matches) to be a bit more precise. We should also trim whitespace if we don't already.

runtime/wasm/src/host.rs
```diff
@@ -684,103 +689,138 @@ impl RuntimeHostTrait for RuntimeHost {
...
-            "handler" => &event_handler.handler,
-            "data_source" => &self.data_source_name,
-            "address" => format!("{}", &log.address),
-        );
-
-        // Call the event handler and asynchronously wait for the result
-        let (result_sender, result_receiver) = oneshot::channel();
+        Box::new(
+            future::result(self.handlers_for_log(&log)).and_then(|event_handlers| {
+                stream::iter_ok(event_handlers).fold(state, move |state, event_handler| {
```
leodasvacas: I think we lost some readability in this function, this is using futures and streams event though it's synchronous, it doesn't express that only one event handler will actually be called, and it puts the entire function inside this loop.

A suggestion is to first somehow build a `Vec<(MappingEventHandler, Event)>`, gathering all the ABIs, and then get the correct `event_handler` (for example using [find_map](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.find_map)), so the rest of the function can remain unchanged.

Jannis: I've updated the code, can you take another look?

runtime/wasm/src/host.rs
```diff
@@ -684,103 +689,177 @@ impl RuntimeHostTrait for RuntimeHost {
...
+
+        // If none of the handlers match the event params, fail processing the event
+        if matching_handlers.is_empty() {
+            return Box::new(future::err(format_err!(
+                "No matching event handlers found for in data source \"{}\"",
+                data_source_name,
+            )));
+        }
+
+        // Process the event with all matching handlers; typically there is only one
```
leodasvacas: An event can trigger more than one handler? In what case does this happen? Is it something we want to allow at this point, are we confident that they're processed in a deterministic order?

IMO, until a use case actually requires something more flexible, we should keep a 1:1 relationship between triggers and handlers for the sake of simplicity, though it doesn't look like we actually validate that atm.

runtime/wasm/src/host.rs
```diff
@@ -684,103 +689,177 @@ impl RuntimeHostTrait for RuntimeHost {
...
         };
 
-        debug!(
-            logger, "Start processing Ethereum event";
-            "signature" => &event_handler.event,
-            "handler" => &event_handler.handler,
-            "data_source" => &self.data_source_name,
-            "address" => format!("{}", &log.address),
-        );
+        // Filter out handlers whose signatures don't exist in the contract ABI;
```
leodasvacas: "Filter out" sounds like this situation isn't an error, perhaps "Catch" or "Error on" would better.

leoyvens (commented):
The code is more step-by-step now, I like it! I'm pretty sure the warnings you're getting are fixed on master, you should rebase. I still have a couple comments.

runtime/wasm/src/host.rs
Jannis: We're confident about the deterministic order, because the handlers are matched and executed in the same order they appear in the manifest.

We should disallow it in the manifest; I've filed https://github.com/graphprotocol/graph-cli/issues/286 to track that. Will update the code here to just use the first match.

Jannis:
@leodasvacas Updated to process only one matching handler and fail if there are more than one.

runtime/wasm/src/host.rs
```diff
@@ -684,49 +689,122 @@ impl RuntimeHostTrait for RuntimeHost {
...
+                "No matching event handlers found for in data source \"{}\"",
+                data_source_name,
+            )));
+        }
+
+        // Process the event with the matching handler
+        let (event_handler, params) = matching_handlers.pop().unwrap();
+
+        // Fail if there is more than one matching handler
+        if matching_handlers.len() > 1 {
```
leodasvacas: `pop` will remove the element, so the correct check here is `!matching_handlers.is_empty()`.

runtime/wasm/src/host.rs
```diff
@@ -684,49 +689,122 @@ impl RuntimeHostTrait for RuntimeHost {
...
+                &event_handler.event
+            )));
+        }
+
+        //  let logger = logger.clone();
+        //  let mapping_request_sender = mapping_request_sender.clone();
+        //  let data_source_name = data_source_name.clone();
+
+        //  let block = block.clone();
+        //  let transaction = transaction.clone();
```
leodasvacas: Remove commented out code.

leoyvens (changes requested):
The diff looks cleaner now! Still some comments.

runtime/wasm/src/host.rs
Jannis: Oops, yes, indeed! I moved the `pop` up there after I wrote the length check.

Jannis:
@leodasvacas Updated 😄

leoyvens (approved):

PR #962, opened 2019-05-23
runtime: Cleaner trap error message

### LINKED ISSUE #931: Improve trap logging
Right now when a handler crashes, we log this:
```
Trap: Trap { kind: Host(HostExportError("Failed to parse byte array using `toString()`. This may be caused by attempting to convert a value such as an address that cannot be parsed to a unicode string. Try \'toHexString()\' instead. Bytes: `[191, 219, 80, 220, 102, 200, 223, 159, 217, 104, 141, 143, 229, 160, 195, 65, 38, 66, 118, 69]`. Error: invalid utf-8 sequence of 1 bytes from index 0")) }
```
It would be much more developer friendly if we could just log the error message itself. In this case we would log
```
Failed to parse byte array using `toString()`. This may be caused by attempting to convert a value such as an address that cannot be parsed to a unicode string. Try \'toHexString()\' instead.
```
This would be in addition to the rest of the log line which includes the function that failed.

leoyvens:
Resolves #931.

Message before:
Subgraph instance failed to run: Failed to process trigger: Failed to handle Ethereum event with handler "handleMint": Trap: Trap { kind: Host(HostExportError("Mapping aborted at src/mapping.ts, line 9, column 2, with no message")) }, code: SubgraphSyncingFailure, id: QmW5tn3sxeja657qrupM9wc2aaw16xyJWACNtJu5Mz2Fzq

Message after:
Subgraph instance failed to run: Failed to process trigger: Failed to handle Ethereum event with handler "handleMint": Mapping aborted at src/mapping.ts, line 9, column 2, with no message, code: SubgraphSyncingFailure, id: QmW5tn3sxeja657qrupM9wc2aaw16xyJWACNtJu5Mz2Fzq

### DIFF
--- a/runtime/wasm/src/module/mod.rs
+++ b/runtime/wasm/src/module/mod.rs
@@ -65,6 +65,17 @@ const DATA_SOURCE_CREATE_INDEX: usize = 35;
 const ENS_NAME_BY_HASH: usize = 36;
 const LOG_LOG: usize = 37;
 
+/// A common error is a trap in the host, so simplify the message in that case.
+fn format_wasmi_error(e: Error) -> String {
+    match e {
+        Error::Trap(trap) => match trap.kind() {
+            wasmi::TrapKind::Host(host_error) => host_error.to_string(),
+            _ => trap.to_string(),
+        },
+        _ => e.to_string(),
+    }
+}
+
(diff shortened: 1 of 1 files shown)


Jannis (approved):
Nice!

PR #976, opened 2019-06-03
Add ethereum call `from`

olibearo:
Gives a mapping sender of the `EthereumCall` it is triggering on.

### DIFF
--- a/runtime/wasm/src/module/mod.rs
+++ b/runtime/wasm/src/module/mod.rs
@@ -326,8 +326,7 @@ where
     ) -> Result<BlockState, FailureError> {
         self.start_time = Instant::now();
 
-        // Prepare an EthereumCall for the WASM runtime
-        let arg = EthereumCallData {
+        let call = EthereumCallData {
             to: call.to,
             from: call.from,
             block: EthereumBlockData::from(&self.ctx.block.block),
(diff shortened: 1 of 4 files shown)


Jannis:
@leodasvacas This is a breaking change due to the memory organization of structs in AssemblyScript, right? So we'll have to version the `EthereumCall` struct, similar to how we do it here:
https://github.com/graphprotocol/graph-node/blob/9c4cd1d6e5ab5f52df6a3eff1c3d19a43cf7d8fb/runtime/wasm/src/module/mod.rs#L245-L274

olibearo:
@Jannis ahh, that makes sense and is unfortunate

leodasvacas:
Yep, that will be necessary.

runtime/wasm/src/module/mod.rs
```diff
@@ -326,20 +326,24 @@ where
...
             block: EthereumBlockData::from(&self.ctx.block.block),
             transaction: EthereumTransactionData::from(transaction.deref()),
             inputs,
             outputs,
         };
+        let arg = if self.host_exports().api_version >= Version::new(0, 0, 3) {
+            RuntimeValue::from(self.asc_new::<AscEthereumCall_0_0_3, _>(&call))
+        } else {
+            RuntimeValue::from(self.asc_new::<AscEthereumCall, _>(&call))
+        };
```
leodasvacas: clever!

Issue #987, opened 2019-06-06
Deadlock during subgraph deployment


lutter:
The production database logs show the following deadlock at `2019-06-06T15:10:03.583326Z`:
```
[535302]: [1-1] db=graph,user=postgres ERROR:  deadlock detected
[535302]: [2-1] db=graph,user=postgres DETAIL:  Process 535302 waits for ShareLock on transaction 285547095; blocked by process 535268.
	Process 535268 waits for ShareRowExclusiveLock on relation 827968 of database 827859; blocked by process 535302.
	Process 535302: DELETE FROM large_notifications
	                 WHERE created_at < current_timestamp - interval '300s'
	Process 535268: create table sgd1205.entity_history
	         (
	           id           serial primary key,
	           event_id     integer references event_meta_data(id)
	                          on update cascade on delete cascade,
	           entity       varchar not null,
	           entity_id    varchar not null,
	           data_before  jsonb,
	           reversion    bool not null default false,
	           op_id        int2 NOT NULL
	         )
[535302]: [3-1] db=graph,user=postgres HINT:  See server log for query details.
[535302]: [4-1] db=graph,user=postgres CONTEXT:  while deleting tuple (7,7) in relation "large_notifications"
[535302]: [5-1] db=graph,user=postgres STATEMENT:  DELETE FROM large_notifications
	                 WHERE created_at < current_timestamp - interval '300s'
```

Relation `827968` is the `event_meta_data` table.

To make what follows easier to understand, I'll call what process `535268` did `A`, and what `535302` did `B`.

What's happening is that `A` does some metadata operations prior to issuing the `create table`, which must have resulted in inserting a row into `large_notifications`. At the same time, `B` must have performed normal entity operations, also resulting in a large notification (sending a large notification is what triggers the `DELETE` statement) Both processes will also have taken a lock on `event_meta_data`. Now `A` is waiting for `B`to release its locks on `event_meta_data`, while `B` is waiting for `A` to release its (row) lock on `large_notifications`.

To avoid this particular deadlock, we should change how we do the clean up of old `large_notifications`: we should try to acquire the appropriate table-level lock explicitly with `NOWAIT`, and possibly also select rows to be deleted with `SKIP LOCKED`. It is not a big deal if clean up of `large_notifications` is skipped occasionally.

We should also catch the deadlock error on subgraph creation, and retry the whole thing after a little while for a limited number of times.

lutter:
[...]

PR #989, opened 2019-06-06
store/postgres: Reduce locking when cleaning up large notifications

lutter:
The locks we need to acquire to clean up old large notifications can make a
transaction wait for one (or several) slow transactions, stalling the whole
system, or even lead to a deadlock.

With this change, we only clean up old large notifications that are not
currently locked by other transactions so that we will never wait for
them.

Fixes https://github.com/graphprotocol/graph-node/issues/987

### DIFF
--- a/store/postgres/src/notification_listener.rs
+++ b/store/postgres/src/notification_listener.rs
@@ -255,11 +255,27 @@ impl JsonNotification {
 
             // Delete any notifications older than 5 minutes; this serves
             // as a regular cleanup so the table doesn't grow in size indefinitely
-            diesel::sql_query(
-                "DELETE FROM large_notifications
-                 WHERE created_at < current_timestamp - interval '300s'",
-            )
-            .execute(conn)?;
+            //
+            // We make sure that we do not wait for any locks when we do the
+            // cleanup to avoid stalling this transaction. There is not much
+            // danger in this cleanup sometimes leaving rows behind, as long
+            // as they get cleaned up eventually
+
+            // 'delete' locks the table in 'row exclusive' mode; see that we
+            // can get that lock without waiting
+            let lock =
+                diesel::sql_query("lock table large_notifications in row exclusive mode nowait")
+                    .execute(conn);
+            if lock.is_ok() {
+                diesel::sql_query(
+                    "delete from large_notifications
+                     where id in
+                        (select id from large_notifications
+                         where created_at < current_timestamp - interval '300s'
+                         for update skip locked)",
+                )
+                .execute(conn)?;
+            }
         }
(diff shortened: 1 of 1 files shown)


store/postgres/src/notification_listener.rs
```diff
@@ -255,11 +255,27 @@ impl JsonNotification {
...
+            // We make sure that we do not wait for any locks when we do the
+            // cleanup to avoid stalling this transaction. There is not much
+            // danger in this cleanup sometimes leaving rows behind, as long
+            // as they get cleaned up eventually
+
+            // 'delete' locks the table in 'row exclusive' mode; see that we
+            // can get that lock without waiting
+            let lock =
+                diesel::sql_query("lock table large_notifications in row exclusive mode nowait")
+                    .execute(conn);
```
leodasvacas: Does this keep the lock after the execute? Maybe the lock and the delete should be in the same transaction?

store/postgres/src/notification_listener.rs
```diff
@@ -255,11 +255,27 @@ impl JsonNotification {
...
+            if lock.is_ok() {
+                diesel::sql_query(
+                    "delete from large_notifications
+                     where id in
+                        (select id from large_notifications
+                         where created_at < current_timestamp - interval '300s'
+                         for update skip locked)",
```
leodasvacas: What does `for update skip locked` do and why do we need both the table lock and this?

store/postgres/src/notification_listener.rs
lutter: This code is run when we already are in a transaction, and Postgres doesn't support autonomous transactions. Yes, the lock is kept until the end of the overall txn (that happens for all locks, they only get released when the txn commits or rolls back)

All that lock statement is doing is explicitly acquiring the lock that the `delete` would acquire implicitly, with the only difference being that `delete` would wait for this lock, and the above code reports an error if it can't get the lock immediately, in which case we don't even try to delete.

store/postgres/src/notification_listener.rs
lutter: The code does the locking that `delete` does explicitly: it takes a `row exclusive` lock on the table, and another lock on every row affected by the `delete`. The explicit locking is done in a way that it will never wait for the lock. The `for update` says "lock each selected row for modification" and the `skip locked` says "don't bother with rows that are already locked".

Together, this gives us the effect of "delete what we can without waiting for locks, and leave everything else for somebody else to clean up". All this code is just so that the `large_notifications` table doesn't grow without bounds; it won't really hurt if stuff sticks around for a while longer than 5 minutes.

PR #990, opened 2019-06-06
Fetch dynamic data sources _after_ the manifest

leoyvens:
So we don't block the IPFS network call on the graphql query, which may cause timeouts. Also time the query to confirm this theory.

### DIFF
--- a/core/src/subgraph/loader.rs
+++ b/core/src/subgraph/loader.rs
@@ -268,8 +268,11 @@ where
                     })
             })
             .map(move |state| {
-                trace!(timing_logger, "loaded dynamic data sources";
-                                      "ms" => start_time.elapsed().as_millis());
+                trace!(
+                    timing_logger,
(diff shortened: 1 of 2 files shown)


Jannis:
For the subgraph for which we've recently seen IPFS resolution issues on startup (in the form of `deadline elapsed` errors), I've confirmed that the dynamic data source queries take a long time:

The current Melon subgraph has 770 dynamic data sources, so we have to run 8 GraphQL queries to get all entities needed to start things up. I only ran the first of these 8 and it took longer than 30s. Total for all of them would be around 4+ minutes.

Since we used to join the IPFS resolution and the data source loading futures, it makes sense that the first timed out while the 8 queries were still running. That would explain the `deadline elapsed` errors.

@leodasvacas I think this PR is right on point. I've slightly changed the log message and will merge this now and roll it out to the hosted service.

PR #991, opened 2019-06-07
Update subgraph reassignment conditional to allow all deployment version types

fordN:
The `AbortUnless` conditional operation being generated by the `subgraph_reassign` RPC method was only allowing deployments that are the current version to be reassigned. 

This PR updates the conditional simply to simply check if any existing deployment version matches the subgraph name and deployment id provided.

### DIFF
--- a/core/src/subgraph/registrar.rs
+++ b/core/src/subgraph/registrar.rs
@@ -846,14 +846,13 @@ fn reassign_subgraph(
     ops.extend(read_summaries_abort_ops);
 
     ops.push(EntityOperation::AbortUnless {
-        description:
-            "Provided name-deploymentId pair must match an existing, current subgraph version"
-                .to_owned(),
+        description: "Provided name-deploymentId pair must match an existing subgraph version"
+            .to_owned(),
         query: SubgraphEntity::query()
             .filter(EntityFilter::new_in("name", vec![name.clone().to_string()])),
         entity_ids: version_summaries
             .iter()
-            .filter(|version| version.current)
+            .filter(|version| version.deployment_id == hash)
             .map(move |summary| summary.clone().subgraph_id)
             .collect(),
     });
--- a/core/src/subgraph/registrar.rs
+++ b/core/src/subgraph/registrar.rs
@@ -284,13 +284,11 @@ where
 
     fn reassign_subgraph(
         &self,
-        name: SubgraphName,
         hash: SubgraphDeploymentId,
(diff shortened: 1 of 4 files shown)


core/src/subgraph/registrar.rs
```diff
@@ -846,14 +846,13 @@ fn reassign_subgraph(
...
-            "Provided name-deploymentId pair must match an existing, current subgraph version"
-                .to_owned(),
+        description: "Provided name-deploymentId pair must match an existing subgraph version"
+            .to_owned(),
         query: SubgraphEntity::query()
             .filter(EntityFilter::new_in("name", vec![name.clone().to_string()])),
         entity_ids: version_summaries
             .iter()
-            .filter(|version| version.current)
+            .filter(|version| version.deployment_id == hash)
```
lutter: I am not convinced of the locking logic here: what we want to make sure of is that the assignment is unchanged when we actually do the reassignment operation in the database. I don't think it matters which subgraphs use that deployment, and it's totally fine if some of them stop using this deployment while we reassign, as long as _something_ is still using the deployment.

lutter: I think something like
```
let current_node_id = lookup the assignment and get its node_id;
ops.push(EntityOperation::AbortUnless {
        description: "Assignment is unchanged".to_owned(),
        query: SubgraphDeploymentAssignmentEntity::query().filter(EntityFilter::And(vec![
            EntityFilter::new_equal("nodeId", current_node_id.to_string()),
            EntityFilter::new_equal("id", hash.clone().to_string()),
        ])),
        entity_ids: vec![hash.clone().to_string()],
});
```

Jannis: Agreed, that makes more sense. I can see where @fordN is coming from though.

fordN: That's a good point; the subgraph <-> deployment relationship doesn't need to be a concern of the reassigner, but we do need to ensure the deployment doesn't change underneath us.  

Along the same vein,  I don't think we need the version summaries or the version abort operations that come with them. (https://github.com/graphprotocol/graph-node/blob/cad6309086c5754987918a9cae1fb71a9dcb0210/core/src/subgraph/registrar.rs#L844-L846)

lutter: Yes, I completely agree. The logic should simply be "if that assignment still exists, change the node for the assignment"

PR #997, opened 2019-06-07
Start subgraphs in parallel

leoyvens:
Inside the blocking thread pool, and being careful to start them all in a single future. This means subgraphs no longer have to wait for others to start before they can be started.

### DIFF
--- a/core/src/subgraph/registrar.rs
+++ b/core/src/subgraph/registrar.rs
@@ -214,10 +214,25 @@ where
                     .collect::<Result<HashSet<SubgraphDeploymentId>, _>>()
             })
             .and_then(move |subgraph_ids| {
-                let provider = provider.clone();
-                stream::iter_ok(subgraph_ids).for_each(move |id| {
-                    start_subgraph(id, &*provider, logger.clone()).map_err(|()| unreachable!())
-                })
+                // This operation should finish only after all subgraphs are
+                // started. We wait for the spawned tasks to complete by giving
+                // each a `sender` and waiting for all of them to be dropped, so
+                // the receiver terminates without receiving anything.
+                let (sender, receiver) = tokio::sync::mpsc::channel::<()>(1);
+                for id in subgraph_ids {
+                    let sender = sender.clone();
+                    tokio::spawn(
+                        graph::util::futures::blocking(start_subgraph(
+                            id,
+                            &*provider,
+                            logger.clone(),
+                        ))
+                        .map(move |()| drop(sender))
+                        .map_err(|()| unreachable!()),
+                    );
+                }
+                drop(sender);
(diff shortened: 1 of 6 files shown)


core/src/subgraph/registrar.rs
```diff
@@ -214,10 +214,25 @@ where
...
+                            id,
+                            &*provider,
+                            logger.clone(),
+                        ))
+                        .map(move |()| drop(sender))
+                        .map_err(|()| unreachable!()),
+                    );
+                }
+                drop(sender);
+                receiver.collect().then(|_| future::ok(()))
```
lutter: I am definitely missing some background, but this looks to me like it's doing something like reference counting using the sender and receiver. What I am not clear on is how does `receiver.collect()` know how many senders to waits for? Does `sender.clone()` somehow tell the receiver about the existence of a new sender?

lutter: Just read the source, and yes, there's a count of senders that the sender and receiver share, amongst other things, a count of existing senders. That all makes sense now!

graph/src/util/futures.rs
```diff
@@ -486,3 +486,15 @@ mod tests {
         assert_eq!(result, Ok(10));
     }
 }
+
+/// Convenient way to annotate a future with `tokio_threadpool::blocking`.
+///
+/// Panics if called from outside a tokio runtime.
+pub fn blocking<T, E>(mut f: impl Future<Item = T, Error = E>) -> impl Future<Item = T, Error = E> {
```
lutter: Nice! This will be very useful to send DB work to the background, too.

lutter (approved):

core/src/subgraph/registrar.rs
```diff
@@ -214,10 +214,25 @@ where
...
             .and_then(move |subgraph_ids| {
-                let provider = provider.clone();
-                stream::iter_ok(subgraph_ids).for_each(move |id| {
-                    start_subgraph(id, &*provider, logger.clone()).map_err(|()| unreachable!())
-                })
+                // This operation should finish only after all subgraphs are
+                // started. We wait for the spawned tasks to complete by giving
+                // each a `sender` and waiting for all of them to be dropped, so
+                // the receiver terminates without receiving anything.
+                let (sender, receiver) = tokio::sync::mpsc::channel::<()>(1);
```
Jannis: Why not a oneshot? (I'm assuming those are a little cheaper to create, which may not be true.)

Jannis (approved):

core/src/subgraph/registrar.rs
leodasvacas: The oneshot sender is not clonable, so we'd have to wrap it an `Arc`. The approaches seem equivalent to me so I'll just merge this as is.

PR #1001, opened 2019-06-08
Add custom one-line log format

### LINKED ISSUE #994: Logs are hard to grep for subgraph-specific messages
The current compact logging is easy to read by humans but hard to grep than one-line logging where all fields (subgraph ID, component, block number, message) are on a single line.

For any production deployments of graph-node, the ability to grep logs is more important than sheer readability.

Jannis:
Resolves #994.

Changes the log format from:
image
to:
image

Advantages:

- Each log message is on a single line
- Each log message for a subgraph has the subgraph ID in the same line
- Subgraph IDs and components are highlighted
- Generally much easier to grep for anything, subgraph logs specifically

Disadvantages:

- Lines are a little longer and more cluttered visually

### DIFF
--- a/graph/src/log/mod.rs
+++ b/graph/src/log/mod.rs
@@ -1,8 +1,8 @@
-use slog::{o, Drain, FilterLevel, Logger};
+use slog::*;
 use slog_async;
 use slog_envlogger;
-use slog_term;
-use std::env;
+use slog_term::*;
+use std::{env, fmt, io, result};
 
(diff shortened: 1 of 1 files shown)


lutter (approved):
Wow. That's quite a lot of ceremony. But the logging looks great to me!

The tests failed because of a transient error downloading ipfs) I restarted the test; if it doesn't go through, can you make sure it does before committing?

Issue #996, opened 2019-06-07
Remove the `event_meta_data` table


lutter:
The table `event_meta_data` is shared by all subgraphs, and is very large. Because subgraphs' `entity_history` references it, subgraph deployment has to get a `SHARE UPDATE EXCLUSIVE` lock on `event_meta_data`; because of that, subgraph deployment can get stuck behind an autovacuum of `event_meta_data`, and writes to `event_meta_data` can get stuck behind either of these operations. Autovacuum of `event_meta_data` can take an hour, and will only become longer as the table grows.

The table doesn't serve much of a purpose: it mostly makes it so we do not store the full block hash for a history event in `entity_history`, but only an integer `event_id` which points into `event_meta_data`.

We should make this table unnecessary by storing the block hash as the `event_source` in `entity_history` directly.

lutter:
We won't do this. Instead, we will move to relational schemas which do not need `event_meta_data`

PR #1002, opened 2019-06-09
store/postgres: Make subgraph creation not block writers

lutter:
Creating a subgraph can block all other subgraph writers if the creation
itself has to wait for its lock, e.g., because of an ongoing autovacuum of
event_meta_data. That wait can be long (an hour or more in the case of
autovacuum)

We now do subgraph creation in such a way that we wait for a lock only a
short time (2s) and then sleep for a bit so that other write activity can
proceed.

### DIFF
--- a/graphql/tests/query.rs
+++ b/graphql/tests/query.rs
@@ -13,7 +13,8 @@ lazy_static! {
     static ref TEST_SUBGRAPH_ID: SubgraphDeploymentId = {
         // Also populate the store when the ID is first accessed.
         let id = SubgraphDeploymentId::new("graphqlTestsQuery").unwrap();
-        STORE.create_subgraph_deployment(&id, vec![]).unwrap();
+        let logger = Logger::root(slog::Discard, o!());
+        STORE.create_subgraph_deployment(&logger, &id, vec![]).unwrap();
         insert_test_entities(&**STORE, id.clone());
         id
     };
--- a/store/postgres/src/store.rs
+++ b/store/postgres/src/store.rs
@@ -1023,6 +1023,7 @@ impl StoreTrait for Store {
 
     fn create_subgraph_deployment(
         &self,
+        subgraph_logger: &Logger,
         subgraph_id: &SubgraphDeploymentId,
         ops: Vec<EntityOperation>,
     ) -> Result<(), StoreError> {
@@ -1075,9 +1076,8 @@ impl StoreTrait for Store {
                 // it was because we timed out on the lock and try again.
                 if start.elapsed() >= Duration::from_secs(LOCK_TIMEOUT) {
                     debug!(
-                        self.logger,
-                        "could not acquire lock for creation of subgraph {}, trying again in {}s",
-                        &subgraph_id,
+                        subgraph_logger,
+                        "Could not acquire lock for subgraph creation, trying again in {}s",
                         delay.as_secs()
                     );
                     std::thread::sleep(delay);
--- a/store/postgres/src/store.rs
+++ b/store/postgres/src/store.rs
@@ -1026,13 +1026,69 @@ impl StoreTrait for Store {
         subgraph_id: &SubgraphDeploymentId,
         ops: Vec<EntityOperation>,
     ) -> Result<(), StoreError> {
+        // Various timing parameters, all in seconds
+        const INITIAL_DELAY: u64 = 2;
+        const MAX_DELAY: u64 = 64;
+        const LOCK_TIMEOUT: u64 = 2;
+
         let conn = self.get_conn().map_err(Error::from)?;
         let econn = e::Connection::new(&conn);
+        let mut delay = Duration::from_secs(INITIAL_DELAY);
 
-        conn.transaction(|| {
-            self.apply_entity_operations_with_conn(&econn, ops, None)?;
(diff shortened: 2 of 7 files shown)


lutter:
This PR addresses an indexing stall that happened in production. The stall was cause by subgraph creation getting stuck behind an autovacuum of `event_meta_data`, which in turn blocked all subgraph indexing.

To test this, you need to do the following:
1. Start `graph-node` and have it index one (or many) subgraphs, ideally busy ones
2. Connect to the database with `psql` and enter `begin; lock table event_meta_data in share update exclusive mode;` (this simulates the lock that autovacuum takes)
3. Observe that subgraphs continue to write to the database and do not get stuck
4. Deploy a subgraph. After a while, debug log messages like `could not acquire lock for creation of subgraph {}, trying again in {}s` should appear, but the indexing activity should continue with very little delay
5. Type `rollback;` in the `psql` session. Subgraph creation should now resume and succeed, and indexing should continue as expected

lutter:
With this PR, issue #996 becomes less pressing; we should still do that, but with this PR, `event_meta_data` should not block indexing any more.

store/postgres/src/store.rs
```diff
@@ -1026,13 +1026,69 @@ impl StoreTrait for Store {
...
+            });
+            if let Err(StoreError::Unknown(_)) = &result {
+                // There is no robust way to actually find out that we timed
+                // out on the lock from the error message; diesel shields us
+                // from these details too much. Rather than grep the error
+                // message, which would be very fragile, we assume that if a
+                // failure occurred after more than LOCK_TIMEOUT seconds that
+                // it was because we timed out on the lock and try again.
+                if start.elapsed() >= Duration::from_secs(LOCK_TIMEOUT) {
+                    debug!(
```
Jannis: Do we want this to show up in the subgraph logs in the explorer? If so, we'd need to pass the `LoggerFactory` to the store and do `factory.subgraph_logger(&subgraph_id)` somewhere in this method.

If we don't use a subgraph logger, then the `subgraph_id` should be a field on the message, and the message should start with a capital letter:

```rust
debug!(
  self.logger,
  "Failed to acquire lock for creating a subgraph schema, trying again in {}s",
  delay.as_secs();
  "subgraph_id" => &subgraph_id)
);
```

store/postgres/src/store.rs
```diff
@@ -1026,13 +1026,69 @@ impl StoreTrait for Store {
...
+                        self.logger,
+                        "could not acquire lock for creation of subgraph {}, trying again in {}s",
+                        &subgraph_id,
+                        delay.as_secs()
+                    );
+                    std::thread::sleep(delay);
```
Jannis: If this method was returning a future, I'd advise using
```
futures::future::loop_fn(|state| {
  tokio::timer::Delay(state.delay.clone())
    .and_then(|| {
       // - try creating the subgraph
       // - adjust `state.delay`
       // - return state
    })
})
```
instead of `thread:sleep`. Will `thread::sleep` not block other work?

store/postgres/src/store.rs
lutter: Yes, I'll change that.

store/postgres/src/store.rs
lutter: Yes, `thread::sleep` will block the tokio worker, but that's no different from what was happening before where the database query itself could block the tokio worker. Getting into this retry loop should be very rare though, and I didn't want to rework too much code to handle that. Basically, this change doesn't change anything from the tokio worker's point of view (it'll block just as much as it did before), but it allows other db work to proceed without blocking their workers.

Jannis: Ok, thanks for the explanation!

lutter:
I changed things so that the message about not getting a lock gets logged on the subgraph logger.

graphql/tests/query.rs
```diff
@@ -13,7 +13,8 @@ lazy_static! {
     static ref TEST_SUBGRAPH_ID: SubgraphDeploymentId = {
         // Also populate the store when the ID is first accessed.
         let id = SubgraphDeploymentId::new("graphqlTestsQuery").unwrap();
-        STORE.create_subgraph_deployment(&id, vec![]).unwrap();
+            let logger = Logger::root(slog::Discard, o!());
```
Jannis: The indentation here is one level too deep.

store/postgres/src/store.rs
```diff
@@ -1023,16 +1023,72 @@ impl StoreTrait for Store {
...
+                // There is no robust way to actually find out that we timed
+                // out on the lock from the error message; diesel shields us
+                // from these details too much. Rather than grep the error
+                // message, which would be very fragile, we assume that if a
+                // failure occurred after more than LOCK_TIMEOUT seconds that
+                // it was because we timed out on the lock and try again.
+                if start.elapsed() >= Duration::from_secs(LOCK_TIMEOUT) {
+                    debug!(
+                        subgraph_logger,
+                        "could not acquire lock for subgraph creation, trying again in {}s",
```
Jannis: `Could not` or `failed to` instead of `could not`

lutter: Fixed

graphql/tests/query.rs
lutter: Fixed.

PR #1004, opened 2019-06-12
Make event matching as backwards-compatible as possible

Jannis:
The recent changes we made to be able to distinguish between event variants with `indexed` and non-`indexed` parameters (#960) are breaking existing subgraphs. This PR provides a fallback for subgraphs that haven't migrated to `indexed` in event signatures yet:

If there is only one event variant with the same name as in the manifest, consider it a match to the signature from the manifest even if the `indexed` hints are missing in the manifest.


ghost (approved):

fordN (approved):
Looks good to me!

Issue #995, opened 2019-06-07
Attribute indexes on `BigInt` columns are not usable


lutter:
This query to find CryptoKitties that were born in a certain time window took about ~12s in production:
```
SELECT "sgd1171"."entities"."data", "sgd1171"."entities"."entity"
  FROM "sgd1171"."entities"
  WHERE "sgd1171"."entities"."entity" = ANY('{CryptoKitty}')
    AND (data -> 'birthTime'->> 'data')::numeric >= 1514761200
    AND (data -> 'birthTime'->> 'data')::numeric < 1519858800
    LIMIT 100
    OFFSET 100
```

The query plan uses the index `attr_5_5_crypto_kitty_birth_time` but since it is defined as
```
"attr_5_5_crypto_kitty_birth_time" btree (((data -> 'birthTime'::text) ->> 'data'::text)) WHERE entity::text = 'CryptoKitty'::text
```
which makes it a BTree on strings, not on numbers, so that the range query can't exploit the BTree structure, instead doing a full scan.

Changing the index to an index on numbers:
```
create index attr_5_5_crypto_kitty_birth_time on sgd1171.entities (((data->'birthTime'->>'data')::numeric)) where entity = 'CryptoKitty';
```
makes this query fast again (~ 300ms)

lutter:
Fixed by PR #1006 for new subgraphs. For existing subgraphs, we'd need to write and run a migration.

PR #1006, opened 2019-06-13
store/postgres: Index BigInt/BigDecimal attributes as numbers

lutter:
Partial fix for https://github.com/graphprotocol/graph-node/issues/995

### DIFF
--- a/store/postgres/src/entities.rs
+++ b/store/postgres/src/entities.rs
@@ -1454,7 +1454,15 @@ impl Table {
             ValueType::String => (String::from("gin"), String::from("gin_trgm_ops"), "->>"),
             ValueType::List => (String::from("gin"), String::from("jsonb_path_ops"), "->"),
         };
-
+        // Cast between the type we store in JSONB for the field and the type
+        // as which comparisons should be made. For example, we store BigInt
+        // as a string in JSONB, but the comparison needs to be made as
+        // a number
+        let type_cast = match index.field_value_type {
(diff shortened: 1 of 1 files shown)


store/postgres/src/entities.rs
```diff
@@ -1445,16 +1445,32 @@ impl Table {
...
         let (index_type, index_operator, jsonb_operator) = match index.field_value_type {
-            ValueType::Boolean
-            | ValueType::BigInt
+            ValueType::Boolean => {
+                // There's no point in having an index on a boolean attribute;
+                // since there are only two possible values, the index will
+                // not be very selective, and Postgres will just scan the
+                // table anyway
+                return Ok(1);
+            }
```
leodasvacas: If we adopt a query costing model that boils down to `constant + size(result)`, then a good property to have is that cheap queries have small result sizes and expensive queries have large result sizes. Having indexes on booleans helps maintain that property.

If an entity has evenly distributed `false` and `true`, then a full scan will be done but will generate a proportionally large result set, so we're good. However if the table for example has only a few `true` values, then the index is important. Without the index, selecting the `true` rows will cause a full scan but return a small result set, breaking the property. With an index, we'd use it and therefore have a cheap query for a small result set, maintaining the property.

lutter: You are right that this is not the right way to address our overindexing, and we should look more to see what indexes are/are not used before removing some. I took that part out of the PR.

Issue #988, opened 2019-06-06
Unable to create a subgraph when Github username has numbers in it


nenadjaja:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
Users are unable to create subgraphs when their Github usernames start with numbers. 

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem.**
- Signup fr  Explorer using a Github username that starts with a number. Go to your Dashboard and click on Add subgraph. Once you filled out the fields, click on Create Subgraph, and it will return an error below:
<img width="713" alt="Screen Shot 2019-06-06 at 8 45 04 AM" src="https://user-images.githubusercontent.com/1865664/59051573-0edb8f80-8842-11e9-98de-8a6d5da86d3d.png">

Error Message:
```
jsonRpcError:  { code: -32602,
  message:
   'Invalid params: invalid value: string "2nena4test/lala", expected valid subgraph name.' }
```

**What is the expected behavior?**
Users should be able to create subgraphs regardless of their GIthub usernames

PR #1007, opened 2019-06-13
graph/data: Relax restrictions on subgraph names

lutter:
Allow subgraph names where a part starts with a number, but still forbid
ones where a part does not contain a single letter (e.g., just a number)

Fixes https://github.com/graphprotocol/graph-node/issues/988


nenadjaja:
yay!

PR #1012, opened 2019-06-18
Only use terminal colors if stdout is a tty

Jannis:
Avoids writing `[36m` type term color strings to non-TTY terminals.

This is not 100% ideal, as the output may be to stderr instead of stdout, but we have no way of knowing where the `TermDecorator` is writing to.


fordN (approved):
👍

PR #1014, opened 2019-06-20
store/postgres: Index string attributes as btrees

lutter:
We used to create indexes on string attributes as GIN indexes; with the way
we translate filters into SQL, these indexes do not get used. Since they
are only on individual attributes, there's no downside to using a standard
BTree index, which makes them usable for our filters.


lutter:
This PR does not change existing indexes, only changes what indexes get created for new subgraphs.

In one example from production it brings the time for a query down from 6s to 200ms.

Issue #973, opened 2019-05-31
GraphQL `_contains` filter do not work as expected


lutter:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
When using a `_contains` filter in GraphQL, results are not returned as expected. For example, for [ENS](https://thegraph.com/explorer/subgraph/ensdomains/ens), this query should return a lot of results, but returns none:
```
{
  domains(first: 5, where:{name_contains:"e"}) {
    name
  }
}
```

**What is the expected behavior?**
The above query returns results.

The underlying issue is that the argument of the `_contains` filter is passed verbatim into a SQL `like` clause, i.e. the query contains something along the lines of `name like 'e'`. The transformation needs to be changed to generate SQL like `name like '%e%'`. In fact, this query does return results, though users should not have to write that:
```
{
  domains(first: 5, where:{name_contains:"%e%"}) {
    name
  }
}
```

Jannis:
Fixed as of #1015.

PR #1015, opened 2019-06-21
store/postgres: Fix `_contains` filter for String/Bytes (#973)

Jannis:
The bug was that `LIKE` in Postgres requires `%...%` around the search value.


lutter (approved):
LGTM

Jannis:
This needs one change: Don't surround the search text with `%`s if they are already present in the search text. Otherwise this change would be breaking.

PR #1022, opened 2019-07-01
Don't fail when no matching event handlers are found (#1021)

### LINKED ISSUE #1021: Graph processing should not fail when no matching handlers are found
The new support for trying multiple handlers is excellent, but graph processing still fails if at least one handler matches `topics[0]`, but none of them can successfully decode the event data.

This makes graphs that subscribe to events from multiple addresses still vulnerable to disruption. Instead, a warning should be logged and graph processing should continue.

Jannis:
Instead, just log a warning and skip processing the event.

Fixes #1021.

### DIFF
--- a/runtime/wasm/src/host.rs
+++ b/runtime/wasm/src/host.rs
@@ -774,17 +774,20 @@ impl RuntimeHostTrait for RuntimeHost {
             })
             .collect::<Vec<_>>();
 
-        // If none of the handlers match the event params, fail processing the event
+        // If none of the handlers match the event params, log a warning and
+        // skip the event. See #1021.
         if matching_handlers.is_empty() {
-            return Box::new(future::err(format_err!(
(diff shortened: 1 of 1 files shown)


ghost (approved):

runtime/wasm/src/host.rs
```diff
@@ -774,17 +774,20 @@ impl RuntimeHostTrait for RuntimeHost {
             })
             .collect::<Vec<_>>();
 
-        // If none of the handlers match the event params, fail processing the event
+        // If none of the handlers match the event params, log a warning and
+        // skip the event. See https://github.com/graphprotocol/graph-node/issues/1021
```
leodasvacas: I'm fine with referencing issues with just `#1021` if you also prefer that.

Jannis: I am :)

PR #1027, opened 2019-07-03
Support connecting to multiple Ethereum networks on a single graph-node

### LINKED ISSUE #1018: Allow Graph Nodes to connect to multiple networks

fordN:
Resolves #1018
Updating the graph-node to handle multiple, different Ethereum networks on a single instance allows for more flexibility in devops because one does not have to run separate nodes to support subgraphs on different Ethereum chains. 

Summary:
The `ethereum-rpc`, `ethereum-ipc`, `ethereum-ws` CLI arguments now accept multiple values each. These supplied Ethereum networks are used to create sets of stores, runtime host builders and block stream builders (containing one item per Ethereum network). The subgraph instance manager and the subgraph registrar now take these sets (implemented as HashMap) as input and when deploying a subgraph version will decide which store, runtime, and block stream to use based on the Ethereum network name defined in the subgraph's manifest.

### DIFF
--- a/core/src/subgraph/validation.rs
+++ b/core/src/subgraph/validation.rs
@@ -2,7 +2,7 @@ use graph::prelude::*;
 
 pub fn validate_manifest(
     manifest: SubgraphManifest,
-) -> Result<SubgraphManifest, SubgraphRegistrarError> {
+) -> Result<(String, SubgraphManifest), SubgraphRegistrarError> {
     let mut errors: Vec<SubgraphManifestValidationError> = Vec::new();
 
     // Validate that the manifest has a `source` address in each data source
@@ -57,8 +57,14 @@ pub fn validate_manifest(
         errors.push(SubgraphManifestValidationError::DataSourceBlockHandlerLimitExceeded)
     }
 
+    let mut network_name = String::from("none");
+    match manifest.network_name() {
+        Ok(n) => network_name = n,
+        Err(e) => errors.push(e),
+    };
+
     if errors.is_empty() {
-        return Ok(manifest);
+        return Ok((network_name, manifest));
     }
 
     return Err(SubgraphRegistrarError::ManifestValidationError(errors));
--- a/core/src/subgraph/validation.rs
+++ b/core/src/subgraph/validation.rs
@@ -2,7 +2,7 @@ use graph::prelude::*;
 
 pub fn validate_manifest(
     manifest: SubgraphManifest,
-) -> Result<(String, SubgraphManifest), SubgraphRegistrarError> {
+) -> Result<SubgraphManifest, SubgraphRegistrarError> {
     let mut errors: Vec<SubgraphManifestValidationError> = Vec::new();
 
     // Validate that the manifest has a `source` address in each data source
@@ -57,14 +57,8 @@ pub fn validate_manifest(
         errors.push(SubgraphManifestValidationError::DataSourceBlockHandlerLimitExceeded)
     }
 
-    let mut network_name = String::from("none");
-    match manifest.network_name() {
-        Ok(n) => network_name = n,
-        Err(e) => errors.push(e),
-    };
-
     if errors.is_empty() {
-        return Ok((network_name, manifest));
+        return Ok(manifest);
     }
 
     return Err(SubgraphRegistrarError::ManifestValidationError(errors));
--- a/graph/src/data/subgraph/mod.rs
+++ b/graph/src/data/subgraph/mod.rs
@@ -264,6 +264,12 @@ impl From<Error> for SubgraphRegistrarError {
     }
 }
 
+impl From<SubgraphManifestValidationError> for SubgraphRegistrarError {
+    fn from(e: SubgraphManifestValidationError) -> Self {
+        SubgraphRegistrarError::ManifestValidationError(vec![e])
+    }
+}
+
 #[derive(Fail, Debug)]
 pub enum SubgraphAssignmentProviderError {
     #[fail(display = "Subgraph resolve error: {}", _0)]
@@ -312,7 +318,7 @@ pub enum SubgraphAssignmentProviderEvent {
 pub enum SubgraphManifestValidationError {
     #[fail(display = "subgraph source address is required")]
     SourceAddressRequired,
-    #[fail(display = "subgraph cannot index data from multiple different ethereum networks")]
+    #[fail(display = "subgraph cannot index data from different Ethereum networks")]
     MultipleEthereumNetworks,
     #[fail(display = "subgraph must have at least one Ethereum network data source")]
     EthereumNetworkRequired,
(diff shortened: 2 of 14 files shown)


fordN:
The `subgraph_provider_events` test has been hanging which causes the test suite to fail.  I am looking into why this is happening today and will update the PR accordingly.

fordN:
Note: These changes could further complicate the Ethereum network version consistency issue we already have with the DB.  
The current state of the PR relies on consistent (in the context of a single `graph-node`) Ethereum network names. The Ethereum network name of a deployed subgraph must match one of the names provided on node startup as an ethereum-rpc, ipc, or ws arguments.  Currently some users have been using the process of incrementing the network name, so it is entered as a new network in the database and does not conflict with existing networks. With my changes the subgraph manifests will need to be updated as well, making it less viable and leaving only the other option of starting from a fresh database.  

The Ethereum network name could be represented by an enum with the supported network types and then user supplied network names will resolve to one of the supported names.  This would allow graph-nodes to be consistent with which Ethereum network names they support.  

On the other hand, the current setup does open up other options with the flexible network naming.  One example is starting up a `graph-node` with multiple mainnet providers (example: mainnet_alchemy and mainnet_infura) and then running a subgraph simultaneously on both to compare performance and data from each of the providers or to leverage the capabilities of each for different subgraphs.

graph/src/data/subgraph/mod.rs
```diff
@@ -312,6 +312,10 @@ pub enum SubgraphAssignmentProviderEvent {
 pub enum SubgraphManifestValidationError {
     #[fail(display = "subgraph source address is required")]
     SourceAddressRequired,
+    #[fail(display = "subgraph cannot index data from multiple different ethereum networks")]
```
Jannis: Capital E in "Ethereum". Multiple different -> different

node/src/main.rs
```diff
@@ -391,26 +405,26 @@ fn async_main() -> impl Future<Item = (), Error = ()> + Send + 'static {
...
-
-    // If we drop the event loop the transport will stop working.
-    // For now it's fine to just leak it.
-    std::mem::forget(transport_event_loop);
+    let eth_adapters: Vec<(
+        &str,
+        Arc<graph_datasource_ethereum::EthereumAdapter<graph_datasource_ethereum::Transport>>,
+    )> = ethereum_rpc
+        .map(|arguments| parse_ethereum_networks_and_nodes(arguments, ConnectionType::RPC))
+        .or_else(|| {
```
Jannis: Instead of using `or_else` like this, I think it's best if we combine all adapters for all provided flags into a flat hash map. And fail if for any network name, more than one URL is provided.

node/src/main.rs
```diff
@@ -422,57 +436,63 @@ fn async_main() -> impl Future<Item = (), Error = ()> + Send + 'static {
...
+            info!(
+                logger, "Connecting to Ethereum...";
+                "network" => &network_name,
+            );
+            match eth_adapter.net_identifiers(&logger).wait() {
+                Ok(network_identifier) => {
+                    info!(
+                    logger, "Connected to Ethereum";
+                    "network" => &network_name,
+                    "network version" => &network_identifier.net_version,
```
Jannis: Add an underscore between network and version.

core/src/subgraph/validation.rs
```diff
@@ -57,8 +57,14 @@ pub fn validate_manifest(
...
 
+    let mut network_name = String::from("none");
+    match manifest.network_name() {
+        Ok(n) => network_name = n,
+        Err(e) => errors.push(e),
+    };
+
     if errors.is_empty() {
-        return Ok(manifest);
+        return Ok((network_name, manifest));
```
Jannis: Can we keep this return value/type as it is? We're going to eventually add non-Ethereum subgraphs and is like to avoid too much Ethereum specific information exposed through generic APIs like this one.

node/src/main.rs
```diff
@@ -481,45 +501,64 @@ fn async_main() -> impl Future<Item = (), Error = ()> + Send + 'static {
...
 
-    // Prepare a block stream builder for subgraphs
-    let block_stream_builder = BlockStreamBuilder::new(
-        store.clone(),
-        store.clone(),
-        eth_adapter.clone(),
-        node_id.clone(),
-        *REORG_THRESHOLD,
-    );
+    let mut block_stream_builders = HashMap::new();
```
Jannis: I haven't checked the internals of the stream and host builders yet but could we keep these to one instance each, have them take the stores and eth adapters map and then take the network name as a build parameter?

node/src/main.rs
```diff
@@ -391,26 +405,25 @@ fn async_main() -> impl Future<Item = (), Error = ()> + Send + 'static {
...
-    std::mem::forget(transport_event_loop);
+    let eth_adapters = [
+        (ConnectionType::RPC, ethereum_rpc),
+        (ConnectionType::IPC, ethereum_ipc),
+        (ConnectionType::WS, ethereum_ws),
+    ]
+    .iter()
+    .cloned()
+    .filter(|(_, values)| values.is_some())
+    .fold(HashMap::new(), |networks, (connection_type, values)| {
```
Jannis: I'd change the `networks` variable name here, it's more like adapters?

node/src/main.rs
```diff
@@ -391,26 +405,25 @@ fn async_main() -> impl Future<Item = (), Error = ()> + Send + 'static {
...
+        match parse_ethereum_networks_and_nodes(logger.clone(), values.unwrap(), connection_type) {
+            Ok(adapters) => networks.into_iter().chain(adapters).collect(),
+            Err(e) => {
+                panic!(
+                    "Failed to parse ethereum networks and create ethereum adapters: {}",
```
Jannis: 2x ethereum -> Ethereum

node/src/main.rs
```diff
@@ -422,57 +435,62 @@ fn async_main() -> impl Future<Item = (), Error = ()> + Send + 'static {
...
+        .iter()
+        .map(|(network_name, eth_adapter)| {
+            info!(
+                logger, "Connecting to Ethereum...";
+                "network" => &network_name,
+            );
+            match eth_adapter.net_identifiers(&logger).wait() {
+                Ok(network_identifier) => {
+                    info!(
+                    logger, "Connected to Ethereum";
```
Jannis: The formatting/indentation here is a bit funny.

runtime/wasm/src/host.rs
```diff
@@ -70,14 +75,29 @@ where
     fn build(
         &self,
         logger: &Logger,
+        network_name: String,
         subgraph_id: SubgraphDeploymentId,
         data_source: DataSource,
     ) -> Result<Self::Host, Error> {
+        let store = self.stores.get(&network_name).ok_or_else(|| {
+            format_err!(
+                "No Store found that matches subgraph network: \"{}\"",
```
Jannis: I'd do Store -> store here.

PR #1029, opened 2019-07-04
store/postgres: Log errors when reading db notifications

lutter:
We used to silently ignore errors when reading notifications from the
database. We now log them to make it easier to tell if anything is going
wrong there, e.g., to at least get a message if the connection gets dropped
for whatever reason.


lutter:
No matter whether there are any problems with the notification listener, we should log any errors, even if they turn out to be harmless. Whoever reviews this, if its ok with you, please merge on my behalf.

PR #1049, opened 2019-07-09
Use an async client for sending Elasticsearch logs

Jannis:
Using a synchronous client can lead to `attempted to run an executor while another executor is already running` errors like the one reported on https://github.com/seanmonstar/reqwest/issues/541.

As part of switching to the asynchronous reqwest client, this commit also reduces the time a lock on the queued logs is held. Instead of locking until all logs have been sent, the queued logs are cloned and the lock is released immediately.

This should fix the problem with no logs being captured on the hosted service right now.

### DIFF
--- a/graph/src/log/elastic.rs
+++ b/graph/src/log/elastic.rs
@@ -6,9 +6,11 @@ use std::sync::{Arc, Mutex};
 use std::time::Duration;
 
 use chrono::prelude::{SecondsFormat, Utc};
+use futures::future;
 use futures::{Future, Stream};
 use itertools;
 use reqwest;
+use reqwest::r#async::Client;
 use serde::ser::Serializer as SerdeSerializer;
 use serde_json;
 use slog::*;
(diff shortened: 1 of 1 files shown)


graph/src/log/elastic.rs
```diff
@@ -6,9 +6,11 @@ use std::sync::{Arc, Mutex};
 use std::time::Duration;
 
 use chrono::prelude::{SecondsFormat, Utc};
+use futures::future;
 use futures::{Future, Stream};
 use itertools;
 use reqwest;
+use reqwest::r#async::Client;
```
olibearo: Where can I read more about this language feature? What is it called?

ghost (approved):

graph/src/log/elastic.rs
Jannis: https://doc.rust-lang.org/edition-guide/rust-2018/module-system/raw-identifiers.html

PR #1051, opened 2019-07-09
Order all entity queries by id to get deterministic ordering of results

### LINKED ISSUE #1036: order is not deterministic, which is a problem for paging
A query that uses `orderBy` is not always giving deterministic results - if the field has the same values on different entities.

This is a problem for paging.

I was wondering if you ys could consider making the orderBy determistic after all. I.e. you could translate the graphql's `(orderBy: fieldName)` into postgres `ORDER BY fieldName, id` (or, alternatively, if you would consider adding support to order by multiple fields, so I can do `(orderBy: [balance, id])`

fordN:
Resolves #1036 

Entity queries are each appended with an order by id filter, so query results are always consistent. This is useful for when no order by is specified or the field being ordered by has duplicates. 

----

| GraphQL Sorting Filter | SQL Order By |
|:-:|:-|
| (OrderBy: AttributeName)  | ORDER BY 'data'->AttributeName->>'data' ASC, id ASC  |
| ()  | ORDER BY id ASC  |

### DIFF
--- a/store/postgres/src/entities.rs
+++ b/store/postgres/src/entities.rs
@@ -875,6 +875,7 @@ impl Table {
                             .sql(" NULLS LAST"),
                     );
                 }
+                query = query.then_order_by(public::entities::id.asc());
 
                 // Add range filter to query
                 if let Some(first) = first {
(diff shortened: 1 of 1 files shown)


jellegerbrandy:
:heart:

PR #1052, opened 2019-07-10
Include a "Content-type" header in http query response

### LINKED ISSUE #1032: Content-Type header is not present in response headers

Bug

GraphQL queries to the node do not return a `Content-Type `header. This breaks some graphQL clients like node's `graphql-request`.

This can be tested in the chrome web console by loading a demo page and testing a query.
Here is a specific example of a client that breaks with no content-type:

```
import { request } from 'graphql-request'

const GET_FILLS = `
  query getFills {
    fills {
      createdAt
      makerAssetFilledAmount
      takerAssetFilledAmount
      order {
        makerAssetData
        takerAssetData
      }
    }
  }
`

async function getFills() {
  await request('https://api.thegraph.com/subgraphs/name/dekz/zeroex', GET_FILLS)
}

getFills()
  .then(() => console.log('Everything worked OK'))
  .catch(e => console.log('Query returned an error'))
```

Expected behaviour is that `Content-Type: application/json` is present.

fordN:
Resolves #1032 

GraphQL query responses from subgraph endpoints use the  application/json content type.

### DIFF
--- a/server/http/src/response.rs
+++ b/server/http/src/response.rs
@@ -62,6 +62,7 @@ impl Future for GraphQLResponse {
             .header("Access-Control-Allow-Origin", "*")
             .header("Access-Control-Allow-Headers", "Content-Type")
             .header("Access-Control-Allow-Methods", "GET, OPTIONS, POST")
+            .header("Content-Type", "application/json")
             .body(Body::from(json))
             .unwrap();
         Ok(Async::Ready(response))


server/http/src/response.rs
```diff
@@ -62,6 +62,7 @@ impl Future for GraphQLResponse {
             .header("Access-Control-Allow-Origin", "*")
             .header("Access-Control-Allow-Headers", "Content-Type")
             .header("Access-Control-Allow-Methods", "GET, OPTIONS, POST")
+            .header("Content-type", "application/json")
```
Jannis: `Content-Type` please, otherwise this is good.

PR #1068, opened 2019-07-15
Allow creating data sources in `ipfs_map` and fix dynamic data sources missing events

### LINKED ISSUE #1035: Events seem to get skipped when templates contain block handlers
The [beetoken subgraph](https://github.com/Betoken/betoken-subgraph/commit/2701d2c2a24e01836650a9875625767630cb912d) uses templates that contain block handlers. Syncing it we can see that it handles one event in the main contract and spawns two data sources:

```
Jul 07 11:59:14.948 INFO Create data source, params: 0x1689dcfef3e695ac4cc1e5b7e77f9135f1d58a50, name: MiniMeToken, block_hash: 0x6fa3c8aef55e3c8b2096cb57350327c00920feecf26f06bca8280105870a90a4, block_number: 8064080, subgraph_id: QmZcMgr2CDU3vkyGABG3nNWSwpPkGxZZ1WXV77Bnkm74f5, component: SubgraphInstanceManager
Jul 07 11:59:14.978 INFO Create data source, params: 0x881a0bdf9514c116f4576f4fba263bf5397fca83, name: BetokenFund, block_hash: 0x6fa3c8aef55e3c8b2096cb57350327c00920feecf26f06bca8280105870a90a4, block_number: 8064080, subgraph_id: QmZcMgr2CDU3vkyGABG3nNWSwpPkGxZZ1WXV77Bnkm74f5, component: SubgraphInstanceManager
Jul 07 11:59:14.978 INFO Done processing Ethereum event, waiting_ms: 0, total_ms: 49370, handler: handleUpdatedFundAddress, signature: UpdatedFundAddress(address), block_hash: 0x6fa3c8aef55e3c8b2096cb57350327c00920feecf26f06bca8280105870a90a4, block_number: 8064080, subgraph_id: QmZcMgr2CDU3vkyGABG3nNWSwpPkGxZZ1WXV77Bnkm74f5, component: SubgraphInstanceManager
```

Contract `0x881a0bdf9514c116f4576f4fba263bf5397fca83` clearly has emitted a [bunch of events ](https://etherscan.io/txs?a=0x881a0bdf9514c116f4576f4fba263bf5397fca83&p=2)that are tracked by the data source. However the subgraph seems to not pick up on any of them, and the block handler is not fired on every block. Needs further investigation to find out what exactly is going on, but it looks like there are graph-node bugs involved.

### LINKED ISSUE #1067: Data sources created in `ipfs.mapJSON` seem to be being ignored
Bug

Data sources created from inside `ipfs.mapJSON` callback are not being processed.

**Steps to reproduce:**
- Create a subgraph with a data source that also define a dynamic data source (template).
- Add a mapping handler to that data source. That handler should contain a call to `ipfs.mapJSON` function with a given callback function.
- On the callback function, instantiate the dynamic data source.

Events from dynamic data source should be processed.

leodasvacas:
The first commit fixes data sources created in `ipfs_map` being ignored, which was a straightforward fix.

The second fixes a more subtle and more general issue, that we'd only restart the block stream to pick up new data sources when receiving the _next_ block in the original data source. This could cause missed events or even the stream to never be restarted because there was only one event in the original data source.

Fixes #1067. Fixes #1035.

### DIFF
--- a/core/src/subgraph/instance_manager.rs
+++ b/core/src/subgraph/instance_manager.rs
@@ -2,7 +2,6 @@ use futures::future::{loop_fn, Loop};
 use futures::sync::mpsc::{channel, Receiver, Sender};
 use std::collections::HashMap;
 use std::ops::Deref;
-use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::RwLock;
 use uuid::Uuid;
 
@@ -278,7 +277,6 @@ where
     let id_for_err = ctx.inputs.deployment_id.clone();
     let store_for_err = ctx.inputs.store.clone();
     let logger_for_err = logger.clone();
-    let logger_for_restart_check = logger.clone();
     let logger_for_block_stream_errors = logger.clone();
 
     let block_stream_canceler = CancelGuard::new();
@@ -307,28 +305,14 @@ where
 
     debug!(logger, "Starting block stream");
 
-    // Flag that indicates when the subgraph needs to be
-    // restarted due to new dynamic data sources being added
-    let needs_restart_set = Arc::new(AtomicBool::new(false));
-
-    // Handle to the flag for checking it before every block
-    let needs_restart_check = needs_restart_set.clone();
+    // The processing stream may be end due to an error or for restarting to
+    // account for new data sources.
+    enum StreamEnd<B: BlockStreamBuilder, S: Store + ChainStore, T: RuntimeHostBuilder> {
+        Error(CancelableError<Error>),
+        NeedsRestart(IndexingContext<B, S, T>),
+    }
 
     block_stream
-        // Take blocks from the stream as long as no dynamic data sources
-        // have been created. Once that has happened, we need to restart
-        // the stream to include blocks for these data sources as well.
-        .take_while(move |_| {
-            if needs_restart_check.load(Ordering::SeqCst) {
-                debug!(
-                    logger_for_restart_check,
-                    "New data sources added, restart the subgraph",
-                );
-                Ok(false)
-            } else {
-                Ok(true)
-            }
-        })
         // Log and drop the errors from the block_stream
         // The block stream will continue attempting to produce blocks
         .then(move |result| match result {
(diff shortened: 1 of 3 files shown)


runtime/wasm/src/module/mod.rs
```diff
@@ -647,7 +645,16 @@ where
                         "entity_operations" => ops.len(),
                         "time" => format!("{}ms", start_time.elapsed().as_millis())
                     );
-                    self.ctx.state.entity_operations.extend(ops);
+                    for op in ops {
```
Jannis: I think it's worth changing the variable name from `ops` to `output_states` or something that is more accurate.

core/src/subgraph/instance_manager.rs
```diff
@@ -307,28 +305,14 @@ where
...
-    let needs_restart_check = needs_restart_set.clone();
+    // The processing stream may be end due to an error or for restarting to
+    // account for new data sources.
+    enum StreamEnd<B: BlockStreamBuilder, S: Store + ChainStore, T: RuntimeHostBuilder> {
+        Error(CancelableError<Error>),
+        NeedsRestart(IndexingContext<B, S, T>),
+    }
 
     block_stream
-        // Take blocks from the stream as long as no dynamic data sources
```
Jannis: Nice! I love to see the flag to disappear.

runtime/wasm/src/host_exports.rs
```diff
@@ -377,7 +377,7 @@ where
         callback: &str,
         user_data: store::Value,
         flags: Vec<String>,
-    ) -> Result<Vec<EntityOperation>, HostExportError<impl ExportError>> {
```
Jannis: I think the comment above this method could use updating now because it only mentions entity operations.

Jannis:
Good catch btw, the cause of that bug was everything but obvious!

leodasvacas:
@Jannis Thanks for the feedback, review addressed.

PR #1069, opened 2019-07-16
Increase NodeId max length

ghost:
An increase in the max length of `NodeId` allows us to have more descriptive k8s pod names.


fordN (approved):
👍

Issue #1043, opened 2019-07-08
Make entity notifications robust


Jannis:

leodasvacas:
@olibearo, @fordN and me met to discuss what sort of measures can be taken here, at different time frames. Pinging @lutter since this is db-related.

In the short-term, the question is how should the graph-node respond if it detects that notifications may have been missed, be it due to the timed cleanup or an error connecting to the db. Here we came up with two ideas:

1. The node should crash. It would then be restarted and query the db source of truth again. This works ok for internal subscriptions (such as subgraph assignments), but for user subscriptions we're relying on the graphql client re-establishing the connection.

2. The node has a redundant polling system, that's less responsive but picks up what might be missed by notifications. This is would be a robust way to deal with the timed cleanup case, but is probably more code complexity than crashing, and is a performance cost when everything is going ok.

I proposed 1 and @olibearo proposed 2. @fordN feels the possibility of still missing a notification with the larger timeout is very unlikely. it'd be good to get more opinions here.

A mid-term goal: Switch to an tokio-postgres so we handle notifications in an async way. Tracked in #1053.

A Long-term solution is to split up the firehose of notifications, we'd have more fine-grained postgres channels, instead of a single channel for everything, and keep something like a `(channel_id, subcription_topic, time_to_live)` tuple in the DB. These would be created by subscriptions, and indexing nodes would query which channels they should send the updates to.

The even longer term question is what concept of subscriptions will exist in the hybrid network and its read interface, and how that will be supported. It will likely build on the lower-level parts of our existing subscription system, so it seems worth maintaining that even if its future is uncertain.

lutter:
Looking at this in some more detail, the only time we can possibly miss notifications is when they have to spill into the `large_notifications` table, i.e. when the message itself is pretty big (~ 7kB) This only happens if a subgraph touches a lot of entities in one block and is relatively rare, and missing notifications will manifest as GraphQL subscriptions not updating when they should have.

The bigger problem we had was that a missed large notification would panic the notification listener, which would then result in a node missing all subsequent notifications, small and large. To me, the way to make this more robust is
* make sure the notification listener can never panic (I'll post a PR for that in a minute)
[...]

PR #1070, opened 2019-07-18
store/postgres: Make sure missing notification does not panic listener

lutter:
When the payload for a large notification was cleaned up too early and went
missing, the notification listener would panic because we
postgres::row::Rows::get panics when we try to access the nonexistent row
in the result.

We now check that the result is not empty before accessing values to avoid
that panic.

This partly addresses https://github.com/graphprotocol/graph-node/issues/1043

### DIFF
--- a/store/postgres/src/notification_listener.rs
+++ b/store/postgres/src/notification_listener.rs
@@ -245,8 +245,20 @@ impl JsonNotification {
                         "SELECT payload FROM large_notifications WHERE id = $1",
                         &[&(payload_id as i32)],
                     )
-                    .map_err(|_| format_err!("No payload found for notification {}", payload_id))?;
+                    .map_err(|e| {
+                        format_err!(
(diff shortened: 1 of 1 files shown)


store/postgres/src/notification_listener.rs
```diff
@@ -245,8 +245,20 @@ impl JsonNotification {
...
-
+                    .map_err(|e| {
+                        format_err!(
+                            "Error retrieving payload for notification {}: {}",
+                            payload_id,
+                            e
+                        )
+                    })?;
+
+                if payload_rows.is_empty() || payload_rows.get(0).is_empty() {
```
leodasvacas: In what case would `payload_rows.get(0).is_empty()` be true?

lutter: I thought the same when I wrote that: I don't think it could ever happen, but out of an overabundance of caution, I covered that case, too.

PR #1071, opened 2019-07-22
server/json-rpc: Add dbg! in reassign handler

olibearo:


fordN:
The logging being added here appears redundant since we are logging the params in the reassign_handler:  https://github.com/graphprotocol/graph-node/blob/master/server/json-rpc/src/lib.rs#L151 
Is this still something you were planning on merging @olibearo?

olibearo:
Nope, closing

PR #1078, opened 2019-07-23
Remove public entities table

lutter:
This pull request does two things:

1. Remove the `public.entities` and `public.entity_history` table, together with all the metadata for the subgraphs that were stored there; we do keep the entries in `event_meta_data` for these entities since removing that causes a lot of headaches in hosted
2. Remove any Rust code related to storing entities in `public.entities`

I've tested this with a recent dump of the production database, and everything seems to be ok.

### DIFF
--- a/store/postgres/migrations/2019-07-20-195916_remove_subgraph/up.sql
+++ b/store/postgres/migrations/2019-07-20-195916_remove_subgraph/up.sql
@@ -1,0 +1,121 @@
+-- List all metadata entities for a set of subgraph deployments.
+-- The entities are returned as a table of (id, entity) pairs, in no particular
+-- order. The list contains all the entities that are reachable through
+-- associations from a SubgraphDeployment, and include the SubgraphDeployment
+-- itself.
+create or replace function
+  list_deployment_entities(ids varchar[])
+returns table(id varchar, entity varchar) as
+$$
+begin
+  create temp table tmp_ids
+      on commit drop as
+  select * from unnest(ids) as id;
+  create index tmp_ids_id on tmp_ids(id);
+
+  -- Put the ids of all dynamic data sources for the
+  -- subgraph deployments in ids into tmp_dds
+  create temp table tmp_dds
+      on commit drop as
+  select dds.id
+    from subgraphs.entities dds
+   where dds.entity = 'DynamicEthereumContractDataSource'
+     and dds.data->'deployment'->>'data' = any(ids);
+   create index tmp_dds_id on tmp_dds(id);
+
+   -- The metadata entities for a subgraph deployment consist of two
+   -- distinct groups of entities:
+   --   * static metadata, whose id starts with the id of the
+   --     subgraph deployment. Subgraph deployment ids are 46 chars long
+   --   * metadata for dynamic data sources, whose ids start with
+   --     the id of the dynamic data source. Dynamic data source ids are
+   --     40 chars long
+   return query
+   select e.id, e.entity
+     from subgraphs.entities e, tmp_ids
+    where substring(e.id, 1, 46) = tmp_ids.id
+   union all
+   select e.id, e.entity
+     from subgraphs.entities e, tmp_dds
+    where substring(e.id, 1, 40) = tmp_dds.id;
+end;
+$$ language plpgsql;
+
+-- Remove the metadata for the deployments with the given ids. None of the
+-- deployments must currently be assigned to a node for indexing.
+-- Returns the number of entities removed.
+-- Note that this function only removes the metadata from
+-- subgraphs.entities, but not the actual data for the subgraph
+create or replace function
+  remove_deployment_metadata(ids varchar[])
+returns int as
+$$
+declare
+  rows_deleted int;
+begin
+  if exists (select 1
+               from subgraphs.entities
+              where entity = 'SubgraphDeploymentAssignment'
+                and id = any(ids)) then
+     raise 'Can not remove assigned subgraph. One of % is currently assigned',
+           ids;
+  end if;
+
+  drop table if exists tmp_deployments;
+  create temp table tmp_deployments
+      on commit drop as
+  select * from list_deployment_entities(ids);
+
+  delete from subgraphs.entities as e
+  using tmp_deployments as md
+  where e.entity = md.entity and e.id = md.id;
+  GET DIAGNOSTICS rows_deleted = ROW_COUNT;
+  return rows_deleted;
+end;
+$$ language plpgsql;
+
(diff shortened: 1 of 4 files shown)


store/postgres/migrations/2019-07-20-195916_remove_subgraph/up.sql
```diff
@@ -0,0 +1,122 @@
...
+   --     the id of the dynamic data source. Dynamic data source ids are
+   --     40 chars long
+   return query
+   select e.id, e.entity
+     from subgraphs.entities e, tmp_ids
+    where substring(e.id, 1, 46) = tmp_ids.id
+   union all
+   select e.id, e.entity
+     from subgraphs.entities e, tmp_dds
+    where substring(e.id, 1, 40) = tmp_dds.id;
```
leoyvens: Clever to use the id prefix on all entities.

store/postgres/migrations/2019-07-20-195916_remove_subgraph/up.sql
```diff
@@ -0,0 +1,122 @@
...
+     raise 'Can not remove assigned subgraph. One of % is currently assigned',
+           ids;
+  end if;
+
+  drop table if exists tmp_deployments;
+  create temp table tmp_deployments
+      on commit drop as
+  select * from list_deployment_entities(ids);
+
+  -- TODO: Nuke entity_history and event_meta_data
```
leoyvens: Is this TODO supposed to be here?

store/postgres/migrations/2019-07-20-195916_remove_subgraph/up.sql
```diff
@@ -0,0 +1,122 @@
...
+begin
+  if exists (select 1
+               from subgraphs.entities
+              where entity = 'SubgraphDeploymentAssignment'
+                and id = any(ids)) then
+     raise 'Can not remove assigned subgraph. One of % is currently assigned',
+           ids;
+  end if;
+
+  drop table if exists tmp_deployments;
```
leoyvens: Why have this drop for `tmp_deployments` but not for `tmp_ids` and `tmp_dds`?

store/postgres/src/entities.rs
```diff
@@ -20,16 +20,16 @@
...
 use diesel::pg::{Pg, PgConnection};
-use diesel::sql_types::{Bool, Integer, Jsonb, Nullable, Text};
+use diesel::sql_types::{Integer, Jsonb, Nullable, Text};
 use diesel::BoolExpressionMethods;
 use diesel::Connection as _;
 use diesel::ExpressionMethods;
-use diesel::{debug_query, insert_into};
 use diesel::{OptionalExtension, QueryDsl, RunQueryDsl};
-use diesel_dynamic_schema::{schema, Column, Table as DynamicTable};
+use diesel_dynamic_schema::{schema as dds_schema, Column, Table as DynamicTable};
```
leoyvens: This PR uses `dds` to mean both "dynamic data source" and "diesel dynamic schema", this name is only used once, I don't mind not importing and writing `diesel_dynamic_schema::schema`.

store/postgres/src/entities.rs
```diff
@@ -1230,81 +966,48 @@ impl Table {
...
-                Ok(ents
-                    .table
-                    .select(ents.entity.clone())
-                    .filter(ents.entity.eq(any(entities)))
-                    .filter(ents.id.eq(entity_id))
-                    .first(conn)
-                    .optional()?)
-            }
-        }
+        let ents = self.clone();
```
leoyvens: image

store/postgres/src/subgraphs.graphql
```diff
@@ -89,16 +89,16 @@ type EthereumContractAbi @entity {
...
+    filter: EthereumBlockHandlerFilterEntity
 }
 
-type EthereumBlockFilter @entity {
+type EthereumBlockHandlerFilterEntity @entity {
      kind: String!
 }
 
-type EthereumCallHandler @entity {
+type EthereumCallHandlerEntity @entity {
```
leoyvens: Why rename these types? I think in the code we put 'Entity' as a suffix to all these types, I don't see a need to change the convention. If we do want to rename, there are more types that need renaming such as `EthereumContractAbi`.

store/postgres/migrations/2019-07-20-195916_remove_subgraph/up.sql
lutter: I started with carefully chasing pointers through the metadata, but that turned out to be incredibly slow. This works much faster.

store/postgres/migrations/2019-07-20-195916_remove_subgraph/up.sql
lutter: No, that was left in by accident. Removing it.

store/postgres/migrations/2019-07-20-195916_remove_subgraph/up.sql
lutter: IIRC, I put that in when I was testing it. I am a little loathe to remove it since I'd need to retest, and it won't cause any harm there.

store/postgres/src/entities.rs
lutter: Good point. Changing that.

store/postgres/src/subgraphs.graphql
lutter: It's not a renaming, it's reflecting what's in the database. That came from when I was chasing pointers. Here are the different `entity` values we use in the database:
```
                  entity                  
------------------------------------------
 DynamicEthereumContractDataSource
 EthereumBlockHandlerEntity
 EthereumBlockHandlerFilterEntity
 EthereumCallHandlerEntity
 EthereumContractAbi
 EthereumContractDataSource
 EthereumContractDataSourceTemplate
 EthereumContractDataSourceTemplateSource
 EthereumContractEventHandler
 EthereumContractMapping
 EthereumContractSource
 Subgraph
 SubgraphDeployment
 SubgraphDeploymentAssignment
 SubgraphManifest
 SubgraphVersion
```

lutter:
Updated with review comments and rebased to latest master.

store/postgres/src/subgraphs.graphql
leoyvens: Ah alright then.

leoyvens:
It's nice to keep the commit history clean, however when reviewing it's a bit hard to tell what changed since the last review. So I'd suggest making new commits during the review process, and if cleanup is desired do it after the changes are fully approved.

Issue #968, opened 2019-05-28
Make number of DB connections in the pool configurable through an env var


leoyvens:
So deployments can increase it when dealing with more than 10 subgraphs.

lutter:
Looks like this is done.

PR #1080, opened 2019-07-25
Add ability to set store connection pool size

ghost:
Addresses https://github.com/graphprotocol/graph-node/issues/968

The Store's connection pool size is defaulted to 10.

### DIFF
--- a/node/src/main.rs
+++ b/node/src/main.rs
@@ -349,7 +349,9 @@ fn async_main() -> impl Future<Item = (), Error = ()> + Send + 'static {
         .expect("invalid --store-connection-pool-size/STORE_CONNECTION_POOL_SIZE value");
 
     // Minimum of two connections needed for the pool in order for the Store to bootstrap
-    assert!(store_conn_pool_size > 1);
+    if store_conn_pool_size <= 1 {
+        panic!("--store-connection-pool-size/STORE_CONNECTION_POOL_SIZE must be > 1")
+    }
 
     // Set up Sentry, with release tracking and panic handling;
     // fall back to an empty URL, which will result in no errors being reported
--- a/node/src/main.rs
+++ b/node/src/main.rs
@@ -531,6 +531,8 @@ fn async_main() -> impl Future<Item = (), Error = ()> + Send + 'static {
         // database.
         assert!(*ANCESTOR_COUNT >= *REORG_THRESHOLD);
 
+        info!(logger, "Starting block ingestor");
+
         // Create Ethereum block ingestors and spawn a thread to run each
         eth_adapters.iter().for_each(|(network_name, eth_adapter)| {
             let block_ingestor = graph_datasource_ethereum::BlockIngestor::new(
--- a/node/src/main.rs
+++ b/node/src/main.rs
@@ -341,7 +341,7 @@ fn async_main() -> impl Future<Item = (), Error = ()> + Send + 'static {
         .parse()
         .expect("Invalid disable block ingestor value");
 
-    // Obtain MAX_CONN_STORE_POOL setting
+    // Obtain STORE_CONNECTION_POOL_SIZE setting
     let store_conn_pool_size = matches
         .value_of("store-connection-pool-size")
         .unwrap()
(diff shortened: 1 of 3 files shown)


fordN (commented):
👍

node/src/main.rs
```diff
@@ -263,6 +263,22 @@ fn async_main() -> impl Future<Item = (), Error = ()> + Send + 'static {
...
+                .default_value("false")
+                .help("Ensures that the block ingestor component does not execute"),
+        )
+        .arg(
+            Arg::with_name("store-connection-pool-size")
+                .long("store-connection-pool-size")
+                .value_name("STORE_CONNECTION_POOL_SIZE")
+                .default_value("10")
+                .env("STORE_CONNECTION_POOL_SIZE")
+                .help("Limits the maximum number of connections in the Store's connection pool"),
```
Jannis: `Store's` -> `store's`

Jannis: `maximum number` -> `number`

node/src/main.rs
```diff
@@ -302,12 +318,12 @@ fn async_main() -> impl Future<Item = (), Error = ()> + Send + 'static {
         .value_of("http-port")
         .unwrap()
         .parse()
-        .expect("invalid GraphQL HTTP server port");
+        .expect("Invalid GraphQL HTTP server port");
```
Jannis: Our convention is for `expect` messages to start with lower-case letters. According to @leoyvens it is more idiomatic.

node/src/main.rs
```diff
@@ -302,12 +318,12 @@ fn async_main() -> impl Future<Item = (), Error = ()> + Send + 'static {
...
     let ws_port = matches
         .value_of("ws-port")
         .unwrap()
         .parse()
-        .expect("invalid GraphQL WebSocket server port");
+        .expect("Invalid GraphQL WebSocket server port");
```
Jannis: Same here

node/src/main.rs
```diff
@@ -318,6 +334,20 @@ fn async_main() -> impl Future<Item = (), Error = ()> + Send + 'static {
 
     debug!(logger, "Setting up Sentry");
 
+    // Obtain DISABLE_BLOCK_INGESTOR setting
+    let disable_block_ingestor: bool = matches
+        .value_of("disable-block-ingestor")
+        .unwrap()
+        .parse()
+        .expect("Invalid disable block ingestor value");
```
Jannis: And here. Also, I'd make this message
```
invalid --disable-block-ingestor/DISABLE_BLOCK_INGESTOR value
```
or something like that to tie it to the flag/var.

node/src/main.rs
```diff
@@ -318,6 +334,20 @@ fn async_main() -> impl Future<Item = (), Error = ()> + Send + 'static {
...
+
+    // Obtain STORE_CONNECTION_POOL_SIZE setting
+    let store_conn_pool_size = matches
+        .value_of("store-connection-pool-size")
+        .unwrap()
+        .parse()
+        .expect("Invalid store connection pool size value");
```
Jannis: Same here

Jannis (changes requested):

node/src/main.rs
```diff
@@ -318,6 +334,23 @@ fn async_main() -> impl Future<Item = (), Error = ()> + Send + 'static {
...
+
+    // Obtain STORE_CONNECTION_POOL_SIZE setting
+    let store_conn_pool_size = matches
+        .value_of("store-connection-pool-size")
+        .unwrap()
+        .parse()
+        .expect("invalid --store-connection-pool-size/STORE_CONNECTION_POOL_SIZE value");
+
+    // Minimum of two connections needed for the pool in order for the Store to bootstrap
+    assert!(store_conn_pool_size > 1);
```
Jannis: This will generate a pretty ugly error. Also: Are assertions still being executed in production builds?

How about this?
```rust
if (store_conn_pool_size < 1) {
  panic!("--store-connection-pool-size/STORE_CONNECTION_POOL_SIZE must be > 1");
}
```

Jannis (changes requested):

Jannis (approved):

Issue #1023, opened 2019-07-01
Top-level data source templates


Jannis:

leodasvacas:
## Summary

Based on the discussion on https://github.com/graphprotocol/support/issues/19, and https://github.com/graphprotocol/graph-node/issues/1017, we want to move data source templates from being owned by a data source to being its own top-level field in the manifest, like this:

```
dataSources:
  - ...
templates:
  - ...
```

## Requirements

The new requirement being covered is that any data source should be able to instantiate templates declared in the `template` section, otherwise this should behave exactly like templates do today.

We need to decide on a migration path.

For graph-node we need to maitain backwards-compatibility, I suggest we do this by supporting _either_ the old form or the new form, but not a mix of both in the same manifest. We should log a deprecation warning when a subgraph uses the old form.

For graph-cli, a migration might not be a good idea, because if we simply move the nested data sources to the top level field, that might create an invalid manifest because of name conflicts. Also we haven't yet figured out how to keep yaml comments on complex migrations. To aviod a migration, we can either error or warn (which one should it be?) if the manifest uses the old form.

## Proposed Implementation

graph-cli: Implement the first step of the migration path we decide on, accept and validate top-level templates. Codegen for top-level templates. Forbid mixing the old and new forms.

graph-node: Parse the new form. Detect whether the subgraph uses the old or the new form, error if it mixes both. Make sure all data sources can access all top-level templates. Make the code as generic as possible over both forms. Add the top-level templates to the subgraph of subgraphs.

graph-ts: No change.

## Proposed Documentation Updates

- Update the subgraph manifest reference in https://github.com/graphprotocol/graph-node/blob/master/docs/subgraph-manifest.md by marking the deprecated field as such and documenting the new field.
- Update the data source templates section of the docs https://thegraph.com/docs/define-a-subgraph#data-source-templates.

## Proposed Tests / Acceptance Criteria

- Validation tests on graph-cli.
- Manual testing of graph-node functionality.

## Tasks

Task list so far:

- [ ] (1d) graph-cli: Accept, validate and do codegen for templates in the top level field.
- [ ] (0.5d) graph-cli: Forbid mixing the new and old form of templates.
- [ ] (0.5d) graph-node: Parse top-level templates.
- [ ] (1d) graph-node: Add an enum to support both forms, detect and error if they are mixed, log warning on use of the old form.
- [ ] (0.5d) graph-node: Make the runtime take a list of templates instead of directly taking `DataSource`.
[...]

PR #1085, opened 2019-07-29
Top level templates

leoyvens:
See https://github.com/graphprotocol/graph-node/issues/1023. Both formats are supported, but not a mix of both in the same subgraph. If a subgraph tries to mix both, only the top-level templates are considered. The `RuntimeHostBuilder` is given the data and the logic to make the decision of which template format is in use. See commits for more details.

A current limitation is that a data source cannot spawn another in the same block that it was created, for now just detect and error in this case, but this should be fixed in a follow up. See https://github.com/graphprotocol/graph-node/issues/1105.

graph-cli PR https://github.com/graphprotocol/graph-cli/pull/319

### DIFF
--- a/core/src/subgraph/instance_manager.rs
+++ b/core/src/subgraph/instance_manager.rs
@@ -507,7 +507,8 @@ where
                     true => Err(err_msg(
                         "A dynamic data source, in the same block that it was created,
                         attempted to create another dynamic data source.
-                        Let us know in graph-node issue #1105 that you hit this.",
+                        To let us know that you are affected by this bug, please comment in
+                        https://github.com/graphprotocol/graph-node/issues/1105",
                     )
                     .into()),
                 }
--- a/core/src/subgraph/instance_manager.rs
+++ b/core/src/subgraph/instance_manager.rs
@@ -21,7 +21,6 @@ struct IndexingInputs<B, S, T> {
     pub store: Arc<S>,
     pub stream_builder: B,
     pub host_builder: T,
-    pub templates: Vec<(String, DataSourceTemplate)>,
     pub include_calls_in_blocks: bool,
 }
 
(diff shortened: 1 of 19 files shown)


core/src/subgraph/instance_manager.rs
```diff
@@ -508,7 +501,17 @@ where
...
                 triggers,
             )
-            .map(move |block_state| (ctx, block_state, data_sources, runtime_hosts))
+            .and_then(move |block_state| {
+                match block_state.created_data_sources.len() > created_ds_count {
+                    false => Ok((ctx, block_state, data_sources, runtime_hosts)),
+                    true => Err(err_msg(
+                        "A dynamic data source, in the same block that it was created,
+                        attempted to create another dynamic data source.
+                        Let us know in graph-node issue #1105 that you hit this.",
```
Jannis: This is a great idea. Can we replace `Let us know in graph-node issue #1105 that you hit this` with `Let us know if you hit this on https://github.com/graphprotocol/graph-node/issues/1105`? That way people can click on the link in their terminals if they run their own Graph node.

graph/src/data/subgraph/mod.rs
```diff
@@ -898,27 +900,51 @@ impl UnresolvedSubgraphManifest {
...
             data_sources,
+            templates,
         } = self;
 
+        match semver::Version::parse(&spec_version) {
+            Ok(ref ver) if *ver <= semver::Version::new(0, 0, 2) => {}
+            _ => {
+                return Box::new(future::err(format_err!(
+                    "This Graph Node only supports manifest spec versions <= 0.0.2,
+                but subgraph `{}` uses `{}`",
```
Jannis: Is the message formatting / indentation correct here? Looks like it'll print an error message with the second line way indented.

graph/src/data/subgraph/mod.rs
```diff
@@ -898,27 +900,51 @@ impl UnresolvedSubgraphManifest {
...
+                    id,
+                    spec_version
+                ))) as Box<dyn Future<Item = _, Error = _> + Send>;
+            }
+        }
+
         // resolve each data set
```
Jannis: I'm just noticing now that we're using `data set` and `data_set` in this method. Fancy changing that to `data source` / `data_source` (optional for this review)?

Jannis (changes requested):
Looks good, only a couple of small comments.

leoyvens:
@Jannis addressed your comments.

graph/src/data/subgraph/mod.rs
leoyvens: This should print as a single line with a single space in the line break, but I've aligned it nicely.

Jannis (approved):

core/src/subgraph/instance_manager.rs
```diff
@@ -508,7 +501,18 @@ where
...
             )
-            .map(move |block_state| (ctx, block_state, data_sources, runtime_hosts))
+            .and_then(move |block_state| {
+                match block_state.created_data_sources.len() > created_ds_count {
+                    false => Ok((ctx, block_state, data_sources, runtime_hosts)),
+                    true => Err(err_msg(
+                        "A dynamic data source, in the same block that it was created,
+                        attempted to create another dynamic data source.
+                        To let us know that you are affected by this bug, please comment in
+                        https://github.com/graphprotocol/graph-node/issues/1105",
```
fordN: Nice, I like the approach of pushing users to a specific issue if they encounter the problem.

fordN (approved):

Issue #1025, opened 2019-07-02
side chain RPC issue


vishpari:
if putting side chain RPC url its showing invalid responce
![Screen Shot 2019-07-02 at 1 51 30 PM](https://user-images.githubusercontent.com/13795288/60496706-ff305900-9cd0-11e9-9dc7-b8fb34e8edb6.png)

vishpari:
same issue with goerli RPC(https://rpc.goerli.mudit.blog) also

leodasvacas:
Could you try to reproduce this against the Infura goerli endpoint? This behaviour suggests that the RPC endpoint is returning an empty response, which is against the ethereum JSON-RPC spec.

Jannis:
We should double-check on our side, because the endpoints do seem to be valid:

```
$ curl https://rpc.goerli.mudit.blog/ -H "content-type: application/json" --data '{"jsonrpc":"2.0","method":"eth_getBlockByNumber","params":["latest", true],"id":1}'
[...]

PR #1089, opened 2019-07-30
Upgrade web3, ethabi and jsonrpc-http-server

leoyvens:
Along with the minor version of other dependencies.

The ethabi graph-patches was rebased to the latest upstream.

web3 was weird. In principle we can use the upstream web3, however it broke our json-rpc calls because it could not parse the `id` field of json-rpc responses when called from graph-node, which is super mysterious. I hacked around this by putting a [patch](https://github.com/graphprotocol/rust-web3/tree/leo/hack-id-deserialization) on web3 that sets the `id` field from responses to null. It's a hack but arguably permitted under the json-rpc spec and didn't cause any further issues, the json-rpc implementation we use doesn't seem check ids at all.

I tested that this works against existing data in the db.

This does not solve #1025, I'll continue the investigation now the dependencies are updated.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -96,7 +96,7 @@ dependencies = [
  "environment 0.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "failure 0.1.5 (registry+https://github.com/rust-lang/crates.io-index)",
  "failure_derive 0.1.5 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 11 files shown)


olibearo:
Thanks for doing this, looks good. Can we deploy this branch to staging to verify the `rust-web3` patch before merging?

ghost (approved):

leoyvens:
@olibearo Thanks for the review, yhea we should watch how it goes on the hosted service staging.

Issue #1088, opened 2019-07-30
Regression in block ingestor/stream between 0.12 and 0.13


leoyvens:
It was reported by @joeandrews from [AZTEC](https://github.com/AztecProtocol/AZTEC/tree/feat-extension) that when upgrading from 0.12 their ganache setup stopped working, the issue manifests as early as 0.13. Their setup is always run from a clean DB. With debug logs this error is observed in the newer versions:
```
Jul 30 14:25:49.436 DEBG Trying again after eth_getBlockByHash RPC call failed (attempt #14) with result Err(CallError(ErrorMessage { msg: "RPC error: Error { code: ServerError(-32000), message: "Key not found in database", data: Some(Object({"name": String("NotFoundError"),
``` 
I was not yet able to set up the project for local reproduction. But the error clearly indicates that we requested a block hash which was not present in the chain. Looking at the commits between 0.12 and 0.13 it seems very probable that this is the culprit https://github.com/graphprotocol/graph-node/commit/97f0ceb01daf25cdc528b5d266fc68f1eda34b9c, since it changed the core of the block ingestor. Ping @lutter who authored that.

PR #1094, opened 2019-08-02
store: Do not get stuck on genesis block's missing parent

lutter:
In Ganache setups, we pull in the genesis block as the only block in the
beginning. That block does not have a valid parent, and we must not include
it in our parent checking.

This chainge also adds tests for various issues we've had with
attempt_chain_head_update to avoid repeating the same bugs.

Fixes https://github.com/graphprotocol/graph-node/issues/1088

### DIFF
--- a/store/postgres/tests/chain_head.rs
+++ b/store/postgres/tests/chain_head.rs
@@ -1,0 +1,265 @@
+//! Test ChainStore implementation of Store, in particular, how
+//! the chain head pointer gets updated in various situations
+
+use diesel::prelude::*;
+use diesel::RunQueryDsl;
+use futures::future::{self, IntoFuture};
+use lazy_static::lazy_static;
+use std::fmt::Debug;
+use std::sync::Arc;
+
+use graph::components::store::ChainStore;
+use graph::serde_json;
+use graph_store_postgres::db_schema_for_tests as db_schema;
+use graph_store_postgres::Store as DieselStore;
+
+use test_store::*;
+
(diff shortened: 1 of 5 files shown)


lutter:
@leoyvens I'd appreciate if you could run this through the setup you used to repro the problem. I think the `genesis_only` unit test covers that case, but better to make sure. Also, if you can think of other weird situations we might encounter with what we store in the DB, I'd appreciate it and would add more unit tests.

store/postgres/tests/chain_head.rs
```diff
@@ -0,0 +1,290 @@
...
+use graph_store_postgres::Store as DieselStore;
+
+use test_store::*;
+
+// The ancestor count we use for chain head updates. We keep this very small
+// to make setting up the tests easier
+const ANCESTOR_COUNT: u64 = 3;
+
+// This is duplicated from `src/db_schema.rs` since there is no good way
+// to keep db_schema private, but allow accessing it from integration tests
```
leoyvens: maybe something like:
```
#[cfg(debug_assertions)]
pub mod pub_db_schema {
   use db_schema::*;
}

lutter: I added a commit to that effect - it's not perfect, in that `cargo test --release` will fail to compile, but I can't figure out a way to have `cargo` set a custom feature flag when it's compiling tests.

leoyvens: Cool, yhea it's a hack but I don't think anyone runs `test --release`.

PR #1099, opened 2019-08-02
Ingest blocks with no transactions

### LINKED ISSUE #1025: side chain RPC issue

leoyvens:
Resolves #1025. The issue that we sent an empty batch, resulting in an empty response which web3 then failed to parse so we considered that the block failed to ingest. I confirmed that Goerli blocks often do not contain transactions.

Filed https://github.com/tomusdrw/rust-web3/issues/244 to see if web3 can be more lenient here,
but it's an easy fix in our code. This is the only place where we batch requests.


ghost (approved):

leoyvens:
@olibearo I've added a comment to clarify the need for this.

PR #1104, opened 2019-08-06
Enforce input length for EthereumCall

ghost:
When generating `EthereumCall`s for call handlers, `graph node` needs to filter out `CALL`s with less than 4 bytes of input since the block stream assumes `EthereumCall`s will have at least 4 bytes of input specifying the method id.

The `CALL`s we now filter out are contract to contract value transfers.

### DIFF
--- a/graph/src/components/ethereum/types.rs
+++ b/graph/src/components/ethereum/types.rs
@@ -83,9 +83,11 @@ pub struct EthereumCall {
 
 impl EthereumCall {
     pub fn try_from_trace(trace: &Trace) -> Option<Self> {
-        if trace.result.is_none() || trace.error.is_some() {
+        // Filter out traces from operations which produced errors
+        if trace.error.is_some() {
             return None;
         }
+        // We are only interested in traces from CALLs
         let call = match &trace.action {
             // Contract to contract value transfers compile to the CALL opcode
             // and have no input. Call handlers are for triggering on explicit method calls right now.
--- a/graph/src/components/ethereum/types.rs
+++ b/graph/src/components/ethereum/types.rs
@@ -99,43 +99,17 @@ impl EthereumCall {
             Some(Res::Call(result)) => (result.output.clone(), result.gas_used),
             _ => return None,
         };
-        Self::new(
-            call.from,
-            call.to,
-            call.value,
-            gas_used,
-            call.input.clone(),
-            output,
-            trace.block_number,
-            trace.block_hash,
-            trace.transaction_hash,
-        )
-    }
 
-    pub fn new(
-        from: Address,
-        to: Address,
-        value: U256,
-        gas_used: U256,
-        input: Bytes,
-        output: Bytes,
-        block_number: u64,
-        block_hash: H256,
-        transaction_hash: Option<H256>,
-    ) -> Option<EthereumCall> {
-        if input.0.len() < 4 {
-            return None;
-        }
         Some(EthereumCall {
-            from,
-            to,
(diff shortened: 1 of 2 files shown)


graph/src/components/ethereum/types.rs
```diff
@@ -81,32 +81,59 @@ pub struct EthereumCall {
...
+        to: Address,
+        value: U256,
+        gas_used: U256,
+        input: Bytes,
+        output: Bytes,
+        block_number: u64,
+        block_hash: H256,
+        transaction_hash: Option<H256>,
+    ) -> Option<EthereumCall> {
+        if input.0.len() < 4 {
```
leoyvens: Could this be an `assert` since we check this at the only call site? In fact, if `new` is only used in `try_from_trace`, I don't see a need for having `new` at all.

graph/src/components/ethereum/types.rs
```diff
@@ -81,32 +81,59 @@ pub struct EthereumCall {
...
-            Action::Call(call) => (call.from, call.to, call.value, call.input.clone()),
-            _ => (
-                Address::zero(),
-                Address::zero(),
-                U256::zero(),
-                Bytes::from(vec![]),
-            ),
+impl EthereumCall {
+    pub fn try_from_trace(trace: &Trace) -> Option<Self> {
+        if trace.result.is_none() || trace.error.is_some() {
```
leoyvens: `trace.result.is_none()` is already checked in the `match` below. In what cases do we get errors? Should we justify why the error content is discarded?

leoyvens (changes requested):
Reviewed the code, did not try to repro myself. The `filter_map` is super elegant!

graph/src/components/ethereum/types.rs
```diff
@@ -81,32 +81,61 @@ pub struct EthereumCall {
...
-            Action::Call(call) => (call.from, call.to, call.value, call.input.clone()),
-            _ => (
-                Address::zero(),
-                Address::zero(),
-                U256::zero(),
-                Bytes::from(vec![]),
-            ),
+impl EthereumCall {
+    pub fn try_from_trace(trace: &Trace) -> Option<Self> {
+        // Filter out traces from operations which produced errors
```
leoyvens: I'd still like to know more here, in which conditions is this error produced? Are they determinstic? Could this cause us to miss a call trigger?

olibearo: Nope, is error exists on the Trace it's because it was an operation that failed.

graph/src/components/ethereum/types.rs
```diff
@@ -81,32 +81,62 @@ pub struct EthereumCall {
...
-            _ => (
-                Address::zero(),
-                Address::zero(),
-                U256::zero(),
-                Bytes::from(vec![]),
-            ),
+impl EthereumCall {
+    pub fn try_from_trace(trace: &Trace) -> Option<Self> {
+        // The parity-ethereum tracing api returns traces for operations which had execution errors.
+        // Filter errorful traces out, since call handlers should only run on successful CALLs.
```
leoyvens: Thanks, I'm happy now.

graph/src/components/ethereum/types.rs
olibearo: Removed

fordN:
Looks good! I ran a quick test and it successfully processed the block with contract to contract value transfers that was causing issues before.

leoyvens (approved):

PR #1108, opened 2019-08-07
Improve block cache robustness

leoyvens:
If we error when fetching a block from the cache, try to get it from the ethereum node.

This is important right now because #1089 lost this patch https://github.com/graphprotocol/rust-web3/pull/1 to our web3 fork, causing errors with graph nodes that have blocks cached from old versions of the ethereum node. This is a way to fix this without reviving the patch.


fordN (approved):
LGTM!

PR #1109, opened 2019-08-07
graph dependency cleanup

leoyvens:
First, this removes the dead `backtrace` dependency. Then I went on to clean up the use of `extern crate` in the graph crate.


Jannis:
@leoyvens I'm curious what made you move from `graph::web3` etc. to using `web3` directly in other crates? I vaguely recall you being in favor of re-exporting dependencies through `graph` so all downstream crates would be guaranteed to use the same versions.

leoyvens:
@Jannis In this case I didn't change any `Cargo.toml`s, that was just a consequence of exporting `web3` in the prelude.

Jannis:
@leoyvens Ooooooh! That wasn't obvious but it makes sense now 😁

Jannis (approved):

PR #1110, opened 2019-08-07
Optimize `blocks_with_calls` memory usage

leoyvens:
This could previously collect a vector sized in GBs, by filtering and extracting the block pointer
before collecting, we can optimize the memory usage by at least 10x.

@olibearo could you please double check this code is still correct and I didn't take out anything important?


leoyvens:
Actually the way I'm deduping here is still not optimal, I'll update this to fold into a set and then this will be optimal.

leoyvens:
Now we dedup before collecting, so this should very memory efficient.

PR #1118, opened 2019-08-15
Address warnings emitted by Rust 1.37

lutter:
The warnings are all instances of 'trait objects without an explicit `dyn`
are deprecated'


leoyvens:
That was fast, did you use `cargo fix`? This needs to be formatted.

lutter:
Just ran `cargo fmt` on it again; no other change.

No, I actually didn't know about `cargo fix`, and did it be patiently clicking around VSCode. Next time!

PR #1121, opened 2019-08-16
Add a new index node GraphQL server

Jannis:
Resolves #1121.

This adds a new GraphQL server specifically for indexing information. This server runs at port `8030` by default. Its first version includes an _indexing status_ feature, that allows to query the indexing status of all or a subset of deployed subgraphs:

For example, the query 
```graphql
{
  indexingStatuses {
    subgraph
    failed
    synced
    error
    chains {
      network
      ... on EthereumIndexingStatus {
        chainHeadBlock {
          number
          hash
        }
        latestBlock {
          number
          hash
        }
        earliestBlock {
          number
          hash
        }
      }
    }
  }
}
```

might return the following result:

```json
{
  "data": {
    "indexingStatuses": [
      {
        "chains": [
          {
            "chainHeadBlock": {
              "hash": "70a1737e45dcbe60d841e262a41d8718a5fd08093f949667d2a7259c94b7460c",
              "number": "6204937"
            },
            "earliestBlock": {
              "hash": "41941023680923e0fe4d74a34bdac8141f2540e3ae90623718e47d66d1ca4a2d",
              "number": "0"
            },
            "latestBlock": {
              "hash": "8d4d913fa0fcb75fd8ec8a9b2d79d5ce7377a25002baab979cf30f6a547d0b35",
              "number": "2300000"
            },
            "network": "ropsten"
          }
        ],
        "error": null,
        "failed": false,
        "subgraph": "QmTk7hoDDFuo7BVh4v4EUFHk8PdpcSNajfkMhGXttdZhjj",
        "synced": false
      }
    ]
  }
}
```


core/src/graphql/runner.rs
```diff
@@ -30,6 +30,11 @@ lazy_static! {
         .map(|s| u8::from_str(&s)
             .unwrap_or_else(|_| panic!("failed to parse env var GRAPH_GRAPHQL_MAX_DEPTH")))
         .unwrap_or(u8::max_value());
+    static ref GRAPHQL_MAX_FIRST: u32 = env::var("GRAPHQL_MAX_FIRST")
```
olibearo: Docs please!

lutter:
One general comment: do we want this to be on a new port, or just use a new URL on an existing port? Generally, exposing more ports just complicates setup.

olibearo:
Alternatively, another port makes it clearer to control how this server is exposed (i.e. only exposing it on the local instead of the public network)

olibearo:
Apollo injects the `__typename` into each request with default configuration, so the `__typename` field is requested on all entities in the GraphQL responses.

Jannis:
@lutter Yeah, I would like to keep this separate from the other servers. The index node server should contain all the APIs/endpoints required for index nodes. Think of it like Ethereum nodes exposing their JSON-RPC API at 8454 and the network communication at 30303.

@olibearo I've added the missing `__typename`, will add docs for the new env var next.

Jannis:
@olibearo Comments addressed! Would be great if you could re-run your local test to see if the `__typename` issue is fixed.

olibearo:
@Jannis I've pull down the latest and I'm still getting missing`__typename` errors from Apollo:

```GraphQL error: Null value resolved for non-null field `__typename```

Jannis:
@olibearo I think I've fixed it. Can you try again?

olibearo:
The `__typename` issue is resolved, thanks

ghost (approved):

Issue #1128, opened 2019-08-22
Runtime sends operations to store that contain redundant work


lutter:
When `Store::apply_entity_operations_with_conn` is called by the runtime, the vector of operations may contain multiple entries for the same entity, and may also contain `Set` operations that do not actually change the entity.

lutter:
[This gist](https://gist.github.com/lutter/94c6e13497b5d8bfff040af264d21e9f) contains an example of both multiple operations for the same entity in one block as well as setting entities without changing any of their attributes.

lutter:
The way to address this would be to change how the three store operations in mappings are implemented in the runtime. That store should be backed by two maps `EntityKey` -> `Entity` (or just `(entity_type, entity_id) -> Entity`), one map would be called `current`, and only ever contain entities as they exist in the database, and one would be called `updates`, which contains the entity as the user has modified it.

* `get`: look the entity up from the database and put it into `current`; if the entity does not exist in the db, put `None` into `current`
* `set`: merge the new entity into the entity that is already in `updates`; if there is no entry yet, implicitly perform a `get` first (we load the entity from the db already in `Store::apply_set_operation`, so we are just moving this work to a different place)
* `remove`: leave `None` in `updates`

When we go to transact the entity operations for a block, we can deduce whether we need to perform a `Remove` or a `Set` based on the entries in `current` and `updates`. If the entries in `current` and `updates` are identical, we do not produce any `EntityOperation` for that entity.

lutter:
This is addressed by PR #1159.

PR #1129, opened 2019-08-22
Support for time-travel queries with the relational schema

lutter:
This PR applies on top of #1122 and adds support for block reversion and time-travel queries.

When trying this with a real subgraph (dice2win), I noticed that there are a lot of entities with empty block ranges (indicating that the same entity was updated more than once in the store in one block) and also a lot of entities with no changes to their data, just fragmented into multiple versions, indicating that unchanged entities are written to the store.

Both of these issues will cause problems with relational schemas, as it will lead to an inflation of the number of entities that we store. Even without relational schemas, this causes unnecessary database load.

I added some code that checks in `apply_entity_operations_with_conn` whether entities are listed multiple times in the same vector of operations, or if they are written but unchanged. I opened #1128 to track that issue; this will need to be addressed before we can use time-travel query support.


lutter:
Added a change that deduplicates a list of entity operations to avoid accumulating a lot of redundant history. That patch isn't as nice as it could be as it struggles with the fact that we send both subgraph data and metadata changes through the same interface, and metadata changes can contain `EntityOperations` that we can not combine, especially `AbortUnless`. Longer term, it would be good to decouple the store interface for subgraph data and metadata to clean up how this works.

lutter:
Did a bunch of code cleanup, since there were quite a few things that could be tightened down after separating metadata from entity operations. I also pulled the `EntityCache` into the runtime, to cut down on unnecessary entity fetches.

lutter:
This is now superseded by PR #1159

Issue #1133, opened 2019-08-26
cannot index string with length more than 8191 bytes.


orenyodfat:
**Do you want to request a *feature* or report a *bug*?**
bug 
**What is the current behavior?**
cannot store and index string with length more than 8191 bytes.
**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem.**

**What is the expected behavior?**
unlimited string length store and index .

guess that this bug is because pr  https://github.com/graphprotocol/graph-node/pull/1014 which use btree index for strings.

orenyodfat:
we are indexing an ipfs data which is used as a dao proposal meta data (title + description).
this ipfs data can be big ( it currently has a limitation of 100k as ipfs data size set as env var at the graph node).
yes, it will be possible to annotate our GraphQL schema with a hint about that.(large string).

orenyodfat:
yes, I would like to be able to search by this data .
e.g gives me all proposals which had description that contain the word "word".

orenyodfat:
it also require us to update our client dapp.
as it is not backward compatible.

orenyodfat:
does @noindex annotation give the same behaviour as before the btree index on String update?

orenyodfat:
any how,  trying to store string > 8191 bytes should not crash the graph. it should warn about.

Jannis:
Here's what I propose:

We use a truncated btree index (let's say on the first 8k characters) for all `String` fields.
 
This means `_equals` and comparison filters, but also `_contains` filters will not be 100% accurate. `_equals` and `_contains` may not return all matches; `_equals` may also return long strings that don't match, if the value passed to `_equals` matches the 8k prefix. _All filters would still be accurate for fields where no value exceeds 8k._

Searching for a word or phrase in a string field is really fulltext search. We don't support that yet but it's on the roadmap: #300.

Jannis:
I'm copying my `@noindex` critique here from the PR:

> I'm afraid this isn't good enough. We simply don't know what size values for any field will have upfront. Subgraph developers don't know either. Anyone with enough ETH can send arbitrarily long strings to a contract. For this directive to protect subgraphs from failing, you'd have to add it to any String field in the schema.
>
> We'll have to find a solution that works for all strings without any special hints in the subgraph.

PR #1135, opened 2019-08-26
graph: Support a @noindex directive for fields

lutter:
When a field is marked with `@noindex`, we do not generate a database index
for it. This is needed since the BTree indexes we put on String attributes
by default introduce a size limit on the strings we can store.

Fixes https://github.com/graphprotocol/graph-node/issues/1133


Jannis:
I'm afraid this isn't good enough. We simply don't know what size values for any field will have upfront. Subgraph developers don't know either. Anyone with enough ETH can send arbitrarily long strings to a contract. For this directive to protect subgraphs from failing, you'd have to add it to any `String` field in the schema.

We'll have to find a solution that works for all strings without any special hints in the subgraph.

PR #1138, opened 2019-08-27
Index only a prefix of string attributes

lutter:
That avoids the problem described in issue #1133 

The changes should not lead to any observable change in behavior, except that when ordering by a string field, entries with very long values (> 2k characters) are only ordered by their prefix, not their full value.

### DIFF
--- a/store/postgres/src/entities.rs
+++ b/store/postgres/src/entities.rs
@@ -353,7 +353,7 @@ impl<'a> Connection<'a> {
         subgraph: &SubgraphDeploymentId,
         entity_types: Vec<String>,
         filter: Option<EntityFilter>,
-        order: Option<(String, &str, &str)>,
+        order: Option<(String, ValueType, &str, &str)>,
         first: Option<u32>,
         skip: u32,
     ) -> Result<Vec<(serde_json::Value, String)>, QueryExecutionError> {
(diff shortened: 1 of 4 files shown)


store/postgres/src/entities.rs
```diff
@@ -51,6 +51,11 @@ use crate::filter::{build_filter, store_filter};
 use crate::functions::set_config;
 use crate::jsonb::PgJsonbExpressionMethods as _;
 
+/// The size of string prefixes that we index. This should be large enough
+/// that we catch most strings, but small enough so that we can still insert
+/// it into a Postgres BTree index
+pub(crate) const STRING_PREFIX_SIZE: usize = 2048;
```
Jannis: If we're not sure, should we make this a `GRAPH_STORE_POSTGRES_STRING_INDEX_PREFIX_SIZE` (or something else that is prefixed with GRAPH_) environment variable?

Jannis: Ok

Jannis:
Any redeploy will generate the new index, so that's alright.

Jannis:
Ready to merge? 😄

Jannis:
Something's up with the go-ipfs server. All tests ran in the `push` build and passed.

PR #1140, opened 2019-08-28
Fix traces being requested when not necessary

leoyvens:
This fixes a bug which is easy to reproduce on master:
1. Deploy a subgraph with dynamic data sources but that doesn't use traces.
2. Let it spawn a dynamic data source.
3. Restart the graph-node, the subgraph will start requesting traces even though they're not required.

I believe this bug was introduced in #1087. This regresses indexing speed and may cause the subgraph to stop indexing if the Ethereum node doesn't support traces.

The root cause is that we use `Option<EthereumCallFilter>`, and we request traces if that's `Some`. When first deploying a subgraph that would be `None` so all good, but when restarting after #1087 that would be `Some` but with an empty filter, and the code that requests traces can't tell that `None` and an empty filter are the same thing, which is annoying.

I solved this the hard way by refactoring away the `Option` around most cases of `Option<Vec<_>>` and `Option<Ethereum*Filter>`. It is confusing to have `Option` around something that has its own concept of being empty, since `None` and `Some(empty)` are usually equivalent.

### DIFF
--- a/core/src/subgraph/instance.rs
+++ b/core/src/subgraph/instance.rs
@@ -35,7 +35,7 @@ where
     ) -> Result<Self, Error> {
         let manifest_id = manifest.id.clone();
         let network_name = manifest.network_name()?;
-        let templates = manifest.templates.unwrap_or_default();
+        let templates = manifest.templates;
 
         // Create a new runtime host for each data source in the subgraph manifest;
         // we use the same order here as in the subgraph manifest to make the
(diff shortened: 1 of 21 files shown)


fordN (approved):
Good catch!   Removing the options cleans things up nicely too.

leoyvens:
@fordN Thanks for the review!

PR #1154, opened 2019-09-03
Fix WASM memory allocation

leoyvens:
#1145 made the assumption that two consecutive `memory.allocate` calls would allocate contiguous memory. This is of course false if AS allocated any memory for itself between the calls. This would cause memory corruption on handlers that did a lot of things and needed to call `memory.allocate` more than once in `raw_new`. This was the case with the Betoken subgraph, I reproduced the error locally and verified that this PR fixes it.


Jannis:
@leoyvens Can we move that explanation into the commit message?

leoyvens:
@Jannis oops. It was originally in the commit message, I lost it somehow and then re-wrote in the PR description. Now it's merged, oh well.

PR #1175, opened 2019-09-09
store: Remove foreign key into event_meta_data

lutter:
The foreign key constraint from a subgraph's entity_history table into
event_meta_data requires a lock on subgraph creation that prevents subgraph
deployment during an autovacuum of event_meta_data.

We remove the foreign key constraint so that subgraph deploys can proceed
even when event_meta_data is being autovacuumed.


lutter:
This does not yet remove some of the code around dealing with this lock during subgraph deployments; once we have confirmed that this indeed avoids the problems we are seeing in hosted, we should do the code removal described in #1176

Issue #1161, opened 2019-09-04
Relational schema: handle GraphQL enums


lutter:
In the GraphQL schema, users can define enums. Those show up as strings in the `Entity`.

Change the relational schema code to accept enums.

Example in the [decentraland subgraph](https://github.com/graphprotocol/decentraland-subgraph/blob/master/schema.graphql#L54)

PR #1178, opened 2019-09-11
store: Support GraphQL enums with relational schema

lutter:
We create a Postgres enum for each GraphQL enum and use that as the type of
the column. That has the advantage that enums only take 4 bytes to store,
no matter how long their label.

Fixes https://github.com/graphprotocol/graph-node/issues/1161


store/postgres/src/relational.rs
```diff
@@ -123,6 +123,9 @@ pub struct Layout {
     /// database tables that contain entities implementing
     /// that interface
     pub interfaces: HashMap<String, Vec<Rc<Table>>>,
+    /// Enums defined in the schema and their possible values. The names
+    /// are the original GraphQL names
+    pub enums: HashMap<String, Vec<String>>,
```
leoyvens: Lets use a BTreeMap instead, otherwise the `create type` statements will be emitted in random order.

store/postgres/src/relational.rs
```diff
@@ -257,6 +264,17 @@ impl Layout {
...
         let mut out = String::new();
 
+        // Output enums first
+        for (name, values) in &self.enums {
+            let mut sep = "";
+            let name = SqlName::from(name.as_str());
+            write!(out, "create type {}.{}\n    as enum (", self.schema, name)?;
+            for value in values {
+                write!(out, "{}'{}'", sep, value)?;
+                sep = ", "
```
leoyvens: Could you use [join](https://doc.rust-lang.org/std/slice/trait.SliceConcatExt.html#tymethod.join) here?

store/postgres/src/relational.rs
```diff
@@ -541,14 +582,22 @@ impl Column {
...
     ///
     /// See the unit tests at the end of this file for the actual DDL that
     /// gets generated
     fn as_ddl(&self, out: &mut String) -> fmt::Result {
         write!(out, "    ")?;
-        write!(out, "{:20} {}", self.name, self.sql_type())?;
+        let type_name = match &self.enum_type {
+            Some(ref type_name) => type_name.as_str(),
+            None => self.sql_type(),
+        };
```
leoyvens: This code might indicate the abstraction is not quite right, I'd expect `Column::sql_type` to return the enum type since enums are also sql. Maybe enum types should be a variant of `ColumnType`?

store/postgres/src/relational.rs
```diff
@@ -767,6 +825,12 @@ mod tests {
             bigThing: Thing!
         }
 
+        enum Color {
+            yellow,
+            red,
+            BLUE
+        }
```
leoyvens: Can we have a test with two different enum types? That would catch the HashMap ordering thing.

store/postgres/src/relational_queries.rs
```diff
@@ -157,21 +157,29 @@ fn query_builder_error(msg: String) -> diesel::result::Error {
 }
 
 /// A `QueryValue` makes it possible to bind a `Value` into a SQL query
-/// where the needed SQL type is `ColumnType`
-struct QueryValue<'a>(&'a Value, ColumnType);
+/// using the metadata from Column
+struct QueryValue<'a>(&'a Value, &'a Column);
```
leoyvens: I suppose if enums where a `ColumnType` this wouldn't need to change.

store/postgres/src/relational.rs
lutter: I was trying to avoid allocating a `Vec` just so we can join it

store/postgres/src/relational.rs
lutter: Ok.

store/postgres/src/relational.rs
lutter: True; I tried it with a `ColumnType::Enum(_)` variant, and that does clean up the code quite a bit.

store/postgres/src/relational.rs
lutter: Ok

store/postgres/src/relational_queries.rs
lutter: You are right - it has to become a `&ColumnType` because `ColumnType` can't be `Copy` anymore, but yes, we don't need the whole column.

lutter:
I've added commit eb0c5172 which addresses the review comments; I'll squash these before merging.

store/postgres/src/relational.rs
```diff
@@ -652,17 +722,23 @@ impl Table {
             block_range = BLOCK_RANGE
         )?;
 
-        for (i, column) in self.columns.iter().enumerate() {
+        // Create indexes. Don't bother with enum types, since an index on a
+        // column that can take only a handful of values is pointless
```
leoyvens: This may be a controversial topic, but as we discussed when we considered not indexing boolean columns, I don't think it's pointless to index a column of low cardinality. It is pointless if the values are uniformly distributed, but if some values occur rarely than an index can be useful. To me, "index everything" still seems like a good strategy.

store/postgres/src/relational.rs
leoyvens: Nice!

store/postgres/src/relational.rs
lutter: I don't want to hold this PR up with this debate, so changed the code to index enum columns. But I would like to see a query plan from production where an index on a boolean column actually gets used, or at the very least some evidence from `pg_stat_user_indexes` that some boolean index has been used and we are not just slowing down writes unnecessarily with these indexes.

PR #1180, opened 2019-09-11
store: Quote table and column names when creating relational schema

lutter:
We make sure that the type and attribute names from GraphQL schemas are
safe to use in SQL, but users may use names that conflict with reserved
words in SQL; for example, a type 'User' will result in the creation of a
table 'user'. That causes a syntax error in Postgres, and we therefore need
to enclose these names in double-quotes during schema creation. Proper
queries do not have this problem since we have Diesel escape them as
identifiers already.

### DIFF
--- a/store/postgres/src/relational.rs
+++ b/store/postgres/src/relational.rs
@@ -48,6 +48,10 @@ impl SqlName {
         &self.0
     }
 
+    pub fn quoted(&self) -> String {
+        format!("\"{}\"", self.0)
+    }
+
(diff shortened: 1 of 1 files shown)


lutter:
Note that this PR applies on top of #1178

leoyvens:
Looks like the PR is not against master.

lutter:
@leoyvens It was after merging #1178; it was just still marked as against the temporary branch

leoyvens:
The diff is the same but it is necessary to change the base branch, right?

PR #1182, opened 2019-09-12
store: Work around a problem with how Diesel serializes BigDecimal

lutter:


lutter:
> This is unfourtunate, is it an upstream bug?

I think so; I haven't completely followed the logic inside Diesel around BigDecimal, but as this shows, it looks like they do not properly handle negative scale

lutter:
Rebased to latest master and switched base branch to master.

PR #1183, opened 2019-09-12
Correctly initialize deployments with the genesis block

Jannis:
This broke on master recently and caused subgraphs to jump straight to the chain head without indexing anything.


lutter (approved):
I could reproduce that without this, a newly deployed subgraph immediately jumps to the chain head, and with this, it starts indexing from genesis.

Issue #1177, opened 2019-09-10
Cache storage layout across transactions


lutter:
The `entities::Storage` enum contains data that we need to access entities in the store. So far, this data was quick to compute, and only cached for the duration of a transaction as it may change because of migrations. With the relational schema, there's quite a bit more computation that needs to happen to get a `entities::Storage::Relational` instance, even from the already cached GraphQL schema, computation that needs to happen on every query, and that is based on immutable data.

Instances of `entities::Storage` for which there are no pending migrations can be safely cached across transactions.

PR #1184, opened 2019-09-13
store: Cache Storage objects in the Store

lutter:
With relational schemas, we need a bit of computation to go from the
GraphQL schema to the relational Layout. If possible, we cache the Layout
for the lifetime of the Store

Fixes https://github.com/graphprotocol/graph-node/issues/1177

### DIFF
--- a/store/postgres/src/entities.rs
+++ b/store/postgres/src/entities.rs
@@ -31,11 +31,11 @@ use diesel::Connection as _;
 use diesel::ExpressionMethods;
 use diesel::{OptionalExtension, QueryDsl, RunQueryDsl};
 use inflector::cases::snakecase::to_snake_case;
-use std::cell::RefCell;
 use std::collections::hash_map::DefaultHasher;
 use std::collections::HashMap;
 use std::convert::TryInto;
 use std::hash::{Hash, Hasher};
+use std::sync::{Arc, Mutex};
 use std::time::Instant;
 
 use graph::data::schema::Schema as SubgraphSchema;
@@ -300,13 +300,21 @@ impl QueryableByName<Pg> for RawHistory {
     }
 }
 
+/// A cache for storage objects as constructing them takes a bit of
+/// computation. The cache lives as an attribute on the Store, but is managed
+/// solely from this module
+pub(crate) type StorageCache = Mutex<HashMap<SubgraphDeploymentId, Arc<Storage>>>;
+
+pub(crate) fn make_storage_cache() -> StorageCache {
+    Mutex::new(HashMap::new())
+}
+
 /// A connection into the database to handle entities which caches the
 /// layout to actual database tables. Instances of this struct must not be
 /// cached across transactions as there is no mechanism in place to notify
 /// other index nodes that a subgraph has been migrated
 pub(crate) struct Connection<'a> {
     pub conn: PooledConnection<ConnectionManager<PgConnection>>,
-    cache: RefCell<HashMap<SubgraphDeploymentId, Storage>>,
     store: &'a Store,
 }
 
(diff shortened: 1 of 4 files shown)


lutter:
Rebased to latest master and switched base branch to master

store/postgres/src/entities.rs
```diff
@@ -322,18 +332,35 @@ impl<'a> Connection<'a> {
...
         let mut cache = self.cache.borrow_mut();
 
-        match cache.get(subgraph) {
-            Some(storage) => Ok(storage.clone()),
-            None => {
-                let storage = Storage::new(&self.conn, subgraph, self.store)?;
-                cache.insert(subgraph.clone(), storage.clone());
-                Ok(storage)
-            }
+        if let Some(storage) = cache.get(subgraph) {
```
leoyvens: Is it not overkill to keep two caches, the connection and the store one?

store/postgres/src/entities.rs
```diff
@@ -1357,7 +1379,10 @@ impl Storage {
 /// Delete all entities. This function exists solely for integration tests
 /// and should never be called from any other code. Unfortunately, Rust makes
 /// it very hard to export items just for testing
-pub fn delete_all_entities_for_test_use_only(conn: &PgConnection) -> Result<usize, StoreError> {
+pub fn delete_all_entities_for_test_use_only(
```
leoyvens: Want to add `#[cfg(debug_assert)]` since we've been doing that in other test-only code?

store/postgres/src/entities.rs
```diff
@@ -649,9 +676,13 @@ impl<'a> Connection<'a> {
...
-            let storage = storage.migrate(&self.conn, logger, &schema)?;
-            let needs_migrating = storage.needs_migrating();
-            self.cache.borrow_mut().insert(subgraph.clone(), storage);
+            // Do the actual migration, and return an updated storage
+            // object, something like
+            //
+            // let storage = storage.migrate(&self.conn, logger, &schema)?;
+            // let needs_migrating = storage.needs_migrating();
+            // self.cache.borrow_mut().insert(subgraph.clone(), Arc::new(storage));
+            //
```
leoyvens: What's up with the commented out code?

store/postgres/src/entities.rs
lutter: There's two reasons why I did it this way:

1. In a txn with a lot of operations, we do not need to get the Mutex every time (the per-connection cache is lockless)
2. For storage objects that have a migration pending, we can not put them into the `Store` cache (cross txn), but would still like to avoid recomputing the `Layout` for each operation in one txn

store/postgres/src/entities.rs
lutter: The migration machinery was quite painful to get right, so I want to keep it around, even if it's not being used at the moment. This code would need more fixing up than I like to keep compiling as `storage.migrate` took ownership of `storage` and possibly returned a new one. That doesn't work with how we cache anymore and needs to be rethought.

leoyvens: You could put `unreachable` somewhere to indicate that it's ok for this code to not be working.

store/postgres/src/entities.rs
lutter: Ok; taking out the per-txn cache

store/postgres/src/entities.rs
lutter: Yep, added `#[cfg(debug_assertions)]` :grinning:

store/postgres/src/entities.rs
lutter: Did that and a bunch of `allow` to make rustc not complain

lutter:
@leoyvens Addressed all the review comments

store/postgres/src/entities.rs
leoyvens: Ah yhea that's what it's called

lutter:
Squashed review comments

PR #1185, opened 2019-09-13
Allow a data source to create another in the same block it was created

### LINKED ISSUE #1105: Allow a data source to create another in the same block it was created
This is a limitation in the initial implementation of top-level templates, because our algorithm is:
- Create data sources
- Process current block for those data sources

That algorithm should loop until no more data sources are created.

leoyvens:
Resolves #1105, this involved some refactoring to `instance_manager::process_block`.

This was tested by uncommenting this line on this subgraph https://github.com/graphprotocol/aragon-subgraph/blob/sistemico/canary/canary/src/mappings/apm.ts#L42. Previously it would print the expected error, now it spawns the data source and processes its events on the same block.

### DIFF
--- a/core/src/subgraph/instance_manager.rs
+++ b/core/src/subgraph/instance_manager.rs
@@ -519,23 +519,23 @@ where
             })
         })
     })
-    .and_then(move |(ctx, block_state, data_sources, runtime_hosts)| {
+    .and_then(move |(mut ctx, mut block_state, data_sources, hosts)| {
+        // If new data sources have been added, indicate that the subgraph
+        // needs to be restarted after this block.
+        let needs_restart = !data_sources.is_empty();
+
         // Add entity operations for the new data sources to the block state
-        // and add runtimes for the data sources to the subgraph instance
+        // and add runtimes for the data sources to the subgraph instance.
         future::result(
             persist_dynamic_data_sources(
                 logger3,
-                ctx,
-                block_state,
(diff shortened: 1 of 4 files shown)


core/src/subgraph/instance_manager.rs
```diff
@@ -456,83 +451,101 @@ where
...
+                    Ok(ok) => ok,
+                    Err(err) => return Box::new(future::err(err.into())),
+                };
+
+                // Reprocess the triggers from this block that match the new data sources
+
+                let block_with_calls = EthereumBlockWithCalls {
+                    ethereum_block: block.deref().clone(),
+                    calls: calls.clone(),
+                };
```
leoyvens: This code is wrong afaict, if the new data source contains call handlers for new addresses the call won't be contained here. But this is the way it was before so I will open an issue for fixing this.

olibearo: Looked into this and found that the calls will not be included in the block in the case where the data sources at the beginning of the block did not require them. The calls will be complete when they are included though.

ghost (commented):
LGTM, let's just fix the issue with the calls not being included if the data sources at the beginning of the block did not require them

ghost (approved):
Talked with @leoyvens, will merge now and fix the issue with calls not being populated during reprocessing separately

leoyvens:
Thanks for the review!

PR #1190, opened 2019-09-14
Improve separation of responsibilities between Store and entities::Connection

lutter:
The `entities::Connection` struct looked like an instance would be used with any number of subgraphs, when we really only ever use it for one subgraph plus the metadata subgraph. These changes make that separation clearer and also remove some of the ugly back-and-forth between the store and the connection.

### DIFF
--- a/store/postgres/src/entities.rs
+++ b/store/postgres/src/entities.rs
@@ -316,7 +316,7 @@ pub(crate) fn make_storage_cache() -> StorageCache {
 /// cached across transactions as there is no mechanism in place to notify
 /// other index nodes that a subgraph has been migrated
 pub(crate) struct Connection<'a> {
-    pub conn: PooledConnection<ConnectionManager<PgConnection>>,
+    conn: PooledConnection<ConnectionManager<PgConnection>>,
     store: &'a Store,
 }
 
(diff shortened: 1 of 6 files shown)


store/postgres/src/entities.rs
```diff
@@ -310,37 +320,44 @@ pub(crate) fn make_storage_cache() -> StorageCache {
...
     /// the Store. Storage objects with a pending migration can not be
     /// cached for longer than a transaction since they might change
     /// without us knowing
-    fn storage(&self, subgraph: &SubgraphDeploymentId) -> Result<Arc<Storage>, StoreError> {
-        if let Some(storage) = self.store.storage_cache.lock().unwrap().get(subgraph) {
+    pub(crate) fn storage(
+        conn: &PgConnection,
+        store: &Store,
+        subgraph: &SubgraphDeploymentId,
+    ) -> Result<Arc<Storage>, StoreError> {
```
leoyvens: I think this will now make more sense as a method on `Store`.

store/postgres/src/entities.rs
```diff
@@ -360,37 +400,37 @@ impl<'a> Connection<'a> {
...
 
     pub(crate) fn find(
         &self,
-        subgraph: &SubgraphDeploymentId,
         entity: &String,
         id: &String,
         block: BlockNumber,
     ) -> Result<Option<Entity>, StoreError> {
-        match &*self.storage(subgraph)? {
+        match &*self.storage {
```
leoyvens: `find` and `query`, if run against the metadata subgraph, would be running in a connection exclusive to it, correct?

lutter: Yes, in that case `self.storage` and `self.metadata` are the same thing (that happens for example for `create_schema`) And I just checked that a query against the subgraph of subgraphs actually returns data.

store/postgres/src/entities.rs
lutter: Good point. Commit c28ad232

lutter:
Rebased onto latest master

PR #1198, opened 2019-09-16
Add entity_count and query_price to the Indexing Statuses API

### LINKED ISSUE #1156: Implement pricing interface in the index node GraphQL API

fordN:
Resolves #1156

This PR adds to the surface area of the Indexing Statuses API, now including several additional fields: `entity_count` and `query_price`.  The new fields have also been included in the `SubgraphDeploymentEntity`.
I have created a new branch off master (`hybrid-network`), so we can keep the master branch free of hybrid network specific features.

### DIFF
--- a/graph/src/data/subgraph/schema.rs
+++ b/graph/src/data/subgraph/schema.rs
@@ -214,6 +214,8 @@ pub struct SubgraphDeploymentEntity {
     ethereum_head_block_hash: Option<H256>,
     ethereum_head_block_number: Option<u64>,
     total_ethereum_blocks_count: u64,
+    entity_count: u64,
+    query_price: u64,
 }
 
 impl TypedEntity for SubgraphDeploymentEntity {
@@ -228,6 +230,8 @@ impl SubgraphDeploymentEntity {
         synced: bool,
         earliest_ethereum_block: EthereumBlockPointer,
         latest_ethereum_block: Option<EthereumBlockPointer>,
+        entity_count: u64,
+        query_price: u64,
     ) -> Self {
         let latest_ethereum_block =
             latest_ethereum_block.unwrap_or(earliest_ethereum_block.clone());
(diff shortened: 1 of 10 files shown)


store/postgres/src/subgraphs.graphql
```diff
@@ -17,6 +17,7 @@ type SubgraphVersion @entity {
 type SubgraphDeployment @entity {
     id: ID! # Subgraph IPFS hash
     manifest: SubgraphManifest!
+    assignment: SubgraphDeploymentAssignment!
```
Jannis: This is a great addition.

graph/src/data/subgraph/schema.rs
```diff
@@ -214,6 +214,8 @@ pub struct SubgraphDeploymentEntity {
     ethereum_head_block_hash: Option<H256>,
     ethereum_head_block_number: Option<u64>,
     total_ethereum_blocks_count: u64,
+    entity_count: u64,
```
Jannis: Let's keep these out of of the main deployment entity. How about we introduce a separate entity, like with the deployment assignments: `SubgraphDeploymentPricing`?

store/postgres/src/subgraphs.graphql
Jannis: Does it need to be `@derivedFrom` or does this here already work?

server/index-node/src/schema.graphql
```diff
@@ -14,6 +14,8 @@ type SubgraphIndexingStatus {
   failed: Boolean!
   error: String
   chains: [ChainIndexingStatus!]!
+  entityCount: BigInt!
```
Jannis: I'd like for this to be separate from the `SubgraphIndexingStatus` type.

Can we introduce a separate `SubgraphPricing` type for this with a `subgraphPricing(subgraphs: [String!]!): [SubgraphPricing!]!` resolver?

server/index-node/src/resolver.rs
```diff
@@ -130,6 +134,17 @@ impl IndexingStatus {
             _ => Ok(None),
         }
     }
+    fn u64_from_value(value: &q::Value, key: &'static str) -> Result<u64, Error> {
```
Jannis: There's a `TryFromValue` trait that we should implement for `u64` instead of introducing `u64_from_value`.

graph/src/data/subgraph/schema.rs
Jannis: Actually, `entityCount` does belong here, but the price doesn't. The only question I have: does anything actually set the entity count?

PR #1199, opened 2019-09-16
Create new RPC method: update_query_prices()

### LINKED ISSUE #1157: Add JSON-RPC method for updating pricing data

fordN:
Resolves #1157

Graph-node operators may want to apply regular query pricing updates based upon their system and market analysis. This PR adds an endpoint that allows one with access to update the `query_price` for any existing deployment on the node.

### DIFF
--- a/core/src/subgraph/registrar.rs
+++ b/core/src/subgraph/registrar.rs
@@ -327,23 +327,26 @@ where
         )))
     }
 
-    fn update_subgraph_price(
+    fn update_subgraph_query_price(
         &self,
         hash: SubgraphDeploymentId,
-        price: f32,
+        price: u64,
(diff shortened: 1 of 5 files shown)


core/src/subgraph/registrar.rs
```diff
@@ -939,3 +949,44 @@ fn reassign_subgraph(
 
     Ok(())
 }
+
+fn update_subgraph_query_prices(
```
Jannis: See my comments in #1198. I think query prices don't belong into deployments directly, but rather a separate `SubgraphDeploymentPricing` entity or similar.

leoyvens:
@fordN does this PR still make sense?

PR #1200, opened 2019-09-16
Only report indexing statuses for assigned deployments

Jannis:
Instead of returning the indexing status for all subgraph deployments of the node, only include the deployments that are currently assigned.

This is achieved by filtering the deployments before returning them, such that only those are included where a subgraph deployment assignment exists with the same ID. Note: this does not filter by the node ID.

### DIFF
--- a/graph/src/data/graphql/values.rs
+++ b/graph/src/data/graphql/values.rs
@@ -1,6 +1,6 @@
 use failure::Error;
-use graphql_parser::query::Value;
-use std::collections::BTreeMap;
+use graphql_parser::query::{Name, Value};
+use std::collections::{BTreeMap, HashMap};
 use std::str::FromStr;
 
 use crate::prelude::format_err;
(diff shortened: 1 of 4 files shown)


olibearo:
Code looks good, but this isn't what I need for the hybrid network alpha. I need the ability to query just the deployments assigned to the node being queried.

olibearo:
I'll approve now since this seems useful in its own right. 

One solution to the node id filtering is to add the `nodeId` to the graphql schema so that I can filter by?

ghost (approved):

Jannis:
@olibearo Yeah, I think that'd be a good addition to the API. I can add that before we merge.

Jannis:
@olibearo Updated!

I've also added a `indexingStatusesForSubgraphName` query field, which a few projects have been asking for to be able to monitor their subgraphs on the hosted service.

Jannis:
I'm going to go ahead and merge this PR, as it is needed in production.

PR #1205, opened 2019-09-17
Fix querying of interface entities with relational schema

lutter:


store/postgres/src/relational_queries.rs
```diff
@@ -796,12 +800,23 @@ impl<'a> QueryFragment<Pg> for FilterQuery<'a> {
...
         // and sort the whole thing by it. Also add limit
         // and offset
-        for (i, table) in self.tables.iter().enumerate() {
-            if i > 0 {
-                out.push_sql("\nunion all\n");
+        let mut first = true;
+        for entity in &self.entity_types {
+            if let Some(table) = self.layout.table_for_entity(entity).ok() {
+                self.object_query(table, first, out.reborrow())?;
+            } else if let Some(tables) = self.layout.interfaces.get(entity) {
```
leoyvens: I don't get this, there might be an interface in `entity_types`? I thought `entity_types` was a list of implementors for an interface.

lutter: I should have looked at the surrounding code more carefully - yes, it looks like `entity_types` is broken down to concrete types by the resolver already. The same seems to be true for `FindQuery`. I'll fix that up

lutter:
Removed the unnecessary interface resolution

PR #1209, opened 2019-09-18
Add some tests to check that store.get and store.set behave as expected

lutter:


runtime/wasm/src/module/test.rs
```diff
@@ -22,7 +22,7 @@ use crate::failure::Error;
 
 use super::*;
 
-use self::graph_mock::FakeStore;
+use self::graph_mock::MockStore;
```
leoyvens: Would it be desirable to use our `test-store` here to test against postgres?

lutter: I wanted to keep this at the level of unit tests, rather than integration tests; ultimately, the only involvement from the store is to create the `EntityCache`.

leoyvens: I'm of the general opinion that testing real code is more useful than testing fake code, even if it's just a bonus and not the focus of the test. But if this works better, I'm fine with it.

lutter:
Rebased to latest master

PR #1210, opened 2019-09-19
Address two performance issues with relational schema

lutter:


store/postgres/src/block_range.rs
```diff
@@ -17,6 +16,16 @@ pub type BlockNumber = i32;
 
 pub const BLOCK_NUMBER_MAX: BlockNumber = std::i32::MAX;
```
leoyvens: This should now be `std::i32::MAX - 1` since `std::i32::MAX` is infinity.

store/postgres/src/relational_queries.rs
```diff
@@ -932,22 +932,22 @@ impl<'a> QueryFragment<Pg> for RevertClampQuery<'a> {
...
-        out.push_identifier(BLOCK_RANGE)?;
+        out.push_identifier(BLOCK_RANGE_COLUMN)?;
         out.push_sql("), null)\n where");
-        out.push_identifier(BLOCK_RANGE)?;
+        out.push_identifier(BLOCK_RANGE_COLUMN)?;
         out.push_sql(" @> ");
         out.push_bind_param::<Integer, _>(&self.block)?;
-        out.push_sql(" and not upper_inf(");
-        out.push_identifier(BLOCK_RANGE)?;
+        out.push_sql(" and not (");
```
leoyvens: Are the parenthesis still required/desired here?

store/postgres/src/relational.rs
```diff
@@ -376,6 +376,14 @@ impl Layout {
             Some(skip.to_string())
         };
 
+        let order = match (order, tables.iter().next()) {
```
leoyvens: Would be good to comment on why you can look only at the first table, also this match suggests `tables` might be empty, is that possible? If not we should error or panic in that case.

store/postgres/src/block_range.rs
lutter: infinity is still `Unbounded`; I am happy to change this, but don't understand your reasoning. We use the constant to generate the same condition that is in `BLOCK_RANGE_CURRENT` for with `Store.get` and `Store.find`

leoyvens: Yhea my comment didn't make much sense, this is fine.

store/postgres/src/relational_queries.rs
lutter: Removed them. Commit 5dc5aa08

store/postgres/src/relational.rs
lutter: Added an explanation. Commit 5dc5aa08

lutter:
Addressed review comments and rebased to latest master

PR #1212, opened 2019-09-19
Allow nested filter

leoyvens:
By adding the filters after the `where` types are added. Resolves https://github.com/graphprotocol/support/issues/25.

### DIFF
--- a/graphql/src/schema/api.rs
+++ b/graphql/src/schema/api.rs
@@ -27,9 +27,9 @@ pub fn api_schema(input_schema: &Document) -> Result<Document, APISchemaError> {
     let mut schema = input_schema.clone();
     add_builtin_scalar_types(&mut schema)?;
     add_order_direction_enum(&mut schema);
-    add_field_arguments(&mut schema, &input_schema)?;
     add_types_for_object_types(&mut schema, &object_types)?;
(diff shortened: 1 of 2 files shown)


graphql/src/schema/api.rs
```diff
@@ -27,9 +27,9 @@ pub fn api_schema(input_schema: &Document) -> Result<Document, APISchemaError> {
     let mut schema = input_schema.clone();
     add_builtin_scalar_types(&mut schema)?;
     add_order_direction_enum(&mut schema);
-    add_field_arguments(&mut schema, &input_schema)?;
     add_types_for_object_types(&mut schema, &object_types)?;
     add_types_for_interface_types(&mut schema, &interface_types)?;
+    add_field_arguments(&mut schema, &input_schema)?;
```
Jannis: This may be worth a comment about why it's important to do this after adding the input object types first. But since we're adding tests here and they will break if anyone messes with this, I'm fine without a comment.

Jannis: Curious though: How did this even work for top-level fields?

leoyvens: That is in `add_query_type` which is done after. Yhea I added the test because I was too lazy to write a comment 😛

PR #1215, opened 2019-09-20
Remove Sentry dependency

leoyvens:
It's not being used.


lutter:
Looks good; there's one remaining log message in `node/src/main.rs:337` that should be removed, too.

leoyvens:
@lutter fixed, thanks!

lutter (approved):

PR #1217, opened 2019-09-20
Check templates to decide if calls should be included in block

leoyvens:
We were almost doing this correctly, but not quite.


Jannis:
@leoyvens I'll merge it for you, so I can roll it out. 😄

PR #1218, opened 2019-09-20
Refactor triggers

leoyvens:
Simplify how we order triggers by taking advantage of logs and calls already knowing their transaction index. And some small cleanups.

### DIFF
--- a/graph/src/components/ethereum/types.rs
+++ b/graph/src/components/ethereum/types.rs
@@ -15,7 +15,7 @@ pub struct EthereumBlockWithCalls {
     pub calls: Option<Vec<EthereumCall>>,
 }
 
-#[derive(Clone, Debug, Deserialize, Serialize)]
+#[derive(Clone, Debug, Default, Deserialize, Serialize)]
 pub struct EthereumBlock {
     pub block: Block<Transaction>,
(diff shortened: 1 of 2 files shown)


graph/src/components/ethereum/types.rs
```diff
@@ -110,48 +82,37 @@ impl EthereumCall {
...
-            }
-            EthereumTrigger::Call(call) => {
-                match transaction_hash_index_lookup.get(&call.transaction_hash.unwrap()) {
-                    Some(index) => Ok(Some(*index)),
-                    None => Err(()),
-                }
-            }
-            EthereumTrigger::Block(_call) => Ok(None),
+            // We only handle logs that are in a block and therefore have a `transaction_index`.
+            EthereumTrigger::Log(log) => Some(log.transaction_index.unwrap().as_u64()),
```
Jannis: Do we know under what conditions logs might _lack_ a transaction index?

leoyvens: I believe that only happens for pending transactions.

Jannis: True, that makes sense.

Jannis (approved):

Jannis:
@leoyvens I'll merge for you, so we can roll it out. 😄

Issue #1164, opened 2019-09-06
Relational schema: add synthetic primary key


lutter:
Add two different serial/synthetic id's to each table in a relational schema: one for each version of an entity which can serve as the pk, and one for each entity. This might be useful for SQL query combination and IPFS data sources, and it's better to have these in the schema already and not need them than to have to migrate to add them to existing tables.

They should both be 8 bytes so we can accommodate more than 2B entities.

lutter:
To support joining, it will be good to also track references (again) in the relational schema and add columns that use the entity id. To illustrate this, assume a schema
```
type Account {
  id: ID!,
  owner: Person!
}

type Person {
  id: ID!,
  name: String!
}
```

With additional primary keys, we would create these tables:
```
create table account(
  id: text not null,
  owner: text not null,
  -- Fill this in as soon as we have an owner with
  -- select eid from person p where p.name = owner
  -- Requires additional code in insert/update for Person
  owner_eid: int8 references person on delete set null,
  
  block_range: int4range,
  -- Uniquely identifies this version of this entity
  vid: bigserial primary key,
  -- Identifies this entity, i.e., is the same for all versions of the same entity
  eid: bigserial
 
  exclude using gist   (eid with =, {block_range} with &&)
);

create table person (
  id: text not null,
  name: text not null,

  -- same block_range, vid, eid, and exclusion constraint as above
);
```

Without these columns, a query for all accounts owned by `alex` will look like
```
  select a.*
    from account a, person p
   where p.name = 'alex'
     and a.owner = p.id
```

With these columns, the same query would look like
```
  select a.*
    from account a, person p
   where p.name = 'alex'
     and a.owner_eid = p.eid
```

Doing things this way has two advantages:
* the index for the exclusion constraint is smaller (8+8 bytes per version rather than length(id) + 8 bytes) and therefore faster to traverse
* joins between tables will similarly be faster as we join on 8 byte integers rather than arbitrary-length strings; not just is comparing integers faster than strings, Postgres can also fit more integers than strings into one 8k page, assuming that `id`'s on average are longer than 8 characters

Downside:
* the `_eid` columns need to be maintained so that every time we insert or update into a table that is referenced from somewhere, we also need to update the corresponding `_eid` column in the referrer. In a way, we precompute the work the database would have to do if we joined on the `text` rather than the `bigserial` column
* the `eid`, `vid`, and `_eid` columns need to be indexed

lutter:
[...]

PR #1219, opened 2019-09-20
Add a generated primary key to all entities

lutter:
This addresses #1164

### DIFF
--- a/store/postgres/src/relational_queries.rs
+++ b/store/postgres/src/relational_queries.rs
@@ -822,7 +822,7 @@ impl<'a> QueryFragment<Pg> for FilterQuery<'a> {
             // Overall, we generate a query
             //
             // with matches as (
-            //   select '...' as entity, id
+            //   select '...' as entity, id, vid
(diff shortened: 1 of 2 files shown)


store/postgres/src/relational_queries.rs
```diff
@@ -844,7 +844,7 @@ impl<'a> QueryFragment<Pg> for FilterQuery<'a> {
                 }
                 out.push_sql("select ");
                 out.push_bind_param::<Text, _>(&table.object)?;
-                out.push_sql(" as entity, e.id");
+                out.push_sql(" as entity, e.id, e.vid");
```
leoyvens: Interesting, you're using the `vid` here since it potentially speeds up the join?

lutter: Yes, exactly; it probably won't make a big difference, but should be a bit faster than joining on `block_range` and `id`.

lutter:
Rebased to latest master

PR #1220, opened 2019-09-21
store: Improve error message for missing non-nullable attributes

lutter:


store/postgres/src/relational_queries.rs
```diff
@@ -596,7 +596,8 @@ impl<'a> InsertQuery<'a> {
         for column in table.columns.iter() {
             if !column.is_nullable() && !entity.contains_key(&column.field) {
                 return Err(StoreError::QueryExecutionError(format!(
-                    "can not insert entity {}[{}] since value for {} is missing",
+                    "can not insert entity {}[{}] since value for non-nullable attribute {} is missing. \
+                     To fix this, mark the attribute as nullable in the GraphQL schema.",
```
Jannis: `or make sure to set the attribute.`?

lutter: Added that to the error message.

PR #1229, opened 2019-09-23
Set min_idle_connections on ConnectionPool

olibearo:
Since adding support for multiple networks on one `graph-node`, there are nodes with idle connection pools for networks they are not using.

The `idle_timeout` defaults to `10` minutes: https://docs.diesel.rs/r2d2/struct.Builder.html#method.idle_timeout.

### DIFF
--- a/store/postgres/src/store.rs
+++ b/store/postgres/src/store.rs
@@ -158,6 +158,7 @@ impl Store {
             // turns off this timeout and makes it possible that work needing
             // a database connection blocks for a very long time
             .connection_timeout(Duration::from_secs(6 * 60 * 60))
+            .min_idle(Some(1))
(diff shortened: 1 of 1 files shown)


Jannis:
@olibearo Doesn't this mean that each node will maintain at least 1 idle connection, but could still ramp it up to 10 eventually? Shouldn't we set min = max so all nodes maintain as many idle connections as they might eventually use?

olibearo:
I dug into the R2D2 connection pool lib today and decided it's best to avoid setting `min_idle` right now. Right now the Postgres server being exhausted with connections produces a clear error on start up.

Let's prioritize this work (https://github.com/graphprotocol/graph-node/issues/1232) to reduce the number of connections used in the hosted service.

PR #1233, opened 2019-09-23
Do not allocate event_meta_data when it is not needed

lutter:


store/postgres/src/entities.rs
```diff
@@ -1103,7 +1126,12 @@ impl JsonStorage {
             None => return Ok(()),
             Some(event) => event,
         };
-
+        let event_id = match history_event.id {
+            Some(id) => id,
+            None => unreachable!(
+                "we always allocate an event_id when we need to record metadata history"
+            ),
+        };
```
leoyvens: This is an `expect`.

leoyvens:
I'd rather pass this review to someone who knows how history events work, possibly @fordN or @Jannis.

store/postgres/src/entities.rs
lutter: D'oh. Thanks for pointing that out. Fixed.

lutter:
Rebased to latest master and addressed review comment.

PR #1234, opened 2019-09-24
Use one conn pool across all store instances

ghost:
These changes will lower a `graph-node`s Postgres connection usage when it is syncing more than one Ethereum network.

### DIFF
--- a/node/src/main.rs
+++ b/node/src/main.rs
@@ -467,11 +467,8 @@ fn async_main() -> impl Future<Item = (), Error = ()> + Send + 'static {
         "conn_pool_size" => store_conn_pool_size,
     );
 
-    let postgres_conn_pool = DieselStore::create_connection_pool(
-        postgres_url.clone(),
-        store_conn_pool_size,
-        &logger,
-    );
+    let postgres_conn_pool =
+        DieselStore::create_connection_pool(postgres_url.clone(), store_conn_pool_size, &logger);
     let stores: HashMap<String, Arc<DieselStore>> = eth_adapters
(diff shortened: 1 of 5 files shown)


Jannis (changes requested):
Looks like the formatting needs to be fixed.

node/src/main.rs
```diff
@@ -467,6 +467,11 @@ fn async_main() -> impl Future<Item = (), Error = ()> + Send + 'static {
         "conn_pool_size" => store_conn_pool_size,
     );
 
+    let postgres_conn_pool = DieselStore::create_conn_pool(
```
Jannis: I'm not a fan of shortening words unnecessarily myself, can we use `create_connection_pool` for the method?

Jannis (changes requested):

lutter (changes requested):
Code looks good. Since `create_conn_pool` has no direct relation to `Store` anymore, it would be nice to move `create_conn_pool` and the error handler to a new module, say `store/postgres/src/connection_pool.rs`

Longer term, I would love to unbundle the connection pool more from the store since the `Store` right now has a lot of responsibilities that are colocated simply because that's where the connection pool lives, and a separate module would be a first step in that direction.

olibearo:
@lutter Good idea, let me know how you like what I pushed

@Jannis Addressed the naming formatting

lutter (approved):

lutter:
Looks great!

Jannis (approved):

Issue #1137, opened 2019-08-27
Modify relational schema to only index string prefix


lutter:
For issue #1133, we modified index generation for JSONB storage to index only a string prefix. Similar changes need to be made to relational storage.

lutter:
We create the index on the string prefix, but queries need to be changed so that Postgres can actually use this index when we query those fields.

PR #1235, opened 2019-09-25
store: Make partial index on text columns usable in comparisons

lutter:
Fixes https://github.com/graphprotocol/graph-node/issues/1137

The existing code already creates an index on `left(column, 2048)`, but queries didn't take advantage of that yet. This PR closes that gap.

### DIFF
--- a/store/postgres/src/relational_queries.rs
+++ b/store/postgres/src/relational_queries.rs
@@ -262,9 +262,9 @@ struct PrefixComparison<'a> {
 }
 
 impl<'a> PrefixComparison<'a> {
-    fn push_column_prefix(&self, mut out: AstPass<Pg>) -> QueryResult<()> {
+    fn push_column_prefix(column: &Column, mut out: AstPass<Pg>) -> QueryResult<()> {
         out.push_sql("left(");
-        out.push_identifier(self.column.name.as_str())?;
+        out.push_identifier(column.name.as_str())?;
         out.push_sql(", ");
         out.push_sql(&STRING_PREFIX_SIZE.to_string());
         out.push_sql(")");
@@ -281,7 +281,7 @@ impl<'a> PrefixComparison<'a> {
     }
 
     fn push_prefix_cmp(&self, op: Comparison, mut out: AstPass<Pg>) -> QueryResult<()> {
-        self.push_column_prefix(out.reborrow())?;
+        Self::push_column_prefix(self.column, out.reborrow())?;
         out.push_sql(op.as_str());
         self.push_value_prefix(out.reborrow())
     }
@@ -590,7 +590,22 @@ impl<'a> QueryFilter<'a> {
         }
 
         if have_non_nulls {
-            out.push_identifier(column.name.as_str())?;
+            if column.is_text()
+                && values.iter().all(|v| match v {
+                    Value::String(s) => s.len() <= STRING_PREFIX_SIZE - 1,
+                    _ => false,
+                })
+            {
+                // If all values are shorter than STRING_PREFIX_SIZE - 1,
+                // only check the prefix of the column; that's a fairly common
+                // case and we present it in the best possible way for
+                // Postgres' query optimizer
+                // See PrefixComparison for a more detailed discussion of what
+                // is happening here
+                PrefixComparison::push_column_prefix(&column, out.reborrow())?;
+            } else {
+                out.push_identifier(column.name.as_str())?;
+            }
             if negated {
                 out.push_sql(" not in (");
             } else {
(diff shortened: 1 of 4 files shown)


store/postgres/src/relational_queries.rs
```diff
@@ -223,6 +224,126 @@ impl<'a> QueryFragment<Pg> for QueryValue<'a> {
...
+    fn walk_ast(&self, mut out: AstPass<Pg>) -> QueryResult<()> {
+        use Comparison::*;
+
+        // Write uv for a string where u is a string at most N long and v is
+        // empty if u is shorter than N, same for another string st.
+        // For the comparison operators '=', '!=', '<', '>' we have the
+        // relation
+        //   uv op st <=> u op s || u = s && uv op st
+        //            <=> u op s || u = s &&  v op  t
+        // If t is the empty string, v op t will always be false so that the
```
leoyvens: > If t is the empty string, v op t will always be false

This is not true since v is any string. The case where it's sufficient to compare the prefix is not so simple to explain, I don't know what would be the best way.

store/postgres/src/relational_queries.rs
```diff
@@ -223,6 +224,126 @@ impl<'a> QueryFragment<Pg> for QueryValue<'a> {
...
+        // relation must already be true for the prefixes, and we can skip
+        // the check after the '||'.
+        // When the operator is '<=' or '>=', the first part of the '||' must
+        // use the strict form '<' and '>' when t is not the empty string.
+        // When op is '=', the above boils down to
+        //                u = s || uv = st
```
leoyvens: I think you meant `&&` not `||` here.

store/postgres/src/relational_queries.rs
```diff
@@ -223,6 +224,126 @@ impl<'a> QueryFragment<Pg> for QueryValue<'a> {
...
+        QueryValue(self.text, &self.column.column_type).walk_ast(out)
+    }
+}
+
+impl<'a> QueryFragment<Pg> for PrefixComparison<'a> {
+    fn walk_ast(&self, mut out: AstPass<Pg>) -> QueryResult<()> {
+        use Comparison::*;
+
+        // Write uv for a string where u is a string at most N long and v is
+        // empty if u is shorter than N, same for another string st.
```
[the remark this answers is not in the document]
lutter: What I am trying to say is that `v` is only nonempty if `u` is exactly `N` long. Or put another way, if `u` is shorter than `N`, `v` is empty; if `u` is exactly `N`, `v` may or may not be empty.

store/postgres/src/relational_queries.rs
lutter: Yeah, I've been struggling with writing this up cleanly. What do you think of the following, which would require a small code change, and I agree there need to be more tests:

Looking at the comparison operators besides `<=` and `>=`, the general statement
```
      uv op st <=> u op s || u = s && uv op st
```
is true because the comparisons are all lexicographic, i.e., there's a first position in the two strings where the characters at that position compare according to `op`, and up to that position the two strings are equal.

We want to know under what circumstances we do not need to check `u = s && uv op st`. If we look at strings `st` where `len(st) <= N - 1`, `t` will be empty, and `len(s) <= N - 1`.

Since we compare prefixes of length `N`, either `len(uv) <= N - 1`, in which case `v` is empty, and `u op s = uv op st`, or `len(uv) >= N`, and therefore `len(u) = N`, `u != s` so that the check `u = s && uv op st` will always be false.

lutter: And for `<=` and `>=` we apply the above logic to (using `<=` as an example`)
```
  uv <= st <=> u < s || u = s && uv <= st
```

store/postgres/src/relational_queries.rs
lutter: Yes, true.

store/postgres/src/relational_queries.rs
lutter: For `=` and `!=` this boils down to
```
  uv = st <=> u = s || u = s && uv = st <=> u = s && uv = st
  uv != st <=> u != s || u = s && uv != st <=> u != s || uv != st
```

The `!=` case won't buy us much, since the `|| uv != st` will make Postgres not use the index in almost all cases.

leoyvens: This is mostly making sense to me, but `uv = st <=> u = s || ...` is not true.

lutter: Yeah, you're right. I'll have another attempt at cleaning this up.

lutter:
I revamped the PR, rebased to latest master, and think came up with an explanation of what's happening that's a lot less complicated than my comments above

PR #1238, opened 2019-09-26
Use a separate timeout for loading subgraph files from IPFS

Jannis:
Before, whatever you set as the IPFS timeout was used not just for `ipfs.cat` in mappings, but also for loading subgraphs. Now, when you set the timeout to a short duration (like 2s), subgraphs would often fail to start up. Restricting the time spent in `ipfs.cat` is reasonable thing to do though, so the timeout used for `ipfs.cat` and subgraph loading should be independent.

This change adds a `with_timeout` method to `LinkResolver` to create a cloned resolver that has its own timeout. This is then used in the `SubgraphRegistrar` to set the timeout for manifest resolution based on the `GRAPH_IPFS_SUBGRAPH_LOADING_TIMEOUT` environment variable (default: 60 seconds).

### DIFF
--- a/docs/environment-variables.md
+++ b/docs/environment-variables.md
@@ -45,7 +45,8 @@ those.
   take (in seconds, default is unlimited)
 - `GRAPH_IPFS_SUBGRAPH_LOADING_TIMEOUT`: timeout for IPFS requests made to load
   subgraph files from IPFS (in seconds, default is 60).
-- `GRAPH_IPFS_TIMEOUT`: timeout for IPFS requests (in seconds, default is 60).
+- `GRAPH_IPFS_TIMEOUT`: timeout for IPFS requests from mappings using `ipfs.cat`
+  or `ipfs.map` (in seconds, default is 60).
 - `GRAPH_MAX_IPFS_FILE_BYTES`: maximum size for a file that can be retrieved
   with `ipfs.cat` (in bytes, default is unlimited)
 - `GRAPH_MAX_IPFS_MAP_FILE_SIZE`: maximum size of files that can be processed
--- a/docs/environment-variables.md
+++ b/docs/environment-variables.md
@@ -43,8 +43,9 @@ those.
 
 - `GRAPH_MAPPING_HANDLER_TIMEOUT`: amount of time a mapping handler is allowed to
   take (in seconds, default is unlimited)
-- `GRAPH_IPFS_TIMEOUT`: timeout for ipfs requests. In seconds, default is 60.
-  seconds.
+- `GRAPH_IPFS_SUBGRAPH_LOADING_TIMEOUT`: timeout for IPFS requests made to load
+  subgraph files from IPFS (in seconds, default is 60).
+- `GRAPH_IPFS_TIMEOUT`: timeout for IPFS requests (in seconds, default is 60).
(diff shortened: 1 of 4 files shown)


graph/src/components/link_resolver.rs
```diff
@@ -18,6 +19,9 @@ pub type JsonValueStream =
 
 /// Resolves links to subgraph manifests and resources referenced by them.
 pub trait LinkResolver: Send + Sync + 'static + Sized {
+    /// Creates a new link resolver that implements the given timeout.
+    fn with_timeout(&self, timeout: Duration) -> Box<Self>;
```
leoyvens: I'd think this `Box` could be avoided since we have a `Sized` bound on the trait.

leoyvens (approved):

graph/src/components/link_resolver.rs
Jannis: I tried just `Self` but it didn't compile.

Jannis: Compiles now, I'll push an update.

leoyvens (approved):

docs/environment-variables.md
```diff
@@ -43,8 +43,9 @@ those.
 
 - `GRAPH_MAPPING_HANDLER_TIMEOUT`: amount of time a mapping handler is allowed to
   take (in seconds, default is unlimited)
-- `GRAPH_IPFS_TIMEOUT`: timeout for ipfs requests. In seconds, default is 60.
-  seconds.
+- `GRAPH_IPFS_SUBGRAPH_LOADING_TIMEOUT`: timeout for IPFS requests made to load
+  subgraph files from IPFS (in seconds, default is 60).
+- `GRAPH_IPFS_TIMEOUT`: timeout for IPFS requests (in seconds, default is 60).
```
lutter: Could we say something like 'timeout for IPFS requests from mapping functions (like ipfs.cat and ipfs.map)` to make it clearer that one is about the internal workings of graph-node and the other is what people are more likely to want to influence?

graph/src/components/link_resolver.rs
```diff
@@ -18,6 +19,9 @@ pub type JsonValueStream =
 
 /// Resolves links to subgraph manifests and resources referenced by them.
 pub trait LinkResolver: Send + Sync + 'static + Sized {
+    /// Creates a new link resolver that implements the given timeout.
+    fn with_timeout(&self, timeout: Duration) -> Self;
```
lutter: Why not take `self` here instead of `&self` and avoid a clone (or leave it to the caller to determine whether a clone is needed or not)

Jannis: Works for me! @lutter PR updated.

docs/environment-variables.md
Jannis: Yep, can do.

Jannis: Updated.

graph/src/components/link_resolver.rs
lutter: Looks good to me !

docs/environment-variables.md
lutter: Great! Thanks!

lutter (approved):

Issue #1093, opened 2019-08-01
Subgraph fails due to `AbortUnless` condition


leoyvens:
These should never fail a subgraph, but we got a report where a subgraph failed with message:
```
RRO Subgraph instance failed to run: Error while processing block stream for a subgraph:
store transaction failed, need to retry: AbortUnless triggered abort, expected ["Qmcr7X8xrcKnhm1QGD1oFrUC1AGQpwYigifzy2LSNFCPsV"] but got []:
update did not change any rows, code: SubgraphSyncingFailure, id: Qmcr7X8xrcKnhm1QGD1oFrUC1AGQpwYigifzy2LSNFCPsV
```
Possiby we should just retry in this case.

lutter:
The message `update did not change any rows` indicates that we tried to change an entity, but that entity didn't exist. Since update operations are only used for metadata operations (all in `graph/src/data/subgraph/schema.rs`), that's a little disconcerting.

We could log this as a warning, but I am concerned that this is hiding some other problem.

orenyodfat:
we are getting this error even though setting DISABLE_BLOCK_INGESTOR: 'true' for having different nodes for queries and indexing.

lutter:
Based on the code, this error can only happen when we try to update the block pointer for a subgraph, and indicates that the block pointer in the database is not what we think it is in the app. There are three reasons for updating the block pointer:

* storing entity changes collected while processing one block
* skipping over blocks with no changes
* reverting a block

One scenario where this could happen is if two graph-node processes work on the same subgraph, and both work on the same block and finish at slightly different times. Another one might be that there is something afoot with block reversions. The changes in #1241 should help shed light on this, but until we get to the bottom of this, raising this error is the safe thing to do.

Jannis:
Closing due to inactivity. Please reopen if this is still an issue.

PR #1241, opened 2019-09-26
Include more detail in 'update did not change any rows' error

lutter:
This should help in getting to the bottom of why we sporadically see #1093


Jannis:
This is good!

Issue #1130, opened 2019-08-24
Optimize historical event scanning


Jannis:
**Do you want to request a *feature* or report a *bug*?**

Performance bug.

**What is the current behavior?**

Scanning for common events like `Transfer` is currently slow. Scanning for events from many data sources (e.g. created by templates) at once is also slow. Both used in combination... you can guess for yourself.

The reason for this is that all event signatures that we are scanning for are hashed together for the `eth_getLogs` filter. The more event signatures are hashed together, the less efficient the filter is. Also, as soon as there is more than one contract / data source involved, the filter we're building is no longer tied to a single contract address and therefor much less efficient.

**What is the expected behavior?**

Scanning for any event should be fast, regardless of how many data sources there are. It should not take days to handle `Transfer` events.

leoyvens:
Really we should do the bloom filter check locally by first pulling in all block headers, which we probably already do anyways.

leoyvens:
Expanding on the previous comment, scanning logs through `eth_getLogs` has a few issues:

1.  Robustness. We've seen different issues arise with `eth_getLogs` behaviour changing under our feet such as #776 and #1111. It's a thing Ethereum node services tend to tinker with, because they have more general requirements than ours, so reducing our exposure to this seems worth it.

2. Performance. `eth_getLogs` is fast but we wish it was faster in cases where the filter is broad or when scanning through the initial blocks which are not relevant for most subgraphs. A local implementation will be at least as fast, more predictable and also tunable for our use case, just skipping the roundtrip should be enough to see a nice speedup.

3. Code simplicity. The `eth_getLogs` interface takes block numbers not hashes, which results in [this complexity in the code](https://github.com/graphprotocol/graph-node/blob/master/datasource/ethereum/src/block_stream.rs#L355). In order to implement #902 we need a function that can scan and re-scan block ranges correctly and quickly, we could probably work with the complexity of the existing situation but doing a full bottom-up refactoring may be worthwhile.

#### Proposed Solution

The block ingestor would always keep a complete and consistent chain of block headers in the DB. This way, we can scan the bloom filter locally and find the relevant block pointers. Note that we don't really use the logs returned from `eth_getLogs`, we only need to know if a block is potentially relevant. The `ethereum_blocks` table would be re-created to have the header fields in the ddl instead of in jsonb, and a unique constraint on the block number, having a local consistent view of the chain should be a generally useful thing. Storing all block headers should require under 1GB, even with indexes.

[...]

PR #1242, opened 2019-09-26
Detect Alchemy timeouts on eth_getLogs

leoyvens:
Treat it the same as the Infura log limit, and reduce the range.

I still want to do #1130 and try separating the log calls, but this is a quick way for affected subgraphs make progress again, and is desirable anyways.

This was tested for the Nuo Network subgraph, at a block where it would previously get stuck with the 503 error.


Jannis (approved):
Nice!

PR #1243, opened 2019-09-26
Query all subgraphs/versions/deployments for indexing statuses

Jannis:
Before, the indexing status API would implicitly only query the first 100 subgraphs/versions/deployments. This commit changes that to query up to 1,000,000 of these. For the forseeable future, that should be sufficient to query _everything_.


fordN (approved):
LGTM 👍

PR #1245, opened 2019-09-27
store: Reduce the size of string prefixes we index

lutter:
The previous size of 2k was the largest we could store in a Postgres BTree
index, but it's excessive for supporting normal string operations; anything
beyond a few hundred bytes is text, and very unlikely to be accessed
through 'hard' string comparisons; it's much more likely that such text
will be accessed through word searches and similar for which the BTree
index is of no use anyway.

We lower the prefix size to 256 to reduce the size of the index for
attributes that do store large text entries.

### DIFF
--- a/store/postgres/src/entities.rs
+++ b/store/postgres/src/entities.rs
@@ -55,10 +55,13 @@ use crate::notification_listener::JsonNotification;
 use crate::relational::{IdType, Layout};
 use crate::store::Store;
 
-/// The size of string prefixes that we index. This should be large enough
-/// that we catch most strings, but small enough so that we can still insert
-/// it into a Postgres BTree index
-pub(crate) const STRING_PREFIX_SIZE: usize = 2048;
+/// The size of string prefixes that we index. This is chosen so that we
+/// will index strings that people will do string comparisons like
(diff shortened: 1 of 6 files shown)


Jannis:
@lutter This will represent a breaking change for projects that depend on long string comparisons beyond 256 bytes. We should double-check especially with DAOstack if this will affect them.

Jannis:
Other than that, the changes look good.

lutter:
> @lutter This will represent a breaking change for projects that depend on long string comparisons beyond 256 bytes. We should double-check especially with DAOstack if this will affect them.

It's not a breaking change, in fact, it's not even user-visible; the only time this _might_ make a difference is when users search for very long strings, but even then the search will narrow down possible results by using the index for the first 256 bytes of the search text.

TLDR: users can perform the exact same searches with and without this change, it's simply avoiding indexing way more than we will reasonably use.

Jannis:
@lutter Ah, so we'll still perform a complete equality check / comparison after narrowing down the candidates?

lutter:
Yes, absolutely. PR #1235 went to great length to write these comparisons in a way that (a) produce the correct result on the entire string and (b) let Postgres use just the index when we know that that is enough to produce the correct result.

Issue #1213, opened 2019-09-19
Reject subgraphs with invalid @derivedFrom fields


lutter:
Graph CLI, since [this commit](https://github.com/graphprotocol/graph-cli/commit/d1d5ee3480) disallows a number of errors around the `@derivedFrom` annotation. It is still possible to deploy such subgraphs to `graph-node`, e.g., by using an older version of `graph-cli`. The issues that `graph-cli` flags will become serious issues with the relational schema, and `graph-node` should reject new deployments of subgraphs with these problems

PR #1247, opened 2019-09-28
graph: Validate @derivedFrom directives on subgraph schemas

lutter:
These validations follow the corresponding validations in graph-cli, and make sure users do not deploy invalid subgraphs, e.g., because they have not updated their subgraph or graph-cli

Fixes https://github.com/graphprotocol/graph-node/issues/1213

### DIFF
--- a/graph/src/data/graphql/validation.rs
+++ b/graph/src/data/graphql/validation.rs
@@ -1,6 +1,7 @@
 use crate::prelude::Fail;
 use graphql_parser::schema::*;
 use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
 use std::fmt;
 
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
@@ -27,11 +28,17 @@ pub enum SchemaValidationError {
         _0, _1, _2
     )]
     CannotImplement(String, String, Strings), // (type, interface, missing_fields)
+    #[fail(
+        display = "Field `{}` in type `{}` has invalid @derivedFrom: {}",
+        _1, _0, _2
+    )]
+    DerivedFromInvalid(String, String, String), // (type, field, reason)
 }
 
 /// Validates whether a GraphQL schema is compatible with The Graph.
 pub(crate) fn validate_schema(schema: &Document) -> Result<(), SchemaValidationError> {
-    validate_schema_types(schema)
+    validate_schema_types(schema)?;
+    validate_derived_from(schema)
 }
 
 /// Validates whether all object types in the schema are declared with an @entity directive.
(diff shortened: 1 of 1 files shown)


graph/src/data/graphql/validation.rs
```diff
@@ -93,10 +100,199 @@ pub fn get_object_type_definitions(schema: &Document) -> Vec<&ObjectType> {
...
+            reason.to_owned(),
+        )
+    }
+
+    let type_definitions = get_object_type_definitions(schema);
+    let object_and_interface_type_fields = get_object_and_interface_type_fields(schema);
+
+    // Iterate over all derived fields in all entity types; include the
+    // `field` argument of @derivedFrom directive
+    for (object_type, field, target_field) in type_definitions
```
Jannis: We should probably move the schema validation code into the `graph-graphql` crate. Like you already mentioned, a number of helpers have already been added there but can't be used. This is another: `graph_graphql::schema::ast::get_derived_from_directive` and `graph_graphql::schema::ast::get_derived_from_field` already exist there.

Jannis: We don't have to do it now though.

graph/src/data/graphql/validation.rs
```diff
@@ -93,10 +100,199 @@ pub fn get_object_type_definitions(schema: &Document) -> Vec<&ObjectType> {
...
+        .clone()
+        .iter()
+        .map(|object_type| {
+            object_type
+                .fields
+                .iter()
+                .map(move |field| (object_type, field))
+        })
+        .flatten()
```
Jannis: The above `map` and `flatten` could be combined with `flat_map`.

graph/src/data/graphql/validation.rs
```diff
@@ -93,10 +100,199 @@ pub fn get_object_type_definitions(schema: &Document) -> Vec<&ObjectType> {
...
+        .filter_map(|(object_type, field)| {
```
Jannis: The `if let Some` logic here could be simplified a little by using `map`:
```rust
...
    .filter_map(|(object_type, field)| {
        field
            .directives
            .iter()
            .find(|directive| directive.name == "derivedFrom")
            .map(|directive| {
                (
                    object_type,
                    field,
                    directive
                        .arguments
                        .iter()
                        .find(|(name, _)| name == "field")
                        .map(|(_, value)| (object_type, field, target_field))
                )
            })
    })
```

graph/src/data/graphql/validation.rs
Jannis: The alternative would be to make the `graph_graphql` crate unaware of an Graph-specific data structures and make it a dependency of `graph`.

graph/src/data/graphql/validation.rs
```diff
@@ -93,10 +100,199 @@ pub fn get_object_type_definitions(schema: &Document) -> Vec<&ObjectType> {
...
+
+        // Check that the type we are deriving from exists
+        let target_type_name = get_base_type(&field.field_type);
+        let target_fields = object_and_interface_type_fields
+            .get(target_type_name)
+            .ok_or_else(|| {
+                invalid(
+                    object_type,
+                    &field.name,
+                    "the type of the field must be an existing entity type",
```
Jannis: I'd change this to `entity or interface type`.

graph/src/data/graphql/validation.rs
```diff
@@ -93,10 +100,199 @@ pub fn get_object_type_definitions(schema: &Document) -> Vec<&ObjectType> {
...
+            .iter()
+            .find(|field| &field.name == target_field)
+            .ok_or_else(|| {
+                let msg = format!(
+                    "field `{}` does not exist on type `{}`",
+                    target_field, target_type_name
+                );
+                invalid(object_type, &field.name, &msg)
+            })?;
+        // The field we are deriving from has to point back to us; as an
```
Jannis: Nit: add an empty line above.

graph/src/data/graphql/validation.rs
lutter: I agree that that would be good to do, but I also do not want to do that as part of this PR which just builds on what's there already. I opened #1265 to track that.

lutter:
Addressed review comments and rebased to latest master.

PR #1260, opened 2019-09-30
Smarter use of eth_getLogs

### LINKED ISSUE #1130: Optimize historical event scanning

leoyvens:
Currently we merge all log filters for a subgraph in a single large filter. This is can be bad for two reasons:
- It returns false positives, which may be slowing down the calls.
- The filter can become too broad with many data sources, timing out on Alchemy or returning too many logs on Infura, requiring the range to be further split up.

This PR splits the `eth_getLogs` calls so that we don't request contract-event combinations we don't actually need, and even split it up a bit more than necessary for more parallel requests.

I tested with the NuoNetwork subgraph for speed and correctness, and we no longer get timeouts on it, probably because we're splitting up the call for the thousands of dynamic data sources per event, which splits it in five parallel requests. I Also tested wildcard event signatures still work.

The ERC20 subgraph didn't spend too much time getting logs, though I didn't wait for it to get very far and I'm not sure what the situation was before. At least on a cold DB the slowest thing was downloading receipts, as usual.

I expected we'll be tweaking this based on the behaviour of specific subgraphs on specific Ethereum node providers, but for now I'll consider that this fixes #1130.

### DIFF
--- a/graph/src/components/ethereum/adapter.rs
+++ b/graph/src/components/ethereum/adapter.rs
@@ -10,7 +10,7 @@ use web3::types::*;
 use super::types::*;
 use crate::prelude::*;
 
-pub type EventSig = H256;
+pub type EventSignature = H256;
 
 /// A collection of attributes that (kind of) uniquely identify an Ethereum blockchain.
 pub struct EthereumNetworkIdentifier {
@@ -89,14 +89,14 @@ impl From<Error> for EthereumAdapterError {
 #[derive(Copy, Clone, Debug, PartialEq, Eq, Ord, PartialOrd, Hash)]
 enum LogFilterNode {
     Contract(Address),
-    Event(EventSig),
+    Event(EventSignature),
 }
 
 /// Corresponds to an `eth_getLogs` call.
 #[derive(Clone)]
 pub struct EthGetLogsFilter {
     pub contracts: Vec<Address>,
-    pub event_sigs: Vec<EventSig>,
+    pub event_signatures: Vec<EventSignature>,
 }
 
 impl fmt::Display for EthGetLogsFilter {
@@ -104,15 +104,15 @@ impl fmt::Display for EthGetLogsFilter {
         if self.contracts.len() == 1 {
             write!(
                 f,
-                "contract {}, {} events",
+                "contract {:?}, {} events",
                 self.contracts[0],
-                self.event_sigs.len()
+                self.event_signatures.len()
             )
-        } else if self.event_sigs.len() == 1 {
+        } else if self.event_signatures.len() == 1 {
             write!(
                 f,
-                "event {}, {} contracts",
-                self.event_sigs[0],
+                "event {:?}, {} contracts",
+                self.event_signatures[0],
                 self.contracts.len()
             )
         } else {
@@ -129,7 +129,7 @@ pub struct EthereumLogFilter {
     contracts_and_events_graph: GraphMap<LogFilterNode, (), petgraph::Undirected>,
 
     // Event sigs with no associated address, matching on all addresses.
-    wildcard_events: HashSet<EventSig>,
+    wildcard_events: HashSet<EventSignature>,
 }
 
 impl EthereumLogFilter {
@@ -151,11 +151,11 @@ impl EthereumLogFilter {
                 // The `Log` matches the filter either if the filter contains
                 // a (contract address, event signature) pair that matches the
                 // `Log`, or if the filter contains wildcard event that matches.
+                let contract = LogFilterNode::Contract(log.address.clone());
+                let event = LogFilterNode::Event(*sig);
                 self.contracts_and_events_graph
                     .all_edges()
                     .any(|(s, t, ())| {
-                        let contract = LogFilterNode::Contract(log.address.clone());
-                        let event = LogFilterNode::Event(*sig);
                         (s == contract && t == event) || (t == contract && s == event)
                     })
                     || self.wildcard_events.contains(sig)
@@ -217,7 +217,7 @@ impl EthereumLogFilter {
         for wildcard_event in self.wildcard_events {
             filters.push(EthGetLogsFilter {
                 contracts: vec![],
-                event_sigs: vec![wildcard_event],
+                event_signatures: vec![wildcard_event],
             })
         }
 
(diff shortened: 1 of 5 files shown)


graph/src/components/ethereum/adapter.rs
```diff
@@ -84,9 +86,50 @@ impl From<Error> for EthereumAdapterError {
...
+    pub contracts: Vec<Address>,
+    pub event_sigs: Vec<EventSig>,
+}
+
+impl fmt::Display for EthGetLogsFilter {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        if self.contracts.len() == 1 {
+            write!(
+                f,
+                "contract {}, {} events",
```
Jannis: I noticed in the logs that the contract address is ellipsized in the middle (e.g. `contract 0x1bfd…59ba`). Could we make logs contain the complete address so it's easier to copy and paste it?

graph/src/components/ethereum/adapter.rs
```diff
@@ -84,9 +86,50 @@ impl From<Error> for EthereumAdapterError {
...
+                self.contracts[0],
+                self.event_sigs.len()
+            )
+        } else if self.event_sigs.len() == 1 {
+            write!(
+                f,
+                "event {}, {} contracts",
```
Jannis: Same goes with the event signature here, that's also shortened.

graph/src/components/ethereum/adapter.rs
```diff
@@ -1,15 +1,17 @@
...
 use std::collections::{HashMap, HashSet};
-use std::iter::FromIterator;
+use std::fmt;
 use tiny_keccak::keccak256;
 use web3::types::*;
 
 use super::types::*;
 use crate::prelude::*;
 
+pub type EventSig = H256;
```
Jannis: Any reason to shorten this name to `EventSig`? How about `EventSignature`? :wink:

graph/src/components/ethereum/adapter.rs
```diff
@@ -84,9 +86,50 @@ impl From<Error> for EthereumAdapterError {
...
+enum LogFilterNode {
+    Contract(Address),
+    Event(EventSig),
+}
+
+/// Corresponds to an `eth_getLogs` call.
+#[derive(Clone)]
+pub struct EthGetLogsFilter {
+    pub contracts: Vec<Address>,
+    pub event_sigs: Vec<EventSig>,
```
Jannis: Same here – I'd prefer `event_signatures`.

Jannis:
I was concerned the reduced parallel block ranges might make scanning logs slower for subgraphs with e.g. just one data source, so I benchmarked this PR with both Gravity (1 data source, 2 events) and Melon (9 data sources, 16 events if I've counted correctly). Here are the results:

## Gravity

Scanning past logs for Gravity against Alchemy, before and after:

**Before**

```
Oct 01 13:45:19.369 INFO Start subgraph, data_sources: 1, subgraph_id: Qmdp5oSxE28VNF17LsoojjdYKMMDKDvfLxQLzXacW3kCSC, component: SubgraphInstanceManager
Oct 01 13:48:00.589 DEBG Scanning blocks [6000001, 6010000], subgraph_id: Qmdp5oSxE28VNF17LsoojjdYKMMDKDvfLxQLzXacW3kCSC, component: SubgraphInstanceManager > BlockStream
```

Time: 02:41 mins

**After**

```
Oct 01 13:57:25.438 INFO Start subgraph, data_sources: 1, subgraph_id: Qmdp5oSxE28VNF17LsoojjdYKMMDKDvfLxQLzXacW3kCSC, component: SubgraphInstanceManager
Oct 01 14:00:32.118 DEBG Scanning blocks [6000001, 6010000], subgraph_id: Qmdp5oSxE28VNF17LsoojjdYKMMDKDvfLxQLzXacW3kCSC, component: SubgraphInstanceManager > BlockStream
```

Time: 02:53 mins

## Melon

Scanning past logs for Melon against Alchemy, before and after:

**Before**

```
Oct 01 14:23:01.647 INFO Start subgraph, data_sources: 9, subgraph_id: QmbfdXRXpeQL3vafKVLYDjev7avzDoUGp1BKQBfgJhpRo9, component: SubgraphInstanceManager
Oct 01 14:26:37.194 DEBG Scanning blocks [6000001, 6010000], subgraph_id: QmbfdXRXpeQL3vafKVLYDjev7avzDoUGp1BKQBfgJhpRo9, component: SubgraphInstanceManager > BlockStream
```

Time: 3:36 mins

**After**

```
Oct 01 14:09:53.718 INFO Start subgraph, data_sources: 9, subgraph_id: QmbfdXRXpeQL3vafKVLYDjev7avzDoUGp1BKQBfgJhpRo9, component: SubgraphInstanceManager
Oct 01 14:13:03.868 DEBG Scanning blocks [6000001, 6010000], subgraph_id: QmbfdXRXpeQL3vafKVLYDjev7avzDoUGp1BKQBfgJhpRo9, component: SubgraphInstanceManager > BlockStream
```

Time: 3:10 mins

## Conclusion

This is just a small sample, but it suggests the performance has degraded slightly for subgraphs with just one data source and few events and improved slightly for subgraphs with many data sources and events.

Overall the differences are small enough to not be a concern, especially since subgraphs with just one data source are rare and so we're likely going to see an improvement with most subgraphs.

graph/src/components/ethereum/adapter.rs
```diff
@@ -84,9 +86,50 @@ impl From<Error> for EthereumAdapterError {
...
+#[derive(Clone, Debug, Default)]
 pub struct EthereumLogFilter {
-    pub contract_address_and_event_sig_pairs: HashSet<(Option<Address>, H256)>,
+    /// Log filters can be represented as a bipartite graph between contracts and events. An edge
+    /// exists between a contract and an event if a data source for the contract has a trigger for
+    /// the event.
+    contracts_and_events_graph: GraphMap<LogFilterNode, (), petgraph::Undirected>,
+
+    // Event sigs with no associated address, matching on all addresses.
+    wildcard_events: HashSet<EventSig>,
```
Jannis: I wonder if these could be represented as another `LogFilterNode` variant to avoid the special casing. It might even make the filter splitting algorithm more efficient?

graph/src/components/ethereum/adapter.rs
```diff
@@ -103,64 +146,124 @@ impl EthereumLogFilter {
...
-                }),
+
+            Some(sig) => {
+                // The `Log` matches the filter either if the filter contains
+                // a (contract address, event signature) pair that matches the
+                // `Log`, or if the filter contains wildcard event that matches.
+                self.contracts_and_events_graph
+                    .all_edges()
+                    .any(|(s, t, ())| {
+                        let contract = LogFilterNode::Contract(log.address.clone());
```
Jannis: I'd create `contract` and `event` upfront to avoid creating it over and over again.

Jannis (changes requested):
Love it! I've got a few small comments but this is close to ready.

graph/src/components/ethereum/adapter.rs
leoyvens: Ah yhea that's annoying.

graph/src/components/ethereum/adapter.rs
leoyvens: I preferred to keep this separate because imo it makes the code more obvious and readable.

I considered putting it into the graph but we'd also need a special contract vertex representing all contracts. There might be an optimization there if a wildcard event also appears tied to a contract, but the ideal behaviour is unclear. Wildcards are a niche feature anyhow, people usually want to control what contract is emitting an event so that random contracts can't put data into their subgraph.

graph/src/components/ethereum/adapter.rs
leoyvens: Good catch.

leoyvens:
@Jannis Great that you did some benchmarking! It does seem like Melon might have benefited from the parallelization, I do not know about Gravity since nothing should have changed there.

graph/src/components/ethereum/adapter.rs
Jannis: That's true. It was just an idea.

Jannis (approved):

PR #1262, opened 2019-10-01
Define start_block in the subgraph manifest

### LINKED ISSUE #1131: Subgraph indexing start block in the manifest

Feature, but it's also sort of a performance bug.

Subgraphs always start indexing from the genesis block. In the majority of cases, the contracts that the subgraph indexes are only deployed later (e.g. in block 5,000,000). In many cases, scanning the history of the chain is expensive/slow (think: call filters, block filters, common event signatures like with `Transfer` events).

Indexing of subgraphs starts with the block that the first of their contracts were deployed at. This should save a lot of time.

**Proposed solution**

1. A new, optional `startBlock` (or similarly named) on each contract data source.
2. When a subgraph is deployed, we use call traces to verify that the `startBlock` blocks include a contract creation for the corresponding data source contract. If this is not the case, the deployment is rejected.

This way we save time and guarantee that we don't miss out on any data before the contract creation.

fordN:
Resolves #1131

Start blocks can now be defined per data source in the manifest and are used to resolve a subgraph wide earliest Ethereum block value.

The subgraph registrar logs warnings if the start blocks do not represent the block where the data source's contract (if specified) was created. 
The permissive nature of the start block validation allows flexibility for a subgraph owner to filter the chain as they see fit.

### DIFF
--- a/core/src/subgraph/registrar.rs
+++ b/core/src/subgraph/registrar.rs
@@ -1,5 +1,4 @@
 use lazy_static::lazy_static;
-use std::cmp::Ordering;
 use std::collections::{HashMap, HashSet};
 use std::time::{Duration, Instant, SystemTime, UNIX_EPOCH};
 use std::{env, iter};
@@ -680,84 +679,79 @@ fn create_subgraph_version(
         },
     });
 
-    let eth_adapter = ethereum_adapter.clone();
-
-    let _starting_blocks: Option<Vec<EthereumBlockPointer>> = manifest
+    let start_blocks: Option<Vec<EthereumBlockPointer>> = manifest
         .data_sources
         .clone()
         .into_iter()
         .map(|data_source| {
             let source_address = data_source.source.address;
-            let block_num = match data_source
+            let block_number = match data_source
                 .source
                 .start_block
                 .and_then(|block_str| block_str.parse::<u64>().ok())
             {
-                Some(block_num) => block_num,
-                None => return None,
+                Some(number) => number,
+                None => {
+                    warn!(
+                        logger,
+                        "Start block invalid or not specified for the '{}' data source",
+                        &data_source.name
+                    );
+                    return None
+                },
             };
 
-            // Safe to unwrap?
-            //            let block_num = start_block.parse::<u64>().unwrap();
-            let full_block = match eth_adapter
-                .block_hash_by_block_number(logger, block_num)
-                .and_then(move |block_hash_opt| {
-                    block_hash_opt.ok_or_else(|| {
-                        format_err!(
-                            "Ethereum node could not find block with number {}",
-                            block_num
-                        )
-                    })
-                })
-                .and_then(|block_hash| eth_adapter.block_by_hash(logger, block_hash))
-                .and_then(move |block_opt| {
-                    block_opt.ok_or_else(|| {
-                        format_err!(
-                            "Ethereum node could not find block with number {}",
-                            block_num
-                        )
-                    })
-                })
-                .from_err()
-                .and_then(|block| eth_adapter.load_full_block(logger, block))
-                .wait() {
-                Ok(block) => block,
-                Err(e) => return None,
-            };
-
-            if full_block
-                .transaction_receipts
-                .iter()
-                .filter(|receipt| {
-                    receipt.contract_address == source_address && receipt.contract_address.is_some()
-                })
-                .any(|receipt| {
-                    match full_block
-                        .block
-                        .transactions
-                        .iter()
-                        .find(|&transaction| transaction.hash == receipt.transaction_hash)
-                    {
-                        Some(t) => t.to.is_none(),
-                        None => false,
-                    }
-                })
+            // Validate that the start block provided for the Data Source is indeed the block that
+            // contains the contract creation transaction. Allow the block to be used as the
+            // subgraph level start block regardless, but emit a warning.
+            match ethereum_adapter
+                .validate_start_block(logger, block_number, source_address)
+                .wait()
             {
-                Some(EthereumBlockPointer::from(full_block))
-            } else {
-                None
+                Ok((block, validated)) => {
+                    if !validated {
+                        warn!(
+                            logger,
+                            "Unable to confirm block number {} as the start block for the '{}' data source, address: {}",
+                            &block_number.to_string(),
+                            &data_source.name,
+                            source_address
+                                .map(|a| a.to_string())
+                                .unwrap_or("unspecified".to_string())
+                        );
+                    }
+                    Some(block)
+                },
+                Err(e) => {
+                    warn!(
+                        logger,
+                        "Start block number {} for the '{}' data source will not be used to filter the stream due to an error validating. Error: {}",
+                        &block_number.to_string(),
+                        &data_source.name,
+                        e
+                    );
+                    None
+                }
             }
         })
+        .filter(|&n| n.is_some())
         .collect();
 
     // Create deployment only if it does not exist already
     if !deployment_exists {
         let chain_head_block = chain_store.chain_head_ptr()?;
-        let genesis_block = dbg!(match starting_blocks {
-            Some(blocks) => EthereumBlockPointer::minimum_block_number(blocks),
-            None => chain_store.genesis_block_ptr().ok(),
-        }
-        .unwrap());
+        let start_blocks = start_blocks.unwrap();
+        let genesis_block = if start_blocks.is_empty() {
+            chain_store.genesis_block_ptr()?
+        } else {
+            EthereumBlockPointer::earliest_block(start_blocks)
+                .or_else(|| chain_store.genesis_block_ptr().ok())
+                .unwrap()
+        };
+        info!(
+            logger,
+            "Set the earliest available Ethereum block, start_block: {}", genesis_block.number
+        );
         ops.extend(
             SubgraphDeploymentEntity::new(&manifest, false, false, genesis_block, chain_head_block)
                 .create_operations(&manifest.id),
--- a/core/src/subgraph/registrar.rs
+++ b/core/src/subgraph/registrar.rs
@@ -526,6 +526,7 @@ fn resolve_subgraph_chain_blocks(
             .data_sources
             .clone()
             .into_iter()
+            .map(|data_source| data_source.source.start_block)
             .min()
             .map(|block_number| {
                 ethereum_adapter
@@ -913,7 +914,7 @@ fn create_subgraph_version(
                         .create_operations(&manifest.id),
                     );
                     deployment_store
-                        .create_subgraph_deployment(&logger, &manifest.schema, ops)
+                        .create_subgraph_deployment(&manifest.schema, ops)
                         .map_err(|e| SubgraphRegistrarError::SubgraphDeploymentError(e))
                 }
             })
--- a/core/src/subgraph/registrar.rs
+++ b/core/src/subgraph/registrar.rs
@@ -507,7 +507,7 @@ fn create_subgraph(
     Ok(CreateSubgraphResult { id: entity_id })
 }
 
-/// Resolves the chain head block and subgraph start block
+/// Resolves the chain head block and the subgraph's earliest block
 fn resolve_subgraph_chain_blocks(
     manifest: SubgraphManifest,
     chain_store: Arc<impl ChainStore>,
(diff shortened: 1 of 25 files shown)


graph/src/components/ethereum/types.rs
```diff
@@ -276,6 +277,19 @@ impl EthereumBlockPointer {
...
+            .iter()
+            .fold(None, |a: Option<EthereumBlockPointer>, &b| match a {
+                Some(a) => match PartialOrd::partial_cmp(&a.number, &b.number) {
+                    None => None,
+                    Some(Ordering::Less) => Some(a),
+                    Some(_) => Some(b),
+                },
+                None => Some(b),
+            })
+    }
```
leoyvens: Would it be possible to use [min_by_key](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.min_by_key) here?

graph/src/components/ethereum/adapter.rs
```diff
@@ -343,6 +343,14 @@ pub trait EthereumAdapter: Send + Sync + 'static {
         block: Block<Transaction>,
     ) -> Box<dyn Future<Item = EthereumBlock, Error = EthereumAdapterError> + Send>;
 
+    /// Load full information for the specified `block number` (in particular, transaction receipts).
```
leoyvens: This comment got copy pasted.

store/postgres/src/subgraphs.graphql
```diff
@@ -73,6 +73,7 @@ type EthereumContractSource @entity {
     id: ID!
     address: String!
     abi: String!
+    start_block: String
```
leoyvens: Shouldn't this be `Int`?

core/src/subgraph/registrar.rs
```diff
@@ -668,10 +679,79 @@ fn create_subgraph_version(
...
+                },
+                Err(e) => {
+                    warn!(
+                        logger,
+                        "Unable to validate start block for '{}', block_num: {}, error: {}",
+                        &block_number.to_string(),
+                        &data_source.name,
+                        e
+                    );
+                    None
```
leoyvens: An error here should be propagated up, not ignored.

core/src/subgraph/registrar.rs
```diff
@@ -668,10 +679,79 @@ fn create_subgraph_version(
...
+                    return None
+                },
+            };
+
+            // Validate that the start block provided for the Data Source is indeed the block that
+            // contains the contract creation transaction. Allow the block to be used as the
+            // subgraph level start block regardless, but emit a warning.
+            match ethereum_adapter
+                .validate_start_block(logger, block_number, source_address)
+                .wait()
```
leoyvens: Don't use `.wait()`, it tends to hang. Turning the whole function into a future might is annoying but we'll have to do it if there's no other way of avoiding `.wait()`.

core/src/subgraph/registrar.rs
```diff
@@ -668,10 +679,79 @@ fn create_subgraph_version(
...
+        })
+        .filter(|&n| n.is_some())
+        .collect();
+
     // Create deployment only if it does not exist already
     if !deployment_exists {
         let chain_head_block = chain_store.chain_head_ptr()?;
-        let genesis_block = chain_store.genesis_block_ptr()?;
+        let start_blocks = start_blocks.unwrap();
+        let genesis_block = if start_blocks.is_empty() {
```
leoyvens: Since this variable might not be the genesis block, lets rename it to something else.

core/src/subgraph/registrar.rs
```diff
@@ -668,10 +679,79 @@ fn create_subgraph_version(
         },
     });
 
+    let start_blocks: Option<Vec<EthereumBlockPointer>> = manifest
```
leoyvens: `Option<Vec<_>>` is always weird, can this be just `Vec<_>`?

leoyvens:
We've agreed that the desired behaviour is for the subgraph to start at the minimum start block, and for each data source to start triggering at their respective start block. The second part has not been done in this PR, to avoid that becoming a future breaking change we should either do it or make the warning a hard error until we resolve this limitation.

core/src/subgraph/registrar.rs
```diff
@@ -668,10 +679,79 @@ fn create_subgraph_version(
...
+            EthereumBlockPointer::earliest_block(start_blocks)
+                .or_else(|| chain_store.genesis_block_ptr().ok())
+                .unwrap()
+        };
+        info!(
+            logger,
+            "Set the earliest available Ethereum block, start_block: {}", genesis_block.number
+        );
         ops.extend(
             SubgraphDeploymentEntity::new(&manifest, false, false, genesis_block, chain_head_block)
```
leoyvens: Have you tested this to work? I don't get how the start block here gets picked up by the block stream.

fordN: Ya, I've tested it successfully. 
Store.block_ptr() picks up the `latestEthereumBlock` and uses it to reconcile the next step.   It's tough to grep the codebase for it because of the switching between camel and snake case between our database naming and rust field naming.

core/src/subgraph/registrar.rs
```diff
@@ -668,10 +679,79 @@ fn create_subgraph_version(
...
+        .into_iter()
+        .map(|data_source| {
+            let source_address = data_source.source.address;
+            let block_number = match data_source
+                .source
+                .start_block
+                .and_then(|block_str| block_str.parse::<u64>().ok())
+            {
+                Some(number) => number,
+                None => {
```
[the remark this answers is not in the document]
fordN: No, though originally I did use this for that purpose.  I've since made it more flexible; it doesn't end the iterator it just returns none for the current map then I filter out any `None` values in the next step: `.filter(|&n| n.is_some())`.

core/src/subgraph/registrar.rs
leoyvens: Ah I see now, thanks. But I believe this will actually start from `genesis_block + 1`, because the subgraph ptr is assumed to have been processed. This is really a more fundamental bug because it means we never process the genesis block. I suppose that's hard to fix, but it's important to work around this for the start block.

fordN:
@leoyvens sounds good. 
In order for each data source to start triggering at its respective start block I'm planning on updating the way we create and use EthereumFilters to be filtered by an earliest Ethereum block.

leoyvens:
@fordN That's seems like a smart way of doing it!

leoyvens:
We should also go ahead and remove `ETHEREUM_START_BLOCK`, that was always a hack.

graph/src/components/ethereum/adapter.rs
```diff
@@ -269,7 +269,7 @@ impl EthereumLogFilter {
 
 #[derive(Clone, Debug)]
 pub struct EthereumCallFilter {
-    pub contract_addresses_function_signatures: HashMap<Address, HashSet<[u8; 4]>>,
+    pub contract_addresses_function_signatures: HashMap<Address, (u64, HashSet<[u8; 4]>)>,
```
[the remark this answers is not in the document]
leoyvens: Maybe we want to make this as struct with start block and signature.

graph/src/components/ethereum/adapter.rs
```diff
@@ -331,23 +334,43 @@ impl EthereumCallFilter {
         } = self;
         contract_addresses_function_signatures.is_empty()
     }
+
+    pub fn only_activated_filters(self, current_block: u64) -> Self {
```
leoyvens: This function looks dead.

graph/src/components/ethereum/adapter.rs
```diff
@@ -331,23 +334,43 @@ impl EthereumCallFilter {
...
-                set
+        let mut lookup: HashMap<Address, (u64, HashSet<[u8; 4]>)> = HashMap::new();
+        iter.into_iter()
+            .for_each(|(start_block, address, function_signature)| {
+                if !lookup.contains_key(&address) {
+                    lookup.insert(address, (start_block, HashSet::default()));
+                }
+                lookup.get_mut(&address).map(|set| {
+                    set.1.insert(function_signature);
+                    set
```
leoyvens: You also need to check for a larger start block here.

leoyvens: This also applies to the `extend` method.

graph/src/components/ethereum/adapter.rs
```diff
@@ -360,15 +383,15 @@ impl From<EthereumBlockFilter> for EthereumCallFilter {
             contract_addresses_function_signatures: ethereum_block_filter
                 .contract_addresses
                 .into_iter()
-                .map(|address| (address, HashSet::default()))
-                .collect::<HashMap<Address, HashSet<[u8; 4]>>>(),
+                .map(|(start_block_opt, address)| (address, (start_block_opt, HashSet::default())))
+                .collect::<HashMap<Address, (u64, HashSet<[u8; 4]>)>>(),
```
leoyvens: This will take the last `start_block` value, but the correct behaviour is the minimum.

graph/src/components/ethereum/adapter.rs
```diff
@@ -412,6 +438,14 @@ impl EthereumBlockFilter {
         self.trigger_every_block = self.trigger_every_block || other.trigger_every_block;
         self.contract_addresses.extend(other.contract_addresses);
```
leoyvens: Again, extending is not the right way to handle the start block.

graph/src/components/ethereum/adapter.rs
```diff
@@ -412,6 +438,14 @@ impl EthereumBlockFilter {
...
     }
+
+    pub fn start_blocks(&self) -> Vec<u64> {
+        self.contract_addresses
+            .iter()
```
leoyvens: You can chain a `.cloned()` here to avoid the `&` and `*` dance.

graph/src/components/ethereum/adapter.rs
```diff
@@ -270,6 +270,8 @@ impl EthereumLogFilter {
 
 #[derive(Clone, Debug)]
 pub struct EthereumCallFilter {
+    // Each call filter has a map of filters keyed by adress, each containing a tuple with
```
leoyvens: s/adress/address

graph/src/data/subgraph/schema.rs
```diff
@@ -712,6 +715,7 @@ impl TryFromValue for EthereumContractSourceEntity {
         Ok(Self {
             address: map.get_optional("address")?,
             abi: map.get_required("abi")?,
+            start_block: map.get_required("startBlock")?,
```
leoyvens: This needs to be `get_optional("startBlock").unwrap_or_default()`.

leoyvens:
Played around with this a bit, it's working great! There's one bikeshed I'd like to decide while we have the chance, do we want `startBlock` to be under `source` or directly in the data source? Personally I favor a flatter layout with `startBlock` in the data source.

core/src/subgraph/registrar.rs
```diff
@@ -651,96 +728,182 @@ fn create_subgraph_version(
...
+                manifest.clone(),
+                chain_store.clone(),
+                ethereum_adapter.clone(),
+                &logger.clone(),
+            )
+            .and_then(move |(chain_head_block, earliest_block)| {
+                info!(
+                    logger,
+                    "Set the earliest available Ethereum block, start_block: {}",
+                    earliest_block.number + 1
```
Jannis: This line confuses me. I'm assuming `earliest_block` is the smallest start block across all data sources. Shouldn't we then start `earliest_block.number` by setting the "latest block" to `earliest_block.number - 1` when creating the deployment?

Jannis: Or is this completely unrelated to data source start blocks?

datasource/ethereum/src/block_stream.rs
```diff
@@ -379,9 +385,21 @@ where
...
+                            .filter(|block_num| block_num > &from)
+                            .min()
+                            .unwrap_or(std::u64::MAX);
+
+                        // End either just before the the next data source start_block or
+                        // just prior to the reorg threshold. It isn't safe to go any farther
+                        // due to race conditions.
+                        let to_limit = cmp::min(
+                            head_ptr.number - reorg_threshold,
+                            next_start_block - 1,
```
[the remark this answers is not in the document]
fordN: In the block just above that resolves the `next_start_block` we filter out any values before the `from` value, so it cannot be lower than the `from` value or genesis block.

Sledro:
> We should also go ahead and remove `ETHEREUM_START_BLOCK`, that was always a hack.

Is there another way to define the start block for the node? Our L2 chain starts with block 1 not 0 resulting in the following error: `Server error: could not find block in the database: 0`

Issue #615, opened 2018-11-29
Validate that entities put in the store contain only fields that exist in the schema


leoyvens:
Right now we'll hapilly put anything the mapping sends us into the store without validating that the fields actually belong to the schema for that entity. See https://github.com/graphprotocol/graph-node/issues/600 for a case of this happening due to a bug. But could also happen due to user error.

We should also validate that the fields can be converted to the type declared in the schema.

Jannis:
Yeah, this would be lovely to catch. It would require schema knowledge in e.g. the host `WasmiModule` so that we can validate `store.set` operations. Not _that_ hard I imagine, the only thing I'm not sure is how we'd best pass the schema to the module.

leodasvacas:
@Jannis I thought a bit about this. Since the store will soon become the source of truth for schemas, then `store.set` will have easy access to the schema and can do the validation.

Jannis:
@leodasvacas Will it? In what way?

lutter:
This has crept up again in the context of https://github.com/graphprotocol/graph-ts/issues/79

I'll add some validation; in particular, this will check for a `store.set` that

* the entity type actually exists
* the values being set actually have the proper type for the field
* non-nullable fields are not set to null

PR #1268, opened 2019-10-02
Validate entities on `store.set`

lutter:
These changes make it so that we now validate entities against the subgraph schema when mappings call `store.set`.

This is a user-visible change, and can lead to subgraphs that were previously working to now fail because of the additional validation - those subgraphs would have produced errors when invaid data was queried, but syncing would have seemingly worked.

Fixes https://github.com/graphprotocol/graph-node/issues/615

### DIFF
--- a/runtime/wasm/src/host_exports.rs
+++ b/runtime/wasm/src/host_exports.rs
@@ -174,7 +174,7 @@ where
         let result = ctx
             .state
             .entity_cache
-            .get(self.store.as_ref(), store_key)
+            .get(self.store.as_ref(), &store_key)
             .map_err(HostExportError)
             .map(|ok| ok.to_owned());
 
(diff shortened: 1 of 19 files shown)


runtime/wasm/src/host_exports.rs
```diff
@@ -139,7 +141,20 @@ where
...
-        ctx.state.entity_cache.set(key, Entity::from(data));
+
+        // Validate the changes against the subgraph schema
+        let updates = Entity::from(data);
+        let mut entity = ctx
+            .state
+            .entity_cache
+            .get(self.store.as_ref(), &key)
+            .map_err(|e| HostExportError(e.to_string()))?
+            .unwrap_or_else(|| Entity::new());
```
leoyvens: What would you think of doing the `ctx.state.entity_cache.set(key, updates);` first, then you can just unwrap here, and the `entity.merge(updates.clone());` can be removed because it will be done in the entity cache.

lutter: Good idea. If validation fails, the subgraph is toast anyway, and it doesn't matter what winds up in the `EntityCache`

lutter:
I changed it so that validation only happens for relational schemas.

PR #1270, opened 2019-10-04
Cache contract view function calls

### LINKED ISSUE #1230: Cache Ethereum contract call results

leoyvens:
Resolves #1230. This mostly just follows the plan in the issue, I got a bit creative with the query to look up a cached result and date of last access.

So far I've tested this with Moloch. The cache works and is correct, but the speedup was not so impressive on that subgraph, I want to try others that are heavier on calls, but we can do that on staging. One thing interesting about Moloch is that their return values are quite big, between 400-600 bytes, higher than expected but should be ok.


graph/src/components/store.rs
```diff
@@ -1047,6 +1047,25 @@ pub trait ChainStore: Send + Sync + 'static {
     ) -> Result<Option<EthereumBlock>, Error>;
 }
 
+pub trait EthereumCallCache: Send + Sync + 'static {
+    /// Cached return value.
+    fn get_call(
```
lutter: I was going to say that `get` (and `set`) would be enough here, seeing how the trait is called `EthereumCallCache`, but I assume that you called it `get_call` so it doesn't conflict with `Store.get`?

store/postgres/src/store.rs
```diff
@@ -1220,6 +1220,94 @@ impl ChainStore for Store {
     }
 }
 
+impl EthereumCallCache for Store {
+    fn get_call(
+        &self,
+        contract_address: ethabi::Address,
+        encoded_call: &[u8],
+        block: EthereumBlockPointer,
+    ) -> Result<Option<Vec<u8>>, Error> {
```
lutter: The DB work should happen inside a transaction.

store/postgres/src/store.rs
```diff
@@ -1220,6 +1220,94 @@ impl ChainStore for Store {
...
+        use crate::db_schema::{eth_call_cache, eth_call_meta};
+        use diesel::dsl::sql;
+
+        let id = contract_call_id(contract_address, encoded_call, block);
+        let conn = &*self.get_conn()?;
+
+        if let Some((return_value, update_accessed_at)) = eth_call_cache::table
+            .find(id.as_ref())
+            .inner_join(eth_call_meta::table)
```
lutter: The db schema is missing an index on `eth_call_cache.contract_address`; but that index would be pretty terrible, since there will be a lot of entries with the same contract address.

It might be better to get even fancier with this query by doing something like
```sql
  select return_value,
             (select current_date > accessed_at
                from eth_call_meta m
               where m.contract_address = c.contract_address)
   from eth_call_cache c
  where c.id = $id
```

That way, we avoid the join, don't need the (not very useful) index, and do all lookups on primary keys.

store/postgres/src/store.rs
```diff
@@ -1220,6 +1220,94 @@ impl ChainStore for Store {
...
+        block: EthereumBlockPointer,
+        return_value: &[u8],
+    ) -> Result<(), Error> {
+        use crate::db_schema::{eth_call_cache, eth_call_meta};
+        use diesel::dsl::sql;
+
+        let id = contract_call_id(contract_address, encoded_call, block);
+        let conn = &*self.get_conn()?;
+
+        insert_into(eth_call_cache::table)
```
lutter: The DB work here should go inside a transaction.

lutter (approved):
Good stuff!

lutter (changes requested):
Good stuff! (and now with the right review status)

graph/src/components/store.rs
leoyvens: That's exactly what happened, I tried just `get` but it started conflicting.

store/postgres/src/store.rs
leoyvens: I decided to not fear the join because I imagined the query as it is would be executed like the query you describe. The "nested loop join" should be the obvious choice for the planner, and it works exactly like that query, I don't think that index would be used by this join.

leoyvens:
Put the db work in a transaction and rebased.

store/postgres/src/store.rs
lutter: We should check that that is what actually happens once we have some real-world data in these tables; but it's a reasonable assumption.

lutter (approved):

datasource/ethereum/src/ethereum_adapter.rs
```diff
@@ -978,20 +979,43 @@ where
...
+                                    error!(logger, "call cache set error";
+                                                   "error" => e.to_string())
+                                });
+                            result.0
+                        }),
+                    )
+                }
+            }
+            // Decode the return values according to the ABI
+            .and_then(move |output| call.function.decode_output(&output).map_err(From::from)),
```
Jannis: I wonder if we should maybe only cache the result if wee succeed in decoding the outputs, to avoid caching invalid results caused e.g. by bugs in the Ethereum node. Because otherwise we might be forced to delete call results from the cache manually under certain situations.

halink0803: Sorry to bring this back but we got the problem with cached of invalid (failed) response from eth_call, and this behavior make retry meaningless and subgraph stuck at the block of eth_call failed. I think it is better to have the option to not save the result if the eth_call failed or cannot be decoded. Otherwise, there should be an easier way to clear the call_cached.

azf20: hey @halink0803 - is this on the hosted service (if so can you share the specific subgraph and cached call), or is this on a local Graph Node?

halink0803: @azf20 it happened in both hosted service and my local graph-node also, because in the graph-node we call eth_call to rpc and sometimes, rpc returns with the bad result (empty or revert), here is the subgraph I got the issue on hosted service:  https://thegraph.com/hosted-service/subgraph/halink0803/kyberswap-exchange-bsc, I also tried on local graph-node and my solution was to clear the call_cache.

leoyvens: @halink0803 call reverts appear in the logs, but for that subgraph I'm not seeing any call reverts logged.

halink0803: @leoyvens I don't get it. How the call reverts appear in the log but you are not seeing any call reverts logged? From the eth_call and the call_cache, it is actually show with empty response instead of throwing call error.

halink0803: How could I make the subgraph resume after this? Is this from subgraph code logic create the issue? It is showed eth_call reverted.

leoyvens: Could you post the exact log line you are seeing, in the hosted service, indicating that a revert happened?

PR #1277, opened 2019-10-08
Limit number of Ethereum request retries (defaults to 10)

### LINKED ISSUE #1276: Don't retry Ethereum calls like `eth_getLogs` forever
We put this in place to get past non-deterministic Infura errors, but retrying forever means subgraphs can end up in infinite loops without failing, and without it being noticed by developers.

We should retry maybe 10-20 times (we can make it configurable with an environment variable) and then fail the subgraph. This will surface problems better.

Jannis:
Resolves #1276.

This adds a new `GRAPH_ETHEREUM_REQUEST_RETRIES` environment variable that defines the number of times Graph Node retries requests made to Ethereum. This avoids retrying requests indefinitely, which can lead to subgraphs being stuck and problems not being noticed.

### DIFF
--- a/datasource/ethereum/src/ethereum_adapter.rs
+++ b/datasource/ethereum/src/ethereum_adapter.rs
@@ -40,6 +40,12 @@ lazy_static! {
         .unwrap_or("500".into())
         .parse::<u64>()
         .expect("invalid number of parallel Ethereum block ranges to scan");
+
+    /// Number of retries to attempt for Ethereum requests. Defaults to 10.
+    static ref REQUEST_RETRIES: usize = std::env::var("GRAPH_ETHEREUM_REQUEST_RETRIES")
+        .unwrap_or("10".into())
+        .parse::<usize>()
+        .expect("invalid number of Ethereum request retries");
 }
 
 impl<T> EthereumAdapter<T>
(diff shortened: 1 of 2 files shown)


Jannis:
Now that I've created the PR I'm already reconsidering it.

The indefinite `eth_getLogs` retries allowed the 2key subgraphs to recover once Alchemy fixed the issue that caused requests to fail, without us having to do anything. If the subgraphs had failed, we would've had to restart the graph-node manually.

Instead of failing after e.g. 10 retries like this PR suggests, I think it would be better to add probes for problems like indefinite retries and expose them to developers (e.g. on a health tab in the explorer subgraph details). @olibearo Do you think that would be possible with Prometheus?

Jannis:
@lutter suggested we retry with exponential fallback (until the retry interval is like 5s) and log a warning as soon as we're reaching e.g. that max retry interval of 5s, and expose the problem to users that way.

lutter:
Sounds good. I would make the maximum wait time longer though, something like 5 minutes, in case a lot of subgraphs hit problems that need retrying, so that we don't do too much work with the retries.

leoyvens:
We ended up doing this in #1396.

PR #1278, opened 2019-10-09
graph: Allow interfaces as the type of a field from which to derive

lutter:


graph/src/data/graphql/validation.rs
```diff
@@ -245,13 +291,18 @@ type D @entity { id: ID! }
...
+type Transaction @entity { from: Address! }
+interface Address { txn: Transaction! @derivedFrom(field: \"from\") }
+type Account implements Address @entity { id: ID!, txn: Transaction! @derivedFrom(field: \"from\") }";
 
     fn validate(field: &str, errmsg: &str) {
         let raw = format!("type A @entity {{ id: ID!\n {} }}\n{}", field, OTHER_TYPES);
 
         let document = graphql_parser::parse_schema(&raw).expect("Failed to parse raw schema");
-        match validate_derived_from(&document) {
+        match dbg!(validate_derived_from(&document)) {
```
Jannis: Let's remove the `dbg!` here 😉

lutter: Woops. Fixed

PR #1291, opened 2019-10-16
Notify of store events in separate transaction

leoyvens:
Right now we sent the notifications in the same transaction as all the entity operations. This creates a probable point of contention between subgraphs, and it seems fine to transact the notifications separately.


lutter (approved):
Looks great!

PR #1293, opened 2019-10-17
store: Make sure query generation doesn't fail

lutter:
We had a few places were query generation could fail because of an unknown
table or column. That then makes it impossible to print the failed query
for debugging purposes.

This commit changes things so that these errors are detected before we
construct QueryFragments.

### DIFF
--- a/store/postgres/src/relational_queries.rs
+++ b/store/postgres/src/relational_queries.rs
@@ -11,7 +11,6 @@ use diesel::query_dsl::{LoadQuery, RunQueryDsl};
 use diesel::result::QueryResult;
 use diesel::sql_types::{Array, Binary, Bool, Integer, Jsonb, Numeric, Range, Text};
 use diesel::Connection;
-use failure::Fail;
 use std::convert::TryFrom;
 use std::str::FromStr;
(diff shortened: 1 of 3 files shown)


store/postgres/src/relational_queries.rs
```diff
@@ -896,8 +936,7 @@ impl<'a, Conn> RunQueryDsl<Conn> for ConflictingEntityQuery<'a> {}
 #[derive(Debug, Clone, Constructor)]
 pub struct FilterQuery<'a> {
     schema: &'a str,
-    tables: Vec<&'a Table>,
-    filter: Option<EntityFilter>,
+    tables: Vec<(&'a Table, Option<QueryFilter<'a>>)>,
```
leoyvens: It's a bit confusing that this field is tables and filters but is called just `tables`. Would it be too horrible to give this a longer name?

lutter: Good point. I was just being lazy renaming things. Renamed this to `table_and_filter_pairs`

lutter:
Rebased on top of latest master

PR #1305, opened 2019-10-22
store: Handle empty and/or filters in the relational schema

lutter:
Empty and/or clauses would generate a `()` which leads to SQL syntax
errors. This behavior is a backstop; what should really happen is that
and/or clauses always have at least one predicate in them.

With the previous behavior, we would generate queries like `select ... from
... where ()` which is not valid SQL.

### DIFF
--- a/store/postgres/src/relational_queries.rs
+++ b/store/postgres/src/relational_queries.rs
@@ -409,16 +409,21 @@ impl<'a> QueryFilter<'a> {
         &self,
         filters: &Vec<EntityFilter>,
         op: &str,
+        on_empty: &str,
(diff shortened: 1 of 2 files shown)


leoyvens:
What would be the ideal fix and why can't we do it now?

lutter:
The ideal fix would be to panic when we create an empty and/or filter since it indicates some logic bug. The only way I can think of enforcing that is to switch to a DSL for constructing filters that checks that the arguments to `and` and `or` are not empty vec's.

leoyvens:
I suppose rejecting these queries might break users in the wild, so it might be best to deal with them somehow, as this PR does.

PR #1307, opened 2019-10-22
Fix following interface references in queries (#1303)

Jannis:
This wasn't a problem with JSONB schemas, as all entities were in a single table. With relational schemas, however, we're resolving entity types into table names and we didn't do that correctly when following non-derived individual interface relationships.

E.g. given a schema

```graphql
interface Legged { legs: Int! }
type Animal implements Legged @entity {
  id: ID!
  parent: Legged!
}
```

the following query would fail with `Invalid entity unknown table 'Legged'`:

```graphql
{
  animal {
    parent { id }
  }
}
```

This commit changes the logic in `StoreResolver::resolve_object` to ensure that interfaces are always resolved to entity type names rather than being used as the relationship target types directly.

### DIFF
--- a/core/tests/interfaces.rs
+++ b/core/tests/interfaces.rs
@@ -201,6 +201,44 @@ fn reference_interface() {
 }
 
 #[test]
+fn follow_interface_reference() {
+    let subgraph_id = "FollowInterfaceReference";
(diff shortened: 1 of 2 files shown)


lutter (approved):
LGTM

Issue #902, opened 2019-04-30
Historical Block Scanning For Dynamic Data Sources


dOrgJelli:
**Do you want to request a *feature* or report a *bug*?**
Feature.

**What is the current behavior?**
From https://github.com/graphprotocol/graph-node/pull/832 :
"Whenever we we process a block and mappings request new data sources, these data sources are collected and, after having processed the block, are instantiated from templates. **We then process the current block again but only with those new data sources.** The entity operations from this are merged into the ones we already have for the block. After that, the dynamic data sources are persisted by adding the data sources to the subgraph instance and by adding entity operations to store them in the db (for 1.)."

**What is the desired behavior?**
".... We then process all blocks in the range of [contract creation -> current block] for each newly created data source."

**More explanation.**
In the DAOstack protocol, new DAOs can be created in a number of different ways. In order to support this (and minimize spam), the desired behavior for the subgraph is to only index DAOs that're added to a universal registry of "accepted" DAOs. Given the graph-node's current functionality, a lot of information about these newly registered DAOs would never be read/processed/stored because it has taken place prior to the current block. In order to support this scenario (which IMO is bound to be commonplace for other projects), being able to process the entire block history for a given contract upon instantiating it as a data source would be ideal.

Thread here: https://github.com/daostack/subgraph/issues/197

leodasvacas:
This is a use case that's likely to come up again and we'll want to support it in some way, however the design and implementation is not simple. It sounds like you can solve this in a different, less ideal way, so I recommend you take that route while we figure this one out.

Jannis:
The three main parts that are difficult about this:

1. There's currently one block stream per subgraph. It only moves forward. When a data source template is instantiated, the block stream is restarted with the additional filters from the new data sources. Making this stream go back and process all past events of a new data source will be complicated. It may be better to spawn separate block streams for each data source but have them synced somehow.

2. Processing past events would require accessing the store at different points in the past. This is not currently possible. Even if it was, it could introduce a time travel problem, where changes generated by new data source B on an older block would have changed the behavior of an already existing data source A.

3. Chain reorgs. A block in which a dynamic data source is created may be removed from the chain again later. When that happens, the dynamic data source needs to be removed again and all its data from that block needs to be removed as well. Now, if we include all past events of such a new data source, we'd have to also remove all entity changes from those past events.

dOrgJelli:
[...]

PR #1308, opened 2019-10-22
Pull in transaction receipts only when necessary

leoyvens:
When processing blocks considered final, we fetch the triggers only to drop them and re-process them from the transaction receipts. This is a wasteful process, particularly of disk space because we cache hundreds of GBs worth of transaction receipts we don't really need.

This PR makes it so that we only request and store transaction receipts when they are actually necessary, which is when processing blocks that may still be reorged. `blocks_with_triggers` now returns richer information instead of dropping it, the return type was changed from `EthereumBlockPointer` to `EthereumBlockWithTriggers`.

One thing that makes this change tricky is that we may need to re-process a block for dynamic data sources, at which point we again need to know if the block is final or not, and need extra information for non-final blocks. This is encoded in the `BlockFinality` enum and the trigger re-processing is encapsulated `triggers_in_block`. Some code got moved from the block stream to the ethereum adapter so that it could be shared between the block stream and the instance manager.

Since this touched some of the block stream, I took the opportunity to refactor it a bit. We no longer have special logic for advancing through an empty range, instead we insert a dummy block for `to` with no triggers, and use the normal processing logic to advance. The special casing was in some sense an optimization, but it was optimizing the fastest possible case which is an empty range, so it was considerable complexity for no real gain. In doing this, `transact_block_operations` got a bit simpler, it was only used to advance one block a time when it's perfectly capable of skipping blocks. Also it no longer takes a `block_ptr_from`, it doesn't make sense for that to be anything other than the current block pointer, so we just assume that. Also, some of the types in the block stream state machine got simplified a bit.

Sorry if this is a lot in one PR! Please ask if some change needs clarification or more comments. I tested this with Moloch and Betoken, both synced fine. I avoided any non-trivial change to reversion code, any changes there are just formatting and minor refactors.

### DIFF
--- a/core/src/subgraph/instance_manager.rs
+++ b/core/src/subgraph/instance_manager.rs
@@ -50,6 +50,8 @@ struct IndexingContext<B, T: RuntimeHostBuilder, S> {
 
     /// Sensors to measue the execution of eth rpc calls
     pub ethrpc_metrics: Arc<SubgraphEthRpcMetrics>,
+
+    pub block_stream_metrics: Arc<BlockStreamMetrics>,
 }
 
 pub struct SubgraphInstanceManager {
@@ -354,6 +356,11 @@ impl SubgraphInstanceManager {
             registry.clone(),
             deployment_id.to_string(),
         ));
+        let block_stream_metrics = Arc::new(BlockStreamMetrics::new(
+            registry.clone(),
+            ethrpc_metrics.clone(),
+            deployment_id.clone(),
+        ));
         let instance =
             SubgraphInstance::from_manifest(&logger, manifest, host_builder, host_metrics.clone())?;
 
@@ -381,6 +388,7 @@ impl SubgraphInstanceManager {
             subgraph_metrics,
             host_metrics,
             ethrpc_metrics,
+            block_stream_metrics,
         };
 
         // Keep restarting the subgraph until it terminates. The subgraph
@@ -453,7 +461,7 @@ where
             ctx.state.call_filter.clone(),
             ctx.state.block_filter.clone(),
             ctx.inputs.templates_use_calls,
-            ctx.ethrpc_metrics.clone(),
+            ctx.block_stream_metrics.clone(),
         )
         .from_err()
         .cancelable(&block_stream_canceler, || CancelableError::Cancel);
--- a/core/src/subgraph/instance_manager.rs
+++ b/core/src/subgraph/instance_manager.rs
@@ -608,7 +608,7 @@ where
     }
 
     // Obtain current and new block pointer (after this block is processed)
-    let thin_block = Arc::new(block.thin_block());
+    let light_block = Arc::new(block.light_block());
     let block_ptr_after = EthereumBlockPointer::from(&block);
     let block_ptr_for_new_data_sources = block_ptr_after.clone();
 
@@ -620,7 +620,7 @@ where
         logger.clone(),
         ctx,
         BlockState::default(),
-        thin_block.clone(),
+        light_block.clone(),
         triggers,
     )
     .and_then(move |(ctx, block_state)| {
@@ -658,7 +658,7 @@ where
                 // Reprocess the triggers from this block that match the new data sources
                 let logger = logger.clone();
                 let logger1 = logger.clone();
-                let thin_block = thin_block.clone();
+                let light_block = light_block.clone();
                 Box::new(
                     eth_adapter
                         .clone()
@@ -706,7 +706,7 @@ where
                                         SubgraphInstance::<T>::process_trigger_in_runtime_hosts(
                                             &logger,
                                             runtime_hosts.iter().cloned(),
-                                            thin_block.clone(),
+                                            light_block.clone(),
                                             trigger,
                                             block_state,
                                         )
--- a/core/src/subgraph/instance_manager.rs
+++ b/core/src/subgraph/instance_manager.rs
@@ -601,7 +601,6 @@ where
 
     // Obtain current and new block pointer (after this block is processed)
     let thin_block = Arc::new(block.thin_block());
-    let block_ptr_now = thin_block.parent_ptr();
     let block_ptr_after = EthereumBlockPointer::from(&block);
     let block_ptr_for_new_data_sources = block_ptr_after.clone();
 
@@ -741,12 +740,7 @@ where
         let start = Instant::now();
         ctx.inputs
             .store
-            .transact_block_operations(
-                ctx.inputs.deployment_id.clone(),
-                block_ptr_now,
-                block_ptr_after,
-                mods,
-            )
+            .transact_block_operations(ctx.inputs.deployment_id.clone(), block_ptr_after, mods)
             .map(|should_migrate| {
                 let elapsed = start.elapsed().as_secs_f64();
                 metrics.block_ops_transaction_duration.observe(elapsed);
--- a/core/src/subgraph/instance_manager.rs
+++ b/core/src/subgraph/instance_manager.rs
@@ -650,11 +650,14 @@ where
 
                 // Reprocess the triggers from this block that match the new data sources
                 let logger = logger.clone();
+                let logger1 = logger.clone();
                 let thin_block = thin_block.clone();
                 Box::new(
                     eth_adapter
+                        .clone()
                         .triggers_in_block(
-                            &logger,
+                            logger,
+                            ctx.inputs.store.clone(),
                             ctx.ethrpc_metrics.clone(),
                             EthereumLogFilter::from_data_sources(data_sources.iter()),
                             EthereumCallFilter::from_data_sources(data_sources.iter()),
@@ -666,12 +669,12 @@ where
 
                             if triggers.len() == 1 {
                                 info!(
-                                    logger,
+                                    logger1,
                                     "1 trigger found in this block for the new data sources"
                                 );
                             } else if triggers.len() > 1 {
                                 info!(
-                                    logger,
+                                    logger1,
                                     "{} triggers found in this block for the new data sources",
                                     triggers.len()
                                 );
@@ -680,14 +683,14 @@ where
                             // Add entity operations for the new data sources to the block state
                             // and add runtimes for the data sources to the subgraph instance.
                             persist_dynamic_data_sources(
-                                logger.clone(),
+                                logger1.clone(),
                                 &mut ctx,
                                 &mut block_state.entity_cache,
                                 data_sources,
                                 block_ptr_for_new_data_sources,
                             );
 
-                            let logger = logger.clone();
+                            let logger = logger1.clone();
                             let thin_block = thin_block.clone();
                             Box::new(
                                 stream::iter_ok(triggers)
--- a/core/src/subgraph/instance_manager.rs
+++ b/core/src/subgraph/instance_manager.rs
@@ -502,9 +502,11 @@ where
         .fold(ctx, move |ctx, block| {
             let subgraph_metrics = ctx.subgraph_metrics.clone();
             let start = Instant::now();
-            subgraph_metrics
-                .block_trigger_count
-                .observe(block.triggers.len() as f64);
+            if block.triggers.len() > 0 {
+                subgraph_metrics
+                    .block_trigger_count
+                    .observe(block.triggers.len() as f64);
+            }
             process_block(
                 logger.clone(),
                 ctx.inputs.eth_adapter.clone(),
--- a/core/src/subgraph/instance_manager.rs
+++ b/core/src/subgraph/instance_manager.rs
@@ -1,7 +1,6 @@
 use futures::future::{loop_fn, Loop};
 use futures::sync::mpsc::{channel, Receiver, Sender};
 use std::collections::HashMap;
-use std::ops::Deref;
 use std::sync::RwLock;
 use std::time::Instant;
 use uuid::Uuid;
@@ -20,6 +19,7 @@ struct IndexingInputs<B, S> {
     network_name: String,
     start_blocks: Vec<u64>,
     store: Arc<S>,
+    eth_adapter: Arc<dyn EthereumAdapter>,
     stream_builder: B,
     templates_use_calls: bool,
     top_level_templates: Vec<DataSourceTemplate>,
@@ -162,6 +162,7 @@ impl SubgraphInstanceManager {
     pub fn new<B, S, M>(
         logger_factory: &LoggerFactory,
         stores: HashMap<String, Arc<S>>,
+        eth_adapters: HashMap<String, Arc<dyn EthereumAdapter>>,
         host_builder: impl RuntimeHostBuilder,
         block_stream_builder: B,
         metrics_registry: Arc<M>,
@@ -182,6 +183,7 @@ impl SubgraphInstanceManager {
             logger_factory,
             subgraph_receiver,
             stores,
+            eth_adapters,
             host_builder,
             block_stream_builder,
             metrics_registry.clone(),
@@ -198,6 +200,7 @@ impl SubgraphInstanceManager {
         logger_factory: LoggerFactory,
         receiver: Receiver<SubgraphAssignmentProviderEvent>,
         stores: HashMap<String, Arc<S>>,
+        eth_adapters: HashMap<String, Arc<dyn EthereumAdapter>>,
         host_builder: impl RuntimeHostBuilder,
         block_stream_builder: B,
         metrics_registry: Arc<M>,
@@ -236,6 +239,13 @@ impl SubgraphInstanceManager {
                                     .get(&n)
                                     .expect(&format!(
                                         "expected store that matches subgraph network: {}",
+                                        &n
+                                    ))
+                                    .clone(),
+                                eth_adapters
+                                    .get(&n)
+                                    .expect(&format!(
+                                        "expected eth adapter that matches subgraph network: {}",
                                         &n
                                     ))
                                     .clone(),
@@ -283,6 +293,7 @@ impl SubgraphInstanceManager {
         host_builder: impl RuntimeHostBuilder,
         stream_builder: B,
         store: Arc<S>,
+        eth_adapter: Arc<dyn EthereumAdapter>,
         manifest: SubgraphManifest,
         registry: Arc<M>,
     ) -> Result<(), Error>
@@ -346,6 +357,7 @@ impl SubgraphInstanceManager {
                 network_name,
                 start_blocks,
                 store,
+                eth_adapter,
                 stream_builder,
                 templates_use_calls,
                 top_level_templates,
@@ -478,6 +490,7 @@ where
                 .observe(block.triggers.len() as f64);
             process_block(
                 logger.clone(),
+                ctx.inputs.eth_adapter.clone(),
                 ctx,
                 block_stream_cancel_handle.clone(),
                 block,
@@ -548,6 +561,7 @@ where
 /// whether new dynamic data sources have been added to the subgraph.
 fn process_block<B, T: RuntimeHostBuilder, S>(
     logger: Logger,
+    eth_adapter: Arc<dyn EthereumAdapter>,
     ctx: IndexingContext<B, T, S>,
     block_stream_cancel_handle: CancelHandle,
     block: EthereumBlockWithTriggers,
@@ -556,13 +570,13 @@ where
     B: BlockStreamBuilder,
     S: ChainStore + Store + EthereumCallCache + SubgraphDeploymentStore,
 {
-    let calls = block.calls;
     let triggers = block.triggers;
     let block = block.ethereum_block;
 
+    let block_ptr = EthereumBlockPointer::from(&block);
     let logger = logger.new(o!(
-        "block_number" => format!("{:?}", block.block.number.unwrap()),
-        "block_hash" => format!("{:?}", block.block.hash.unwrap())
+        "block_number" => format!("{:?}", block_ptr.number),
+        "block_hash" => format!("{:?}", block_ptr.hash)
     ));
     let logger1 = logger.clone();
 
@@ -577,9 +591,9 @@ where
     }
 
     // Obtain current and new block pointer (after this block is processed)
-    let block = Arc::new(block);
-    let block_ptr_now = EthereumBlockPointer::to_parent(&block);
-    let block_ptr_after = EthereumBlockPointer::from(&*block);
+    let thin_block = Arc::new(block.thin_block());
+    let block_ptr_now = EthereumBlockPointer::from(&block).to_parent();
+    let block_ptr_after = EthereumBlockPointer::from(&block);
     let block_ptr_for_new_data_sources = block_ptr_after.clone();
 
     let metrics = ctx.subgraph_metrics.clone();
@@ -590,7 +604,7 @@ where
         logger.clone(),
         ctx,
         BlockState::default(),
-        block.clone(),
+        Arc::new(block.thin_block()),
         triggers,
     )
     .and_then(move |(ctx, block_state)| {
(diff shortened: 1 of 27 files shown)


store/postgres/src/store.rs
```diff
@@ -815,37 +815,13 @@ impl StoreTrait for Store {
...
         subgraph_id: SubgraphDeploymentId,
-        block_ptr_from: EthereumBlockPointer,
         block_ptr_to: EthereumBlockPointer,
         mods: Vec<EntityModification>,
     ) -> Result<bool, StoreError> {
-        // Sanity check on block numbers
-        if block_ptr_from.number != block_ptr_to.number - 1 {
-            panic!("transact_block_operations must transact a single block only");
-        }
+        let block_ptr_from = self.block_ptr(subgraph_id.clone())?;
```
lutter: I am all for getting rid of `set_block_ptr_with_no_changes`, but I don't like that this change causes a query for data the caller already has. We should continue to pass `block_ptr_from` into `transact_block_operations` but change the rules for it: if `block_ptr_from + 1 < block_ptr_to`, `mods` has to be empty, and we must always have `block_ptr_from < block_ptr_to`.

leoyvens: It's interesting observation that at some earlier point the block stream called made this same DB lookup, and we could plumb that through in order to save this call. But if we wanted to keep `block_ptr` cached in memory for performance, I'd approach that differently by having the store keep that cache without complicating other code. However this query should be cheap enough to do in both places, if it's currently slow that's because we're pulling the entire subgraph only to read the block pointer, which is an issue regardless of this change.

I can assert that `block_ptr_from < block_ptr_to`, though I don't follow why if `block_ptr_from + 1 < block_ptr_to` then `mods` must be empty, that's not the case in this PR, is that a problem?

datasource/ethereum/src/block_stream.rs
```diff
@@ -401,145 +306,107 @@ where
...
+                                        call_filter.clone(),
+                                        block_filter.clone(),
+                                    )
+                                    .map(move |descendant_blocks| {
                                         debug!(
-                                            ctx.logger,
-                                            "Found {} relevant block(s)",
-                                            descendant_ptrs.len()
+                                            logger,
+                                            "Found {} trigger(s)",
```
olibearo: I think it's useful here to log the number of blocks, don't we log triggers farther down the pipeline?

leoyvens: You're right, seems I got confused here. I've moved this log into `triggers_in_block` and it now logs the number of blocks again.

datasource/ethereum/src/block_stream.rs
```diff
@@ -651,97 +525,42 @@ where
...
-                    // Revert entity changes from this block, and update subgraph ptr.
-                    future::result(
-                        ctx.subgraph_store
-                            .revert_block_operations(
-                                ctx.subgraph_id.clone(),
-                                subgraph_ptr,
-                                parent_ptr,
+                Box::new(
+                    self.eth_adapter
+                        .load_block(&ctx.logger, subgraph_ptr.hash)
```
olibearo: Curious why expect this to succeed when we already checked that this block has been uncled, do ethereum node reliably maintain uncles?

datasource/ethereum/src/ethereum_adapter.rs
```diff
@@ -840,144 +859,19 @@ where
...
-
     fn blocks_with_logs(
         &self,
         logger: &Logger,
         subgraph_metrics: Arc<SubgraphEthRpcMetrics>,
         from: u64,
         to: u64,
         log_filter: EthereumLogFilter,
-    ) -> Box<dyn Future<Item = Vec<EthereumBlockPointer>, Error = Error> + Send> {
+    ) -> Box<dyn Future<Item = Vec<Log>, Error = Error> + Send> {
```
olibearo: Seems that the method name needs to change now, something like `logs_in_block_range`

datasource/ethereum/src/ethereum_adapter.rs
```diff
@@ -991,7 +885,7 @@ where
         from: u64,
         to: u64,
         call_filter: EthereumCallFilter,
-    ) -> Box<dyn Future<Item = HashSet<EthereumBlockPointer>, Error = Error> + Send> {
+    ) -> Box<dyn Stream<Item = EthereumCall, Error = Error> + Send> {
```
olibearo: Same here, `(filter_)calls_in_block_range`

datasource/ethereum/src/block_stream.rs
```diff
@@ -1164,45 +816,22 @@ where
 
                 // Yielding blocks from reconciliation process
                 BlockStreamState::YieldingBlocks(mut next_blocks) => {
-                    match next_blocks.poll() {
+                    match next_blocks.pop_front() {
                         // Yield one block
-                        Ok(Async::Ready(Some(next_block))) => {
+                        Some(next_block) => {
```
olibearo: I think it's odd to push blocks without triggers to the subgraph instance, I think it's going to produce metrics observations of `0` for the triggers in block histogram and logs which say something like `0 triggers in blocks`. 

At a higher level, I like that that block stream would advance the subgraph pointer in this case even though at a lower level it adds more paths in the block streams code.

leoyvens: I see that a block with zero triggers is still a sort of special case, but it already happens in the current code for non-final blocks and so far it's a special case that requires less complexity, the cases you mention with logs and metrics are easy to handle. I'm not fundamentally opposed to keeping `set_block_ptr_with_no_changes`, though I'm personally not a fan and @lutter also found it annoying, I'd be interested to know what @Jannis thinks.

datasource/ethereum/src/block_stream.rs
```diff
@@ -401,145 +306,93 @@ where
...
+                            // Ethereum RPC call to find the first few blocks that have event(s) we
+                            // are interested in that lie within the block range between the subgraph ptr
+                            // and either the next data source start_block or the reorg threshold.
+                            // Note that we use block numbers here.
+                            // This is an artifact of Ethereum RPC limitations.
+                            // It is only safe to use block numbers because we are beyond the reorg
+                            // threshold.
+
+                            // Start with first block after subgraph ptr
+                            let from = subgraph_ptr.number + 1;
```
Jannis: I know this code was here before, but the name `from` is pretty misleading (especially when looking at the `let to = if block_filter.trigger_every_block { from }` assignment below. What do you think about updating it to `to` or `next_block_num`?

datasource/ethereum/src/block_stream.rs
Jannis: I like the idea of skipping irrelevant blocks more efficiently but I think `set_block_ptr_with_no_changes` unnecessarily triggered the logic that checks if subscriptions need to be updated? I'm all for dumping it _if_ it doesn't have a detrimental impact on indexing performance.

datasource/ethereum/src/ethereum_adapter.rs
```diff
@@ -1084,4 +971,209 @@ where
...
+                                .map(move |block_hash_opt| {
+                                    let parent_block_pointer = EthereumBlockPointer {
+                                        hash: block_hash_opt.unwrap(),
+                                        number: descendant_block_pointer.number - 1,
+                                    };
+                                    (descendant_block_pointer, parent_block_pointer)
+                                }),
+                        )
+                    })
+                    .collect()
```
Jannis: How large are the block ranges we'd pass in here? While simple and easy to follow, populating the vector one item at a time could be slow — a more performant (but more complicated) approach would be to optimistically resolve all hashes for the `(to, from)` range in parallel (or in parallel batches) and checking them for consistency afterwards.

graph/src/components/ethereum/adapter.rs
```diff
@@ -661,31 +710,129 @@ pub trait EthereumAdapter: Send + Sync + 'static {
...
+
+        if !call_filter.is_empty() {
+            trigger_futs.push(Box::new(
+                eth.blocks_with_calls(&logger, subgraph_metrics.clone(), from, to, call_filter)
+                    .map(EthereumTrigger::Call)
+                    .collect(),
+            ));
+        }
+
+        match block_filter.contract_addresses.len() {
```
Jannis: Should we put this one right next to the `block_filter.trigger_every_block` condition above, perhaps even in an `else` branch to make it more obvious that the block filter is only applied once, depending on which type of block filter it is.

graph/src/components/ethereum/adapter.rs
```diff
@@ -661,31 +710,129 @@ pub trait EthereumAdapter: Send + Sync + 'static {
...
+                    block_hashes.insert(to_hash.unwrap());
+                    triggers_by_block.entry(to).or_insert(Vec::new());
+
+                    (block_hashes, triggers_by_block)
+                })
+                .and_then(move |(block_hashes, mut triggers_by_block)| {
+                    self.load_blocks(logger1, chain_store, block_hashes)
+                        .map(move |block| {
+                            EthereumBlockWithTriggers::new(
+                                // All blocks with triggers should are in `triggers_by_block`, and
```
Jannis: Typo here: "... triggers should are in...".

datasource/ethereum/src/ethereum_adapter.rs
Jannis: Yup

graph/src/components/ethereum/types.rs
```diff
@@ -1,12 +1,96 @@
...
+    fn transaction_for_call(&self, call: &EthereumCall) -> Option<Transaction> {
+        call.transaction_hash
+            .and_then(|hash| self.transactions.iter().find(|tx| tx.hash == hash))
+            .cloned()
+    }
+
+    fn parent_ptr(&self) -> EthereumBlockPointer {
+        EthereumBlockPointer {
+            hash: self.parent_hash,
+            number: self.number() - 1,
```
Jannis: This needs to check for an overflow: the parent of the genesis block with number `0` is `None`. #1297 turns the result of this to be `Option<EthereumBlockPointer>`.

Jannis: I'm ok if we don't do that in this PR but leave it to #1297 to rebase and add this in.

Jannis (changes requested):
I've reviewed a bit more than 2/3 of the PR. I like most changes _a lot_. The things I still need to think about a bit are `ThinEthereumBlock` and `BlockFinality` and how they are used. Will do that tomorrow.

datasource/ethereum/src/block_stream.rs
olibearo: Do we want to give the block stream a property of only emitting relevant blocks (with triggers) to the subgraph instance? It seems cleaner that way to me. If we don't want that or we don't want that in this PR, we should make sure to filter which blocks get logged and added to metrics on the subgraph instance side

Jannis: I just remembered another reason we had this: to get more frequent updates on indexing progress for any subgraph while scanning historic blocks. That is an important feature to preserve.

store/postgres/src/store.rs
leoyvens: I measured the speed of this query for the Moloch subgraph, it never took more than 1 ms, so we're good on that.

datasource/ethereum/src/block_stream.rs
leoyvens: I don't mind the 0 triggers log, during syncing most subgraphs will log that only once every 10k blocks. When synced, it seems relevant to log how many triggers are found on each new block, even if 0. If this is not helpful for the metrics, I can filter out.

I spent some time measuring performance vs master, I used Moloch because it has few events so it's seems like a good one for measuring overhead per processing cycle. Sometimes master was faster, sometimes the branch was faster, so any difference seems to be in the noise. Afaict the block stream is network bound once the blocks are cached, and this PR didn't change the rpc calls done in that scenario.

datasource/ethereum/src/block_stream.rs
leoyvens: I didn't change this code, the diff is from formatting churn. This relies on the block ingestor always pulling in the blocks within the reorg threshold.

datasource/ethereum/src/block_stream.rs
```diff
@@ -250,47 +191,11 @@ where
...
-        to: EthereumBlockPointer,
-    ) -> Result<(), Error> {
-        let result =
-            self.subgraph_store
-                .set_block_ptr_with_no_changes(self.subgraph_id.clone(), from, to);
-
-        match result {
-            Ok(should_migrate) => {
-                if should_migrate {
-                    self.subgraph_store.migrate_subgraph_deployment(
```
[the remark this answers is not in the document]
leoyvens: It's still called by the instance manager after we transact block operations.

graph/src/components/ethereum/types.rs
leoyvens: Yes lets fix this when rebasing, I don't mind which PR gets rebased on which.

datasource/ethereum/src/block_stream.rs
olibearo: Ahh, makes sense re: logs. Let's just filter out the zero trigger blocks from metrics then

store/postgres/src/store.rs
lutter: What I was getting at is that current callers have `block_ptr_from` already, so why not continue to pass that in rather than look it up? Yes, it's a very cheap lookup, at least on an empty, uncontested DB, but I'd rather not query at all when it's not necessary.

As for empty `mods`, I was going by what the current `set_block_ptr_with_no_changes` does; what is in `mods` when we are jumping more than one block? I'd like to assert/check _something_ about them; part of the reason for this is #1093 where we never got to the bottom of why we trip over a metadata update, even though that should be impossible, and I'd like to make sure we are not just silently overwriting stuff.

datasource/ethereum/src/ethereum_adapter.rs
```diff
@@ -1084,4 +971,209 @@ where
...
+                .map(|blocks| {
+                    assert!(blocks.len() <= 1);
+                    blocks
+                        .into_iter()
+                        .next()
+                        .unwrap_or(EthereumBlockWithTriggers::new(vec![], ethereum_block))
+                }),
+            )
+                as Box<dyn Future<Item = _, Error = _> + Send>,
+            BlockFinality::NonFinal(fat_block) => Box::new(future::ok({
```
Jannis: Could we rename `fat_block` to `full_block`, since that's more in line with e.g. `load_full_block` etc.

graph/src/components/ethereum/types.rs
```diff
@@ -1,12 +1,96 @@
...
+        }
+    }
+}
+
+#[derive(Clone, Debug)]
+pub enum BlockFinality {
+    /// If a block is final, we only need the header and the triggers.
+    Final(ThinEthereumBlock),
+
+    // If a block may still be reorged, we need to work with more local data.
```
Jannis: By `local data` you mean that we may still have to load the full block from Ethereum when processing the block, rather than already having everything in the database?

I still can't quite follow how `Final` vs. `NonFinal` are used throughout the code base and what makes them different. I think some more documentation would be good here.

One small request: Could we rename `ThinEthereumBlock` to `LightEthereumBlock`? Even though used in a different context, the light vs. full nomenclature is already established in the Ethereum space and feels more natural here. Even outside this context, "light" is probably more commonly used than "thin" (apart from when talking about wrappers and layers).

Jannis (changes requested):

graph/src/components/ethereum/adapter.rs
leoyvens: done thanks

graph/src/components/ethereum/adapter.rs
leoyvens: Good idea, I've refactored this a bit.

datasource/ethereum/src/ethereum_adapter.rs
leoyvens: I've improved this to fetch the block hashes in parallel, and refactored the code that does regular block fetching so that they're similar. I also removed the special case for the scan range of block triggers, this helps a subgraph like Betoken to be less slow, though the situation is still not ideal because we don't cache the number -> hash association in the database. Consistency checking is not necessary since this range is for final blocks.

datasource/ethereum/src/block_stream.rs
leoyvens: I removed that special case, at some point I added it because block triggers were new and I was afraid of them, but they seem to work fine with the normal range.

store/postgres/src/store.rs
Jannis: The check we had `transact_block_operations` caught bugs in my early block explorer data work, so I second the request for keeping the from pointer and performing the sanity checks that @lutter is recommending.

core/src/subgraph/instance_manager.rs
```diff
@@ -577,9 +608,8 @@ where
     }
 
     // Obtain current and new block pointer (after this block is processed)
-    let block = Arc::new(block);
-    let block_ptr_now = EthereumBlockPointer::to_parent(&block);
-    let block_ptr_after = EthereumBlockPointer::from(&*block);
+    let thin_block = Arc::new(block.thin_block());
```
Jannis: `thin` -> `light`

datasource/ethereum/src/block_stream.rs
Jannis: Hehe, ok 😁

datasource/ethereum/src/block_stream.rs
```diff
@@ -104,7 +82,7 @@ enum ReconciliationStep {
 /// The result of performing a single ReconciliationStep.
 enum ReconciliationStepOutcome {
     /// These blocks must be processed before reconciliation can continue.
-    YieldBlocks(Box<dyn Stream<Item = EthereumBlockWithTriggers, Error = Error> + Send>),
+    YieldBlocks(Vec<EthereumBlockWithTriggers>),
```
Jannis: Is there a good reason why `YieldBlocks` and `ProcessingDescendantBlocks` take a `Vec` but `YieldingBlocks` takes a `VecDeque`?

datasource/ethereum/src/ethereum_adapter.rs
```diff
@@ -461,6 +467,67 @@ where
                     .map_err(|e| e.into_inner().unwrap_or(EthereumContractCallError::Timeout))
             })
     }
+
+    /// Request blocks by hash through JSON-RPC.
+    fn load_blocks_rpc(
```
Jannis: Do we need this as a dedicated method? Can't `load_blocks` be used without caring whether or not blocks come from the Ethereum node or the database?

datasource/ethereum/src/ethereum_adapter.rs
```diff
@@ -461,6 +467,67 @@ where
...
+                .no_limit()
+                .timeout_secs(60)
+                .run(move || {
+                    web3.eth()
+                        .block_with_txs(BlockId::Hash(hash))
+                        .from_err::<Error>()
+                        .map_err(|e| e.compat())
+                        .and_then(move |block| {
+                            block.ok_or_else(|| {
+                                format_err!("eth node did not find block {:?}", hash).compat()
```
Jannis: I prefer spelling out `Ethereum` in errors.

datasource/ethereum/src/ethereum_adapter.rs
```diff
@@ -461,6 +467,67 @@ where
...
+                .no_limit()
+                .timeout_secs(60)
+                .run(move || {
+                    web3.eth()
+                        .block(BlockId::Number(BlockNumber::Number(block_num)))
+                        .from_err::<Error>()
+                        .map_err(|e| e.compat())
+                        .and_then(move |block| {
+                            block.ok_or_else(|| {
+                                format_err!("eth node did not find block {:?}", block_num).compat()
```
Jannis: Same here. Having said this, if we already use `eth node` in all other errors in this module, you can just ignore me and resolve my comments.

datasource/ethereum/src/ethereum_adapter.rs
```diff
@@ -1084,4 +1009,156 @@ where
...
+                    blocks.extend(new_blocks);
+                    blocks.sort_by_key(|block| block.number);
+                    stream::iter_ok(blocks)
+                })
+                .flatten_stream(),
+        )
+    }
+
+    /// Reorg safety: `to` must be a final block.
+    fn block_range_to_ptrs(
```
Jannis: Is this wrapper around `load_block_ptrs_rpc` necessary at all?

graph/src/components/store.rs
```diff
@@ -1000,19 +984,20 @@ pub trait SubgraphDeploymentStore: Send + Sync + 'static {
...
     where
         B: Stream<Item = EthereumBlock, Error = E> + Send + 'a,
-        E: From<Error> + Send + 'a;
+        E: From<Error> + Send + 'a,
+        Self: Sized,
+    {
+        unimplemented!()
+    }
+
+    fn upsert_thin_blocks(&self, blocks: Vec<LightEthereumBlock>) -> Result<(), Error>;
```
Jannis: `upsert_thin_blocks` -> `upsert_light_blocks`

Jannis (changes requested):

datasource/ethereum/src/ethereum_adapter.rs
leoyvens: Done, thanks

graph/src/components/ethereum/types.rs
leoyvens: That is a better name, I've renamed it.

I added a comment above `BlockFinality` to clarify. Actually I don't expect this enum to be around for long, for #902 I imagine we'll want a single trigger scanning function that can validate the block hashes and signal a revert if necessary. But to avoid this change becoming bigger than it already is, I introduced this enum which does the job for our current need of re-scanning only one block.

core/src/subgraph/instance_manager.rs
leoyvens: Ah I missed some, I think I got them all now, thanks.

datasource/ethereum/src/block_stream.rs
leoyvens: I just avoided the conversion until the point where it's actually needed. We need a `VecDeque` at some point because we want to pop from the front, like a queue. Another possibility is to sort the blocks backwards when returning them from `blocks_with_triggers`.

datasource/ethereum/src/ethereum_adapter.rs
leoyvens: You mean we could inline this into `load_blocks`? Sure, but this code is very similar to `load_block_ptrs_rpc`, so I put them as their own methods and next to each other, as a reminder that if we figure out how to optimize one of them, we should do the same optimization to the other.

datasource/ethereum/src/ethereum_adapter.rs
leoyvens: Looks like `eth node` is something I made up, `Ethereum node` is more consistent with the rest of the code base.

datasource/ethereum/src/ethereum_adapter.rs
leoyvens: It would be in the same "layer" as `load_blocks`. If we ever figure out how to cache this data to the DB, this is where we'd handle that.

datasource/ethereum/src/block_stream.rs
leoyvens: @olibearo Done, let me know if any other metrics need to check for this.

datasource/ethereum/src/ethereum_adapter.rs
leoyvens: The name does read weird, renamed it.

leoyvens:
Some comment threads are still on going but I've responded to everything. I did a sample of the size of an `ethereum_blocks` row before and after this change, and there was a 10x size reduction, from around 300k to around 30k. Yay!

Jannis (changes requested):
There's a typo in one of the commit messages:
```
ethereum: Rename methdos
```

leoyvens:
@Jannis rebased and fixed the message.

datasource/ethereum/src/block_stream.rs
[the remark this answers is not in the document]
Jannis: Ah, yes, I see. Thanks :)

datasource/ethereum/src/block_stream.rs
Jannis: It does. I'm not sure Ethereum nodes keep _everything_ about uncles available, like transaction receipts, but the block headers are stored on chain. This is necessary for e.g. applying and verifying uncle rewards.

datasource/ethereum/src/ethereum_adapter.rs
Jannis: Ok, I see it's an internal method anyway. I'll mark this point as resolved.

graph/src/components/ethereum/types.rs
Jannis: Aaaah, now I get it. Thanks for the extra comment!

store/postgres/src/store.rs
Jannis: As am I (fine with the changes, that is).

Jannis (approved):
As far as I am concerned, the changes and post-review updates look good to me. I'd be happy to merge this.

PR #1313, opened 2019-10-28
Handle Geth reverts and empty responses

### LINKED ISSUE #85: Unrecoverable error when calling contract that return 0x
I got the following error when deploy my graph :
```
Subgraph instance failed to run: Failed to process trigger: Failed to handle Ethereum event with handler "handleTransfer": Failed to call function "supportsInterface" of contract "EIP721": ABI error: please ensure the contract and method you're calling exist! failed to decode empty bytes. if you're using jsonrpc this is likely due to jsonrpc returning `0x` in case contract or method don't exist, code: SubgraphSyncingFailure, id: QmPhZ4f4ZJoim2kqaFLgBoHgmWByQYLaE9RxDrVWBMtMrm
```

What I think is happening is that the contract `0x55b9a11c2e8351b4ffc7b11561148bfac9977855` is returning some data, maybe 0x that graph-ts is interpreting as an error but there seems to be no way to react on it. the whole graph shut down.

When I call it through ethers.js is says it revert. So maybe the error is in the internal call and that is why it does not even reach where I check for `result.reverted`

What I want to do is simply check if the contract supports an interface.
If it revert, I consider that it does not. if it does return I check if the value is `true`

Unfortunately doing the following is not sufficient:

```
function supportsInterface(contract: EIP721, interfaceId: String, expected : boolean = true) : boolean {
  let supports = contract.try_supportsInterface(toBytes(interfaceId));
  return !supports.reverted && supports.value == expected;
}
```
So I do that but it is still not enough
```
function supportsInterface(contract: EIP721, interfaceId: String, expected : boolean = true) : boolean {
    let result = contract.tryCall('supportsInterface', [EthereumValue.fromFixedBytes(toBytes(interfaceId))]);
    if (result.reverted) {
        return false;
    }
    let value = result.value;
    if(value.length == 0) {
        return !expected;
    }
    let firstValue = value[0];
    let byteValue = changetype<Bytes>(value[0].data as u32);
    if(byteValue.length == 0) {
        return !expected;
    }
    return firstValue.toBoolean() == expected;
}
```

It would be great if there was a way to allow us to react on wrong data

leoyvens:
We previously held back from accepting a `0x` from Geth as a revert, because it was used for timeouts. Geth has since fixed that and timeouts have a different response, so `0x` is a determinstic response. So using `try_` calls against Geth is now ok.

It can also be that the contract returned an empty response, but that's also an error, even if not techinically a revert, so we treat it the same. Resolves https://github.com/graphprotocol/graph-ts/issues/85.

### DIFF
--- a/datasource/ethereum/src/ethereum_adapter.rs
+++ b/datasource/ethereum/src/ethereum_adapter.rs
@@ -391,6 +391,7 @@ where
                             // 0xfe is the "designated bad instruction" of the EVM, and Solidity
                             // uses it for asserts.
                             const PARITY_BAD_INSTRUCTION_FE: &str = "Bad instruction fe";
+                            const PARITY_BAD_JUMP_PREFIX: &str = "Bad jump";
                             const GANACHE_VM_EXECUTION_ERROR: i64 = -32000;
                             const GANACHE_REVERT_MESSAGE: &str =
                                 "VM Exception while processing transaction: revert";
@@ -425,6 +426,7 @@ where
                                     match rpc_error.data.as_ref().and_then(|d| d.as_str()) {
                                         Some(data)
                                             if data.starts_with(PARITY_REVERT_PREFIX)
+                                                || data.starts_with(PARITY_BAD_JUMP_PREFIX)
                                                 || data == PARITY_BAD_INSTRUCTION_FE =>
                                         {
                                             let reason = if data == PARITY_BAD_INSTRUCTION_FE {
--- a/datasource/ethereum/src/ethereum_adapter.rs
+++ b/datasource/ethereum/src/ethereum_adapter.rs
@@ -1008,11 +1008,11 @@ where
             }
             // Decode the return values according to the ABI
             .and_then(move |output| {
-                // We got a `0x` response. For Geth, this can mean a revert. It can also be that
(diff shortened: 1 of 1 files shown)


datasource/ethereum/src/ethereum_adapter.rs
```diff
@@ -1007,7 +1007,17 @@ where
                 }
             }
             // Decode the return values according to the ABI
-            .and_then(move |output| call.function.decode_output(&output).map_err(From::from)),
+            .and_then(move |output| {
+                // We got a `0x` response. For Geth, this can mean a revert. It can also be that
```
Jannis: What do you think about moving this comment into the `if` branch? It being outside suggests that regardless of the check below, we already got `0x` response.

Jannis (approved):
One minor comment about the comment location, other than that all looks good here.

datasource/ethereum/src/ethereum_adapter.rs
leoyvens: True, moving it.

Jannis (approved):

leoyvens:
This now also resolves #1315, @Jannis could you also take a look at that last commit?

leoyvens:
That last commit should be uncontroversial so merging.

PR #1314, opened 2019-10-28
Add `dataSource.address` and `.network` host exports

leoyvens:
Adds support for basic data source introspection. See https://github.com/graphprotocol/graph-ts/pull/90.

Resolves https://github.com/graphprotocol/support/issues/3. Resolves https://github.com/graphprotocol/support/issues/37.

### DIFF
--- a/runtime/wasm/src/host_exports.rs
+++ b/runtime/wasm/src/host_exports.rs
@@ -1,5 +1,5 @@
 use crate::UnresolvedContractCall;
-use ethabi::Token;
+use ethabi::{Address, Token};
 use futures::sync::oneshot;
 use graph::components::ethereum::*;
 use graph::components::store::EntityKey;
@@ -41,6 +41,8 @@ pub(crate) struct HostExports {
     subgraph_id: SubgraphDeploymentId,
     pub(crate) api_version: Version,
     data_source_name: String,
+    data_source_address: Option<Address>,
+    data_source_network: String,
     templates: Vec<DataSourceTemplate>,
     abis: Vec<MappingABI>,
     ethereum_adapter: Arc<dyn EthereumAdapter>,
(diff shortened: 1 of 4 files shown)


runtime/wasm/src/host_exports.rs
```diff
@@ -41,6 +41,8 @@ pub(crate) struct HostExports {
     subgraph_id: SubgraphDeploymentId,
     pub(crate) api_version: Version,
     data_source_name: String,
+    data_source_address: Option<Address>,
+    data_source_network: String,
```
Jannis: Given that we'll be adding non-Ethereum data sources eventually, shouldn't this be `Option<String>`, like the address?

runtime/wasm/src/host_exports.rs
```diff
@@ -630,6 +636,14 @@ impl HostExports {
             panic!("Critical error logged in mapping");
         }
     }
+
+    pub(crate) fn data_source_address(&self) -> H160 {
+        self.data_source_address.clone().unwrap_or_default()
+    }
+
+    pub(crate) fn data_source_network(&self) -> String {
+        self.data_source_network.clone()
```
Jannis: And shouldn't these two methods return `Option<_>` as well to map to `Address | null` / `string | null` in AssemblyScript?

leoyvens: The reasoning to represent `None` as an empty string is that data sources with an address (and network) are the common case, and if you don't have an address you wouldn't be calling this export anyways, so I tried to save users the hassle of checking null and casting to `String`.

Jannis: While I think that's reasonable, I could see users writing helper functions used in multiple data sources where they _don't_ have a priori knowledge of whether there is an address / network.

Then again, in those cases checking for `''` is no worse than checking for `null`.

Jannis:
@leoyvens I still approve. 😁

leoyvens:
@Jannis ha thanks.

PR #1317, opened 2019-10-29
Use an env var for json-rpc timeouts, increase default

leoyvens:
I noticed with erc20 that some eth_getLogs requests could take over a minute and still complete successfully.

### DIFF
--- a/docs/environment-variables.md
+++ b/docs/environment-variables.md
@@ -34,6 +34,7 @@ those.
   calls to make when scanning logs for a subgraph. Defaults to 100.
 - `GRAPH_ETHEREUM_MAX_EVENT_ONLY_RANGE`: Maximum range size for `eth.getLogs`
   requests that dont filter on contract address, only event signature.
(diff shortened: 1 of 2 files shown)


docs/environment-variables.md
```diff
@@ -34,6 +34,7 @@ those.
   calls to make when scanning logs for a subgraph. Defaults to 100.
 - `GRAPH_ETHEREUM_MAX_EVENT_ONLY_RANGE`: Maximum range size for `eth.getLogs`
   requests that dont filter on contract address, only event signature.
+- `ETHEREUM_JSON_RPC_TIMEOUT`: Timeout for Ethereum JSON-RPC requests.
```
Jannis: I recall you being a proponent of prefixing our environment variables with `GRAPH_`. I've since stuck to the prefix for everything, let's keep using it. 😉

leoyvens: We really should! Done.

PR #1318, opened 2019-10-29
Refactor away `guard` from `MetadataOperation::Update`

leoyvens:
Following up to #1308 with some more simplification.

The guard was only used in `update_ethereum_block_pointer_operations`, which in turn is only used in reverts or transact block.

For transact block, the guard was made totally redundant as we now fetch the block in the same transaction. For reverts, I was more conservative and just added an assert to replace the guard.


leoyvens:
@lutter thanks for the review!

PR #1326, opened 2019-10-31
Optimize `blocks` to not fetch receipts from the DB

leoyvens:
If receipts were present, they'd be loaded in memory only to be discarded. This could cause a memory usage spike. Now we only query the block header from the DB.


Jannis (approved):
Good catch!

PR #1338, opened 2019-11-07
Indexing time instrumentation

### LINKED ISSUE #1302: Improve instrumentation to measure how much time all parts of indexing take

leoyvens:
Resolves #1302.

This adds Prometheus metrics for measuring the total time to sync a subgraph, broken down by sections. This PR marks some sections and a convenient API is introduced for marking sections as needed. This also introduces a metric for measuring the impact of DB contention on performance. Grafana panels were added to display these metrics, which looks like this (timings for syncing erc20 up to about block 5400000):

Captura de Tela 2019-11-07 às 13 10 07

See https://github.com/graphprotocol/instrumentation/pull/1 for the panels.

### DIFF
--- a/core/src/subgraph/instance_manager.rs
+++ b/core/src/subgraph/instance_manager.rs
@@ -364,7 +364,8 @@ impl SubgraphInstanceManager {
         ));
         let instance =
             SubgraphInstance::from_manifest(&logger, manifest, host_builder, host_metrics.clone())?;
-        let stopwatch_metrics = StopwatchMetrics::new(deployment_id.clone(), logger.clone());
+        let stopwatch_metrics =
+            StopwatchMetrics::new(logger.clone(), deployment_id.clone(), registry.clone());
 
         // The subgraph state tracks the state of the subgraph instance over time
         let ctx = IndexingContext {
--- a/core/src/subgraph/instance_manager.rs
+++ b/core/src/subgraph/instance_manager.rs
@@ -32,7 +32,6 @@ struct IndexingState<T: RuntimeHostBuilder> {
     log_filter: EthereumLogFilter,
     call_filter: EthereumCallFilter,
     block_filter: EthereumBlockFilter,
-    stopwatch_metrics: StopwatchMetrics,
     restarts: u64,
 }
 
@@ -344,6 +343,8 @@ impl SubgraphInstanceManager {
 
         // Create a subgraph instance from the manifest; this moves
         // ownership of the manifest and host builder into the new instance
+        let stopwatch_metrics =
+            StopwatchMetrics::new(logger.clone(), deployment_id.clone(), registry.clone());
         let subgraph_metrics = Arc::new(SubgraphInstanceMetrics::new(
             registry.clone(),
             deployment_id.clone().to_string(),
@@ -352,6 +353,7 @@ impl SubgraphInstanceManager {
         let host_metrics = Arc::new(HostMetrics::new(
             registry.clone(),
             deployment_id.clone().to_string(),
+            stopwatch_metrics.clone(),
         ));
         let ethrpc_metrics = Arc::new(SubgraphEthRpcMetrics::new(
             registry.clone(),
@@ -361,11 +363,10 @@ impl SubgraphInstanceManager {
             registry.clone(),
             ethrpc_metrics.clone(),
             deployment_id.clone(),
+            stopwatch_metrics,
         ));
         let instance =
             SubgraphInstance::from_manifest(&logger, manifest, host_builder, host_metrics.clone())?;
-        let stopwatch_metrics =
-            StopwatchMetrics::new(logger.clone(), deployment_id.clone(), registry.clone());
 
         // The subgraph state tracks the state of the subgraph instance over time
         let ctx = IndexingContext {
@@ -386,7 +387,6 @@ impl SubgraphInstanceManager {
                 log_filter,
                 call_filter,
                 block_filter,
-                stopwatch_metrics,
                 restarts: 0,
             },
             subgraph_metrics,
@@ -750,6 +750,7 @@ where
 
         // Transact entity operations into the store and update the
         // subgraph's block stream pointer
+        let _section = ctx.host_metrics.stopwatch.start_section("transact_block");
         let start = Instant::now();
         ctx.inputs
             .store
--- a/core/src/subgraph/instance_manager.rs
+++ b/core/src/subgraph/instance_manager.rs
@@ -32,6 +32,7 @@ struct IndexingState<T: RuntimeHostBuilder> {
     log_filter: EthereumLogFilter,
     call_filter: EthereumCallFilter,
     block_filter: EthereumBlockFilter,
+    stopwatch_metrics: StopwatchMetrics,
     restarts: u64,
 }
 
@@ -363,6 +364,7 @@ impl SubgraphInstanceManager {
         ));
         let instance =
             SubgraphInstance::from_manifest(&logger, manifest, host_builder, host_metrics.clone())?;
+        let stopwatch_metrics = StopwatchMetrics::new(deployment_id.clone(), logger.clone());
 
         // The subgraph state tracks the state of the subgraph instance over time
         let ctx = IndexingContext {
@@ -383,6 +385,7 @@ impl SubgraphInstanceManager {
                 log_filter,
                 call_filter,
                 block_filter,
+                stopwatch_metrics,
                 restarts: 0,
             },
             subgraph_metrics,
--- a/graph/src/components/metrics/stopwatch.rs
+++ b/graph/src/components/metrics/stopwatch.rs
@@ -31,6 +31,13 @@ impl StopwatchMetrics {
         registry: Arc<dyn MetricsRegistry>,
     ) -> Self {
         let mut inner = StopwatchInner {
+            total_counter: *registry
+                .new_counter(
+                    format!("{}_total_secs", subgraph_id),
+                    format!("total time spent syncing"),
+                    HashMap::new(),
+                )
+                .expect("failed to register total_secs prometheus counter"),
             logger,
             subgraph_id,
             registry,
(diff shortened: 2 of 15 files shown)


graph/src/components/metrics/stopwatch.rs
```diff
@@ -0,0 +1,156 @@
...
+    }
+}
+
+#[derive(Clone)]
+pub struct StopwatchMetrics {
+    disabled: Arc<AtomicBool>,
+    inner: Arc<Mutex<StopwatchInner>>,
+}
+
+impl StopwatchMetrics {
```
lutter: Could you add a comment on the typical use of this, i.e., that you do `let _section = start_section(..)` and then have `drop` end the section? It required me to look into the implementation to understand how the uses in the rest of the PR worked.

lutter:
One thing I am wondering: it might be useful to also measure how much time we spend in block reversions. Would that be terribly hard to add?

leoyvens:
@lutter I made this stop registering metrics when the subgraph becomes synced, I think what we care about measuring right now is how fast a subgraph becomes synced. Since reverts usually happen after the subgraph has synced, it doesn't fit with the theme of this PR, though it would be a good standalone metric to add.

Jannis:
@leoyvens I'll review later but one quick comment: any chance we could benchmark the WASM memory allocation? You improved that recently but the time we spend there used to be a significant unknown between `store.get` returning and the mapping execution being able to use the value. Would be nice to know how long this takes now. What do you think?

lutter:
@leoyvens what made me mention reverts is that I have seen some queries related to reverts take a long time (something like 2.5s) and having a better understanding of how much time we spend there would be great. But I agree with your point about time-to-sync being more interesting right now.

graph/src/components/metrics/stopwatch.rs
leoyvens: I tried doing that, let me know if it's helpful.

leoyvens:
@Jannis I've added a section for `store_get_asc_new`, it's about 10% of the total `store.get` time. It might become something we need to optimize but not an issue right now.

lutter (approved):
LGTM

leoyvens:
@Jannis do you want to take a look? Should I merge?

core/src/subgraph/instance_manager.rs
```diff
@@ -746,6 +750,7 @@ where
 
         // Transact entity operations into the store and update the
         // subgraph's block stream pointer
+        let _section = ctx.host_metrics.stopwatch.start_section("transact_block");
```
Jannis: Having the stopwatch stop implicitly when the section is dropped is almost too magical for my taste. I wonder if there is a better name than `start_section` that makes this more obvious. Like `measure_until_dropped` (not a great name) or something.

graph/src/components/metrics/stopwatch.rs
```diff
@@ -0,0 +1,169 @@
...
+    }
+}
+
+/// Usage example:
+/// ```ignore
+/// // Start counting time for the "main_section".
+/// let _main_section = stopwatch.start_section("main_section");
+/// // do stuff...
+/// // Pause timer for "main_section", start for "child_section".
+/// let chid_section = stopwatch.start_section("child_section");
```
Jannis: Typo here: `chid` -> `child`

graph/src/components/metrics/stopwatch.rs
```diff
@@ -0,0 +1,169 @@
...
+impl StopwatchMetrics {
+    pub fn new(
+        logger: Logger,
+        subgraph_id: SubgraphDeploymentId,
+        registry: Arc<dyn MetricsRegistry>,
+    ) -> Self {
+        let mut inner = StopwatchInner {
+            total_counter: *registry
+                .new_counter(
+                    format!("{}_total_secs", subgraph_id),
```
Jannis: I'm not sure `total_secs` is the right term. Maybe `{}_secs_to_synced` or `{}_syncing_secs`? Who knows what we might use `total` for in the future, maybe _all_ time spent on a subgraph?

graph/src/components/metrics/stopwatch.rs
```diff
@@ -0,0 +1,169 @@
...
+
+    pub fn start_section(&self, id: &str) -> Section {
+        let id = id.to_owned();
+        if !self.disabled.load(Ordering::SeqCst) {
+            self.inner.lock().unwrap().start_section(id.clone())
+        }
+        Section {
+            id,
+            stopwatch: self.clone(),
+        }
```
Jannis: I don't understand this part yet: what does it mean to start a section on `inner` if the section is disabled and what does it mean to clone it? How does this make child sections work?

Deserves some explanatory comments I feel 😄

graph/src/components/metrics/stopwatch.rs
```diff
@@ -0,0 +1,169 @@
...
+    subgraph_id: SubgraphDeploymentId,
+    registry: Arc<dyn MetricsRegistry>,
+
+    // Counter for the total time the subgraph spent syncing.
+    total_counter: Counter,
+
+    // Counts the seconds spent in each section of the indexing code.
+    counters: HashMap<String, Counter>,
+
+    // The top section is the one that's currently executing.
```
Jannis: By `top`, do you mean the last section in the vector or the first?

graph/src/components/metrics/stopwatch.rs
Jannis: Maybe this is fine actually, as the stopwatch metrics could be used in other places too. But then I'd maybe pass in a `sync` prefix here that allows this to become `{}_sync_total_secs`.

graph/src/components/metrics/stopwatch.rs
```diff
@@ -0,0 +1,169 @@
...
+
+impl StopwatchInner {
+    fn record_and_reset(&mut self) {
+        if let Some(section) = self.section_stack.last() {
+            // Get or create the counter.
+            let counter = if let Some(counter) = self.counters.get(section) {
+                counter.clone()
+            } else {
+                let name = format!("{}_{}_secs", self.subgraph_id, section);
+                let help = format!("indexing section {}", section);
```
Jannis: Here also I wonder if `indexing` isn't tying this too closely to indexing even though we may want to use stopwatches in other places as well.

Jannis: e.g. query execution

graph/src/components/metrics/stopwatch.rs
```diff
@@ -0,0 +1,169 @@
...
+
+impl Section {
+    /// A more readable `drop`.
+    pub fn end(self) {}
+}
+
+impl Drop for Section {
+    fn drop(&mut self) {
+        self.stopwatch
+            .end_section(std::mem::replace(&mut self.id, String::new()))
```
Jannis: What's the `std::mem::replace` trick here doing?

runtime/wasm/src/module/mod.rs
```diff
@@ -464,7 +464,13 @@ where
         )?;
 
         Ok(Some(match entity_option {
-            Some(entity) => RuntimeValue::from(self.asc_new(&entity)),
+            Some(entity) => {
+                let _section = self
+                    .host_metrics
+                    .stopwatch
+                    .start_section("store_get_asc_new");
+                RuntimeValue::from(self.asc_new(&entity))
```
Jannis: Nice!

store/postgres/src/store.rs
```diff
@@ -641,7 +649,24 @@ impl Store {
...
         let conn = self.get_conn()?;
+        self.get_entity_conn_timers
+            .lock()
+            .unwrap()
+            .entry(subgraph.clone())
+            .or_insert_with(|| {
+                *self
+                    .registry
+                    .new_counter(
+                        format!("{}_get_conn_secs", subgraph),
```
Jannis: Should this be `{}_get_entity_conn_secs` perhaps?

Jannis (changes requested):

core/src/subgraph/instance_manager.rs
leoyvens: The idea is to keep metrics from cluttering the code, so I figured one line of code is better than two. I'm up for suggestions on naming.

graph/src/components/metrics/stopwatch.rs
leoyvens: Interesting, I hadn't considered the reuse potential. Should all metrics be prefixed with `_sync` or just the one for the totals?

graph/src/components/metrics/stopwatch.rs
leoyvens: Simulating a move to prevent cloning `self.id`. We can't actually move because drop doesn't take `self`.

graph/src/components/metrics/stopwatch.rs
leoyvens: true, I've removed the `indexing`.

graph/src/components/metrics/stopwatch.rs
leoyvens: The top of the stack is the last element, I've added a comment.

graph/src/components/metrics/stopwatch.rs
leoyvens: I've added comments, if the stopwatch is disabled the section won't be started and nothing will happen when the returned value is dropped.

store/postgres/src/store.rs
leoyvens: done

leoyvens:
@Jannis thanks for the review, I've responded to all comments.

graph/src/components/metrics/stopwatch.rs
Jannis: Initially, I thought we should prefix all but I think just the totals works. `{subgraph}_total_secs` is just too generic, but `{subgraph}_store_get` (or whatever) makes sense even if it's just capture during syncing for now. What do you think?

core/src/subgraph/instance_manager.rs
Jannis: I don't have any good ideas, so let's keep it.

Jannis:
@leoyvens Ok, just the one comment about `_total_secs` remains.

graph/src/components/metrics/stopwatch.rs
leoyvens: Agreed, I'll change just the `total_secs` to `sync_total_secs`.

leoyvens:
@Jannis renamed that.

Jannis (approved):
LGTM!

Issue #325, opened 2018-09-04
Implement complete GraphQL query validation


Jannis:
We should implement full query validation so that, after a query is checked, later parts of the code have less to worry about and can make the assumption that the query is valid and matches the schema.

See [the GraphQL specification](http://facebook.github.io/graphql/draft/#sec-Validation) for the validation algorithm.

Jannis:
This is particularly important because broken queries can cause panics at the moment:
> If user enters a blank list or one that contains different types of values a panic ends the node process.  Include query validation allowing filter generation to fail and report back the error to the client instead of panicking. 
Also check to see if list is empty, index access currently panics if empty.

timmclean:
Bumping up the milestone on this one as we need to get rid of these panics

PR #1339, opened 2019-11-08
Validate for unknown fields in graphql query

leoyvens:
Cases such as the one in the `follow_interface_reference` test were not being caught, were a field that exists only on the concrete type is selected directly on the interface. Part of #325.

### DIFF
--- a/core/tests/interfaces.rs
+++ b/core/tests/interfaces.rs
@@ -211,24 +211,7 @@ fn follow_interface_reference_invalid() {
 
     let query = "query { legged(id: \"child\") { parent { id } } }";
 
-    let parent = (
-        Entity::from(vec![
-            ("id", Value::from("parent")),
-            ("legs", Value::from(4)),
-            ("parent", Value::Null),
-        ]),
-        "Animal",
-    );
-    let child = (
-        Entity::from(vec![
-            ("id", Value::from("child")),
-            ("legs", Value::from(3)),
-            ("parent", Value::String("parent".into())),
-        ]),
-        "Animal",
-    );
-
-    let res = insert_and_query(subgraph_id, schema, vec![parent, child], query).unwrap();
+    let res = insert_and_query(subgraph_id, schema, vec![], query).unwrap();
 
     match &res.errors.unwrap()[0] {
         QueryError::ExecutionError(QueryExecutionError::UnknownField(_, type_name, field_name)) => {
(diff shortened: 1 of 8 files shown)


core/tests/interfaces.rs
```diff
@@ -230,6 +230,46 @@ fn follow_interface_reference() {
 
     let res = insert_and_query(subgraph_id, schema, vec![parent, child], query).unwrap();
 
+    match &dbg!(res.errors).unwrap()[0] {
```
lutter: Leftover `dbg!`

graphql/src/execution/execution.rs
```diff
@@ -220,6 +205,97 @@ where
                 .and_then(|complexity| total_complexity.checked_add(complexity).ok_or(Overflow))
             })
     }
+
+    // Checks for invalid selections.
+    pub(crate) fn validate_fields(
```
lutter: It would be good to have a test or two for this so it doesn't break inadvertently if this code ever gets changed.

graph/src/data/query/error.rs
```diff
@@ -32,7 +32,7 @@ pub enum QueryExecutionError {
     OrderByNotSupportedError(String, String),
     OrderByNotSupportedForType(String),
     FilterNotSupportedError(String, String),
-    UnknownField(Pos, String, String),
+    UnknownField(String, String),
```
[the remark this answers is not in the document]
leoyvens: I removed it because it wasn't used in the error mesage, but I'll put it back.

graphql/src/execution/execution.rs
leoyvens: I've added another test.

leoyvens:
Thanks for the reviews, comments addressed!

graph/src/data/query/error.rs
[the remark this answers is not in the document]
Jannis: It's true it's not used in `Display` but it could be used in `Deserialize` further below. Can we add this variant there? See https://github.com/graphprotocol/graph-node/blob/90ce7e355bf6d42ca92a9be5e9381e6e8e1f917d/graph/src/data/query/error.rs#L326

graph/src/data/query/error.rs
```diff
@@ -120,8 +121,8 @@ impl fmt::Display for QueryExecutionError {
             FilterNotSupportedError(value, filter) => {
                 write!(f, "Filter not supported by value `{}`: `{}`", value, filter)
             }
-            UnknownField(_, t, s) => {
-                write!(f, "Type `{}` has no field `{}`", t, s)
+            UnknownField(p, t, s) => {
+                write!(f, "Type `{}` has no field `{}`, at position {}:{}", t, s, p.line, p.column)
```
Jannis: Rather than including the position in the error string, can we add the `UnknownField` variant here?
https://github.com/graphprotocol/graph-node/pull/1339/files#diff-b50193ba4cc86fb70dca7e10af594488R328-R336

That way clients will receive the error as an object, with `line`/`column` fields, as is ideal in GraphQL.

leoyvens: Oh I did not notice we have that, everything makes sense now! I've added it there, thanks.

Jannis: Yep, not exactly obvious 😁

Jannis: I still think we should remove the `, at position {}:{}` here because now the position will show up twice in query results w/ this error. 😉

leoyvens: Ah of course 🤦‍♂

PR #1352, opened 2019-11-14
Remove use of `.wait()`

leoyvens:
This was causing the node to hang at startup if the `network_identifiers` calls had to retry, for example due to rate limits. Most of the diff is from having to indent the rest of main.

### DIFF
--- a/node/src/main.rs
+++ b/node/src/main.rs
@@ -486,247 +486,256 @@ fn async_main() -> impl Future<Item = (), Error = ()> + Send + 'static {
 
     let postgres_conn_pool =
         create_connection_pool(postgres_url.clone(), store_conn_pool_size, &logger);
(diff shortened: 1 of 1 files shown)


node/src/main.rs
```diff
@@ -486,247 +486,256 @@ fn async_main() -> impl Future<Item = (), Error = ()> + Send + 'static {
...
+    .map_err(move |e| {
+        error!(stores_error_logger, "Was a valid Ethereum node provided?");
+        panic!("Failed to connect to Ethereum node: {}", e);
+    })
+    .map(move |(network_name, network_identifier)| {
+        info!(
+        stores_logger,
+        "Connected to Ethereum";
+        "network" => &network_name,
+        "network_version" => &network_identifier.net_version,
```
Jannis: The indentation here is off.

PR #1361, opened 2019-11-18
Fix race condition in stopwatch

leoyvens:
We've seen the `transact_block` section start before `run_handler` ended because the section was being dropped only after the result was sent.


lutter (approved):
LGTM

Issue #1359, opened 2019-11-18
Graph node docker process gets killed abruptly


puneet-khushwani-eth:
**What is the current behavior?**

Our process starts docker containers for graph-node (with a fresh data directory) and its dependencies. To this we deploy a sub-graph and then fire some calls to it to verify if deployment was successful. In most cases the docker process for graph-node gets killed abruptly (some times even before we start deploying sub-graph)

Versions we use -
docker compose : 1.23.2
geth : 1.9.5
the-graph : 0.16.0

logs we see:

```
wait-for-it.sh: ipfs:5001 is available after 0 seconds
wait-for-it.sh: waiting 30 seconds for postgres:5432
wait-for-it.sh: postgres:5432 is available after 6 seconds
++ echo dev:http://172.16.0.201:40005
+ graph-node --postgres-url postgresql://graph-node:let-me-in@postgres:5432/graph-node --ethereum-rpc dev:http://172.16.0.201:40005 --ipfs ipfs:5001
Nov 18 10:20:30.151 INFO Graph Node version: v0.16.0 (32a58d94a 2019-11-01)
Nov 18 10:20:30.151 INFO Starting up
Nov 18 10:20:30.154 INFO Trying IPFS node at: http://ipfs:5001/
Nov 18 10:20:30.179 INFO Creating transport, url: http://172.16.0.201:40005, network: dev
Nov 18 10:20:30.182 INFO Successfully connected to IPFS node at: http://ipfs:5001/
Nov 18 10:20:30.191 INFO Connecting to Postgres, conn_pool_size: 10, url: postgresql://graph-node:HIDDEN_PASSWORD@postgres:5432/graph-node
Nov 18 10:20:30.339 INFO Connected to Postgres, url: postgresql://graph-node:HIDDEN_PASSWORD@postgres:5432/graph-node, component: Store
Nov 18 10:20:30.339 INFO Connecting to Ethereum..., network: dev
Nov 18 10:20:30.345 INFO Connected to Ethereum, network_version: 5, network: dev
Nov 18 10:20:30.352 INFO Waiting for other graph-node instances to finish migrating, component: Store
Nov 18 10:20:30.354 INFO Running migrations, component: Store
Nov 18 10:20:30.367 INFO Migrations finished, component: Store
Nov 18 10:20:30.367 INFO Completed pending Postgres schema migrations, component: Store
[...]

PR #1366, opened 2019-11-19
Detect invalid ETHEREUM_REORG_THRESHOLD

leoyvens:
A better error message for the issue in #1359.

### DIFF
--- a/chain/ethereum/src/block_stream.rs
+++ b/chain/ethereum/src/block_stream.rs
@@ -275,6 +275,14 @@ where
             }
         }
 
+        if head_ptr.number < reorg_threshold {
+            panic!(
+                "The reorg threshold {} is larger than the size of the chain {}, \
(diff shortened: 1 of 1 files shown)


datasource/ethereum/src/block_stream.rs
```diff
@@ -275,6 +275,14 @@ where
             }
         }
 
+        if head_ptr.number < reorg_threshold {
+            panic!(
+                "The reorg threshold {} is larger than the size of the chain {}, \
+                 you probably want to set the ETHEREUM_REORG_THRESHOLD env var to 0",
```
Jannis: `env var` -> `environment variable` please 😄

Jannis (approved):

datasource/ethereum/src/block_stream.rs
leoyvens: Fixed

PR #1367, opened 2019-11-19
Add Aggregate metric utility

Jannis:
This adds `Aggregate` metrics for a single value.

Given a `name`, this registers the following gauges:

- `{name}_count` (counting all updates to the value)
- `{name}_sum` (the sum of all updates to the value over time)
- `{name}_avg` (the moving average of all updates to the value over time)
- `{name}_cur` (the current or latest update to the value

I've been using this for the network subgraphs feature, but I thought it could be useful to other parts, so I wanted to submit it earlier.

### DIFF
--- a/graph/src/components/metrics/aggregate.rs
+++ b/graph/src/components/metrics/aggregate.rs
@@ -1,0 +1,96 @@
+use std::collections::HashMap;
+use std::time::Duration;
+
+use crate::prelude::*;
+
+/// Aggregate metrics for a single value.
+///
+/// Given a `name`, this registers the following gauges:
+///
+/// - `{name}_count` (counting all updates to the value)
+/// - `{name}_sum` (the sum of all updates to the value over time)
+/// - `{name}_avg` (the moving average of all updates to the value over time)
+/// - `{name}_cur` (the current or latest update to the value)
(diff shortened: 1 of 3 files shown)


graph/src/components/metrics/aggregate.rs
```diff
@@ -0,0 +1,96 @@
...
+                HashMap::new(),
+            )
+            .expect(format!("failed to register metric `{}_avg`", name).as_str());
+        let cur_gauge = registry
+            .new_gauge(
+                format!("{}_cur", name),
+                format!("{} (cur)", help),
+                HashMap::new(),
+            )
+            .expect(format!("failed to register metric `{}_cur`", name).as_str());
```
leoyvens: We do this in other places as well, but it makes me nervous that the graph node would crash because of a metrics bug.

leoyvens:
I'd expect some of this stuff to be doable with PromQL, at least in theory, but I had a hard time trying to work with gauges, for example I couldn't figure out how to sum the values in PromQL so I like this. At least the count I'd expect PromQL to be able to do somehow.

Jannis:
Obsoleted by #1383.

PR #1370, opened 2019-11-21
Dynamic block range based on triggers per range

### LINKED ISSUE #1350: Dynamic, per-subgraph block step
Currently the block range size is a fixed env var for all subgraphs. This is starting to cause operational issues, because subgraphs that find too many triggers in a range will use too much RAM, make unreasonably expensive ethereum calls and make expensive DB queries. However setting the block range too low will needlessly punish performance for light subgraphs.

I propose a relatively simple solution which is to have the block stream dynamically set the block step for each subgraph, based on how many triggers were found in the previous range. We target an ideal range for the number of triggers, I propose 100 to 1000 triggers. If the number of triggers is below that we increase the step 10x, if it is above we decrease 10x. Possible step values can be 1, 10, 100, 1k, 10k up to 100k blocks.

leoyvens:
Use the amount of triggers found in the previous range to adjust the next range, trying to meet a configured target triggers per range. For dense subgraphs this allows the complexity of ethereum calls and memory usage to be kept under control, without compromising speed for cheap subgraphs or initial speed for subgraphs that are denser towards the end of the chain.

Testing with erc20, this was effective in keeping the trigger count around the default target of 1000. The memory usage sat at ~100mb, which seems ok for an actively syncing subgraph, and suggests that each trigger takes ~100kb in average. The range size was at about 3000 blocks, which is expected to decrease at higher block numbers.

Resolves #1350. The implementation is different from what I suggested in the issue in that the range steps are not tied to multiples of 10.

### DIFF
--- a/chain/ethereum/src/block_stream.rs
+++ b/chain/ethereum/src/block_stream.rs
@@ -406,9 +406,9 @@ where
                             // - Scan 10 blocks:
                             //   2 triggers found, 0.2 per block, range_size = 1000 / 0.2 = 5000
                             // - Scan 5000 blocks:
-                            //   500 triggers found, 0.1 per block, range_size = 500 / 0.2 = 2500
-                            // - Scan 2500 blocks:
-                            //   500 triggers found, 0.2 per block, range_size = 500 / 0.2 = 2500
+                            //   10000 triggers found, 2 per block, range_size = 1000 / 2 = 500
+                            // - Scan 500 blocks:
+                            //   1000 triggers found, 2 per block, range_size = 1000 / 2 = 500
                             let max_range_size =
                                 MAX_BLOCK_RANGE_SIZE.min(ctx.previous_block_range_size * 10);
                             let range_size = if ctx.previous_triggers_per_block == 0.0 {
--- a/chain/ethereum/src/block_stream.rs
+++ b/chain/ethereum/src/block_stream.rs
@@ -822,7 +822,9 @@ where
                             self.ctx.previous_triggers_per_block =
                                 total_triggers as f64 / block_range_size as f64;
                             self.ctx.previous_block_range_size = block_range_size;
-                            debug!(self.ctx.logger, "Processing {} triggers", total_triggers);
+                            if total_triggers > 0 {
+                                debug!(self.ctx.logger, "Processing {} triggers", total_triggers);
+                            }
 
                             // Switch to yielding state until next_blocks is depleted
                             state = BlockStreamState::YieldingBlocks(next_blocks);
--- a/chain/ethereum/src/block_stream.rs
+++ b/chain/ethereum/src/block_stream.rs
@@ -398,6 +398,17 @@ where
                             // Calculate the range size according to the target number of triggers,
                             // respecting the global maximum and also not increasing too
                             // drastically from the previous block range size.
+                            //
+                            // An example of the block range dynamics:
+                            // - Start with a block range of 1, target of 1000.
+                            // - Scan 1 block:
+                            //   0 triggers found, max_range_size = 10, range_size = 10
+                            // - Scan 10 blocks:
+                            //   2 triggers found, 0.2 per block, range_size = 1000 / 0.2 = 5000
+                            // - Scan 5000 blocks:
+                            //   500 triggers found, 0.1 per block, range_size = 500 / 0.2 = 2500
+                            // - Scan 2500 blocks:
+                            //   500 triggers found, 0.2 per block, range_size = 500 / 0.2 = 2500
                             let max_range_size =
                                 MAX_BLOCK_RANGE_SIZE.min(ctx.previous_block_range_size * 10);
                             let range_size = if ctx.previous_triggers_per_block == 0.0 {
--- a/chain/ethereum/src/block_stream.rs
+++ b/chain/ethereum/src/block_stream.rs
@@ -12,19 +12,18 @@ use graph::prelude::{
 };
 use tokio::timer::Delay;
 
-const FAST_SCAN_SPEEDUP: u64 = 10;
-
 lazy_static! {
-    /// Number of blocks to request in each chunk.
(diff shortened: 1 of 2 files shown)


chain/ethereum/src/block_stream.rs
```diff
@@ -12,19 +12,18 @@ use graph::prelude::{
...
 use tokio::timer::Delay;
 
-const FAST_SCAN_SPEEDUP: u64 = 10;
-
 lazy_static! {
-    /// Number of blocks to request in each chunk.
-    static ref ETHEREUM_BLOCK_RANGE_SIZE: u64 = ::std::env::var("ETHEREUM_BLOCK_RANGE_SIZE")
-        .unwrap_or("10000".into())
+    /// Maximum number of blocks to request in each chunk.
+    static ref MAX_BLOCK_RANGE_SIZE: u64 = std::env::var("GRAPH_ETHEREUM_MAX_BLOCK_RANGE_SIZE")
```
olibearo: Should we keep this default in production? Right now the existing env var in community is set to `3000` and `10000` for customers

leoyvens: The `MAX_BLOCK_RANGE_SIZE` is essentially to prevent us from scanning the entire chain at once. Scanning 100k blocks at once may cause us to go way over the target in bad cases, but that will happen only once and then the range will adapt.

From my testing I believe these defaults should work out fine and we won't need to tinker with these variables, but if they turn out to be too aggresive we can always lower as a short term solution.

Jannis: Won't reducing 100k to 10k have a negative performance impact on early block scanning of most subgraphs?

chain/ethereum/src/block_stream.rs
```diff
@@ -173,6 +180,10 @@ where
                 start_blocks,
                 templates_use_calls,
                 metrics,
+
+                // A high number here forces a slow start, with a range of 1.
```
Jannis: Why don't we want to start fast? 😉

chain/ethereum/src/block_stream.rs
```diff
@@ -383,20 +395,28 @@ where
...
+                            let max_range_size =
+                                MAX_BLOCK_RANGE_SIZE.min(ctx.previous_block_range_size * 10);
+                            let range_size = if ctx.previous_triggers_per_block == 0.0 {
+                                max_range_size
+                            } else {
+                                (*TARGET_TRIGGERS_PER_BLOCK_RANGE as f64
+                                    / ctx.previous_triggers_per_block)
+                                    .max(1.0)
+                                    .min(max_range_size as f64)
+                                    as u64
```
Jannis: My brain is really bad at doing math unless it's written down. It would help if we had 2-3 examples in a comment here. Like what range do we get if there were a lot of triggers  previously, what if there were very few triggers and what if there was the ideal size?

```rust
// Examples (target triggers: 1000, max range size 10000, previous block range size 500):
// New max range size              -> min(10000, 5000 * 10) = 10000
// 1 triggers/block previously     -> min(max(1000 / 1, 1.0), 10000) = min(1000, 10000) = 1000
// 1000 triggers/block previously  -> min(max(1000 / 1000, 1.0), 10000) = min(1.0, 10000) = 1
// 10000 triggers/block previously -> min(max(1000 / 10000, 1.0), 10000) = min(1.0, 10000) = 1

Is there a mistake in my math or is the formula not ideal? In the second example, we're at the target triggers/block rate, so shouldn't the new block range size be exactly the same as before?

Comment (leoyvens) [reply chain/ethereum/src/block_stream.rs]: I've added some examples in a comment.

If there are 1000 triggers per block in 1000 blocks that's 1 million triggers total, which is way above the target that's why the range decreases to 1 in your example.

File: chain/ethereum/src/block_stream.rs [hunk 4]
```diff
@@ -398,6 +398,17 @@ where
...
                             // drastically from the previous block range size.
+                            //
+                            // An example of the block range dynamics:
+                            // - Start with a block range of 1, target of 1000.
+                            // - Scan 1 block:
+                            //   0 triggers found, max_range_size = 10, range_size = 10
+                            // - Scan 10 blocks:
+                            //   2 triggers found, 0.2 per block, range_size = 1000 / 0.2 = 5000
+                            // - Scan 5000 blocks:
+                            //   500 triggers found, 0.1 per block, range_size = 500 / 0.2 = 2500
```
Jannis: Hmm. If I scan 5000 blocks and I find 500 triggers but my target is 1000, shouldn't I increase the block range to 10000, not keep it at 500?

Similar with scanning 2500 blocks and finding 500 triggers. In that case I'd also think that I have to double my range size to get to the target amount of triggers in the range.

chain/ethereum/src/block_stream.rs
Jannis: Oh, yeah, that part I got completely wrong. 1000 triggers as the target means 1000 triggers per _block_ range_, not per block. 🤦‍♂

chain/ethereum/src/block_stream.rs
leoyvens: Starting slow is more conservative. For example if we're coming from a large range size and suddenly hit a ton of triggers, causing the node to crash oom, starting slow gives us a chance to recover and find the correct range.

The range can increase by 10x on each scan, so it takes only 5 scans for this to hit the default maximum of 100k.

chain/ethereum/src/block_stream.rs
leoyvens: Right this is totally wrong, I think I fixed it.

chain/ethereum/src/block_stream.rs
Jannis: Right, like when a node restarts and the subgraphs pick up at a range with an enormous amount of triggers. Ok, yes, makes sense. I was only thinking about the start case / genesis block, but with the start block feature, you could start at any block and too large a range could cause trouble.

leoyvens: Ah yes there's that as well, when you start you don't know what your previous range was.

PR #1371, opened 2019-11-21
Log the transaction hash for the trigger when it fails to run

olibearo:
Addresses https://github.com/graphprotocol/support/issues/38

### DIFF
--- a/core/src/subgraph/instance_manager.rs
+++ b/core/src/subgraph/instance_manager.rs
@@ -796,6 +796,11 @@ where
                 EthereumTrigger::Call(_) => TriggerType::Call,
                 EthereumTrigger::Block(..) => TriggerType::Block,
             };
+            let transaction_id = match &trigger {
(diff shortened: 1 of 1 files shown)


core/src/subgraph/instance_manager.rs
```diff
@@ -805,7 +810,14 @@ where
                     subgraph_metrics.observe_trigger_processing_duration(elapsed, trigger_type);
                     (ctx, block_state)
                 })
-                .map_err(|e| format_err!("Failed to process trigger: {}", e))
+                .map_err(|e| match transaction_hash {
+                    Some(tx_hash) => format_err!(
+                        "Failed to process trigger for transaction {}: {}",
```
Jannis: Is `for` the right word or is it `in`?

olibearo: Good point, fixing

PR #1374, opened 2019-11-22
Fix db contention metric, introduce global counters

leoyvens:
It wouldn't work in a multi-network node because the "generic store" would register the metric first and then the store which really owned the subgraph could not register. This removes the counters from the store and lets the metrics registry centralize keeping track of them.

### DIFF
--- a/core/src/metrics/registry.rs
+++ b/core/src/metrics/registry.rs
@@ -1,5 +1,5 @@
 use std::collections::HashMap;
-use std::sync::Arc;
+use std::sync::{Arc, RwLock};
 
 use graph::prelude::{MetricsRegistry as MetricsRegistryTrait, *};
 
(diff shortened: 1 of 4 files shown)


core/src/metrics/registry.rs
```diff
@@ -197,6 +202,21 @@ impl MetricsRegistryTrait for MetricsRegistry {
...
+            let help = "global counter".to_owned();
+            let counter = *self.new_counter(name.clone(), help, HashMap::new())?;
+            self.global_counters
+                .write()
+                .unwrap()
+                .insert(name.clone(), counter.clone());
+            Ok(counter)
+        }
+    }
+
```
Jannis: I wonder if this could be made more atomic with e.g. `.entry(&name).or_insert_with()` but with the error handling and the `RwLock` around the hash map I don't see an obvious way.

leoyvens: Yhea that's the first thing I tried but doesn't work for the reasons you mention.

PR #1375, opened 2019-11-22
Consider decode failures as reverts

### LINKED ISSUE #33: `try_` calls don't handle ABI decoding errors / mismatches
Contract calls made with `try` handle assertion failures in contracts but they don't currently handle ABI decoding errors / ABI mismatches.

For example, different contracts could implement a `name` function with different return types. A call like
```ts
let result = contract.try_name()
if (!result.reverted) {
  let s: string = result.value
}
```
currently works with contracts that have a `name` function that returns a string, but will fail with contracts that have a `name` function that returns e.g. a `bytes32`.

`try` exists not just to detect reverts (although that was its original purpose), but could also allow to test whether a method can be called on a contract at all. Cases that would be good to cover:

1. The function doesn't exist on the contract.
2. The function ABI doesn't match (i.e. the result fails to decode).

One idea could be to add an `resultDecodingFailed` field similar to `reverted` so you could do
```ts
let result = contract.try_name()
if (!reverted && !result.resultDecodingFailed) {
  let s: string = result.value
}
```

leoyvens:
Resolves https://github.com/graphprotocol/support/issues/33.

### DIFF
--- a/chain/ethereum/src/ethereum_adapter.rs
+++ b/chain/ethereum/src/ethereum_adapter.rs
@@ -1030,7 +1030,11 @@ where
                     // support/#85 for a use case.
                     Err(EthereumContractCallError::Revert("empty response".into()))
                 } else {
-                    call.function.decode_output(&output).map_err(From::from)
+                    // Decode failures are reverts. The reasoning is that if Solidity fails to
+                    // decode an argument, that's a revert, so the same goes for the output.
+                    call.function.decode_output(&output).map_err(|e| {
+                        EthereumContractCallError::Revert(format!("failed to decode output: {}", e))
+                    })
                 }
             }),
         )


chain/ethereum/src/ethereum_adapter.rs
```diff
@@ -1030,7 +1030,11 @@ where
                     // support/#85 for a use case.
                     Err(EthereumContractCallError::Revert("empty response".into()))
                 } else {
-                    call.function.decode_output(&output).map_err(From::from)
+                    // Decode failures are reverts. The reasoning is that if Solidity fails to
+                    // decode an argument, that's a revert, so the same goes for the output.
```
Jannis: I'd say it's a little different with the outputs. Us being able to not decode the output can just mean that there is an ABI mismatch and the contract returns different values/types from the call. I could be wrong?

Jannis: I still think this is the right thing to do. Just not sure the explanation is 100% accurate and won't confuse us later.

Jannis (approved):

chain/ethereum/src/ethereum_adapter.rs
leoyvens: > Us being able to not decode the output can just mean that there is an ABI mismatch and the contract returns different values/types from the call

How is that different from the input decoding? Both are caused either by an ABI mismatch or by bad data.

leoyvens: Oh I see, we encode the inputs and they should always be valid, so that's actually a real error, yhea what I wrote doesn't make sense, I'll change this.

leoyvens: We concluded this analogy is pretty much correct as it is.

PR #1382, opened 2019-11-25
Fix extending EthereumCallFilter

olibearo:
Addresses https://github.com/graphprotocol/support/issues/42

### DIFF
--- a/graph/src/components/ethereum/adapter.rs
+++ b/graph/src/components/ethereum/adapter.rs
@@ -858,3 +858,58 @@ pub trait EthereumAdapter: Send + Sync + 'static {
         ethereum_block: BlockFinality,
     ) -> Box<dyn Future<Item = EthereumBlockWithTriggers, Error = Error> + Send>;
 }
+
+#[cfg(test)]
+mod tests {
+    use super::EthereumCallFilter;
+
+    use web3::types::Address;
+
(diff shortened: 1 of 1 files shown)


graph/src/components/ethereum/adapter.rs
```diff
@@ -326,12 +326,21 @@ impl EthereumCallFilter {
...
+            match self.contract_addresses_function_signatures.get_mut(&key) {
+                Some((existing_start_block, existing_sigs)) => {
+                    *existing_start_block =
+                        cmp::min(proposed_start_block, existing_start_block.clone());
+                    existing_sigs.extend(new_sigs);
+                }
+                None => {
+                    self.contract_addresses_function_signatures
+                        .insert(key, (proposed_start_block, new_sigs));
+                }
```
Jannis: We should call `key` `address` in this method. I got confused because I didn't know what `key` meant.

Jannis: The code looks good otherwise. It would be nice to have a test for this though.

olibearo: Addressed ;)

olibearo:
Okay, let's add a test

olibearo:
Test added @Jannis

Issue #1381, opened 2019-11-25
Optimize store.get and store.set


Jannis:
In https://github.com/graphprotocol/graph-node/issues/1347#issuecomment-555539043 we have identified `store.get`/`store.set` to take up the majority of indexing time (~30-50%). Optimizing these should significantly boost indexing performance.

Next steps:

1. Plan specific optimizations (like for non-existent entities).
2. Implement them.

leoyvens:
Discussing with @lutter, we decided to start with some straightforward improvements to the entity cache, which are not going to the DB for validation if we can help it, and loading entities in a single DB query in `as_modifications`. After these we can evaluate doing a bloom filter for non-existing entities or better caching for existing entities.

At least for now we don't see a need for a user-facing `store.overwrite` or `store.insert`, we'd like to optimize as much as possible without complicating things for the user.

leoyvens:
Update: The optimizations for `store.set` are done, the next target is `store.get`, I'll evaluate the impact of the bloom filter and plan it in detail if it has the large impact.

leoyvens:
Optimizations have now been implemented for store.set and store.get, so that they are no longer among the most time consuming sections. There is more that can be done, such as the bloom filter, but I feel we've done enough in the context of this issue.

PR #1384, opened 2019-11-26
Optimize store.set entity validation

leoyvens:
If the entity that comes from the mapping is already valid, avoid going to the DB at all. Also add a metrics section for `as_modifications`, since some DB accesses shift from `store_set` to `as_modifications`. Part of #1381.

PR for grafana https://github.com/graphprotocol/instrumentation/pull/3.

### DIFF
--- a/runtime/wasm/src/host_exports.rs
+++ b/runtime/wasm/src/host_exports.rs
@@ -142,17 +142,20 @@ impl HostExports {
             entity_type,
             entity_id,
         };
+        let entity = Entity::from(data);
+        state.entity_cache.set(key.clone(), entity.clone());
 
-        state.entity_cache.set(key.clone(), Entity::from(data));
-
-        if self.store.uses_relational_schema(&self.subgraph_id)? {
-            // Validate the changes against the subgraph schema
+        // Validate the changes against the subgraph schema.
+        // If the set of fields we have is already valid, avoid hitting the DB.
+        let schema = self.store.input_schema(&self.subgraph_id)?;
+        if validate_entity(&schema.document, &key, &entity).is_err()
+            && self.store.uses_relational_schema(&self.subgraph_id)?
+        {
             let entity = state
                 .entity_cache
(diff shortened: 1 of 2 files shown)


lutter:
This looks good, but you should also change the lookup that happens in [EntityCache](https://github.com/graphprotocol/graph-node/blob/master/graph/src/components/store.rs#L1210-L1212) from looking up missing entities one by one to using a bulk query (`EntityQuery` with a `EntityFilter::In`) and then put the entities you get back or `None` into `self.current`.

runtime/wasm/src/host_exports.rs
```diff
@@ -142,17 +142,20 @@ impl HostExports {
...
-
-        if self.store.uses_relational_schema(&self.subgraph_id)? {
-            // Validate the changes against the subgraph schema
+        let entity = Entity::from(data);
+        state.entity_cache.set(key.clone(), entity.clone());
+
+        // Validate the changes against the subgraph schema.
+        // If the set of fields we have is already valid, avoid hitting the DB.
+        let schema = self.store.input_schema(&self.subgraph_id)?;
+        if validate_entity(&schema.document, &key, &entity).is_err()
```
lutter: You could save a clone of the entity if you did the first `validate_entity` before the `entity_cache.set`.

Jannis:
@leoyvens Were you able to measure the impact of this change compared to before?

runtime/wasm/src/host_exports.rs
leoyvens: Ah yes I should do that.

leoyvens:
@Jannis I only observed that `store.set` now takes less time than `as_modifications`, as expected. I want to do the bulk loading as well and then I'll measure.

Jannis:
@leoyvens Works for me!

runtime/wasm/src/host_exports.rs
leoyvens: Done, merging.

PR #1385, opened 2019-11-27
adding missing port redirections in docker-compose

Amxx:
Following discussion on discord, Needed to acces monitoring interfaces


Jannis:
Great, thank you!

PR #1396, opened 2019-12-06
Limit JSON-RPC retries

leoyvens:
So that we don't get stuck, for example requesting a block hash that got reorged. The block stream already resets on error so this will not fail the subgraph.

### DIFF
--- a/chain/ethereum/src/ethereum_adapter.rs
+++ b/chain/ethereum/src/ethereum_adapter.rs
@@ -58,10 +58,10 @@ lazy_static! {
     /// This is used for requests that will not fail the subgraph if the limit is reached, but will
     /// simply restart the syncing step, so it can be low. This limit guards against scenarios such
     /// as requesting a block hash that has been reorged.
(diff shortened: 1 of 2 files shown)


lutter (changes requested):
Could you also document this environment variable in `docs/environment-variables.md`? Other than that, looks good to me.

Jannis:
@leoyvens The changes look good. Could you briefly explain how this is not failing the subgraph but resets the block stream and tries again _outside_ the adapter. I forgot how that worked.

Jannis:
Heh, wait. Isn't this more or less the same as https://github.com/graphprotocol/graph-node/pull/1277/files?

leoyvens:
@Jannis it is 🤦‍♂. The difference is that this only limits retries for calls used in the block stream. For example I'm not sure if limiting retries in the block ingestor would be prudent.

I like the env var name in that PR better so I renamed it in this PR, and documented it.

Jannis:
@leoyvens I think my PR caused more drastic failures in situations where they weren't desirable. So let's go with yours.

Jannis (approved):

lutter (approved):

PR #1397, opened 2019-12-06
Enable time-travel GraphQL queries

### LINKED ISSUE #813: Time travel queries
We currently only run queries as of the latest block. We will want a way to run queries as of a specified block. There are also use cases where people want to query information about how an entity has changed over time. That could be considered as part of this epic or broken out into a separate one.

### LINKED ISSUE #1329: Time-travel queries
With relational schemas (#1159) we now have the backend plumbing for time-travel queries in place. What's missing to close out #813 is a way for users to access that functionality, and pass a specific block (number) at which a query should happen to the store.

There's a few ways in which this could be surfaced. My favorite options are either through a directive or a filter. We'd really want to restrict this to the top level of a query since it doesn't really make sense to mix different block heights in the same query. Either way, users should be able to either tell us a block number or a block hash at which a query should be run.

This would look like one of these variants:
```
users(...) @block(number: 5) { .. } 
users(...) @block(hash: "0xdeadbeef") { .. }
users(block_number: 5, ..) { .. }
users(block_hash: "0xdeadbeef", ..) { .. }
```

If they give us a hash, we first map it to the block number; strictly speaking, we should check that that hash is an ancestor of the current head block of the subgraph, but that can be an expensive query, since it needs to traverse the `hash -> parent_hash` pointers in blocks in the database. We could limit that search by treating blocks a certain distance behind head as final, as we do with the ingestor though.

lutter:
With this PR, it is possible to run queries against the state of a subgraph at an arbitrary block height, and not just the latest block. So far, a query like `transactions { id }` would always return the data for the latest set of transactions, and rerunning it might return a different result if the subgraph had ingested more transactions in the mean time.

It is now possible to fix the block at which to query by passing an additional `block` argument to top-level fields: `transactions(block: { number: 7000 }) { id }` will return the transactions as of the block with that number, and `transactions(block: { hash: "0xdeadbeef" }) { id }` will return transactions as of that block hash. To make absolutely sure the query result is reproducible, the form with the `hash` should be used as which block is designated by `number` may change if the chain is reorganized (though using `number` far enough away from the chain head will be fine)

If a query does not explicitly specify a block, we continue to run it against the latest available data, so that the behavior for existing queries does not change.

Fixes #1329 
Fixes #813

### DIFF
--- a/graph/src/components/store.rs
+++ b/graph/src/components/store.rs
@@ -260,10 +260,14 @@ pub struct EntityQuery {
 }
 
 impl EntityQuery {
-    pub fn new(subgraph_id: SubgraphDeploymentId, collection: EntityCollection) -> Self {
+    pub fn new(
+        subgraph_id: SubgraphDeploymentId,
+        block: BlockNumber,
+        collection: EntityCollection,
+    ) -> Self {
         EntityQuery {
             subgraph_id,
-            block: BLOCK_NUMBER_MAX,
+            block,
             collection,
             filter: None,
             order_by: None,
--- a/graph/src/components/store.rs
+++ b/graph/src/components/store.rs
@@ -1159,6 +1159,14 @@ pub trait Store: Send + Sync + 'static {
         subgraph_id: &SubgraphDeploymentId,
         block_ptr: &EthereumBlockPointer,
     );
+
+    /// Return the number of the block with the given hash for the given
+    /// subgraph
+    fn block_number(
+        &self,
+        subgraph_id: &SubgraphDeploymentId,
+        block_hash: H256,
+    ) -> Result<BlockNumber, StoreError>;
 }
 
 pub trait SubgraphDeploymentStore: Send + Sync + 'static {
--- a/graph/src/components/store.rs
+++ b/graph/src/components/store.rs
@@ -217,6 +217,12 @@ pub enum EntityCollection {
     /// column `b`; they will be grouped by using `A.a` and `B.b` as the keys
     Window(Vec<EntityWindow>),
 }
+/// The type we use for block numbers. This has to be a signed integer type
+/// since Postgres does not support unsigned integer types. But 2G ought to
+/// be enough for everybody
+pub type BlockNumber = i32;
+
+pub const BLOCK_NUMBER_MAX: BlockNumber = std::i32::MAX;
 
 /// A query for entities in a store.
 ///
--- a/graph/src/components/store.rs
+++ b/graph/src/components/store.rs
@@ -233,6 +233,13 @@ pub struct EntityQuery {
     /// ID of the subgraph.
     pub subgraph_id: SubgraphDeploymentId,
 
+    /// The block height at which to execute the query. Set this to
+    /// `BLOCK_NUMBER_MAX` to run the query at the latest available block.
+    /// If the subgraph uses JSONB storage, anything but `BLOCK_NUMBER_MAX`
+    /// will cause an error as JSONB storage does not support querying anything
+    /// but the latest block
+    pub block: BlockNumber,
+
     /// The names of the entity types being queried. The result is the union
     /// (with repetition) of the query for each entity.
     pub collection: EntityCollection,
@@ -256,6 +263,7 @@ impl EntityQuery {
     pub fn new(subgraph_id: SubgraphDeploymentId, collection: EntityCollection) -> Self {
         EntityQuery {
             subgraph_id,
+            block: BLOCK_NUMBER_MAX,
             collection,
             filter: None,
             order_by: None,
@@ -771,10 +779,11 @@ pub trait Store: Send + Sync + 'static {
         subgraph_id: SubgraphDeploymentId,
     ) -> Result<Option<EthereumBlockPointer>, Error>;
 
-    /// Looks up an entity using the given store key.
+    /// Looks up an entity using the given store key at the latest block.
     fn get(&self, key: EntityKey) -> Result<Option<Entity>, QueryExecutionError>;
 
-    /// Look up multiple entities. Returns a map of entities by type.
+    /// Look up multiple entities as of the latest block. Returns a map of
+    /// entities by type.
     fn get_many(
         &self,
         subgraph_id: &SubgraphDeploymentId,
--- a/graph/src/components/store.rs
+++ b/graph/src/components/store.rs
@@ -1166,7 +1166,7 @@ pub trait Store: Send + Sync + 'static {
         &self,
         subgraph_id: &SubgraphDeploymentId,
         block_hash: H256,
-    ) -> Result<BlockNumber, StoreError>;
+    ) -> Result<Option<BlockNumber>, StoreError>;
 }
 
 pub trait SubgraphDeploymentStore: Send + Sync + 'static {
--- a/graphql/src/query/ext.rs
+++ b/graphql/src/query/ext.rs
@@ -77,21 +77,22 @@ impl FieldExt for q::Field {
             );
         if let Some(value) = value {
             if let q::Value::Object(map) = value {
-                if map.len() != 1 || !(map.contains_key("hash") || map.contains_key("number")) {
+                let hash = map.get("hash");
+                let number = map.get("number");
+                if map.len() != 1 || (hash.is_none() && number.is_none()) {
                     return Err(invalid_argument("block", self, value));
                 }
                 let subgraph = parse_subgraph_id(object_type)?;
-                if let Some(hash) = map.get("hash") {
-                    TryFromValue::try_from_value(hash)
+                match (hash, number) {
+                    (Some(hash), _) => TryFromValue::try_from_value(hash)
                         .map_err(|_| invalid_argument("block.hash", self, value))
                         .map(|hash| {
                             Some(BlockConstraint {
                                 subgraph,
                                 block: BlockLocator::Hash(hash),
                             })
-                        })
-                } else if let Some(number_value) = map.get("number") {
-                    TryFromValue::try_from_value(number_value)
+                        }),
+                    (_, Some(number_value)) => TryFromValue::try_from_value(number_value)
                         .map_err(|_| invalid_argument("block.number", self, number_value))
                         .and_then(|number: u64| {
                             TryFrom::try_from(number)
@@ -102,9 +103,8 @@ impl FieldExt for q::Field {
                                 subgraph,
                                 block: BlockLocator::Number(number),
                             })
-                        })
-                } else {
-                    unreachable!("We already checked that there is a hash or number entry")
+                        }),
+                    _ => unreachable!("We already checked that there is a hash or number entry"),
                 }
             } else {
                 Err(invalid_argument("block", self, value))
--- a/graphql/src/query/ext.rs
+++ b/graphql/src/query/ext.rs
@@ -3,6 +3,16 @@
 use graphql_parser::query as q;
 
 use std::collections::BTreeMap;
+use std::convert::TryFrom;
+
+use graph::data::graphql::TryFromValue;
+use graph::data::query::QueryExecutionError;
+use graph::data::subgraph::SubgraphDeploymentId;
+use graph::prelude::web3::types::H256;
+use graph::prelude::BlockNumber;
+
+use crate::execution::ObjectOrInterface;
+use crate::store::parse_subgraph_id;
 
 pub trait ValueExt {
     fn as_object(&self) -> &BTreeMap<q::Name, q::Value>;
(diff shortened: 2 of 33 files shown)


lutter:
Note that this PR applies on top of #1386 as it requires some conveniences introduced in that PR

lutter:
Rebased to latest master and made master the base branch for this PR

graphql/src/query/ext.rs
```diff
@@ -24,3 +34,83 @@ impl ValueExt for q::Value {
...
+                        .and_then(|number: u64| {
+                            TryFrom::try_from(number)
+                                .map_err(|_| invalid_argument("block.number", self, number_value))
+                        })
+                        .map(|number| {
+                            Some(BlockConstraint {
+                                subgraph,
+                                block: BlockLocator::Number(number),
+                            })
+                        })
```
Jannis: Slightly more compact and therefor a little easier to read:

Do
```rust
let hash = map.get("hash");
let number = map.get("number");
```
first, simplify the first `if` to
```
if (map.len() != 1 || !(hash.is_some() || number.is_some()))
```
and use `match` below:
```
match (hash, number) => {
  (Some(hash), _) => { ... },
  (_, Some(number)) => { ... },
  _ => unreachable!("We already checked that there is a hash or number"),
}
```

graphql/src/schema/api.rs
```diff
@@ -417,30 +497,53 @@ fn add_subscription_type(
...
+fn block_argument() -> InputValue {
+    InputValue {
+        position: Pos::default(),
+        description: None, /*Some(
+                               "The block at which the query should be executed. \
+                                Can either be an `Int` containing the block number or a `Bytes` \
+                                value containing a block hash. Defaults to the latest block when \
+                                omitted."
+                                   .to_owned(),
+                           ), */
```
Jannis: Is there any reason for this to be commented out?

graphql/src/store/resolver.rs
```diff
@@ -246,6 +247,24 @@ where
...
+    fn locate_block(&self, bc: &BlockConstraint) -> Result<BlockNumber, QueryExecutionError> {
+        match bc.block {
+            BlockLocator::Number(number) => Ok(number),
+            BlockLocator::Hash(hash) => self
+                .store
+                .block_number(&bc.subgraph, hash)
+                .map_err(|e| e.into())
+                .and_then(|number| {
+                    number.ok_or_else(|| {
+                        QueryExecutionError::ValueParseError(
```
Jannis: Is that a `ValueParseError` then or an invalid argument error?

graphql/src/subscription/mod.rs
```diff
@@ -222,6 +223,7 @@ where
         variable_values,
         deadline: timeout.map(|t| Instant::now() + t),
         max_first,
+        block: BLOCK_NUMBER_MAX,
```
Jannis: Subscribing to specific blocks is funny... at least for historic blocks that data is never going to change. There may be situations where it makes sense though (future block heights maybe)?

graphql/tests/query.rs
```diff
@@ -1240,3 +1255,44 @@ fn can_use_nested_filter() {
...
+        format!("hash : \"0x{}\"", block.hash)
+    }
+
+    musicians_at("number: 7000", vec!["m1", "m2", "m3", "m4"], "n7000");
+    musicians_at("number: 0", vec!["m1", "m2"], "n0");
+    musicians_at("number: 1", vec!["m1", "m2", "m3", "m4"], "n1");
+
+    musicians_at(&hash(&*GENESIS_BLOCK), vec!["m1", "m2"], "h0");
+    musicians_at(&hash(&*BLOCK_ONE), vec!["m1", "m2", "m3", "m4"], "h1");
+    musicians_at(&hash(&*BLOCK_TWO), vec!["m1", "m2", "m3", "m4"], "h2");
```
Jannis: Lovely 😍

store/postgres/src/entities.rs
```diff
@@ -454,7 +455,19 @@ impl Connection {
...
     ) -> Result<Vec<Entity>, QueryExecutionError> {
         match &*self.storage {
-            Storage::Json(json) => json.query(&self.conn, collection, filter, order, range),
+            Storage::Json(json) => {
+                // JSON storage can only query at the latest block
+                if block != BLOCK_NUMBER_MAX {
+                    return Err(StoreError::QueryExecutionError(
+                        "This subgraph uses JSONB storage, which does not \
+                         support querying at a specific block height. Redeploy \
+                         the subgraph to remove this error message."
```
Jannis: `remove this error` -> `to enable this feature` maybe?

Jannis: Also, simply redeploying won't do it if the deployment ID doesn't change. So perhaps it should say `Deploy a new version of this subgraph to enable this feature`?

store/postgres/src/store.rs
```diff
@@ -6,20 +6,31 @@ use diesel::{insert_into, select, update};
...
+use graph::prelude::{
+    bail, debug, ethabi, format_err, info, o, serde_json, stream, tiny_keccak, tokio, trace, warn,
+    web3, AttributeIndexDefinition, BigInt, BlockNumber, ChainHeadUpdateListener as _,
+    ChainHeadUpdateStream, ChainStore, Entity, EntityKey, EntityModification, EntityOrder,
+    EntityQuery, EntityRange, Error, EthereumBlock, EthereumBlockPointer, EthereumCallCache,
+    EthereumNetworkIdentifier, EventProducer as _, Future, LightEthereumBlock, Logger,
+    MetadataOperation, MetricsRegistry, QueryExecutionError, Schema, Sink as _, StopwatchMetrics,
+    StoreError, StoreEvent, StoreEventStream, StoreEventStreamBox, Stream,
+    SubgraphAssignmentProviderError, SubgraphDeploymentId, SubgraphDeploymentStore,
+    SubgraphEntityPair, TransactionAbortError, Value, BLOCK_NUMBER_MAX,
```
Jannis: Why not import `*`? 😁

store/postgres/src/store.rs
```diff
@@ -1051,6 +1062,35 @@ impl StoreTrait for Store {
...
+        use crate::db_schema::ethereum_blocks::dsl;
+
+        let hash = format!("{:x}", hash);
+
+        // We should also really check that the block with the given hash is
+        // on the chain starting at the subgraph's current head. That check is
+        // very expensive though with the data structures we have currently
+        // available. Ideally, we'd have the last REORG_THRESHOLD blocks in
+        // memory so that we can check against them, and then mark in the
+        // database the blocks on the main chain that we consider final
```
Jannis: #1383 will make this easier and cheaper. We'll be able to query the blocks for a network as entities and filter by `isOmmer: false` to only match blocks that are _truly_ on the official chain.

Jannis: For now I suggest we leave this comment in. How we access blocks for a network will change and that will take us back to this comment eventually.

graph/src/components/store.rs
```diff
@@ -217,6 +217,12 @@ pub enum EntityCollection {
     /// column `b`; they will be grouped by using `A.a` and `B.b` as the keys
     Window(Vec<EntityWindow>),
 }
+/// The type we use for block numbers. This has to be a signed integer type
+/// since Postgres does not support unsigned integer types. But 2G ought to
+/// be enough for everybody
+pub type BlockNumber = i32;
```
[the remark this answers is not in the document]
lutter: I did :smiley:  At 4 blocks/minute, block 2,147,483,647 will be mined in the year 3036. Even for a chain 100x as fast as Ethereum, it would take 10 years to exhaust the `i32` space.

This is also something that was introduced with relational schemas as block ranges are a pair of `int4` in the database, changing that would require a migration of data. Seeing how much runway we have there, there's no need to tackle that now.

graph/src/components/store.rs
```diff
@@ -1140,6 +1159,14 @@ pub trait Store: Send + Sync + 'static {
...
         block_ptr: &EthereumBlockPointer,
     );
+
+    /// Return the number of the block with the given hash for the given
+    /// subgraph
+    fn block_number(
+        &self,
+        subgraph_id: &SubgraphDeploymentId,
+        block_hash: H256,
+    ) -> Result<Option<BlockNumber>, StoreError>;
```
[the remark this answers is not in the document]
lutter: Yes, that is possible for blocks that are not final yet. It's ultimately a flaw in Ethereum though, as that has no concept of finality - anything we do here would close the window of the target moving only with respect to our view of the world; and we'd still be subject to races caused by our view of the world changing. Querying Ethereum close to the chain head is risky business no matter what we do.

graphql/src/query/ext.rs
lutter: Fixed with commit ce5d10eb

graphql/src/schema/api.rs
lutter: Nope, just an oversight from testing. Putting it back in

graphql/src/store/resolver.rs
lutter: I was wondering the same thing when I wrote the code; `ValueParseError` gets turned into `Failed to decode 'block.hash' value: 'no block with that hash found'` which IMHO fits the bill. And it has the added charm that it doesn't require a `Position` which I don't have at that point.

graphql/src/subscription/mod.rs
lutter: Not sure if it ever makes sense to subscribe at a specific block height, and `BLOCK_NUMBER_MAX` means 'give me whatever the latest is' which is what we want for subscriptions.

store/postgres/src/entities.rs
lutter: Yes on both counts. Changed.

store/postgres/src/store.rs
lutter: With `*` imports, you're never quite sure where something comes from; this makes it explicit (IIRC this was prompted by me tearing my hair out over some name clash)

store/postgres/src/store.rs
lutter: Yes, that was my intent: that soon we'll make this comment go away, but doing so is outside the scope of time-travel queries, since it requires _a lot_ of heavy lifting in other places.

lutter:
Addressed all review comments

graph/src/components/store.rs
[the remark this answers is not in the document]
Jannis: That's not going to work. We have to guarantee that if the caller specifies `block: { hash: "abcdef" }`, that all fields are resolved against that particular block (or whatever state they had up to that block). The results cannot come from a different block.

graphql/src/store/resolver.rs
Jannis: Ok, let's roll with that :)

graph/src/components/store.rs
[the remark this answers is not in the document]
lutter: The heart of the matter here seems to be that we do not have a way to translate block hashes to block number in a way that 

* guarantees that the block hash is actually on the chain for that subgraph 
* the part of the subgraph's chain needed to construct the result doesn't change while we construct the result

In the first case, we'd return a result even though we should have returned an error indicating that the block hash is not on the chain.

The second case only affects queries for blocks that are not final yet.

Getting this right will require both some careful design, and some serious changes to how we store information about blocks; at the same time, even with these limitations I feel this feature is very useful as-is. I suggest that we merge this PR, clearly document the limitations of querying by block hash for users, and address these limitations in a separate issue.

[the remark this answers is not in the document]
leoyvens: I agree, it's not like this is introducing new consistency issues that we don't already have.

graph/src/components/store.rs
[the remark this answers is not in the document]
Jannis: Fair enough 😁

lutter:
Rebased to latest master

ernaneluis:
is it possible to disable this feature? cuz i check the database and looks like it keeps all the records states but in my case I just need the latest state.

Jannis:
@ernaneluis Entity versions are also kept for handling reorgs. It's true that old entity versions could be pruned if they are not needed. However, it's the first time we hear someone asking for it. What's your concern about past entity versions being kept in the db?

ernaneluis:
in our use-case, we need to keep balance of a token of a address which is linked to a profile data, we them dump the graphql layer to query direct the database, we implemented our mapping  in a way to update the entity as its described in the docs  https://thegraph.com/docs/assemblyscript-api then when we tried to query the balance we realised they were duplicated because of this new feature of time-queries. So we wonder if there is a way to keep a trim table with latest balances(where it updates the values rather then create a new row)

lutter:
@ernaneluis As @Jannis said, there is no way to turn this off, and it's really important to keep the data correct when a revert happens. I would strongly recommend using GraphQL queries rather than going to the database directly; we make no guarantees that the database schema will stay the way it is, and it's not consodered part of the `graph-node` API.

Despite all this, if you want to get the latest version of an entity, you can simply add `block_range @> 2147483647` (that's the largest positive `i32`, any block number that is in the future will do here) so that a query, for example, for an account by id would become
```sql
select *
  from sgd42.account
 where id = '0xdeadbeef'
   and block_range @> 2147483647
```

Just out of curiosity, why do you want to query the database directly, rather than going through GraphQL?

ernaneluis:
The graphql dont fit our needs, something like pagination is not good enough, also we need to do some joins with other tables to return what we want so we decided to dump the graphql and query directly the sql database.

Jannis:
@ernaneluis That's outside the scope of what we're building.

ernaneluis:
> @ernaneluis That's outside the scope of what we're building.

what exactly "thats" means? just to avoid confusion on my side

lutter:
@ernaneluis I think "that's" refers to querying the database directly - the details of how `graph-node` stores data is not part of our public/stable API, and we may break that in unpredictable ways, and with little fanfare.

I'd still be curious to see more of the details of how you are querying the database - is there any code with the queries that I could look at?

ernaneluis:
we are using typeorm to model the graphnode tables and using the ORM to query and do the relations we need.

PR #1403, opened 2019-12-11
Create metrics to track block ingestion and GraphQL query execution

fordN:
This PR creates two new metrics structs for instrumenting the Block Ingestor and the GraphQL server. 

1. `BlockIngestorMetrics` with a gauge vec to track the number of blocks being synced from Ethereum, labeled by networks name. 

2. `GraphQLServiceMetrics` which includes a histogram vec for tracking query execution time of successful queries, labeled by subgraph deployment id.

### DIFF
--- a/chain/ethereum/src/block_ingestor.rs
+++ b/chain/ethereum/src/block_ingestor.rs
@@ -47,18 +47,15 @@ impl<S> BlockIngestor<S>
 where
     S: ChainStore,
 {
-    pub fn new<M>(
+    pub fn new(
         chain_store: Arc<S>,
         eth_adapter: Arc<dyn EthereumAdapter>,
         ancestor_count: u64,
         network_name: String,
         logger_factory: &LoggerFactory,
         polling_interval: Duration,
-        registry: Arc<M>,
-    ) -> Result<BlockIngestor<S>, Error>
-    where
-        M: MetricsRegistry,
-    {
+        registry: Arc<impl MetricsRegistry>,
+    ) -> Result<BlockIngestor<S>, Error> {
         let logger = logger_factory.component_logger(
             "BlockIngestor",
             Some(ComponentLoggerConfig {
--- a/chain/ethereum/src/block_ingestor.rs
+++ b/chain/ethereum/src/block_ingestor.rs
@@ -39,7 +39,6 @@ where
     ancestor_count: u64,
     network_name: String,
     logger: Logger,
-    ingestor_metrics: Arc<BlockIngestorMetrics>,
     polling_interval: Duration,
 }
 
@@ -54,7 +53,6 @@ where
         network_name: String,
         logger_factory: &LoggerFactory,
         polling_interval: Duration,
-        registry: Arc<impl MetricsRegistry>,
     ) -> Result<BlockIngestor<S>, Error> {
         let logger = logger_factory.component_logger(
             "BlockIngestor",
@@ -64,7 +62,6 @@ where
                 }),
             }),
         );
-        let ingestor_metrics = Arc::new(BlockIngestorMetrics::new(registry.clone()));
 
         Ok(BlockIngestor {
             chain_store,
@@ -72,7 +69,6 @@ where
             ancestor_count,
             network_name,
             logger,
-            ingestor_metrics,
             polling_interval,
         })
     }
@@ -144,10 +140,6 @@ where
                                     LogCode::BlockIngestionStatus
                                 };
                                 if distance > 0 {
-                                    self.ingestor_metrics.observe_blocks_synced(
-                                        &network_name,
-                                        latest_number,
-                                    );
                                     info!(
                                         self.logger,
                                         "Syncing {} blocks from Ethereum.",
--- a/chain/ethereum/src/block_ingestor.rs
+++ b/chain/ethereum/src/block_ingestor.rs
@@ -1,8 +1,34 @@
+use std::collections::HashMap;
 use std::time::Duration;
 use std::time::Instant;
 
 use graph::prelude::*;
 use web3::types::*;
+
+pub struct BlockIngestorMetrics {
+    pub block_number: Box<GaugeVec>,
+}
+
+impl BlockIngestorMetrics {
+    pub fn new<M: MetricsRegistry>(registry: Arc<M>) -> Self {
+        Self {
+            block_number: registry
+                .new_gauge_vec(
+                    String::from("ethereum_block_number"),
+                    String::from("Records the number of the most recent block synced on the Ethereum network"),
+                    HashMap::new(),
+                    vec![String::from("network")],
+                )
+                .unwrap(),
+        }
+    }
+
+    pub fn observe_blocks_synced(&self, network_name: &str, latest_block_number: i64) {
+        self.block_number
+            .with_label_values(vec![network_name].as_slice())
+            .set(latest_block_number as f64);
+    }
+}
 
 pub struct BlockIngestor<S>
 where
@@ -13,6 +39,7 @@ where
     ancestor_count: u64,
     network_name: String,
     logger: Logger,
+    ingestor_metrics: Arc<BlockIngestorMetrics>,
     polling_interval: Duration,
 }
 
@@ -20,14 +47,18 @@ impl<S> BlockIngestor<S>
 where
     S: ChainStore,
 {
-    pub fn new(
+    pub fn new<M>(
         chain_store: Arc<S>,
         eth_adapter: Arc<dyn EthereumAdapter>,
         ancestor_count: u64,
         network_name: String,
         logger_factory: &LoggerFactory,
         polling_interval: Duration,
-    ) -> Result<BlockIngestor<S>, Error> {
+        registry: Arc<M>,
+    ) -> Result<BlockIngestor<S>, Error>
+    where
+        M: MetricsRegistry,
+    {
         let logger = logger_factory.component_logger(
             "BlockIngestor",
             Some(ComponentLoggerConfig {
@@ -36,6 +67,7 @@ where
                 }),
             }),
         );
+        let ingestor_metrics = Arc::new(BlockIngestorMetrics::new(registry.clone()));
 
         Ok(BlockIngestor {
             chain_store,
@@ -43,6 +75,7 @@ where
             ancestor_count,
             network_name,
             logger,
+            ingestor_metrics,
             polling_interval,
         })
     }
@@ -87,7 +120,6 @@ where
     fn do_poll<'a>(&'a self) -> impl Future<Item = (), Error = EthereumAdapterError> + 'a {
         trace!(self.logger, "BlockIngestor::do_poll");
         let network_name = self.network_name.clone();
-
         // Get chain head ptr from store
         future::result(self.chain_store.chain_head_ptr())
             .from_err()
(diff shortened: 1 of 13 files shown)


chain/ethereum/src/block_ingestor.rs
```diff
@@ -1,9 +1,35 @@
...
+pub struct BlockIngestorMetrics {
+    pub block_number: Box<GaugeVec>,
+}
+
+impl BlockIngestorMetrics {
+    pub fn new<M: MetricsRegistry>(registry: Arc<M>) -> Self {
+        Self {
+            block_number: registry
+                .new_gauge_vec(
+                    String::from("ethereum_block_number"),
```
Jannis: We've been calling this the chain head in graph-node. Let's stick to that and call this `ethereum_chain_head_number`.

chain/ethereum/src/block_ingestor.rs
```diff
@@ -1,9 +1,35 @@
...
 use web3::types::*;
 
+pub struct BlockIngestorMetrics {
+    pub block_number: Box<GaugeVec>,
+}
+
+impl BlockIngestorMetrics {
+    pub fn new<M: MetricsRegistry>(registry: Arc<M>) -> Self {
+        Self {
+            block_number: registry
```
Jannis: And let's rename the variable to `chain_head_number` as well.

chain/ethereum/src/block_ingestor.rs
```diff
@@ -1,9 +1,35 @@
...
+                .new_gauge_vec(
+                    String::from("ethereum_block_number"),
+                    String::from("Records the number of the most recent block synced on the Ethereum network"),
```
Jannis: Shorten this to `Block number of the most recent block synced from Ethereum`?

chain/ethereum/src/block_ingestor.rs
```diff
@@ -1,9 +1,35 @@
...
+                )
+                .unwrap(),
+        }
+    }
+
+    pub fn observe_blocks_synced(&self, network_name: &str, latest_block_number: i64) {
+        self.block_number
+            .with_label_values(vec![network_name].as_slice())
+            .set(latest_block_number as f64);
+    }
```
Jannis: Let's also call this `set_chain_head_number`?

chain/ethereum/src/block_ingestor.rs
```diff
@@ -115,6 +147,10 @@ where
                                     LogCode::BlockIngestionStatus
                                 };
                                 if distance > 0 {
+                                    self.ingestor_metrics.observe_blocks_synced(
```
Jannis: This is not the right place IMHO. At this point we have't written the block to our db yet, so as far as our system is concerned, we are not up to date with the chain yet.

Jannis: I think the right place would be an `inspect` here: https://github.com/graphprotocol/graph-node/pull/1403/files#diff-dc8c393601039a99522fbd9bdd37b8dbR233-R235

chain/ethereum/src/block_ingestor.rs
```diff
@@ -1,9 +1,35 @@
...
 
 use graph::prelude::*;
 use web3::types::*;
 
+pub struct BlockIngestorMetrics {
+    pub block_number: Box<GaugeVec>,
+}
+
+impl BlockIngestorMetrics {
+    pub fn new<M: MetricsRegistry>(registry: Arc<M>) -> Self {
```
Jannis: We've been able to get away with `Arc<dyn MetricsRegistry>` in a lot of places. This should work here as well; then we don't have to use generics.

chain/ethereum/src/block_ingestor.rs
```diff
@@ -13,21 +39,26 @@ where
...
+    ingestor_metrics: Arc<BlockIngestorMetrics>,
     polling_interval: Duration,
 }
 
 impl<S> BlockIngestor<S>
 where
     S: ChainStore,
 {
-    pub fn new(
+    pub fn new<M>(
```
Jannis: Same here regarding `Arc<dyn MetricsRegistry>`.

server/http/src/server.rs
```diff
@@ -65,15 +67,17 @@ impl<Q, S> GraphQLServer<Q, S> {
...
+            metrics_registry,
             graphql_runner,
             store,
             node_id,
         }
     }
 }
 
-impl<Q, S> GraphQLServerTrait for GraphQLServer<Q, S>
+impl<M, Q, S> GraphQLServerTrait for GraphQLServer<M, Q, S>
```
Jannis: Same here regarding `Arc<dyn MetricsRegistry>`.

server/http/src/service.rs
```diff
@@ -11,6 +13,40 @@ use hyper::{Body, Method, Request, Response, StatusCode};
...
+        write!(f, "GraphQLServiceMetrics {{ }}")
+    }
+}
+
+impl GraphQLServiceMetrics {
+    pub fn new<M: MetricsRegistry>(registry: Arc<M>) -> Self {
+        let query_execution_time = registry
+            .new_histogram_vec(
+                format!("subgraph_query_execution_time"),
+                String::from("Measures the execution time for GraphQL queries"),
```
Jannis: Drop the `Measures the`. I think that's obvious. :)

server/http/src/service.rs
```diff
@@ -225,27 +266,35 @@ where
...
+                                    "GraphQL query served";
+                                    "subgraph_deployment" => sd_id.deref(),
+                                    "query_time_ms" => elapsed.as_millis(),
+                                    "code" => LogCode::GraphQlQuerySuccess,
+                                )
+                            }
+                            Err(ref e) => error!(
+                                logger,
+                                "GraphQL query failed";
+                                "subgraph_deployment" => sd_id.deref(),
```
Jannis: Is this not also worth measuring the time for?

server/http/src/service.rs
```diff
@@ -11,6 +13,40 @@ use hyper::{Body, Method, Request, Response, StatusCode};
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "GraphQLServiceMetrics {{ }}")
+    }
+}
+
+impl GraphQLServiceMetrics {
+    pub fn new<M: MetricsRegistry>(registry: Arc<M>) -> Self {
+        let query_execution_time = registry
+            .new_histogram_vec(
+                format!("subgraph_query_execution_time"),
```
Jannis: I wonder if we should shorten this to `query_time`, like in the logs.

Jannis (changes requested):

chain/ethereum/src/block_ingestor.rs
That3Percent: Why do we avoid using generics?

Jannis: I've been trying to avoid them where I could to simplify type/method signatures. Admittedly, I don't understand the drawbacks of either approach.

leoyvens: The conclusion is that because this is function parameter the best syntax is `fn new(registry: Arc<impl MetricsRegistry>)`

server/http/src/service.rs
fordN: Want to go with the shorter, `query_time`?  
I vacillated between the two options myself and ended up using `query_execution_time`  because I think it's more clear. In the logs we use `query_time_ms`;  I think aside from being useful to express the unit of measurement,  the `_ms` suffix makes the distinction between execution time and time of execution clear.

server/http/src/service.rs
fordN: Definitely. I had questions about how we want to handle this, so I started with the simplest case.  The error case could return early from a bad request and I don't want those times to skew the aggregate numbers.  

How do you feel about keeping `subgraph_query_execution_time` (or `subgraph_query_time`) for successful queries and add another histogram called `subgraph_failed_query_execution_time`?

chain/ethereum/src/block_ingestor.rs
Jannis: dyn -> impl

server/http/src/service.rs
Jannis: `subgraph_query_time` + `subgraph_failed_query_time`? Sounds good to me 😄

server/http/src/service.rs
Jannis: Ok, let's keep the longer version. We'll want to add more detailed metrics for query execution at some point. We'll see about the names again then.

chain/ethereum/src/block_ingestor.rs
fordN: Hmmm.. `new(registry: Arc<impl MetricsRegistry>)` does not compile for `BlockIngestorMetrics` and `GraphQLServiceMetrics`. 

Screen Shot 2019-12-12 at 10 46 59

Btw, the format I started with, using generics, matches all of the existing metrics struct implementations:
  - `SubgraphInstanceManagerMetrics`
  - `SubgraphInstanceMetrics`
  - `HostMetrics`
  - `ProviderEthRpcMetrics`
  - `SubgraphEthRpcMetrics`
  - `BlockStreamMetrics`

I've updated each of these  according to this conversation and they all successfully compile with `new(registry: Arc<impl MetricsRegistry>)`.

chain/ethereum/src/block_ingestor.rs
fordN: I'm not sure I follow how that would work. Also generally, it is tricky because `ingest_blocks()` will get called again for any missing blocks.  

What about observing the `chain_head_number` from the `ChainHeadUpdateListener`?

Jannis: I like that idea!

chain/ethereum/src/block_ingestor.rs
Jannis: image

Works fine for me?

Jannis: That compiles over here.

fordN: But that's without the signature cleanup suggestions you made about the `BlockIngestor` struct right? ;)

server/http/src/service.rs
```diff
@@ -11,6 +13,58 @@ use hyper::{Body, Method, Request, Response, StatusCode};
...
+
+        let failed_query_execution_time = registry
+            .new_histogram_vec(
+                format!("subgraph_failed_query_execution_time"),
+                String::from("Execution time for failed GraphQL queries"),
+                HashMap::new(),
+                vec![String::from("subgraph_deployment")],
+                vec![0.1, 0.5, 1.0, 10.0, 100.0],
+            )
+            .expect("failed to create `subgraph_query_execution_time` histogram");
```
Jannis: This should say `subgraph_failed_query_execution_time`.

store/postgres/src/chain_head_listener.rs
```diff
@@ -56,6 +68,10 @@ impl ChainHeadUpdateListener {
                                 notification.payload
                             )
                         });
+                    metrics.set_chain_head_number(
```
Jannis: This deserves a comment IMHO

Jannis (changes requested):

store/postgres/src/chain_head_listener.rs
fordN: @jannis, I've added a comment describing the action and intent of using set_chain_head_number() here, please confirm the content is clear and then we should be good to merge this.

Jannis: Yup, that works!

Jannis (approved):

PR #1406, opened 2019-12-12
Simplify passing around MetricsRegistry

Jannis:
Simple changes to simplify all places where `MetricsRegistry` is passed around.


Jannis:
Merged along with #1403.

PR #1408, opened 2019-12-14
node, graph, store, chain: Remove unneeded blocks from the block cache

lutter:
Fixes https://github.com/graphprotocol/support/issues/45

### DIFF
--- a/chain/ethereum/src/block_ingestor.rs
+++ b/chain/ethereum/src/block_ingestor.rs
@@ -111,7 +111,7 @@ where
                         // Continue polling even if polling failed
                         future::ok(())
                     })
-                    .then(move |result| {
+                    .inspect(move |_| {
                         if cleanup_freq.as_secs() > 0 && last_cleanup.elapsed() > cleanup_freq {
                             match static_self.chain_store.cleanup_cached_blocks() {
                                 Ok((min_block, count)) => {
@@ -134,7 +134,6 @@ where
                             }
                             last_cleanup = Instant::now();
                         }
-                        result
                     })
             })
     }
--- a/chain/ethereum/src/block_ingestor.rs
+++ b/chain/ethereum/src/block_ingestor.rs
@@ -1,9 +1,17 @@
+use lazy_static;
 use std::collections::HashMap;
 use std::time::Duration;
 use std::time::Instant;
 
 use graph::prelude::*;
 use web3::types::*;
+
+lazy_static! {
+    static ref CLEANUP_BLOCKS: bool = std::env::var("GRAPH_ETHEREUM_CLEANUP_BLOCKS")
+        .ok()
+        .map(|s| s.eq_ignore_ascii_case("true"))
+        .unwrap_or(false);
+}
 
 pub struct BlockIngestorMetrics {
     chain_head_number: Box<GaugeVec>,
@@ -73,11 +81,9 @@ where
         })
     }
 
-    pub fn into_polling_stream(self, cleanup_freq: Duration) -> impl Future<Item = (), Error = ()> {
+    pub fn into_polling_stream(self) -> impl Future<Item = (), Error = ()> {
         // Currently, there is no way to stop block ingestion, so just leak self
         let static_self: &'static _ = Box::leak(Box::new(self));
-
-        let mut last_cleanup = Instant::now();
 
         // Create stream that emits at polling interval
         tokio::timer::Interval::new(Instant::now(), static_self.polling_interval)
@@ -112,7 +118,7 @@ where
                         future::ok(())
                     })
                     .inspect(move |_| {
-                        if cleanup_freq.as_secs() > 0 && last_cleanup.elapsed() > cleanup_freq {
+                        if *CLEANUP_BLOCKS {
                             match static_self
                                 .chain_store
                                 .cleanup_cached_blocks(static_self.ancestor_count)
@@ -135,7 +141,6 @@ where
                                     "network_name" => &static_self.network_name,
                                 ),
                             }
-                            last_cleanup = Instant::now();
                         }
                     })
             })
--- a/chain/ethereum/src/block_ingestor.rs
+++ b/chain/ethereum/src/block_ingestor.rs
@@ -73,9 +73,11 @@ where
         })
     }
 
-    pub fn into_polling_stream(self) -> impl Future<Item = (), Error = ()> {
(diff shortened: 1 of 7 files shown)


chain/ethereum/src/block_ingestor.rs
```diff
@@ -73,40 +73,69 @@ where
...
+                                }
                             }
-                            EthereumAdapterError::Unknown(inner_err) => {
-                                warn!(
+                        }
+
+                        // Continue polling even if polling failed
+                        future::ok(())
+                    })
+                    .then(move |result| {
```
Jannis: You could use `inspect` here, then you don't need to pass `result` through:
```rust
.inspect(move |_| {
  ...
})
```
I do wonder though, wouldn't this better be implemented via a separate Tokio task / thread?

node/src/main.rs
```diff
@@ -51,6 +51,13 @@ lazy_static! {
         .map(|s| usize::from_str(&s)
              .unwrap_or_else(|_| panic!("failed to parse env var GRAPH_TOKIO_THREAD_COUNT")))
         .unwrap_or(100);
+
+    static ref BLOCK_CACHE_CLEANUP_FREQ: Duration = env::var("GRAPH_BLOCK_CACHE_CLEANUP_FREQ")
+        .ok()
+        .map(|s| u64::from_str(&s)
+             .unwrap_or_else(|_| panic!("failed to parse env var GRAPH_BLOCK_CACHE_CLEANUP_FREQ")))
+        .map(|secs| Duration::from_secs(secs))
+        .unwrap_or(Duration::from_secs(0));
```
Jannis: Can we move this into `BlockIngestor`, rename it to `GRAPH_ETHEREUM_BLOCK_CLEANUP_INTERVAL`, and also drop the frequency argument to `into_polling_stream()`? I feel this is internal to Ethereum, for other chains in the future we may want to configure different intervals.

store/postgres/src/store.rs
```diff
@@ -1268,6 +1268,50 @@ impl ChainStore for Store {
...
+        struct MinBlock {
+            #[sql_type = "Integer"]
+            block: i32,
+        };
+
+        // Remove all blocks from the cache that are behind the slowest
+        // subgraph's head block, but retain the genesis block. We stay
+        // behind the slowest subgraph so that we do not interfere with its
+        // syncing activity.
+        // Only consider active subgraphs that have not failed
```
Jannis: That's pretty cool!

store/postgres/src/store.rs
```diff
@@ -1268,6 +1268,50 @@ impl ChainStore for Store {
...
+        let conn = self.get_conn()?;
+        let query = "select min((d.data->'latestEthereumBlockNumber'->>'data')::int) as block
+                       from subgraphs.entities d,
+                            subgraphs.entities a,
+                            subgraphs.entities ds
+                      where d.entity='SubgraphDeployment'
+                        and a.entity='SubgraphDeploymentAssignment'
+                        and ds.entity = 'EthereumContractDataSource'
```
Jannis: Nit: The spacing around `=` is inconsistent.

leoyvens:
Exactly how it all works is not clear in my mind right now, but the block ingestor makes sure to keep the latest `REORG_THRESHOLD` blocks in the database, and I believe the block stream relies on that assumption. Could deleting these recent blocks be a potential issue?

lutter:
> Exactly how it all works is not clear in my mind right now, but the block ingestor makes sure to keep the latest `REORG_THRESHOLD` blocks in the database, and I believe the block stream relies on that assumption. Could deleting these recent blocks be a potential issue?

Good point. I think the minimum block number we keep needs to be `min(subgraph heads, chain head - REORG_THRESHOLD)`

node/src/main.rs
lutter: I am actually thinking that it shouldn't be an interval at all, and we should do that cleanup after each block we ingest, i.e., make that environment variable a boolean. That way, we will most often clean up just a small number of blocks, and it'll help in keeping `ethereum_blocks` small.

chain/ethereum/src/block_ingestor.rs
lutter: What's the concern here? When you do cleanups, the overall slowdown of the system because you need to continually download blocks will far outweigh any slowdown from adding a `delete` query into the blockstream.

node/src/main.rs
lutter: I made it a boolean, and changed the name to `GRAPH_ETHEREUM_CLEANUP_BLOCKS` and moved it into the block ingestor

lutter:
Addressed all review comments. We now only clean up blocks that are before the minimum of any subgraph heads and before `ANCESTOR_COUNT` of the chain head.

chain/ethereum/src/block_ingestor.rs
Jannis: I'm not sure I understand the "continually download blocks" part. I was just wondering if running this delete query every time there is a new chain head would slow down the ingestion, but since we can configure how frequently that happens, I think we're good here. It was more a question around a separation of concerns.

chain/ethereum/src/block_ingestor.rs
```diff
@@ -84,29 +92,57 @@ where
...
+
+                        // Continue polling even if polling failed
+                        future::ok(())
+                    })
+                    .inspect(move |_| {
+                        if *CLEANUP_BLOCKS {
+                            match static_self
+                                .chain_store
+                                .cleanup_cached_blocks(static_self.ancestor_count)
+                            {
```
Jannis: I liked the ability to control the interval. Do we think this is cheap enough to run every time?

Jannis: Ah, well, it's a development time thing, so let's not worry about it. 😄

lutter: That, and I really wouldn't know what to tell users how to choose a good interval. Testing this locally with one syncing subgraph, it would generally remove one or two blocks, and that did not slow the block stream down at all, it's very quick. I had (from previous syncs) ~ 20k blocks in the DB; cleanup whittled it down to about 2k blocks while it was syncing without causing any noticeable delay.

On the 'continually download blocks': yeah, that phrasing could be improved, I just wanted to say something sufficiently scary so people don't turn this on in production settings. But I am open to rewording that, just not sure how.

lutter:
Rebased to latest master

PR #1413, opened 2019-12-16
Ensure correct Ethereum trigger order

Jannis:
Ethereum nodes do not guarantee returning events in the right order, especially not if they come from the same transaction. This is why we need ensure the right order before processing.

This PR sorts Ethereum events by their `log_index` (which for each event is their index in the block) and verifies that this order has not been messed with just before the triggers of a block are processed.

### DIFF
--- a/graph/src/components/ethereum/types.rs
+++ b/graph/src/components/ethereum/types.rs
@@ -77,20 +77,7 @@ pub struct EthereumBlockWithTriggers {
 impl EthereumBlockWithTriggers {
     pub fn new(mut triggers: Vec<EthereumTrigger>, ethereum_block: BlockFinality) -> Self {
         // Sort the triggers
-        triggers.sort_by(|a, b| {
-            let a_tx_index = a.transaction_index();
-            let b_tx_index = b.transaction_index();
-            if a_tx_index.is_none() && b_tx_index.is_none() {
-                return Ordering::Equal;
-            }
-            if a_tx_index.is_none() {
-                return Ordering::Greater;
-            }
-            if b_tx_index.is_none() {
-                return Ordering::Less;
-            }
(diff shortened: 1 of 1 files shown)


graph/src/components/ethereum/types.rs
```diff
@@ -78,8 +78,9 @@ impl EthereumBlockWithTriggers {
     pub fn new(mut triggers: Vec<EthereumTrigger>, ethereum_block: BlockFinality) -> Self {
         // Sort the triggers
         triggers.sort_by(|a, b| {
-            let a_tx_index = a.transaction_index();
-            let b_tx_index = b.transaction_index();
+            let a_tx_index = a.sort_index();
```
olibearo: Let's change the var names to reflect sort_index?

graph/src/components/ethereum/types.rs
```diff
@@ -177,10 +179,10 @@ pub enum EthereumBlockTriggerType {
 }
 
 impl EthereumTrigger {
-    fn transaction_index(&self) -> Option<u64> {
+    fn sort_index(&self) -> Option<u64> {
         match self {
-            // We only handle logs that are in a block and therefore have a `transaction_index`.
-            EthereumTrigger::Log(log) => Some(log.transaction_index.unwrap().as_u64()),
+            // We only handle logs that are in a block and therefore have a `log_index`.
+            EthereumTrigger::Log(log) => Some(log.log_index.unwrap().as_u64()),
```
olibearo: Doesn't this mean we will be sorting by two different types of numbers? `EthereumTrigger::Call` will be sorted by its `transaction_index` and `EthereumTrigger::Log` will be sorted by its `log_index`. That is deterministic, but it could result in odd ordering.

ghost (approved):
Nice

PR #1416, opened 2019-12-19
Introduce an `LfuCache` for entities

leoyvens:
Part of #1381. The cache works by balancing frequency and recency, it's basically an LFU cache but can also detect stale entries. The cache size is limited by the estimated memory usage. It is integrated into the `EntityCache`, which is initialized with the starting state of the cache and returns the updated cache from `as_modifications` to be stashed in the `IndexingState` and passed in again on the next block to be processed.

The `BlockStream` now emits a special item when a revert happens, so we can clear the cache on reverts.

Edit: I gave up on the previous "frecency" based idea, and went for something simpler.


leoyvens:
This is ready for review. Local benchmark results with ERC20:

Captura de Tela 2020-01-05 às 19 28 17

Comparing with the previous benchmark which can be found in #1388. `store.get` time is down from ~15% of indexing time to ~7%. The subgraph got to block 4880791 vs 4844255 previously. This is not meant to necessarily represent production performance but to show that `store.get` got faster and overall things got faster, which is sufficient to merge this and try it out in production.

graph/src/data/store/mod.rs
```diff
@@ -521,6 +523,32 @@ pub trait ToEntityKey {
...
+    /// The weight of an entity in the cache is the approximate amount of bytes occupied by it.
+    fn weight(&self) -> u64 {
+        use std::mem::size_of_val;
+        self.0
+            .iter()
+            .map(|(key, value)| size_of_val(key) as u64 + key.len() as u64 + value.weight())
+            .sum()
+    }
+}
+
```
Jannis: Nice!

graph/src/util/lfu_cache.rs
```diff
@@ -0,0 +1,192 @@
...
+        if self.total_weight <= max_weight {
+            return;
+        }
+
+        self.stale_counter += 1;
+        if self.stale_counter == STALE_PERIOD {
+            self.stale_counter = 0;
+
+            // Entries marked `will_stale` are effectively set as stale.
+            // All entries are then marked `will_stale`.
```
Jannis: I understand how it works but the comment is confusing. How about something like
```
// Entries marked `will_stale` were not updated in this period. Properly mark them
// as stale in their priorities. Also mark all entities as `will_stale` for the _next_ period
// so that they will be marked stale next time unless they are updated between now
// and then. 
```

About the stale behavior itself: Entries that are not updated in a period become stale. Do we care if they are updated at all? If they are read a lot but never written to, I'd assume we want them to remain cached.

graph/src/util/lfu_cache.rs
```diff
@@ -0,0 +1,192 @@
...
+            for (e, p) in self.queue.iter_mut() {
+                p.0 = e.will_stale;
+                e.will_stale = true;
+            }
+        }
+
+        while self.total_weight > max_weight {
+            // Unwrap: If there were no entries left, `total_weight` would be 0.
+            let entry = self.queue.pop().unwrap().0;
```
Jannis: Could be worth changing `unwrap` to `expect("no entries left in cache but total_weight incorrectly is > max_weight")`? Then the comment becomes the error and we're told what's up if this is ever triggered.

graph/src/util/lfu_cache.rs
```diff
@@ -0,0 +1,192 @@
...
+    fn into_iter(self) -> Self::IntoIter {
+        Box::new(self.queue.into_iter())
+    }
+}
+
+impl<K: Ord + Eq + Hash, V> Extend<(CacheEntry<K, V>, Priority)> for LfuCache<K, V> {
+    fn extend<T: IntoIterator<Item = (CacheEntry<K, V>, Priority)>>(&mut self, iter: T) {
+        self.queue.extend(iter);
+    }
+}
```
Jannis: The `LfuCache` deserves unit tests, wouldn't you agree? 😉

graph/src/components/store.rs
```diff
@@ -1410,11 +1415,14 @@ impl EntityCache {
...
     /// `EntityModification`, making sure to only produce one when a change
-    /// to the current state is actually needed
+    /// to the current state is actually needed.
+    ///
+    /// Also returns the updated `LfuCache`.
     pub fn as_modifications(
         mut self,
         store: &(impl Store + ?Sized),
-    ) -> Result<Vec<EntityModification>, QueryExecutionError> {
+    ) -> Result<(Vec<EntityModification>, LfuCache<EntityKey, Option<Entity>>), QueryExecutionError>
```
Jannis: I feel it would be easier to understand code that uses this method if we returned a struct instead of a tuple here. Expressions like
```rust
.as_modifications(...).0
```
are sufficiently obvious inside newtype implementations but don't make for a good external API. 
```rust
.as_modifications(...).modifications
.as_modifications(...).entity_cache
```
would be more readable. This is definitely on the nit-picky side though. What do you think?

core/src/subgraph/instance_manager.rs
```diff
@@ -503,33 +515,46 @@ where
...
-        })
+        // Process events from the stream as long as no restart is needed
+        .fold(
+            ctx,
+            move |mut ctx, event| -> Box<dyn Future<Item = _, Error = _> + Send> {
+                let block = match event {
+                    BlockStreamEvent::Revert => {
+                        // On revert, clear the entity cache.
+                        ctx.state.entity_lfu_cache = LfuCache::new();
+                        return Box::new(future::ok(ctx));
```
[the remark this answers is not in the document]
leoyvens: This is a coercion and therefore can be implicit, but a type hint is needed somewhere. In this case I found it cleaner to put the type hint in the closure return type.

core/src/subgraph/instance_manager.rs
```diff
@@ -34,6 +44,7 @@ struct IndexingState<T: RuntimeHostBuilder> {
     call_filter: EthereumCallFilter,
     block_filter: EthereumBlockFilter,
     restarts: u64,
+    entity_lfu_cache: LfuCache<EntityKey, Option<Entity>>,
```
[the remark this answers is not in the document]
leoyvens: This is so we don't confuse it with the `EntityCache`.

core/src/subgraph/instance_manager.rs
```diff
@@ -625,8 +650,8 @@ where
     // collected previously to every new event being processed
     process_triggers(
         logger.clone(),
+        BlockState::with_cache(std::mem::take(&mut ctx.state.entity_lfu_cache)),
```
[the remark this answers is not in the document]
leoyvens: Both, but more the second, it's indeed like `Option::take`.

core/src/subgraph/instance_manager.rs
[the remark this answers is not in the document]
Jannis: Ok, point taken.

core/src/subgraph/instance_manager.rs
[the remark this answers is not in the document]
Jannis: I like it!

graph/src/components/store.rs
leoyvens: This really isn't very readable unless you destructure as `let (modifications, _) = ...`. I replaced this with a struct so the modifications and cache can be accessed by name.

graph/src/util/lfu_cache.rs
leoyvens: Yes! I've added one.

graph/src/util/lfu_cache.rs
leoyvens: I don't like how `expect` reads and how it clutters the code, but since `unwrap` doesn't yet give line numbers, it's reasonable that we should use a specific message.

graph/src/util/lfu_cache.rs
leoyvens: That is a great comment, I've used it.

It was a mistake for `get` to not reset the staleness! I've fixed that now.

core/src/subgraph/instance_manager.rs
```diff
@@ -747,6 +772,16 @@ where
...
         section.end();
 
+        let section = ctx
+            .host_metrics
+            .stopwatch
+            .start_section("entity_cache_evict");
+        cache.evict(*ENTITY_CACHE_SIZE);
+        section.end();
+
+        assert!(ctx.state.entity_lfu_cache.is_empty());
```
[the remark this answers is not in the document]
leoyvens: This would say `assertion failed: ctx.state.entity_lfu_cache.is_empty()` which lets us know that this is the assert being hit, which is really all we need for debugging. For clarity, I'll add a comment here.

leoyvens:
@Jannis I've addressed all comments.

graph/src/util/lfu_cache.rs
Jannis: Yeah, I share the sentiment about `expect`. I'm always inclined to write `.expect("something to happen")` but really you need to write `.expect("something didn't happen")` which just feels _so_ wrong from a grammar standpoint...

Jannis: It should be `.or_panic("something didn't happen")`.

graph/src/util/lfu_cache.rs
```diff
@@ -0,0 +1,253 @@
...
+            return;
+        }
+
+        self.stale_counter += 1;
+        if self.stale_counter == STALE_PERIOD {
+            self.stale_counter = 0;
+
+            // Entries marked `will_stale` were not accessed in this period. Properly mark them as
+            // stale in their priorities. Also mark all entities as `will_stale` for the _next_
+            // period so that they will be marked stale next time unless they are updated between
```
Jannis: `updated` -> `updated or looked up`

leoyvens: Thanks, fixed

core/src/subgraph/instance_manager.rs
```diff
@@ -10,9 +11,18 @@ use graph::data::subgraph::schema::{
     DynamicEthereumContractDataSourceEntity, SubgraphDeploymentEntity,
 };
 use graph::prelude::{SubgraphInstance as SubgraphInstanceTrait, *};
+use graph::util::lfu_cache::LfuCache;
 
 use super::SubgraphInstance;
 
+lazy_static! {
+    pub static ref ENTITY_CACHE_SIZE: u64 = 1000
```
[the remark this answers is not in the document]
leoyvens: Added a comment

graph/src/data/store/mod.rs
```diff
@@ -521,6 +523,32 @@ pub trait ToEntityKey {
     fn to_entity_key(&self, subgraph: SubgraphDeploymentId) -> EntityKey;
 }
 
+impl CacheWeight for Value {
+    fn weight(&self) -> u64 {
+        use std::mem::size_of_val;
+        size_of_val(self) as u64
+            + match self {
+                Value::String(s) => s.len() as u64,
+                Value::BigDecimal(d) => (d.digits() as f32).log2() as u64,
```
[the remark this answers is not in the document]
leoyvens: That would be more direct, but this is the best I could get out of the `BigDecimal` public api.

graph/src/data/store/mod.rs
```diff
@@ -521,6 +523,32 @@ pub trait ToEntityKey {
...
+                Value::List(values) => values.iter().map(|value| value.weight()).sum(),
+                Value::Bytes(bytes) => bytes.as_slice().len() as u64,
+                Value::BigInt(n) => n.bits() / 8 as u64,
+                Value::Int(_) | Value::Bool(_) | Value::Null => 0,
```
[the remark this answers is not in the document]
leoyvens: There's nothing to add here beyond `size_of_val`, these values are fixed size so kept entirely in the stack. The size of the key dwarfs them anyways.

graph/src/data/store/mod.rs
```diff
@@ -521,6 +523,32 @@ pub trait ToEntityKey {
...
+    }
+}
+
+impl CacheWeight for Entity {
+    /// The weight of an entity in the cache is the approximate amount of bytes occupied by it.
+    fn weight(&self) -> u64 {
+        use std::mem::size_of_val;
+        self.0
+            .iter()
+            .map(|(key, value)| size_of_val(key) as u64 + key.len() as u64 + value.weight())
```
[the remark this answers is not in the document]
leoyvens: `size_of_val` will give us the stack size of the value. Well, the stack size if it were placed on the stack. The crucial thing is that it won't follow pointers, so it won't know the size of a variable length type such as a string or array. That part we need to estimate manually.

leoyvens:
@Jannis I checked this on the Moloch subgraph and also with a test subgraph that records the number and hash of each block, and waited until that handled a revert correctly. We also have tests that use `EntityCache`, and those will transitively test the lru cache.

PR #1422, opened 2019-12-28
Fix error in GraphQL execution caused by overly aggressive assertion in prefetch

lutter:
The assertion in `prefetch.rs` that this PR removes was too aggressive, as there is no reason for it to be true in the situation created by the test `interface_inline_fragment_with_subquery` in `core/tests/interfaces.rs`

This change simply replaces that assertion by skipping the attempt to add children to a parent whose type is not mentioned in any join condition. The queries did the right thing already.

The reverse assertion on line 338 in `prefetch.rs` is still valid: if a child type is not mentioned in a join condition, that join condition should not even be there, which would indicate some logic error somewhere else in the code.

### DIFF
--- a/core/tests/interfaces.rs
+++ b/core/tests/interfaces.rs
@@ -474,6 +474,80 @@ fn interface_inline_fragment() {
 }
 
 #[test]
+fn interface_inline_fragment_with_subquery() {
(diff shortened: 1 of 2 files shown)


Jannis:
Tested locally with the Livepeer subgraph where we encountered the problem initially.

PR #1428, opened 2020-01-02
Fix setting fields to null being ignored in certain cases

Jannis:
This fixes a bug in `EntityCache` where setting an entity field to `Value::Null` would not take effect when there were two changes to the same entity, with any number of changes in-between, where the first change would set the field to a non-null value and the later change would set the field to `Value::Null`.

The problem was caused by `EntityCache` merging entities twice: all in-flight changes were merged together with `Entity::merge`, dropping all `Value::Null` fields. The combined changes were later merged into the original entity from the store using the same merge algorithm; except that now all fields that should be set to `Value::Null` were already removed from the combined changes and it was as if they had never happened.

This PR renames the existing `Entity::merge` method to `Entity::merge_remove_null_fields` to make obvious what it does. In addition, a new `Entity::merge` method is added that preserves `Value::Null` fields.

The `EntityCache` uses the new `Entity::merge` method to merge in-flight changes, preserving requests to set fields to `Value::Null`. It then uses the `Entity::merge_remove_null_fields` method to merge these changes into the original entity, so that any null fields are removed from the final entity, as expected.

All places using the old `Entity::merge` method are updated to use `Entity::merge_remove_null_fields`.

Basic tests are added for `EntityCache`. These would've caught the bug (I verified this by making `merge` and `merge_remove_null_fields` behave the same).

### DIFF
--- a/graph/src/components/store.rs
+++ b/graph/src/components/store.rs
@@ -2,6 +2,8 @@ use failure::Error;
 use futures::stream::poll_fn;
 use futures::{Async, Future, Poll, Stream};
 use lazy_static::lazy_static;
+use mockall::predicate::*;
+use mockall::*;
 use serde::{Deserialize, Serialize};
 use std::collections::{BTreeMap, HashMap, HashSet};
 use std::env;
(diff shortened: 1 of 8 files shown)


Jannis:
I'll merge it to unblock projects on the hosted service. I'd still appreciate a post-merge review. 😄

PR #1429, opened 2020-01-03
graph, graphql: Log variables passed into GraphQL query with query

lutter:
For some queries, there is huge variation in the time a query takes depending on the values of the variables passed in. Log GraphQL variables together with the query to make it easier to understand why a certain query is slow.


Jannis:
@lutter Don't forget to merge this one. 😉

lutter:
Somehow missed that that had been approved

PR #1430, opened 2020-01-04
Use automocked store and Ethereum adapter in more tests

Jannis:
Not a particularly interesting change, just getting rid of some mock code and replacing it with automocking.


leoyvens (approved):
Yay for less fiddling with mocks.

PR #1437, opened 2020-01-07
Reject time-travel queries with block numbers in the future

lutter:


graphql/src/store/resolver.rs
```diff
@@ -249,7 +249,25 @@ where
...
+                .block_ptr(bc.subgraph.clone())
+                .map_err(|e| StoreError::from(e).into())
+                .and_then(|ptr| {
+                    let ptr = ptr.expect("we should have already checked that the subgraph exists");
+                    if ptr.number < number as u64 {
+                        Err(QueryExecutionError::ValueParseError(
+                            "block.number".to_owned(),
+                            format!(
+                                "subgraph {} has only indexed up to block number {} \
+                                 and data for block number {} is therefore not yet available",
```
Jannis: People are so going to mis-use this as a status API by querying against with `{ block: { number: MAX_INT } }`. 😉 It's the right thing to do though.

Jannis:
The only comment I'm pondering is whether we should quote the subgraph name in the error message... I'm undecided and it's a minor decision anyway.

PR #1440, opened 2020-01-08
Ethereum: Add support for calling overloaded contract functions

Jannis:
Part of resolving https://github.com/graphprotocol/support/issues/48.
Depends on https://github.com/graphprotocol/ethabi/pull/11 and must be rebased after that is merged.

This matches contract calls made from mappings against function definitions in the ABI based on their signature rather than just their name. This allows to correctly identify which exact function definition should be used, which was so far impossible / broken with overloaded functions.

### DIFF
--- a/runtime/wasm/src/mapping.rs
+++ b/runtime/wasm/src/mapping.rs
@@ -160,7 +160,7 @@ impl Clone for MappingContext {
 /// A pre-processed and valid WASM module, ready to be started as a WasmiModule.
 pub(crate) struct ValidModule {
     pub(super) module: wasmi::Module,
-    pub(super) user_module: Option<String>,
+    pub(super) host_module_names: Vec<String>,
 }
 
 impl ValidModule {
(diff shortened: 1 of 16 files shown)


runtime/wasm/src/module/mod.rs
```diff
@@ -998,7 +998,12 @@ where
             }
             ETHEREUM_CALL_FUNC_INDEX => {
                 let _section = stopwatch.start_section("host_export_ethereum_call");
-                self.ethereum_call(args.nth_checked(0)?)
+                let arg = if self.ctx.host_exports.api_version >= Version::new(0, 0, 4) {
+                    self.asc_get::<_, AscUnresolvedContractCall_0_0_4>(args.nth_checked(0)?)
```
fordN: It isn't clear to me how the call function indices differ per APIVersion. Does the `_0_0_4` suffix simply mean that struct was created for version 0.0.4?

runtime/wasm/src/to_from/external.rs
```diff
@@ -390,6 +390,19 @@ impl FromAscObj<AscUnresolvedContractCall> for UnresolvedContractCall {
             contract_name: heap.asc_get(asc_call.contract_name),
             contract_address: heap.asc_get(asc_call.contract_address),
             function_name: heap.asc_get(asc_call.function_name),
+            function_signature: None,
+            function_args: heap.asc_get(asc_call.function_args),
+        }
+    }
+}
+
+impl FromAscObj<AscUnresolvedContractCall_0_0_4> for UnresolvedContractCall {
```
fordN: Ok the `FromAscObj` impl difference for version 0.0.4 is clear here, you can disregard my question above.  Maybe it's worth including a brief comment explaining this situation on this impl or the ETHEREUM_CALL_FUNC_INDEX where it is used?

runtime/wasm/src/mapping.rs
```diff
@@ -173,34 +173,20 @@ impl ValidModule {
...
-            format_err!(
-                "Invalid module `{}`: {}",
-                user_module.as_ref().unwrap_or(&String::new()),
-                e
-            )
-        })?;
+        host_module_names.dedup();
+
+        let module = wasmi::Module::from_parity_wasm_module(parsed_module)
+            .map_err(|e| format_err!("Invalid WASM module: {}", e))?;
```
fordN: Though this isn't a user controlled value it is usually helpful to see the item name in the error; should we keep the module name in the error message here?

runtime/wasm/src/module/mod.rs
Jannis: I believe you introduced this `_x_y_z` version notation yourself. (I could be wrong about this 😉.)

`_0_0_4` means introduced in `apiVersion: 0.0.4`.

runtime/wasm/src/to_from/external.rs
Jannis: Yep, I've added a comment right where `ETHEREUM_CAL_FUNC_INDEX` is handled.

runtime/wasm/src/mapping.rs
Jannis: There isn't a single `user_module` to log anymore though...

runtime/wasm/src/module/mod.rs
fordN: Indeed I did. 😊

runtime/wasm/src/mapping.rs
fordN: Good point; I got `parsed_module` and `user_module` mixed up here. 
Ok, let's leave it as is.

fordN (approved):
LGTM 👍

PR #1441, opened 2020-01-08
Add integration test for calling overloaded contract functions

Jannis:
Adds an integration test for https://github.com/graphprotocol/support/issues/48.
Depends on `graph test` (https://github.com/graphprotocol/graph-cli/pull/414).


Jannis:
This test will not succeed until `graph test` and overloaded contract functions support have been merged into graph-cli and graph-ts.

Jannis:
The integration test suite can be seen running successfully here: https://travis-ci.org/graphprotocol/graph-node/jobs/637589758#L3528

leoyvens:
@Jannis what would you think of using a Rust test to run these, instead of a shell script, so we can leverage the Rust test runner?

Jannis:
@leoyvens Could work, let me try that.

leoyvens (approved):

Jannis:
@leoyvens I made more changes:

* Added an `.ignore` file for `cargo watch` to ignore changes to certain integration test dirs (`node_modules/`, `build/`, `generated/`).
* Moved the integration tests into a different directory to make the above easier.
* Switch to `duct` for running child processes and only print command output when a command fails. (Using duct here mainly for its `stderr_to_stdout()` feature that allows to get _everything_ the command prints into a single output string.)

leoyvens (approved):

PR #1443, opened 2020-01-10
Implement schema merging

olibearo:
Implements https://github.com/graphprotocol/rfcs/pull/6


store/postgres/src/store.rs
```diff
@@ -698,9 +703,26 @@ impl Store {
...
+                    .unstable_imports
+                    .iter()
+                    .any(|(schema_reference, subgraph_id_opt)| {
+                        self.resolve_schema_reference(schema_reference)
+                            .map(|schema| match subgraph_id_opt {
+                                Some(subgraph_id) => schema.id.eq(subgraph_id),
+                                None => true,
+                            })
+                            .unwrap_or(false)
+                    });
```
Jannis: Don't we also have to check the previously available schemas that were imported? Some of them might be gone now.

store/postgres/src/store.rs
```diff
@@ -698,9 +703,26 @@ impl Store {
...
     }
 
-    fn cached_schema(&self, subgraph_id: &SubgraphDeploymentId) -> Result<SchemaPair, Error> {
-        if let Some(pair) = self.schema_cache.lock().unwrap().get(&subgraph_id) {
-            return Ok(pair.clone());
+    fn cached_schema(&self, subgraph_id: &SubgraphDeploymentId) -> Result<SchemaCacheEntry, Error> {
+        if let Some(entry) = self.schema_cache.lock().unwrap().get(&subgraph_id) {
+            // Cache entry is stale and an unstable import has changed or is now available
+            // TODO: Is it okay to ignore unstable imports which are no longer available?
+            // TODO: Make this threshold an envvar?
```
Jannis: Yes, an env var sounds good.

store/postgres/src/store.rs
```diff
@@ -727,18 +749,44 @@ impl Store {
...
         // API schema have a @subgraphId directive as well
+        let (imported_schemas, schema_import_errors) = self.resolve_import_graph(&input_schema);
+        let mut unstable_schemas = imported_schemas
+            .iter()
+            .filter_map(|(schema_reference, schema)| match schema_reference {
+                SchemaReference::ByName(_) => {
+                    Some((schema_reference.clone(), Some(schema.id.clone())))
+                }
+                _ => None,
+            })
```
Jannis: Aren't references by ID also unstable, as they may or may not be available all the time?

PR #1444, opened 2020-01-10
Improve block ingestor logging

Jannis:
A tiny improvement to be able to trace block ingestor logs back to specific networks.


lutter (approved):
LGTM

Issue #1414, opened 2019-12-17
Allow using `Bytes` for `ID` and switch block explorer to that


lutter:
Currently, the `ID` type for entities is always mapped to a `String`/`text` column in the database. For id's that are really hex strings/bytes that's wasteful as the string representation takes about 2x as much storage as a representation as `Bytes`/`bytea`.

The relational storage code has some provisions for using either `String` or `Bytes`, but they are not fully finished and tested. For the block explorer, we should switch id's to `Bytes`. To keep the changes in the rest of the code to a minimum, we will do this solely within the storage code. The rest of the code (when dealing with `Entity`) will still get a `String` id, though that is expected to be formatted as a hex string `0xdeadbeef...`

The following needs to happen for that:

* Make `IdType` an argument for `Store::create_subgraph_deployment`; the block explorer will pass `IdType::Bytes` there, while all other subgraphs will pass `IdType::String`
* When a subgraph is started, look at `information_schema.columns` to decide whether the subgraph uses `String` or `Bytes` for ids
* Fix up the places in `relational_queries.rs` that currently assume that ids are strings and write tests for that (mostly requires switching `out.push_bind_param::<Text, _>(self.entity_id)` to something better)

PR #1445, opened 2020-01-12
Allow using `Bytes` for `ID` and switch block explorer to that

lutter:
Map between strings that represent bytes and `bytea` in the database; the difference between subgraphs that use `Bytes` for `ID` and those that use `String` is transparent to subgraph users. The only exception is that when an entity is stored with an id `0xdeadbeef` it will be returned as the string `deadbeef` stripping the `0x` prefix.

Currently, this is only used for network indexing; all user-supplied subgraphs still use `String` as their `ID` and there is no way for users to switch to `Bytes`.

Fixes https://github.com/graphprotocol/graph-node/issues/1414

### DIFF
--- a/chain/ethereum/src/network_indexer/subgraph.rs
+++ b/chain/ethereum/src/network_indexer/subgraph.rs
@@ -88,8 +88,12 @@ fn create_subgraph(
         spec_version: String::from("0.0.1"),
         description: None,
         repository: None,
-        schema: Schema::parse(include_str!("./ethereum.graphql"), subgraph_id.clone())
-            .expect("valid Ethereum network subgraph schema"),
+        schema: Schema::parse(
+            include_str!("./ethereum.graphql"),
+            subgraph_id.clone(),
+            IdType::String,
+        )
+        .expect("valid Ethereum network subgraph schema"),
         data_sources: vec![],
         templates: vec![],
     };
--- a/chain/ethereum/src/network_indexer/subgraph.rs
+++ b/chain/ethereum/src/network_indexer/subgraph.rs
@@ -91,7 +91,7 @@ fn create_subgraph(
         schema: Schema::parse(
             include_str!("./ethereum.graphql"),
             subgraph_id.clone(),
-            IdType::String,
+            IdType::Bytes,
         )
         .expect("valid Ethereum network subgraph schema"),
         data_sources: vec![],
--- a/server/http/src/request.rs
+++ b/server/http/src/request.rs
@@ -85,18 +85,25 @@ mod tests {
 
     const EXAMPLE_SCHEMA: &'static str = "type Query @entity { users: [User!] }";
 
+    fn example_schema() -> Schema {
+        Schema::parse(
+            EXAMPLE_SCHEMA,
+            SubgraphDeploymentId::new("test").unwrap(),
+            IdType::String,
+        )
+        .unwrap()
+    }
+
     #[test]
     fn rejects_invalid_json() {
-        let schema =
-            Schema::parse(EXAMPLE_SCHEMA, SubgraphDeploymentId::new("test").unwrap()).unwrap();
+        let schema = example_schema();
         let request = GraphQLRequest::new(hyper::Chunk::from("!@#)%"), Arc::new(schema));
         request.wait().expect_err("Should reject invalid JSON");
     }
 
     #[test]
     fn rejects_json_without_query_field() {
-        let schema =
-            Schema::parse(EXAMPLE_SCHEMA, SubgraphDeploymentId::new("test").unwrap()).unwrap();
+        let schema = example_schema();
         let request = GraphQLRequest::new(hyper::Chunk::from("{}"), Arc::new(schema));
         request
             .wait()
@@ -105,8 +112,7 @@ mod tests {
 
     #[test]
     fn rejects_json_with_non_string_query_field() {
-        let schema =
-            Schema::parse(EXAMPLE_SCHEMA, SubgraphDeploymentId::new("test").unwrap()).unwrap();
+        let schema = example_schema();
         let request = GraphQLRequest::new(hyper::Chunk::from("{\"query\": 5}"), Arc::new(schema));
         request
             .wait()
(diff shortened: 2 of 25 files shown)


lutter:
Rebased to latest master

lutter:
Rebased to latest master

chain/ethereum/src/network_indexer/subgraph.rs
```diff
@@ -88,8 +88,12 @@ fn create_subgraph(
         spec_version: String::from("0.0.1"),
         description: None,
         repository: None,
-        schema: Schema::parse(include_str!("./ethereum.graphql"), subgraph_id.clone())
-            .expect("valid Ethereum network subgraph schema"),
+        schema: Schema::parse(
+            include_str!("./ethereum.graphql"),
+            subgraph_id.clone(),
+            IdType::Bytes,
```
Jannis: Before I review the entire PR: Is this really a schema-wide setting and not a per-entity setting?

lutter: It is, but only as far as expressing what should be `IdType::Bytes` is concerned; the `Layout` does not assume that it's schema-wide. If we want this to be per entity, we need to figure out a way how users express which entities use `IdType::Bytes`; I can think of two ways: either add a directive so that users say `type Foo { id: ID! @OnlyBytes }` or that we allow users to declare `id` as `Bytes`; either won't be much work, but I am not sure if using anything but type `ID` has ripple effects in other places - we would still represent `id` in an `Entity` as a Rust `String`; anything else would be a huge change all over the code base.

Jannis: Agreed. `id: ID! @bytes` or `id: Bytes!` could both work. Not sure right now, which one would be better.

server/http/src/request.rs
```diff
@@ -85,18 +85,25 @@ mod tests {
 
     const EXAMPLE_SCHEMA: &'static str = "type Query @entity { users: [User!] }";
 
+    fn example_schema() -> Schema {
+        Schema::parse(
+            EXAMPLE_SCHEMA,
+            SubgraphDeploymentId::new("test").unwrap(),
+            IdType::String,
+        )
```
Jannis: We could get away with a smaller diff in this PR if we kept `Schema::parse` the way it is, had it set `IdType::String` internally, and added a separate `Schema::parse_with_id_type` method. I'll leave it to you to decide what you prefer, it's just an idea.

store/postgres/src/entities.rs
```diff
@@ -180,6 +180,18 @@ mod public {
             state -> crate::entities::public::DeploymentSchemaStateMapping,
         }
     }
+
+    // Access to the bits of the information_schema we care about. It conists
```
Jannis: Typo: conists -> consists

store/postgres/src/entities.rs
```diff
@@ -894,6 +907,51 @@ fn entity_from_json(json: serde_json::Value, entity: &str) -> Result<Entity, Sto
...
+        .select((columns::data_type, columns::table_name))
+        .filter(columns::column_name.eq("id"))
+        .filter(columns::table_schema.eq(schema_name.as_str()))
+        .first(conn)?;
+    let id_type = match data_type.as_str() {
+        "character varying" | "text" => IdType::String,
+        "bytea" => IdType::Bytes,
+        _ => {
+            return Err(StoreError::Unknown(format_err!(
+                "unsupported type `{}` for primary key column in table {} in schema {}(`{}`)",
```
Jannis: Shouldn't we put ``` `` ``` around the table `{}` placeholder too?

store/postgres/src/entities.rs
```diff
@@ -894,6 +907,51 @@ fn entity_from_json(json: serde_json::Value, entity: &str) -> Result<Entity, Sto
...
+        // that from the information_schema reliably, because `entity_history`
+        // has an integer `id` column
+        return Ok(IdType::String);
+    }
+
+    let (data_type, table_name): (String, String) = columns::table
+        .select((columns::data_type, columns::table_name))
+        .filter(columns::column_name.eq("id"))
+        .filter(columns::table_schema.eq(schema_name.as_str()))
+        .first(conn)?;
```
Jannis: Ok, I can see how this could easily be changed to obtain the ID type per entity.

store/postgres/src/relational.rs
```diff
@@ -113,15 +113,6 @@ impl fmt::Display for SqlName {
     }
 }
 
-/// The SQL type to use for GraphQL ID properties. We support
-/// strings and byte arrays
-#[derive(Copy, Clone, Debug, Eq, PartialEq)]
-pub enum IdType {
-    String,
-    #[allow(dead_code)]
-    Bytes,
```
Jannis: Interesting, I didn't know you had already prepared this. 😉

store/postgres/src/relational.rs
```diff
@@ -568,6 +564,17 @@ impl ColumnType {
...
+        }
+    }
+
+    /// Return the `IdType` corresponding to this column type. This can only
+    /// be called on a column that stores an `ID` and will panic otherwise
+    pub(crate) fn id_type(&self) -> IdType {
+        match self {
+            ColumnType::String => IdType::String,
+            ColumnType::BytesId => IdType::Bytes,
+            _ => unreachable!("only String and Bytes are allowed as primary keys"),
```
Jannis: This is `unimplemented!()` more than `unreachable!()`.

store/postgres/src/relational_queries.rs
```diff
@@ -179,6 +227,11 @@ impl<'a> QueryFragment<Pg> for QueryValue<'a> {
                     out.push_sql(name.as_str());
                     Ok(())
                 }
+                ColumnType::Bytes | ColumnType::BytesId => {
+                    let bytes = scalar::Bytes::from_str(&s)
+                        .map_err(|e| DieselError::SerializationError(Box::new(e)))?;
```
Jannis: Isn't this identical to the new `str_as_bytes` helper?

store/postgres/tests/relational_bytes.rs
```diff
@@ -0,0 +1,311 @@
...
+
+fn scrub(entity: &Entity) -> Entity {
+    let mut scrubbed = Entity::new();
+    // merge has the sideffect of removing any attribute
+    // that is Value::Null
+    scrubbed.merge(entity.clone());
+    scrubbed
+}
+
+macro_rules! assert_entity_eq {
```
Jannis: I have nothing against macros, but this to me looks like it could also just be a regular function; unless I'm missing something?

PR #1449, opened 2020-01-13
Make sure to check the block cache in reverts

leoyvens:
This solution is not yet ideal because it's possible that the block is neither in the cache nor in the chain, for example if the node is shutdown for a long time before the revert is processed, but it's still expected that it will be in the cache.


Jannis:
The CI failure is unrelated.

Issue #1394, opened 2019-12-03
Migrate subgraph metadata to relational storage


lutter:
The metadata for subgraphs (in the `subgraphs` database schema) still uses JSONB storage; it needs to be moved to relational storage so we can eventually delete the JSONB storage code.

There's two big things that need to be done here:

* write a database migration that creates a relational `subgraphs` schema and copies the data from the `entities` and `entity_history` tables into that schema
* make it possible to not record history for certain entities/tables as only subgraph metadata related to dynamic data sources should be versioned, and at least for the case of `SubgraphDeployment` versioning the data is completely nonsensical and would lead to enormously bloated tables

leoyvens:
We can only delete JSONB code after dropping support for subgraphs that were deployed under it, right? Or if we put those on a legacy system that stops getting updates.

Regarding the issue with only the dynamic data sources needing history, I wonder if we want to keep those in a shared table or move them into the per-subgraph schema.

lutter:
Yes, we absolutely need to have all subgraphs use relational storage before we can delete the JSONB code. We will also need to come up with and communicate a deprecation plan to users (on their end, it will consist of "Redeploy your subgraph by this date/release")

In prod, we currently have 460 JSONB subgraphs and 322 relational ones, so it looks like the natural cycle of updating subgraphs is taking care of a lot of this; but there will be a point where we need to identify inactive JSONB subgraphs and come up with a plan for how to move them forward or delete them with the owner's consent.

Either way, as long as subgraph metadata uses JSONB storage, it won't be possible to move away from JSONB.

As for dynamic data sources: I was going to leave them in shared tables, simply because it will reduce risk when making this change. Longer term, I think it's a worthwhile goal.

PR #1457, opened 2020-01-16
Switch subgraph metadata to relational storage

lutter:
Move subgraph metadata from JSONB storage to relational storage. One nice effect of this is that the one entity we change all the time, `SubgraphDeployment`, now lives in its own table rather than being intermingled with all the other metadata. Since we have ~ 700 deployments, but between 600k and 900k metadata entries, we are now making changes in a much smaller table.

The migration in this PR takes about 5 minutes against a production dump from yesterday.

I tested this by deploying new subgraphs locally (mostly Uniswap), and by running against a migrated production dump and indexing (from their state in the dump, not from the start) the subgraphs `protofire/makerdao-governance`, `graphprotocol/livepeer`, `graphprotocol/uniswap`, `protofire/token-registry`, `melonproject/melon`, and `daostack/alchemy`

This fixes https://github.com/graphprotocol/graph-node/issues/1394


store/postgres/src/subgraphs.graphql
```diff
@@ -71,7 +71,7 @@ type DynamicEthereumContractDataSource @entity {
 
 type EthereumContractSource @entity {
     id: ID!
-    address: String!
+    address: Bytes
```
Jannis: Contracts always have an address, I'm pretty sure, so we should keep the field as required with `!`.

store/postgres/src/entities.rs
```diff
@@ -527,8 +491,8 @@ impl Connection {
         entity: &Entity,
     ) -> Result<usize, StoreError> {
         match &*self.metadata {
-            Storage::Json(json) => json.update_metadata(&self.conn, key, entity),
-            Storage::Relational(_) => unreachable!("relational storeage is not used for metadata"),
+            Storage::Json(_) => unreachable!("JSONB storage of subgraph metadata is not supported"),
```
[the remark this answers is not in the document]
lutter: This really is unreachable: we've set up things so that we will never get here (if we do that indicates a bug somewhere) rather than something that's unimplemented, i.e., something that we should have implemented but haven't really gotten around to.

store/postgres/src/subgraphs.graphql
lutter: Unfortunately, they don't; we have 29 `EthereumContractSource` in hosted without an address. They belong to these subgraphs: 
```
ensdomains/ens
ensdomains/ensgoerli
ensdomains/ensrinkeby
ensdomains/ensropsten
luckyyang/ela-ens
luckyyang/elastos-eth-ens
makoto/ens-goerli
makoto/ens-mainnet
makoto/ens-subgraph-goerli-text
makoto/ens-subgraph-goerli-text-indexed
radek1st/metamultisig
radek1st/meta-multisig
radek1st/multisig
wighawag/eip721-subgraph
```

The `EthereumContractSource` in question are
```
QmPcFGJPhiq7NXUXeBUUiVbjv5D1RKKsg5Y29GNmP838bf-manifest-data-source-5-source
QmReUGTmnBqrjZierx4rvjgRMDxkZ5icHa6ECv1WfdUvw9-manifest-data-source-5-source
QmU7igtqrBQhLeD6SafF6YqUMvrtkbgfJCPmM17NRbneyR-manifest-data-source-1-source
QmTBmMKqcDH7GQegG6ktk6hWFzi2YRXAJU3XSvBrXPovJJ-manifest-data-source-1-source
QmRYNumG74pNtCA4jGftJkJm3hBgapEvDFN6yojjdBypm4-manifest-data-source-0-source
QmVEoWSQ8eNnVkL1uw2toDSc1xXKi1ADYUm36p961Q1EzT-manifest-data-source-5-source
QmZTHHbtEyGcDRp2mWiuSHMKU1PrHKh7KaMp1jDPSNDqRv-manifest-data-source-1-source
QmafxPzuuWFtLzxkEXMVBFpkkKcwavaK2D9BSLFDyMHQxK-manifest-data-source-1-source
QmXieLMydGH6rT8bN1jCi5JeLmsfkyjNKhxGoveWffVxBe-manifest-data-source-5-source
QmZUg1pXA7izP1TQ6krMy7wS2YzHSTbCwjnUPtfJnL2RkN-manifest-data-source-1-source
QmeqpV7maxD3pFZmcVSLyAuDyhFL6E7Vfw39ZmG1P1cdMM-manifest-data-source-0-source
QmTBmMKqcDH7GQegG6ktk6hWFzi2YRXAJU3XSvBrXPovJJ-manifest-data-source-5-source
QmUHceFFmgeR4tBF3F38DsZiGvozx6CEM4H5xXw7JYeU4t-manifest-data-source-1-source
QmReUGTmnBqrjZierx4rvjgRMDxkZ5icHa6ECv1WfdUvw9-manifest-data-source-1-source
QmQ491oCHnZiRS9LtS9nsTfbPd6EGLpcebcwMzUsou5aM7-manifest-data-source-1-source
QmU7igtqrBQhLeD6SafF6YqUMvrtkbgfJCPmM17NRbneyR-manifest-data-source-5-source
QmVEoWSQ8eNnVkL1uw2toDSc1xXKi1ADYUm36p961Q1EzT-manifest-data-source-1-source
```

leoyvens:
I don't mean to pile up additional changes, but the `templates` field within data sources is long deprecated, do you think this PR would be an opportunistic moment to drop support for it by not including it in the relational schemas?

lutter:
> I don't mean to pile up additional changes, but the `templates` field within data sources is long deprecated, do you think this PR would be an opportunistic moment to drop support for it by not including it in the relational schemas?

The change to drop that data is pretty easy: delete it form the GraphQL schema and regenerate the various generated bits of SQL. Are we sure that removing the `templates` attribute won't lead to trouble anywhere else? If so, I can add this when I brush up this PR.

leoyvens:
@lutter It would also need to be removed from the dynamic data sources query and from the corresponding data structures in `data/schema/subgraphs.rs`.

lutter:
@leoyvens We still have some subgraphs that have templates associated with data sources; they are all fairly old and are `aragonblack/aragonpm`, `iherger/melon`, `zeframlou/betoken`, and `zeframlou/betoken-mainnet-test`. I assume they won't break if I remove the `templates` attribute from data sources.

lutter:
Actually, code like [this](https://github.com/graphprotocol/graph-node/blob/master/core/src/subgraph/instance_manager.rs#L323-L347) makes me think there is more going on than just removing an unused data field. I now think it's safer to do that removal as a separate issue/PR. When we do that, the main work we'd save doing the change now is dropping the `templates` column from the two data source tables we'll have - that won't be a big deal since there are only ~ 150k data sources, so not a huge migration.

lutter:
Rebased to latest master

store/postgres/src/entities.rs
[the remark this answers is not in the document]
Jannis: True

store/postgres/src/subgraphs.graphql
Jannis: Of course, you're right. Addresses are optional on data sources.

leoyvens:
@lutter this would break those subgraphs, melon and betoken have their official subgraphs under a different org, the only one I'm unsure about is in use is `aragonblack/aragonpm`.

That code actually looks wrong in that it's not taking top level templates in consideration.

It's fine if you'd rather merge this and leave that to a future migration.

lutter:
Rebased to latest master

store/postgres/src/entities.rs
```diff
@@ -528,8 +492,8 @@ impl Connection {
...
+            Storage::Relational(layout) => layout.update_metadata(&self.conn, key, entity),
```
leoyvens: I love the elegance of this swap of the match arms.

store/postgres/src/relational.rs
```diff
@@ -460,6 +472,35 @@ impl Layout {
         Ok(())
     }
 
+    pub fn update_metadata(
```
leoyvens: Should this be `update_unversioned` to keep the naming consistent?

store/postgres/src/relational_queries.rs
```diff
@@ -926,6 +932,127 @@ impl<'a> QueryId for InsertQuery<'a> {
...
+impl<'a> QueryId for UpdateQuery<'a> {
+    type QueryId = ();
+
+    const HAS_STATIC_QUERY_ID: bool = false;
+}
+
+impl<'a, Conn> RunQueryDsl<Conn> for UpdateQuery<'a> {}
+
+/// A query that removes all versions whose block range lies entirely
+/// beyond `block`
```
leoyvens: This comment is incorrect due to copy-paste.

store/postgres/src/relational.rs
lutter: Good point. Renaming it.

lutter:
Addressed Leo's comments

lutter:
Rebase to latest master; remove commit that duplicated #1526

lutter:
I just reran this against the most recent production dump, and the migration took 10 minutes; before we deploy this to production, I will remove unsued subgraph deployments, which will help make this faster since it reduces the amount of metadata we have. The fact that the production DB has faster IO than my test instance should also help in speeding this up.

PR #1476, opened 2020-01-31
store: Speed up simple queries by delaying JSONB generation in the database

lutter:


lutter:
> Surprising that the two forms aren't equivalent.

The reason is that `order by` (and `limit`/`offset`) is applied after the `select` clause has been evaluated; in the original form, that caused generation of a lot of JSONB that the `limit` clause would then throw out. In the new form, we reduce the result set in 'native' form before converting to JSONB

This is a really cool illustration of how queries are evaluated: https://jvns.ca/blog/2019/10/03/sql-queries-don-t-start-with-select/

PR #1477, opened 2020-02-01
Use cached block for number -> hash mapping if that is unambiguous

lutter:
One thing I am unsure about with this is whether it will interfere with reverts, since during a revert, we'll have multiple blocks with the same number in the database. If doing the revert requires that the orphaned block remains in the database, this code could cause problems.

### DIFF
--- a/chain/ethereum/src/block_stream.rs
+++ b/chain/ethereum/src/block_stream.rs
@@ -346,6 +346,7 @@ where
                             ctx.eth_adapter.is_on_main_chain(
                                 &ctx.logger,
                                 ctx.metrics.ethrpc_metrics.clone(),
+                                ctx.chain_store.clone(),
                                 ptr,
                             )
                         },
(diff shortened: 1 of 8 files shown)


leoyvens:
Good thing you mentioned it, because now I remember there is an interaction between this and reverts. Here https://github.com/graphprotocol/graph-node/blob/master/chain/ethereum/src/block_stream.rs#L575 we need the reverted block to be in the DB, so we can't de-duplicate blocks within the reorg threshold.

lutter:
> Good thing you mentioned it, because now I remember there is an interaction between this and reverts. Here https://github.com/graphprotocol/graph-node/blob/master/chain/ethereum/src/block_stream.rs#L575 we need the reverted block to be in the DB, so we can't de-duplicate blocks within the reorg threshold.

Thanks for finding that. I'll try to change things so that we send all the information we need with the `ReconciliationStep::RevertBlock`

lutter:
I decided to not try to be too clever here, and simply clean up duplicate blocks only when they have become final. Being more aggressive with cleanups would probably help drive down the number of RPC calls we make further, but since that will interfere with the revert logic, it's better to play it safe and see how much of a drop in RPC calls we get with this approach first.

PR #1482, opened 2020-02-04
store: Lower the threshold for a notification to be large to 128 bytes

lutter:
We are seeing a lot of commits taking 300ms or so in production; when that happens lots of commits pile up and it looks like they are all waiting for a lock on Postgres' internal notification queue. Forcing more notifications to be bounced through the `large_notifications` table should reduce the amount of work Postgres has to do for each notification internally during commit, and the hope is that this will reduce the contention on the notification queue.

### DIFF
--- a/store/postgres/src/notification_listener.rs
+++ b/store/postgres/src/notification_listener.rs
@@ -218,8 +218,10 @@ pub struct JsonNotification {
     pub payload: serde_json::Value,
 }
 
-// Any payload bigger than this is considered large.
-static LARGE_NOTIFICATION_THRESHOLD: usize = 7800;
+// Any payload bigger than this is considered large. Any notification larger
+// than this will be put into the `large_notifications` table, and only
+// its id in the table will be sent via `notify`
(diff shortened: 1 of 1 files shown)


store/postgres/src/notification_listener.rs
```diff
@@ -218,8 +218,10 @@ pub struct JsonNotification {
     pub payload: serde_json::Value,
 }
 
-// Any payload bigger than this is considered large.
-static LARGE_NOTIFICATION_THRESHOLD: usize = 7800;
+// Any payload bigger than this is considered large. Any notification larger
+// than this will be put into the `large_notifications` table, and only
+// its id in the table will be sent via `notify`
+static LARGE_NOTIFICATION_THRESHOLD: usize = 128;
```
olibearo: Is this worth making an env var with the default defined in the code?

lutter: I don't think that's useful here - I have no idea how to tell people what a good value for it would be. If this helps with long commit times, we'll leave it at this, and if not, I might revert it. But either way, it's not really something that can be tuned.

PR #1484, opened 2020-02-06
Add counter to store connection error

ghost:
This is needed for monitoring

### DIFF
--- a/store/postgres/src/connection_pool.rs
+++ b/store/postgres/src/connection_pool.rs
@@ -4,13 +4,22 @@ use diesel::r2d2::{self, ConnectionManager, Pool};
 use graph::prelude::*;
 use graph::util::security::SafeDisplay;
 
+use std::collections::HashMap;
+use std::fmt;
+use std::sync::Arc;
 use std::time::Duration;
 
-#[derive(Debug)]
-struct ErrorHandler(Logger);
+struct ErrorHandler(Logger, Box<Counter>);
+
+impl Debug for ErrorHandler {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "ErrorHandler")
+    }
+}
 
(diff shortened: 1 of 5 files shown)


store/postgres/src/connection_pool.rs
```diff
@@ -4,13 +4,22 @@ use diesel::r2d2::{self, ConnectionManager, Pool};
...
+
+impl Debug for ErrorHandler {
+    fn fmt(&self, _f: &mut fmt::Formatter) -> fmt::Result {
+        fmt::Result::Ok(())
+    }
+}
 
 impl r2d2::HandleError<r2d2::Error> for ErrorHandler {
     fn handle_error(&self, error: r2d2::Error) {
+        self.1.inc();
```
fordN: What does .inc() do?

store/postgres/src/connection_pool.rs
```diff
@@ -19,10 +28,18 @@ pub fn create_connection_pool(
     postgres_url: String,
     pool_size: u32,
     logger: &Logger,
+    registry: Arc<dyn MetricsRegistry>,
```
fordN: I'm a bit rusty on setting up metrics counters; can we just pass the counter into `create_connection_pool()` rather than the entire `MetricsRegistry`?

store/postgres/src/connection_pool.rs
olibearo: It just increments the counter

store/postgres/src/connection_pool.rs
olibearo: I'd rather not create the counter in `main` and keep the counter specific knowledge in the crate/components that needs it

fordN: Yep, sounds good. 
Looking closer now and seeing the pattern of passing in the registry is consistent with the other counters too. 👍

fordN (approved):
LGTM once the tests pass!

PR #1492, opened 2020-02-13
Merged schema

olibearo:
Repointing this [PR](https://github.com/graphprotocol/graph-node/pull/1443) to master.


graph/src/data/subgraph/schema.rs
```diff
@@ -1241,35 +1242,36 @@ pub fn attribute_index_definitions(
...
-                            &document.definitions,
-                        ) {
-                            Ok(value_type) => value_type,
-                            Err(_) => continue,
-                        },
-                        attribute_name: entity_field.name,
-                        entity_name: schema_object.name.clone(),
-                    });
+        if let Definition::TypeDefinition(TypeDefinition::Object(schema_object)) = schema_type {
+            if schema_object.name.eq(SCHEMA_TYPE_NAME) {
```
lutter: This feels like it's going to lead to surprises for anybody who's enumerating document definitions. Can we add an extension trait for `Document` (in `graphql/src/schema/ext.rs`) that defines a function `entity_definitions(&self) -> Vec<Definition>` that can be used to iterate over 'real' definitions? Then the for loop could just be
```rust
   for (entity_number, schema_type) in document.entity_definitions().into_iter().enumerate()
```
and callers don't need to worry about special meta types

graphql/src/schema/ast.rs
```diff
@@ -411,6 +412,23 @@ pub fn get_input_object_definitions(schema: &Document) -> Vec<InputObjectType> {
         .collect()
 }
 
+/// Determines if the field is an imported type
+pub fn is_field_imported(
```
lutter: This would also be really nice to put into the extension trait on `Document` I mentioned above.

graphql/src/schema/ast.rs
```diff
@@ -411,6 +412,23 @@ pub fn get_input_object_definitions(schema: &Document) -> Vec<InputObjectType> {
...
+    subgraph_id: &SubgraphDeploymentId,
+    schema: &Document,
+    field_definition: &Field,
+) -> bool {
+    let field_type_name = get_field_name(&field_definition.field_type);
+    Schema::new(subgraph_id.clone(), schema.clone())
+        .imported_types()
```
lutter: Could `imported_types` also go into the extension trait? That would make it unnecessary to construct a `Schema` just to extract imported types.

graphql/src/schema/ast.rs
```diff
@@ -411,6 +412,23 @@ pub fn get_input_object_definitions(schema: &Document) -> Vec<InputObjectType> {
...
+        .iter()
+        .find(|(imported, _)| match imported {
+            ImportedType::Name(name) => name.eq(&field_type_name),
+            ImportedType::NameAs(_, az) => az.eq(&field_type_name),
```
lutter: If you add a `ImportedType::name(&self) -> &str`, this could just be `&field_type_name == imported.name()` which would make this code easier to read.

store/postgres/src/store.rs
```diff
@@ -118,16 +119,20 @@ pub struct StoreConfig {
 }
 
 /// Commonly needed information about a subgraph that we cache in
-/// `Store.subgraph_cache`. Only immutable subgraph data can be cached this
+/// `Store.schema_cache`. Only immutable subgraph data can be cached this
 /// way as the cache lives for the lifetime of the `Store` object
 #[derive(Clone)]
-struct SubgraphInfo {
+struct SchemaCacheEntry {
```
lutter: Why the rename? Each cache entry is still about a specific subgraph, right?

store/postgres/src/store.rs
```diff
@@ -118,16 +119,20 @@ pub struct StoreConfig {
...
     /// The schema as supplied by the user
     input: Arc<Schema>,
     /// The schema we derive from `input` with `graphql::schema::api::api_schema`
     api: Arc<Schema>,
     /// The name of the network from which the subgraph is syncing
     network: Option<String>,
+    /// The imported schemas which are referenced by name or were not available to merge
+    unstable_imports: Vec<(SchemaReference, Option<SubgraphDeploymentId>)>,
+    /// Timestamp for the last merge
+    last_refresh: Instant,
```
lutter: This requires more explanation; the comment at the top says that these entries are immutable. At the very least, there should be something about what happens in between refreshes when data underneath us changes. What happens when queries run while the cache is out-of-date? Is it a problem that different query nodes could have more or less out-of-date caches?

store/postgres/src/store.rs
```diff
@@ -118,16 +119,20 @@ pub struct StoreConfig {
...
-struct SubgraphInfo {
+struct SchemaCacheEntry {
     /// The schema as supplied by the user
     input: Arc<Schema>,
     /// The schema we derive from `input` with `graphql::schema::api::api_schema`
     api: Arc<Schema>,
     /// The name of the network from which the subgraph is syncing
     network: Option<String>,
+    /// The imported schemas which are referenced by name or were not available to merge
+    unstable_imports: Vec<(SchemaReference, Option<SubgraphDeploymentId>)>,
```
lutter: If there are unavailable schemas, can the subgraph still function? Can it be queried?

store/postgres/src/store.rs
```diff
@@ -712,9 +715,26 @@ impl Store {
...
+                    .unstable_imports
+                    .iter()
+                    .any(|(schema_reference, subgraph_id_opt)| {
+                        self.resolve_schema_reference(schema_reference)
+                            .map(|schema| match subgraph_id_opt {
+                                Some(subgraph_id) => schema.id.eq(subgraph_id),
+                                None => true,
+                            })
+                            .unwrap_or(false)
+                    });
```
lutter: The body of this `any` took me quite a bit to understand. I think what it is saying is "we need to refresh this entry if (a) we have an unstable import that we couldn't resolve last time around or (b) is an unstable import points at a different subgraph; we ignore errors in resolving a reference in all cases" It would be a bit clearer if the `match subgraph_id_opt` happened before the call to `resolve_schema_reference`

Jannis:
Closing for now. We'll revisit composition in the future.

PR #1493, opened 2020-02-14
Fulltext search graphql

ghost:
Implements the subgraph schema validation and api schema generation for fulltext search.

### DIFF
--- a/graph/src/data/schema.rs
+++ b/graph/src/data/schema.rs
@@ -131,7 +131,7 @@ impl TryFrom<&String> for FulltextLanguage {
             "SWEDISH" => Ok(FulltextLanguage::Swedish),
             "TURKISH" => Ok(FulltextLanguage::Turkish),
             invalid => Err(format!(
-                "Provided langauge for full text search is invalid: {}",
+                "Provided langauge for fulltext search is invalid: {}",
                 invalid
             )),
         }
--- a/graph/src/data/schema.rs
+++ b/graph/src/data/schema.rs
@@ -97,7 +97,7 @@ enum FulltextLanguage {
     Dutch,
     English,
     Finnish,
-    Fresh,
+    French,
     German,
     Hungarian,
     Italian,
--- a/graph/src/data/schema.rs
+++ b/graph/src/data/schema.rs
@@ -87,7 +87,7 @@ pub enum SchemaValidationError {
     FulltextAlgorithmInvalid(String),
     #[fail(display = "Fulltext include undefined")]
     FulltextIncludeUndefined,
-    #[fail(display = "Fulltext include invalid")]
+    #[fail(display = "Fulltext include is invalid")]
     FulltextIncludeInvalid,
 }
 
--- a/graph/src/data/schema.rs
+++ b/graph/src/data/schema.rs
@@ -76,7 +76,7 @@ pub enum SchemaValidationError {
     #[fail(display = "Fulltext directive name undefined")]
     FulltextNameUndefined,
     #[fail(display = "Fulltext directive name overlaps with type: {}", _0)]
-    FulltextNameOverlap(String),
+    FulltextNameConflict(String),
     #[fail(display = "Fulltext language is undefined")]
     FulltextLanguageUndefined,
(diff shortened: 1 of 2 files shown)


graph/src/data/schema.rs
```diff
@@ -606,6 +714,192 @@ impl Schema {
...
+            .imported_types()
+            .into_iter()
+            .map(|(imported_type, _)| match imported_type {
+                ImportedType::Name(name) => name.clone(),
+                ImportedType::NameAs(_, az) => az.clone(),
+            })
+            .collect::<Vec<String>>();
+        all_types.append(&mut imported_types);
+
+        if let Some(_) = all_types.iter().find(|typ| typ.eq(&name)) {
```
Jannis: I think this check doesn't really catch conflicts. The fulltext directive name defines the query field that will be generated. The type names _map_ to query fields but they are query fields themselves. So unless I'm missing something, this is comparing 🍎s to 🍊s.

graph/src/data/schema.rs
```diff
@@ -606,6 +714,192 @@ impl Schema {
...
+            .iter()
+            .find(|(key, _)| key.eq("name"))
+            .map(|(_, value)| value);
+        let name = match name_value {
+            Some(Value::String(name)) => name,
+            _ => return vec![SchemaValidationError::FulltextNameUndefined],
+        };
+
+        // Validate that fulltext.name does not overlap with any of the types
+        // TODO: Validate that each fulltext directive has a distinct name
```
Jannis: Mmh, yeah, this is a rather important check that we shouldn't leave out. It could be better placed in the `validate_fulltext_directives` method though.

graph/src/data/schema.rs
```diff
@@ -606,6 +714,192 @@ impl Schema {
...
+
+        // Validate that each entity in fulltext.include exists
+        let include_value = fulltext
+            .arguments
+            .iter()
+            .find(|(key, _)| key.eq("include"))
+            .map(|(_, value)| value);
+        let includes = match include_value {
+            Some(Value::List(includes)) if includes.len() > 0 => includes,
+            _ => return vec![SchemaValidationError::FulltextIncludeUndefined],
```
Jannis: "Undefined" is misleading here as it _is_ may be defined. "invalid" or "not a list" would be more appropriate for cases where there is a value but it's not a list.

graph/src/data/schema.rs
```diff
@@ -606,6 +714,192 @@ impl Schema {
...
+        };
+
+        match includes.iter().find(|include| match include {
```
Jannis: I feel this code can be broken up to be easier to understand, similarly to how `validate_fulltext_directives` delegates to various subroutines for validating parts of each directive.

graphql/src/schema/api.rs
```diff
@@ -989,4 +1145,43 @@ mod tests {
...
+  id: ID!
+  owner: Bytes!
+  displayName: String!
+  imageUrl: String!
+}
+"#;
+        let input_schema = parse_schema(SCHEMA).expect("Failed to parse input schema");
+        let schema = api_schema(&input_schema).expect("Failed to derive API schema");
+
+        dbg!(&schema);
```
Jannis: Drop

graph/src/data/schema.rs
Jannis: That would also make it straight-forward to have different per-include / per-include-field errors.

Jannis (changes requested):

graph/src/data/schema.rs
```diff
@@ -73,6 +73,111 @@ pub enum SchemaValidationError {
...
+    FulltextIncludeInvalid,
+}
+
+enum FulltextLanguage {
+    Simple,
+    Danish,
+    Dutch,
+    English,
+    Finnish,
+    Fresh,
```
fordN: French?

olibearo: Maybe we should support `Fresh` as well?

graph/src/data/schema.rs
olibearo: Ahh totally, removing.

PR #1494, opened 2020-02-14
Do not clone templates for each data source

leoyvens:
This wasted memory.

Testing on a local daostack deployment, that wasn't even fully synced, this dropped memory usage from ~800mb to under 200mb. There may be memory usage problems aside from this but it's clearly an important optimization.


Jannis (approved):
Woah!

PR #1495, opened 2020-02-14
Fix field validation for enums

ghost:
Fix a regression in user defined enums introduced during subgraph composition.


Jannis (approved):
LGTM!

PR #1498, opened 2020-02-20
IPFS selects files from fastest client & lots of async work

That3Percent:
I need to test this still, but if anyone wants to get a head start on reviewing, they can.

I promise I didn't _try_ to refactor everything. `.boxed().compat()` requiring `'static` futures was the main reason that async kept moving up the stack. I think you'll agree that the async code turns out much nicer anyway.

@leoyvens You may be interested in some of the changes here.

### DIFF
--- a/core/src/link_resolver.rs
+++ b/core/src/link_resolver.rs
@@ -1,6 +1,7 @@
 use bytes::BytesMut;
-use futures::{stream::poll_fn, try_ready};
-use ipfs_api;
+use futures01::{stream::poll_fn, try_ready};
+use futures03::stream::FuturesUnordered;
+use ipfs_api::{response::ObjectStatResponse, IpfsClient};
 use lazy_static::lazy_static;
 use lru_time_cache::LruCache;
 use std::env;
@@ -11,25 +12,25 @@ use std::time::Duration;
 use graph::prelude::{LinkResolver as LinkResolverTrait, *};
 use serde_json::Value;
 
-// Environment variable for limiting the `ipfs.map` file size limit.
+/// Environment variable for limiting the `ipfs.map` file size limit.
 const MAX_IPFS_MAP_FILE_SIZE_VAR: &'static str = "GRAPH_MAX_IPFS_MAP_FILE_SIZE";
 
-// The default file size limit for `ipfs.map` is 256MiB.
+/// The default file size limit for `ipfs.map` is 256MiB.
 const DEFAULT_MAX_IPFS_MAP_FILE_SIZE: u64 = 256 * 1024 * 1024;
 
-// Environment variable for limiting the `ipfs.cat` file size limit.
+/// Environment variable for limiting the `ipfs.cat` file size limit.
 const MAX_IPFS_FILE_SIZE_VAR: &'static str = "GRAPH_MAX_IPFS_FILE_BYTES";
 
 lazy_static! {
-    // The default file size limit for the IPFS cahce is 1MiB.
+    /// The default file size limit for the IPFS cache is 1MiB.
     static ref MAX_IPFS_CACHE_FILE_SIZE: u64 = read_u64_from_env("GRAPH_MAX_IPFS_CACHE_FILE_SIZE")
         .unwrap_or(1024 * 1024);
 
-    // The default size limit for the IPFS cache is 50 items.
+    /// The default size limit for the IPFS cache is 50 items.
     static ref MAX_IPFS_CACHE_SIZE: u64 = read_u64_from_env("GRAPH_MAX_IPFS_CACHE_SIZE")
         .unwrap_or(50);
 
-    // The timeout for IPFS requests in seconds
+    /// The timeout for IPFS requests in seconds
     static ref IPFS_TIMEOUT: Duration = Duration::from_secs(
         read_u64_from_env("GRAPH_IPFS_TIMEOUT").unwrap_or(60)
     );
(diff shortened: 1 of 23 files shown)


core/src/subgraph/registrar.rs
```diff
@@ -292,88 +294,60 @@ where
...
-                        chain_store.clone(),
-                        ethereum_adapter.clone(),
-                        name,
-                        manifest,
-                        node_id,
-                        version_switching_mode,
-                    )
-                    .map(|_| (manifest_id, validation_warnings))
-                },
+            .await?;
```
Jannis: The indentation (or lack thereof) here looks odd. Was this formatted with `rustfmt`?

core/src/subgraph/registrar.rs
```diff
@@ -416,61 +387,62 @@ where
         AssignmentEvent::Add {
             subgraph_id,
             node_id: _,
-        } => Box::new(start_subgraph(subgraph_id, &*provider, logger).map_err(|()| unreachable!())),
+        } => {
+            start_subgraph(subgraph_id, provider.clone(), logger).await;
+            Ok(())
```
Jannis: Is the extra `Ok(())` needed or could we just return the `.await` result?

graph/src/data/subgraph/mod.rs
```diff
@@ -568,12 +566,31 @@ pub struct Mapping {
     pub link: Link,
 }
 
+// TODO: This seems to be a part of the stdlib (FuturesOrdered.collect()), but I can't get it working
```
Jannis: Wanna try again together?

graph/src/data/subgraph/mod.rs
```diff
@@ -588,18 +605,17 @@ impl UnresolvedMapping {
...
-        .join(
-            resolver.cat(&logger, &link).and_then(|module_bytes| {
-                Ok(Arc::new(parity_wasm::deserialize_buffer(&module_bytes)?))
-            }),
+        let (abis, runtime) = try_join(
+            // resolve each abi
+            collect_futures(
+                abis.into_iter()
+                    .map(|unresolved_abi| unresolved_abi.resolve(resolver, logger)),
+            ),
```
Jannis: I _may_ be wrong, but we might be able to use unordered futures here to resolve all these files in parallel. I don't think the order of the abis in `Mapping` matters, since they are identified by their names anyway (e.g. when the mapping makes contract calls).

graph/src/data/subgraph/mod.rs
```diff
@@ -614,6 +630,15 @@ impl UnresolvedMapping {
...
 
+async fn cat_module(
+    resolver: &impl LinkResolver,
+    logger: &Logger,
+    link: &Link,
+) -> Result<Arc<Module>, failure::Error> {
+    let module_bytes = resolver.cat(logger, link).await?;
+    Ok(Arc::new(parity_wasm::deserialize_buffer(&module_bytes)?))
+}
+
```
Jannis: Hah, this function has more signature/boilerplate code than actual code. 😁 It's still fine, just made me giggle.

graph/src/data/subgraph/mod.rs
```diff
@@ -660,24 +685,25 @@ impl UnresolvedDataSource {
...
-                templates,
-            })
+
+        let (mapping, templates) = try_join(
+            mapping.resolve(&*resolver, logger),
+            collect_futures(
+                templates
+                    .into_iter()
+                    .map(|template| template.resolve(resolver, logger)),
+            ),
```
Jannis: Here, too, I think we can get away with unordered futures (and hence, a small speed-up).

graph/src/data/subgraph/mod.rs
```diff
@@ -773,15 +799,14 @@ impl UnresolvedDataSourceTemplate {
...
-                mapping,
-            })
+        let mapping = mapping.resolve(resolver, logger).await?;
+        Ok(DataSourceTemplate {
+            kind,
+            network,
+            name,
+            source,
+            mapping,
+        })
```
Jannis: I like to avoid variable assignments where possible. Would this work?
```rust
Ok(DataSourceTemplate {
    kind,
    network,
    name,
    source,
    mapping: mapping.resolve(resolver, logger).await?,
})
```

graph/src/data/subgraph/mod.rs
```diff
@@ -837,12 +862,13 @@ impl UnvalidatedSubgraphManifest {
...
+    pub async fn resolve(
         link: Link,
         resolver: Arc<impl LinkResolver>,
-        logger: Logger,
-    ) -> impl Future<Item = Self, Error = SubgraphManifestResolveError> + Send {
-        SubgraphManifest::resolve(link, resolver, logger).map(|manifest| Self(manifest))
+        logger: &Logger,
+    ) -> Result<Self, SubgraphManifestResolveError> {
+        let manifest = SubgraphManifest::resolve(link, resolver.deref(), logger).await?;
+        Ok(Self(manifest))
```
Jannis: Similarly here, this could also be
```rust
Ok(Self(SubgraphManifest::resolve(link, resolver.deref(), logger).await?))
```

Jannis: Just wanted to point out two examples, there's probably more places like this. Also note this is me being nit-picky, this is not something that I _expect_ to be changed. Up to you what you prefer 😄

graph/src/data/subgraph/mod.rs
```diff
@@ -1029,44 +1054,39 @@ impl UnresolvedSubgraphManifest {
...
+            collect_futures(
+                data_sources
+                    .into_iter()
+                    .map(|ds| ds.resolve(resolver, logger)),
+            ),
+            collect_futures(
+                templates
+                    .into_iter()
+                    .map(|template| template.resolve(resolver, logger)),
+            ),
```
Jannis: Here, also, I believe we don't actually care about the order.

Jannis (changes requested):
Nice! Looking really good. Lovely cleanup of some of the async methods (`create_subgraph_version` for instance) as well! Exactly what was I looking forward to with async/await.

core/src/subgraph/registrar.rs
That3Percent: Yes, it has been formatted.

core/src/subgraph/registrar.rs
```diff
@@ -121,7 +121,13 @@ where
                     })
                     .for_each(move |assignment_event| {
                         assert_eq!(assignment_event.node_id(), &node_id);
-                        handle_assignment_event(assignment_event, provider.clone(), &logger_clone1)
+                        handle_assignment_event(
+                            assignment_event,
+                            provider.clone(),
+                            logger_clone1.clone(),
```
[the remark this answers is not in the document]
That3Percent: Removing the clone here produces an error. This will be fixed when more async-await is implemented.

core/src/subgraph/registrar.rs
That3Percent: start_subgraph doesn't return Result, so an Ok is needed. I moved it up to just wrap the outer value though.

graph/src/data/subgraph/mod.rs
That3Percent: Using `FuturesOrdered` does not restrict the parallelism for retrieving the files. Each file is retrieved concurrently, and `FuturesOrdered` buffers any results that are available but not yet ready to be yielded. So, while there is a performance benefit to choose `FuturesUnordered`, it's not that big.

So, while you're probably right, I'm a bit hesitant to stir the pot more on this change for such a small gain. If we intend to make this an unordered collection I'd prefer it to be a HashSet or something.

core/src/subgraph/registrar.rs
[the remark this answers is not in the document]
Jannis: Ok, np!

core/src/subgraph/registrar.rs
Jannis: Ok, nevermind then. 😉

graph/src/data/subgraph/mod.rs
Jannis: Oh, you're absolutely right, they are still resolved in parallel... this is fine then.

graph/src/data/subgraph/mod.rs
Jannis: Nevermind (see earlier discussion about the same thing).

graph/src/data/subgraph/mod.rs
Jannis: Nvmd

graph/src/data/subgraph/mod.rs
That3Percent: I can try again. I just wasn't sure it was worth delaying the launch of the network since this task is taking longer than expected.

graph/src/data/subgraph/mod.rs
That3Percent: I think I can fix this. I didn't know about `async move` yet when I wrote it.

That3Percent:
@Jannis I pushed a lot of small improvements after this review. I think you'll like how it turned out.

core/src/link_resolver.rs
```diff
@@ -90,25 +87,26 @@ async fn select_fastest_client_with_stat<'a>(
...
-                max_file_bytes,
-                stat.cumulative_size
-            ));
-        }
+/// Returns an error if the stat is bigger than the `MAX_IPFS_MAP_FILE_SIZE_VAR`
+fn restrict_file_size(path: &str, stat: &ObjectStatResponse) -> Result<(), failure::Error> {
+    // FIXME: Having an env variable here seems like a problem for consensus.
+    // Index Nodes should not disagree on whether the file should be read.
+    let max_file_size =
+        read_u64_from_env(MAX_IPFS_MAP_FILE_SIZE_VAR).unwrap_or(DEFAULT_MAX_IPFS_MAP_FILE_SIZE);
```
Jannis: Ah, this unfortunately doesn't work. The file size limits for `ipfs.cat` and `ipfs.map` are deliberately split up into two environment variables, as `ipfs.map` is intended for larger files than `ipfs.cat`.

Plus both env vars are already in use, so we can't simply drop the one for `ipfs.cat`.

Jannis: So I think we still have to pass the file size in.

Jannis (changes requested):
Apart from the comment I left about the IPFS file size limits, I have one more request:

We use the following commit message format fairly consistently:
```
graph, core, store: Make some change

<Commit message body>
```
So we try to mention each crate/path touched in the commit. In some cases we touch (almost) all, in which case we use
```
*: Update dependencies

<body>
```

The goal of this format is to encourage small, atomic commits but also to make it easy to filter commits by what crates they change (e.g. when you're hunting a bug in the store, you can probably ignore all commits that tough the `graphql` directory/crate.

Long story short: could you rebase this branch and update the commit messages accordingly?

Jannis:
Code looks good to me, but CI is currently failing.

That3Percent:
@Jannis The build now passes.

Jannis (approved):
LGTM!

PR #1500, opened 2020-02-21
Update tokio tungstenite

leoyvens:
Our version is really old. This fixes an issue where our 404 responses where instead immediately closing the connection.

Tested locally that subscriptions still work. We should check on staging that this resolves the issue.


Jannis:
@leoyvens I'll merge this PR in your absence. Hope that's ok 😄

PR #1504, opened 2020-02-25
Use monomorphic result type for fulltext searches

fordN:
The initial implementation of fulltext search queries only supports the inclusion of one entity type, so the result type for a fulltext query is the type of the entity included.


olibearo:
We should remove the Union type from the api schema as well then since it won't be used

PR #1505, opened 2020-02-25
store: Only try to cleanup large notifications once per cleanup interval

lutter:
Right now, we try to cleanup large notifications every time one is sent. That is way too often; instead, only do the cleanup once per the configured notification interval.

This change makes sense regardless, but I am also hoping that it will reduce how often we see very slow commits in production.

### DIFF
--- a/store/postgres/src/notification_listener.rs
+++ b/store/postgres/src/notification_listener.rs
@@ -8,9 +8,9 @@ use postgres::{Connection, TlsMode};
 use std::env;
 use std::str::FromStr;
 use std::sync::atomic::{AtomicBool, Ordering};
-use std::sync::{Arc, Barrier};
+use std::sync::{Arc, Barrier, Mutex};
 use std::thread;
-use std::time::Duration;
+use std::time::{Duration, Instant};
 
 use futures::sync::mpsc::{channel, Receiver};
 use graph::prelude::serde_json;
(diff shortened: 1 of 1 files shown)


store/postgres/src/notification_listener.rs
```diff
@@ -303,29 +303,32 @@ impl JsonNotification {
...
-            // We make sure that we do not wait for any locks when we do the
-            // cleanup to avoid stalling this transaction. There is not much
-            // danger in this cleanup sometimes leaving rows behind, as long
-            // as they get cleaned up eventually
-
-            // 'delete' locks the table in 'row exclusive' mode; see that we
-            // can get that lock without waiting
-            let lock =
-                diesel::sql_query("lock table large_notifications in row exclusive mode nowait")
-                    .execute(conn);
```
Jannis: Is it safe to get rid of this Postgres-level lock?

lutter: Yes. When I put that lock there I meant for it to make it so that only one index node at a time would try to clean up. The lock is the same kind that the `delete` statement takes. I didn't read the fine-print well enough though as any number of connections can take a `row exclusive` lock simultaneously, and the lock served no real purpose.

It would be nice to have some sort of singleton worker that can do the cleanup, but doing it no more often than every configured cleanup interval should get us to almost the same place in terms of load/performance.

lutter:
Rebased to latest master

PR #1507, opened 2020-02-27
Kill unresponsive node

leoyvens:
So kubernetes has a chance to restart it. If `GRAPH_ALLOW_UNRESPONSIVE` is set, then this will be disabled.

### DIFF
--- a/docs/environment-variables.md
+++ b/docs/environment-variables.md
@@ -99,3 +99,4 @@ triggers in each request (defaults to 100000).
   [here](https://docs.rs/env_logger/0.6.0/env_logger/)
 - `THEGRAPH_STORE_POSTGRES_DIESEL_URL`: postgres instance used when running
   tests. Set to `postgresql://<DBUSER>:<DBPASSWORD>@<DBHOST>:<DBPORT>/<DBNAME>`
+- `GRAPH_KILL_IF_UNRESPONSIVE`: If set, the process will be killed if unresponsive.
--- a/node/src/main.rs
+++ b/node/src/main.rs
@@ -750,6 +750,10 @@ async fn main() {
                                      "code" => LogCode::TokioContention);
             if timeout < Duration::from_secs(10) {
                 timeout *= 10;
+            } else if std::env::var_os("GRAPH_KILL_IF_UNRESPONSIVE").is_some() {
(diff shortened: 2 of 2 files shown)


node/src/main.rs
```diff
@@ -750,6 +750,10 @@ async fn main() {
                                      "code" => LogCode::TokioContention);
             if timeout < Duration::from_secs(10) {
                 timeout *= 10;
+            } else if std::env::var_os("GRAPH_ALLOW_UNRESPONSIVE").is_none() {
+                // The node is unresponsive, kill it in hopes it will be restarted.
+                crit!(contention_logger, "Node is unresponsive, killing process");
+                std::process::exit(2)
```
Jannis: Shouldn't this be `1` instead of `2`? Exit code `2` stands for "Misuse of shell builtins". 😉

docs/environment-variables.md
```diff
@@ -99,3 +99,4 @@ triggers in each request (defaults to 100000).
   [here](https://docs.rs/env_logger/0.6.0/env_logger/)
 - `THEGRAPH_STORE_POSTGRES_DIESEL_URL`: postgres instance used when running
   tests. Set to `postgresql://<DBUSER>:<DBPASSWORD>@<DBHOST>:<DBPORT>/<DBNAME>`
+- `GRAPH_ALLOW_UNRESPONSIVE`: If set, the process will not be killed if unresponsive.
```
Jannis: I would prefer `GRAPH_KILL_IF_UNRESPONSIVE` and requiring operators to manually set it.

Jannis (changes requested):

node/src/main.rs
leoyvens: Oh didn't know there is a standard for this, we already use `1` for tokio panics. Maybe `3`?

Jannis: `1` is standard for general errors. According to https://stackoverflow.com/a/1535733,
```
#define EX_UNAVAILABLE  69      /* service unavailable */
```
could be an option. But since many scripts only check 0 and 1, I'd stick to 1.

leoyvens: interesting, `1` it is.

leoyvens:
@Jannis I've amended the commit to address your review.

Jannis (approved):

PR #1508, opened 2020-02-28
Use the correct network when resolving a block hash into a block number

lutter:
GraphQL queries that uses time-travel queries with a block hash would sometimes fail with an error of 'no block for this hash found' when that block clearly existed in the database. When we look up a block with a given hash, we look for a block with that hash and the store's network; that is wrong, as for queries the store's network and the subgraph's network might be different. Instead, we now determine the subgraph's network and use that when we check for the hash.

### DIFF
--- a/store/postgres/src/store.rs
+++ b/store/postgres/src/store.rs
@@ -1121,12 +1121,10 @@ impl StoreTrait for Store {
 
     fn block_number(
         &self,
-        _: &SubgraphDeploymentId,
+        subgraph_id: &SubgraphDeploymentId,
         hash: H256,
     ) -> Result<Option<BlockNumber>, StoreError> {
         use crate::db_schema::ethereum_blocks::dsl;
-
-        let hash = format!("{:x}", hash);
 
         // We should also really check that the block with the given hash is
         // on the chain starting at the subgraph's current head. That check is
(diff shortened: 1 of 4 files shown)


store/postgres/src/store.rs
```diff
@@ -1073,47 +1121,59 @@ impl StoreTrait for Store {
...
+                    BlockNumber::try_from(number)
+                        .map_err(|e| StoreError::QueryExecutionError(e.to_string()))
+                } else {
+                    Err(StoreError::QueryExecutionError(format!(
+                        "subgraph {} belongs to network {} but block {:x} belongs to network {}",
+                        subgraph_id,
+                        subgraph_network.unwrap_or("(none)".to_owned()),
+                        hash,
+                        network_name
+                    )))
```
leoyvens: Why bother checking the network? If the hash matches, then the number and all ancestor blocks are the same, and for any network that's not actively trying to mimick another one, the network will also be the same, not that it would matter that the networks are different since they are identical up to this point.

lutter: It guards against mistakenly using a hash from one network against another network.

lutter: To expand on that: if you use a hash that is not from the network in question, you should always get an error. But if that hash happens to be from another network that the node in question is also ingesting, we would resolve the hash to a number (according to the wrong network) and then accept the query. That introduces some reproducibility issues as whether such a query succeeds depends on whether an indexer is also indexing from the other network or not.

leoyvens: Oh I see now, thanks for explaining. The network will be included in the block parameter in the v2 API which means we won't need to infer it from the subgraph, but for now this is a good solution.

lutter:
Rebased to latest master

PR #1510, opened 2020-02-29
Create a coredump on panic

lutter:
Use the [coredump](https://crates.io/crates/coredump) crate so that any panic will cause a core dump (the crate does that by sending `SIGQUIT` to itself)

Also change the behavior when we detect that the node is unresponsive: if `GRAPH_KILL_IF_UNRESPONSIVE` is set, panic so we dump core instead of calling `exit(1)`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -463,6 +463,14 @@ dependencies = [
 ]
 
 [[package]]
+name = "coredump"
+version = "0.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "libc 0.2.66 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
 name = "crossbeam"
 version = "0.2.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


leoyvens:
This will be great, but we're setting a global panic handler that kills the process, yet there are panics which we prefer to catch.

A way to fix this would be to fork coredump (and make a PR) to expose `fn dump_core_and_quit` and `fn enable_core_dumps` as public functions. Then call `enable_core_dumps` on main, and replace our usages of `process::exit` with `dump_core_and_quit`.

lutter:
> This will be great, but we're setting a global panic handler that kills the process, yet there are panics which we prefer to catch.

I didn't know we did that; I was looking for any uses of `catch_unwind` in our code base, and could only find one in `task_spawn` - and there, I don't understand its use. Why do we manually exit the process when that's the default behavior, anyway?

Also, doesn't the default handler for panics only come into play when panics aren't handled otherwise? 

> A way to fix this would be to fork coredump (and make a PR) to expose `fn dump_core_and_quit` and `fn enable_core_dumps` as public functions. Then call `enable_core_dumps` on main, and replace our usages of `process::exit` with `dump_core_and_quit`.

If it turns out that we need to not use a global panic handler, we should only crib the `dump_core` function; I am not a huge fan of everything else that is going on in the crate (changing rlimit, changing current dir) because it tries to override system settings, making it harder to understand what actually happened.

leoyvens:
@lutter tokio treats tasks as threads wrt to panic semantics, so if a task panics tokio will catch it and terminate only the task, that's why we exit the process ourselves.

The panic handler is called before unwinding begins, so it's called regardless of whether the panic is later caught.

Copying only `dump_core` sounds good, we can avoid the `unsafe` by using this safe wrapper https://docs.rs/nix/0.17.0/nix/sys/signal/fn.kill.html.

lutter:
Ah, of course; this becomes even simpler: we can just call `std::process::abort` in those cases then, since the default action for that is also to dump core.

lutter:
Closing this in favor of #1512 which is also much much simpler.

PR #1522, opened 2020-03-03
Spawn websocket connections in the core pool

leoyvens:
Otherwise they would indefinitely take up a thread in the blocking pool, causing that pool to run out of threads under high load and freeze up the node. Now only the execution of the selection set, which is the actually blocking part, is spawned as blocking. Lesson learned, don't put long-running tasks in the blocking pool.

Also bumped the unresponsive timeout from 10s to 100s, since we don't expect it to happen anymore and might want to avoid crashing the node in case of temporary unresponsiveness for some other reason.

Tested locally that the issue no longer reproduces.

Most of the diff if from dropping the lifetime from `Resolver`, it needs to be static now so it can be put in a tokio task.


graph/src/data/query/error.rs
```diff
@@ -57,6 +57,8 @@ pub enum QueryExecutionError {
     UndefinedFragment(String),
     // Using slow and prefetch query resolution yield different results
     IncorrectPrefetchResult { slow: q::Value, prefetch: q::Value },
+    Panic(String),
+    EventStreamError,
```
Jannis: I wonder if these aren't too much of an implementation detail; then again... why not! They might help us debug things in the future.

node/src/main.rs
```diff
@@ -757,7 +757,7 @@ async fn main() {
             debug!(contention_logger, "Possible contention in tokio threadpool";
                                      "timeout_ms" => timeout.as_millis(),
                                      "code" => LogCode::TokioContention);
-            if timeout < Duration::from_secs(10) {
+            if timeout < Duration::from_secs(100) {
```
lutter: We should leave that at 10s - that's a really long time to not make progress for a simple ping/pong test, and increasing it to 100s makes it just more likely that all nodes have frozen, making the whole system unresponsive. Let's just leave that at the old setting until we actually have some false alarms. If anything, I would actually shorten this even more.

graphql/src/subscription/mod.rs
```diff
@@ -228,9 +222,16 @@ where
...
-
-    let result = execute_selection_set(&ctx, selection_set, subscription_type, &None);
+    let subscription_type = sast::get_root_subscription_type(&ctx.schema.document)
+        .unwrap()
+        .clone();
+
+    let result = graph::spawn_blocking_allow_panic(async move {
+        execute_selection_set(&ctx, &selection_set, &subscription_type, &None)
+    })
+    .await
```
lutter: This should helps since now the number of blocking threads we need is no longer dependent on the number of subscriptions, but we can still exhaust the blocking pool because of the thundering herd behavior of subscriptions. If the individual queries for each subscription take a while to run (say 1s which is not that hard to cause) the first few subscriptions through here will exhaust the connection pool, which will cause following subscriptions to wait for those queries to finish, eventually filling up the blocking pool. We should gate spawning the blocking thread on a semaphore that is sized so that we do not exhaust the connection pool (say allows 75% of the connection pool through)

node/src/main.rs
leoyvens: Alright, I've amended the commit to keep it at 10.

graphql/src/subscription/mod.rs
lutter: What we should really do is change the store to move actual work to the blocking pool itself by having some internal function `Store.with_connection(f:Fn(Connection)->Result)` which first acquires a semaphore sized according to the max number of connections in the pool and then executes `f` on the blocking pool. We'd then change all `Store` methods that right now just get a connection to use `with_connection` and do their work inside that.

leoyvens: It does make sense to throttle subscription queries so they don't delay normal queries. Ideally we'd have query nodes dedicated only to subscriptions. I'll put an async semaphore here as you suggest.

graphql/src/subscription/mod.rs
```diff
@@ -2,13 +2,31 @@ use graphql_parser::{query as q, schema as s, Style};
...
 use crate::query::ast as qast;
 use crate::schema::ast as sast;
 
+use lazy_static::lazy_static;
+
+lazy_static! {
+    static ref SUBSCRIPTION_QUERY_SEMAPHORE: Semaphore = {
+        // This is duplicating the logic in main.rs to get the connection pool size, which is
+        // unfourtunate. But because this module has no share state otherwise, it's not simple to
+        // refactor so that the semaphore isn't a global.
```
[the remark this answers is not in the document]
lutter: Added a comment about the 'right' way to do this to #905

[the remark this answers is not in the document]
leoyvens: Thanks for registering this in the issue, merging the PR.

PR #1525, opened 2020-03-04
Fix regression in LinkResolver retry feature

Jannis:
The first commit messages sum it up well:

> Adding support for multiple IPFS clients added a non-retrying `object.stat` into the mix, which unfortunately breaks the retry feature of `LinkResolver`. We added this retry feature to work around intermittent IPFS issues when starting subgraphs (e.g. retry when hitting an IPFS rate limit or temporary DNS error), reducing the chances of subgraphs failing to start at all.

This PR wraps the entire `LinkResolver::cat` body in a retry future. Initially I thought it would be best to simply add a retry future around identifying a suitable IPFS client, so we'd have two separate retries (for finding a client and for retrieving the file). However, that can lead to problems.

The second commit therefore wraps the entire body in a single retry future. From the commit message:

> This makes the retry logic easier to follow and means we are retrying whenever one of the two steps (identify the best IPFS client, retrieve the file) fails. Otherwise we might run into situations where we find an IPFS client but by the time we try to retrieve the file, that client is no longer suitable, e.g. because it no longer has the file.

Tested as follows:

1. Run IPFS, graph-node etc. locally.
2. Deploy an example subgraph locally. It starts indexing fine.
3. Stop graph-node.
4. Run `rm -rf ~/.ipfs && ipfs init && ipfs daemon --offline` to reset IPFS. Now the subgraph files are gone.
5. Start graph-node again.

Before the fix: graph-node would fail starting the subgraph immediately with a `ResolveError` (`merkledag: not found`).

After the fix: graph-node retries resolving subgraph files indefinitely, as expected:
```
Mar 04 15:11:05.471 DEBG Trying again after object.stat + ipfs.cat failed (attempt #50) with result Err(Api(ApiError { message: "merkledag: not found", code: 0 })), subgraph_id: QmcGseoFNsK98KFxeucCHJaRD4zEbqUPM2uieT9jthRr8P, component: SubgraphAssignmentProvider
```

### DIFF
--- a/core/src/link_resolver.rs
+++ b/core/src/link_resolver.rs
@@ -61,37 +61,42 @@ fn read_u64_from_env(name: &str) -> Option<u64> {
 /// make good use of the stat returned.
 async fn select_fastest_client_with_stat<'a>(
     clients: &'a [IpfsClient],
+    logger: &'a Logger,
     path: &'_ str,
     timeout: Duration,
+    do_retry: bool,
 ) -> Result<(ObjectStatResponse, &'a IpfsClient), failure::Error> {
     let mut err: Option<failure::Error> = None;
 
     let mut stats: FuturesUnordered<_> = clients
         .iter()
-        .map(|c| tokio::time::timeout(timeout, c.object_stat(path).map_ok(move |s| (s, c))))
+        .enumerate()
+        .map(|(i, c)| {
+            let retry_fut = if do_retry {
+                retry("object.stat", logger).no_limit()
+            } else {
+                retry("object.stat", logger).limit(1)
+            }
+            .timeout(timeout);
+
+            retry_fut
+                .run(move || c.object_stat(path).map_ok(move |s| (s, i)).boxed().compat())
+                .compat()
+        })
         .collect();
 
     while let Some(result) = stats.next().await {
         match result {
-            // If there is a timeout, use that error if it's the only error.
-            // Prefer other types if error when possible
-            Err(e) => {
-                if err.is_none() {
-                    err = Some(e.into())
-                }
+            Ok((stat, index)) => {
+                return Ok((stat, &clients[index]));
             }
-            Ok(Err(e)) => {
-                err = Some(e.into());
-            }
-            Ok(Ok(value)) => {
-                return Ok(value);
-            }
+            Err(e) => err = Some(e.into()),
         }
     }
 
     Err(err.unwrap_or_else(|| {
         format_err!(
-            "No IPFS clients were supplied to handle the call to object_stat. File: {}",
+            "No IPFS clients were supplied to handle the call to object.stat. File: {}",
             path
         )
     }))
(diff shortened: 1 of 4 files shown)


core/src/link_resolver.rs
```diff
@@ -157,35 +157,36 @@ impl LinkResolverTrait for LinkResolver {
...
+                        trace!(logger, "IPFS cache miss"; "hash" => &path);
+
+                        // `object.stat` fastest
+                        let (stat, client) =
+                            select_fastest_client_with_stat(&self.clients, &path, self.timeout)
+                                .await?;
+
+                        // FIXME: Having an env variable here is a problem for consensus.
+                        // Index Nodes should not disagree on whether the file should be read.
+                        let max_file_size: Option<u64> = read_u64_from_env(MAX_IPFS_FILE_SIZE_VAR);
```
That3Percent: I'm not sure how much of an edge case this is, but it doesn't make sense to retry on errors that are not intermittent. If restrict_file_size fails, this will retry in vain until the timeout is reached. Our retry function doesn't really give you that option though. (And most of our error handling is the anyhow ad-hoc sort where this would be difficult even if the retry function supported it.

I'd like to take a stab at cleaning up more of our async stuff - especially around diesel and retry. But I don't think we'll have time for it.

core/src/link_resolver.rs
```diff
@@ -157,35 +157,36 @@ impl LinkResolverTrait for LinkResolver {
...
             .timeout(self.timeout);
 
-            let result = retry_fut
+            retry_fut
                 .run(move || {
-                    let path = path.clone();
                     async move {
+                        // Pick the client that a) has the file and b) responds to
+                        // Discard the `/ipfs/` prefix (if present) to get the hash.
+                        let path = link.link.trim_start_matches("/ipfs/").to_owned();
```
That3Percent: It looks like you got rid of one clone. Nice. I think we might be able to drop this to_owned() as well.

That3Percent (changes requested):
Just some nitpicks.

That3Percent (approved):

core/src/link_resolver.rs
Jannis: Yeah, this is bad. I've reverted combining everything into a single retry.

core/src/link_resolver.rs
Jannis: I tried but the compiler didn't like it.

runtime/wasm/src/module/test.rs
```diff
@@ -390,10 +390,8 @@ async fn ipfs_map() {
     assert!(errmsg.contains("JSON value is not a string."));
 
     // Bad IPFS hash.
-    let errmsg = run_ipfs_map(BAD_IPFS_HASH.to_string())
-        .unwrap_err()
-        .to_string();
-    assert!(errmsg.contains("api returned error"))
+    let errmsg = dbg!(run_ipfs_map(BAD_IPFS_HASH.to_string()).unwrap_err()).to_string();
+    assert!(dbg!(errmsg).contains("ApiError"));
```
That3Percent: Let's get rid of this `dbg!`

PR #1528, opened 2020-03-05
graph-node: Added StableHash impls for all  Entity update structs

That3Percent:


graph/src/data/store/scalar.rs
```diff
@@ -16,6 +20,51 @@ pub type BigDecimal = bigdecimal::BigDecimal;
...
+    // sequence_number. So, if the &int is negative, when it writes the
+    // is_negative bool it ends up writing that value with a different sequence
+    // number than had we not called next_child. As a result we'll have to
+    // settle with backward compatible changes from any uint type to BigDecimal
+    // and back, but not signed ints.
+    //
+    // My gut feeling is that there is an API that works, and it's somewhere in
+    // the area of having the stable_hash fn yield back some *unused*
+    // SequenceNumber to the caller. Intuitively I think such an API would be
+    // significantly more error prone and will not explore further at this time.
```
Jannis: What's the problem here exactly? Where are we converting between integers (or in what case would we?) and where would we want backwards-compatibility?

graph/src/data/store/scalar.rs
```diff
@@ -16,6 +20,51 @@ pub type BigDecimal = bigdecimal::BigDecimal;
...
+    int: &num_bigint::BigInt,
+    sequence_number: impl SequenceNumber,
+    state: &mut impl StableHasher,
+) {
+    AsInt {
+        is_negative: int.sign() == BigIntSign::Minus,
+        little_endian: &int.to_bytes_le().1,
+    }
+    .stable_hash(sequence_number, state)
+}
```
Jannis: Why not move this into the `stable_hash` implementation below?

And then call `BigInt::stable_hash(&int, sequence_number, state)` in the implementation for `BigDecimal`?

graph/src/components/store.rs
```diff
@@ -41,6 +42,19 @@ pub struct EntityKey {
 
     /// ID of the individual entity.
     pub entity_id: String,
+    // Note: Do not modify fields without making a backward compatible change to
+    // the StableHash impl (below)
```
Jannis: It would be worth describing what a backward-compatible change would look like? Or is this covered in the stable_hash crate docs already?

Jannis (changes requested):

Jannis (changes requested):
Two more quick things:

The commit message subject should be `graph: ...`, since the changes only affect the `graph` directory. And it should be written in present tense, e.g. `graph: Add StableHash impls...`.

And let's rebase this branch on top of master to get rid of that "Merge master into ..." commit.

graph/src/data/store/scalar.rs
That3Percent: Because of the orphan rules.

We can't implement `StableHash` for `bigint::BigInt`, only `data::store::scalar::BigInt`. `big_decimal_stable_hash` only has access to the former and cannot construct a `data::store::scalar::BigInt` without moving out of the `BigDecimal`.

graph/src/components/store.rs
That3Percent: The RFC covers backward compatibility but should be moved into the crate docs when time allows.

graph/src/data/store/scalar.rs
Jannis: Oh, of course. `num_bigint::BigInt` vs. our `sclar::BigInt`. Nevermind.

That3Percent:
@Jannis I moved the comment, rebased, and deleted the comments that you didn't find useful. This PR is dependent on [this other PR](https://github.com/graphprotocol/stable-hash/pull/3) to fix the build so it cannot be merged yet.

Jannis (approved):
Looks good!

Issue #1404, opened 2019-12-12
FEATURE: Parameterized dynamic data source


fubhy:
It would be nice to be able to provide parameters to dynamic data sources that are available in the mappings of the data source.

The use case we have for Melon is this:

When a fund is created, we also pass some addresses from the current `Version` contract of the network. One of these is the `PriceFeed` contract The `PriceFeed` is controlled by an off-chain component that sends an `update` transaction every 24h. This transaction triggers a `PriceUpdate` event that we are currently globally listening to.

The event handler for this event is then used to globally update some values (new prices for the supported assets in the network, etc.). But we also use it to then loops through all funds of the current version and computes a few new performance values for each fund. 

This is problematic because it doesn't scale well and also doesn't work anymore once the `PriceFeed` contract is replaced when a new `Version` is deployed. The loop also only conceptually works because our `Version` contract has a list of all `Funds` that it has deployed. This is a trait of our contract but shouldn't be required for such a setup to work.

I want to be able to generate multiple dynamic data sources for the same eventful target contract (e.g. our `PriceFeed`) and pass additional parameters to it so that during the execution phase of the mapping I can load the `Fund` and re-compute values for it without storing a list of addresses to loop on in our contract.

Jannis:
I'll try to repeat the request in my own words to confirm I understand the use case.

There is one `PriceFeed` contract with a `PriceUpdate` event. Currently, you're handling this in a single, global data source and loop over all `Fund`s to update them after a price update. This only works because there the list of `Fund` contracts is stored in another contract and that allows the `PriceUpdate` handler to loop over them.

What you would like to see is the ability to create data sources from templates with some contextual information. Specifically, in this situation, you would like to create multiple dynamic `PriceFeed` data sources, each with a `Fund` address on its context. Each of these would monitor the same `PriceFeed` contract with the same `PriceUpdate` event handler. In this handler, each of them would then be able to get the `Fund` address from its context and update only this fund.

Instead of looping over all, say, N funds in one handler, there would be N dynamic `PriceFeed` data sources with a different fund address, each responsible only for updating that particular fund.

leoyvens:
An alternative to `dataSource.context()`, which would presumably load all context variables at once, would be an API such as `dataSource.getContext("fund")` to pull only the necessary variables as needed.

ajcronk:
Has this issue been worked on elsewhere? Would you be open to receiving a PR for it?  Or a better route to prioritize?

[...]

PR #1537, opened 2020-03-11
Data source context

leoyvens:
Resolves https://github.com/graphprotocol/graph-node/issues/1404.

Depends on https://github.com/graphprotocol/graph-ts/pull/106 and https://github.com/graphprotocol/graph-cli/pull/446.

See the integration test for usage.

The persistence of the context as metadata was tested manually, since I didn't manage to write a test that would cover that.

Todo:
- Docs.
- Test on a DB with real-world data.

Do we want to continue to commit yarn.lock for tests? It's a large file, and the tests are supposed to work on a fresh build anyways.


graph/src/components/subgraph/instance.rs
```diff
@@ -7,6 +7,7 @@ pub struct DataSourceTemplateInfo {
     pub data_source: String,
     pub template: DataSourceTemplate,
     pub params: Vec<String>,
+    pub context: Option<Entity>,
```
Jannis: Similar to the `Entity` vs. `DataSourceContext` comment in the graph-ts review, I think even for us it'll be confusing to see `Entity` here, say, six months down the road. My suggestion is to add a `DataSourceContext` type here as well.

graph/src/data/subgraph/mod.rs
```diff
@@ -631,13 +636,15 @@ impl From<EthereumContractMappingEntity> for UnresolvedMapping {
...
 
-#[derive(Clone, Debug, Hash, Eq, PartialEq, Deserialize)]
+#[derive(Clone, Debug, Eq, PartialEq, Deserialize)]
 pub struct BaseDataSource<M, T> {
     pub kind: String,
     pub network: Option<String>,
     pub name: String,
     pub source: Source,
     pub mapping: M,
+    pub context: Option<Entity>,
```
Jannis: See the previous comment.

graph/src/data/subgraph/schema.rs
```diff
@@ -572,6 +551,7 @@ pub struct DynamicEthereumContractDataSourceEntity {
     source: EthereumContractSourceEntity,
     mapping: EthereumContractMappingEntity,
     templates: Vec<EthereumContractDataSourceTemplateEntity>,
+    context: Option<Entity>,
```
Jannis: Same here

runtime/wasm/src/host.rs
```diff
@@ -26,6 +26,7 @@ struct RuntimeHostConfig {
     mapping: Mapping,
     data_source_network: String,
     data_source_name: String,
+    data_source_context: Option<Entity>,
```
Jannis: And here :)

runtime/wasm/src/host_exports.rs
```diff
@@ -41,7 +41,8 @@ pub(crate) struct HostExports {
     pub(crate) api_version: Version,
     data_source_name: String,
     data_source_address: Option<Address>,
-    data_source_network: Option<String>,
+    data_source_network: String,
+    data_source_context: Option<Entity>,
```
Jannis: And here

runtime/wasm/src/host_exports.rs
```diff
@@ -64,7 +65,8 @@ impl HostExports {
         api_version: Version,
         data_source_name: String,
         data_source_address: Option<Address>,
-        data_source_network: Option<String>,
+        data_source_network: String,
+        data_source_context: Option<Entity>,
```
Jannis: And here

runtime/wasm/src/host_exports.rs
```diff
@@ -594,6 +597,7 @@ impl HostExports {
         state: &mut BlockState,
         name: String,
         params: Vec<String>,
+        context: Option<Entity>,
```
Jannis: Here!

runtime/wasm/src/module/mod.rs
```diff
@@ -885,6 +888,27 @@ impl WasmiModule {
             &mut self.ctx.state,
             name,
             params,
+            None,
+        )?;
+        Ok(None)
+    }
+
+    /// function createWithContext(name: string, params: Array<string>, context: Entity): void
```
Jannis: Here! (In the comment)

runtime/wasm/src/module/mod.rs
```diff
@@ -903,6 +927,13 @@ impl WasmiModule {
         )))
     }
 
+    /// function dataSource.context(): Entity
```
Jannis: And here (in the comment)!

store/postgres/migrations/2020-03-11-162100_add_data_source_context/up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE subgraphs.dynamic_ethereum_contract_data_source ADD COLUMN context text;
```
Jannis: I assume this will initialize the column for all existing dynamic data sources with `null` or an empty string. Will these data sources still load properly then?

store/postgres/src/subgraphs.graphql
```diff
@@ -54,6 +54,7 @@ type EthereumContractDataSource @entity {
     source: EthereumContractSource!
     mapping: EthereumContractMapping!
     templates: [EthereumContractDataSourceTemplate!]
+    context: String # arbitrary entity serialized to JSON
```
Jannis: `JSON string representing a map with string keys and arbitrary values` perhaps?

Also, let's add this description above the field, like so:
```
"JSON string representing a map with string keys and arbitrary values"
context: String
```
That way it'll show up in the introspection :)

store/postgres/src/subgraphs.graphql
```diff
@@ -67,6 +68,7 @@ type DynamicEthereumContractDataSource @entity {
     ethereumBlockHash: Bytes!
     ethereumBlockNumber: BigInt!
     deployment: SubgraphDeployment!
+    context: String # arbitrary entity serialized to JSON
```
Jannis: Same here

tests/tests/tests.rs
```diff
@@ -51,8 +50,27 @@ fn overloaded_contract_functions() {
...
+            "--node-logs",
             "yarn test",
         ],
         dir,
     );
 }
+
+lazy_static! {
+    // Necessary due to port conflict when running multiple nodes at once.
+    // TODO: Allow tests to run in parallel.
```
Jannis: Shouldn't be that hard to do if e.g. we give `graph test` a `--start-port` and enumerate all ports in the compose file based on that. We can look into that.

store/postgres/src/subgraphs.graphql
leoyvens: I've noticed there's no reason for this field to be on `EthereumContractDataSource`, it should be only in `DynamicEthereumContractDataSource`.

To your point, I should put it as a field description. I made this:
```
"JSON object with user-provided keys and values, can be parsed as an entity"
```
What do you think?

tests/tests/tests.rs
leoyvens: That's a nice idea.

leoyvens:
I've added the `DataSourceContext` alias and addressed all comments.

store/postgres/src/subgraphs.graphql
Jannis: Perhaps `like` an entity instead of `as`?

leoyvens:
I've tested the migration on a real-world DB and it was instant and worked fine. Merge time!

And now to writing docs.

PR #1538, opened 2020-03-11
store: Make sure metadata gets reverted only for one subgraph

lutter:
The way we used to revert metadata when a block was reverted was not
correct: the reversion would also revert metadata for unrelated subgraphs,
leading to potential loss of dynamic data sources.

This code now makes sure we only revert metadata for the subgraph for which
we are reverting a block.

### DIFF
--- a/store/postgres/src/relational.rs
+++ b/store/postgres/src/relational.rs
@@ -18,8 +18,9 @@ use std::sync::Arc;
 use std::time::{Duration, Instant};
 
 use crate::relational_queries::{
-    ClampRangeQuery, ConflictingEntityQuery, DeleteQuery, EntityData, FilterQuery, FindManyQuery,
-    FindQuery, InsertQuery, RevertClampQuery, RevertRemoveQuery, UpdateQuery,
+    ClampRangeQuery, ConflictingEntityQuery, DeleteByPrefixQuery, DeleteDynamicDataSourcesQuery,
+    DeleteQuery, EntityData, FilterQuery, FindManyQuery, FindQuery, InsertQuery, RevertClampQuery,
+    RevertRemoveQuery, UpdateQuery,
 };
 use graph::data::schema::SCHEMA_TYPE_NAME;
 use graph::prelude::{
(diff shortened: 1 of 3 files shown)


store/postgres/src/relational.rs
```diff
@@ -541,6 +542,73 @@ impl Layout {
...
+                operation: EntityChangeOperation::Removed,
+            })
+            .collect();
+
+        if !dds.is_empty() {
+            // Remove subordinate entities for the dynmaic data sources from
+            // the various metadata tables. We do not need to consider the
+            // table for dynamic data sources, not any table whose name starts
+            // with 'Subgraph'. Since the set of metadata entities might change
+            // in the future, it is safer to exclude entity types that we knoe
```
leoyvens: s/knoe/know

store/postgres/src/relational.rs
```diff
@@ -541,6 +542,73 @@ impl Layout {
...
+            .map(|id| EntityChange {
+                subgraph_id: self.subgraph.clone(),
+                entity_type: DDS.to_owned(),
+                entity_id: id.to_owned(),
+                operation: EntityChangeOperation::Removed,
+            })
+            .collect();
+
+        if !dds.is_empty() {
+            // Remove subordinate entities for the dynmaic data sources from
```
leoyvens: s/dynmaic/dynamic

lutter:
> Deleting by prefix and dynamically determining the tables to delete on (be it based on whitelisting or blacklisting) both seem fragile to future changes on the metadata layout. Perhaps ideally we'd have ON DELETE CASCADE constraints but we are not using foreign keys for this. The next best thing imo would be to return the foreign ids in the deleted dynamic data source, and then delete those by the specific id in the specific tables, proceeding recursively if necessary. Don't know if the performance would be better or worse, but performance is not critical here so we should favor robustness.

I agree that this is fragile and depends on assumptions that are hard to enforce, most notably that subordinate entities, and only those, for a dynamic data source have an id that starts with the data source's id. At the same time, coming up with a more robust scheme isn't entirely straightforward: with the current data model, `on delete cascade` doesn't solve it since we have references from the parent to the child; for example, the dynamic data source points at a `EthereumContractSource` which would not disappear with a foreign key constraint.

Recursively deleting subordinate entities is also not straightforward since you need to know to not recurse into e.g. `DynamicEthereumContractDataSource.deployment` If you write the recursion out by hand, I don't feel you win much since you still need to know about this code and that it needs to be fixed up if the data model changes.

Maybe these entities should explicitly mention the data source they belong to and that should be used for deletion? In any event, I agree that something less fragile here would be a good step forward. I just don't know what to write in an issue :(

lutter:
Fixed the typos and added a comment how nobody is really happy with how this metadata gets reverted

leoyvens:
> If you write the recursion out by hand

That's what I was suggesting. This wouldn't delete anything it shouldn't, and we could write it a way that you have to update the code if you add a field.

PR #1539, opened 2020-03-12
-32000 is a code for Alchemy timeout

leoyvens:
We're seeing this a lot on rinkeby requests if the range is above 500, even for empty ranges.


Jannis (approved):

leoyvens:
Snuck in fix for 1.42 warning about deprecated `Error::description` so the build passes.

PR #1541, opened 2020-03-13
Lossy conversion for `bytesToString`

leoyvens:
In the spirit of avoiding fatal errors to subgraphs if at all possible, attempting to parse utf8 will now do lossy conversion instead of failing if invalid utf8 is encountered.


Jannis (approved):
Makes complete sense. We're already doing this in ethabi as well.

Issue #1544, opened 2020-03-18
Deploying a subgraph does not display an error


That3Percent:
This bug is in the latest versions of graph-cli and graph-node.

Deploying commit 068bfb5eb50ca44013e4717840516399e5110711 from [PoCo](https://github.com/iExecBlockchainComputing/PoCo-subgraph) using `yarn deploy-local` reproduces the problem.

The output of the deploy is:
```
Build completed: QmeqFJzXbChLgZnm36oS7HicPbMSkDLrCzE8YvPEKEgyLu

Deployed to http://localhost:8000/subgraphs/name/amxx/iexec-erc1538-goerli/graphql

Subgraph endpoints:
Queries (HTTP):     http://localhost:8000/subgraphs/name/amxx/iexec-erc1538-goerli
Subscriptions (WS): http://localhost:8001/subgraphs/name/amxx/iexec-erc1538-goerli

✨  Done in 4.33s.
```

But, there's an error in the logs:
```
Mar 18 09:44:05.743 ERRO subgraph_deploy failed, params: SubgraphDeployParams { name: SubgraphName("amxx/iexec-erc1538-goerli"), ipfs_hash: SubgraphDeploymentId("QmeqFJzXbChLgZnm36oS7HicPbMSkDLrCzE8YvPEKEgyLu"), node_id: None }, error: ResolveError(ResolveError(InterfaceFieldsMissing("Transfer", "AccountEvent", Strings(["account: Account!"])))), component: JsonRpcServer
```

This error should be displayed in the console.

That3Percent:
Fixed by #1546

PR #1546, opened 2020-03-19
server: Fixed deploy suppressing error

That3Percent:
This PR fixes a regression that caused error messages to be suppressed when deploying a subgraph


That3Percent:
Fixes #1544

Jannis (approved):
Good catch on the `map`/`map_ok`, that's very subtle...

leoyvens:
Oops, thanks for fixing this! You need to `cargo fmt`.

That3Percent:
I just did a significantly more involved refactor, which I don't normally do for a bugfix - but since the old futures api and styles are in some part responsible for the bug this seems warranted.

Jannis (approved):
Looks good!

PR #1547, opened 2020-03-19
Reduce step for trace filter with no address

leoyvens:
We are seeing some of these timeout and 50 should be a good range size. I didn't want to hardcode neither wanted to add another env var, so I made it 1/4 of the regular range size.


Jannis (approved):
LGTM!

PR #1558, opened 2020-03-26
*: Simplify async control flow for block processing

That3Percent:
Refactors block processing to simplify control flow as preparation for adding Proof of Indexing to BlockState.

* No more `and_then()` - “And then” is just “the next line”. No need to pass local variables again and again via closure arguments
* Better performance and readability by reduced boxing, cloning, and dyn futures
* Migrate to futures03 APIs
* Code is generally shorter and follows a straight line
* Migrate towards the use of `CheapClone`

Example:
```rust
		Box::new(future::result(transaction).and_then(|transaction| {
                    stream::iter_ok(matching_hosts).fold(state, move |state, host| {
                        host.process_log(
                            logger.clone(),
                            block.clone(),
                            transaction.clone(),
                            log.clone(),
                            state,
                        )
                    })
                }))
```
Becomes
```
		for host in matching_hosts {
                        state = host
                            .process_log(logger, block, &transaction, &log, state)
                            .await?;
                    }
```

### DIFF
--- a/core/src/subgraph/instance.rs
+++ b/core/src/subgraph/instance.rs
@@ -125,94 +125,71 @@ where
         self.hosts.iter().any(|host| host.matches_log(log))
     }
 
-    fn process_trigger(
-        &self,
-        logger: &Logger,
-        block: Arc<LightEthereumBlock>,
+    fn process_trigger<'a>(
+        &'a self,
+        logger: &'a Logger,
+        block: &'a Arc<LightEthereumBlock>,
         trigger: EthereumTrigger,
         state: BlockState,
-    ) -> Box<dyn Future<Item = BlockState, Error = Error> + Send> {
-        Self::process_trigger_in_runtime_hosts(
-            logger,
-            self.hosts.iter().cloned(),
-            block,
-            trigger,
-            state,
-        )
+    ) -> DynTryFuture<'a, BlockState> {
+        Self::process_trigger_in_runtime_hosts(logger, &self.hosts, block, trigger, state)
     }
 
-    fn process_trigger_in_runtime_hosts(
-        logger: &Logger,
-        hosts: impl Iterator<Item = Arc<T::Host>>,
-        block: Arc<LightEthereumBlock>,
+    fn process_trigger_in_runtime_hosts<'a>(
+        logger: &'a Logger,
+        hosts: &'a [Arc<T::Host>],
+        block: &'a Arc<LightEthereumBlock>,
         trigger: EthereumTrigger,
-        state: BlockState,
-    ) -> Box<dyn Future<Item = BlockState, Error = Error> + Send> {
-        let logger = logger.to_owned();
-        match trigger {
-            EthereumTrigger::Log(log) => {
-                let transaction = block
-                    .transaction_for_log(&log)
-                    .map(Arc::new)
-                    .ok_or_else(|| format_err!("Found no transaction for event"));
-                let matching_hosts: Vec<_> = hosts.filter(|host| host.matches_log(&log)).collect();
-                let log = Arc::new(log);
+        mut state: BlockState,
+    ) -> DynTryFuture<'a, BlockState> {
+        async move {
+            match trigger {
+                EthereumTrigger::Log(log) => {
+                    let log = Arc::new(log);
 
-                // Process the log in each host in the same order the corresponding data
-                // sources appear in the subgraph manifest
-                Box::new(future::result(transaction).and_then(|transaction| {
-                    stream::iter_ok(matching_hosts).fold(state, move |state, host| {
-                        host.process_log(
-                            logger.clone(),
-                            block.clone(),
-                            transaction.clone(),
-                            log.clone(),
-                            state,
-                        )
-                    })
-                }))
-            }
-            EthereumTrigger::Call(call) => {
-                let transaction = block
-                    .transaction_for_call(&call)
-                    .map(Arc::new)
-                    .ok_or_else(|| format_err!("Found no transaction for call"));
-                let matching_hosts: Vec<_> = hosts
-                    .into_iter()
-                    .filter(|host| host.matches_call(&call))
-                    .collect();
-                let call = Arc::new(call);
+                    let transaction = block
+                        .transaction_for_log(&log)
+                        .map(Arc::new)
+                        .ok_or_else(|| format_err!("Found no transaction for event"))?;
+                    let matching_hosts = hosts.iter().filter(|host| host.matches_log(&log));
+                    // Process the log in each host in the same order the corresponding data
+                    // sources appear in the subgraph manifest
+                    let transaction = Arc::new(transaction);
+                    for host in matching_hosts {
+                        state = host
+                            .process_log(logger, block, &transaction, &log, state)
+                            .await?;
+                    }
+                }
+                EthereumTrigger::Call(call) => {
+                    let call = Arc::new(call);
 
-                Box::new(future::result(transaction).and_then(|transaction| {
-                    stream::iter_ok(matching_hosts).fold(state, move |state, host| {
-                        host.process_call(
-                            logger.clone(),
-                            block.clone(),
-                            transaction.clone(),
-                            call.clone(),
-                            state,
-                        )
-                    })
-                }))
-            }
-            EthereumTrigger::Block(ptr, trigger_type) => {
-                let matching_hosts: Vec<_> = hosts
-                    .into_iter()
-                    .filter(|host| host.matches_block(trigger_type.clone(), ptr.number))
-                    .collect();
+                    let transaction = block
+                        .transaction_for_call(&call)
+                        .ok_or_else(|| format_err!("Found no transaction for call"))?;
+                    let transaction = Arc::new(transaction);
+                    let matching_hosts = hosts.iter().filter(|host| host.matches_call(&call));
 
-                Box::new(
-                    stream::iter_ok(matching_hosts).fold(state, move |state, host| {
-                        host.process_block(
-                            logger.clone(),
-                            block.clone(),
-                            trigger_type.clone(),
-                            state,
-                        )
-                    }),
-                )
+                    for host in matching_hosts {
+                        state = host
+                            .process_call(logger, block, &transaction, &call, state)
+                            .await?;
+                    }
+                }
+                EthereumTrigger::Block(ptr, trigger_type) => {
+                    let matching_hosts = hosts
+                        .iter()
+                        .filter(|host| host.matches_block(&trigger_type, ptr.number));
+                    for host in matching_hosts {
+                        state = host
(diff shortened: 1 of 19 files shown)


core/src/subgraph/instance_manager.rs
```diff
@@ -437,175 +436,155 @@ impl EventConsumer<SubgraphAssignmentProviderEvent> for SubgraphInstanceManager
...
+            };
 
-            Err(StreamEnd::Error(CancelableError::Cancel)) => {
-                debug!(
-                    logger_for_err,
-                    "Subgraph block stream shut down cleanly";
-                    "id" => id_for_err.to_string(),
-                );
-                Err(())
+            // Process events from the stream as long as no restart is needed
```
Jannis: Let's move this above the `loop` above.

core/src/subgraph/instance_manager.rs
```diff
@@ -437,175 +436,155 @@ impl EventConsumer<SubgraphAssignmentProviderEvent> for SubgraphInstanceManager
...
+            let start = Instant::now();
 
-                // Set subgraph status to Failed
-                let status_ops =
-                    SubgraphDeploymentEntity::update_failed_operations(&id_for_err, true);
-                if let Err(e) = store_for_err.apply_metadata_operations(status_ops) {
+            let res = process_block(
+                &logger,
+                ctx.inputs.eth_adapter.cheap_clone(),
+                ctx,
```
Jannis: Would it be more idiomatic (and doable?) to take `&mut ctx` now?

core/src/subgraph/instance_manager.rs
```diff
@@ -639,221 +617,195 @@ where
 
     // Process events one after the other, passing in entity operations
     // collected previously to every new event being processed
-    process_triggers(
-        logger.clone(),
+    let (mut ctx, mut block_state) = process_triggers(
+        &logger,
         BlockState::with_cache(std::mem::take(&mut ctx.state.entity_lfu_cache)),
         ctx,
```
Jannis: Hmm, are `&mut ctx` and `&mut block_state` possible here? I see more opportunities for breaking with that functional approach that was necessary before async/await.

core/src/subgraph/instance_manager.rs
```diff
@@ -639,221 +617,195 @@ where
...
-        section.end();
+        // Process the triggers in each host in the same order the
+        // corresponding data sources have been created.
+        for trigger in triggers.into_iter() {
+            block_state = SubgraphInstance::<T>::process_trigger_in_runtime_hosts(
+                &logger,
+                &runtime_hosts,
+                &light_block,
+                trigger,
+                block_state,
```
Jannis: Another candidate for `&mut block_state`.

core/src/subgraph/instance_manager.rs
```diff
@@ -639,221 +617,195 @@ where
...
+    } = block_state
+        .entity_cache
+        .as_modifications(ctx.inputs.store.as_ref())
+        .map_err(|e| {
+            CancelableError::from(format_err!(
+                "Error while processing block stream for a subgraph: {}",
+                e
+            ))
+        })?;
+    section.end();
```
Jannis: Question for @leoyvens: I see that due to the `?`, we may return early from this. Do we have to clean up the stopwatch section in that case or is it ok to return and never call `section.end()`?

There may be a few more places like that; I just want to double-check that the behavior is fine. We were already returning early before, so I'm assuming this was already ok.

graph/Cargo.toml
```diff
@@ -5,7 +5,7 @@ edition = "2018"
 
 [dependencies]
 async-trait = "0.1.25"
-bigdecimal = { version = "0.0.14", features = ["serde"] }
+bigdecimal = { version = "0.1.0", features = ["serde"] }
```
Jannis: This looks like a fairly significant version bump. Is it necessary? I'm asking because there is some risk that they changed things in incompatible ways that would warrant a closer look at the the changes before we update.

graph/src/components/ethereum/types.rs
```diff
@@ -314,8 +314,8 @@ pub struct EthereumTransactionData {
     pub input: Bytes,
 }
 
-impl<'a> From<&'a Transaction> for EthereumTransactionData {
-    fn from(tx: &'a Transaction) -> EthereumTransactionData {
+impl From<&'_ Transaction> for EthereumTransactionData {
+    fn from(tx: &Transaction) -> EthereumTransactionData {
```
Jannis: Oh, cool trick!

graph/src/lib.rs
```diff
@@ -61,7 +61,7 @@ pub mod prelude {
     pub use tokio;
     pub use web3;
 
-    pub type DynTryFuture<'a, Ok, Err = Error> =
+    pub type DynTryFuture<'a, Ok = (), Err = Error> =
```
Jannis: Doesn't this make `DynTryFuture` impossible to use in situations where the result value is not `()`?

runtime/wasm/src/host.rs
```diff
@@ -374,423 +444,258 @@ impl RuntimeHostTrait for RuntimeHost {
...
+                            "Event with the signature \"{}\" not found in \
+                                contract \"{}\" of data source \"{}\"",
+                            event_handler.event,
+                            abi_name,
+                            data_source_name,
+                        )
+                    })?;
+                    Ok((event_handler, event_abi))
+                })
+                .collect::<Result<Vec<_>, failure::Error>>()?;
```
Jannis: Interesting, I didn't know this works with `map`. I thought it would result in a `Vec<Result<..., failure::Error>>` instead. I assume there's an implicit conversion going on from that to `Result<Vec<_>, failure::Error>` using `From`?

runtime/wasm/src/mapping.rs
```diff
@@ -138,6 +139,7 @@ pub(crate) struct MappingContext {
     pub(crate) state: BlockState,
 }
 
+// TODO: This should not be called clone().
```
Jannis: `{clone,copy}_with_empty_block_state()` perhaps? I agree, it isn't _truly_ cloning due to the block state being dropped.

Jannis (changes requested):
Nice! I love it. Consider the `&mut` comments nice to have, I just figured I'd point out additional opportunities I see for simplifying the code.

core/src/subgraph/instance_manager.rs
leoyvens: It's fine to return early, `section.end()` simply drops `section`, and it will also be dropped on return, ending the section.

Jannis: Ok, cool.

core/src/subgraph/instance.rs
```diff
@@ -125,94 +125,70 @@ where
...
-                            logger.clone(),
-                            block.clone(),
-                            trigger_type.clone(),
-                            state,
-                        )
-                    }),
-                )
+        mut state: BlockState,
+    ) -> DynTryFuture<'a, BlockState> {
+        async move {
```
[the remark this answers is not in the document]
That3Percent: Yes/No. I agree that making the function `async` is the best possible. Here is I think what you're intending:

Change the signature from this
```rust
fn process_trigger_in_runtime_hosts<'a>(
        logger: &'a Logger,
        hosts: &'a [Arc<T::Host>],
        block: &'a Arc<LightEthereumBlock>,
        trigger: EthereumTrigger,
        mut state: BlockState,
    ) -> DynTryFuture<'a, BlockState> {
```
to this:
```rust
async fn process_trigger_in_runtime_hosts(
        logger: &Logger,
        hosts: &[Arc<T::Host>],
        block: &Arc<LightEthereumBlock>,
        trigger: EthereumTrigger,
        mut state: BlockState,
    ) -> Result<BlockState> {
```

That would require GATs. Here's why: [Async Methods 1: generic associated types](https://boats.gitlab.io/blog/post/async-methods-i/)

graph/Cargo.toml
That3Percent: Yes, it is necessary. I went through the actual changes in this repository and verified that there are no breaking changes (If I remember there was just a minor new API added).

This resolves an issue with the build where our interop failed to compile since diesel also updated their dependency on bigdecimal.

graph/src/lib.rs
That3Percent: This is just a default. So, you can write `DynTryFuture<'a, u32>` (defaults to Error = failure::Error), `DynTryFuture<'a, u32, MyErrorType>` (explicitly specify everything) or just `DynTryFuture<'a>`.

runtime/wasm/src/host.rs
That3Percent: Yeah, this is something very clever from the Rust team. Once you know the trick, it's obvious that it would be done this way. Unfortunately, it is difficult to discover this API.

The trick is that `Result<T: FromIterator, E>` implements `FromIterator`. (More precise syntax [here](https://doc.rust-lang.org/std/result/enum.Result.html#impl-FromIterator%3CResult%3CA%2C%20E%3E%3E)). It does the thing you would hope for and expect by iterating trying to collect each item unless it encounters the first error than bailing. This is very powerful because you can use it on any iterator (not just `map`) and for any other `FromIterator` type - One could collect into a `Result<HashMap>` for example.

runtime/wasm/src/mapping.rs
That3Percent: Oops, I meant to go back and rename this. The problem is that with the kind of work that I'm going it's desirable to go around and remove calls to `clone()` that were only put there to work around lifetime problems. It's a useful heuristic to assume that converting a `clone()` to a `&` does not change any runtime behavior (besides performance).

I'll go rename this now.

That3Percent:
> Nice! I love it. Consider the `&mut` comments nice to have, I just figured I'd point out additional opportunities I see for simplifying the code.

I went ahead and fixed the minor issues and pushed a commit, but am doing a second pass for some of the more major things we discussed earlier (in particular blocking futures around mapping requests). I'll keep them in separate commits for now to make the review smoother then we can squash later.

graph/Cargo.toml
leoyvens: Ah we also updated diesel. I was suprised to see this bump not cause incompatibility.

core/src/subgraph/instance.rs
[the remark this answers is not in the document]
Jannis:
```rust
async fn process_trigger_in_runtime_hosts(
        logger: &Logger,
        hosts: &[Arc<T::Host>],
        block: &Arc<LightEthereumBlock>,
        trigger: EthereumTrigger,
        mut state: BlockState,
    ) -> Result<()> {
```
was my thinking (modify the `state` in place).

Jannis (approved):
Looks good to me! As discussed, let's postpone the next round of improvements.

PR #1560, opened 2020-03-28
Moar Async

Jannis:
While I was going through our code in preparation of multi-blockchain planning, I thought I'd take the opportunity to convert more code to async. This converts the following traits (and implementations) to use async/await more (primarily in the trait methods):

- `DataSourceLoader`
- `LinkResolver`
- `RuntimeHost`
- `SubgraphAssignmentProvider`
- `SubgraphRegistrar`
- `SubgraphInstance`

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -4,6 +4,7 @@ version = "0.17.1"
 edition = "2018"
 
 [dependencies]
+async-trait = "0.1.25"
 bytes = "0.5"
 futures01 = { package="futures", version="0.1.29" }
 futures = { version="0.3.4", features=["compat"] }
--- a/core/src/subgraph/loader.rs
+++ b/core/src/subgraph/loader.rs
@@ -2,6 +2,8 @@ use std::collections::HashMap;
 use std::iter::FromIterator;
 use std::time::Instant;
 
+use async_trait::async_trait;
+
 use graph::data::subgraph::schema::*;
 use graph::data::subgraph::UnresolvedDataSource;
 use graph::prelude::{DataSourceLoader as DataSourceLoaderTrait, GraphQlRunner, *};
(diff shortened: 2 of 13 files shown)


core/Cargo.toml
```diff
@@ -4,6 +4,7 @@ version = "0.17.1"
 edition = "2018"
 
 [dependencies]
+async-trait = "0.1.25"
```
That3Percent: Great find

core/src/subgraph/loader.rs
```diff
@@ -183,89 +185,62 @@ where
...
+            if next_data_sources.is_empty() {
+                finished = true;
+            } else {
+                data_sources.extend(next_data_sources);
+                skip += data_sources.len() as i32;
+            }
+        }
 
-                let self2 = self1.clone();
-                let self3 = self1.clone();
```
That3Percent: So much better!

core/src/subgraph/provider.rs
```diff
@@ -71,16 +73,17 @@ where
...
 }
 
+#[async_trait]
 impl<L, Q, S> SubgraphAssignmentProviderTrait for SubgraphAssignmentProvider<L, Q, S>
 where
     L: LinkResolver + Clone,
     Q: GraphQlRunner,
     S: Store + SubgraphDeploymentStore,
 {
-    fn start<'a>(
```
That3Percent: Love this.

core/src/subgraph/registrar.rs
```diff
@@ -275,102 +277,85 @@ where
...
+#[async_trait]
 impl<L, P, S, CS> SubgraphRegistrarTrait for SubgraphRegistrar<L, P, S, CS>
 where
     L: LinkResolver,
     P: SubgraphAssignmentProviderTrait,
     S: Store + SubgraphDeploymentStore,
     CS: ChainStore,
 {
-    fn create_subgraph(
+    async fn create_subgraph(
```
That3Percent: It seems that at the moment this does not need to be async. Let's keep it this way though since eventually store will be async.

That3Percent (approved):
💯

PR #1571, opened 2020-03-31
Moar Async II

Jannis:
This time converting `LinkResolver` and `SubgraphInstance` to async traits.


That3Percent (approved):
Straightforward

Issue #1577, opened 2020-04-02
json.fromBytes should not crash subgraph in any case.


orenyodfat:
Possible solutions:

-  providing try /catch pattern on it 
-  return error code or null value if error ...
at https://github.com/graphprotocol/graph-node/blob/4b46753dc5c539da5dd0b0d6c179f5ea683b0fa3/runtime/wasm/src/host_exports.rs#L348

Jannis:
Closed via #1578 and graphprotocol/graph-ts#110.

PR #1578, opened 2020-04-02
Add `json.try_fromBytes` host export for handling JSON parsing errors

Jannis:
Part of resolving #1577.

This adds a `json.try_fromBytes` host export that returns a`Result<serde_json::Value, bool>` back to the mapping, using the new `Result` type added to graph-ts in https://github.com/graphprotocol/graph-ts/pull/110.

This allows mappings to _try_ parsing JSON and gracefully handle errors instead of failing. Right now, subgraphs always fail when failing to parse JSON data, which is a problem, as there is no way to control that IPFS files anchored on chain are always valid.


runtime/wasm/wasm_test/json_parsing.ts
```diff
@@ -0,0 +1,59 @@
...
+  kind: JSONValueKind;
+  data: JSONValuePayload;
+
+  isNull(): boolean {
+    return this.kind == JSONValueKind.NULL;
+  }
+
+  toString(): string {
+    assert(this.kind == JSONValueKind.STRING, "JSON value is not a string.");
+    return changetype<string>(this.data as u32);
```
That3Percent, line 36: Can you explain the cast from u64 to u32?

That3Percent (approved):
I just have a quick question. LGTM though so assuming the question doesn't trigger any red flags for you go ahead and merge.

runtime/wasm/wasm_test/json_parsing.ts
leoyvens: This is copy pasted from graph-ts, where we represent the payload of enums as an u64, in case we need to pass an actual u64, but mostly we just pass pointers around, which are 32 bit in WASM.

runtime/wasm/src/module/mod.rs
```diff
@@ -544,15 +544,15 @@ impl WasmiModule {
         Ok(Some(RuntimeValue::from(i)))
     }
 
-    /// function json.fromBytes(bytes: Bytes): JSONValue
-    fn json_from_bytes(
+    /// function json.try_fromBytes(bytes: Bytes): Result<JSONValue, JSONError>
+    fn json_try_from_bytes(
```
That3Percent, line 565: Don't we need both functions for backward compatibility?

That3Percent (changes requested):
Quick question here, but maybe a serious one.

That3Percent (approved):
Nevermind the previous (deleted) question... reviewing changes incrementally showed a different story. Approved!

PR #1587, opened 2020-04-06
Rework and optimize Elasticsearch logging

Jannis:
This PR does several things:

- It moves all key/value pairs passed to log messages into the `arguments` object sent to Elasticsearch and removes them from the actual message text. This makes it possible to efficiently search for fields that were previously only included as text.

- It renames the `meta` field to `source`. This felt logical because this really is about where in the source code a message comes from.

- It optimizes the logs flushing to send logs in batches of up to 1000 messages at time. Instead of cloning the pending logs every time, they are now moved from the queue into the processing thread. Cloning is reduced overall. This should make logging both more efficient and less memory hungry.

- It simplifies the key/value pair serialization logic.

- It uses the `CheapClone` trait where possible.

_Note: This will require changes in the explorer backend to merge all log `arguments` back into the message. It may also require changes to Grafana._

### DIFF
--- a/graph/src/log/elastic.rs
+++ b/graph/src/log/elastic.rs
@@ -1,6 +1,5 @@
-use std::collections::HashMap;
+use std::collections::{BTreeMap, HashMap};
 use std::fmt;
-use std::fmt::Write;
 use std::result::Result;
 use std::sync::{Arc, Mutex};
 use std::time::Duration;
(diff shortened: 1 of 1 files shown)


azf20:
@Jannis is this PR still valid?

otaviopace:
I'll close this for now, @Jannis feel free to re-open if necessary 🙂

PR #1590, opened 2020-04-07
Add integration test for handling Ganache reverts

Jannis:
This tests that `try_` calls made against Ganache correctly catch reverts. The test makes two calls (one reverting, the other succeeding) and writes the reverted flag and return values into a `Call` entity. This entity is then queried after the subgraph has synced.

Run the test with:
```sh
cargo test ganache_reverts
```
from the root directory of this repo.

### DIFF
--- a/tests/integration-tests/ganache-reverts/package.json
+++ b/tests/integration-tests/ganache-reverts/package.json
@@ -1,0 +1,25 @@
+{
+  "name": "ganache-reverts",
+  "version": "0.1.0",
+  "scripts": {
+    "build-contracts": "rm -rf abis bin && solcjs contracts/Contract.sol --abi -o abis && mv abis/*Contract.abi abis/Contract.abi && solcjs contracts/Contract.sol --bin -o bin && mv bin/*Contract.bin bin/Contract.bin",
+    "codegen": "graph codegen",
+    "test": "yarn build-contracts && truffle test --network test",
+    "create:test": "graph create test/ganache-reverts --node http://localhost:18020/",
+    "deploy:test": "graph deploy test/ganache-reverts --ipfs http://localhost:15001/ --node http://localhost:18020/"
+  },
+  "devDependencies": {
+    "@graphprotocol/graph-cli": "https://github.com/graphprotocol/graph-cli#master",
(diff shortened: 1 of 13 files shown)


tests/integration-tests/ganache-reverts/package.json
```diff
@@ -0,0 +1,25 @@
...
+  "version": "0.1.0",
+  "scripts": {
+    "build-contracts": "rm -rf abis bin && solcjs contracts/Contract.sol --abi -o abis && mv abis/*Contract.abi abis/Contract.abi && solcjs contracts/Contract.sol --bin -o bin && mv bin/*Contract.bin bin/Contract.bin",
+    "codegen": "graph codegen",
+    "test": "yarn build-contracts && truffle test --network test",
+    "create:test": "graph create test/ganache-reverts --node http://localhost:18020/",
+    "deploy:test": "graph deploy test/ganache-reverts --ipfs http://localhost:15001/ --node http://localhost:18020/"
+  },
+  "devDependencies": {
+    "@graphprotocol/graph-cli": "https://github.com/graphprotocol/graph-cli#master",
```
orenyodfat, line 12: can you try that with graph-cli 0.17.0 and graph-ts 0.17.1

Jannis: As commented offline, `graph test` was only added after graph-cli 0.17.1, so no.

tests/integration-tests/overloaded-contract-functions/package.json
```diff
@@ -9,7 +9,7 @@
     "deploy:test": "graph deploy test/overloaded-contract-functions --ipfs http://localhost:15001/ --node http://localhost:18020/"
   },
   "devDependencies": {
-    "@graphprotocol/graph-cli": "https://github.com/graphprotocol/graph-cli#master",
+    "@graphprotocol/graph-cli": "https://github.com/graphprotocol/graph-cli#jannis/graph-test-ipfs-0.4.23",
     "@graphprotocol/graph-ts": "https://github.com/graphprotocol/graph-ts#master",
```
orenyodfat, line 13: can you do that also for graph-ts ?

leoyvens (approved):

PR #1596, opened 2020-04-10
Fix dynamic data source loading

leoyvens:
The previous logic was incorrect for subgraphs with over 200 dynamic data sources.


lutter (approved):
LGTM

Jannis (approved):

PR #1605, opened 2020-04-14
Proof of Indexing

That3Percent:
### Proof of Indexing
This change adds much of the needed framework around the Proof of Indexing digest update and serving. Some of the database work is left for future work in order to integrate what changes have been marked. These are clearly marked.

Changes were made to the network RFC in order to work around limitations of GraphQL. In particular, interfaces cannot be used as inputs.

### Other new APIs
new `object!` macro - removes noise/boilerplate for graphql objects (`IntoValue` added to support this). Lib code has been migrated to use this, There are more than 100 places within tests to migrate later.
`Store::with_entity_connection` - Talk to the database without violating the futures API, allowing timeout, quicker transaction cancellation on future drop, and greater levels of concurrency
`TryFrom<BigInt> for u64` - fallible conversion with custom error
`trait CancelToken` - convenient `token.check_cancel()?` API.
More `CheapClone` impls
`spawn_blocking_async_allow_panic` to eventually replace futures based version

### Fixes
Call `resolve_scalar_value` when outside of an object for GraphQL
No 6 hour timeout for db connection in test means you get an error message when the connection fails

### And more
Various added documentation, migrating to async, and other cleanup

### DIFF
--- a/core/src/subgraph/instance_manager.rs
+++ b/core/src/subgraph/instance_manager.rs
@@ -6,8 +6,9 @@ use std::time::Instant;
 
 use graph::components::ethereum::triggers_in_block;
 use graph::components::store::ModificationsAndCache;
+use graph::components::subgraph::{ProofOfIndexing, ProofOfIndexingDigest};
 use graph::data::subgraph::schema::{
-    DynamicEthereumContractDataSourceEntity, SubgraphDeploymentEntity,
+    DynamicEthereumContractDataSourceEntity, SubgraphDeploymentEntity, POI_OBJECT,
 };
 use graph::prelude::{SubgraphInstance as SubgraphInstanceTrait, *};
 use graph::util::lfu_cache::LfuCache;
@@ -626,6 +627,7 @@ where
     .await?;
 
     // If new data sources have been created, restart the subgraph after this block.
+    // This is necessary to re-create the block stream.
     let needs_restart = !block_state.created_data_sources.is_empty();
     let host_metrics = ctx.host_metrics.clone();
 
@@ -657,7 +659,6 @@ where
             EthereumBlockFilter::from_data_sources(data_sources.iter()),
             block.clone(),
         )
-        .compat()
         .await?;
 
         let triggers = block_with_triggers.triggers;
@@ -687,6 +688,7 @@ where
 
         // Process the triggers in each host in the same order the
         // corresponding data sources have been created.
+
         for trigger in triggers.into_iter() {
             block_state = SubgraphInstance::<T>::process_trigger_in_runtime_hosts(
                 &logger,
@@ -705,6 +707,15 @@ where
     if block_stream_cancel_handle.is_canceled() {
         return Err(CancelableError::Cancel);
     }
+
+    update_proof_of_indexing(
+        &mut block_state.proof_of_indexing,
+        &ctx.host_metrics.stopwatch,
+        ctx.inputs.store.as_ref(),
+        &ctx.inputs.deployment_id,
+        &mut block_state.entity_cache,
+    )
+    .await?;
 
     let section = ctx.host_metrics.stopwatch.start_section("as_modifications");
     let ModificationsAndCache {
@@ -764,6 +775,63 @@ where
             Err(format_err!("Error while processing block stream for a subgraph: {}", e).into())
         }
     }
+}
+
+/// Transform the proof of indexing changes into entity updates that will be
+/// inserted when as_modifications is called.
+async fn update_proof_of_indexing(
+    proof_of_indexing: &mut ProofOfIndexing,
+    stopwatch: &StopwatchMetrics,
+    store: &(impl Store + SubgraphDeploymentStore),
+    deployment_id: &SubgraphDeploymentId,
+    entity_cache: &mut EntityCache,
+) -> Result<(), Error> {
+    // Need to take this out whether or not we hit the early return. Otherwise it accumulates
+    let mut proof_of_indexing = if let Some(proof_of_indexing) = proof_of_indexing.take() {
+        proof_of_indexing
+    } else {
+        return Ok(());
+    };
+
+    let _section_guard = stopwatch.start_section("update_proof_of_indexing");
+
+    // Check if the PoI table actually exists. Do not update PoI unless the subgraph
+    // was deployed after the feature was implemented
+    if !store.supports_proof_of_indexing(&deployment_id).await? {
+        return Ok(());
+    }
+
+    for (causality_region, stream) in proof_of_indexing.drain() {
+        // Create the special POI entity key specific to this causality_region
+        let entity_key = EntityKey {
+            subgraph_id: deployment_id.clone(),
+            entity_type: POI_OBJECT.to_owned(),
+            entity_id: causality_region,
+        };
+
+        // Grab the current digest attribute on this entity
+        let prev_poi = entity_cache
+            .get(store, &entity_key)
+            .map_err(Error::from)?
+            .map(|entity| match entity.get("digest") {
+                Some(Value::String(s)) => ProofOfIndexingDigest(s.clone()),
+                _ => panic!("Expected POI entity to have a digest and for it to be a string"),
+            });
+
+        // Finish the POI stream, getting the new POI value.
+        let ProofOfIndexingDigest(updated_proof_of_indexing) = stream.finish(&prev_poi);
+
+        // Put this onto an entity with the same digest attribute
+        // that was expected before when reading.
+        let new_poi_entity = entity! {
+            id: entity_key.entity_id.clone(),
+            digest: updated_proof_of_indexing,
+        };
+
+        entity_cache.set(entity_key, new_poi_entity);
+    }
+
+    Ok(())
 }
 
 async fn process_triggers<B: BlockStreamBuilder, T: RuntimeHostBuilder, S: Send + Sync>(
--- a/graph/src/components/store.rs
+++ b/graph/src/components/store.rs
@@ -15,6 +15,7 @@ use std::sync::Arc;
 use std::time::{Duration, Instant};
 use web3::types::H256;
 
+use crate::components::subgraph::ProofOfIndexingDigest;
 use crate::data::store::*;
 use crate::data::subgraph::schema::*;
 use crate::prelude::*;
@@ -27,8 +28,8 @@ lazy_static! {
             .map(|s| u64::from_str(&s).unwrap_or_else(|_| panic!(
                 "failed to parse env var SUBSCRIPTION_THROTTLE_INTERVAL"
             )))
-            .map(|millis| Duration::from_millis(millis))
-            .unwrap_or(Duration::from_millis(1000));
+            .map(Duration::from_millis)
+            .unwrap_or_else(|| Duration::from_millis(1000));
 }
 
 // Note: Do not modify fields without making a backward compatible change to
@@ -129,7 +130,7 @@ pub enum EntityOrder {
 
 impl EntityOrder {
     /// Return `"asc"` or `"desc"` as is used in SQL
-    pub fn to_sql(&self) -> &'static str {
+    pub fn to_sql(self) -> &'static str {
         match self {
             EntityOrder::Ascending => "asc",
             EntityOrder::Descending => "desc",
@@ -800,7 +801,6 @@ pub enum TransactionAbortError {
 }
 
 /// Common trait for store implementations.
-#[automock]
 pub trait Store: Send + Sync + 'static {
     /// Get a pointer to the most recently processed block in the subgraph.
     fn block_ptr(
@@ -808,15 +808,31 @@ pub trait Store: Send + Sync + 'static {
         subgraph_id: SubgraphDeploymentId,
     ) -> Result<Option<EthereumBlockPointer>, Error>;
 
+    fn supports_proof_of_indexing<'a>(
+        &'a self,
+        subgraph_id: &'a SubgraphDeploymentId,
+    ) -> DynTryFuture<'a, bool>;
+
+    /// A value of None indicates that the table is not available. Re-deploying
+    /// the subgraph fixes this. It is undesirable to force everything to
+    /// re-sync from scratch, so existing deployments will continue without a
+    /// Proof of Indexing. Once all subgraphs have been re-deployed the Option
+    /// can be removed.
+    fn get_proof_of_indexing<'a>(
+        &'a self,
+        subgraph_id: &'a SubgraphDeploymentId,
+        block_number: u64,
+    ) -> DynTryFuture<'a, Option<ProofOfIndexingDigest>>;
+
     /// Looks up an entity using the given store key at the latest block.
     fn get(&self, key: EntityKey) -> Result<Option<Entity>, QueryExecutionError>;
 
     /// Look up multiple entities as of the latest block. Returns a map of
     /// entities by type.
-    fn get_many<'a>(
+    fn get_many(
         &self,
         subgraph_id: &SubgraphDeploymentId,
-        ids_for_type: BTreeMap<&'a str, Vec<&'a str>>,
+        ids_for_type: BTreeMap<&str, Vec<&str>>,
     ) -> Result<BTreeMap<String, Vec<Entity>>, StoreError>;
 
     /// Queries the store for entities that match the store query.
@@ -1193,6 +1209,137 @@ pub trait Store: Send + Sync + 'static {
         subgraph_id: &SubgraphDeploymentId,
         block_hash: H256,
     ) -> Result<Option<BlockNumber>, StoreError>;
+}
+
+mock! {
+    pub Store {
+        fn get_many_mock<'a>(
+            &self,
+            _subgraph_id: &SubgraphDeploymentId,
+            _ids_for_type: BTreeMap<&'a str, Vec<&'a str>>,
+        ) -> Result<BTreeMap<String, Vec<Entity>>, StoreError>;
+    }
+}
+
+// The store trait must be implemented manually because mockall does not support async_trait, nor borrowing from arguments.
+impl Store for MockStore {
+    fn block_ptr(
+        &self,
+        _subgraph_id: SubgraphDeploymentId,
+    ) -> Result<Option<EthereumBlockPointer>, Error> {
+        unimplemented!();
+    }
+
+    fn supports_proof_of_indexing<'a>(
+        &'a self,
+        _subgraph_id: &'a SubgraphDeploymentId,
+    ) -> DynTryFuture<'a, bool> {
+        unimplemented!();
+    }
+
+    fn get_proof_of_indexing<'a>(
+        &'a self,
+        _subgraph_id: &'a SubgraphDeploymentId,
+        _block_number: u64,
+    ) -> DynTryFuture<'a, Option<ProofOfIndexingDigest>> {
+        unimplemented!();
+    }
+
+    fn get(&self, _key: EntityKey) -> Result<Option<Entity>, QueryExecutionError> {
+        unimplemented!()
+    }
+
+    fn get_many(
+        &self,
+        subgraph_id: &SubgraphDeploymentId,
+        ids_for_type: BTreeMap<&str, Vec<&str>>,
+    ) -> Result<BTreeMap<String, Vec<Entity>>, StoreError> {
+        self.get_many_mock(subgraph_id, ids_for_type)
+    }
+
+    fn find(&self, _query: EntityQuery) -> Result<Vec<Entity>, QueryExecutionError> {
+        unimplemented!()
+    }
+
+    fn find_one(&self, _query: EntityQuery) -> Result<Option<Entity>, QueryExecutionError> {
+        unimplemented!()
+    }
+
+    fn find_ens_name(&self, _hash: &str) -> Result<Option<String>, QueryExecutionError> {
+        unimplemented!()
+    }
+
+    fn transact_block_operations(
+        &self,
+        _subgraph_id: SubgraphDeploymentId,
+        _block_ptr_to: EthereumBlockPointer,
+        _mods: Vec<EntityModification>,
+        _stopwatch: StopwatchMetrics,
+    ) -> Result<bool, StoreError> {
+        unimplemented!()
+    }
+
+    fn apply_metadata_operations(
+        &self,
+        _operations: Vec<MetadataOperation>,
+    ) -> Result<(), StoreError> {
+        unimplemented!()
+    }
+
+    fn build_entity_attribute_indexes(
+        &self,
+        _subgraph: &SubgraphDeploymentId,
+        _indexes: Vec<AttributeIndexDefinition>,
+    ) -> Result<(), SubgraphAssignmentProviderError> {
+        unimplemented!()
+    }
+
+    fn revert_block_operations(
+        &self,
+        _subgraph_id: SubgraphDeploymentId,
+        _block_ptr_from: EthereumBlockPointer,
+        _block_ptr_to: EthereumBlockPointer,
+    ) -> Result<(), StoreError> {
+        unimplemented!()
+    }
+
+    fn subscribe(&self, _entities: Vec<SubgraphEntityPair>) -> StoreEventStreamBox {
+        unimplemented!()
+    }
+
+    fn create_subgraph_deployment(
+        &self,
+        _schema: &Schema,
+        _ops: Vec<MetadataOperation>,
+    ) -> Result<(), StoreError> {
+        unimplemented!()
+    }
+
+    fn start_subgraph_deployment(
+        &self,
+        _logger: &Logger,
+        _subgraph_id: &SubgraphDeploymentId,
+        _ops: Vec<MetadataOperation>,
+    ) -> Result<(), StoreError> {
+        unimplemented!()
+    }
+
+    fn migrate_subgraph_deployment(
+        &self,
+        _logger: &Logger,
+        _subgraph_id: &SubgraphDeploymentId,
+        _block_ptr: &EthereumBlockPointer,
+    ) {
+        unimplemented!()
+    }
+
+    fn block_number(
+        &self,
+        _subgraph_id: &SubgraphDeploymentId,
+        _block_hash: H256,
+    ) -> Result<Option<BlockNumber>, StoreError> {
+        unimplemented!()
+    }
 }
 
 #[automock]
--- a/graph/src/components/subgraph/proof_of_indexing.rs
+++ b/graph/src/components/subgraph/proof_of_indexing.rs
@@ -1,0 +1,142 @@
+use crate::prelude::Value;
+use stable_hash::{prelude::*, utils::StableHasherWrapper, SequenceNumberInt};
+use std::collections::HashMap;
+use std::fmt;
+use strum::AsStaticRef as _;
+use strum_macros::AsStaticStr;
+use twox_hash::XxHash64;
+
+#[derive(Debug)]
+pub struct ProofOfIndexingDigest(pub String);
+
+impl StableHash for ProofOfIndexingDigest {
+    fn stable_hash(&self, sequence_number: impl SequenceNumber, state: &mut impl StableHasher) {
+        self.0.stable_hash(sequence_number, state)
+    }
+}
+
+#[derive(AsStaticStr)]
+pub enum ProofOfIndexingEvent<'a> {
+    RemoveEntity {
+        entity_type: &'a str,
+        id: &'a str,
+    },
+    SetEntity {
+        entity_type: &'a str,
+        id: &'a str,
+        data: &'a HashMap<String, Value>,
+    },
+}
+
+impl StableHash for ProofOfIndexingEvent<'_> {
+    fn stable_hash(&self, mut sequence_number: impl SequenceNumber, state: &mut impl StableHasher) {
+        use ProofOfIndexingEvent::*;
+        match self {
+            RemoveEntity { entity_type, id } => {
+                entity_type.stable_hash(sequence_number.next_child(), state);
+                id.stable_hash(sequence_number.next_child(), state);
+            }
+            SetEntity {
+                entity_type,
+                id,
+                data,
+            } => {
+                entity_type.stable_hash(sequence_number.next_child(), state);
+                id.stable_hash(sequence_number.next_child(), state);
+                data.stable_hash(sequence_number.next_child(), state);
+            }
+        }
+        // Include the discriminant
+        self.as_static().stable_hash(sequence_number, state);
+    }
+}
+
+/// The POI is the StableHash of:
+/// (Vec<ProofOfIndexingEvent>, PreviousDigest)
+/// This struct contains the necessary state to construct that value in a streaming manner
+pub struct ProofOfIndexingStream {
+    previous_digest_sequence_number: SequenceNumberInt<u64>,
+    vec_sequence_number: SequenceNumberInt<u64>,
+    vec_length: usize,
+    digest: StableHasherWrapper<XxHash64>,
+}
+
+impl ProofOfIndexingStream {
+    fn new() -> Self {
+        let mut tuple_sequence_number = SequenceNumberInt::<u64>::root();
+        let vec_sequence_number = tuple_sequence_number.next_child();
+        let previous_digest_sequence_number = tuple_sequence_number.next_child();
+        Self {
+            previous_digest_sequence_number,
+            vec_sequence_number,
+            vec_length: 0,
+            digest: Default::default(),
+        }
+    }
+
+    fn write(&mut self, event: &ProofOfIndexingEvent) {
+        event.stable_hash(self.vec_sequence_number.next_child(), &mut self.digest);
+        self.vec_length += 1;
+    }
+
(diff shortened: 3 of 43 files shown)


store/postgres/src/relational.rs
```diff
@@ -283,6 +283,30 @@ impl Layout {
...
+        writeln!(
+            out,
+            "CREATE TABLE {} (
+                id serial primary key,
+                data_source varchar not null,
+                block_number int not null,
+                artifact bigint not null
+            );",
+            proof_of_indexing_name,
+        )?;
```
lutter, line 303: Not a review, just some thoughts since last we talked about this:

(1) A POI artifact is likely valid for a block range (from the block where the last change happened to just before the block with the next change) That means artifacts should store a `block_range`, and makes them really just another entity type. We should just inject the appropriate `Table` definition into the `Layout`'s list of tables and then manage them with the usual entity methods. 

(2) We should hold off on any marking of artifacts by `data_source` since we don't need it yet. For now, it's enough to have one series of artifacts since each subgraph has a single source of time. We should wait with adding a `data_source` or similar field until we actually implement something where a subgraph has more than one source of time.

That3Percent: @lutter: 
Using the entity API turned out [nicely](https://github.com/graphprotocol/graph-node/blob/df2dc6b9bc31cb3246c4d2661e25ea1a0496b20f/core/src/subgraph/instance_manager.rs#L824):

 Is this what you had in mind for determining [whether the POI table exists](https://github.com/graphprotocol/graph-node/blob/df2dc6b9bc31cb3246c4d2661e25ea1a0496b20f/core/src/subgraph/instance_manager.rs#L811)? I'm unclear as to whether adding this to the `Schema` for the side effect of having the tables created and be queryable is the right approach, but didn't see any straightforward alternatives. Where would you put the code that edits the `Schema` to include the additional type definition?

lutter: > @lutter:
> Using the entity API turned out [nicely](https://github.com/graphprotocol/graph-node/blob/df2dc6b9bc31cb3246c4d2661e25ea1a0496b20f/core/src/subgraph/instance_manager.rs#L824):

Nice! You can't use `$poi` as the entity type though: Postgres names can contain a `$`, but can't start with it [(docs)](https://www.postgresql.org/docs/12/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS).

> Is this what you had in mind for determining [whether the POI table exists](https://github.com/graphprotocol/graph-node/blob/df2dc6b9bc31cb3246c4d2661e25ea1a0496b20f/core/src/subgraph/instance_manager.rs#L811)? I'm unclear as to whether adding this to the `Schema` for the side effect of having the tables created and be queryable is the right approach, but didn't see any straightforward alternatives. Where would you put the code that edits the `Schema` to include the additional type definition?

What I had in mind there was a check against the actual database schema; externally, that should be a method `Store.supports_poi(&subgraph_id)` which is backed by the same method on `entities::Connection` which in turn always returns `false` for `JsonStorage` and for `Storage::Relational` queries [`information_schema.tables`](https://www.postgresql.org/docs/9.6/infoschema-tables.html) to see if an entry with  `table_name = <name of POI table> and table_schema = <layout.schema>` exists. The result of that query should be cached in the `Layout` and so should probably be passed into `Layout::new` (and `struct Layout` gets a new attribute `supports_poi`)

That3Percent:
This is now fully operational.

store/postgres/src/store.rs
```diff
@@ -665,17 +704,116 @@ impl Store {
         Ok(())
     }
 
+    /// Panics:
+    /// * This task will panic if the supplied closure panics
+    /// * This task will panic if the supplied closure returns Err(Cancelled)
+    ///   when the supplied cancel token is not cancelled.
+    // Ideally, we want to pass a borrow of the PooledConnection in. But,
+    // with_entity_conn requires this to be owned to give it to e::Connection
+    // If e::Connection could take a reference this would not be a problem
```
lutter, line 713: What would it take to make it so that we can pass a reference of `PooledConnection` in?

store/postgres/src/store.rs
```diff
@@ -665,17 +704,116 @@ impl Store {
...
+    async fn with_conn<T: Send + 'static>(
```
lutter, line 714: This function and `with_entity_conn` is fundamental enough that it deserves documentation on how to use it properly, in particular, for which operations to use it and for which ones not to use it and instead fall back to `Store::get_conn`.

Your uses call `cancel_handle.check_cancel()?` as the first thing - is that necessary seeing how `with_conn` calls that right before calling `f`. As a user of this, when should I be calling  `cancel_handle.check_cancel()?` ?

store/postgres/src/store.rs
```diff
@@ -665,17 +704,116 @@ impl Store {
...
+
     fn get_conn(&self) -> Result<PooledConnection<ConnectionManager<PgConnection>>, Error> {
         let start_time = Instant::now();
         let conn = self.conn.get();
         let wait = start_time.elapsed();
         if wait > Duration::from_millis(10) {
-            warn!(self.logger, "Possible contention in DB connection pool";
-                               "wait_ms" => wait.as_millis())
+            warn!(&self.logger, "Possible contention in DB connection pool";
+                                "wait_ms" => wait.as_millis())
```
lutter, line 810: I'd be hugely in favor of just getting rid of this log statement. I don't think we get much utility out of it anymore.

store/postgres/src/store.rs
```diff
@@ -854,6 +993,69 @@ impl StoreTrait for Store {
...
+            cancel.check_cancel()?;
+
+            if !conn.supports_proof_of_indexing() {
+                return Ok(None);
+            }
+
+            let entities = conn
+                .query(
+                    &logger,
+                    EntityCollection::All(vec!["poi_artifact".to_owned()]),
```
lutter, line 1024: Using a string constant here would be very good (and this one is actually the entity type, i.e., `Table.object`)

store/postgres/src/relational.rs
```diff
@@ -170,8 +168,43 @@ impl Layout {
...
+                        ))),
+                        column_type: ColumnType::String,
+                        fulltext_fields: None,
+                    },
+                ],
+                /// The position of this table in all the tables for this layout; this
+                /// is really only needed for the tests to make the names of indexes
+                /// predictable
+                position: tables.len() as u32,
+            };
```
lutter, line 202: How fun, a manually constructed `Table` structure :laughing:

runtime/wasm/src/host_exports.rs
```diff
@@ -129,13 +130,30 @@ impl HostExports {
         )))
     }
 
+    /// Some data sources have indeterminism or different notions of time. These
+    /// need to be each be stored separately to separate causality between them,
+    /// and merge the results later. Right now, this is just ethereum but it
+    /// will be expanded for ipfs and the availability chain.
+    fn poi_namespace(&self) -> &str {
+        "ethereum"
```
lutter, line 138: Strictly speaking, this should be the concrete network, not just the family. For now, it doesn't matter since we don't ever have subgraphs that straddle networks, so that for a mainnet subgraph `ethereum` means `mainnet` and for a Rinkeby subgraph it means `rinkeby`, but it would be cleaner to actually use the network.

graphql/src/values/mod.rs
```diff
@@ -1,15 +1,77 @@
...
+    (String, String),
+    (f64, Float),
+    (bool, Boolean),
+    (Vec<Value>, List),
+    (Number, Int)
+];
+
+/// Creates a `graphql_parser::query::Value::Object` from key/value pairs.
+#[macro_export]
+macro_rules! object {
```
lutter, line 63: Neat! That's a nice tool to tighten up some of our code. At least for our tests, would also be nice to have an `entity!` macro along the same lines.

lutter (changes requested):

runtime/wasm/src/host_exports.rs
Jannis: Yeah, agreed

server/index-node/src/resolver.rs
```diff
@@ -34,19 +35,41 @@ impl TryFromValue for DeploymentAssignment {
     }
 }
 
+#[derive(Debug)]
+struct ProofOfIndexingInput {
+    // The id of the subgraph.
+    subgraph_name: String,
```
Jannis, line 41: This should really be the ID (i.e. the `Qm...` hash of the subgraph deployment), not the subgraph name. The name is a moving target, whereas the ID always points to the same subgraph data.

server/index-node/src/resolver.rs
```diff
@@ -34,19 +35,41 @@ impl TryFromValue for DeploymentAssignment {
...
+    // The block number in ethereum to query against. Eventually when the data
+    // availability chain comes online and we support other block chains then
+    // this will be come an optional field.
+    ethereum_block_number: BigInt,
```
Jannis, line 45: IMHO this should be just `block` or `block_number` (with the comment reflecting that it's the block number on the chain that the subgraph is indexing, which could be an Ethereum chain). The reason I'd like to get rid of the `ethereum` here is that this will work just as well for other blockchains.

store/postgres/src/store.rs
That3Percent: `entities::Connection`` would need to have a lifetime on it. This seems like the right thing to do. Let me do a quick dive and see how much lift it would require.

server/index-node/src/resolver.rs
```diff
@@ -470,17 +469,133 @@ where
...
+        let poi = match futures::executor::block_on(poi_fut) {
+            Ok(Some(poi)) => q::Value::String(poi.0),
+            Ok(None) => q::Value::Null,
+            Err(e) => {
+                error!(
+                    self.logger,
+                    "Failed to query proof of indexing";
+                    "subgraph" => subgraph_name,
+                    "error" => format!("{:?}", e)
+                );
```
Jannis, line 592: Wouldn't we want to return the error to the client?

server/index-node/src/schema.graphql
```diff
@@ -7,6 +7,12 @@ scalar String
 type Query {
   indexingStatusesForSubgraphName(subgraphName: String!): [SubgraphIndexingStatus!]!
   indexingStatuses(subgraphs: [String!]): [SubgraphIndexingStatus!]!
+  proofOfIndexing(request: ProofOfIndexingInput!): Bytes
```
Jannis, line 10: In the interest of future-proofing, should this perhaps return an array of proof-of-indexing "slices"/"units"/"segments"... (whatever we call them)? Or would we combine them on the node before returning the result?

server/index-node/src/schema.graphql
```diff
@@ -7,6 +7,12 @@ scalar String
...
+}
+
+input ProofOfIndexingInput {
+  subgraphName: String!
+  ethereumBlockNumber: BigInt!
```
Jannis, line 15: block` or `blockNumber`

server/index-node/src/schema.graphql
```diff
@@ -7,6 +7,12 @@ scalar String
 type Query {
   indexingStatusesForSubgraphName(subgraphName: String!): [SubgraphIndexingStatus!]!
   indexingStatuses(subgraphs: [String!]): [SubgraphIndexingStatus!]!
+  proofOfIndexing(request: ProofOfIndexingInput!): Bytes
+}
+
+input ProofOfIndexingInput {
+  subgraphName: String!
```
Jannis, line 14: `subgraph` is fine, but it should be the ID of the subgraph deployment, not the name of a subgraph.

server/index-node/src/resolver.rs
```diff
@@ -470,17 +469,133 @@ where
...
+
+        let mut subgraphs = subgraphs
+            .get_values::<q::Value>()
+            .expect("invalid subgraph values");
+
+        let subgraph = if let Some(subgraph) = subgraphs.pop() {
+            subgraph
+        } else {
+            return Ok(q::Value::Null);
+        };
```
Jannis, line 560: Most of this we can get rid of when we query by subgraph ID.

store/postgres/src/entities.rs
```diff
@@ -823,6 +832,28 @@ fn find_schema(
...
+    }
+    let query =
+        "SELECT table_name FROM information_schema.tables WHERE table_schema=$1 AND table_name=$2";
+    let result: Vec<Table> = diesel::sql_query(query)
+        .bind::<Text, _>(schema)
+        .bind::<Text, _>("poi_artifact")
+        .load(conn)?;
+    Ok(result.len() > 0)
+}
+
```
Jannis, line 856: I wonder if that should go into the `metadata` module that @lutter is preparing in the grafting PR?

store/postgres/src/entities.rs
```diff
@@ -1291,11 +1322,13 @@ impl Storage {
             }
             V::Relational => {
                 let subgraph_schema = store.input_schema(subgraph)?;
+                let create_poi = supports_proof_of_indexing(conn, subgraph, &schema.name)?;
```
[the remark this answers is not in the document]
Jannis: `supports_poi` sounds good to me.

store/postgres/src/relational.rs
```diff
@@ -170,8 +168,43 @@ impl Layout {
         let mut tables = Vec::new();
         let mut enums = EnumMap::new();
 
-        for defn in &document.definitions {
-            match defn {
+        if create_proof_of_indexing {
+            let poi_table = Table {
+                object: "poi_artifact".to_owned(),
```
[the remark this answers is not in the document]
Jannis: This type (and, more importantly, the generated table) must not collide with subgraph-defined entities. I therefor suggest we use `_PoiArtifact` (or `_POIArtifact`) for the entity type name and `_poi_artifact` for the table name.

Jannis (changes requested):
I'll leave the discussion around `with_conn` to you and @lutter.

This is looking pretty good already!

store/postgres/src/store.rs
That3Percent: You're right. I'll add copious amounts of documentation around `with_conn` and `with_entity_conn` that encompasses a lot of the earlier discussions.

I don't think there is any case where one should fall back on `Store::get_conn`. The one place where it might make seem to make sense _right now_ is where you need multiple connections simultaneously (eg: a normal connection and an entity connection, or multiple entity connections) when you're already inside `with_conn`. But, there is a better way to deal with that for a slightly different API. Instead of using `with_entity_conn` as it exists today, simply have `with_conn` return some kind of wrapper type that has a method `to_entity_conn<'a>(&'a self) -> e::Connection<'a>` which would allow for re-use of the same connection multiple times in the same `with_conn` closure.

I wish I had time to port the rest of the database code and refine the API, but I think this is going to have to wait.

store/postgres/src/relational.rs
That3Percent: Was there a better way to do this? I didn't want to create a schema type and pass that to `Table::new`.

graphql/src/values/mod.rs
That3Percent: I'd be happy to create one. Can you point me to one test that would use this macro so I know what you're looking for?

store/postgres/src/store.rs
That3Percent: Screw it. Maybe I'll go ahead and make that API since now having the good API doesn't take long, it's not inferring that everything needs to be migrated.

That3Percent: Discussed offline, will not make these changes yet.

core/src/subgraph/instance_manager.rs
```diff
@@ -767,6 +778,62 @@ where
...
+    store: &(impl Store + SubgraphDeploymentStore),
+    deployment_id: &SubgraphDeploymentId,
+    entity_cache: &mut EntityCache,
+) -> Result<(), Error> {
+    // Need to take this out whether or not we hit the early return. Otherwise it accumulates
+    let mut proof_of_indexing = if let Some(proof_of_indexing) = proof_of_indexing.take() {
+        proof_of_indexing
+    } else {
+        return Ok(());
+    };
```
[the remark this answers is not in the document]
That3Percent: It takes any poi changes out of the block state to be converted to entities.

core/src/subgraph/instance_manager.rs
```diff
@@ -767,6 +778,62 @@ where
...
+        return Ok(());
+    }
+
+    for (namespace, stream) in proof_of_indexing.drain() {
+        // Create the special POI entity key specific to this namespace
+        let entity_key = EntityKey {
+            subgraph_id: deployment_id.clone(),
+            entity_type: "poi_artifact".to_owned(),
+            entity_id: namespace,
+        };
```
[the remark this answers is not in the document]
That3Percent: I went down this road per your suggestion but the `TypedEntity`s aren't per-subgraph, they have the `SUBGRAPHS_ID` hardcoded in. That could be changed but it seems like more trouble than it's worth just to not construct an `EntityKey` here.

graphql/src/values/mod.rs
That3Percent: Nevermind, I think you mean just the search for `Entity::new`

graph/src/components/store.rs
```diff
@@ -1194,6 +1210,136 @@ pub trait Store: Send + Sync + 'static {
...
+    pub Store {
+        fn get_many_mock<'a>(
+            &self,
+            _subgraph_id: &SubgraphDeploymentId,
+            _ids_for_type: BTreeMap<&'a str, Vec<&'a str>>,
+        ) -> Result<BTreeMap<String, Vec<Entity>>, StoreError>;
+    }
+}
+
+// The store trait must be implemented manually because mockall does not support async_trait, nor borrowing from arguments.
```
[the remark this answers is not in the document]
That3Percent: `DynTryFuture<'a, T>` is enough to create a problem for mockall.

graph/src/components/subgraph/proof_of_indexing.rs
```diff
@@ -0,0 +1,139 @@
...
+        }
+        // Include the discriminant
+        self.as_static().stable_hash(sequence_number, state);
+    }
+}
+
+/// The POI is the StableHash of:
+/// (Vec<ProofOfIndexingEvent, PreviousArtifact>)
+/// This struct contains the necessary state to construct that value in a streaming manner
+pub struct ProofOfIndexingStream {
```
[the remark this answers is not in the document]
That3Percent: The term digest refers to the output of a hash function. So, I'm renaming all cases of "artifact" to "digest" at the moment. For consistency, I think I'll call this `ProofOfIndexingHashFn`.

graph/src/components/subgraph/proof_of_indexing.rs
```diff
@@ -0,0 +1,139 @@
...
+use stable_hash::{prelude::*, utils::StableHasherWrapper, SequenceNumberInt};
+use std::collections::HashMap;
+use std::fmt;
+use strum::AsStaticRef as _;
+use strum_macros::AsStaticStr;
+use twox_hash::XxHash64;
+
+/// An intermediate value of a ProofOfIndexing specific to the events emitted by a single DataSource
+#[derive(Debug)]
+pub struct ProofOfIndexingArtifact(pub String);
```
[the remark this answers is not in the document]
That3Percent: The comment above this was outdated. This is now going to be `ProofOfIndexingDigest` - but it refers to both intermediate digests and final digests, which don't really need to be disambiguated by the type system.

graph/src/components/subgraph/proof_of_indexing.rs
```diff
@@ -0,0 +1,139 @@
...
+    }
+}
+
+#[derive(Default)]
+pub struct ProofOfIndexing {
+    /// The POI is updated for each data source independently. This is necessary because
+    /// some data sources (eg: IPFS files) may be unreliable and therefore cannot mix
+    /// state with other data sources. This may also give us some freedom to change
+    /// the order of triggers in the future.
+    per_namespace: HashMap<String, ProofOfIndexingStream>,
```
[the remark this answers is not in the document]
That3Percent: I've renamed these to "causality_region"

graph/src/components/subgraph/proof_of_indexing.rs
```diff
@@ -0,0 +1,139 @@
...
+        let previous_artifact_sequence_number = tuple_sequence_number.next_child();
+        Self {
+            previous_artifact_sequence_number,
+            vec_sequence_number,
+            vec_length: 0,
+            digest: Default::default(),
+        }
+    }
+
+    fn emit(&mut self, event: &ProofOfIndexingEvent) {
```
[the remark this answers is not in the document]
That3Percent: I've renamed this to "write"

graph/src/ext/futures.rs
```diff
@@ -97,8 +97,30 @@ pub struct CancelHandle {
...
+pub struct NeverCancel;
+
+impl CancelToken for NeverCancel {
+    #[inline]
+    fn is_canceled(&self) -> bool {
+        false
+    }
+}
+
+pub struct Canceled;
```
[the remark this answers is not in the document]
That3Percent: `CancelableError` is a generic type, so that would cause `check_cancel` to also be generic, but then type inference wouldn't know what type `t.check_cancel()?;` should return - `CancelableError<???>`. This way fixes that issue, but also allows `Canceled` to coerce to other types. For example you could sometimes just have `Result<T, Canceled>` if the operation was otherwise infallible.

server/index-node/src/resolver.rs
That3Percent: We had started out with an API in the RFC that was something like `blocks: [Block!]!` where `Block` was an interface and there was `EthereumBlock` etc. That doesn't work, because GraphQL doesn't (yet?) support interfaces in inputs. We still need to be able to support blocks from multiple blockchains in the same query - especially with the availability chain coming.

So, there's a couple of APIs possible. One, you could keep in the spirit of the per/network blocks without the interface, but then it's easy to specify more than one block with the same network. Eg: `[{Block {network: "eth", number: 10}, Block {network: "eth", number: 20}}]` It's unclear what to return (probably an error, even though the data is not malformed). With the current API, this isn't a problem and has the same amount of API churn as the original `EthereumBlock` API.

I'm not really happy with either API really. Maybe we can defer until we actually have multiple chains?

server/index-node/src/resolver.rs
That3Percent: I would have thought so, but this is the pattern that other resolvers in the file are following.

server/index-node/src/schema.graphql
That3Percent: This already combines them, which for The Network at least is already the desired outcome. If we want to output intermediate values for integration tests or something we could create a new API for that. I think we should leave that as a future improvement for now.

server/index-node/src/resolver.rs
That3Percent: I changed it to subgraph id, but unless I'm misunderstanding all this code stays. We still need to go from a subgraph id to a deployment id.

Jannis: Subgraph ID = deployment ID 😉

store/postgres/src/relational.rs
lutter: No, that's perfectly fine and probably the best you can do here w/o huge overhead

store/postgres/src/entities.rs
lutter: Yeah, that would be a good place for it (this query could also be written using Diesel directly, but I don't want @That3Percent have to go through the whole `table!` etc. dance I can fix that up once this is committed. We should just remember to do that.

graphql/src/values/mod.rs
lutter: It's actually pretty easy since `graph::data::store::Value` already implements `From` for a lot of cases:
```rust
macro_rules! entity {
    ($($name:ident: $value:expr,)*) => {
        {
            let mut result = ::graph::prelude::Entity::new();
            $(
                result.insert(stringify!($name).to_string(), Value::from($value));
            )*
            result
        }
    };
    ($($name:ident: $value:expr),*) => {
        entity! {$($name: $value,)*}
    };
}
```

That3Percent:
@Jannis @lutter I think I've addressed all the comments, though some of them I'm not sure there was a satisfactory answer for.

Here is a summary of the latest changes:

- Query POI by subgraph ID, not name
- `with_conn` closure gets a reference to the PgConnection
- Named constants for duplicated string values
- Renamed `create_poi` to `has_poi`
- Renamed "artifact" to "digest"
- Add lots of documentation to `with_conn` and `with_entity_conn`
- Removed logs that are no longer providing value
- Renamed "namespace" to "causality_region"
- Include Ethereum network in causality region
- Figured out how to use GraphQL illegal names in PoI table names
- Added `entity!` and migrated a small handful of places in the code to use it
- Fixed unfinished comment
- Renamed `emit` to `write`

I think everything that wasn't fixed has some comment or question in the discussion here on GitHub.

store/postgres/src/store.rs
```diff
@@ -854,6 +1019,69 @@ impl StoreTrait for Store {
...
+                        _ => Err(format_err!("Entity has non-string digest attribute")),
+                    }?;
+
+                    let poi = ProofOfIndexingDigest(digest);
+
+                    Ok((causality_region, poi))
+                })
+                .collect::<Result<HashMap<_, _>, Error>>()?;
+
+            let hash = stable_hash_with_hasher::<twox_hash::XxHash64, _>(&by_causality_region);
```
Jannis, line 1078: I checked stable-hash to understand how `HashMap` (by definition unordered) hashes deterministically. I still don't understand it. Could you once more explain how that's guaranteed? I see there are tests for `HashMap` in stable-hash, so this is just for me to understand. 😉

Jannis (changes requested):

store/postgres/src/store.rs
That3Percent: The properties we _must_ demonstrate are:

- Determinism: Any value which `PartialEq(a, b) == true` outputs `StableHash(a) == StableHash(b)`.
- Avalanche: Small changes in the input drastically change the output "seemingly randomly"
- Collision resistance: It is not easy to have 2 inputs that hash to the same value.

The algorithm for an unordered collection of *unique* (<- very important to be unique) items is as follows:
```
1: Initialize some binary value (rollup) to all 0 (or some predefined value), and count to 0
2: For each element in the collection:
    a: StableHash it independently of the supplied hash state
    b: XOR the intermediate digest from step a with the rollup
    c: increment count
3: Feed rollup into the StableHash
4: Feed the count into the StableHash
```

### Determinism
We'll take as a given that the StableHash of each element in the sequence has a deterministic stable hash.
Consider the rollup value as an array of bits. Each XOR which operates on the rollup bits operates on each bit independently. Eg: rollup_bits[0] is a function of all of the intermediate_digest_bits[0], and is not a function of any bits[1] or intermediate_digest_bits[1] etc.
Each XOR flips the rollup bit from 0 -> 1 or back if and only if digest_bit is 1. So, the final value of the bit is `count_ones(digest bits) % 2`. The function `count_ones(digest bits)` is independent of the order of `digest bits` and therefore independent of the order of the sequence.
Showing that `rollup` is deterministic was the hard part, it's pretty easy to see that `count` and what we do after is deterministic so that's left as an exercise for the reader.

### Avalanche
This falls out of 2 properties - first that each individual item in the sequence has it's own `StableHash` with the avalanche property, and secondly that XOR acts as a one-time-pad cipher with the probability of each bit being exactly 50%.

### Collision resistance
This falls out of each `StableHash` value being pseudo-unique, there not being duplicate items in the sequence, and including the count.

If there are duplicate items then a hash of `ABA` is the same as the hash of `B` because the XOR cancels.

The count is necessary for any implementation of `StableHasher` which might output `0` for the empty sequence. Such a stable hasher would make it so that `HashSet([])` and `HashSet([0])` collide. I guess this probably isn't necessary in practice.

### Performance

Another property we would _like_ to show is we can get good performance or that the chosen algorithm scales well for performance as compared to other more obvious methods.

The other "obvious" method would have been to sort the sequence and feed the sorted sequence through the hasher.

Sorting takes `O(N*log(N))` instead of `N` but more crucially requires that all the data be loaded and implement `Ord` (ruling out `f64` keys, among others). Although the implementation doesn't make use of this yet, not sorting makes it possible to stream batches of ipfs data sources from the database and update the hash incrementally.

server/index-node/src/schema.graphql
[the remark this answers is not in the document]
That3Percent: It feels more natural _now_, but when there are some arbitrary count of block numbers because of subgraph composition over the availability chain and other non-Ethereum chains it's going to be weird that only the Ethereum chain is special in that it doesn't have a prefix for it's block number, and there are a bunch of optional arguments rather than optional fields on a struct.

I guess the future can wait though - especially since this is an internal API. Kind of hesitant about putting effort into changing it though.

store/postgres/src/entities.rs
```diff
@@ -823,6 +834,28 @@ fn find_schema(
         .optional()?)
 }
 
+fn supports_proof_of_indexing(
```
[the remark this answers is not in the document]
That3Percent: Yeah, I don't know either. It does work though... and only in that specific combination.

store/postgres/src/store.rs
```diff
@@ -665,17 +704,142 @@ impl Store {
...
-        let start_time = Instant::now();
-        let conn = self.conn.get();
-        let wait = start_time.elapsed();
-        if wait > Duration::from_millis(10) {
-            warn!(self.logger, "Possible contention in DB connection pool";
-                               "wait_ms" => wait.as_millis())
+    /// Execute a closure with a connection to the database.
+    ///
+    /// # API
+    ///   The API of using a closure to wrap the use of wrapper serves several
```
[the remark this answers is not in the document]
That3Percent: Sometimes I do this thing where I type a sentence and then start thinking of something else and then dont

store/postgres/src/store.rs
```diff
@@ -665,17 +704,142 @@ impl Store {
...
+        drop(cancel_guard);
+
+        // Finding cancel isn't technically unreachable, since there is nothing
+        // stopping the supplied closure from returning Canceled even if the
+        // supplied handle wasn't canceled. That would be very unexpected, the
+        // doc comment for this function says we will panic in this scenario.
+        match result {
+            Ok(t) => Ok(t),
+            Err(CancelableError::Error(e)) => Err(e),
+            Err(CancelableError::Cancel) => panic!("The closure supplied to with_entity_conn must not return Err(Canceled) unless the supplied token was canceled."),
```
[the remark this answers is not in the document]
That3Percent: The cancel token supplied to the task closure is only canceled if this future is dropped while we were waiting for the task to finish. If we reached this point, the supplied closure must have finished executing before the cancel token was canceled. If the value is then `CancelableError::Cancel` the method decided to return `CancelableError::Cancel` even though cancellation was not requested. In that case, there is no good return value for `with_conn` since it does not return `CancelableError` (nor should it, since it follows the better drop-to-cancel pattern).

store/postgres/src/store.rs
leoyvens: While that's a great explanation, I consider using `HashMap` tricky in code where determinism is a requirement. It's easy to iterate over the items without remembering that the order is not deterministic. So we should prefer `BTreeMap` whenever possible, it would be nice to have a `StableHash` implementation for it.

That3Percent: @leoyvens It would be nice to have a `StableHash` implementation for `BTreeMap`. I'm not convinced about preferring `BTreeMap` in the general case though. [The docs](https://doc.rust-lang.org/std/collections/index.html#use-a-hashmap-when) have some nice guidelines about when to choose `HashMap` or `BTreeMap` which I find agreeable.

leoyvens: Those docs kinda suggest defaulting to `HashMap` if you don't need `BTreeMap`, but I feel the opposite. Not all `HashMap` operations are deterministic. All `BTreeMap` operations are deterministic. In any codebase avoiding accidental non-determinism is good, and in ours I feel `HashMap` is a needless risk.

leoyvens: Anyways I'm not saying this PR needs to change, it's just a general note.

That3Percent: "any codebase avoiding accidental non-determinism" -> almost all codebases? I mean, it's not like our application is unusual in requiring determinism. So, you're pretty much condemning the use of `HashMap` almost everywhere.

In my experience, this is a bug that bites every developer... _once_. And it's usually a pretty easy fix with consequences not larger than other ordinary bugs. When we consider broadly applying a set of rules as a preventative measure we need to take into account the tradeoffs against the likelihood and severity of the problem manifesting.

It's a pretty subtle issue that I think reasonable people can disagree on.  I'd still choose `HashMap` here again. But, I certainly appreciate the sentiment from which your concern emerges and can respect that you may choose `BTreeMap`.

server/index-node/src/schema.graphql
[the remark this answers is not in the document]
Jannis: Yeah, understood. We can leave it the way it is.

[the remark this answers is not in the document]
That3Percent: It's ok, I already changed it to your preference. This makes it the best code according to the current requirements.

Jannis (approved):

lutter (approved):

Issue #1351, opened 2019-11-14
Fail subgraph if contract calls are not supported by the Ethereum node


Jannis:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**

When graph-node runs against a non-archive Ethereum node, contract calls are not supported. However, when making contract calls from a subgraph on such a node, it simply retries forever. All it logs are debug messages that are easy to miss if you don't have debug logs enabled with e.g. `--debug` or `GRAPH_LOG=debug`.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem.**

1. Connect graph-node to a non-archive Ethereum node.
2. Deploy a subgraph that makes contract calls.
3. Observe that it retries making the first contract call forever with no chance of recovering.

**What is the expected behavior?**

When a subgraph on such a node makes a contract call and it fails with an error like
```rust
Nov 14 14:18:17.159 DEBG Trying again after eth_call RPC call failed (attempt #1)
with result Err(Web3Error(Rpc(Error { code: ServerError(-32000), message: "This request
is not supported because your node is running with state pruning. Run with --pruning=archive.",
data: None }))), block_hash: 0xb4d3413066344d8f76368c0bd98c4a52184b9366e401215ffd0bf2e73e369b59,
block_number: 118842, subgraph_id: Qmaf6AFkpPWkULhyza2GT68NotJVL671JK2zjiLcR2Uczd,
component: SubgraphInstanceManager
```
we should immediately fail the subgraph, because there is no chance for it to ever get past this. Not all developers have the graph-node started with e.g. `--debug` and might think the call just hangs forever (which it kinda does).

leoyvens:
We currently get that response routinely when connected to Alchemy, it goes away after retrying a few times. This looks like an issue on their end but for as long as that's happening we don't want to fail the subgraph on this case.

Jannis:
This is expected to only happen with non-archive nodes (e.g. full nodes), which is why it was reported originally. But indeed, we're also seeing this against Alchemy where it is unexpected.

leoyvens:
[...]

PR #1606, opened 2020-04-14
graph: warn after successive failed retry attempts

That3Percent:
Closes part of #1351.

Rather than upgrading the log level for just this operation, this patch has the opinion that if some operation is failing repeatedly it's probably worth promoting the log level for, since repeated failure may indicate a serious condition.

So, by default promote the log level to a warning after 10 attempts. This value is configurable.

### DIFF
--- a/graph/src/util/futures.rs
+++ b/graph/src/util/futures.rs
@@ -1,7 +1,7 @@
 use crate::ext::futures::FutureExtension;
 use failure::Fail;
 use futures::prelude::*;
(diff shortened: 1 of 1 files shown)


Jannis (approved):
I like it. I know the original issue talked about failing the subgraph, but since we can't reliably tell when we're hitting an archive node or not when a call is failing, this at least makes the problem visible to subgraph authors.

osnwt:
Excellent!

PR #1607, opened 2020-04-14
Allow using `Bytes` for `ID` and switch block explorer to that

### LINKED ISSUE #1414: Allow using `Bytes` for `ID` and switch block explorer to that

lutter:
These changes make it so that the `id` field of an entity can either be of type `String` or of type `Bytes` in the subgraph's GraphQL schema. For backwards compatibility, the type `ID` can also be used, though it is now officially a synonym for `String`.

The distinction between `Bytes` and `String` only affects how the `id` of entities is stored in the database, either in a `bytea` or a `text` column. The rest of the code (e.g., `Entity`) uses strings always.

Fixes #1414 

(Replaces PR #1445 as the code had changed a lot since that PR was opened)

### DIFF
--- a/graph/src/data/subgraph/schema.rs
+++ b/graph/src/data/subgraph/schema.rs
@@ -1350,8 +1350,9 @@ fn inner_type_name(field_type: &Type, definitions: &[Definition]) -> Result<Valu
     match field_type {
         Type::NamedType(ref name) => ValueType::from_str(&name).or_else(|e| {
             if is_entity(name, definitions) {
-                // The field is a reference to another type and therefore of type ID
-                Ok(ValueType::ID)
+                // The field is a reference to another type and therefore
+                // of type `String`
+                Ok(ValueType::String)
             } else {
                 Err(e)
             }
--- a/store/postgres/src/relational.rs
+++ b/store/postgres/src/relational.rs
@@ -25,7 +25,7 @@ use crate::relational_queries::{
     DeleteDynamicDataSourcesQuery, DeleteQuery, EntityData, FilterCollection, FilterQuery,
     FindManyQuery, FindQuery, InsertQuery, RevertClampQuery, RevertRemoveQuery, UpdateQuery,
 };
-use graph::data::graphql::ext::ObjectTypeExt;
+use graph::data::graphql::ext::{DocumentExt, ObjectTypeExt};
 use graph::data::schema::{FulltextConfig, FulltextDefinition, Schema, SCHEMA_TYPE_NAME};
 use graph::data::subgraph::schema::{
     DynamicEthereumContractDataSourceEntity, POI_OBJECT, POI_TABLE,
@@ -188,23 +188,13 @@ impl Layout {
         schema_name: &str,
         create_proof_of_indexing: bool,
     ) -> Result<Self, StoreError> {
-        use s::Definition::*;
-        use s::TypeDefinition::*;
-
         SqlName::check_valid_identifier(schema_name, "database schema")?;
 
         // Extract enum types
         let enums: EnumMap = schema
             .document
-            .definitions
+            .get_enum_definitions()
             .iter()
-            .filter_map(|defn| {
-                if let TypeDefinition(Enum(enum_type)) = defn {
-                    Some(enum_type)
-                } else {
-                    None
-                }
-            })
             .map(
                 |enum_type| -> Result<(String, Arc<BTreeSet<String>>), StoreError> {
                     SqlName::check_valid_identifier(&enum_type.name, "enum")?;
@@ -225,15 +215,8 @@ impl Layout {
         // List of all object types that are not __SCHEMA__
         let object_types = schema
             .document
-            .definitions
-            .iter()
-            .filter_map(|defn| {
-                if let TypeDefinition(Object(obj_type)) = defn {
-                    Some(obj_type)
-                } else {
-                    None
-                }
-            })
+            .get_object_type_definitions()
+            .into_iter()
             .filter(|obj_type| obj_type.name != SCHEMA_TYPE_NAME)
             .collect::<Vec<_>>();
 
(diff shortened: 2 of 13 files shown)


lutter:
Rebased to latest master

store/postgres/src/relational_queries.rs
```diff
@@ -1995,8 +2038,24 @@ pub struct RevertEntityData {
     pub id: String,
 }
 
+impl RevertEntityData {
+    /// Convert primary key ids from Postgres' internal form to the format we
+    /// use by stripping `\\x` off the front of bytes strings
+    fn convert(table: &Table, mut data: Vec<RevertEntityData>) -> Vec<RevertEntityData> {
```
Jannis, line 2044: The method name could be more descriptive.

store/postgres/tests/relational_bytes.rs
```diff
@@ -0,0 +1,303 @@
...
+        assert_eq!(
+            "store error: Odd number of digits",
+            res.err().unwrap().to_string()
+        );
+
+        // We do not allow the `\x` prefix that Postgres uses
+        let res = layout.find(conn, "Thing", "\\xbadd", BLOCK_NUMBER_MAX);
+        assert!(res.is_err());
+        assert_eq!(
+            "store error: Invalid character \'\\\\\' at position 0",
```
Jannis, line 177: Who's escaping whom? 🤣

graph/src/data/subgraph/schema.rs
```diff
@@ -1331,8 +1331,9 @@ fn inner_type_name(field_type: &Type, definitions: &[Definition]) -> Result<Valu
     match field_type {
         Type::NamedType(ref name) => ValueType::from_str(&name).or_else(|e| {
             if is_entity(name, definitions) {
-                // The field is a reference to another type and therefore of type ID
-                Ok(ValueType::ID)
+                // The field is a reference to another type and therefore
+                // of type `String`
+                Ok(ValueType::String)
```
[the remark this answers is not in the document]
lutter: Yes, above the store, entity id's are always strings. I _think_ the only way to change that would be to genericize `Entity` which might be the right thing to do, but would have a lot of ripple effects throughout the code base. Instead, the id's that are really `Bytes` need to be stored as string attributes that are valid hex strings.

store/postgres/src/relational.rs
```diff
@@ -147,63 +176,101 @@ pub struct Layout {
...
+                    Some(enum_type)
+                } else {
+                    None
                 }
-                other => {
-                    return Err(StoreError::Unknown(format_err!(
-                        "can not handle {:?}",
-                        other
-                    )))
+            })
```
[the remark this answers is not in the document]
lutter: Nice! Changed.

store/postgres/src/relational.rs
```diff
@@ -147,63 +176,101 @@ pub struct Layout {
...
+            .iter()
+            .filter_map(|defn| {
+                if let TypeDefinition(Object(obj_type)) = defn {
+                    Some(obj_type)
+                } else {
+                    None
                 }
-            }
-        }
+            })
```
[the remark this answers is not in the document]
lutter: Even nicer! Changed

store/postgres/src/relational.rs
```diff
@@ -147,63 +176,101 @@ pub struct Layout {
...
+        let id_types_for_interface = schema.types_for_interface.iter().map(|(interface, types)| {
+            types
+                .iter()
+                .map(|obj_type| IdType::try_from(obj_type))
+                .collect::<Result<HashSet<_>, _>>()
+                .and_then(move |types| {
+                    if types.len() > 1 {
+                        Err(format_err!(
+                            "The implementations of interface \
+                            {} use different types for the `id` field",
```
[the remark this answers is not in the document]
lutter: Changed.

store/postgres/src/relational_queries.rs
lutter: Changed it to `bytes_as_str` since it's doing the same thing as that helper, but on a collection of stuff.

store/postgres/tests/relational_bytes.rs
lutter: Haha .. yes, counting backslashes is my new hobby

lutter:
> You can reproduce this easily by running `graph-node` with `--network-subgraphs ethereum/mainnet` and then running the following query:

I had some troible reproducing this since it seems the network indexer needs a tracing node? (Infura didn't work for me, and gave me obscure error messages, switching to Alchemy fixed it)

In any event, I managed to reproduce it, wrote some tests, and fixed it. I 

With that, all the review comments have been addressed.

store/postgres/tests/relational_bytes.rs
Jannis: 🤣

lutter:
Rebased to latest master

lutter:
Rebased to latest master

PR #1608, opened 2020-04-16
Small changes to validation for subgraph composition

lutter:
Here are a bunch of changes, mostly simplifying code, but also removing the syntax for importing by name that I started on the road towards addressing [this issue](https://github.com/graphprotocol/rfcs/issues/19) It's by no means complete, but good enough to get in right now before I switch to other things.


graph/src/data/schema.rs
```diff
@@ -268,77 +269,103 @@ pub enum SchemaImportError {
...
-            Self::Name(name) => name.hash(state),
-            Self::NameAs(name, az) => {
-                name.hash(state);
-                String::from(" as ").hash(state);
-                az.hash(state);
-            }
-        };
+        self.name.hash(state);
+        self.alias.hash(state);
+        self.explicit.hash(state);
```
leoyvens, line 290: You can derive `Hash`.

graph/src/data/schema.rs
```diff
@@ -268,77 +269,103 @@ pub enum SchemaImportError {
...
+    subgraph: SubgraphDeploymentId,
 }
 
 impl Hash for SchemaReference {
     fn hash<H: Hasher>(&self, state: &mut H) {
-        match self {
-            Self::ById(id) => id.hash(state),
-            Self::ByName(name) => name.hash(state),
-        };
+        self.subgraph.hash(state);
```
leoyvens, line 334: Again, derive `Hash`.

graph/src/data/store/mod.rs
```diff
@@ -133,6 +133,15 @@ impl FromStr for ValueType {
     }
 }
 
+impl ValueType {
+    /// Return `true` if `s` is the name of a builtin scalar type
+    pub fn is_scalar(s: &str) -> bool {
+        Self::from_str(s)
+            .map(|vt| vt != ValueType::List)
```
leoyvens, line 140: I'd let the caller use `from_str` and then have this be `fn is_scalar(&self) -> bool`. And to be more future proof exhaustively match on the `ValueType`.

lutter: I initially had it as `is_scalar(&self)` but by the time you've dealt with the `Result` from `from_str` at the call site, it's much harder to see what is actually being checked by the caller. If this ever gets used from more than one place, we can change it again.

I did make the test for `ValueType::List` and exhaustive `match` for safety; though the existence of `ValueType::List` is very suspect anyway, I think it's only used in one or two places in the code. But that's for another day.

graph/src/data/schema.rs
lutter: Done

graph/src/data/schema.rs
lutter: Done

lutter:
Rebased to latest master

PR #1609, opened 2020-04-17
Improve indexing status

leoyvens:
This partially implements https://github.com/graphprotocol/rfcs/pull/18, namely the indexing API and metadata schema changes, leaving out:
-  Returning a graphql error for failed or unhealthy subgraphs.
-  Continuing in the presence of errors.

So even though this introduces an `unhealthy` state, no subgraph can actually get in this state yet, or get any `nonFatalErrors`.

The biggest change between the implementation and the RFC, is that instead of having a single `errors` array in the metadata and indexing status, there is now a `fatalError` and an array of `nonFatalErrors`.

This also does a little bit of asyncification by turning `start_subgraph` into an `async fn`.

An integration test is added to test querying an error on the current version.


core/src/subgraph/instance_manager.rs
```diff
@@ -236,68 +240,69 @@ impl SubgraphInstanceManager {
...
-                                "expected eth adapter that matches subgraph network: {}",
-                                &network
-                            ))
-                            .clone(),
-                        manifest,
-                        metrics_registry_for_subgraph.clone(),
-                    )
-                    .map_err(|err| {
-                        error!(
+        graph::spawn_blocking(async move {
```
leoyvens, line 243: No logical changes here, just asyncifying.

core/src/subgraph/loader.rs
```diff
@@ -98,37 +98,6 @@ where
         })
     }
 
-    async fn query_dynamic_data_sources(
```
leoyvens, line 101: This was moved to the declaration of `GraphQlRunner`.

store/postgres/migrations/2020-04-10-111111_add_deployment_errors/up.sql
```diff
@@ -0,0 +1,54 @@
+-- create subgraph_error
+create table subgraphs."subgraph_error" (
+        "id"                 text not null,
+        "subgraph_id"        text not null,
```
lutter, line 4: Could we add a `references subgraphs.subgraph_deployment on delete cascade`? That way, when we clean up subgraph deployments we don't need to delete errors manually.

store/postgres/migrations/2020-04-10-111111_add_deployment_errors/up.sql
```diff
@@ -0,0 +1,54 @@
...
+        "message"            text not null,
+        "block_number"       numeric,
```
lutter, line 6: Can this ever be different from `lower(block_range)`?

store/postgres/src/subgraphs.graphql
```diff
@@ -21,8 +21,16 @@ type SubgraphVersion @entity {
...
-    failed: Boolean!
+    failed: Boolean! @deprecated(reason: "Use `health`.")
+    health: Health!
     synced: Boolean!
+
+    "If the subgraph has failed, this is the error caused it"
+    fatalError: SubgraphError
+
+    "Sorted by block number, limited to 1000"
+    nonFatalErrors: [SubgraphError!]!
```
lutter, line 32: It's a little awkward that this is not derived - is that because we couldn't distinguish between fatal and non-fatal errors if we derived this field? It might be better to have a data model where the error just has a `is_fatal` flag and the deployment has an `errors` field.

store/postgres/src/subgraphs.graphql
```diff
@@ -21,8 +21,16 @@ type SubgraphVersion @entity {
...
     manifest: SubgraphManifest!
-    failed: Boolean!
+    failed: Boolean! @deprecated(reason: "Use `health`.")
+    health: Health!
     synced: Boolean!
+
+    "If the subgraph has failed, this is the error caused it"
+    fatalError: SubgraphError
+
+    "Sorted by block number, limited to 1000"
```
lutter, line 31: Does that mean we will only ever store the latest 1000 errors for a subgraph, or that we will only return the 1000 latest errors when queried?

store/postgres/migrations/2020-04-10-111111_add_deployment_errors/up.sql
leoyvens: The operation to create this entity is currently unversioned. If you think it is viable I could make this a versioned entity, which would make sense since non-fatal errors should be cleared on a revert.

It will be a bit inconvenient to lose the ability to sort on `block_number` in a graphql query, but I can just write a lower-level query to order the `nonFatalErrros`.

store/postgres/migrations/2020-04-10-111111_add_deployment_errors/up.sql
leoyvens: Awesome! I'll do this.

store/postgres/src/subgraphs.graphql
leoyvens: Tbh I hadn't considered deriving this, though I did consider an `is_fatal` flag, but there is a complication, we only want to keep the first 1000 non-fatal errors, but we do want to make sure to expose any fatal error. 

In terms of API, I like having a separate `fatalError` because it expresses that there is at most one fatal error, and offers a straight forward way to query it.

store/postgres/src/subgraphs.graphql
leoyvens: The plan is to only store the _first_ 1000. Though we don't have non-fatal errors yet so we can still change that after this PR.

store/postgres/migrations/2020-04-10-111111_add_deployment_errors/up.sql
lutter: Yes, I agree; exposing this in GraphQL is valuable, and the redundant data field is the easiest way to get there. Could it be an `i32`?

leoyvens: In our metadata we use `BigInt` for block numbers which corresponds to `numeric` in sql.

leoyvens: Since this can be made versioned in a future PR without changing the schema, I'm thinking of not doing any change here for now.

store/postgres/migrations/2020-04-10-111111_add_deployment_errors/up.sql
leoyvens: PG tells me this would take a unique constraint on `subgraphs.subgraph_deployment(id)`. Should I add such a constraint?

lutter: Yes, I think that's a good thing in any event: for normal entities, we can't have a unique constraint on `id`, but for these metadata entities, we actually want to make sure they are unique.

leoyvens: Done!

lutter (approved):
Good stuff! The tests are failing, but once they are fixed this can be merged.

server/index-node/src/resolver.rs
```diff
@@ -176,14 +214,60 @@ impl TryFromValue for IndexingStatusWithoutNode {
...
+                error_entity.set("__typename".to_string(), "SubgraphError".to_string());
+                error_entity.set("subgraphId".to_string(), subgraph_id.to_string());
+                error_entity.set("message".to_string(), message);
+                error_entity.set("handler".to_string(), handler);
+                BTreeMap::from(error_entity)
+            };
+
+            error_object.insert("block".to_string(), block_value.into());
+
+            q::Value::Object(error_object)
```
Jannis, line 253: Since we have `entity!` now, that could be a cool way to clean this file up a little:
```rust
let error_object = entity! {
    ...
};
```
may already do the job, may not even have to convert it to a `q::Value::Object` explicitly, if `object!` does it for us (I'm not sure).

tests/integration-tests/fatal-error/test/test.js
```diff
@@ -0,0 +1,91 @@
...
+    );
+
+    // Create and deploy the subgraph
+    exec("yarn codegen");
+    exec(`yarn create:test`);
+    exec(`yarn deploy:test`);
+  });
+
+  it("subgraph fails with expected error", async () => {
+    // Wait for the subgraph to be indexed, and not fail
```
Jannis, line 88: We're expecting it to fail. The comment was copied, I think? 😉

server/index-node/src/resolver.rs
leoyvens: This is way better!

leoyvens:
@Jannis I've addressed your review.

I also needed to disable the integration tests with jsonb storage. I don't know exactly why it wasn't working, jsonb is on its way out anyways.

core/src/subgraph/instance_manager.rs
```diff
@@ -558,9 +580,15 @@ where
...
                     );
 
+                    let error = SubgraphError {
+                        subgraph_id: id_for_err.clone(),
+                        message: e.to_string(),
+                        block_ptr: Some(block_ptr),
+                        handler: None,
+                    };
+
                     // Set subgraph status to Failed
```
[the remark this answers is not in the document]
leoyvens: 'Set subgraph health to failed' might be more precise, but this is still accurate as it is.

Jannis (approved):

PR #1614, opened 2020-04-21
Partial multi blockchain refactor

Jannis:
This is a partial refactor of the code base to get us closer to a multi-blockchain architecture.

The following sections list what is included in this refactor and what isn't. The list is probably incomplete.

The main changes in this PR are:

* the creation of a registry of supported networks / network instances,
* the creation of a network store factory,
* moving subgraph indexing logic into the `graph-chain-ethereum` crate (which should really be called `graph-network-ethereum` now).

# Included

## Added

- A `Bytes::as_hex_string()` helper.
- Implementation of `From<Bytes>` for `store::Value`.
- `NetworkRegistry`, `NetworkInstance` and `Blockchain` traits to manage instances of different networks (with e.g. Ethereum being a network and Ethereum mainnet being a network instance and blockchain).
- A `NetworkStoreFactory` trait and implementation that can return stores for blockchains.
- A generic `BlockPointer` type and a `ToBlockPointer` trait.
- A `SubgraphManifest::network()` method.
- An implementations of the `NetworkRegistry` trait in `graph-core`.
- Implementations  of the `NetworkInstance` and `Blockchain` traits in the ethereum crate.

## Changed

- Instantiate network registry, network store factory and different Ethereum chain instances at node startup instead of creating Ethereum adapters etc.
- Moved subgraph instance management to the ethereum crate.
- Replaced events between registrar, provider and instance manager with method calls.
- Renamed `SubgraphInstanceManager` to `SubgraphIndexer`.
- Rework registrar, provider etc. to use network registry and network store factory.

## Removed

- The `--subgraph` option is gone. We can bring it back but I felt like this was a good time to kill it.
- The `SubgraphInstanceManager`, `SubgraphInstance`, `RuntimeHostBuilder`, `RuntimeHost` traits were all removed. Their concrete implementations remain (in the ethereum and WASM runtime crates).

# Not Included

- Make the store completely Ethereum-agnostic. This requires, among other things:
    - Moving from `EthereumBlockPointer` to `BlockPointer`/`ToBlockPointer`
    - Making tables such as `ethereum_blocks` more generic; potentially dropping them altogether.
    - Making metadata  fields like `latestEthereumBlockNumber` more generic.
- Making the WASM runtime completely Ethereum-agnostic. This requires, among other things:
    - Move Ethereum trigger handling from `RuntimeHost` to the ethereum crate. Possibly reduce `RuntimeHost`, `WasmiModule` etc. to just standing up a module and being able to calling handlers in it to which either `ToAscObj` values or `RuntimeValue` values can be passed.
    - Making it possible to inject custom host modules into WASM mappings.
    - Making the AssemblyScript ABI/conversion parts in the runtime crate public.
- Making the `NetworkIndexer` generic over the `Blockchain` trait, rename it to e.g. `BlockchainIndexer` and make it replace `BlockIngestor`, so that we don't even need a global `ethereum_blocks` type table anymore and instead store blockchain data in per-chain database schemas.

# TODO

- [ ] Call `abort()` on `AbortHandle` used for terminating unassigned subgraph deployments. It _seems_ to work without but we should make sure.
- [ ] Remove the now unused `SubgraphAssignmentProviderEvent`.
- [ ] Rename `graph-chain-ethereum` crate to `graph-network-ethereum`.


leoyvens:
> We can bring it back but I felt like this was a good time to kill it.

When we tried just adding a deprecation warning for it, people protested. If we want to try this again, it would be more gentle to start with a deprecation warning.

Jannis:
@leoyvens I'll bring it back, should be easy enough.

graph/src/components/networks/mod.rs
```diff
@@ -0,0 +1,76 @@
...
+use async_trait::*;
+
+use crate::prelude::{format_err, Error, SubgraphManifest};
+use futures03::future::AbortHandle;
+
+pub mod blockchains;
+use blockchains::Blockchain;
+
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub struct NetworkInstanceId {
```
lutter, line 14: Could you add a comment explaining what such an id signifies? Is this something like `ethereum/mainnet` where `ethereum` is the `network` and `name` is the specific instance?

chain/ethereum/src/chain/mod.rs
```diff
@@ -0,0 +1,282 @@
...
+    let (transport_event_loop, transport) = Transport::new_rpc(descriptor.url.as_str());
+    let adapter_metrics = Arc::new(ProviderEthRpcMetrics::new(metrics_registry));
+    let adapter = Arc::new(EthereumAdapter::new(transport, adapter_metrics));
+
+    // Connect to the node and obtain the network ID and version
+    let identifier = adapter.net_identifiers(&logger).compat().await?;
+
+    // If we drop the event loop the transport will stop working.
+    // For now it's fine to just leak it.
+    std::mem::forget(transport_event_loop);
```
lutter, line 126: Would it be better to store it in the `Connection` so it gets dropped when the connection gets dropped? That way, there's no leak.

lutter (changes requested):
This PR is massive, and I can't say I've followed all that much of what it is doing. The PR puts a number of abstractions in place that will only have one implementer for the foreseeable future; how sure are we that all these abstractions will be what we need when we work on multi-blockchain? Are there changes that could be cut? I do like though how some of them componentize Ethereum-specific concepts.

leoyvens:
This is hopelessly conflicted and have a modified plan for multi-blockchain, so closing this PR.

PR #1618, opened 2020-04-22
Pass `log_index` to mappings in place of `transaction_log_index`

leoyvens:
Geth does not support `transaction_log_index`, and subgraphs that use it only care that it identifies the log, the specific value is not important. Still this will change the output of subgraphs that use this field.


Jannis (approved):
I think this is ok. Looking at Geth's GraphQL schema for events (logs), they only have an `index` field and that's the index of the log inside the block: https://github.com/ethereum/go-ethereum/blob/master/graphql/schema.go#L59-L60.

The Ethereum JSON-RPC "spec" also does not include a `transactionLogIndex` field on events: https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_getfilterchanges.

leoyvens:
@Jannis Interesting. We'll also need to properly remove all the unsupported fields in a new api version, but this should be enough to keep subgraphs running on a Geth endpoint.

PR #1620, opened 2020-04-23
Increase connection pool size

leoyvens:
The tests were running out of connections and hanging on my machine with the current size.

### DIFF
--- a/store/test-store/src/lib.rs
+++ b/store/test-store/src/lib.rs
@@ -41,7 +41,7 @@ lazy_static! {
                 net_version: NETWORK_VERSION.to_owned(),
                 genesis_block_hash: GENESIS_PTR.hash,
             };
-            let conn_pool_size: u32 = 10;
+            let conn_pool_size: u32 = 20;
             let postgres_conn_pool = create_connection_pool(
                 postgres_url.clone(),
                 conn_pool_size,


leoyvens:
@lutter It only happened when I ran a large number of tests in parallel, the graph-core interface query tests which do run in parallel. I'm not sure what changed (in my system or in the code) that would cause this to start happening, but it was related to requesting a connection (in `fn api_schema`) while already in a transaction.

If all of the connections are in a transaction which then requests another connection, that's a deadlock. It could conceivably happen in a production environment.

leoyvens:
I'll merge this but we should open an issue to track the deadlock if it's not tracked already.

lutter:
> @lutter It only happened when I ran a large number of tests in parallel, the graph-core interface query tests which do run in parallel. I'm not sure what changed (in my system or in the code) that would cause this to start happening, but it was related to requesting a connection (in `fn api_schema`) while already in a transaction.
> 
> If all of the connections are in a transaction which then requests another connection, that's a deadlock. It could conceivably happen in a production environment.

Yes, this would be very bad in production. We should maybe try and rejigger the code to pass the connection we already have into `api_schema`.

leoyvens:
@lutter opened #1631 to track this.

PR #1622, opened 2020-04-23
Update Ethereum network indexer to include transactions and logs

fordN:
This PR updates the Ethereum network indexer to process transactions and transaction receipts from each block.  Additional entities have been created on the Ethereum subgraph to store the transactions, transaction logs and their relationship. 

Nested queries linking blocks, transactions and logs are now supported:
```graphql
query {
  blocks(orderBy:number orderDirection:desc where:{logs_not:[]}){
    id
    number
    transactionCount
    transactions {
      id 
      logs {
        id
      }
    }
    logs {
      id
      account
      data      
      transaction {
        id
      }
    }
  }
}
```
```graphql
query {
    logs {
      transaction {
        block {
          id
        }
      }
      id
      account
      data      
      transaction {
        id
      }
    }
  }

```

### DIFF
--- a/chain/ethereum/src/network_indexer/convert.rs
+++ b/chain/ethereum/src/network_indexer/convert.rs
@@ -120,6 +120,7 @@ impl TryIntoEntity for Ommer {
                     .collect::<Vec<_>>()
                     .into()),
             ),
+            ("logCount", (0 as i32).into()),
             ("size", inner.size.into()),
             ("sealFields", inner.seal_fields.clone().into()),
         ] as Vec<(_, Value)>))
@@ -130,6 +131,11 @@ impl TryIntoEntity for &BlockWithOmmers {
     fn try_into_entity(self) -> Result<Entity, Error> {
         let inner = self.inner();
         let receipts = &self.block.transaction_receipts;
+        let logs = receipts
+            .iter()
+            .flat_map(|receipt| receipt.logs.iter())
+            .map(|log| format!("{:x}-{}", log.block_hash.unwrap(), log.log_index.unwrap()))
+            .collect::<Vec<_>>();
 
         Ok(Entity::from(vec![
             ("id", format!("{:x}", inner.hash.unwrap()).into()),
@@ -173,15 +179,8 @@ impl TryIntoEntity for &BlockWithOmmers {
                     .collect::<Vec<_>>()
                     .into()),
             ),
-            (
-                "logs",
-                (receipts
-                    .iter()
-                    .flat_map(|receipt| receipt.logs.iter())
-                    .map(|log| format!("{:x}-{}", log.block_hash.unwrap(), log.log_index.unwrap()))
-                    .collect::<Vec<_>>()
-                    .into()),
-            ),
+            ("logCount", (logs.len() as i32).into()),
+            ("logs", logs.into()),
             ("size", inner.size.into()),
             ("sealFields", inner.seal_fields.clone().into()),
         ] as Vec<(_, Value)>))
@@ -208,6 +207,8 @@ impl TryIntoEntity for &LoadedTransaction {
                 "block",
                 format!("{:x}", transaction.block_hash.unwrap()).into(),
             ),
+            ("blockNumber", transaction.block_number.unwrap().into()),
+            ("logCount", (receipt.logs.len() as i32).into()),
             (
                 "logs",
                 receipt
--- a/chain/ethereum/src/network_indexer/convert.rs
+++ b/chain/ethereum/src/network_indexer/convert.rs
@@ -202,7 +202,6 @@ impl TryIntoEntity for &LoadedTransaction {
             ("value", transaction.value.into()),
             ("gasPrice", transaction.gas_price.into()),
             ("gas", transaction.gas.into()),
-            ("inputData", transaction.input.clone().into()),
             (
                 "block",
                 format!("{:x}", transaction.block_hash.unwrap()).into(),
--- a/chain/ethereum/src/network_indexer/convert.rs
+++ b/chain/ethereum/src/network_indexer/convert.rs
@@ -34,6 +34,22 @@ impl ToEntityKey for &BlockWithOmmers {
     }
 }
 
+impl ToEntityId for Transaction {
+    fn to_entity_id(&self) -> String {
+        format!("{:x}", self.0.hash)
+    }
+}
+
+impl ToEntityKey for &Transaction {
+    fn to_entity_key(&self, subgraph_id: SubgraphDeploymentId) -> EntityKey {
+        EntityKey {
+            subgraph_id,
+            entity_type: "Transaction".into(),
+            entity_id: format!("{:x}", self.0.hash),
+        }
+    }
+}
+
 impl TryIntoEntity for Ommer {
     fn try_into_entity(self) -> Result<Entity, Error> {
         let inner = &self.0;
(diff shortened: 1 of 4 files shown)


chain/ethereum/src/network_indexer/convert.rs
```diff
@@ -113,9 +169,101 @@ impl TryIntoEntity for &BlockWithOmmers {
...
+        let inner = &self.0;
+
+        Ok(Entity::from(vec![
+            (
+                "id",
+                format!(
+                    "{:x}-{}",
+                    inner.block_hash.unwrap(),
+                    inner.log_index.unwrap()
+                )
```
fordN, line 245: I use "{block_hash}-{log_index}" here to build unique ids for log entities,  another option is to use "{transaction_hash}-{log_index}".  It's not obvious to me which approach will be more clear to users, or will support more useful queries.
   
I don't foresee either id structure being necessary for querying since one can query for transaction or block entities at the top level and get logs as a sub-entity. For example, getting all logs from a specific block or transaction is easily doable without filtering on the prefix of the log id.

Jannis: Yeah, I think the IDs should be treated as opaque values. Nobody should really filter logs with `id_starts_with: { "sometxhash-" }`.

What we should really do though is use `Bytes` for all these IDs because they are more efficiently stored and indexed by Postgres.

chain/ethereum/src/network_indexer/ethereum.graphql
```diff
@@ -62,25 +62,126 @@ type Block @entity {
...
+  value: BigInt!
+
+  """ The price offered to miners for gas, in wei per unit. """
+  gasPrice: BigInt!
+
+  """ The maximum amount of gas this transaction can consume. """
+  gas: BigInt!
+
+  #  """ The data supplied to the target of the transaction. """
+  #  inputData: Bytes!
```
Jannis, line 123: Are we not able to include this data?

chain/ethereum/src/network_indexer/ethereum.graphql
```diff
@@ -62,25 +62,126 @@ type Block @entity {
...
+  """
+  The block this transaction was mined in.
+  This will be null if the transaction has not yet been mined.
+  """
+  block: Block
+
+  """ The number of the block this transaction was mined in. """
+  blockNumber: Int!
+
+  """The number of logs emitted in this transaction."""
```
Jannis, line 134: Inconsistency in formatting. Add space between the text and `"""` at both ends.

chain/ethereum/src/network_indexer/ethereum.graphql
```diff
@@ -62,25 +62,126 @@ type Block @entity {
...
+  """
+  logs: [Log!]
+
+  """
+  The return status of the transaction.
+  This will be 1 if the transaction succeeded,
+  or 0 if it failed (due to a revert, or due to running out of gas).
+  If the transaction has not yet been mined, this field will be null.
+  """
+  status: BigInt
```
Jannis, line 149: If the only possible values are `0`, `1` and `null` can just be `Int`.

chain/ethereum/src/network_indexer/ethereum.graphql
```diff
@@ -62,25 +62,126 @@ type Block @entity {
...
+  """
+  The account that was created by a contract creation transaction.
+  If the transaction was not a contract creation transaction, or it has not yet been mined, this field will be null.
+  """
+  createdContract: Bytes
+}
+
+""" An Ethereum event log. """
+type Log @entity {
+  id: ID!
```
Jannis, line 172: Make this `Bytes!`, which is now supported.

chain/ethereum/src/network_indexer/ethereum.graphql
```diff
@@ -62,25 +62,126 @@ type Block @entity {
...
   """Size of the block in bytes."""
   size: BigInt
 
   """Seal fields."""
   sealFields: [Bytes!]!
 }
+
+""" An Ethereum transaction. """
+type Transaction @entity {
+  id: ID!
```
Jannis, line 93: Make this `Bytes!`, which is now supported.

chain/ethereum/src/network_indexer/ethereum.graphql
```diff
@@ -62,25 +62,126 @@ type Block @entity {
...
+  If the transaction has not yet been mined, this field will be null.
+  """
+  cumulativeGasUsed: BigInt
+
+  """
+  The account that was created by a contract creation transaction.
+  If the transaction was not a contract creation transaction, or it has not yet been mined, this field will be null.
+  """
+  createdContract: Bytes
+}
```
Jannis, line 168: I see that Geth also includes
```graphql
r: BigInt!
s: BigInt!
v: BigInt!
```

But I agree that we don't need this information for now; besides, it's not directly available from the rust-web3 types.

chain/ethereum/src/network_indexer/mod.rs
```diff
@@ -83,6 +85,40 @@ impl fmt::Display for BlockWithOmmers {
     }
 }
 
+// Helper type to represent transactions.
+#[derive(Clone, Debug, Default, PartialEq)]
+pub struct LoadedTransaction {
```
Jannis, line 90: Let's call it `FullTransaction`

chain/ethereum/src/network_indexer/convert.rs
```diff
@@ -34,6 +34,46 @@ impl ToEntityKey for &BlockWithOmmers {
     }
 }
 
+impl ToEntityId for LoadedTransaction {
```
Jannis, line 37: @lutter How would we best deal with the fact that references are strings but the IDs should be represented as `Bytes!`?

Do we keep this `ToEntityId` impl for references but when we create the `Transaction` and `Log` entities and entity keys for them, we generate the corresponding value as `Bytes`?

chain/ethereum/src/network_indexer/convert.rs
```diff
@@ -113,9 +169,101 @@ impl TryIntoEntity for &BlockWithOmmers {
...
+            ),
+            ("index", inner.log_index.unwrap().into()),
+            ("account", inner.address.into()),
+            (
+                "topics",
+                inner
+                    .topics
+                    .iter()
+                    .cloned()
+                    .map(move |topic| topic)
```
Jannis, line 256: I don't think this is needed if we use `cloned()`.

chain/ethereum/src/network_indexer/convert.rs
```diff
@@ -113,9 +169,101 @@ impl TryIntoEntity for &BlockWithOmmers {
...
+        ] as Vec<(_, Value)>))
+    }
+}
+
+impl TryIntoEntity for &LoadedTransaction {
+    fn try_into_entity(self) -> Result<Entity, Error> {
+        let transaction = &self.transaction;
+        let receipt = &self.receipt;
+
+        Ok(Entity::from(vec![
```
Jannis, line 195: It could be worth switching to the `object!` macro in a separate commit. We can always do that later though, so I'll leave it up to you.

chain/ethereum/src/network_indexer/ethereum.graphql
fordN: There are some blocks with exceedingly large `inputData` which causes insert errors since the btree index row size limits are hit. 
My thinking is the way past the size error is with custom indices, so the inputData field can be set as one that does not get indexed.

leoyvens:
Network indexer as it currently exists is being removed in #3055.

PR #1623, opened 2020-04-23
Fix entities updated after remove

### LINKED ISSUE #1598: store.remove('TypeName', id) does not (always) remove an entity
Handing an event, I want to clear most optional props of an entity and keep just few like `isRemoved` with other data nulled (to show that the instance was created but later removed).

To make life easier I try to remove the instance from the store, then recreate it as a new and set just few props:

```
  store.remove('DocumentObject', id)

  let obj = new DocumentObject(id)
  obj.isRemoved = true
  obj.save()
```

But instead of making a new fresh object, this code revives the existing instance with all fields kept from the past. 

What is the proper way to do this?

leoyvens:
Previously we'd ignore the removal, merging the update with the version in the store.

The fix required changing the structure of `EntityCache` to explicitly keep track of entities which were removed at some point.

Resolves #1598. Includes integration test.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,9 @@
 **/*.idea
 **/*.DS_STORE
 
+# IDE files
+.vscode/
+
 # Ignore data files created by running the Docker Compose setup locally
 /docker/data/
 /docker/parity/chains/
--- a/chain/ethereum/src/network_indexer/block_writer.rs
+++ b/chain/ethereum/src/network_indexer/block_writer.rs
@@ -85,7 +85,7 @@ impl BlockWriter {
             logger,
             subgraph_id: self.subgraph_id.clone(),
             store: self.store.clone(),
-            cache: EntityCache::new(),
+            cache: EntityCache::new(self.store.clone()),
             metrics: self.metrics.clone(),
         };
         context.write(block)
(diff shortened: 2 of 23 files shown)


lutter:
Still trying to understand what went wrong here. I think the sequence of events is something like:
1. Entity is deleted -> `None` entry in `EntityCache.updates`
2. Entity is recreated -> `Some(entity)` in `EntityCache.updates`

At that point, we can't tell whether `entity` should be merged with `current` or written to the store as is; i.e. with the two steps above, any attribute not set in `entity` should be null, whereas in a 'normal' update any attribute not set should retain the value it has in `current`. 

It seems another way to fix this would be to notice that we are overwriting a `None` in `EntityCache.updates` with `Some(entity)` and setting all attributes in `entity` to null that are not set explicitly (we could get the list of attributes from `current`)

I am not sure if that would be better than what you have done, but it might be a smaller change.

lutter:
One more note on integration tests: it's great to have that to make sure this works end-to-end. But it's also important to have cheaper tests that will fail if we break this behavior, since running integration tests is much more expensive than unit tests. Could you add a test for `EntityCache` that fixes that sequence of steps?

leoyvens:
@lutter I tried out the fix you propose, since it really requires less changes to the internal logic of the `EntityCache`. The diff ends up being larger, mostly because `EntityCache::set` needs to return an error which is a bit annoying, but overall I think I prefer this solution.

I did have to change from `.merge` to `.merge_remove_null_fields` when merging the update with the current entity to return it to mapping. This was due to https://github.com/graphprotocol/graph-cli/pull/481 but I think it makes sense anyways.

leoyvens:
On the tests, if these integration tests become too slow to be run in most situations where we run other tests, they will lose an important part of their usefulness. That doesn't seem to be the case yet, but as the number of tests grows that may happen. I rather address that by improving the performance of the integration tests than duplicating them as unit tests, because I find the integration tests easier to write, read and maintain through refactors.

PR #1629, opened 2020-04-29
Don't return `__typename` to mappings

leoyvens:
We didn't mean for that to happen, and also causes a difference between entities that are set and then returned from cache vs entities returned from the store.

Also the `ipfs_map` test was causing trouble for me locally, and refactoring it a bit fixed it.

### DIFF
--- a/graph/src/components/store.rs
+++ b/graph/src/components/store.rs
@@ -1717,7 +1717,11 @@ impl LfuCache<EntityKey, Option<Entity>> {
     ) -> Result<Option<Entity>, QueryExecutionError> {
         match self.get(&key) {
             None => {
-                let entity = store.get(key.clone())?;
+                let mut entity = store.get(key.clone())?;
+                if let Some(entity) = &mut entity {
+                    // `__typename` is for queries not for mappings.
+                    entity.remove("__typename");
+                }
                 self.insert(key.clone(), entity.clone());
                 Ok(entity)
             }
--- a/runtime/wasm/src/module/test.rs
+++ b/runtime/wasm/src/module/test.rs
@@ -377,20 +377,17 @@ async fn ipfs_map() {
     let ipfs = Arc::new(ipfs_api::IpfsClient::default());
     let subgraph_id = "ipfsMap";
 
-    let run_ipfs_map = move |json_string| {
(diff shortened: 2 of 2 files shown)


graph/src/components/store.rs
```diff
@@ -1717,7 +1717,11 @@ impl LfuCache<EntityKey, Option<Entity>> {
     ) -> Result<Option<Entity>, QueryExecutionError> {
         match self.get(&key) {
             None => {
-                let entity = store.get(key.clone())?;
+                let mut entity = store.get(key.clone())?;
+                if let Some(entity) = &mut entity {
+                    // `__typename` is for queries not for mappings.
+                    entity.remove("__typename");
+                }
```
lutter, line 1724: As I said in slack, I think the right place to do this is in `HostExports::store_set` before we hand the entity to POI, and we should remove _everything_ that is not an attribute mentioned in the GraphQL schema - that way, we make sure that POI does not change if we ever need to add additional attributes into entities to support queries.

lutter (approved):
Let's see if this improves things; if it does, we should either put logic to remove any non-entity attributes from entities we return from `store.get` into `store_get` or make it part of the contract of `EntityCache.get`

runtime/wasm/src/module/test.rs
```diff
@@ -377,20 +377,17 @@ async fn ipfs_map() {
     let ipfs = Arc::new(ipfs_api::IpfsClient::default());
```
Jannis, line 377: Did this file have to change at all?

graph/src/components/store.rs
That3Percent: I think the comment may not reflect your latest thinking, but I want to make clear for the record here as was discussed on slack that I strongly disagree with any change that moves code before the PoI event emit

runtime/wasm/src/module/test.rs
leoyvens: These changes are unrelated, I had issues running the tests locally which apparently resolved with this changes.

leoyvens:
Merging in the spirit of moving forward (which we all share), the ideal fix is tbd.

PR #1630, opened 2020-04-29
graph, runtime: Env variable to log PoI events deterministically

That3Percent:
This PR adds an environment variable `GRAPH_LOG_POI_EVENTS` which when set to `true` will log the PoI stream deterministically within a given subgraph for efficient log diffing.

### DIFF
--- a/graph/src/components/subgraph/proof_of_indexing.rs
+++ b/graph/src/components/subgraph/proof_of_indexing.rs
@@ -1,10 +1,19 @@
-use crate::prelude::Value;
+use crate::prelude::{debug, Logger, Value};
 use stable_hash::{prelude::*, utils::StableHasherWrapper, SequenceNumberInt};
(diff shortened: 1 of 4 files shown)


graph/src/components/subgraph/proof_of_indexing.rs
```diff
@@ -28,6 +37,42 @@ pub enum ProofOfIndexingEvent<'a> {
...
+                builder.field("entity_type", entity_type);
+                builder.field("id", id);
+                builder.field("data", &data.iter().collect::<BTreeMap<_, _>>());
+            }
+        }
+        builder.finish()
+    }
+}
+
+impl slog::Value for ProofOfIndexingEvent<'_> {
```
Jannis, line 65: Sweeet

Jannis (approved):
Perfect!

PR #1632, opened 2020-04-29
server: Add test for canonical serialization of JSON responses

lutter:
The JSON that `serde_json` produces is already what we want. This adds tests that make sure that the serialization of responses remains what we expect, even if the underlying implementation changes.


Jannis:
Nice. The one test that fails does so because of network errors. This is good to merge.

PR #1640, opened 2020-05-04
Normalize BigDecimal values

leoyvens:
By putting any trailing zeros in the scale. This solves determinism issues because when an entity makes a roundtrip through the store it gets normalized by postgres anyways, at least in some cases.

This is mostly just plumbing through the impls for the `BigDecimal` newtype.

One strange consequence is that the workaround in https://github.com/graphprotocol/graph-node/pull/1182 is no longer necessary. It seems that issue only manifested in non-normalized BigDecimals.

### DIFF
--- a/graph/src/data/store/scalar.rs
+++ b/graph/src/data/store/scalar.rs
@@ -1,3 +1,6 @@
+use diesel::deserialize::FromSql;
+use diesel::serialize::ToSql;
+use diesel_derives::{AsExpression, FromSqlRow};
 use failure::Fail;
 use hex;
 use num_bigint;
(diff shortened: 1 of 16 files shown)


graph/src/data/store/scalar.rs
```diff
@@ -10,14 +13,158 @@ use stable_hash::{
...
-pub type BigDecimal = bigdecimal::BigDecimal;
+/// All operations on `BigDecimal` return a normalized value.
+// Caveat: The exponent is currently an i64 and may overflow. See
+// https://github.com/akubera/bigdecimal-rs/issues/54.
+// Using `#[serde(from = "BigDecimal"]` makes sure deserialization calls `BigDecimal::new()`.
+#[derive(
+    Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize, AsExpression, FromSqlRow,
+)]
+#[serde(from = "BigDecimal")]
+#[sql_type = "diesel::sql_types::Numeric"]
```
Jannis, line 30: I would prefer not to add any more diesel-dependent code to the `graph` crate. Could this be done in the diesel store? (Might require an extra newtype around `BigDecimal`?)

leoyvens: It would require a newtype, and between this and a newtype, I see this as the lesser evil.

graph/src/data/store/scalar.rs
```diff
@@ -10,14 +13,158 @@ use stable_hash::{
...
+        self.0.digits()
+    }
+
+    // Copy-pasted from `bigdecimal::BigDecimal::normalize`. We can use the upstream version once it
+    // is included in a released version supported by Diesel.
+    pub fn normalize(&self) -> BigDecimal {
+        let (bigint, exp) = self.0.as_bigint_and_exponent();
+        let (_, mut digits) = bigint.to_u32_digits();
+        let mut digits_iter = digits.iter().rev();
+        let mut trailing_count = 0;
```
That3Percent, line 58: One-liner: `let trailing_count = digits.iter().rev().take_while(|i| **i == 0).count() as i64;`

graph/src/data/store/scalar.rs
```diff
@@ -10,14 +13,158 @@ use stable_hash::{
...
+        while digits_iter.next() == Some(&0) {
+            trailing_count += 1;
+        }
+        digits.truncate(digits.len() - trailing_count as usize);
```
That3Percent, line 62: I copied the tests from the [bigdecimal PR](https://github.com/akubera/bigdecimal-rs/commit/03f56367e27b3f4fcaf70b117bcd6fc49c64593a#diff-b4aea3e418ccdb71239b96952d9cddb6R2822) into this repo, and they failed. So, something is awry.

Here's what I got...
```
thread 'data::store::scalar::test::test_normalize' panicked at 'assertion failed: `(left == right)`
  left: `BigDecimal(BigDecimal("1000"))`,
 right: `BigDecimal(BigDecimal("10"))`', graph/src/data/store/scalar.rs:594:13
```

This works if the exponent is not negated in our newtype constructor.

graph/src/data/store/scalar.rs
```diff
@@ -10,14 +13,158 @@ use stable_hash::{
...
+        let int_val = num_bigint::BigInt::new(self.0.sign(), digits);
```
That3Percent, line 63: After bringing in the tests I also added the line `assert_eq!(not_normalized.normalize().to_string(), string);` between the other asserts. This causes the tests to fail:
```
thread 'data::store::scalar::test::test_normalize' panicked at 'assertion failed: `(left == right)`
  left: `"0.10"`,
 right: `"0.1"`', graph/src/data/store/scalar.rs:602:13
```

graph/src/data/store/scalar.rs
```diff
@@ -10,14 +13,158 @@ use stable_hash::{
...
+        self.0.as_bigint_and_exponent()
+    }
+
+    pub(crate) fn digits(&self) -> u64 {
+        self.0.digits()
+    }
+
+    // Copy-pasted from `bigdecimal::BigDecimal::normalize`. We can use the upstream version once it
+    // is included in a released version supported by Diesel.
+    pub fn normalize(&self) -> BigDecimal {
```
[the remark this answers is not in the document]
leoyvens: Done, nice tip with `#[must_use]`.

graph/src/data/store/scalar.rs
```diff
@@ -10,14 +13,158 @@ use stable_hash::{
...
+/// All operations on `BigDecimal` return a normalized value.
+// Caveat: The exponent is currently an i64 and may overflow. See
+// https://github.com/akubera/bigdecimal-rs/issues/54.
+// Using `#[serde(from = "BigDecimal"]` makes sure deserialization calls `BigDecimal::new()`.
+#[derive(
+    Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize, AsExpression, FromSqlRow,
+)]
+#[serde(from = "bigdecimal::BigDecimal")]
+#[sql_type = "diesel::sql_types::Numeric"]
+pub struct BigDecimal(bigdecimal::BigDecimal);
```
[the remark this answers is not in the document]
leoyvens: Done.

graph/src/data/store/scalar.rs
```diff
@@ -10,14 +13,158 @@ use stable_hash::{
...
+
+impl From<bigdecimal::BigDecimal> for BigDecimal {
+    fn from(big_decimal: bigdecimal::BigDecimal) -> Self {
+        BigDecimal(big_decimal).normalize()
+    }
+}
+
+impl BigDecimal {
+    pub fn new(digits: BigInt, exp: i64) -> Self {
+        Self::from(bigdecimal::BigDecimal::new(digits.0, -exp))
```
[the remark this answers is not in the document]
leoyvens: Added a comment, bidecimal chose to flip this, strangely.

graph/src/data/store/scalar.rs
```diff
@@ -10,14 +13,158 @@ use stable_hash::{
...
+
+    pub(crate) fn digits(&self) -> u64 {
+        self.0.digits()
+    }
+
+    // Copy-pasted from `bigdecimal::BigDecimal::normalize`. We can use the upstream version once it
+    // is included in a released version supported by Diesel.
+    pub fn normalize(&self) -> BigDecimal {
+        let (bigint, exp) = self.0.as_bigint_and_exponent();
+        let (_, mut digits) = bigint.to_u32_digits();
```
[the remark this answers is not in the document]
leoyvens: Done.

graph/src/data/store/scalar.rs
leoyvens: Nice!

graph/src/data/store/scalar.rs
leoyvens: I copied over the test for good measure, but they were testing fine (though they were wrong as you noticed below). Because our constructor handles the exponent in the more natural way while bigdecimal flips it, the test needs to be adapted for our constructor.

graph/src/data/store/scalar.rs
leoyvens: Huge thanks for testing this better than I did! The normalization was totally wrong because I wasn't getting the digits in the correct way, and the test was wrong because it was testing numerical equality and not the equality in the representation. Fixed now, thanks again!

leoyvens: Also I'm normalizing `0` now.

leoyvens:
@That3Percent I've addressed all points in your review, I'll also submit a PR to fix upstream once this is approved. Thanks for diving into this and noticing it was incorrect! Fixed now.

leoyvens:
Problems remain, I suspect it's because we allow values which diesel will overflow, investigating.

graph/src/data/store/scalar.rs
```diff
@@ -167,7 +319,7 @@ impl BigInt {
         }
         let mut byte_array = if exp >= 0.into() { [0; 8] } else { [255; 8] };
         byte_array[..bytes.len()].copy_from_slice(&bytes);
-        BigDecimal::new(self.0, -i64::from_le_bytes(byte_array))
```
That3Percent, line 170: I reviewed all the callers of this. Some places enforce an exponent of 0 (resulting often in non-normalized values) but those places then use the decimal in a way that shouldn't be a problem for consistency. Other places that use this code immediately do something with the decimal so that the result is normalized (like divide). I wonder if it's a  🐛💣  to not normalize here though.

Taking that further, should normalization just move into `new`?. I tried to evaluate each code path and everything looks good now, but it's unintuitive that the normalization happens via round-tripping through the underlying type. This structure lead to more code needing to be checked than necessary.

graph/src/data/store/scalar.rs
```diff
@@ -10,30 +13,179 @@ use stable_hash::{
...
+    // Copy-pasted from `bigdecimal::BigDecimal::normalize`. We can use the upstream version once it
+    // is included in a released version supported by Diesel.
+    #[must_use]
+    pub fn normalized(&self) -> BigDecimal {
+        if self == &BigDecimal::zero() {
+            return BigDecimal::zero();
+        }
+        let (bigint, exp) = self.0.as_bigint_and_exponent();
+        let (sign, mut digits) = bigint.to_radix_be(10);
+        let trailing_count = digits.iter().rev().take_while(|i| **i == 0).count() as i64;
```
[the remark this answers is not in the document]
leoyvens: The cast is needed in an expression below, so I moved it there.

leoyvens:
A few more changes were necessary to get this working and improve the state of our BigDecimal support:
- Don't use Diesel's support for BigDecimal when inserting. It is fishy and they understandably don't have the intention of fixing that, see https://github.com/diesel-rs/diesel/issues/2385.
- Refactor away `to_big_decimal`, as per @That3Percent's feedback above.
- Check that the BigDecimal exponent is within the bounds accepted by the bigdecimal128 standard. We needed a limit anyways, and conforming to the standard might make it easier to switch to it later on. This has the potential of breaking subgraphs that have decimals outside these bounds.
- `_in` filters for BigDecimal and BigInt no longer work in JSONB subgraphs. JSONB is to be removed soon anyways, I couldn't be bothered to keep that working.

That3Percent:
> * `_in` filters for BigDecimal and BigInt no longer work in JSONB subgraphs. JSONB is to be removed soon anyways, I couldn't be bothered to keep that working.

Are we really not concerned about this? JSONB has been "to be removed soon" for months now.  I don't understand how we can be confident this doesn't break existing deployments until JSONB is actually removed with all JSONB subgraphs removed from the hosted service.

store/postgres/tests/relational.rs
```diff
@@ -121,12 +123,22 @@ lazy_static! {
         entity.set("id", "one");
         entity.set("bool", true);
         entity.set("int", std::i32::MAX);
-        entity.set("bigDecimal", (*LARGE_DECIMAL).clone());
+        let decimal = (*LARGE_DECIMAL).clone();
```
That3Percent, line 126: Let's not make this change now, but in the future if you encounter a test which is constructing an entity, consider re-factoring it to use the new `entity!` macro. Ditto for `object!` in GraphQL.

leoyvens:
@That3Percent I'm also unsure, I expect @lutter can clarify whether that's ok or not.

store/postgres/tests/relational.rs
leoyvens: Thanks for the tip, I only used the `object!` macro so far.

store/postgres/src/relational_queries.rs
```diff
@@ -338,31 +338,26 @@ impl<'a> QueryFragment<Pg> for QueryValue<'a> {
...
-                    out.push_bind_param::<Numeric, _>(d)
-                }
+                out.push_bind_param::<Text, _>(&d.to_string())?;
+                out.push_sql("::numeric");
+                Ok(())
             }
             Value::Bool(b) => out.push_bind_param::<Bool, _>(b),
             Value::List(values) => {
-                let values = SqlValue::new_array(values.clone());
+                let sql_values = SqlValue::new_array(values.clone());
```
lutter, line 347: Minor quibble about the variable renaming: I usually rebind the same variable in situations like this to indicate to the reader that the old value will not be used in the code that follows. That reduces the number of things to think about, and helps making things more understandable.

leoyvens: Nothing against that, it's just that I actually needed the old value in this case, and couldn't come up with something more elegant.

lutter: Oh, woops .. I'll get off my soapbox then ;)

PR #1643, opened 2020-05-06
Add logging of DB connection contention back in

lutter:
I was too hasty when I said we don't need to log about excessive waits when getting a connection from the database pool. Turns out when a node is going sideways it's useful to see when contention starts.

This PR adds logging about that back in, but done as a r2d2 event handler, and with a threshold of 100ms instead of 10ms. It also adds a Prometheus gauge that reports the number of currently checked-out connections.

### DIFF
--- a/store/postgres/src/connection_pool.rs
+++ b/store/postgres/src/connection_pool.rs
@@ -27,23 +27,47 @@ impl r2d2::HandleError<r2d2::Error> for ErrorHandler {
     }
 }
 
-#[derive(Debug)]
-struct EventHandler(Logger);
+struct EventHandler {
+    logger: Logger,
+    gauge: Box<Gauge>,
+}
+
+impl EventHandler {
+    fn new(logger: Logger, registry: Arc<dyn MetricsRegistry>) -> Self {
+        let gauge = registry
+            .new_gauge(
+                String::from("store_connection_checkout_count"),
+                String::from("The number of Postgres connections currently checked out"),
+                HashMap::new(),
+            )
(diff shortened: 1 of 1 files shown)


store/postgres/src/connection_pool.rs
```diff
@@ -24,6 +27,49 @@ impl r2d2::HandleError<r2d2::Error> for ErrorHandler {
...
+}
+
+impl HandleEvent for EventHandler {
+    fn handle_acquire(&self, _: e::AcquireEvent) {}
+    fn handle_release(&self, _: e::ReleaseEvent) {}
+    fn handle_checkout(&self, event: e::CheckoutEvent) {
+        self.gauge.inc();
+        if event.duration() > Duration::from_millis(CONTENTION_LOG_THRESHOLD) {
+            warn!(self.logger, "Excessive wait time on checkout";
+            "wait_ms" => event.duration().as_millis())
```
Jannis, line 61: nit: The code formatting is a bit off here

store/postgres/src/connection_pool.rs
```diff
@@ -24,6 +27,49 @@ impl r2d2::HandleError<r2d2::Error> for ErrorHandler {
...
+        }
+    }
+    fn handle_timeout(&self, event: e::TimeoutEvent) {
+        error!(self.logger, "Connection checkout timed out";
+            "wait_ms" => event.timeout().as_millis())
```
Jannis, line 66: nit: Here as well

lutter:
Fixed formatting and rebased to latest master

PR #1644, opened 2020-05-07
Revamp how we build containers

lutter:
These changes accomplish a few things:

* the `Dockerfile` produces three containers: a (very large) build container, a smaller `graph-ndoe` runtime container, and a `graph-node-dbg` debug container that can be used to investigate core dumps produced by the runtime container. The `docker/build.sh` script shows how to build the three containers using plain `docker` (resp `podman`) The `Dockerfile` expects that the current working directory for a build is the root of the repo
* the `start` script makes it possible to run an all-in-one `graph-node` (`ENV node_role=combined-node`, the default), or separate `query-node` or `index-node` containers. For `index-node`, it is possible to select which instance of such a container runs a block ingestor
* the `start` script will capture a core dump and other useful debugging information if the main process aborts for any reason as long as the directory `/var/lib/graph` exists. When starting a container, map that directory to a persistent directory on the host or via NFS to enable this feature
* the `graph-node-dbg` container also contains utility scripts that make it possible to create/deploy/reassign subgraphs that are already present on the IPFS server
* there is a `cloudbuild.yaml` that can be used to build containers in Google Cloud Build; in the future, we should also fix up the support for building images in Docker Hub (I'll need access to our account there to set that up)

The PR starts out by moving the container build to Alpine, but because of [this issue](https://andygrove.io/2020/05/why-musl-extremely-slow/) that causes Rust programs to run very slow under Alpine, a later commit moves the build back to Debian buster. The slowness was clearly observable both during container builds and in test environments indexing simple subgraphs.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -4,7 +4,7 @@
 # by running something like the following
 #   docker build --target STAGE -f docker/Dockerfile .
 
-FROM rust:alpine as graph-node-bld
+FROM rust:latest as graph-node-bld
 
 ARG COMMIT_SHA=unknown
 ARG REPO_NAME=unknown
@@ -13,12 +13,8 @@ ARG TAG_NAME=unknown
 
 ADD . /graph-node
 
-# See https://github.com/rust-lang/cargo/issues/7563 for why we use -crt-static
-# in the RUSTFLAGS
-RUN apk update \
- && apk add musl-dev openssl-dev postgresql-dev git \
- && cd /graph-node \
- && RUSTFLAGS="-g -C target-feature=-crt-static" cargo install --locked --path node \
+RUN cd /graph-node \
+ && RUSTFLAGS="-g" cargo install --locked --path node \
  && cargo clean \
  && objcopy --only-keep-debug /usr/local/cargo/bin/graph-node /usr/local/cargo/bin/graph-node.debug \
  && strip -g /usr/local/cargo/bin/graph-node \
@@ -32,7 +28,7 @@ RUN apk update \
  && echo "RUST_VERSION='$(rustc --version)'" >> /etc/image-info
 
 # The graph-node runtime image with only the executable
-FROM alpine:latest as graph-node
+FROM debian:buster-slim as graph-node
 ENV RUST_LOG ""
 ENV GRAPH_LOG ""
 ENV EARLY_LOG_CHUNK_SIZE ""
@@ -61,7 +57,8 @@ EXPOSE 8001
 # JSON-RPC port
 EXPOSE 8020
 
-RUN apk add libgcc libssl1.1 libcrypto1.1 postgresql-libs bash
+RUN apt-get update \
+ && apt-get install -y libpq-dev ca-certificates
 
 ADD docker/wait_for docker/start /usr/local/bin/
 COPY --from=graph-node-bld /usr/local/cargo/bin/graph-node /usr/local/bin
@@ -71,8 +68,8 @@ CMD start
 
 # Debug image to access core dumps
 FROM graph-node-bld as graph-node-dbg
-RUN apk add gdb postgresql-client bash curl
-ADD docker/debug /usr/local/bin
-COPY --from=graph-node-bld /etc/image-info /etc/image-info
+RUN apt-get update \
+ && apt-get install -y curl gdb postgresql-client
+
 COPY docker/Dockerfile /Dockerfile
-COPY bin/* /usr/local/bin
+COPY docker/bin/* /usr/local/bin/
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -51,6 +51,8 @@ ENV ethereum ""
 # The role the node should have, one of index-node, query-node, or
 # combined-node
 ENV node_role "combined-node"
+# The name of this node
+ENV node_id "default"
 
 # HTTP port
 EXPOSE 8000
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -6,6 +6,11 @@
 
 FROM rust:alpine as graph-node-bld
 
+ARG COMMIT_SHA=unknown
+ARG REPO_NAME=unknown
+ARG BRANCH_NAME=unknown
+ARG TAG_NAME=unknown
+
 ADD . /graph-node
 
 # See https://github.com/rust-lang/cargo/issues/7563 for why we use -crt-static
@@ -19,10 +24,12 @@ RUN apk update \
  && strip -g /usr/local/cargo/bin/graph-node \
  && cd /usr/local/cargo/bin \
  && objcopy --add-gnu-debuglink=graph-node.debug graph-node \
- && echo "repo: $(git config --get remote.origin.url)" > /image-info \
- && echo "commit: $(git describe --always)" >> /image-info \
- && echo "cargo: $(cargo --version)" >> /image-info \
- && echo "rust: $(rustc --version)" >> /image-info
+ && echo "repo:   $REPO_NAME" > /image-info \
+ && echo "tag:    $TAG_NAME" >> /image-info \
+ && echo "branch: $BRANCH_NAME" >> /image-info \
+ && echo "commit: $COMMIT_SHA" >> /image-info \
+ && echo "cargo:  $(cargo --version)" >> /image-info \
+ && echo "rust:   $(rustc --version)" >> /image-info
 
 # The graph-node runtime image with only the executable
 FROM alpine:latest as graph-node-run
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -71,7 +71,8 @@ CMD start
 
 # Debug image to access core dumps
 FROM graph-node-bld as graph-node-dbg
-RUN apk add gdb
+RUN apk add gdb postgresql-client bash curl
 ADD docker/debug /usr/local/bin
 COPY --from=graph-node-bld /etc/image-info /etc/image-info
 COPY docker/Dockerfile /Dockerfile
+COPY bin/* /usr/local/bin
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,23 +1,23 @@
 # Full build with debuginfo for graph-node
 FROM rust:alpine as graph-node-bld
 
-# Replace this with the graph-node branch you want to build the image from;
-# Note: Docker Hub substitutes this automatically using our hooks/post_checkout script.
-ENV SOURCE_BRANCH "master"
+ADD . /graph-node
 
 # See https://github.com/rust-lang/cargo/issues/7563 for why we use -crt-static
 # in the RUSTFLAGS
 RUN apk update \
  && apk add musl-dev openssl-dev postgresql-dev git \
- && git clone https://github.com/lutter/graph-node /graph-node \
  && cd /graph-node \
- && git checkout "$SOURCE_BRANCH" \
  && RUSTFLAGS="-g -C target-feature=-crt-static" cargo install --locked --path node \
  && cargo clean \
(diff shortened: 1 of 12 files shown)


docker/Dockerfile
```diff
@@ -1,27 +1,34 @@
-FROM rust:latest
+# Full build with debuginfo for graph-node
+#
+# The expectation if that the docker build uses the parent directory as PWD
+# by running something like the following
+#   docker build --target STAGE -f docker/Dockerfile .
 
-# Replace this with the graph-node branch you want to build the image from;
-# Note: Docker Hub substitutes this automatically using our hooks/post_checkout script.
-ENV SOURCE_BRANCH "master"
```
Jannis, line 5: I'm not sure why I wrote "our" `hooks/post_checkout` script, but Docker Hub I think needs the `SOURCE_BRANCH` environment variable to pick the branch / commit to build.

docker/Dockerfile
```diff
@@ -1,27 +1,34 @@
...
+ && objcopy --only-keep-debug /usr/local/cargo/bin/graph-node /usr/local/cargo/bin/graph-node.debug \
+ && strip -g /usr/local/cargo/bin/graph-node \
+ && cd /usr/local/cargo/bin \
+ && objcopy --add-gnu-debuglink=graph-node.debug graph-node \
+ && echo "REPO_NAME='$REPO_NAME'" > /etc/image-info \
+ && echo "TAG_NAME='$TAG_NAME'" >> /etc/image-info \
+ && echo "BRANCH_NAME='$BRANCH_NAME'" >> /etc/image-info \
+ && echo "COMMIT_SHA='$COMMIT_SHA'" >> /etc/image-info \
+ && echo "CARGO_VERSION='$(cargo --version)'" >> /etc/image-info \
+ && echo "RUST_VERSION='$(rustc --version)'" >> /etc/image-info
```
Jannis, line 28: Good idea!

docker/cloudbuild.yaml
```diff
@@ -0,0 +1,32 @@
...
+         '--build-arg', 'TAG_NAME=$TAG_NAME',
+         '-t', 'gcr.io/$PROJECT_ID/alp-graph-node:$SHORT_SHA',
+         '-f', 'docker/Dockerfile', '.']
+- name: 'gcr.io/cloud-builders/docker'
+  args: ['build', '--target', 'graph-node-dbg',
+         '--build-arg', 'COMMIT_SHA=$COMMIT_SHA',
+         '--build-arg', 'REPO_NAME=$REPO_NAME',
+         '--build-arg', 'BRANCH_NAME=$BRANCH_NAME',
+         '--build-arg', 'TAG_NAME=$TAG_NAME',
+         '-t', 'gcr.io/$PROJECT_ID/alp-graph-node-dbg:$SHORT_SHA',
```
Jannis, line 27: Is the `alph-` prefix in all these image tags desirable?

docker/start
```diff
@@ -0,0 +1,105 @@
...
+    graph-node \
+        --postgres-url "$postgres_url" \
+        --ethereum-rpc $ethereum \
+        --ipfs "$ipfs"
+}
+
+postgres_url="postgresql://$postgres_user:$postgres_pass@$postgres_host/$postgres_db"
+
+wait_for_ipfs "$ipfs"
+wait_for "$postgres_host:5432" -t 30
```
Jannis, line 84: Would be nice to wait for Postgres for 120 seconds as well. Postgres is the component that takes the longest to come up locally, so if this is the start script that people would use when testing a local Docker build, they'll run into this timeout randomly. The Postgres initialization is the reason for the 120s. 😉

docker/start
```diff
@@ -0,0 +1,105 @@
...
+        ;;
+    index-node)
+        start_index_node
+        ;;
+    combined-node)
+        start_combined_node
+        ;;
+    *)
+        echo "Unknown mode for start-node: $1"
+        echo "usage: start-node (query-node|index-node)"
```
Jannis, line 103: I guess `start-node` is `start` now?

docker/cloudbuild.yaml
```diff
@@ -0,0 +1,46 @@
...
+- name: 'gcr.io/cloud-builders/docker'
+  args: ['push', 'gcr.io/$PROJECT_ID/alp-graph-node:$SHORT_SHA']
+images:
+  - 'gcr.io/$PROJECT_ID/alp-graph-node-bld:$SHORT_SHA'
+  - 'gcr.io/$PROJECT_ID/alp-graph-node:$SHORT_SHA'
+  - 'gcr.io/$PROJECT_ID/alp-graph-node-dbg:$SHORT_SHA'
+secrets:
+ - kmsKeyName: projects/the-graph-staging/locations/global/keyRings/docker/cryptoKeys/docker-hub-push
+   secretEnv:
+     PASSWORD: 'CiQAdfFldbmUiHgGP1lPq6bAOfd+VQ/dFwyohB1IQwiwQg03ZE8STQDvWKpv6eJHVUN1YoFC5FcooJrH+Stvx9oMD7jBjgxEH5ngIiAysWP3E4Pgxt/73xnaanbM1EQ94eVFKCiY0GaEKFNu0BJx22vCYmU4'
```
Jannis, line 46: Is this safe to share in a public repo?

docker/Dockerfile
lutter: Now that we are building in Cloud Build, this is moot anyway

docker/cloudbuild.yaml
lutter: I just did that to avoid possible clashes with existing images. I've removed the `alp-` prefix now that things seem to be working

docker/start
lutter: Changed that

docker/start
lutter: It is indeed. Changed.

docker/cloudbuild.yaml
lutter: Yes, it's encrypted with the key stored in `kmsKeyName` which can not be retrieved

lutter: To add to this: it's an access token under my Docker Hub account, which we can revoke/replace at any time.

lutter:
This should be good to go now; the only thing still outstanding is to push images to the `graphprotocol/` org in Docker Hub rather than into my own personal namespace - I kept that there to make sure people don't see a change just yet until this is approved. I'll fix that up just before merging the PR by editing `docker/tag.sh`

docker/Dockerfile
Jannis: True

docker/build.sh
```diff
@@ -0,0 +1,14 @@
...
+# file should ultimately be moved to the corresponding commands for Cloud Build
+# in the cloudbuild.*.yaml files
+
+type -p podman > /dev/null && docker=podman || docker=docker
+
+cd $(dirname $0)/..
+
+for stage in graph-node-build graph-node graph-node-debug
+do
+    $docker build -t $stage --target $stage -f docker/Dockerfile .
```
Jannis, line 13: Should this also pass the `--build-arg`s in if the directory has a `.git` subdirectory?

docker/cloudbuild.yaml
```diff
@@ -0,0 +1,52 @@
+options:
+  machineType: "N1_HIGHCPU_32"
+timeout: 1800s
+steps:
+- name: 'gcr.io/cloud-builders/docker'
+  entrypoint: 'bash'
+  args: ['-c', 'docker login --username=lutter --password=$$PASSWORD']
```
Jannis, line 7: I'd make the username a variable too.

Jannis, line 7: What does the `$$` do here?

docker/cloudbuild.yaml
```diff
@@ -0,0 +1,52 @@
...
+  secretEnv: ['PASSWORD']
+- name: 'gcr.io/cloud-builders/docker'
+  args: ['build', '--target', 'graph-node-build',
+         '--build-arg', 'COMMIT_SHA=$COMMIT_SHA',
+         '--build-arg', 'REPO_NAME=$REPO_NAME',
+         '--build-arg', 'BRANCH_NAME=$BRANCH_NAME',
+         '--build-arg', 'TAG_NAME=$TAG_NAME',
+         '-t', 'gcr.io/$PROJECT_ID/graph-node-build:$SHORT_SHA',
```
Jannis, line 15: If we use `docker login`, does that log us into docker hub? Would it then be possible to use `-t graphprotocol/graph-node:$SHORT_SHA` and have it push to the docker hub registry?

docker/cloudbuild.yaml
```diff
@@ -0,0 +1,52 @@
...
+         '--build-arg', 'COMMIT_SHA=$COMMIT_SHA',
+         '--build-arg', 'REPO_NAME=$REPO_NAME',
+         '--build-arg', 'BRANCH_NAME=$BRANCH_NAME',
+         '--build-arg', 'TAG_NAME=$TAG_NAME',
+         '-t', 'gcr.io/$PROJECT_ID/graph-node-debug:$SHORT_SHA',
+         '-f', 'docker/Dockerfile', '.']
+- name: 'gcr.io/cloud-builders/docker'
+  args: ['tag',
+         'gcr.io/$PROJECT_ID/graph-node:$SHORT_SHA',
+         'lutter/graph-node:$SHORT_SHA']
```
Jannis, line 36: I'm assuming the `lutter` here is for testing? Or would you update the `graphprotocol/graph-node` tags on docker hub separately?

docker/tag.sh
```diff
@@ -0,0 +1,27 @@
...
+           lutter/graph-node:$tag
+    docker push lutter/graph-node:$tag
+
+    docker tag gcr.io/$PROJECT_ID/graph-node-debug:$SHORT_SHA \
+           lutter/graph-node-debug:$tag
+    docker push lutter/graph-node-debug:$tag
+}
+
+echo "Logging into Docker Hub"
+echo $PASSWORD | docker login --username=lutter --password-stdin
```
Jannis, line 17: Same questions here regarding the use `lutter` and using a variable for the username.

docker/cloudbuild.yaml
lutter: It keeps cloud build from expanding `$PASSWORD` since that's an environment variable passed at runtime. But this step isn't needed anymore anyway, since that all now happens in `tag.sh`. Removing this step.

docker/build.sh
lutter: I now tried my best to gather that info, though this file is really just meant to try things out locally

docker/cloudbuild.yaml
lutter: This step isn't needed anymore anyway, since that all now happens in tag.sh. Removing this step.

docker/cloudbuild.yaml
lutter: Yes, and that's how I had it initially, until I realized there's a little more logic around tagged builds that I don't think can be expressed in a Cloud Build file which is why I added `tag.sh`

docker/cloudbuild.yaml
lutter: No, this is just for testing, and I plan on doing `s/lutter/graphprotocol` in most places to make this the official way we build images.

docker/tag.sh
lutter: This one unfortunately has to stay, or, at least be _some_ user rather than just an organization; AFAICT, Docker Hub can only generate tokens for users, not for orgs and you have to log in as that user.

lutter:
Addressed all review comments

docker/cloudbuild.yaml
Jannis: Ok, fair enough

docker/cloudbuild.yaml
Jannis: Got it

PR #1645, opened 2020-05-08
Less expensive block polling

leoyvens:
This asyncifies most of the block ingestor and does the following functional changes, aiming at reducing the cost of block polling:
- Increase the default polling interval from 500ms to 1s. This is to reduce the overall number of requests made, though it will still add up to millions in month.
- Fetch only the header when checking for a new block, without full transactions. This is a cheaper call which suffices for most polls where we find no new block.
- A minor change: Use a delay at the end of the loop instead of a tokio interval stream. The interval stream would queue up triggers if the polling took longer than the interval duration, which could then lead to burst. Waiting a fixed amount after a poll has more predictable behavior.


Jannis (approved):
LGTM!

PR #1648, opened 2020-05-10
Speed up GraphQL query execution

lutter:
These two changes speed up GraphQL query execution by quite a bit by avoiding a number of clones.

Running an introspection query (the one GraphiQL issues when opening it against the metadata subgraph) used to take 180ms (measured on my machine with `--release`) at commit 606e460c. At commit a567a4f7 this had been reduced to 65ms because of various refactorings in GraphQL query execution. With the changes in this PR, that time goes down to 20ms


leoyvens:
Would it be simple to remove the `Clone` impl from the `ExecutionContext` and refactor the places that use it? From a quick look it seems doable.

lutter:
> Would it be simple to remove the `Clone` impl from the `ExecutionContext` and refactor the places that use it? From a quick look it seems doable.

Excellent point! I had another look, and the only reason we are cloning `ExecutionContext` is to maintain a path from the query root to the current field in `fields`. That path is not used for anything during query execution. I got rid of it, and removed `Clone` from `ExecutionContext`. It doesn't improve the runtime of my benchmark much (maybe by 1ms) but helps make the code clearer.

I also realized that the next batch of improvements for query execution should look at `collect_fields`: that currently gets run for every field in the _response_ even though it does not depend on any response values. We should switch that to run only once before execution starts, and use a data structure that makes it easy to look up whatever we need during execution rather than recomputing that a huge number of times.

PR #1658, opened 2020-05-12
Fetch genesis block by number 0 instead of 'earliest'

Jannis:
There are some implementations of Ethereum that don't support `eth_getBlockByNumber` requests where the requested block number is `"earliest"`. It seems safe to just request the block by number (`0`). At least I am not aware of any chains where the block number starts with.. `1` or `500`.

Reported by @fubhy.


leoyvens (approved):

leoyvens:
It is silly that `'earliest'` exists.

Issue #400, opened 2018-09-20
Find way to call contract methods by block hash, not block number


timmclean:
The `eth_call` in [web3](https://docs.rs/web3/0.4.0/web3/api/struct.Eth.html#method.call) and in the [Ethereum JSON-RPC spec](https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_call) has a block number as parameter. Block numbers do not reliably identify recent blocks due to the possibility of chain reorganizations. This will lead to incorrect data being returned when the Graph node and the Ethereum node have a different interpretation of what blocks are on the "main chain".

leoyvens:
This EIP should allow us to do this https://eips.ethereum.org/EIPS/eip-1898.

leoyvens:
Done in #1905.

PR #1662, opened 2020-05-14
Rebase ethabi and web3 forks on upstream

leoyvens:
The web3 fork also includes this patch https://github.com/tomusdrw/rust-web3/pull/345 which allows doing calls by block hash, but we don't do so yet because of issues with support for EIP 1898 by Ethereum node providers.

This is as far as we can go on #400 for now.


Jannis:
@leoyvens Wouldn't it be better to rebase our rust-web3 and ethabi `master` branches on top of upstream and then depend on these `master` branches instead of adding new branches?

Jannis:
It also looks like this PR is accidentally removing `Cargo.lock`.

leoyvens:
@Jannis I've updated the master branches. `Cargo.lock` is still there, it's just that these updates removed a bunch of transitive dependencies so the diff is very negative.

Jannis (approved):
Looks good to me. And yeah, 🤦 for thinking `Cargo.lock` was gone...

PR #1663, opened 2020-05-14
Remove `graph` parity-wasm dependency

leoyvens:
In preparation for the wasmtime switch, this makes the `graph` crate no longer depend on `parity-wasm` to parse modules, and instead treat them as a blob of bytes. This way all our wasm runtime dependencies are restricted to our runtime crate.

This makes an optimization to the `module_cache` so it uses the hash of the module as key. Though modules are typically only a few kb.


Jannis (approved):
Love it.

PR #1666, opened 2020-05-15
Graft by only copying data at the graft point

lutter:
Rather than grafting onto a subgraph by copying the entire source subgraph, graft by only copying the data that is valid at the block at which the graft is performed. This should greatly reduce the time that it takes to initialize a graft.


Jannis:
That's a great idea.

lutter:
Closing this for now - we'll have to revisit grafting more comprehensively, but truncating history might not be all that useful seeing how a lot of subgraph queries ask for historical data

PR #1671, opened 2020-05-15
notification listener: Exit process on panic

leoyvens:
Catch panics so we can exit the process and not just the thread. Disable whitespace diff to read it.


lutter:
Nice! This will make rollouts a lot less nerve-wrecking.

PR #1680, opened 2020-05-18
Collision resistant contract call cache

leoyvens:
The reason for using a hash as a key is to have a compact primary key, so I purposefully used a hash with only 128 bits of output when first implementing this, probably because I only considered protection against accidental collisions. However in the context of the network this could be targeted by a birthday attack, so we need a 256 bit secure hash to be unquestionably protected against that. This PR switches the hash to blake3 which is the same we use for PoI v2.

We don't want to lose the cache we have in the hosted service, so if the lookup by the new hash fails, we fallback to the old hash. If it's present in the old format, we set a new entry in the upgraded format and delete the old entry, making a smooth migration. Once this has been running in the hosted service for a few months and most of the old cache has been migrated, we can remove this fallback code.

A small patch to our web3 fork was necessary due to a dependency conflict with `blake3`. We'll be able to remove this patch once the problematic dependency `secp256k1` publishes a new version to crates.io.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -208,6 +208,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "blake3"
+version = "0.3.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "68df31bdf2bbb567e5adf8f21ac125dc0e897b1381e7b841f181521f06fc3134"
+dependencies = [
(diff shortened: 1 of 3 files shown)


That3Percent (approved):

lutter:
Look good to me. I like the gradual migration strategy, nice!

PR #1681, opened 2020-05-19
Handle database tables that have `text` columns where they should have enums

lutter:
We had a bug that if the GraphQL schema for a subgraph used an enum before it was defined, the generated database schema would have a 'text' column instead of an 'enum' column. Since that bug was (inadvertently) fixed in commit 3d23bccf, queries against such tables would fail if they tried to filter by the presumptive enum attribute.
    
With this commit, we check whether we have a text column where we should have an enum, and then switch the internal table representation to treat the enum as 'String' and not as an enum when querying or storing data.

Verifying that this actually fixes the problem is unfortunately a little complicated. [This gist](https://gist.github.com/lutter/70158472fc237b188e33ee8d6a1a201a) contains instructions on how to do that. I have also verified that use-before-declare of enums was not a problem before 3d23bccf (by running the test at commit 77e697b4)

### DIFF
--- a/store/postgres/examples/layout.rs
+++ b/store/postgres/examples/layout.rs
@@ -320,7 +320,7 @@ pub fn main() {
     let schema = ensure(fs::read_to_string(schema), "Can not read schema file");
     let schema = ensure(Schema::parse(&schema, subgraph), "Failed to parse schema");
     let catalog = ensure(
-        Catalog::new(db_schema.to_owned()),
+        Catalog::make_empty(db_schema.to_owned()),
         "Failed to construct catalog",
     );
     let layout = ensure(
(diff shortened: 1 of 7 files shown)


lutter:
> One request, shouldn't we include the `find_forward_enum_equal` test here?

We could, though it won't be testing this issue; nor would it have found the initial issue. As long as how we are generating the `Layout` is internally consistent, the test won't pick up anything and we are just repeating what `find_enum_equal` does already

lutter:
I have now also verified that this fixes the issue with production data by running the query that exposed this both against `master` and this branch, with the expected result (`master` returns an error, this branch returns a result)

lutter:
I just added the test and rebased to latest `master`

leoyvens:
> We could, though it won't be testing this issue; nor would it have found the initial issue.

I see. Can we assert that the layout column type is for an enum, so it would have at least caught the initial issue?

lutter:
> I just added the test and rebased to latest `master`

I got rid of this test again since it is pretty pointless, and added a unit test that checks that we actually create enum columns for use-before-declare enums.

Issue #1415, opened 2019-12-17
Improve `RevertRemoveQuery`


lutter:
This query currently finds the blocks that should be removed during a revert by filtering on `lower(block_range) >= $block` which forces a full table scan. This should be changed to `int4range($block, null) @> block_range` which can use the GIST index on `(id, block_range)`

For the block explorer, this brings the revert query from 7s (warm caches; 16s with cold caches) to ~ 2-3s. Reducing the storage size of the `id` of blocks will bring that down further.

lutter:
The way to make reverts fast is to use a [BRIN index](https://www.postgresql.org/docs/12/brin-intro.html) on `lower(block_range)` as that can exploit the fact that `lower(block_range)` never changes and increases monotonically as the table grows, i.e., there is close correlation between `lower(block_range)` and the physical location of the record on disk.

Doing some experimentation, the current slowness of reverts is due to the fact that we need to sift through a lot of data to find the records actually affected by the revert. And since most of that data is not in cache, most of the time for reverts is taken up by reading from disk. In its current form, the revert query does a full table scan; it's common for that to take 20s with cold caches, and then 500ms with a warm cache. But even switching to an index scan as suggested above shows similar behavior because of cache effects. A BRIN index effectively cuts down the number of blocks that need to be read from disk, especially for reverts where `lower(block_number)` is very recent and the index therefore very selective. The BRIN index reduces the number of blocks that need to be read often by one or two orders of magnitude.

lutter:
When we revert block `$block`, we run two queries for each table:

(1) Remove all entity versions that would now lie in the future:
```sql
delete from things where lower(block_range) >= $bock
```

(2) Undo deletes/updates for entity versions that happened at `$block`:
```sql
update things 
   set block_range = int4range(lower(block_range),null)
 where block_range @> $block
   and not block_range @> 2147483647
```

Both of these queries lead to full table scans, which can be extremely slow (over a minute for a table I tried all this with, but can be longer). Creating a [BRIN index](https://www.postgresql.org/docs/12/brin-intro.html)
```sql
create index things_brin 
    on things using brin(lower(block_range), coalesce(upper(block_range), 2147483647))
```
and rewriting the second query by adding redundant constraints that make the BRIN index usable for this query:
```sql
update things
   set block_range = int4range(lower(block_range),null)
 where block_range @> $block
   and not block_range @> 2147483647
   and lower(block_range) <= $block
[...]

PR #1683, opened 2020-05-20
Add BRIN indexes to speed up reverts

lutter:
The reasoning behind these changes is in https://github.com/graphprotocol/graph-node/issues/1415

### DIFF
--- a/store/postgres/src/relational.rs
+++ b/store/postgres/src/relational.rs
@@ -34,6 +34,7 @@ use graph::prelude::{
     format_err, info, BlockNumber, Entity, EntityChange, EntityChangeOperation, EntityCollection,
     EntityFilter, EntityKey, EntityOrder, EntityRange, EthereumBlockPointer, Logger,
     QueryExecutionError, StoreError, StoreEvent, SubgraphDeploymentId, Value, ValueType,
+    BLOCK_NUMBER_MAX,
 };
 
 use crate::block_range::{BLOCK_RANGE_COLUMN, BLOCK_UNVERSIONED};
@@ -1229,6 +1230,16 @@ impl Table {
             block_range = BLOCK_RANGE_COLUMN
         )?;
 
+        // Add a BRIN index on the block_range bounds to exploit the fact
+        // that block ranges closely correlate with where in a table an
+        // entity appears physically
+        write!(out,"create index brin_{table_name}\n    \
+                    on {schema_name}.{table_name}\n \
+                       using brin(lower(block_range), coalesce(upper(block_range), {block_max}), vid, id);\n",
+            table_name = self.name,
+            schema_name = layout.catalog.schema,
+            block_max = BLOCK_NUMBER_MAX)?;
+
         // Create indexes. Skip columns whose type is an array of enum,
         // since there is no good way to index them with Postgres 9.6.
         // Once we move to Postgres 11, we can enable that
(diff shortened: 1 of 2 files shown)


leoyvens:
Is possible and beneficial to have a BRIN index directly on the range? The PG docs only indicate that it is possible but not exactly how it works. I found [these docs](https://postgis.net/docs/using_postgis_dbmanagement.html#brin_indexes) for GIS types, which are more general but explain how BRIN works for operators like `@>`:

> The idea of a BRIN index is to store only the bouding box englobing all the geometries contained in all the rows in a set of table blocks, called a range. Obviously, this indexing method will only be efficient if the data is physically ordered in a way where the resulting bouding boxes for block ranges will be mutually exclusive. The resulting index will be really small, but will be less efficient than a GiST index in many cases.

This is exactly our case where rows are physically ordered in non-overlapping ranges. With an index directly on the range, the redundant clauses in the revert query might be necessary, if the planner can figure out that the BRIN index will be better than the GIST.

lutter:
> This is exactly our case where rows are physically ordered in non-overlapping ranges. With an index directly on the range, the redundant clauses in the revert query might not be necessary, if the planner can figure out that the BRIN index will be better than the GIST.

IIRC, I tried that and the planner didn't use the BRIN index on `block_range`, but I can try again just to make absolutely sure.

leoyvens:
@lutter Got it, it would be nice to comment on the BRIN index creation why each of those indexes may be worth it and the experience with trying to create one on the range itself.

lutter:
I just tried with a BRIN index on `(lower(block_range), block_range)` and Postgres is not using the index for the `update` query, and going back to a sequential scan. (We need to have `lower(block_range)` in the index anyway because we do a `delete from things where lower(block_range) > ...` to remove all versions now in the future)

leoyvens:
@lutter interesting, was that tried with the redundant clause on `lower(block_range)`? I can understand a BRIN index on the lower part working, I don't get how it works on the upper part.

Thinking again it actually makes sense that an index directly on the range shouldn't work, because we have a bunch of overlapping ranges. The ranges only follow the BRIN index conditions when looking at a single `id`.

lutter:
> Thinking again it actually makes sense that an index directly on the range shouldn't work, because we have a bunch of overlapping ranges. The ranges only follow the BRIN index conditions when looking at a single `id`.

@leoyvens I also suspect that the fact that we use ranges with an upper bound of `NULL` to indicate 'version is the most recent and valid for all future' has something to do with it since nulls in SQL are really strange beasts. In retrospect, it would have been better to use `i32::MAX` for this, but sadly, that ship has sailed. But that's the reason behind using the `coalesce`: to keep nulls out of the index.

leoyvens:
Exploiting BRIN for entity ids is interesting though I expect it to basically never work with our string ids since those are ordered alphanumerically. When we have autogenerated integer IDs, for which we can have a scheme such that higher blocks always have higher IDs, they should work great.

leoyvens:
@lutter With null or `i32::MAX`, I still don't expect a BRIN index to make sense for the upper part of the range, and consequentially it doesn't make sense for the entire range. Was that index on upper necessary in your tests?

lutter:
> Exploiting BRIN for entity ids is interesting though I expect it to basically never work with our string ids since those are ordered alphanumerically. When we have autogenerated integer IDs, for which we can have a scheme such that higher blocks always have higher IDs, they should work great.

Yes, putting `id` in the index is highly speculative, and I agree will mostly not be used, but since the index is so small, there's not much harm in doing that.

leoyvens:
Since indexing everything is already what we do, I wonder what subgraphs would benefit from BRIN indexes on integer fields, especially fields called `timestamp` or `block` :stuck_out_tongue:

lutter:
Just added a comment on some of the ideas behind using a BRIN index.

PR #1686, opened 2020-05-21
Round BigDecimals with too negative exp

leoyvens:
Numbers with too much precision after the dot can be rounded to an acceptable amount of precision. This should fix subgraphs with numbers which would otherwise not pass the validation due to a too negative exponent.


graph/src/data/store/scalar.rs
```diff
@@ -19,7 +19,15 @@ use std::str::FromStr;
 
 pub use num_bigint::Sign as BigIntSign;
 
-/// All operations on `BigDecimal` return a normalized value.
+/// These are the limits of IEEE-754 decimal128, a format we may want to switch to. See
+/// https://en.wikipedia.org/wiki/Decimal128_floating-point_format.
+pub const BIG_DECIMAL_MIN_EXP: i32 = -6143;
```
That3Percent, line 24: This is nit-picky, but these can be made associated constants and not pollute the already large prelude namespace:
```rust
impl BigDecimal {
    pub const MIN_EXP: i32 = -6143;
}

fn test() {
   let exp = BigDecimal::MIN_EXP;
}
```

tests/integration-tests/big-decimal/test/test.js
```diff
@@ -0,0 +1,77 @@
+const path = require("path");
```
That3Percent, line 1: I really like that we moved to integration tests, and that I consistently see new integration tests being added with your changes.

The amount of code that these require as compared to the size of the change is intimidating. Is there a way to move a lot of this into a library or something?

leoyvens: I carn try to move the contract, migration, and other files that repeat between many tests to a shared directory. The `waitForSubgraphToBeSynced` function can also be shared.

graph/src/data/store/scalar.rs
leoyvens: That really is nicer!

leoyvens:
@That3Percent Review addressed, I've moved contracts and migrations which are reused under `integration-tests/common`, which is a start to reduce duplication in tests.

leoyvens:
I've changed this to always round to 34 decimal digits (the test was also updated). This change may affect users, but it's hard to evaluate whether this rounding will have undesirable effects to a subgraph without knowing the requirements of the applications, so I suggest that we try it out and see if users report issues.

I also added normalization after the BigDecimal math operations.

graph/src/data/store/scalar.rs
```diff
@@ -71,20 +69,18 @@ impl BigDecimal {
         if self == &BigDecimal::zero() {
             return BigDecimal::zero();
         }
-        let (bigint, exp) = self.0.as_bigint_and_exponent();
+
+        // Round to the maximum significant digits.
+        let big_decimal = self.0.with_prec(Self::MAX_SIGNFICANT_DIGITS as u64);
```
That3Percent, line 74: When switching to decimal128 we may need to remember that this method has a specific rounding mode and we need to match it or make a version bump.

tests/integration-tests/big-decimal/src/mapping.ts
```diff
@@ -6,21 +6,27 @@ export function handleTrigger(event: Trigger): void {
...
       BigDecimal.fromString("9999999999999999999999999999999997").digits,
-    "fug " + big_decimal.digits.toString()
+    "wrong digits " + small.digits.toString()
   );
+
+  // This has 35 nines, but we round to 34 decimal digits.
+  let big = BigDecimal.fromString("99999999999999999999999999999999999");
+
+  // This has 35 zeros.
+  let rounded = BigDecimal.fromString("100000000000000000000000000000000000");
```
That3Percent, line 29: This test case doesn't prove as much as it could. It would be better to use `8` instead of `9` to demonstrate exactly where the rounding occurs.

graph/src/data/store/scalar.rs
```diff
@@ -115,23 +126,23 @@ impl Add for BigDecimal {
...
+        Self::from(self.0.add(other.0)).normalized()
     }
 }
 
 impl Sub for BigDecimal {
     type Output = Self;
 
     fn sub(self, other: Self) -> Self {
-        Self::from(self.0.sub(other.0))
+        Self::from(self.0.sub(other.0)).normalized()
```
That3Percent, line 137: `from` already normalizes - so all these math operations now normalize twice.

leoyvens: True, I got confused here and thought we weren't normalizing the output. Removed this commit.

graph/src/data/store/scalar.rs
leoyvens: Hopefully there isn't too much variation in how to round a decimal number, but versioning might be the best anyways to make sure we don't miss any subtle behavior difference.

tests/integration-tests/big-decimal/src/mapping.ts
leoyvens: I've added that test case.

leoyvens: I've added that test case.

PR #1691, opened 2020-05-26
store: Do not include 'id' in BRIN indexes

lutter:
Including the id column makes Postgres use this index when it really shouldn't and actually slows some queries down catastrophically from 50ms to 2s.


leoyvens:
There are tests that need fixing.

lutter:
> There are tests that need fixing.

Ugh .. of course, the unit tests that check the generated SQL needed an update.

PR #1693, opened 2020-05-28
Refactor the event listeners

lutter:
Each `graph-node` process would have one `ChainHeadUpdateListener` and one `StoreEventListener` for each network that the process was dealing with. With 8 networks, that would eat up 16 database connections, which adds up if you run a lot of `graph-node` processes.

With these changes, each process only uses one listener each, independent of the number of networks that are being processed, so that we only need 2 database connections per process.

Thanks to @leoyvens for helping me with some gnarly futures issues with the `ChainHeadUpdateListener`

### DIFF
--- a/core/src/subgraph/registrar.rs
+++ b/core/src/subgraph/registrar.rs
@@ -158,12 +158,13 @@ where
             .map(|event| {
                 // We're only interested in the SubgraphDeploymentAssignment change; we
                 // know that there is at least one, as that is what we subscribed to
-                stream::iter_ok(
-                    event
-                        .changes
-                        .into_iter()
-                        .filter(|change| change.entity_type == "SubgraphDeploymentAssignment"),
-                )
+                let assignments = event
+                    .changes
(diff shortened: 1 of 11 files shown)


lutter:
My preference would be for (2): even the fastest network generates new blocks every few seconds, which reduces the likelihood for a collision; not changing the code more reduces the risk that anything here goes wrong; and we already fall behind quite regularly in a lot of subgraphs (which is what would happen if we tripped over this) on reverts, so impact is relatively small

leoyvens:
That's fine by me, could you put that in a comment?

lutter:
I added a comment about the `watch::Receiver` behavior.

Looking at the code I also remembered that I had always planned to send events around as `Arc<StoreEvent>` instead of the full clone we do right now; that's particularly important as we do not filter on entity types until after we cloned. I added a commit that turns all streams of `StoreEvent` into streams of `Arc<StoreEvent>` to reduce cloning.

PR #1695, opened 2020-05-29
store: Properly escape single quotes in SafeString

lutter:


lutter:
I just verified with a copy of production that this does indeed fix the issue of not escaping single quotes in strings.

leoyvens:
Can we get a test?

PR #1703, opened 2020-06-03
Avoid adding 'order by' clauses for single-object lookups

lutter:
With a GraphQL schema like 
```graphql
type Parent {
  id: ID!
  child: Child
}
type Child { id: ID! }
```
the query that looks up children for a given parent would always include a clause `order by id` even though it makes no sense to ask for an order here: we know there is only one child per parent. We already check that there is only one by actually asking for two children, and generating an error if there are at least two.

A successful query would therefore only ever return one child per parent, which makes the `order by` clause superfluous, but can force Postgres to use a suboptimal query plan, since it is looking for query plans that also optimize for the requested sort operation.

This PR makes it so that queries for single-object relationships never contain an `order by` clause, which (hopefully) lets Postgres choose better query plans.

Right now, we have at least one query in production where the `order by` causes the query to take much longer than needed. On a clone of the production database, that query takes 2s with the `order by` clause, but only 500ms without it. It is hard to gauge the speedup in production since there that same query (with `order by`) can now take 20s and more.


graphql/src/store/query.rs
```diff
@@ -30,12 +36,19 @@ pub fn build_query<'a>(
...
     }
-    if let Some(order_by) = build_order_by(entity, arguments)? {
-        query = query.order_by_attribute(order_by);
-    }
-    if let Some(direction) = build_order_direction(arguments)? {
-        query = query.order_direction(direction);
-    }
+    let order = match (
+        dbg!(build_order_by(entity, arguments)?),
+        dbg!(build_order_direction(arguments)?),
```
leoyvens, line 41: You've got some `dbg!`s left behind.

graphql/src/store/query.rs
```diff
@@ -238,14 +251,15 @@ fn build_fulltext_order_by_from_object(
...
         .get(&"orderDirection".to_string())
-        .and_then(|value| match value {
-            q::Value::Enum(name) if name == "asc" => Some(EntityOrder::Ascending),
-            q::Value::Enum(name) if name == "desc" => Some(EntityOrder::Descending),
-            _ => None,
-        }))
+        .map(|value| match value {
+            q::Value::Enum(name) if name == "asc" => OrderDirection::Ascending,
+            q::Value::Enum(name) if name == "desc" => OrderDirection::Descending,
+            _ => OrderDirection::Ascending,
```
leoyvens, line 260: Should it not be an error if the `orderDirection` is present but something we don't recognize?

store/postgres/src/relational_queries.rs
```diff
@@ -1646,25 +1646,53 @@ impl<'a> FilterCollection<'a> {
 pub struct SortKey<'a> {
     column: Option<&'a Column>,
     value: Option<&'a str>,
-    direction: EntityOrder,
+    direction: &'static str,
+    /// Entirely skip creating an `order by` clause. The other fields of this
+    /// struct will be ignored if this is `true`
+    omit: bool,
```
leoyvens, line 1652: It's a bit strange to have an "ignore me" field in a struct. Would it be possible to represent this as `Option<SortKey>` in the `FilterQuery`?

store/postgres/src/relational_queries.rs
```diff
@@ -1646,25 +1646,53 @@ impl<'a> FilterCollection<'a> {
...
+            out.push_sql("order by ");
+            self.sort_expr(out)
+        }
+    }
+
+    /// Generate
+    ///   order by g$parent_id, [name direction], id
+    fn order_by_parent(&self, out: &mut AstPass<Pg>) -> QueryResult<()> {
+        if self.omit {
+            out.push_sql("order by g$parent_id ");
```
leoyvens, line 1685: This is confusing, not that I understand much about the window queries, but why is ordering by the parent id still important even `omit` was specified?

graphql/src/store/query.rs
lutter: I just followed what the code did before - the old `None` really ended up as `EntityOrder::Ascending` somewhere in the guts of the machinery. There's good reason to make that an error, as you say, but I didn't want to mess with that in this PR.

graphql/src/store/query.rs
lutter: Ugh .. removing them.

store/postgres/src/relational_queries.rs
lutter: Doing it that way would require that every caller checks whether they have a `SortKey` or `None`; I went this route to keep the code using `SortKey` more uniform and easier to follow. Another option would be to make `SortKey` a trait with a noop and a actual implementation, but I felt that was taking this a little too far. It is a bit strange, but the strangeness is all inside the `SortKey` implementation.

store/postgres/src/relational_queries.rs
lutter: Because it's only done for the outer part of queries where we already have collected actual children. I should probably check, but at least in some incarnation of `prefetch.rs`, the join logic expected children to come back sorted by parent, and that's another aspect of this logic that I didn't want to touch right now.

lutter:
Addressed review comments

store/postgres/src/relational_queries.rs
lutter: I just looked at the code that does the [join](https://github.com/graphprotocol/graph-node/blob/master/graphql/src/store/prefetch.rs#L372-L385) and it does not require that children are sorted by parent id. They do need to appear in the requested order for each parent, but in the `Unordered` case that doesn't matter. I'll change `order_by_parent`

store/postgres/src/relational_queries.rs
leoyvens: You could have
`
enum SortKey {
  None,
  Key {
   // fields
  }
}
```
and destructure it at the start of every method, returning early in case of `None`.

graphql/src/store/query.rs
leoyvens: Make sense to separate the changes, but now that we noticed it's a good opportunity to try to fix this, even if it risks breaking queries in the wild. Could you make a follow up PR for that?

store/postgres/src/relational_queries.rs
leoyvens: So in that case, `omit` is true even though the query is expected to return more than one row? And this makes sense because we are re-ordering the children which are already in the correct relative order for each parent?

PR #1708, opened 2020-06-09
Quick and dirty query cache

leoyvens:
Only for the subgraphs listed in the env var `GRAPH_CACHED_SUBGRAPH_IDS`, we keep a result cached for `GRAPH_CACHE_EXPIRY_SECS`. This might serve stale data but is useful as a way to temporarily alleviate pressure on the DB.

I did a manual test that this works as expected, logging `Query cache hit` on a second query, but expiring one minute after the first query.

### DIFF
--- a/graphql/src/execution/execution.rs
+++ b/graphql/src/execution/execution.rs
@@ -15,9 +15,57 @@ use crate::prelude::*;
 use crate::query::ast as qast;
 use crate::schema::ast as sast;
 use crate::values::coercion;
+use lru_time_cache::LruCache;
+use std::sync::Mutex;
 
 lazy_static! {
     static ref NO_PREFETCH: bool = std::env::var_os("GRAPH_GRAPHQL_NO_PREFETCH").is_some();
+
+    // Comma separated subgraph ids to cache queries for.
+    static ref CACHED_SUBGRAPH_IDS: Vec<String> = {
+        std::env::var("GRAPH_CACHED_SUBGRAPH_IDS")
+        .unwrap_or_default()
+        .split(',')
+        .map(|s| s.to_owned())
+        .collect()
+    };
+
+    // Cache expiry time, 1 minute by default.
+    static ref CACHE_EXPIRY: Duration = {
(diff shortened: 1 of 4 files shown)


graphql/src/execution/execution.rs
```diff
@@ -15,9 +15,57 @@ use crate::prelude::*;
...
+    let mut hasher = blake3::Hasher::new();
+    // subgraph id
+    hasher.update(ctx.query.schema.id.to_string().as_bytes());
+
+    // selection set
+    hasher.update(selection_set.to_string().as_bytes());
+
+    // variables
+    for (key, value) in &ctx.query.variables {
+        hasher.update(key.as_bytes());
```
That3Percent, line 58: This is not collision-resistant.

Consider, for example, a query variable with the key "KEY", and the value "VALUE". This hashes the same as a query variable with the key "KEYV" and the value "ALUE". In both cases, all that the hasher sees is the string "KEYVALUE", so the hash is the same.

We worked through all this and more in the `StableHash` crate. I would strongly consider using that which will help to get it right automatically. I can help with that effort if necessary.

graphql/src/execution/execution.rs
```diff
@@ -106,6 +154,20 @@ pub fn execute_root_selection_set(
...
+    if CACHED_SUBGRAPH_IDS.contains(&ctx.query.schema.id) {
+        // Check if the response is cached.
+        let mut cache = QUERY_CACHE.lock().unwrap();
+
+        // Remove expired entries.
+        cache.notify_iter();
+
+        // Peek because we want even hot entries to invalidate after the expiry period.
+        if let Some(response) = cache.peek(&cache_key(ctx, selection_set)) {
+            debug!(ctx.logger, "Query cache hit");
```
That3Percent, line 166: Consider removing this log. Logging in bulk can be a huge drain on performance. If this query is very hot this may be quite a lot of logs. A metric might be better.

leoyvens:
This is not collision resistant but we fix that in a follow up using stable hash. Thanks @That3Percent for pointing that out.

PR #1709, opened 2020-06-09
graphql: query hash collision resistance

That3Percent:


graphql/src/execution/execution.rs
```diff
@@ -156,15 +197,22 @@ pub fn execute_root_selection_set(
...
         // Check if the response is cached.
         let mut cache = QUERY_CACHE.lock().unwrap();
 
         // Remove expired entries.
         cache.notify_iter();
 
         // Peek because we want even hot entries to invalidate after the expiry period.
-        if let Some(response) = cache.peek(&cache_key(ctx, selection_set)) {
+        if let Some(response) = cache.peek(key.as_ref().unwrap()) {
             debug!(ctx.logger, "Query cache hit");
```
lutter, line 216: I second the comment from the previous PR: we should remove this log statement, since it can cause a lot of logging activity under load

That3Percent: Done. Logs: the (not so?) silent killer.

PR #1710, opened 2020-06-10
Fix issue inserting fulltext fields when all included field values are null

fordN:
This PR fixes an issue causing invalid insert queries to be constructed for fulltext fields.  

In the case of an insert of an entity with a fulltext field, an issue arose when all fields included in the fulltext field are null.  When walking the ast of the `QueryFragments` in an `InsertQuery` this led to an invalid query fragment, `())`, being constructed  [here](https://github.com/graphprotocol/graph-node/blob/master/store/postgres/src/relational_queries.rs#L367-L378).

### DIFF
--- a/store/postgres/src/relational_queries.rs
+++ b/store/postgres/src/relational_queries.rs
@@ -1049,12 +1049,17 @@ impl<'a> InsertQuery<'a> {
         for column in table.columns.iter() {
             match column.fulltext_fields.as_ref() {
                 Some(fields) => {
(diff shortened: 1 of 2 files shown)


Jannis:
Also, it looks like the formatting isn't up to date with the latest `rustfmt`.

fordN:
@Jannis I added a simple test of inserting an entity with null values for each of the fields included in its fulltext index and I updated the formatting using the latest version of `rustfmt`.

PR #1712, opened 2020-06-10
Optimize graphql execution

leoyvens:
This does a pass at simplifying and optimizing our graphql execution. Non-prefetched execution is removed. I couldn't get very amazing numbers locally, maybe a 10% speed up and a 20% resource usage reduction, but we'll see how it goes in real workloads.

The commits can be read separately. The first commit makes so `coerce` doesn't clone. The annoying part here is that it has to return the value on a failure to coerce so we can have good errors, so most of the diff is about the change to a `Result<Value, Value>` return type.

The second commit completes values in place so we don't allocate a new vec for it.

The third commit is probably the most impactful one. It removes non-prefetched execution. The optimization is instead of passing down a reference to the prefetched parent object, only so that the field being resolved is cloned later, we remove the field being resolved from the parent object in `execute_selection_set_to_map`, and then pass down the field value. This is more efficient and makes the code nicer overall.

### DIFF
--- a/graphql/src/execution/execution.rs
+++ b/graphql/src/execution/execution.rs
@@ -24,8 +24,6 @@ use std::sync::Mutex;
 type QueryHash = <SetHasher as StableHasher>::Out;
 
 lazy_static! {
-    static ref NO_PREFETCH: bool = std::env::var_os("GRAPH_GRAPHQL_NO_PREFETCH").is_some();
-
     // Comma separated subgraph ids to cache queries for.
     static ref CACHED_SUBGRAPH_IDS: Vec<String> = {
         std::env::var("GRAPH_CACHED_SUBGRAPH_IDS")
@@ -111,12 +109,6 @@ fn cache_key(ctx: &ExecutionContext<impl Resolver>, selection_set: &q::Selection
     stable_hash::<SetHasher, _>(&query)
 }
 
-#[derive(Clone, Copy, Debug, PartialEq)]
-pub enum ExecutionMode {
-    Prefetch,
-    Verify,
-}
-
 /// Contextual information passed around during query execution.
 pub struct ExecutionContext<R>
 where
@@ -136,8 +128,6 @@ where
 
     /// Max value for `first`.
     pub max_first: u32,
-
-    pub mode: ExecutionMode,
 }
 
 // Helpers to look for types and fields on both the introspection and regular schemas.
@@ -174,7 +164,6 @@ where
             query: self.query.as_introspection_query(),
             deadline: self.deadline,
             max_first: std::u32::MAX,
-            mode: ExecutionMode::Prefetch,
         }
     }
 }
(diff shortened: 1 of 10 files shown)


graphql/src/values/coercion.rs
```diff
@@ -117,55 +132,69 @@ pub(crate) fn coerce_input_value<'a>(
...
 
+/// On error, the `value` is returned as `Err(value)`.
 pub(crate) fn coerce_value<'a>(
-    value: &Value,
+    value: Value,
     ty: &Type,
     resolver: &impl Fn(&Name) -> Option<&'a TypeDefinition>,
     variable_values: &HashMap<q::Name, q::Value>,
-) -> Option<Value> {
+) -> Result<Value, Value> {
```
lutter, line 151: This change is a really nice way to avoid cloning just so you can generate an error!

graphql/src/execution/execution.rs
```diff
@@ -183,13 +172,7 @@ pub fn prefetch(
...
     selection_set: &q::SelectionSet,
 ) -> Result<Option<q::Value>, Vec<QueryExecutionError>> {
-    // Allow turning prefetch off as a safety valve. Once we are confident
-    // prefetching contains no more bugs, we should remove this env variable
-    if *NO_PREFETCH {
-        Ok(None)
-    } else {
-        ctx.resolver.prefetch(&ctx, selection_set)
-    }
+    ctx.resolver.prefetch(&ctx, selection_set)
```
lutter, line 175: Now that this is a one liner, you could just inline it in the one place where it is called.

graphql/src/execution/resolver.rs
```diff
@@ -72,69 +72,65 @@ pub trait Resolver: Clone + Send + Sync {
     /// Resolves entities referenced by a parent object.
     fn resolve_objects(
         &self,
-        parent: &Option<q::Value>,
+        objects_value: Option<q::Value>,
```
lutter, line 75: I liked that this used to be called `parent` - it helped me in initially understanding this code to know that this object lives one level above where we currently are. It would be good to have a comment here what exactly we expect in the `q::Value`. The dynamic typing via `q::Value` makes it easy to lose track of what you are looking at.

graphql/src/store/resolver.rs
```diff
@@ -356,144 +217,24 @@ where
...
-        if !is_derived {
-            Self::add_filter_for_reference_field(&mut query, parent, field_definition, object_type);
-        }
-
-        let mut entity_values = Vec::new();
-        for entity in self.store.find(query)? {
-            entity_values.push(entity.into())
-        }
-        Ok(q::Value::List(entity_values))
+        self.resolve_objects_prefetch(objects_value, field, object_type)
```
lutter, line 226: You could inline this here -  I only made that a separate method to help readability.

graphql/src/store/resolver.rs
```diff
@@ -356,144 +217,24 @@ where
...
-                        Some(q::Value::String(id)) => resolve_object_with_id(id)?,
-                        _ => None,
-                    },
-                    _ => panic!("top level queries must either take an `id` or return a list"),
-                }
-            }
-        };
-
-        Ok(entity.map_or(q::Value::Null, Into::into))
+        self.resolve_object_prefetch(object_value, field, field_definition, object_type)
```
lutter, line 237: You could inline this here -  I only made that a separate method to help readability.

PR #1716, opened 2020-06-12
Fix host functions in globals

leoyvens:
The first commit is the regression test, plus a pure refactor which renames `WasmInstanceHandle` -> `WasmInstance`, and pulls the `wasmtime::Instance` to it. The other rename is `WasmInstance` -> `WasmInstanceContext`. These names hopefully better reflect the purpose of each struct.

The second commit is where it gets wild. To support calling host functions in the Wasm start function so that AS can use them when initializing globals, we lazily initialize the `WasmInstanceContext` in the host export if necessary, taking from the `Caller` the memory and the memory allocation function. I needed to patch wasmtime to support that last part. But wasmtime would rather not merge that upstream and instead suggests that AS makes changes on their side, that discussion will happen here https://github.com/AssemblyScript/assemblyscript/issues/1333.

On the side, this also makes the timeout apply to the start function.

### DIFF
--- a/runtime/wasm/src/module/mod.rs
+++ b/runtime/wasm/src/module/mod.rs
@@ -330,63 +330,65 @@ impl WasmInstance {
 
         let modules = valid_module
             .import_name_to_modules
-            .get("store.get")
-            .into_iter()
-            .flatten();
-
-        for module in modules {
-            let func_shared_ctx = Rc::downgrade(&shared_ctx);
-            linker.func(module, "store.get", move |entity_ptr: u32, id_ptr: u32| {
-                let start = Instant::now();
-                let instance = func_shared_ctx.upgrade().unwrap();
-                let mut instance = instance.borrow_mut();
-                let instance = instance.as_mut().expect("called store.get in a global");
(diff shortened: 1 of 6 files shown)


leoyvens:
Merging for expediency but I'd still like @That3Percent to take a look.

runtime/wasm/src/module/mod.rs
```diff
@@ -35,43 +35,56 @@ mod test;
...
-        assert_eq!(Rc::strong_count(&self.instance), 1);
+        assert_eq!(Rc::strong_count(&self.instance_ctx), 1);
     }
 }
 
-impl WasmInstanceHandle {
+/// Proxies to the WasmInstanceContext.
+impl AscHeap for WasmInstance {
+    fn raw_new(&mut self, bytes: &[u8]) -> u32 {
+        let mut ctx = RefMut::map(self.instance_ctx.borrow_mut(), |i| i.as_mut().unwrap());
```
That3Percent, line 59: I'd never seen `RefMut::map`. Useful!

PR #1719, opened 2020-06-14
Remove "Graphql query served" log

leoyvens:
We already have a metric for this and also the `GRAPH_LOG_QUERY_TIMING` env var which can be turned on. This log is noisy and not useful, and I could even measure a throughput increase of about 5% in fast queries by removing it.


leoyvens:
I've checked and those are calculated with reports from the DB. It used to be that @fordN use these logs to calculate volume, but he gave the ok to remove this.

PR #1728, opened 2020-06-16
Fix duplicated scalar fields

leoyvens:
This was a regression in the optimization to not clone all fields before resolving them. The problematic case is when the same scalar field redundantly appears in multiple response keys, in which we really have to clone the field. But we don't want to reverse the optimization and go back to cloning everything, so this checks which fields are duplicated and clones only those.

### DIFF
--- a/core/tests/interfaces.rs
+++ b/core/tests/interfaces.rs
@@ -566,7 +566,17 @@ fn alias() {
                     parent: Legged
                   }";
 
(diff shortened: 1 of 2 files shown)


leoyvens:
@lutter It's a very good insight that the work for collecting fields is done redundantly for objects in lists, I'll have to try optimizing that in a follow up.

PR #1730, opened 2020-06-17
Reduce cloning in graphql execution

leoyvens:
The first commit removes `merge_selection_sets`, because it does an expensive clone in `extend_from_slice`. This procedure turns out to not be necessary if we make `collect_fields` directly handle a sequence of selection sets, merging them into the collected fields. Read with whitespace diff disabled.

The second commit removes some completely unecessary clones, found by running `cargo flamegraph`.

There are other ideas such as caching the `collect_fields` procedure, but as far as easy optimizations go these were all I could identify.

### DIFF
--- a/graphql/src/store/prefetch.rs
+++ b/graphql/src/store/prefetch.rs
@@ -493,7 +493,7 @@ fn execute_root_selection_set(
         items: Vec::new(),
     };
 
-    for (_, type_fields) in collect_fields(ctx, &query_type.into(), selection_set, None) {
+    for (_, type_fields) in collect_fields(ctx, &query_type.into(), vec![selection_set], None) {
         let fields = match type_fields.get(&TypeCondition::Any) {
             None => return Ok(vec![]),
             Some(fields) => fields,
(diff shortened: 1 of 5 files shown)


graphql/src/store/prefetch.rs
```diff
@@ -543,17 +543,17 @@ fn object_or_interface_by_name<'a>(
...
 }
 
-fn execute_selection_set(
+fn execute_selection_set<'a>(
     resolver: &StoreResolver<impl Store>,
-    ctx: &ExecutionContext<impl Resolver>,
+    ctx: &'a ExecutionContext<impl Resolver>,
     mut parents: Vec<Node>,
-    selection_set: &q::SelectionSet,
+    selection_sets: Vec<&'a q::SelectionSet>,
```
lutter, line 542: Just out of curiosity: why a `Vec` here instead of an iterator?

leoyvens: There's a story here, I got a `recursion limit reached` error from rustc when I tried using a generic here, which indicates it somehow lead to generating an infinitely recursive type or something. I could not understand why that error happened so I switched to vec which fixed it.

PR #1734, opened 2020-06-20
Optimize queries for single-object lookups

lutter:
When we query for children where we know that each parent can have at most one child, it is possible to simplify queries considerably. With this PR, we pass the necessary information to know whether we expect one or many children per parent into the store, and use it to generate better queries.

The background on why these queries look the way they do can be found in [the RFC](https://graphprotocol.github.io/rfcs/engineering-plans/0001-graphql-query-prefetching.html)

### DIFF
--- a/store/postgres/src/relational_queries.rs
+++ b/store/postgres/src/relational_queries.rs
@@ -1368,7 +1368,7 @@ enum TableLink<'a> {
 impl<'a> TableLink<'a> {
     fn new(child_table: &'a Table, link: EntityLink) -> Result<Self, QueryExecutionError> {
         match link {
-            EntityLink::Direct(attribute) => {
+            EntityLink::Direct(attribute, _) => {
                 let column = child_table.column_for_field(attribute.name())?;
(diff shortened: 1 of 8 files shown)


store/postgres/src/relational_queries.rs
```diff
@@ -1406,6 +1406,15 @@ impl<'a> ParentLimit<'a> {
         }
         Ok(())
     }
+
+    /// Include a 'limit {num_parents}+1' clause for single-object queries
+    /// if that is needed
+    fn single_limit(&self, num_parents: usize, out: &mut AstPass<Pg>) {
+        if let ParentLimit::Ranked(_, _) = self {
```
leoyvens, line 1413: Could this do a match on `ParentLimit`, for future-proofing, and give an explanation of why the `Outer` variant doesn't need this?

lutter: Good point. Added a commit that does that

lutter:
Rebase to latest master

PR #1743, opened 2020-06-22
Prefetch directly as `q::Value` instead of `Entity`s

leoyvens:
The purpose of this PR is to optimize the CPU usage of query nodes. There are no functional changes.

Currently, the conversions involved in serving a query result are:
jsonb (DB) -> json::Value (EntityData) -> Entity (prefetch) -> q::Value (graphql resolver)

After this PR it is:
jsonb (DB) -> json::Value (EntityData) -> q::Value (prefetch and graphql resolver)

Removing one round of conversion from `HashMap` to `BTreeMap`, which is heavy on allocations, and also avoids some conversions between `BigInt` and `Bytes` to string.

This make it so that the prefetch logic uses `q::Values` instead of entities, and abstracts a bit of the store so that it can serve queries as either `Vec<Entity>` or `Vec<q::Value>`. While `q::Value` has less information than `Entity`, we don't need the extra information in the query processing, and we need efficiency improvements here.

I expect this to have a large impact on the CPU usage of the query nodes, but we'll see.

### DIFF
--- a/graphql/src/store/prefetch.rs
+++ b/graphql/src/store/prefetch.rs
@@ -5,15 +5,14 @@ use graphql_parser::query as q;
 use graphql_parser::schema as s;
 use lazy_static::lazy_static;
 use std::collections::{BTreeMap, HashMap, HashSet};
-use std::ops::Deref;
 use std::rc::Rc;
 use std::time::Instant;
 
 use graph::data::graphql::ext::ObjectTypeExt;
 use graph::prelude::{
-    BlockNumber, ChildMultiplicity, Entity, EntityCollection, EntityFilter, EntityLink,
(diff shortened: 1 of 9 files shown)


graphql/src/store/prefetch.rs
```diff
@@ -87,7 +86,7 @@ impl From<Option<q::TypeCondition>> for TypeCondition {
 /// has an entry mapping the response key to the list of nodes.
 #[derive(Debug, Clone)]
 struct Node {
-    entity: Entity,
+    entity: BTreeMap<String, q::Value>,
```
lutter, line 89: We should really have a type name for this, since it's all over the codebase.

leoyvens: We could call this `QueryValueObject` or something. But I can never recall what a type alias stands for when reading, so personally I'd keep it as it is.

PR #1746, opened 2020-06-24
Allow to run against read-only Postgres replicas

Jannis:
This is useful when running a Graph Node against a Postgres read replica.

It disables:
- Chain head listeners.
- All indexing functionality.
- The JSON-RPC admin server.
- The GraphQL subscription server (which currently relies on Postgres `LISTEN`/`NOTIFY`, which doesn't work with replicas).

### DIFF
--- a/node/src/main.rs
+++ b/node/src/main.rs
@@ -259,6 +259,12 @@ async fn main() {
                 .value_name("URL")
                 .help("HTTP endpoint for 3box profiles"),
         )
+        .arg(
+            Arg::with_name("readonly-db")
+                .takes_value(false)
+                .long("readonly-db")
+                .help("Run against a readonly db (no subscriptions etc.)"),
+        )
         .get_matches();
 
     // Set up logger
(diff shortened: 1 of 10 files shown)


node/src/main.rs
```diff
@@ -561,11 +577,19 @@ async fn main() {
...
+            // database
+            if !readonly_db {
+                let subscription_server = GraphQLSubscriptionServer::new(
+                    &logger,
+                    graphql_runner.clone(),
+                    generic_store.clone(),
+                );
+
+                // Serve GraphQL subscriptions over WebSockets
+                graph::spawn(subscription_server.serve(ws_port));
```
That3Percent, line 591: I'm a little uncomfortable with spawning subscriptions here potentially introducing a regression. Why was this spawned last before this commit? It's probably ok if there was any dependency on ordering before then there was already a bug because there are no synchronization primitives to indicate this dependency resolving. Just a concern.

Jannis: There shouldn't be any race conditions here, no. I think we just spawned all the servers at the bottom because it looked... clean.
image

PR #1747, opened 2020-06-24
Better query cache

leoyvens:
The current one may serve stale data, since the expiry is time-based, and also has a bug where it wasn't evicting entries, and even if that was fixed the eviction would be slow because the used library did a linear scan to evict entries.

This revamps the query cache so that entries are cached by block pointer. It is designed with the assumption that most queries are for the latest block, so historical queries are not cached. This is implemented with a ring buffer of caches per block, the buffer has a configurable length corresponding to the amount of recent blocks we wish to keep cached. I suggest we start with length 1 when trying it in a production setting.

Because of the way the code is structured right now, subscriptions are no longer cached.

Adds a config option to turn on the cache globally, for all subgraphs.

### DIFF
--- a/graphql/src/execution/execution.rs
+++ b/graphql/src/execution/execution.rs
@@ -5,9 +5,10 @@ use lazy_static::lazy_static;
 use stable_hash::crypto::SetHasher;
 use stable_hash::prelude::*;
 use stable_hash::utils::stable_hash;
-use std::collections::{BTreeMap, HashMap, HashSet};
+use std::collections::{BTreeMap, HashMap, HashSet, VecDeque};
 use std::iter;
 use std::sync::atomic::AtomicBool;
+use std::sync::RwLock;
 use std::time::Instant;
 
 use graph::prelude::*;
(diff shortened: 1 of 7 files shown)


graphql/src/execution/execution.rs
```diff
@@ -244,10 +263,36 @@ pub fn execute_root_selection_set(
         )?);
     }
 
-    if let Some(key) = key {
-        // Insert into the cache.
-        let mut cache = QUERY_CACHE.lock().unwrap();
-        cache.insert(key, values.clone());
+    // Check if this query should be cached.
```
leoyvens, line 266: The code for writing to the cache is a bit tricky, and probably could be written in different, better ways. But as long as we can agree that the current form is correct, performs well even if there are micro-optimizations that could be made and reads ok, I'm happy with its current form.

graphql/src/execution/execution.rs
```diff
@@ -34,18 +40,22 @@ lazy_static! {
...
-        .unwrap()
+    static ref CACHE_ALL: bool = CACHED_SUBGRAPH_IDS.contains(&"*".to_string());
+
+    // How many blocks should be kept in the query cache. When the limit is reached, older blocks
+    // are evicted. This should be kept small since a lookup to the cache is O(n) on this value, and
+    // the cache memory usage also increases with larger number. Set to 0 to disable the cache,
+    // defaults to disabled.
+    static ref QUERY_CACHE_BLOCKS: usize = {
+        std::env::var("GRAPH_QUERY_CACHE_BLOCKS")
+        .unwrap_or("0".to_string())
```
That3Percent, line 51: I would set the default to 1. (can be future PR). Setting CACHED_SUBGRAPH_IDS and not getting a cache is surprising behavior.

leoyvens: Yes, this is envisioning a future where the default of `CACHED_SUBGRAPH_IDS` is `*`. So that this is an on/off switch for operators, manually adding ids is not fun.

leoyvens:
Tests passed locally, merging.

PR #1751, opened 2020-06-25
Shared queries

That3Percent:


graphql/src/execution/cache.rs
```diff
@@ -0,0 +1,240 @@
...
+        }
+    }
+    /// Assumption: Whatever F is passed in consistently returns the same
+    /// value for any input - for all values of F used with this Cache.
+    pub fn cached_query<F: FnOnce() -> R>(&self, hash: Hash, f: F) -> CachedResponse<R> {
+        // This holds it's own lock so make sure that this happens outside of
+        // holding any other lock.
+        let cleanup = self.cleanup.take();
+
+        let mut cache = self.cache.lock().unwrap();
```
leoyvens, line 193: We should release this lock before running `f`!

graphql/src/execution/execution.rs
```diff
@@ -53,9 +58,25 @@ lazy_static! {
         .expect("Invalid value for GRAPH_QUERY_CACHE_BLOCKS environment variable")
     };
 
-    // News blocks go on the front, so the oldest block will be at the back.
+    // New blocks go on the front, so the oldest block will be at the back.
     // This `VecDeque` works as a ring buffer with a capacity of `QUERY_CACHE_BLOCKS`.
     static ref QUERY_CACHE: RwLock<VecDeque<CacheByBlock>> = RwLock::new(VecDeque::new());
+    static ref QUERY_HERD_CACHE: QueryCache<QueryResponse> = QueryCache::new();
```
leoyvens, line 64: :cow: :cow: :cow:

graphql/src/execution/execution.rs
```diff
@@ -178,46 +199,17 @@ where
             query: self.query.as_introspection_query(),
             deadline: self.deadline,
             max_first: std::u32::MAX,
-            cached: AtomicBool::new(false),
+            cached: AtomicBool::new(true),
```
leoyvens, line 202: Don't get the changes on `cached` in this PR.

graphql/src/execution/execution.rs
```diff
@@ -178,46 +199,17 @@ where
...
-                        return Ok(response.clone());
-                    }
-                }
-
-                key = Some(cache_key);
-            }
-        }
-    }
+) -> QueryResponse {
+    ctx.cached.store(false, std::sync::atomic::Ordering::SeqCst);
```
leoyvens, line 212: Don't get the changes on `cached` in this PR.

graphql/src/execution/execution.rs
```diff
@@ -178,46 +199,17 @@ where
...
             max_first: std::u32::MAX,
-            cached: AtomicBool::new(false),
+            cached: AtomicBool::new(true),
         }
     }
 }
 
-/// Executes the root selection set of a query.
-pub fn execute_root_selection_set(
+pub fn execute_root_selection_set_uncached(
```
leoyvens, line 207: Nice, thanks for cleaning up the mix of cache code and execution code.

graphql/src/execution/cache.rs
That3Percent: Excellent catch!

graphql/src/execution/execution.rs
That3Percent: The `QueryCache` isn't aware of the `ctx` and therefore doesn't set `ctx.cached`. Rather than make every cache layer aware of this variable just move the setting of cached to false in one spot where the query is in fact executed.

Jannis:
The commit messages make me think a rebase to clean them up would be a good idea.

That3Percent:
> The commit messages make me think a rebase to clean them up would be a good idea.

That is the idea. It's just this way to make review easier.

That3Percent:
> We'll need to test this properly before rolling it out. The changes (especially the inner synchronization in `QueryCache` and the interplay between the two caches) are complex enough that there may be bugs hidden in there.

I agree that this is quite complex.

I did a fair amount of `dbg!()` testing and adding calls to `std::thread::sleep_ms()` to ensure that various expected conditions were met in regards to the cache interplay, non-reentrancy, query uniqueness, and keeping the memory utilization within expected bounds. With more time we can definitely test more though. Pushing right now is probably not urgent.

graphql/src/execution/execution.rs
leoyvens: Ah makes sense. The comment on the declaration of this field should be updated.

graphql/src/execution/execution.rs
```diff
@@ -305,32 +314,37 @@ pub fn execute_root_selection_set(
...
-            // Otherwise this is a historical query which will not be cached.
-            let should_insert = match cache.iter().next() {
-                None => true,
-                Some(highest) if highest.block.number <= block_ptr.number => true,
-                Some(_) => false,
-            };
+        // Share errors from the herd cache, but don't store them in generational cache.
+        // In particular, there is a problem where asking for a block pointer beyond the chain
+        // head can cause the legitimate cache to be thrown out.
+        if cached.is_ok() {
```
leoyvens, line 320: Style nit: We could avoid the extra indentation by moving this as an `Ok(_)` pattern match in the `if let` above.

PR #1752, opened 2020-06-25
Log the calculated complexity of each query executed 

fordN:
Adds a `complexity` field to the `Query` object and includes it in the query execution logging.

### DIFF
--- a/graphql/src/execution/query.rs
+++ b/graphql/src/execution/query.rs
@@ -42,6 +42,7 @@ pub struct Query {
     /// have dummy values
     pub(crate) query_text: Arc<String>,
     pub(crate) variables_text: Arc<String>,
+    pub(crate) complexity: u64,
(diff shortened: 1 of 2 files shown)


fordN:
I could avoid using the mutable reference by converting `complexity()` to an anonymous function that gets run before instantiating the query object...or shadowing the query object.  Thoughts on whether using the mutable reference to query here will degrade performance?

graphql/src/execution/query.rs
```diff
@@ -161,6 +163,7 @@ impl Query {
             kind: self.kind,
             query_text: self.query_text.clone(),
             variables_text: self.variables_text.clone(),
+            complexity: self.complexity.clone(),
```
leoyvens, line 166: `complexity` is `Copy`, no need for a clone

PR #1753, opened 2020-06-25
Revert experimental read only

leoyvens:
We're now going for every query node having read access, and optionally having additional access to a read replica.


leoyvens:
Tests pass locally, merging.

PR #1754, opened 2020-06-26
Support read replicas

leoyvens:
This adds a configuration to pass in a sequence of postgres urls for replicas, which is turned into a sequence of DB connections pools passed to each store. In the store, each use of a connection is either in `ConnMode::Read` or `ConnMode::Write`. Writes always use the main DB. Reads are distributed between all available DBs, including the main one.

### DIFF
--- a/store/postgres/src/query_store.rs
+++ b/store/postgres/src/query_store.rs
@@ -1,0 +1,48 @@
+use std::collections::BTreeMap;
+
+use crate::store::ReplicaId;
+use graph::components::store::QueryStore as QueryStoreTrait;
+use graph::prelude::{Store as _, *};
+
+pub(crate) struct QueryStore {
+    replica_id: ReplicaId,
+    store: Arc<crate::Store>,
+    for_subscription: bool,
+}
+
+impl QueryStore {
+    pub(crate) fn new(store: Arc<crate::Store>, for_subscription: bool) -> Self {
+        // Subscriptions always go to the main replica.
+        let replica_id = match for_subscription {
+            false => store.pick_replica(),
+            true => ReplicaId::Main,
+        };
+        QueryStore {
+            replica_id,
+            store,
+            for_subscription,
+        }
+    }
+}
+impl QueryStoreTrait for QueryStore {
+    fn find_query_values(
+        &self,
+        query: EntityQuery,
+    ) -> Result<Vec<BTreeMap<String, graphql_parser::query::Value>>, QueryExecutionError> {
+        let conn = self
+            .store
+            .get_entity_conn(&query.subgraph_id, self.replica_id)
+            .map_err(|e| QueryExecutionError::StoreError(e.into()))?;
+        self.store.execute_query(&conn, query)
+    }
+
+    fn subscribe(&self, entities: Vec<SubgraphEntityPair>) -> StoreEventStreamBox {
+        assert!(self.for_subscription);
+        assert_eq!(self.replica_id, ReplicaId::Main);
+        self.store.subscribe(entities)
+    }
+
+    fn is_deployment_synced(&self, id: SubgraphDeploymentId) -> Result<bool, Error> {
+        self.store.is_deployment_synced(id)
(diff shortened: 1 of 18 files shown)


Jannis:
@leoyvens I've tested this with https://github.com/Jannis/docker-postgres-replication#changes-compared-to-danieldentmaster.

Steps:
1. In `docker-postgres-replication` repo: `docker build .`, copy image hash into `docker-compose.yml`
2. Anywhere: Run `createdb -h localhost -p 5432 -U postgres graph` (password: `postgres`)
3. In `graph-node` repo: 
    ```sh
cargo watch -x "run -p graph-node -- --debug --ethereum-rpc mainnet:https://eth-mainnet.alchemyapi.io/v2/iemhQFg2k89zAO1-gSngWqkLHTZ-Byc_ --ipfs "127.0.0.1:5001" --postgres-url postgresql://postgres:postgres@localhost:5432/graph --postgres-secondary-hosts postgresql://postgres:postgres@localhost:5433/graph"
    ```
4. Deploy a subgraph to this and send queries and subscriptions to `graph-node`.

I also added a `dbg!` statement to the code in this branch to verify that it's picking a different connection pool with every query. I can confirm that `graph-node` connects to both stores
```rust
Jun 29 13:10:27.215 INFO Connected to Postgres, url: postgresql://postgres:HIDDEN_PASSWORD@localhost:5432/graph, component: Store
Jun 29 13:10:27.294 INFO Connected to Postgres, url: postgresql://postgres:HIDDEN_PASSWORD@localhost:5433/graph, component: Store
```
and sends queries to the primary and secondary Postgres instances fine. Subscriptions also work, and indexing as well.

leoyvens:
This has been revamped to be more careful with the use of replicas. `ConnMode` has been refactored to `ReplicaId`. Generally `ReplicaId::Main` is used, except for queries. Each query now utilizes a `QueryStore`, which picks a replica and utilizes it throughout the query, guaranteeing a consistent view of the database. Subscriptions use the main db server.

store/postgres/src/query_store.rs
```diff
@@ -0,0 +1,48 @@
...
+    replica_id: ReplicaId,
+    store: Arc<crate::Store>,
+    for_subscription: bool,
+}
+
+impl QueryStore {
+    pub(crate) fn new(store: Arc<crate::Store>, for_subscription: bool) -> Self {
+        // Subscriptions always go to the main replica.
+        let replica_id = match for_subscription {
+            false => store.pick_replica(),
```
lutter, line 17: I find this back-and-forth between `Store` and `QueryStore` a little strange; it would be more straightforward to make `QueryStore::new` part of `Store.pick_replica` so that this whole logic is in one place. All of this logic could probably go into `Store::query_store`

leoyvens: You're saying we can remove `QueryStore::new` by inlining it in `Store::query_store`? Makes sense. I suppose I can then also do the same with `pick_replica`.

leoyvens:
@lutter I've addressed your comment by putting all of the logic regarding choosing the replica in `Store::query_store`.

store/postgres/src/store.rs
```diff
@@ -1346,7 +1333,28 @@ impl StoreTrait for Store {
         self: Arc<Self>,
         for_subscription: bool,
     ) -> Arc<(dyn QueryStore + Send + Sync + 'static)> {
-        Arc::new(crate::query_store::QueryStore::new(self, for_subscription))
+        use std::sync::atomic::Ordering;
+
+        let replica_id = match for_subscription {
```
lutter, line 1338: This is a bit nitpicky, but I would have written `if for_subscription { .. } else { .. }` here.

leoyvens: The match does look nicer when the branches are single liners, but still I like being able to put the `false` case first when it is the common case, without using `!` to negate the condition.

leoyvens:
@lutter thanks for the review!

PR #1771, opened 2020-07-08
graph: Turn jailing of queries off by default

lutter:


lutter:
@leoyvens I added another commit before I saw that you had approved this already. I just re-requested a review for that second commit.

graph/src/data/graphql/effort.rs
```diff
@@ -92,11 +94,8 @@ impl QueryEffort {
     /// for the query
     pub fn current_effort(&self, shape_hash: u64) -> (Option<Duration>, Duration) {
         let inner = self.inner.read().unwrap();
-        let total_effort = inner.total.average().unwrap_or(*ZERO_DURATION);
```
leoyvens, line 95: An improvement unrelated to this commit: The constructors for `Duration` are `const`, so I believe you can pull some `Duration` constants out of `lazy_static` and into an ordinary constant.

lutter: Ooh .. nice; I didn't realize that. Added another commit to address that.

PR #1772, opened 2020-07-09
Remove various log statements from the store

lutter:
These log statements are either already captured by Prometheus metrics by themselves, or are not very useful on their own.

### DIFF
--- a/store/postgres/src/store.rs
+++ b/store/postgres/src/store.rs
@@ -125,10 +125,7 @@ fn initiate_schema(logger: &Logger, conn: &PgConnection, blocking_conn: &PgConne
         // run, which is not easy to tell from the Diesel API, and reset the
         // query statistics since a schema change makes them not all that
         // useful. An error here is not serious and can be ignored.
-        let res = conn.batch_execute("select pg_stat_statements_reset()");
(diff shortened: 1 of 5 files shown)


store/postgres/src/store.rs
```diff
@@ -1619,9 +1615,6 @@ impl EthereumCallCache for Store {
             {
                 use crate::db_schema::eth_call_cache::dsl;
 
-                // Once we stop getting these logs, this code can be removed.
-                trace!(self.logger, "Updating eth call cache entry");
```
leoyvens, line 1623: We should only remove this log if we remove the call cache migration (this `else` branch). Which I think we're ready to do, I fund only very few of these in the recent hosted service logs.

lutter: I don't understand why this log is useful - what does it tell us?

leoyvens: It tell us that there is some entry in the old format that's being updated to the new one. As long as this is happening for some entries, we want to keep this fallback.

lutter: Ok. I've put that statement back in and rebased the PR against latest master

PR #1774, opened 2020-07-09
Correctly resolve overload situation

lutter:
Erroneous logic in how we step down the `kill_rate` caused it to never fully go to 0

### DIFF
--- a/server/websocket/src/connection.rs
+++ b/server/websocket/src/connection.rs
@@ -325,10 +325,19 @@ where
                             // Send errors back to the client as GQL_DATA
                             match e {
                                 SubscriptionError::GraphQLError(e) => {
-                                    let result = Arc::new(QueryResult::from(e));
-                                    let msg =
-                                        OutgoingMessage::from_query_result(err_id.clone(), result);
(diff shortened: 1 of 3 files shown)


server/websocket/src/connection.rs
```diff
@@ -325,10 +325,18 @@ where
...
                             match e {
                                 SubscriptionError::GraphQLError(e) => {
-                                    let result = Arc::new(QueryResult::from(e));
-                                    let msg =
-                                        OutgoingMessage::from_query_result(err_id.clone(), result);
-                                    error_sink.unbounded_send(msg.into()).unwrap();
+                                    // Don't bug clients with transient `TooExpensive` errors,
+                                    // simply skip updating them
+                                    if e.iter().all(|err| {
+                                        !matches!(err, QueryExecutionError::TooExpensive)
```
leoyvens, line 331: This is better written as `if !e.iter.any(|err| { matches!(...`.

lutter: Yep; I though `!matches!` looked funny, but didn't think of writing it that way. Fixed that and rebased to latest master

PR #1779, opened 2020-07-10
More accurate logging of load manager actions

lutter:
Previously, we logged `Overload resolved` when we were merely stepping down the kill rate.  Also, back off killing queries twice as fast as we ratchet up during an overload

### DIFF
--- a/graph/src/data/graphql/effort.rs
+++ b/graph/src/data/graphql/effort.rs
@@ -43,8 +43,6 @@ lazy_static! {
             .unwrap_or(1e9)
     };
 
-    static ref WAIT_STAT_CHECK_INTERVAL: Duration = Duration::from_secs(1);
-
     // Load management can be disabled by setting the threshold to 0. This
     // makes sure in particular that we never take any of the locks
     // associated with it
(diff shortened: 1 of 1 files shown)


graph/src/data/graphql/effort.rs
```diff
@@ -389,18 +394,24 @@ impl LoadManager {
             if overloaded {
                 kill_rate = (kill_rate + KILL_RATE_STEP * (1.0 - kill_rate)).min(1.0);
             } else {
-                kill_rate = (kill_rate - KILL_RATE_STEP).max(0.0);
+                kill_rate = (kill_rate - 2.0 * KILL_RATE_STEP).max(0.0);
```
leoyvens, line 397: Do we want to document somewhere that the rate of decay is `2*KILL_RATE_STEP`? Maybe make this its own constant?

lutter: There's a whole section of the `graph-node` manual that needs to be written about load management and how to use and tune it; I'll try and write some of that up next week. I don't want to bug people with another tuneable, since I view tuneables largely as us saying "we couldn't figure it out so we'll have you do that". But I am also not sure whether stepping down at twice `KILL_RATE_STEP` is the right way to go anyway - it's a tradeoff between stopping to drop queries quickly when an overload is brief vs settling on the 'right' kill rate when the overload is prolonged. Thinking about this more, I wonder if stepping down too quickly could be problematic during the latter.

leoyvens: I wasn't necessarily suggesting to make this configurable, just to extract this to a constant so you can have a name and a comment for it.

lutter: I totally misunderstood - I turned that factor into a constant, pulled all the constants used in this calculation into `update_kill_rate` and added comments about the calculations in the code.

graph/src/data/graphql/effort.rs
```diff
@@ -381,26 +381,52 @@ impl LoadManager {
...
+        //
+        // The idea behind this is that we want to be conservative when we drop
+        // queries, but aggressive when we reduce the amount of queries we drop
+        // to disrupt traffic for as little as possible.
+        const KILL_RATE_STEP_UP: f64 = 0.1;
+        const KILL_RATE_STEP_DOWN: f64 = 2.0 * KILL_RATE_STEP_UP;
+
+        lazy_static! {
+            static ref KILL_RATE_UPDATE_INTERVAL: Duration = Duration::from_millis(1000);
+        }
```
leoyvens, line 402: This doesn't need `lazy_static`, I believe it can be an ordinary constant.

lutter: True; fixed that.

PR #1783, opened 2020-07-14
Refactor servers, log graphql panics

leoyvens:
First this refactors the http servers, removing the types `GraphQLResponse` and `IndexNodeResponse` which were largely redundant with `QueryResult`. There were minor functional changes, such as malformed graphql queries returning just an error string instead of a graphql error, but I don't think this is too important. This also async-ifies the servers a bit more.

We also refactor `block_in_place` into `spawn_blocking`, this made it easier to make the main change, which is to catch and log panics during query execution.


graphql/src/execution/query.rs
```diff
@@ -60,16 +60,9 @@ impl Query {
...
-                    .document
-                    .format(&Style::default().indent(0))
-                    .replace('\n', " "),
-                serde_json::to_string(&query.variables).unwrap_or_default(),
-            )
-        } else {
-            ("(gql logging turned off)".to_owned(), "".to_owned())
+        let (query_text, variables_text) = match *graph::log::LOG_GQL_TIMING {
+            true => (query.query_text(), query.variables_text()),
+            false => ("(gql logging turned off)".to_owned(), "".to_owned()),
```
lutter, line 65: Nice factoring out the details of formatting queries, but I feel this would be more readable as an if statement rather than as a match

server/http/src/service.rs
```diff
@@ -272,9 +272,41 @@ where
         let query = GraphQLRequest::new(body, schema, network).compat().await;
 
         let result = match query {
-            Ok(query) => graph::spawn_blocking_allow_panic(service.graphql_runner.run_query(query))
-                .await
-                .unwrap(),
+            Ok(query) => {
+                // Not great that we serialize the query here only for the panic case.
+                let query_text = query.query_text();
+                let variables_text = query.variables_text();
```
lutter, line 278: This to me is a big problem as we will wind up formatting each query twice (once here, once for logging) Could we have `execute_query` add the query text it already has on unwind and just log that?

Alternatively, we could construct the `execution::Query` here already and pass it into the runner, though that might be a pretty intrusive change

server/http/src/service.rs
```diff
@@ -272,9 +272,41 @@ where
...
+
+                let result =
+                    graph::spawn_blocking_allow_panic(service.graphql_runner.run_query(query))
+                        .await;
+
+                match result {
+                    Ok(res) => res,
+
+                    // `JoinError` means a panic.
```
lutter, line 287: This comment seems like it was left over from a previous incarnation of the code

graphql/src/execution/query.rs
leoyvens: I like using `match` because it's one line shorter than `if`, but sure I can make this an `if` as it was originally.

server/http/src/service.rs
leoyvens: The type of `e` here is `JoinError`. I'll make this `Err(JoinError)` in the comment for clarity.

server/http/src/service.rs
leoyvens: It's unfortunate but generally the cost of processing of query inputs is small compared to processing large query outputs. But I'll see about moving `execution::Query` into `graph` so we can construct it directly.

leoyvens:
@lutter It turned out to be easy to avoid duplicating the query text, I just moved it to the `Query` in `graph`. I've addressed your comments.

PR #1784, opened 2020-07-15
Add env variable options for sensitive arguments

adzialocha:
This PR adds the option to pass sensitive arguments like database passwords to the application without leaking them in bash history, docker logs or process managers (like `htop` / `top`).

Related discussion: https://github.com/graphprotocol/graph-node/issues/849#issuecomment-658731288

### DIFF
--- a/node/src/main.rs
+++ b/node/src/main.rs
@@ -109,6 +109,7 @@ async fn main() {
                 .takes_value(true)
                 .long("subgraph")
                 .value_name("[NAME:]IPFS_HASH")
+                .env("SUBGRAPH")
                 .help("name and IPFS hash of the subgraph manifest"),
         )
         .arg(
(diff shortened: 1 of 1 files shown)


node/src/main.rs
```diff
@@ -128,6 +130,7 @@ async fn main() {
                 .conflicts_with_all(&["ethereum-ws", "ethereum-ipc"])
                 .long("ethereum-rpc")
                 .value_name("NETWORK_NAME:URL")
+                .env("ETHEREUM_RPC")
```
Jannis, line 133: Can we add the sae for `--ethereum-ws` and `--ethereum-ipc` (that's usually local, so perhaps not important).

Jannis: Should probably also add an environment variable for `--ipfs`.

adzialocha:
@Jannis :+1:  added!

PR #1796, opened 2020-07-20
Fix subscription semaphore

leoyvens:
This currently isn't doing anything because we were not `await`ing for the semaphore.


leoyvens:
I'm working on using a single semaphore for subscriptions and queries so this won't matter.

PR #1797, opened 2020-07-20
Resolve subgraph names in a blocking task

leoyvens:
This is a relatively cheap DB operation, but at the volume of thousands of queries per second this could clog the tokio threadpool.


leoyvens:
Thanks for the review

Issue #1405, opened 2019-12-12
Address certain corner cases of time-travel queries by block hash


lutter:
Time-travel queries introduced in PR #1397 might still return incorrect results in certain circumstances. A query like `query { things(block: { hash: $hash }) { ... } }` might return data from different blocks if either

* **[H1]** the block with the given hash is not on the chain starting at the subgraph's current head
* **[H2]** the block gets removed from the subgraph's chain because of a block reorganization happening while the query is executing

Users can make sure they do not encounter this problem by only querying per block hash for blocks that are final, and by ensuring that the block hash is on the main chain.

Problem H1 can happen if we downloaded a block with `$hash` that was later removed from the subgraph's chain because of a reorganization, and is possible for any block, no matter whether it can be considered final or not. The query should return an error, as the block is not on the main chain, but might return successfully. The data returned will be from the block on the main chain that has the same number as block `$hash`.

Problem H2 can only happen for blocks that can not be considered final yet; for `graph_node`, that means that they are within `REORG_THRESHOLD` of the current chain head, but in practice only affects blocks within two or three blocks of the chain head. The query should return an error, as the block `$hash` is no longer on the main chain when the query finishes, but will return successfully with data that might come from either the block with `$hash` or one of the siblings with the same number that `graph-node` downloaded while it was processing the query.

#### Solving H1

Solving this issue requires that we have the infrastructure in place to answer the question "Is the block `$hash` on the chain starting at the subgraph's current head?"

Phase 1 of the [block explorer](https://github.com/graphprotocol/graph-node/pull/1383) will make it possible to distinguish between blocks on the main chain and ommers. If block `$hash` is at least `REORG_THRESHOLD` away from the chain head, we can reject the query if the block is an ommer. For blocks `$hash` within `REORG_THRESHOLD` of the chain head, we will keep a list of `(block_hash, block_number)` for each subgraph, tying the subgraph's current head to the finalized chain, and reject the query if `$hash` is not in that list.

#### Solving H2

For this, we need to detect that the subgraph's view of the chain was reorganized while the query was running; this is only necessary for blocks that are not final yet, i.e., that are within `REORG_THREASHOLD` of the chain head. To support this check, we will add a `generation` attribute to each `SubgraphDeployment` that is incremented every time the subgraph's chain is reorganized. We retrieve the `generation` attribute before we start executing the query, and check that it has not changed after we have finished executing the query. If it did change, we report an error to the user.

#### Time-travel queries by block number

[...]

PR #1801, opened 2020-07-22
Detect reorgs during execution of a GraphQL query

lutter:
With this PR, a graph-node will check whether the portion of a subgraph's data that is needed to formulate the response has changed because of a reorg of the subgraph while the GraphQL query was being executed. If that check indicates that there was a reorg affecting the query result, a GraphQL error is returned.

This implements a good chunk of issue #1405; what's still missing is detecting whether the block hash against which we execute the query is actually on the chain that starts at the subgraph's current head.


leoyvens:
@lutter looks like the integration tests are failing, at least on CI.

store/postgres/src/entities.rs
```diff
@@ -309,7 +309,7 @@ pub(crate) fn make_storage_cache() -> StorageCache {
 /// been migrated
 #[derive(Constructor)]
 pub(crate) struct Connection<'a> {
-    conn: MaybeOwned<'a, PooledConnection<ConnectionManager<PgConnection>>>,
+    pub conn: MaybeOwned<'a, PooledConnection<ConnectionManager<PgConnection>>>,
```
leoyvens, line 312: `pub(crate)`

store/postgres/tests/store.rs
```diff
@@ -2237,3 +2256,123 @@ fn cleanup_cached_blocks() {
...
+        transact_entity_operations(
+            store,
+            TEST_SUBGRAPH_ID.clone(),
+            block.clone(),
+            vec![test_entity_1],
+        )
+        .unwrap();
+    }
+
+    macro_rules! check_state {
```
leoyvens, line 2282: Why can't this macro be a function?

graph/src/data/query/error.rs
```diff
@@ -224,7 +225,8 @@ impl fmt::Display for QueryExecutionError {
             Panic(msg) => write!(f, "panic processing query: {}", msg),
             EventStreamError => write!(f, "error in the subscription event stream"),
             FulltextQueryRequiresFilter => write!(f, "fulltext search queries can only use EntityFilter::Equal"),
-            TooExpensive => write!(f, "query is too expensive")
+            TooExpensive => write!(f, "query is too expensive"),
+            DeploymentReverted => write!(f, "the subgraph deployment was reverted while executing the query"),
```
leoyvens, line 229: This error message isn't very clear, to me it almost sounds as if the deployment was entirely erased. Maybe "The chain had a reorganization while the query was executing".

chain/ethereum/src/block_stream.rs
```diff
@@ -629,7 +629,7 @@ where
         let head_ptr_opt = self.chain_store.chain_head_ptr()?;
         let subgraph_ptr = self.subgraph_store.block_ptr(self.subgraph_id.clone())?;
 
-        if head_ptr_opt != subgraph_ptr {
+        if head_ptr_opt.is_none() || subgraph_ptr.is_none() || head_ptr_opt != subgraph_ptr {
```
leoyvens, line 632: Maybe you found it more readable this way, but checking if `subgraph_ptr` is `None` is redundant with the other two checks. If `head_ptr_opt` is `None` then it's redundant with `head_ptr_opt.is_none()`. If `head_ptr_opt` is `Some` it is redundant with `head_ptr_opt != subgraph_ptr`.

store/postgres/src/subgraphs.graphql
```diff
@@ -43,6 +43,12 @@ type SubgraphDeployment @entity {
     graftBase: SubgraphDeployment
     graftBlockHash: Bytes
     graftBlockNumber: BigInt
+
+    # Tracking of reorgs so that we can spot that a reorg
+    # happened while a query was being processed
+    reorgCount: Int
+    currentReorgDepth: Int
+    maxReorgDepth: Int
```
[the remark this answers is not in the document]
lutter: When I make that `Int!`, I get a bunch of test failures because the storage layer requires that new entities have a value for these fields; but for these fields that's not necessary, since the columns have a `default 0` clause.

store/postgres/src/entities.rs
lutter: The struct us already `pub(crate)`, and the module is not exported outside the crate - is the `pub(crate)` still necessary here then?

store/postgres/tests/store.rs
lutter: It's a macro so that you get accurate line numbers on test failure, i.e., the line number with the `check_state!` rather than the line number of the failed assert inside `fn check_state`

graph/src/data/query/error.rs
lutter: I changed it to `the chain was reorganized while executing the query`. What do you think?

chain/ethereum/src/block_stream.rs
lutter: Yes, I just liked the symmetry better since the redundant check gives you a little less pause when reading that. I think the condition should be changed to have the common case `head_ptr_opt != subgraph_ptr` first though

store/postgres/src/subgraphs.graphql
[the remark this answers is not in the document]
leoyvens: Oh got it

graph/src/data/query/error.rs
leoyvens: Better, thanks

store/postgres/src/entities.rs
leoyvens: Ok this can be kept as it is.

lutter:
I've added a test that makes sure the `GraphQlRunner` detects a reorg and rebased the PR to the latest master.

graphql/tests/query.rs
```diff
@@ -252,6 +252,21 @@ async fn execute_query_document_with_variables(
         .clone()
 }
 
+fn execute_query_document_with_state(query: q::Document, state: DeploymentState) -> QueryResult {
```
leoyvens, line 255: This is called from a `tokio::test`, so it can be made an `async fn`.

leoyvens (changes requested):

graphql/tests/query.rs
```diff
@@ -1454,3 +1469,29 @@ async fn block_extension() {
...
+    assert_eq!(result.data.unwrap(), object!(musician: object!(id: "m1")));
+
+    // Simulate a reorg; this is what store::metadata::revert_block_ptr does
+    // though it goes through the database
+    let mut state = STORE
+        .deployment_state_from_id(TEST_SUBGRAPH_ID.clone())
+        .expect("failed to get state");
+    state.reorg_count += 1;
+    state.max_reorg_depth += 1;
+    let result = execute_query_document_with_state(query.clone(), state);
```
leoyvens, line 1492: So what will happen in this test is that a revert will be detected because the `state` reorg count which is `1` will be different from the store reorg count (which I assume is `0`). But in the real code isn't it an invariant that the query reorg count is always less or equal to the reorg count in the DB? If so that should ideally be respected in this test.

graphql/src/runner.rs
```diff
@@ -106,9 +114,33 @@ where
...
+    fn deployment_changed(
+        &self,
+        state: DeploymentState,
+        latest_block: u64,
+    ) -> Result<(), QueryExecutionError> {
+        if *GRAPHQL_ALLOW_DEPLOYMENT_CHANGE {
+            return Ok(());
+        }
+        let new_state = self.store.deployment_state_from_id(state.id.clone())?;
+        if state.reorg_count != new_state.reorg_count {
```
leoyvens, line 130: Couldn't we assert here that `new_state.reorg_count >= state.reorg_count` and therefore this check can be `state.reorg_count < new_state.reorg_count`. The test would trip this assert, as said in the other comment.

lutter: Yep, changed that.

graphql/tests/query.rs
lutter: Very good point! Fixing this, and having a true test that does a revert and forward move turned out to be pretty involved, since the tests in this file were all readonly and lazily set up the database. WIth a test that mutates, that's not possible any more. I've changed how the tests do their thing, forced them to be sequential, and changed the reorg test to go through reverting and moving forward again.

lutter:
I haven't rebased yet to make it a little easier to get through the review; I'll rebase once you are happy with the PR.

leoyvens (approved):
Nice! Thanks for doing this effort to revamp the testing. Feel free to merge after rebasing.

Issue #1460, opened 2020-01-18
It is not possible to pass the GraphQL block argument as a variable


lutter:
Brendan on Discord reported this: 

This works:
```graphql
query poolContractsDrawQuery($poolAddress: ID!, $drawId: String!, $winnerAddress: String!, $first: Int!, $skip: Int!) {
  poolContract(id: $poolAddress, block: { number: 15561871 }) {
  }
}
```

This doesn't:
```graphql
query poolContractsDrawQuery($poolAddress: ID!, $blockNumber: Int!, $drawId: String!, $winnerAddress: String!, $first: Int!, $skip: Int!) {
  poolContract(id: $poolAddress, block: { number: $blockNumber }) {
  }
}
```

```
Message: Invalid value provided for argument `block.number`: Variable("blockNumber"), Location: [object Object], Path: undefined
```

nionis:
I am experiencing this issue too, any ideas on how to workaround this @lutter ?

lutter:
You can work around this issue by putting the hash or number directly into a query, rather than using a variable.

moodysalem:
Would there be an ETA for this fix? @lutter 
GraphQL queries are usually expected to be immutable, and work best with other tooling when they are

JasoonS:
I'm in the same boat as @moodysalem 

I eagerly await this change!

PR #1803, opened 2020-07-23
Allow using variables in block constraints

lutter:
So far, it was not possible to use a variable in a block constraint. As it turns out, after various refactorings over the last few months, fixing #1460 was rather easy.

### DIFF
--- a/graphql/src/query/ext.rs
+++ b/graphql/src/query/ext.rs
@@ -1,6 +1,6 @@
 //! Extension traits for graphql_parser::query structs
 
-use graphql_parser::query as q;
+use graphql_parser::{query as q, Pos};
 
 use std::collections::{BTreeMap, HashMap};
 use std::convert::TryFrom;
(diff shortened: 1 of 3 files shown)


graphql/src/query/ext.rs
```diff
@@ -58,6 +64,19 @@ impl FieldExt for q::Field {
             )
         }
 
+        fn lookup<'a>(
```
leoyvens, line 67: This looks generally useful, should it go under `FieldExt`?

lutter: Good point - I pulled it into `ValueExt::lookup`. `Field` got only dragged into it so we would have a position for the error message.

leoyvens: Even nicer!

moodysalem:
Thanks for fixing this! What's the ETA for deployment of this fix?

Jannis:
@moodysalem Oh, it's already deployed to production!

Jannis:
@moodysalem Quick demo using httpie:

```sh
$ http post \
  https://api.thegraph.com/subgraphs/id/Qme2hDXrkBpuXAYEuwGPAjr6zwiMZV4FHLLBa3BHzatBWx \
  query='query transfers($block: Block_height!) { transfers(first: 2, block: $block) { id } }' \
  variables:='{ "block": { "number": 10000000 } }'

HTTP/1.1 200 OK
Alt-Svc: clear
Content-Length: 182
Content-Type: application/json
Date: Tue, 28 Jul 2020 12:32:08 GMT
Server: nginx/1.19.1
Via: 1.1 google
access-control-allow-headers: Content-Type, User-Agent
access-control-allow-methods: GET, OPTIONS, POST
access-control-allow-origin: *

{
    "data": {
        "transfers": [
            {
                "id": "0x0000090a7b11fa8646ac3c9214af4c21c2b525ab72488dde243b4f2a0ef7d896-11"
            },
            {
                "id": "0x0000a98fd5282163c8a29c174360d6c2ba5f51053cee2f1002798bc975ce8cb5-78"
            }
        ]
    }
}
```

Issue #1805, opened 2020-07-23
overflow when subtracting duration from instant


kevholder:
**Do you want to request a *feature* or report a *bug*?**

bug

**What is the current behavior?**

panic

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem.**

```
    Finished release [optimized] target(s) in 0.30s
     Running `target/release/graph-node --postgres-url 'postgresql://user@localhost:5432/graph-node' --ethereum-rpc 'mainnet:http://localhost:8333' --ipfs '127.0.0.1:5001'`
Jul 23 17:40:17.334 INFO Graph Node version: 0.18.0 :: fraction+4 (dc826116b 2020-07-20) dirty 1 modification
Jul 23 17:40:17.335 INFO Starting up
Jul 23 17:40:17.335 INFO Trying IPFS node at: http://127.0.0.1:5001/
Jul 23 17:40:17.335 INFO Creating transport, url: http://localhost:8333, network: mainnet
Jul 23 17:40:17.335 INFO Connecting to Postgres, conn_pool_size: 10, url: postgresql://user@localhost:5432/graph-node
Jul 23 17:40:17.335 INFO Successfully connected to IPFS node at: http://127.0.0.1:5001/
Jul 23 17:40:17.395 INFO Connected to Postgres, url: postgresql://user@localhost:5432/graph-node, component: Store
Jul 23 17:40:17.402 INFO Connecting to Ethereum..., network: mainnet
Jul 23 17:40:17.412 INFO Connected to Ethereum, network_version: 1515, network: mainnet
Jul 23 17:40:17.413 INFO Waiting for other graph-node instances to finish migrating, component: Store
Jul 23 17:40:17.413 INFO Running migrations, component: Store
Jul 23 17:40:17.414 INFO Migrations finished, component: Store
Jul 23 17:40:17.414 INFO Completed pending Postgres schema migrations, component: Store
Jul 23 17:40:17.415 INFO Creating LoadManager in disabled mode, component: LoadManager
thread 'tokio-runtime-worker' panicked at 'overflow when subtracting duration from instant', src/libstd/time.rs:374:9
stack backtrace:
   0:        0x10a8454ce - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::he3a6eecb2201efb0
   1:        0x10a86bc0c - core::fmt::write::h8ef98027ac1df1be
[...]

PR #1806, opened 2020-07-24
graph: Fix an error in KillState::new

lutter:
On Mac OSX, Instant::now() is the time since the last boot, which might
have been less than a day ago, in which case KillState::new would panic. We
now use 'now - 60s' as a time in the past for initialization, and if that
fails, we'll use just 'now'. There's not much danger in using 'now', it
might just lead to spurious logging and a too-soon update of the kill_rate
if the node becomes immediately overloaded after starting.

Thanks to kevholder for root-causing the issue and demonstrating a fix.

Fixes https://github.com/graphprotocol/graph-node/issues/1805

### DIFF
--- a/graph/src/data/graphql/effort.rs
+++ b/graph/src/data/graphql/effort.rs
@@ -149,9 +149,19 @@ struct KillState {
 
 impl KillState {
     fn new() -> Self {
-        let long_ago = Duration::from_secs(60);
-        let now = Instant::now();
-        let before = now.checked_sub(long_ago).unwrap_or(now);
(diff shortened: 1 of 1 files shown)


leoyvens:
This seems worth a comment

lutter:
True - I made the mistake of putting the comment into the commit message. Fixed now.

leoyvens:
@lutter I think you forgot to push the change.

lutter:
> @lutter I think you forgot to push the change.

I did indeed; rebased to latest master and pushed for realz

PR #1810, opened 2020-07-24
Support multiple Ethereum endpoints per network with different capabilities

fordN:
### Description
With this PR the `--ethereum-rpc`, `--ethereum-ipc`, and `ethereum-ws` arguments now accept a list of network capabilities along with the network name and url. Subgraph deployments will now use the "cheapest" available ethereum adapter for the network that supports its features. The supported network capabilities are: `Full`, `Archive`, and `Traces` with possible combinations of the three. If no capabilities are provided it will default to "archive,trace".  

Although not all components for a subgraph require the full capabilities each is still considered, so that adapter choice per subgraph is deterministic. For example, the BlockStream does not ever need an Archive Ethereum node because contract calls only go through the runtime host, but in that case the network with `Archive` enabled will still be used. Not only does this provide consistent across each subgraph's adapters, it allows for the capability list to be used for coarse load balancing.  With this API one can setup a node that moves all ethereum activity for subgraphs with call handlers or call filters to a specific Ethereum endpoint. 

#### Example:
```
--ethereum-rpc mainnet:full,tracing:http://localhost:8545
```

##### Subgraph features with require specific network capabilities::
Call handlers -> Traces
Block handlers with call filters -> Traces
Contract calls -> Archive


core/src/subgraph/instance_manager.rs
```diff
@@ -270,12 +274,14 @@ impl SubgraphInstanceManager {
                                     &network
                                 ))
                                 .clone(),
-                            eth_adapters
-                                .get(&network)
+                            eth_networks
+                                .get_adapter_with_requirements(
```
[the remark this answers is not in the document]
Jannis: So this method would become
```rust
eth_networks
    .adapter_with_capabilities(network.clone(), &required_capabilities)
```
What do you think?

core/src/subgraph/instance_manager.rs
```diff
@@ -358,13 +364,10 @@ impl SubgraphInstanceManager {
...
-        // Identify whether there are templates with call handlers or
+        // Identify whether there are mappings with call handlers or
         // block handlers with call filters; in this case, we need to
-        // include calls in all blocks so we cen reprocess the block
-        // when new dynamic data sources are being created
-        let templates_use_calls = templates.iter().any(|template| {
-            template.has_call_handler() || template.has_block_handler_with_call_filter()
-        });
+        // include calls in all blocks
+        let include_calls_in_blocks = manifest.requires_traces();
```
Jannis, line 370: Here too I would do something like
```rust
let include_calls_in_blocks = manifest
    .required_ethereum_capabilities()
    .contains(&NetworkCapability::Archive);
```

Jannis: On the other hand, using `requires_traces()` / `required_ethereum_capabilities()` for determining whether to include calls doesn't seem right. I think the check before was easier to understand and less likely to lead to an incorrect result. Just imagine we add something else that also requires traces. Suddenly we'd include calls by accident even though the feature may not require them at all.

graph/src/components/ethereum/network.rs
```diff
@@ -0,0 +1,273 @@
...
+            [NetworkCapability::Traces] => network.contains(&NetworkCapability::Traces),
+            [NetworkCapability::Full, NetworkCapability::Traces] => {
+                network.contains(&NetworkCapability::Traces)
+            }
+            [NetworkCapability::Archive] => network.contains(&NetworkCapability::Archive),
+            [NetworkCapability::Archive, NetworkCapability::Traces] => {
+                network == &vec![NetworkCapability::Archive, NetworkCapability::Traces]
+            }
+            _ => false,
+        }
```
Jannis, line 57: `required` isn't guaranteed to be sorted, right? So we could be looking at any permutation of the capabilities. There must be a more elegant, less error-prone way of performing this check.

Since we're implementing `Ord`, `min`/`max` should help. _Something_ like

```rust
let max_network = network.iter().max();
let max_required = required.iter().max();
return max_network >= max_required;
```

graph/src/components/ethereum/network.rs
```diff
@@ -0,0 +1,273 @@
...
+    pub fn cheapest() -> Self {
+        NetworkCapability::Full
+    }
+}
+
+impl_slog_value!(NetworkCapability, "{:?}");
+
+#[derive(Clone)]
+pub struct EthereumNetworkAdapters {
+    pub adapters: BTreeMap<Vec<NetworkCapability>, Arc<dyn EthereumAdapter>>,
```
Jannis, line 69: I think this is too complicated. We're going to deal with a handful of adapters perhaps. A simple
```rust
struct EthereumNetworkAdapter {
  capabilities: Vec<NetworkCapability>,
  adapter: dyn EthereumAdapter
}
```
and
```rust
  pub adapters: Vec<EthereumNetworkAdapter>,
```
should be sufficient.

graph/src/components/ethereum/network.rs
```diff
@@ -0,0 +1,273 @@
...
+                    NetworkCapability::sufficient_capability(capabilities, required_capabilities)
+                })
+                .collect();
+        if sufficient_adapters.len() == 0 {
+            return Err(format_err!(
+                "A matching Ethereum network with {:?} was not found.",
+                required_capabilities
+            ));
+        }
+        Ok(sufficient_adapters.values().next().unwrap())
```
Jannis, line 90: With the above construction (simple vector of adapters), this can be simplified to something like
```rust
let min_required = required_capabilities.iter().max();

let capable_adapters = self.adapters
    .filter(|adapter|  adapter.capabilities.iter().max() >= min_required)
    .collect();

if capable_adapters.is_empty() {
  return Err(...)
} else {
  // pick a random adapter from `capable_adapters` and return it
}
```

Jannis: Everything else in this file becomes simpler as well.

graph/src/components/ethereum/network.rs
```diff
@@ -0,0 +1,273 @@
...
+    pub fn get_adapter_with_requirements(
+        &self,
+        network_name: String,
+        requirements: &Vec<NetworkCapability>,
+    ) -> Result<&Arc<dyn EthereumAdapter>, Error> {
+        self.networks
+            .get(&network_name)
+            .ok_or(format_err!("network not supported: {}", &network_name))
+            .and_then(|adapters| {
+                println!("{:?}", adapters.adapters.clone().keys());
```
Jannis, line 152: Drop this

graph/src/data/subgraph/mod.rs
```diff
@@ -271,6 +273,11 @@ pub enum SubgraphRegistrarError {
     NameNotFound(String),
     #[fail(display = "Ethereum network not supported by registrar: {}", _0)]
     NetworkNotSupported(String),
+    #[fail(
+        display = "Ethereum network {} not found with capabilities: {:?}",
```
Jannis, line 277: I'm not sure I'd say `network {} not found`. Maybe something like `Ethereum nodes for network {} are missing the following capabilities: {}`?

graph/src/data/subgraph/mod.rs
```diff
@@ -604,6 +611,43 @@ pub struct Mapping {
     pub link: Link,
 }
 
+impl Mapping {
+    pub fn calls_host_fn(&self, host_fn: &str) -> bool {
```
Jannis, line 615: We probably want to call this a few times for each mapping, so it may make sense to have a method that returns the parsed payloads that we can then run multiple checks on.

Something like
```rust
struct ParsedMapping (Vec<Payload>);

impl ParsedMapping {
    fn calls_host_fn(&self, host_fn: &str) -> bool {
        ...
    }
}

impl Mapping {
    fn parse(&self) -> Result<ParsedMapping, Error> {
        ...
    }
}
```

graph/src/data/subgraph/mod.rs
```diff
@@ -1110,6 +1138,41 @@ impl SubgraphManifest {
             .map(|data_source| data_source.source.start_block)
             .collect()
     }
+
+    pub fn all_mappings(&self) -> Vec<Mapping> {
```
Jannis, line 1142: I'd call this just `mappings`

graph/src/data/subgraph/mod.rs
```diff
@@ -1110,6 +1138,41 @@ impl SubgraphManifest {
...
+                    .map(|source| source.mapping.clone()),
+            )
+            .collect()
+    }
+
+    pub fn requires_traces(&self) -> bool {
+        self.all_mappings().iter().any(|mapping| {
+            dbg!(mapping.has_call_handler()) || dbg!(mapping.has_block_handler_with_call_filter())
+        })
+    }
```
Jannis, line 1158: I'd drop this

graph/src/data/subgraph/mod.rs
```diff
@@ -1110,6 +1138,41 @@ impl SubgraphManifest {
...
+
+    pub fn requires_archive(&self) -> bool {
+        self.all_mappings()
+            .iter()
+            .any(|mapping| mapping.calls_host_fn("ethereum.call"))
+    }
```
Jannis, line 1164: And this too

graph/src/data/subgraph/mod.rs
```diff
@@ -1110,6 +1138,41 @@ impl SubgraphManifest {
...
+    pub fn all_eth_requirements(&self) -> Vec<NetworkCapability> {
+        let mut caps = Vec::new();
+        if self.requires_traces() {
+            caps.push(NetworkCapability::Traces)
+        }
+        if self.requires_archive() {
+            caps.push(NetworkCapability::Archive)
+        }
+        caps
+    }
```
Jannis, line 1175: This I'd rename to `required_ethereum_capabilities` and internalize the `requires_traces` and `requires_archive` methods.

node/src/main.rs
```diff
@@ -551,15 +554,18 @@ async fn main() {
     let expensive_queries = read_expensive_queries().unwrap();
 
     graph::spawn(
-        futures::stream::FuturesOrdered::from_iter(stores_eth_adapters.into_iter().map(
-            |(network_name, eth_adapter)| {
+        futures::stream::FuturesOrdered::from_iter(stores_eth_networks.flatten().into_iter().map(
+            |(network_name, capabilities, eth_adapter)| {
+                let capabilities_string = format!("{:?}", &capabilities);
```
Jannis, line 559: Nit: sort these so they are ordered min->max or the other way around?

Jannis, line 559: I don't think this has to be variable. Can just inline it in the `info!` call.

node/src/main.rs
```diff
@@ -568,12 +574,14 @@ async fn main() {
             error!(stores_error_logger, "Was a valid Ethereum node provided?");
             panic!("Failed to connect to Ethereum node: {}", e);
         })
-        .map(move |(network_name, network_identifier)| {
+        .map(move |(network_name, capabilities, network_identifier)| {
+            let capabilities_string = format!("{:?}", &capabilities);
```
Jannis, line 578: Same here about sorting and dropping the variable.

graph/src/components/ethereum/network.rs
```diff
@@ -0,0 +1,273 @@
...
+use std::collections::{BTreeMap, HashMap};
+use std::fmt;
+use std::str::FromStr;
+use std::sync::Arc;
+
+use crate::components::ethereum::EthereumAdapter;
+pub use crate::impl_slog_value;
+
+#[derive(Clone, Copy, Debug, Hash, PartialEq, Eq, PartialOrd, Ord)]
+pub enum NetworkCapability {
```
Jannis, line 12: I'm strongly tempted to suggest that `(Ethereum)NodeCapability` is the better name. Everywhere this shows up it feels weird to me to talk about the network, when there may be other adapters for the same network that _have_ the right capability. So it's per node/provider, not per network.

node/src/main.rs
```diff
@@ -866,70 +880,187 @@ async fn main() {
...
+                    "Ethereum network name cannot be an empty string"
+                ));
+            }
 
-                if name.is_empty() {
-                    return Err(format_err!(
-                        "Ethereum network name cannot be an empty string"
-                    ));
-                }
+            let http_split_at = rest.find("http").ok_or_else(|| {
```
Jannis, line 917: `http` alone is not enough. We also support `ipc` and `ws(s)`.

Jannis, line 917: I don't think you want to split at `http` or so, but at the next `:`. That seems more robust.

node/src/main.rs
```diff
@@ -866,70 +880,187 @@ async fn main() {
...
-                if loc.is_empty() {
-                    return Err(format_err!("Ethereum node URL cannot be an empty string"));
+            let (network_capabilities_str, url) = rest.split_at(http_split_at);
+            let network_capabilities: Vec<NetworkCapability> = match network_capabilities_str.len()
+            {
+                0 => vec![NetworkCapability::Archive, NetworkCapability::Traces],
+                length => {
+                    network_capabilities_str[..length - 1].split(',')
+                        .map(|capability_str| {
+                            Ok(NetworkCapability::from_str(dbg!(capability_str))?)
```
Jannis, line 931: Drop the `dbg!`

Jannis: One cool trick I think you can do with `FromStr` impls is to just call `capability_str.parse()?` and it'll automatically coerce the type into `NetworkCapability`.

node/src/main.rs
```diff
@@ -866,70 +880,187 @@ async fn main() {
...
                 }
+            };
 
-                info!(
-                    logger,
-                    "Creating transport";
-                    "network" => &name,
-                    "url" => &loc,
-                );
+            println!("caps: {:?}", network_capabilities);
```
Jannis, line 937: Drop

node/src/main.rs
```diff
@@ -866,70 +880,187 @@ async fn main() {
...
-
-                Ok((
-                    name.to_string(),
-                    Arc::new(graph_chain_ethereum::EthereumAdapter::new(
-                        transport,
-                        eth_rpc_metrics.clone(),
-                    )) as Arc<dyn EthereumAdapter>,
-                ))
+            if network_capabilities.len() > 2 {
+                return Err(format_err!("Maximum number of Ethereum Capabilities exceeded: {}", network_capabilities.len()));
```
Jannis, line 944: No need to capitalize `Capabilities` here.

graph/src/components/ethereum/network.rs
fordN: I was hoping to move toward that, but right now I wanted to use the BTreeMap to enforce uniqueness in (network_name, capabilities) since there currently isn't a selection algorithm for deciding between identical networks.  

Ok, I see below that you suggest randomly picking if there are two, this does open up our options to supply more networks to the node.  Works for me.   It could also be useful to use identical network capabilities as fallback networks, restarting the subgraph with another endpoint if it gets stuck in a retry loop past a threshold.

graph/src/components/ethereum/network.rs
fordN: I like that idea, I have been starting to feel that NetworkCapability is confusing as well.  
I find that when I'm writing about these components I'm unsure whether to use "network", "endpoint", or "node". 
I'll update it to be `NodeCapability`; since it's in the `ethereum` component I don't think we need to include that in the name.

graph/src/components/ethereum/network.rs
leoyvens: I'd suggest using a set for the capabilities, rather than a `Vec`. Or a bitfield :stuck_out_tongue:

core/src/subgraph/instance_manager.rs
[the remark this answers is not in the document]
fordN: Sounds great. That's the delineation I was going for, but it looks like I left some inconsistencies!

graph/src/components/ethereum/network.rs
fordN: I was really hoping to do something like this, but I don't think it works because of the way there are two groups of capability types.  Let's try an example; what happens with a node that has [`Full`, `Trace`] and a subgraph with an `Archive` requirement.  If we order the enum: `Full`, `Archive`, `Trace` then the comparison will return true because `Trace` is higher then `Archive` but the node can't handle historic contract calls so this is incorrect.  

Now if we order the enum variants: `Full`, `Trace`, `Archive` the comparison will be incorrect in the case of node=`Archive` and subgraph requires [`Full`, `Trace`].

fordN: @leoyvens a struct of `bool`s is a good idea. 🤔  
I was hoping to use a flexible data structure like a vec which we could add capabilities into, but really the..structure..of a struct will be useful here, as evidenced by the messiness of this comparison function. Realistically, new Ethereum node features will be very few and far between.

Jannis: @fordN You're right, there's not a single order between the capabilities. My bad.

graph/src/components/ethereum/network.rs
Jannis: Yeah, I don't think there are any uniqueness constraints here.

core/src/subgraph/instance_manager.rs
Jannis: This is still valid.

core/src/subgraph/instance_manager.rs
```diff
@@ -270,12 +272,14 @@ impl SubgraphInstanceManager {
...
-                                .get(&network)
+                            eth_networks
+                                .adapter_with_capabilities(
+                                    network.clone(),
+                                    &required_capabilities)
                                 .expect(&format!(
-                                    "expected eth adapter that matches subgraph network: {}",
-                                    &network
-                                ))
+                                    "expected eth adapter that matches subgraph network {} with required capabilities: {:?}",
```
Jannis, line 280: You can do `{}` now instead of `{:?}`, no?

graph/src/components/ethereum/network.rs
```diff
@@ -0,0 +1,227 @@
...
+use crate::components::ethereum::EthereumAdapter;
+pub use crate::impl_slog_value;
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct NodeCapabilities {
+    pub archive: bool,
+    pub traces: bool,
+}
+
+impl Ord for NodeCapabilities {
```
Jannis, line 16: A little comment here would be good. Greater means... cheaper? More expensive?

graph/src/components/ethereum/network.rs
```diff
@@ -0,0 +1,227 @@
...
+            } => write!(f, "full"),
+            NodeCapabilities {
+                archive: true,
+                traces: false,
+            } => write!(f, "archive"),
+        }
+    }
+}
+
+impl_slog_value!(NodeCapabilities, "{:?}");
```
Jannis, line 61: Should be able to do `{}` now, no?

graph/src/components/ethereum/network.rs
```diff
@@ -0,0 +1,227 @@
...
+            .iter()
+            .filter(|adapter| &adapter.capabilities >= required_capabilities)
+            .collect();
+        if sufficient_adapters.is_empty() {
+            return Err(format_err!(
+                "A matching Ethereum network with {:?} was not found.",
+                required_capabilities
+            ));
+        }
+        // TODO: Randomly choose between capable nodes
```
Jannis, line 90: Looks like this should make it easy: https://docs.rs/rand/0.7.3/rand/seq/trait.IteratorRandom.html

graph/src/components/ethereum/network.rs
```diff
@@ -0,0 +1,227 @@
...
+        Ok(&sufficient_adapters.iter().next().unwrap().adapter)
+    }
+
+    pub fn cheapest(&self) -> Option<&Arc<dyn EthereumAdapter>> {
+        self.adapters
+            .iter()
+            .next()
+            .map(|ethereum_network_adapter| &ethereum_network_adapter.adapter)
```
Jannis, line 98: I don't see `self.adapters` being sorted, so this just returns _some_ adapter, I think?

graph/src/components/ethereum/network.rs
```diff
@@ -0,0 +1,227 @@
...
+            .entry(name)
+            .or_insert(EthereumNetworkAdapters {
+                adapters: vec![EthereumNetworkAdapter {
+                    capabilities,
+                    adapter: adapter.clone(),
+                }],
+            });
+        network_adapters.adapters.push(EthereumNetworkAdapter {
+            capabilities,
+            adapter: adapter.clone(),
```
Jannis, line 131: In case there was no entry for the network yet, this is adding the adapter twice. You could do this:
```rust
        let network_adapters = self
            .networks
            .entry(name)
            .or_insert(EthereumNetworkAdapters {
                adapters: vec![],
            });

        network_adapters.adapters.push(EthereumNetworkAdapter {
            capabilities,
            adapter: adapter.clone(),
        });
```

graph/src/components/ethereum/network.rs
```diff
@@ -0,0 +1,227 @@
...
+}
+
+impl EthereumNetworks {
+    pub fn new() -> EthereumNetworks {
+        EthereumNetworks {
+            networks: HashMap::new(),
+        }
+    }
+
+    pub fn insert_or_update(
```
Jannis, line 114: This _always_ inserts, right? It may update the _network_ but it always inserts the _adapter_.

node/src/main.rs
```diff
@@ -866,70 +880,209 @@ async fn main() {
...
     let eth_rpc_metrics = Arc::new(ProviderEthRpcMetrics::new(registry));
-    networks
-        .map(|network| {
-            if network.starts_with("wss://")
-                || network.starts_with("http://")
-                || network.starts_with("https://")
-            {
-                return Err(format_err!(
-                    "Is your Ethereum node string missing a network name? \
+    networks.fold(Ok(EthereumNetworks::new()), |networks, network_arg| {
```
Jannis, line 890: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.try_fold seems like the right thing to do here.

node/src/main.rs
```diff
@@ -866,70 +880,209 @@ async fn main() {
...
+                    let capabilities = network_capabilities_str.split(',').collect::<Vec<&str>>();
+                    let invalid_capabilities: Vec<&&str> = capabilities
+                        .iter()
+                        .filter(|capability| !vec!["archive", "traces"].contains(capability))
+                        .collect();
+                    if !invalid_capabilities.is_empty() {
+                        return Err(format_err!(
+                            "Invalid Ethereum node capability supplied: {:?}",
+                            invalid_capabilities
+                        ));
```
Jannis, line 944: This would for a nice `FromStr` implementation for `NodeCapabilities`. Not a show-stopper but that would be nice to clean up in a follow-up PR.

I also still see the variable name `network_capabilities` here. ;)

Jannis: Well, ok, here it is:
```rust
impl FromStr for NodeCapabilities {
    type Err = Error;

    fn from_str(s: &str) -> Result<Self, Self::Err> {
        let capabilities: Vec<&str> = str.split(":").collect();
        Ok(NodeCapabilities {
            archive: capabilities.iter().find(cap => cap.eq("archive")).is_some(),
            traces: capabilities.iter().find(cap => cap.eq("traces")).is_some(),
        })
    }
}

graph/src/components/ethereum/network.rs
fordN: Good point, updated to `.insert()`.

node/src/main.rs
fordN: Yeah, that cleans it up a bit.

graph/src/components/ethereum/network.rs
fordN: Ok, now I'm sorting the EthereumAdapters when EthereumNetworks is instantiated, so this should return the Adapter with minimum capabilities.

graph/src/components/ethereum/network.rs
Jannis: @fordN I think the comment can go

core/src/subgraph/instance_manager.rs
Jannis: Actually. I'm ok with this.

PR #1811, opened 2020-07-25
store: Change how we index foreign keys

lutter:


store/postgres/src/relational.rs
```diff
@@ -1273,20 +1273,30 @@ impl Table {
             .filter(|col| !(col.is_list() && col.is_enum()))
             .enumerate()
         {
-            // Attributes that are plain strings are indexed with a BTree; but
-            // they can be too large for Postgres' limit on values that can go
-            // into a BTree. For those attributes, only index the first
-            // STRING_PREFIX_SIZE characters
-            let index_expr = if column.is_text() {
-                format!("left({}, {})", column.name.quoted(), STRING_PREFIX_SIZE)
+            let (method, index_expr) = if column.is_reference() && !column.is_list() {
```
leoyvens, line 1276: Sad that there's no index type that will work for both text arrays and ranges, so we won't be able to use a single index in the list case.

PR #1821, opened 2020-07-30
Replace unwrap with error in block stream

leoyvens:
Now that we support multiple endpoints, this can happen if the call goes to an endpoint that's behind.

### DIFF
--- a/graph/src/components/ethereum/adapter.rs
+++ b/graph/src/components/ethereum/adapter.rs
@@ -946,12 +946,17 @@ pub fn blocks_with_triggers(
     Box::new(
         trigger_futs
             .concat2()
-            .join(adapter.clone().block_hash_by_block_number(
-                &logger,
-                chain_store.clone(),
-                to,
-                true,
-            ))
+            .join(
+                adapter
+                    .clone()
+                    .block_hash_by_block_number(&logger, chain_store.clone(), to, true)
+                    .then(move |to_hash| match to_hash {
+                        Ok(n) => {
+                            n.ok_or_else(|| format_err!("Block {} not found in the chain", to))
+                        }
+                        Err(e) => Err(e),
+                    }),
+            )
             .map(move |(triggers, to_hash)| {
                 let mut block_hashes: HashSet<H256> =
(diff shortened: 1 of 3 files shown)


graph/src/components/ethereum/adapter.rs
```diff
@@ -946,12 +946,17 @@ pub fn blocks_with_triggers(
...
-                to,
-                true,
-            ))
+            .join(
+                adapter
+                    .clone()
+                    .block_hash_by_block_number(&logger, chain_store.clone(), to, true)
+                    .then(move |to_hash| match to_hash {
+                        Ok(n) => {
+                            n.ok_or_else(|| format_err!("Block {} not found in the chain", to))
```
lutter, line 955: Is there any way we can include the Ethereum provider in the error message? Is that info available here?

Jannis: I agree that would be nice. I wouldn't dump the whole URL though for security reasons (e.g. basic auth included in the URL) I'd do only the host perhaps.

leoyvens:
@lutter I've updated this to log the endpoint that's lagging. Though this is far from being the best way of monitoring endpoints that are behind.

graph/src/components/ethereum/adapter.rs
```diff
@@ -942,16 +944,27 @@ pub fn blocks_with_triggers(
...
-            ))
+            .join(
+                adapter
+                    .clone()
+                    .block_hash_by_block_number(&logger, chain_store.clone(), to, true)
+                    .then(move |to_hash| match to_hash {
+                        Ok(n) => n.ok_or_else(|| {
+                            warn!(logger2,
+                                    "Ethereum endpoint is behind";
+                                    "url" => eth_clone.url()
```
Jannis, line 961: Can we only log the hostname here? URLs may include `username:password@host/path` and we don't want that to show up in logs.

lutter: Yep, we just need enough to figure out which provider's dashboard to look at

leoyvens:
I've made this log only the hostname of the url.

PR #1825, opened 2020-07-31
Fix fragment bugs

### LINKED ISSUE #1816: queries should or shouldn't produce duplicate records
In reference to subgraph:  https://thegraph.com/explorer/subgraph/daostack/v39_4

Which is query is producing the correct behavior, this one that produces duplicate vote records:

```
query ProposalsSearchAllData {
  proposals {
    id
    ...ProposalFields
    votes {
      id
    }
  }
}

fragment ProposalFields on Proposal {
  votes {
    id
  }
}
```

or this one that does not:

```
query ProposalsSearchAllData {
  proposals {
    id
    
  votes {
    id
  }

    votes {
      id
    }
  }
}
```

leoyvens:
This fixes issues in how prefetch handles redundant selections in fragments. The changes are mostly around field collection, which will now have a single group of fields for response keys of objects, and for response keys of interfaces it will have a group of fields for the interface and one for each concrete type that appears in a type condition in the query.

Field collection in `execution.rs` was also refactored, even though it's not directly related to the issues.

This also changes where field collection happens, to make selecting only the relevant fields from the DB in a future patch easier.

Resolves #1816.

### DIFF
--- a/graph/src/data/graphql/object_or_interface.rs
+++ b/graph/src/data/graphql/object_or_interface.rs
@@ -1,0 +1,64 @@
+use crate::prelude::Schema;
+use graphql_parser::schema as s;
+
+#[derive(Copy, Clone, Debug)]
+pub enum ObjectOrInterface<'a> {
+    Object(&'a s::ObjectType),
+    Interface(&'a s::InterfaceType),
+}
+
+impl<'a> From<&'a s::ObjectType> for ObjectOrInterface<'a> {
+    fn from(object: &'a s::ObjectType) -> Self {
+        ObjectOrInterface::Object(object)
+    }
+}
+
+impl<'a> From<&'a s::InterfaceType> for ObjectOrInterface<'a> {
+    fn from(interface: &'a s::InterfaceType) -> Self {
+        ObjectOrInterface::Interface(interface)
+    }
+}
+
+impl<'a> ObjectOrInterface<'a> {
+    pub fn is_object(self) -> bool {
+        match self {
+            ObjectOrInterface::Object(_) => true,
+            ObjectOrInterface::Interface(_) => false,
+        }
+    }
+
+    pub fn name(self) -> &'a str {
+        match self {
+            ObjectOrInterface::Object(object) => &object.name,
+            ObjectOrInterface::Interface(interface) => &interface.name,
+        }
+    }
+
+    pub fn directives(self) -> &'a Vec<s::Directive> {
+        match self {
+            ObjectOrInterface::Object(object) => &object.directives,
+            ObjectOrInterface::Interface(interface) => &interface.directives,
+        }
+    }
+
+    pub fn fields(self) -> &'a Vec<s::Field> {
+        match self {
+            ObjectOrInterface::Object(object) => &object.fields,
+            ObjectOrInterface::Interface(interface) => &interface.fields,
+        }
+    }
+
+    pub fn field(&self, name: &s::Name) -> Option<&s::Field> {
+        self.fields().iter().find(|field| &field.name == name)
+    }
+
+    pub fn object_types(self, schema: &'a Schema) -> Option<Vec<&'a s::ObjectType>> {
+        match self {
+            ObjectOrInterface::Object(object) => Some(vec![object]),
+            ObjectOrInterface::Interface(interface) => schema
+                .types_for_interface()
+                .get(&interface.name)
+                .map(|object_types| object_types.iter().collect()),
+        }
+    }
+}
--- a/graph/src/data/graphql/object_or_interface.rs
+++ b/graph/src/data/graphql/object_or_interface.rs
@@ -1,5 +1,6 @@
 use crate::prelude::Schema;
 use graphql_parser::schema as s;
+use std::collections::BTreeMap;
 
 #[derive(Copy, Clone, Debug)]
 pub enum ObjectOrInterface<'a> {
@@ -59,6 +60,21 @@ impl<'a> ObjectOrInterface<'a> {
                 .types_for_interface()
                 .get(&interface.name)
                 .map(|object_types| object_types.iter().collect()),
+        }
+    }
+
+    /// `typename` is the name of an object type. Matches if `self` is an object and has the same
+    /// name, or if self is an interface implemented by `typename`.
+    pub fn matches(
+        self,
+        typename: &str,
+        types_for_interface: &BTreeMap<s::Name, Vec<s::ObjectType>>,
+    ) -> bool {
+        match self {
+            ObjectOrInterface::Object(o) => o.name == typename,
+            ObjectOrInterface::Interface(i) => types_for_interface[&i.name]
+                .iter()
+                .any(|o| o.name == typename),
         }
     }
 }
--- a/graph/src/data/graphql/object_or_interface.rs
+++ b/graph/src/data/graphql/object_or_interface.rs
@@ -28,6 +28,13 @@ impl<'a> ObjectOrInterface<'a> {
         }
     }
 
+    pub fn is_interface(self) -> bool {
+        match self {
+            ObjectOrInterface::Object(_) => false,
+            ObjectOrInterface::Interface(_) => true,
+        }
+    }
+
     pub fn name(self) -> &'a str {
         match self {
             ObjectOrInterface::Object(object) => &object.name,
--- a/graphql/src/execution/execution.rs
+++ b/graphql/src/execution/execution.rs
@@ -581,39 +581,33 @@ fn execute_selection_set_to_map<'a>(
             _ => (),
         }
 
-        // If the field exists on the object, execute it and add its result to the result map
-        if let Some(ref field) = sast::get_field(object_type, &fields[0].name) {
-            // Check if we have the value already.
-            let field_value = prefetched_object
-                .as_mut()
-                .map(|o| {
-                    // Prefetched objects are associated to `prefetch:response_key`.
-                    if let Some(val) = o.remove(&format!("prefetch:{}", response_key)) {
-                        return Some(val);
-                    }
+        // Unwrap: The query was validated to contain only valid fields.
+        let field = sast::get_field(object_type, &fields[0].name).unwrap();
 
-                    // Scalars and scalar lists are associated to the field name.
-                    // If the field has more than one response key, we have to clone.
-                    match multiple_response_keys.contains(fields[0].name.as_str()) {
-                        false => o.remove(&fields[0].name),
-                        true => o.get(&fields[0].name).cloned(),
-                    }
-                })
-                .flatten();
-            match execute_field(&ctx, object_type, field_value, &fields[0], field, fields) {
-                Ok(v) => {
-                    result_map.insert(response_key.to_owned(), v);
+        // Check if we have the value already.
+        let field_value = prefetched_object
+            .as_mut()
+            .map(|o| {
+                // Prefetched objects are associated to `prefetch:response_key`.
+                if let Some(val) = o.remove(&format!("prefetch:{}", response_key)) {
+                    return Some(val);
                 }
-                Err(mut e) => {
-                    errors.append(&mut e);
+
+                // Scalars and scalar lists are associated to the field name.
+                // If the field has more than one response key, we have to clone.
+                match multiple_response_keys.contains(fields[0].name.as_str()) {
+                    false => o.remove(&fields[0].name),
+                    true => o.get(&fields[0].name).cloned(),
                 }
+            })
+            .flatten();
+        match execute_field(&ctx, object_type, field_value, &fields[0], field, fields) {
+            Ok(v) => {
+                result_map.insert(response_key.to_owned(), v);
             }
-        } else {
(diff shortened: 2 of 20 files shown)


graphql/src/store/prefetch.rs
```diff
@@ -506,38 +466,15 @@ fn execute_root_selection_set(
...
-        let selections = fields
-            .into_iter()
-            .map(|f| q::Selection::Field((*f).clone()));
-        // See if this is an introspection or data field. We don't worry about
-        // nonexistant fields; those will cause an error later when we execute
-        // the query in `execution::execute_root_selection_set`
-        if sast::get_field(query_type, &name).is_some() {
-            data_set.items.extend(selections)
-        }
-    }
```
leoyvens, line 532: I removed this explicit separation of the introspection fields, because introspection fields will be ignored by field collection anyways.

graphql/src/store/prefetch.rs
```diff
@@ -658,17 +602,73 @@ fn execute_selection_set<'a>(
...
+    interface_cond: Option<&'a s::InterfaceType>,
+    interface_fields: Vec<&'a q::Field>,
+    concrete_types: IndexMap<ObjectCondition<'a>, Vec<&'a q::Field>>,
+}
+
+impl<'a> CollectedResponseKey<'a> {
+    fn collect_field(&mut self, type_condition: ObjectOrInterface<'a>, field: &'a q::Field) {
+        match type_condition {
+            ObjectOrInterface::Interface(i) => {
+                self.interface_cond = Some(i);
```
lutter, line 616: Why is it ok to clobber a potentially already existing `Some(_)` here?

graphql/src/store/prefetch.rs
```diff
@@ -26,61 +28,23 @@ lazy_static! {
...
-        use TypeCondition::*;
-        match (self, other) {
-            (Any, _) => Some(other.clone()),
-            (_, Any) => Some(self.clone()),
-            (On(name1), On(name2)) if name1 == name2 => Some(self.clone()),
-            _ => None,
-        }
-    }
+#[derive(Clone, Debug)]
+struct ObjectCondition<'a>(&'a s::ObjectType);
```
lutter, line 32: Would be good to comment that this is just an `ObjectType` that implements `Hash` and `Eq`

graphql/src/store/prefetch.rs
```diff
@@ -567,66 +504,73 @@ fn execute_selection_set<'a>(
...
+            .map(|cond| (ObjectOrInterface::Interface(cond), interface_fields));
+        for (type_cond, fields) in interface.into_iter().chain(
+            collected_fields
+                .concrete_types
+                .into_iter()
+                .filter(|(c, _)| match &relevant_concrete_types {
+                    None => true,
+                    Some(relevant_set) => relevant_set.contains(c.0.name.as_str()),
+                })
+                .map(|(c, f)| (ObjectOrInterface::Object(c.0), f)),
```
lutter, line 546: There's a lot going on here, even though it's just about getting an iterator over all `(TypeCondition, Vec<Field>)`  pairs in the `collected_fields`. Could constructing the iterator be a function on `CollectedResponseKey`?

graphql/src/store/prefetch.rs
```diff
@@ -658,17 +602,73 @@ fn execute_selection_set<'a>(
...
+        &mut HashSet::new(),
+        &mut grouped_fields,
+    )?;
+
+    // For interfaces, if a response key occurs both under the interface and under concrete types,
+    // we want to add the fields selected at the interface level to the selections in the specific
+    // concrete types, effectively merging the selection sets.
+    // See also: e0d6da3e-60cf-41a5-b83c-b60a7a766d4a
+    for collected_response_key in grouped_fields.values_mut() {
+        for concrete_type_fields in collected_response_key.concrete_types.values_mut() {
```
lutter, line 651: This `for` loop should be a function `merge_interface_fields` or maybe just `finish` on `CollectedResponseKey`

graphql/src/store/prefetch.rs
```diff
@@ -699,101 +699,77 @@ fn collect_fields<'a>(
...
-                            for (type_cond, mut group) in type_fields {
-                                if let Some(cond) = fragment_cond.and(&type_cond) {
-                                    grouped_fields
-                                        .entry(response_key)
-                                        .or_default()
-                                        .entry(cond)
-                                        .or_default()
-                                        .append(&mut group);
-                                }
+                    let fragment_ty = {
```
lutter, line 739: The body of this match arm is the same (I think) as most of the body of the previous arm; could this be factored into a helper function to make it clearer that we do almost the same thing for the two kinds of fragments, and what the differences are in how we treat them?

graphql/src/store/prefetch.rs
leoyvens: We actually check that in too many places. We have proper validation for `UnknownField` errors in `validate_fields`, so I've removed all other checks for these and replaced them with unwraps.

graphql/src/execution/execution.rs
```diff
@@ -556,58 +570,27 @@ pub fn collect_fields<'a>(
...
-                                // with the corresponding response key
-                                for (response_key, mut fragment_group) in fragment_grouped_field_set
-                                {
-                                    grouped_fields
-                                        .entry(response_key)
-                                        .or_default()
-                                        .append(&mut fragment_group);
-                                }
-                            });
+                    if visited_fragments.insert(&spread.fragment_name) {
```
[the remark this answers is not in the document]
leoyvens: Interesting observation. If we were to check for this, it would be checked in a validation step and not here. But I don't see any reason to bother with not supporting cycles when they can be supported just fine. I've added a note here.

graphql/src/store/prefetch.rs
```diff
@@ -717,76 +679,59 @@ fn collect_fields<'a>(
...
-                                }
-                            });
+                    if visited_fragments.insert(&spread.fragment_name) {
+                        let fragment = ctx.query.get_fragment(&spread.fragment_name);
+                        let fragment_ty = {
+                            let q::TypeCondition::On(ty_name) = &fragment.type_condition;
+                            object_or_interface_by_name(schema, ty_name).ok_or_else(|| {
+                                QueryExecutionError::NamedTypeError(ty_name.clone())
+                            })?
+                        };
```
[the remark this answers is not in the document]
leoyvens: Among the changes made there is now a `collect_fragment` function.

graphql/src/execution/execution.rs
```diff
@@ -540,11 +540,25 @@ pub fn collect_fields<'a>(
     ctx: &'a ExecutionContext<impl Resolver>,
     object_type: &s::ObjectType,
     selection_sets: impl Iterator<Item = &'a q::SelectionSet>,
-    visited_fragments: Option<HashSet<&'a q::Name>>,
 ) -> IndexMap<&'a String, Vec<&'a q::Field>> {
```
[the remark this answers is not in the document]
leoyvens: I've added a comment here.

graphql/src/store/prefetch.rs
```diff
@@ -717,76 +679,59 @@ fn collect_fields<'a>(
...
+
+                        if fragment_ty.is_object() {
+                            type_condition = fragment_ty
```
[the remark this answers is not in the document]
leoyvens: `type_condition` was actually never read from after this assignment, I've revamped the logic around this.

graphql/src/store/prefetch.rs
```diff
@@ -411,24 +372,23 @@ impl<'a> Join<'a> {
...
-            // not cause problems in later processing, but make sure that we
-            // do not clobber an entry under this `response_key` that might
-            // have been set by a previous join by appending values rather
-            // than using straight insert into the parent
-            let mut values = parent
+            // not cause problems in later processing.
+            //
+            // This `insert` will overwrite in the case where the response key occurs both at the
+            // interface level and in concrete types. The interface is always joined first, and may
+            // then be overwritten by the merged selection set under. the concrete type condition.
```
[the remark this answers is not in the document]
leoyvens: I've reworded this a bit to try to make it clearer. But yes, we continue to separate queries for separate type conditions returned by `collect_fields`.

graphql/src/store/prefetch.rs
```diff
@@ -567,66 +504,73 @@ fn execute_selection_set<'a>(
...
     let mut errors: Vec<QueryExecutionError> = Vec::new();
 
-    // Group fields with the same response key, so we can execute them together
-    let grouped_field_set = collect_fields(ctx, object_type, selection_sets, None);
+    if parents.is_empty() {
+        return Ok(Vec::new());
+    }
+
+    // Skip dead fragments as an optimization
+    let relevant_concrete_types: Option<HashSet<String>> = if parents[0].entity.is_empty() {
```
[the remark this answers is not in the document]
leoyvens: I've improved this function to fetch and join only for parents which match the type condition. This ended up being necessary for correctness and is also a more general optimization than this was, so this check was removed.

graphql/src/store/prefetch.rs
leoyvens: I've added an assert that if this assigns twice, it will assign the same value. It's not a very obvious invariant but `collect_fields` will never call this for two different interface types.

graphql/src/store/prefetch.rs
leoyvens: I've added a comment.

graphql/src/store/prefetch.rs
leoyvens: I moved this into an `IntoIterator` implementation for `CollectedResponseKey`, now that we don't need the `relevant_set` anymore.

graphql/src/store/prefetch.rs
leoyvens: There is now a `collect_fragment` function.

graphql/src/store/prefetch.rs
leoyvens: I don't think that would gain us much clarity, and that method would only ever be called here, so I'm fine with how this is now.

graph/src/data/graphql/object_or_interface.rs
```diff
@@ -0,0 +1,87 @@
+use crate::prelude::Schema;
+use graphql_parser::schema as s;
+use std::collections::BTreeMap;
+
+#[derive(Copy, Clone, Debug)]
+pub enum ObjectOrInterface<'a> {
```
leoyvens, line 6: This got moved into the `graph` crate to be used in validation.

leoyvens:
@lutter Thanks for the review. I added tests for nested interface fragments which were still buggy. This required changing multiple aspects of this PR, so it will need a re-review (recommend reading with whitespace diff disabled). Though many of your points were still valid and I addressed them.

roienatan:
What is the status of this?

dkent600:
@leoyvens Any ETA on release of this?  Thanks

graphql/src/store/prefetch.rs
```diff
@@ -506,295 +453,309 @@ fn execute_root_selection_set(
...
-            errors.push(QueryExecutionError::EmptySelectionSet(
-                object_type.name().to_owned(),
-            ));
-        }
         Err(errors)
     }
 }
 
+#[derive(Default)]
+struct CollectedResponseKey<'a> {
```
lutter, line 538: Just to make sure I understand what this is (and might be good as a comment): a `CollectedResponseKey` contains a map of the `ObjectType` to the fields we fetch for that object type. If we actually fetch fields for an interface type, i.e., `iface_cond` is `Some(_)`, `iface_fields` contains the fields we fetch from the interface. Do the fields in `obj_types` already contain these fields or are they solely the type-specific fields we fetch? I'll probably find the answer to that question when I read `collect_field`, but it would be nice if it was stated explicitly.

graphql/src/store/prefetch.rs
```diff
@@ -506,295 +453,309 @@ fn execute_root_selection_set(
...
+                visited_fragments,
+                output,
+            );
+        } else {
+            // This is an interface fragment in the root selection for an interface.
+            // We deal with this by expanding the fragment into one fragment for
+            // each type in the intersection between the root interface and the
+            // interface in the fragment type condition.
+            let types_for_interface = ctx.query.schema.types_for_interface();
+            let root_tys = &types_for_interface[outer_type_condition.name()];
```
lutter, line 681: Is it possible that `outer_type_condition` is an object type, but `fragment_ty` is an interface? I think that would lead to a panic here.

lutter (approved):
Looks great! Feel free to merge after addressing the two comments

graphql/src/store/prefetch.rs
leoyvens: The comment states that `This is an interface fragment in the root selection for an interface`. It's a bit subtle, but if the outer type is an object, we consider any condition to be redundant and therefore don't enter this if arm.

graphql/src/store/prefetch.rs
leoyvens: I've added a comment.

PR #1826, opened 2020-08-02
Revamp how we handle Prometheus metrics

lutter:
The biggest change in this PR is that metrics no longer have dynamically constructed names; instead, names are fixed, and labels are used to attach, e.g. the subgraph to the metric.

The PR also changes the `MetricsRegistry` trait a bit to make the various constructor methods a little more ergonomic and clearer in what they should be used for (like `new_counter` vs `new_subgraph_counter`)

### DIFF
--- a/chain/ethereum/src/network_indexer/block_writer.rs
+++ b/chain/ethereum/src/network_indexer/block_writer.rs
@@ -22,7 +22,7 @@ impl BlockWriterMetrics {
     ) -> Self {
         let transaction = Aggregate::new(
             "subgraph_transaction",
-            subgraph_id.to_string(),
+            subgraph_id.as_str(),
             "Transactions to the store",
             registry.clone(),
         );
--- a/chain/ethereum/src/network_indexer/block_writer.rs
+++ b/chain/ethereum/src/network_indexer/block_writer.rs
@@ -21,7 +21,8 @@ impl BlockWriterMetrics {
         registry: Arc<dyn MetricsRegistry>,
     ) -> Self {
         let transaction = Aggregate::new(
-            format!("{}_transaction", subgraph_id.to_string()),
+            "subgraph_transaction",
+            subgraph_id.to_string(),
             "Transactions to the store",
             registry.clone(),
         );
--- a/chain/ethereum/src/network_indexer/block_writer.rs
+++ b/chain/ethereum/src/network_indexer/block_writer.rs
@@ -21,7 +21,7 @@ impl BlockWriterMetrics {
         registry: Arc<dyn MetricsRegistry>,
     ) -> Self {
         let transaction = Aggregate::new(
-            "subgraph_transaction",
+            "deployment_transaction",
             subgraph_id.as_str(),
             "Transactions to the store",
             registry.clone(),
--- a/chain/ethereum/src/network_indexer/metrics.rs
+++ b/chain/ethereum/src/network_indexer/metrics.rs
@@ -84,63 +84,63 @@ impl NetworkIndexerMetrics {
 
             reorg_depth: Aggregate::new(
                 "subgraph_reorg_depth",
-                subgraph_id.to_string(),
+                subgraph_id.as_str(),
                 "The depth of reorgs over time",
                 registry.clone(),
             ),
 
             poll_chain_head: Aggregate::new(
                 "subgraph_poll_chain_head",
-                subgraph_id.to_string(),
+                subgraph_id.as_str(),
                 "Polling the network's chain head",
                 registry.clone(),
             ),
 
             fetch_block_by_number: Aggregate::new(
                 "subgraph_fetch_block_by_number",
-                subgraph_id.to_string(),
+                subgraph_id.as_str(),
                 "Fetching a block using a block number",
                 registry.clone(),
             ),
 
             fetch_block_by_hash: Aggregate::new(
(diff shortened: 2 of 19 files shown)


graph/src/components/metrics/mod.rs
```diff
@@ -11,57 +11,203 @@ pub mod stopwatch;
...
+                common_opts: opts,
+                buckets,
+            },
+            variable_labels
+                .iter()
+                .map(|s| s.as_str())
+                .collect::<Vec<&str>>()
+                .as_slice(),
+        )?);
+        self.register(name, histograms.clone());
```
Jannis, line 206: Have you tested this with multiple subgraphs? I'm wondering if registering the same name twice won't fail.

chain/ethereum/src/network_indexer/block_writer.rs
```diff
@@ -21,7 +21,8 @@ impl BlockWriterMetrics {
         registry: Arc<dyn MetricsRegistry>,
     ) -> Self {
         let transaction = Aggregate::new(
-            format!("{}_transaction", subgraph_id.to_string()),
+            "subgraph_transaction",
```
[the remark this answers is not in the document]
lutter: Could we just do `deployment`?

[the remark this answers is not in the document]
Jannis: I'm ok with that

graph/src/components/metrics/mod.rs
lutter: I have now run it in the integration cluster, and that worked without errors. I'll let it run over the weekend, but it looks like (after the additions I made) it's all working.

lutter:
This is all good to go from my POV

lutter:
Rebased to latest master and addressed an issue introduced in 27fb73c0 which made it possible to register global counters/gauges with different const labels, but you'd only ever use the one defined first since internally the global counters/gauges were kept by name, and not by the full `Desc.id`

graph/src/components/ethereum/adapter.rs
```diff
@@ -580,23 +578,21 @@ impl BlockStreamMetrics {
...
-                String::from(
-                    "Track the number of blocks a subgraph deployment is behind the HEAD block",
-                ),
-                HashMap::new(),
+            .new_deployment_gauge(
+                "deployment_blocks_behind",
+                "Track the number of blocks a subgraph deployment is behind the HEAD block",
+                deployment_id.as_str(),
             )
             .expect("failed to create `subgraph_blocks_behind` gauge");
```
fordN, line 590: Update error message to match the new metric name: `failed to create 'deployment_blocks_behind' gauge`.

graph/src/components/ethereum/adapter.rs
```diff
@@ -580,23 +578,21 @@ impl BlockStreamMetrics {
...
-            .new_gauge(
-                format!("subgraph_reverted_blocks_{}", deployment_id.to_string()),
-                String::from("Track the last reverted block for a subgraph deployment"),
-                HashMap::new(),
+            .new_deployment_gauge(
+                "deployment_reverted_blocks",
+                "Track the last reverted block for a subgraph deployment",
+                deployment_id.as_str(),
             )
             .expect("Failed to create `subgraph_reverted_blocks` gauge");
```
fordN, line 597: `deployment_reverted_blocks`

core/src/subgraph/instance_manager.rs
```diff
@@ -84,10 +84,8 @@ impl SubgraphInstanceManagerMetrics {
     pub fn new(registry: Arc<impl MetricsRegistry>) -> Self {
         let subgraph_count = registry
             .new_gauge(
-                String::from("subgraph_count"),
-                String::from(
-                    "Counts the number of subgraphs currently being indexed by the graph-node.",
-                ),
+                "subgraph_count",
```
fordN, line 87: "deployment_count"?

graph/src/components/ethereum/adapter.rs
lutter: Done

graph/src/components/ethereum/adapter.rs
lutter: Done

core/src/subgraph/instance_manager.rs
lutter: Yep. Fixed.

fordN (approved):
👍

chain/ethereum/src/network_indexer/metrics.rs
```diff
@@ -49,246 +48,152 @@ pub struct NetworkIndexerMetrics {
...
+            chain_head: make_gauge("deployment_chain_head", "The current chain head block"),
 
-            local_head: registry
-                .new_gauge(
-                    format!("{}_local_head", subgraph_id),
-                    "The current local head block".into(),
-                    HashMap::new(),
-                )
-                .expect(format!("failed to register metric `{}_local_head`", subgraph_id).as_str()),
+            local_head: make_gauge("deployment_local_head", "The current local head block"),
```
Jannis, line 76: On these two (and others related to the subgraph network): it might make sense to add a `network` label in order to be future proof for the days where one subgraph indexes across multiple networks. But for now, this is fine.

Jannis (approved):
All looking good to me!

Issue #1841, opened 2020-08-07
Failed call output decoding


leoyvens:
I've now seen a few reports of subgraphs failing on this line with the message: `Generating function outputs for an Ethereum call failed: Invalid data`. The latest example being `QmabWhS38ENWDVYL3MHpF7os57yFrfEFvqE397j5Pauqoo` at https://thegraph.com/explorer/subgraph/tech-nerd/productioneip20token.

Some subgraphs were reported to have "fixed themselves" after a restart. I tried one restart for that one and it did not work.

leoyvens:
[This transaction](https://etherscan.io/tx/0xc58a79a458b677e7ed5f3a9e76b03fdfbac88cdf40805779fb2a6b1764198baa) triggers the issue on [this subgraph](https://github.com/stevepro-lab/Subgraph).

kinrezC:
Hi there,

I'm running into a related issue where the subgraph fails with `Generating function inputs for an Ethereum call failed`. The latest example is at `QmXeyfE9zNJNV6Pugm5zmGfrjbAJ9S2wYayD8SQfV8WTSa` and the line in the mapping where it fails can be found here: https://github.com/asyncart/async-subgraph/blob/v2-contract/src/mapping.ts#L241.

stevepro-lab:
@leoyvens 
I made output params optional in the schema and only handling input params of the call handler. Now I am getting below error.  Seems below error is also related to the same issue 

```
Subgraph instance failed to run: Failed to process trigger in block #10457921 (34e887a7d9ed070a24cdeea3d27d707cde7c6d953e7089f3b61dba4b00110d7d), transaction cb0ba75fcb46c36f6d693e06ba4302784d39aea3626f98c9a2d32e308d283e94: Generating function inputs for an Ethereum call failed: Invalid data, code: SubgraphSyncingFailure, id: QmckE99oyeHMcg8pTFNEvKQEZVJ25muja3hC4DLWMgqCtN
```

epiqueras:
I am getting these all the time for various functions.

epiqueras:
Now I'm also getting the error for generating inputs.

fordN:
@epiqueras I'm diving into this encoding issue and could use any additional detail you have (the `invalid data` error from ethabi isn't very specific unfortunately.)  We've added detailed logs which include the raw data and abi structure of the param, could you share that log here?

epiqueras:
```
[...]

PR #1842, opened 2020-08-07
Log failed output decoding

leoyvens:
See #1841, if we print the raw data this gives us a shot at reproducing the error.


leoyvens:
@lutter I've added logging to the input decoding since we've also seen that fail, and also added the function ABI to the log. This should be a good start.

PR #1844, opened 2020-08-07
docker: Make bin/remove executable

lutter:


leoyvens:
TIL git keeps track of file permissions.

PR #1854, opened 2020-08-11
Tweak the BlockRangeContainsClause to make BRIN indexes usable

lutter:
The BlockRangeContainsClause used to expand to just `block_range @> {block}`; it now expands to that plus redundant clauses that make the BRIN index on block_range usable.
    
The redundant clauses are useful in situations where relatively few entities are updated a lot since the BRIN index is very effective in narrowing down where matching entities can actually be found in the heap.
    
For some problematic queries this can bring execution time from 10+s down to 500ms. It doesn't seem to increase execution times for queries that don't strictly need the extra clauses, but setting `DISABLE_BRIN_BLOCK_RANGE` in the environment can be used to turn this change off in case it does adversely affect query performance.

### DIFF
--- a/store/postgres/src/block_range.rs
+++ b/store/postgres/src/block_range.rs
@@ -20,9 +20,7 @@ lazy_static! {
     // Setting this to anything will disable the expanded BlockRangeContainsClause
     static ref DISABLE_BRIN_BLOCK_RANGE: bool = {
         env::var("DISABLE_BRIN_BLOCK_RANGE")
-            .ok()
-            .map(|_| false)
-            .unwrap_or(true)
+            .ok().is_some()
     };
 }
 
--- a/store/postgres/src/block_range.rs
+++ b/store/postgres/src/block_range.rs
@@ -4,12 +4,27 @@ use diesel::result::QueryResult;
 ///! Utilities to deal with block numbers and block ranges
 use diesel::serialize::{Output, ToSql};
 use diesel::sql_types::{Integer, Range};
+use lazy_static::lazy_static;
+use std::env;
 use std::io::Write;
 use std::ops::{Bound, RangeBounds, RangeFrom};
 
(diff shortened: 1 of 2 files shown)


store/postgres/src/block_range.rs
```diff
@@ -4,13 +4,28 @@ use diesel::result::QueryResult;
...
+lazy_static! {
+    // Make it possible to turn off the expanded BlockRangeContainsClause
+    // if that causes problems in the hosted service. Remove this variable
+    // once we know that it does not lead to performance issues
+    // Setting this to anything will disable the expanded BlockRangeContainsClause
+    static ref DISABLE_BRIN_BLOCK_RANGE: bool = {
+        env::var("DISABLE_BRIN_BLOCK_RANGE")
+            .ok()
+            .map(|_| false)
+            .unwrap_or(true)
```
leoyvens, line 25: Isn't this backwards, disabling by default and enabling when the variable is present? I think this should be just `ok().is_some()`.

store/postgres/src/block_range.rs
```diff
@@ -91,9 +106,42 @@ impl<'a> QueryFragment<Pg> for BlockRangeContainsClause<'a> {
...
     }
 }
+
+#[test]
+fn block_number_max_is_i32_max() {
+    // The code in this file embeds i32::MAX aka BLOCK_NUMBER_MAX in strings
+    // for efficiency. This assertion makes sure that BLOCK_NUMBER_MAX still
+    // is what we think it is
+    assert_eq!(2147483647, BLOCK_NUMBER_MAX);
+}
```
leoyvens, line 147: Smart to add this test!

store/postgres/src/block_range.rs
lutter: Ugh .. yes, you are right. Thanks for spotting that!

PR #1856, opened 2020-08-12
EthereumNetworkAdapters used in place of single adapter

fordN:
With this PR EthereumNetworkAdapters now implements the EthereumAdapter trait which is a retry wrapper on top of the trait impl for EthereumAdapters.  This way a failed Ethereum request will be retried across all sufficient adapters thus increasing robustness and allowing more Ethereum request volume to be split across Ethereum endpoints. 

_Note: I'm currently doing some local testing with "bad" endpoints to ensure the desired robustness is achieved._

### DIFF
--- a/chain/ethereum/src/block_stream.rs
+++ b/chain/ethereum/src/block_stream.rs
@@ -5,7 +5,8 @@ use std::sync::Mutex;
 use std::time::Duration;
 
 use graph::components::ethereum::{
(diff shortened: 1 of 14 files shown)


leoyvens:
One thing to note is that we probably do not want to expose endpoint info to logs that are shown to users (in case this PR does that).

leoyvens:
@fordN This needs a rebase and a green CI run.

leoyvens:
Closing since this is not being actively worked on. And we are focused on Firehose now.

PR #1857, opened 2020-08-12
store: Do not add redundant clauses to BlockRangeContainsClause

lutter:
For entities that are transaction-like in that entities are rarely, if ever, updated, adding redundant `block_range` clauses from PR #1854 slows queries down to the point where the overall performance of the system suffers. At the same time, for entities that are account like in that entities are updated frequently, adding redundant `block_range` clauses is indeed beneficial.

This reverts commit 9d72f3b82212b385a015230769045ab368a8981c and introduces a setting where account-like tables can be specified explicitly. Ultimately, we would like to automatically detect such tables, but to make sure that this is in fact a useful heuristic, we allow setting which tables to treat special in the environment to make it possible to test this hypothesis under load.

### DIFF
--- a/store/postgres/src/block_range.rs
+++ b/store/postgres/src/block_range.rs
@@ -4,25 +4,12 @@ use diesel::result::QueryResult;
 ///! Utilities to deal with block numbers and block ranges
 use diesel::serialize::{Output, ToSql};
 use diesel::sql_types::{Integer, Range};
-use lazy_static::lazy_static;
-use std::env;
 use std::io::Write;
 use std::ops::{Bound, RangeBounds, RangeFrom};
 
-use graph::prelude::{BlockNumber, BLOCK_NUMBER_MAX};
+use graph::prelude::BlockNumber;
 
 use crate::history_event::HistoryEvent;
-
-lazy_static! {
-    // Make it possible to turn off the expanded BlockRangeContainsClause
-    // if that causes problems in the hosted service. Remove this variable
-    // once we know that it does not lead to performance issues
-    // Setting this to anything will disable the expanded BlockRangeContainsClause
-    static ref DISABLE_BRIN_BLOCK_RANGE: bool = {
-        env::var("DISABLE_BRIN_BLOCK_RANGE")
-            .ok().is_some()
-    };
-}
 
 /// The name of the column in which we store the block range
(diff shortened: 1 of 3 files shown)


store/postgres/src/relational.rs
```diff
@@ -42,6 +44,23 @@ use crate::block_range::{BLOCK_RANGE_COLUMN, BLOCK_UNVERSIONED};
...
+    /// the BRIN index on block_range usable
+    ///
+    /// Example: GRAPH_ACCOUNT_TABLES=sgd21902.pair,sgd1708.things
+    static ref ACCOUNT_TABLES: HashSet<String> = {
+        env::var("GRAPH_ACCOUNT_TABLES")
+            .ok()
+            .map(|v| v.split(",").map(|s| s.to_owned()).collect())
+            .unwrap_or(HashSet::new())
+    };
+}
```
leoyvens, line 62: Is there a way to not duplicate this static variable?

store/postgres/src/block_range.rs
```diff
@@ -135,11 +127,3 @@ impl<'a> QueryFragment<Pg> for BlockRangeContainsClause<'a> {
...
     }
 }
-
-#[test]
-fn block_number_max_is_i32_max() {
-    // The code in this file embeds i32::MAX aka BLOCK_NUMBER_MAX in strings
-    // for efficiency. This assertion makes sure that BLOCK_NUMBER_MAX still
-    // is what we think it is
-    assert_eq!(2147483647, BLOCK_NUMBER_MAX);
-}
```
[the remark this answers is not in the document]
lutter: Ugh .. I didn't mean to delete that.

store/postgres/src/relational.rs
lutter: Yes, it's not needed in `block_range.rs`; I moved it from there to `relational.rs` and then forgot to delete it.

lutter:
Fixed up the PR

The longer range plan with this is to use `pg_stats.n_distinct` and compare it to `pg_class.rowestimate`; I'd go with a criterion where for tables where `rowestimate` is beyond a certain threshold (say 50k rows) and the ratio `n_distinct/rowestimate` is low (say < 0.1) we'd add the additional clauses. But this setup makes it possible to try this out, and to hopefully gauge whether this will actually be useful across the board, before investing time in writing code to automate the decision.

PR #1858, opened 2020-08-13
node: Increment ipfs-api to 0.7.2

### LINKED ISSUE #1799: Failed to connect to IPFS: api returned unknwon error '405 - Method Not Allowed ''

bug

```
Failed to connect to IPFS: api returned unknwon error '405 - Method Not Allowed'
```

Rung `ipfs daemon` with `go-ipfs version: 0.6.0` and then:

```bash
  --postgres-url postgresql://user@localhost:5432/graph-node \
  --ethereum-rpc mainnet:https://localhost:8333 \
  --ipfs 127.0.0.1:5001
```

`graph-node` should work with ipfs v6

cpfiffer:
Fixes #1799. `ipfs-api` 0.7.2 uses the correct messaging for newer IPFS clients. 

The fix also introduces a depwarning, since `IpfsClient::new_from_uri` in 

https://github.com/graphprotocol/graph-node/blob/087bc1d0a159c3fce4f613168b4ab43cc9f7c699/node/src/main.rs#L445

will be deprecated in `ipfs-api` 0.8.0, which isn't out yet.

### DIFF
--- a/node/Cargo.toml
+++ b/node/Cargo.toml
@@ -10,7 +10,7 @@ git-testament = "0.1"
 graphql-parser = "0.2.3"
 prometheus = "0.7"
 futures = { version = "0.3.1", features = ["compat"] }
-ipfs-api = { version = "0.7.1", features = ["hyper-tls"] }
+ipfs-api = { version = "0.7.2", features = ["hyper-tls"] }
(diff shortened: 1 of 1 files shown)


leoyvens:
Closing in favor of #2334.

Issue #1861, opened 2020-08-14
Label the metrics coming from a connection pool with the Postgres host name


lutter:
The connection pool reports connection use and wait time in a Prometheus metric. When using read replicas, it's not possible to distinguish against which database these metrics were taken. To make that possible, the metrics need to be labeled with the database's host name and port number

PR #1868, opened 2020-08-17
all: Label connection pool metrics with the pool they come from

lutter:
We identify pools as either 'main' or 'replicaN' in the labels

Fixes https://github.com/graphprotocol/graph-node/issues/1861

### DIFF
--- a/node/src/main.rs
+++ b/node/src/main.rs
@@ -151,6 +151,20 @@ async fn main() {
                 ),
         )
         .arg(
+            Arg::with_name("postgres-host-weights")
+                .multiple(true)
+                .use_delimiter(true)
+                .long("postgres-host-weights")
+                .value_name("WEIGHT,")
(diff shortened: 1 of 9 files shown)


lutter:
Following @leoyvens's idea of statically determining how much traffic to send to each database, added a commit that allows assigning weights to each database host

node/src/main.rs
```diff
@@ -150,6 +150,20 @@ async fn main() {
                        which will share the load with the primary server",
                 ),
         )
+        .arg(
+            Arg::with_name("postgres-host-weigths")
+                .multiple(true)
```
Jannis, line 155: Is this really a flag that requires `multiple(true)`? It's just a single CSV value.

lutter: I just followed what `postgres-secondary-hosts` does; not sure if you would want to specify the weights similarly

PR #1872, opened 2020-08-20
Gracefully fail on null `block.hash` and encoding failures in Ethereum adapter

fordN:
As a more diverse set of indexers use `graph-node` syncing from different Ethereum clients and providers, edge cases have been encountered in which block's are being returned without a `hash`.  This PR removes the assumption that the block.hash will always be `Some()` in these cases, so the cases fail gracefully and the steps are retried. 
Contract_call() has also been updated, so failed call_data input encoding returns an error (no unwrap.)

### DIFF
--- a/chain/ethereum/src/ethereum_adapter.rs
+++ b/chain/ethereum/src/ethereum_adapter.rs
@@ -1204,7 +1204,10 @@ where
         }
 
         // Encode the call parameters according to the ABI
-        let call_data = call.function.encode_input(&call.args).unwrap();
+        let call_data = match call.function.encode_input(&call.args) {
+            Ok(data) => data,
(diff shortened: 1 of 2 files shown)


chain/ethereum/src/ethereum_adapter.rs
```diff
@@ -974,6 +974,12 @@ where
                             );
                         });
                 }
+            } else {
+                warn!(
+                    logger1,
+                    "Failed to fetch block hash for block number";
+                    "number" => block_number
```
lutter, line 981: I am not really deep enough in the logic of all this to review it, but each log message that's related to a specific provider should include an indication of which provider was used. Does `logger1` already contain that? Otherwise, it should be added either to the logger or to this log message.

PR #1873, opened 2020-08-22
Log GraphQL queries only once

lutter:
Because block constraints might cause us to execute a GraphQL query in several steps, we used to log the entire query for each of these steps. That leads to inaccurate query execution counts and times, and to logs that are hugely inflated in size.
    
We now log the overall query execution only once. To get insight into cache usage, we also log cache usage information separately from the overall query execution. To keep log size manageable, we do not log the query text with the cache information. That would have to be inferred from the `query_id` logged with both log statements.


Jannis (approved):
Looks good as far as I can tell!

PR #1878, opened 2020-08-24
Add an LfuCache for historical queries

lutter:
We see a lot of historical queries in the hosted service, and our cache performs poorly for those. Add an `LfuCache` of query results for those to improve the cache hit ratio.

### DIFF
--- a/graphql/src/execution/execution.rs
+++ b/graphql/src/execution/execution.rs
@@ -92,9 +92,9 @@ lazy_static! {
         .expect("Invalid value for GRAPH_QUERY_CACHE_MAX_MEM environment variable")
     };
 
-    // Query cache by network.
+    // Cache query results for recent blocks by network.
     // The `VecDeque` works as a ring buffer with a capacity of `QUERY_CACHE_BLOCKS`.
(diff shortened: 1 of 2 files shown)


graphql/src/execution/execution.rs
```diff
@@ -57,6 +58,100 @@ impl CacheByBlock {
...
+        if let Some(cache_by_block) = cache.iter_mut().find(|c| c.block == block_ptr) {
+            cache_insert = cache_by_block.insert(key, result.cheap_clone(), weight);
+        } else if *QUERY_CACHE_BLOCKS > 0 {
+            // We're creating a new `CacheByBlock` if:
+            // - There are none yet, this is the first query being cached, or
+            // - `block_ptr` is of higher or equal number than the most recent block in the cache.
+            // Otherwise this is a historical query which will not be cached.
+            let should_insert = match cache.iter().next() {
+                None => true,
+                Some(highest) if highest.block.number <= block_ptr.number => true,
```
That3Percent, line 117: This can be simplified a bit:

```
let should_insert = match cache.iter().next() {
    None => true,
    Some(highest) => highest.block.number <= block_ptr.number,
}
```
Removes 1 branch from the `match` expression

lutter: That was in the existing code that I just copied around, but I agree - I just pushed a change to address that.

PR #1879, opened 2020-08-28
Improve estimate of how much memory is used by an LfuCache

lutter:
This PR improves the estimates that `LfuCache` uses to determine how big a cache entry is. Under heavy load, `LfuCache` would appear to leak memory. At least part of that apparent leak was that `LfuCache` underestimated how much memory cache entries use, which would lead to the cache growing much bigger than intended. In some cases, for example for small `BTreeMap`s, the estimate was off by more than an order of magnitude, which in practice has the same consequences as an actual leak.

With the tweaks to memory estimates, `LfuCache` now mostly overestimates the amount of memory an entry takes, though running the `stress` example still shows that overall memory usage can be a factor of 3 larger than the estimate that the cache tracks, for example, when caching strings less than 50 bytes long.

The PR also includes a small tool to show how cache estimates relate to actually used memory. You can run the tool with
```
cargo run -p graph --release --example stress -- \
      --niter 1000000 --print-count 10 --obj-size 50 --template string
```
which will perform 1M insertions of strings up to 50 bytes in size and print information about the cache about 10 times during the run. The information printed looks like
```
evicted:     59  dropped:     19 new_weight:   999991 heap_factor: 3.00
```
where `evicted` is the cache weight removed from the cache by the last call to `LfuCache.evict`, `dropped` is how much memory was actually freed, `new_weight` is the weight of the cache after the last `evict`, and `heap_factor` is the ratio of memory allocated by the cache divided by its weight.

The `stress` example defines tests for a few different data structures which can be selected with the `--template` option:
* `string`: cache `String` of length `obj_size`
* `vec`: cache `Vec<usize>` of length `obj_size`
* `hashmap`: cache `HashMap<String, String>` with `obj_size` entries; keys and values are short strings of less than 10 bytes
* `valuemap`: cache `BTreeMap<String, q::Value>` with `obj_size` entries, where entries cycle through the 9 possible kinds of `q::Value`

### DIFF
--- a/graph/examples/stress.rs
+++ b/graph/examples/stress.rs
@@ -1,0 +1,210 @@
+use std::alloc::{GlobalAlloc, Layout, System};
+use std::collections::{BTreeMap, HashMap};
+use std::iter::FromIterator;
+use std::sync::atomic::{AtomicUsize, Ordering::SeqCst};
+
+use graphql_parser::query as q;
+use rand::{thread_rng, Rng};
+use structopt::StructOpt;
+
+use graph::util::cache_weight::CacheWeight;
+use graph::util::lfu_cache::LfuCache;
+
+struct Counter;
+
+static ALLOCATED: AtomicUsize = AtomicUsize::new(0);
+
+unsafe impl GlobalAlloc for Counter {
+    unsafe fn alloc(&self, layout: Layout) -> *mut u8 {
+        let ret = System.alloc(layout);
+        if !ret.is_null() {
+            ALLOCATED.fetch_add(layout.size(), SeqCst);
+        }
+        return ret;
+    }
+
+    unsafe fn dealloc(&self, ptr: *mut u8, layout: Layout) {
+        System.dealloc(ptr, layout);
+        ALLOCATED.fetch_sub(layout.size(), SeqCst);
+    }
+}
+
+#[global_allocator]
+static A: Counter = Counter;
+
+#[derive(StructOpt)]
+#[structopt(name = "stress", about = "Stress test for the LFU Cache")]
+struct Opt {
+    #[structopt(short, long, default_value = "1000")]
+    niter: usize,
+    #[structopt(short, long, default_value = "10")]
+    print_count: usize,
+    #[structopt(short, long, default_value = "1024")]
+    obj_size: usize,
+    #[structopt(short, long, default_value = "1000000")]
+    cache_size: usize,
+}
+
+trait Template<T> {
+    type Item;
+
+    fn create(size: usize) -> Self;
+    fn sample(&self, size: usize) -> Self::Item;
+}
+
+impl Template<Vec<usize>> for Vec<usize> {
+    type Item = Vec<usize>;
+
+    fn create(size: usize) -> Self {
+        Vec::from_iter(0..size)
+    }
+    fn sample(&self, size: usize) -> Self::Item {
+        self[0..size].into()
+    }
+}
+
+impl Template<HashMap<String, String>> for HashMap<String, String> {
+    type Item = Self;
+
+    fn create(size: usize) -> Self {
+        let mut map = HashMap::new();
+        for i in 0..size {
+            map.insert(format!("key{}", i), format!("value{}", i));
+        }
+        map
+    }
+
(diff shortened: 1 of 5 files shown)


graph/src/util/lfu_cache.rs
```diff
@@ -40,6 +40,12 @@ impl<K, V: Default> CacheEntry<K, V> {
             will_stale: false,
         }
     }
+
+    /// Estimate the size of a `CacheEntry` with the given key and value
+    fn weight(key: &K, value: &V) -> usize {
```
That3Percent, line 45: Would this be simpler if we just required `K` to impl `CacheWeight` as well?

graph/src/util/lfu_cache.rs
```diff
@@ -94,9 +100,11 @@ impl<K: Clone + Ord + Eq + Hash + Debug, V: CacheWeight + Default> LfuCache<K, V
                 );
             }
             Some(entry) => {
+                let old_weight = entry.weight;
```
That3Percent, line 103: A bit simpler (removes temporary):

```
Some(entry) => {
   self.total_weight -= entry.weight;
   self.total_weight += weight;
   entry.weight = weight;
   entry.value = value;
}
```

graph/src/util/lfu_cache.rs
```diff
@@ -145,9 +153,9 @@ impl<K: Clone + Ord + Eq + Hash + Debug, V: CacheWeight + Default> LfuCache<K, V
         self.queue.len()
     }
 
-    pub fn evict(&mut self, max_weight: usize) {
+    pub fn evict(&mut self, max_weight: usize) -> Option<(usize, usize, usize)> {
```
That3Percent, line 156: It would be good to comment on what this return value is.

graph/src/util/lfu_cache.rs
```diff
@@ -79,7 +85,7 @@ impl<K: Clone + Ord + Eq + Hash + Debug, V: CacheWeight + Default> LfuCache<K, V
 
     /// Updates and bumps freceny if already present.
     pub fn insert(&mut self, key: K, value: V) {
-        let weight = value.weight();
+        let weight = CacheEntry::weight(&key, &value);
```
That3Percent, line 88: This is an existing problem so there is no regression, but the weight of the LfuCache itself (in particular the `queue` field) may not be insignificant for small K/V.

graph/examples/stress.rs
```diff
@@ -0,0 +1,293 @@
+use std::alloc::{GlobalAlloc, Layout, System};
```
That3Percent, line 1: I'm skipping the review for this file because I don't think it matters. Let me know if you disagree or need any particular part of it reviewed.

That3Percent:
@lutter Looks like some tests fail as a result of these updates. You might just need to update the expected values according to to the new calculations.

graph/src/util/lfu_cache.rs
lutter: Yeah, I wound up doing that when I fixed the unit test - it is simpler

graph/src/util/lfu_cache.rs
lutter: I wrote it that way since I find it a little easier to follow what is happening there - I bet the compiler generates pretty much the same code for the two variants.

graph/src/util/lfu_cache.rs
lutter: Done

graph/src/util/lfu_cache.rs
lutter: Yeah, there's still quite a bit of wiggle room - if what I have so far makes the total cache size more predictable, I'll stick with it; otherwise, we'll need another round of improving estimates. It's crazy how something that would just be an exercise in pointer chasing in C is so hard here (though it would be similarly hard in any language that hides implementation details, like C++)

graph/examples/stress.rs
lutter: No, this file is really just a tool to explore various scenarios and get a better understanding of memory usage.

lutter:
Updating the test was a little tricky since it requires carefully calculated weights that we don't know a priori. But it looks like I parametrized everything successfully now.

PR #1886, opened 2020-09-02
Bug fix: Fetch correct function abi from contract abi 

### LINKED ISSUE #1841: Failed call output decoding

fordN:
Resolves #1841 

This PR fixes a bug in `contract_function_with_signature()` which allowed for false positives to be returned. This inconsistency led to incorrect function abis being used in function param decoding leading to `invalid data` errors in [ethabi](https://github.com/graphprotocol/ethabi). 

I've also restructured `contract_function_with_signature()` to reduce unnecessary work building function signature hashes for functions that can be excluded early because they are `view` or `pure` functions.

### DIFF
--- a/graph/src/util/ethereum.rs
+++ b/graph/src/util/ethereum.rs
@@ -122,21 +122,25 @@ pub fn contract_function_with_signature<'a>(
     contract: &'a Contract,
     target_signature: &str,
 ) -> Option<&'a Function> {
-    contract.functions().find(|function| {
-        // Construct the argument function signature:
-        // `address,uint256,bool`
-        let mut arguments = function
-            .inputs
-            .iter()
-            .map(|input| format!("{}", input.kind))
-            .collect::<Vec<String>>()
-            .join(",");
-        // `address,uint256,bool)
-        arguments.push_str(")");
-        // `operation(address,uint256,bool)`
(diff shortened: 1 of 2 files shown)


leoyvens (approved):

graph/src/util/ethereum.rs
```diff
@@ -123,20 +123,24 @@ pub fn contract_function_with_signature<'a>(
     target_signature: &str,
 ) -> Option<&'a Function> {
     contract.functions().find(|function| {
```
leoyvens, line 125: Actually this would be better if we checked the mutability in a call to `filter`. Also we should do an exhaustive match on all variants, for better future proofing.

leoyvens (changes requested):
Actually I have a request.

leoyvens (approved):

stevepro-lab:
Hey @fordN  @leoyvens 
This fix is available in hosted service now?

PR #1889, opened 2020-09-03
store: Add administrative views

lutter:


leoyvens (approved):
Nice!

lutter:
I added `with no data` to the materialized views so we don't have to wait for several minutes for the materialized views to build in the hosted service.

It would be nice if we had the ability to periodically run background tasks, like refreshing these views, and some other stuff.

PR #1890, opened 2020-09-04
Move tracking the cache status of queries to a Prometheus metric

lutter:
We currently use a Google cloud log-based metric, but that is both wasteful and seems to be inaccurate. With this PR, it is still possible to log the cache status of queries, but it will also be recorded in a Prometheus metric so that log analysis should in general not be necessary.

### DIFF
--- a/graph/src/data/graphql/effort.rs
+++ b/graph/src/data/graphql/effort.rs
@@ -5,14 +5,15 @@ use lazy_static::lazy_static;
 use rand::{prelude::Rng, thread_rng};
 use std::collections::{HashMap, HashSet};
 use std::env;
+use std::iter::FromIterator;
 use std::str::FromStr;
 use std::sync::{Arc, RwLock};
 use std::time::{Duration, Instant};
 
-use crate::components::metrics::{Gauge, MetricsRegistry};
+use crate::components::metrics::{Counter, Gauge, MetricsRegistry};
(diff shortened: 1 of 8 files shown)


graph/src/data/graphql/effort.rs
```diff
@@ -276,12 +292,19 @@ impl LoadManager {
...
     }
 
-    pub fn add_query(&self, shape_hash: u64, duration: Duration) {
+    /// Record that we spent `duration` amount of work for the query
+    /// `shape_hash`, where `cache_status` indicates whether the query
+    /// was cached or had to actually run
+    pub fn record_work(&self, shape_hash: u64, duration: Duration, cache_status: CacheStatus) {
         if !*LOAD_MANAGEMENT_DISABLED {
             self.effort.add(shape_hash, duration, &self.effort_gauge);
+            self.query_counters
```
That3Percent, line 305: I feel like this logging should be irrespective of LOAD_MANAGEMENT_DISABLED which is for the feature that rejects queries when under load. I think you would want to use the counters regardless of whether we turn that off. So, just move outside the `if` block.

That3Percent (approved):

graph/src/data/graphql/effort.rs
lutter: Good point. Made the counting unconditional.

PR #1903, opened 2020-09-14
Don't crash node and show better error on invalid subgraph transitions

Jannis:
When a subgraph is indexed by two or more graph nodes that share a database, we're currently crashing nodes when they run into from/to block conflicts. What we should do (and what this PR does) is only make the subgraph fail and also print a message that points out the likely cause of the issue: that multiple nodes are indexing the same subgraph.

### DIFF
--- a/graph/src/components/store.rs
+++ b/graph/src/components/store.rs
@@ -741,11 +741,11 @@ pub enum StoreError {
     #[fail(display = "invalid identifier: {}", _0)]
     InvalidIdentifier(String),
     #[fail(
-        display = "subgraph `{}`: invalid transition from block `{}` to `{}`; \
+        display = "subgraph `{}` has already processed block `{}`; \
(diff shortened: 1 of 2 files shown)


lutter (changes requested):
This all looks great. My only issue is calling this a 'subgraph transition' which makes me think subgraphs move through various states. What we are really complaining about is that we are trying to process blocks again (or, more precisely, write the result of processing blocks) that we were already done with.

Can we call this something like `DuplicateBlockIngestion` or `DuplicateBlockProcessing` instead?

Jannis:
@lutter Totally. Let me update.

Jannis:
@lutter Updated. How's this?
https://github.com/graphprotocol/graph-node/pull/1903/files#diff-02527213a13f32ef41afb0e0e7e7df5dR743-R748

lutter (approved):

PR #1904, opened 2020-09-15
Ensure runtime host uniqueness, improve logging

fordN:
This PR adds some safety and clarity around the creating and usage of dynamic data sources. Subgraph developers have been susceptible to silent cases of unexpected behavior due to the ability to create multiple runtime hosts with identical triggers and data transformations.  In practice this can lead to a situation where triggers are processed through a mapping multiple times because several runtime hosts have been created with the same contract address, trigger, and mapping file. 

Now there is a check when creating a runtime host to ensure we are not instantiating any duplicate hosts and emit a warning. Additional logging has also been added to more clearly show the trigger processing path; a runtime_host context is now included which shows the number of matching runtime hosts for the trigger and the index of the current runtime host doing the processing.

### DIFF
--- a/core/src/subgraph/instance.rs
+++ b/core/src/subgraph/instance.rs
@@ -272,7 +272,7 @@ where
         data_source: DataSource,
         top_level_templates: Arc<Vec<DataSourceTemplate>>,
         metrics: Arc<HostMetrics>,
-    ) -> Result<Arc<T::Host>, anyhow::Error> {
+    ) -> Result<Option<Arc<T::Host>>, anyhow::Error> {
         // Protect against creating more than the allowed maximum number of data sources
         if let Some(max_data_sources) = *MAX_DATA_SOURCES {
             if self.hosts.len() >= max_data_sources {
@@ -289,7 +289,12 @@ where
             top_level_templates,
             metrics.clone(),
         )?);
-        self.hosts.push(host.clone());
-        Ok(host)
+
+        Ok(if self.hosts.contains(&host) {
+            None
+        } else {
+            self.hosts.push(host.clone());
+            Some(host)
+        })
     }
 }
--- a/core/src/subgraph/instance.rs
+++ b/core/src/subgraph/instance.rs
@@ -8,7 +8,7 @@ use std::str::FromStr;
 
 use graph::components::subgraph::SharedProofOfIndexing;
 use graph::prelude::{SubgraphInstance as SubgraphInstanceTrait, *};
-use web3::types::Log;
+use web3::types::{Log, H256};
 
 lazy_static! {
     static ref MAX_DATA_SOURCES: Option<usize> = env::var("GRAPH_SUBGRAPH_MAX_DATA_SOURCES")
@@ -168,13 +168,27 @@ where
                     .map(Arc::new)
                     .context("Found no transaction for event")?;
                 let matching_hosts = hosts.iter().filter(|host| host.matches_log(&log));
+                let hosts_count = matching_hosts.clone().count();
+
+                if hosts_count > 1 {
+                    info!(
+                        logger,
+                        "{} matching runtime hosts found for log trigger.", hosts_count;
+                        "address" => &log.address.to_string(),
+                        "topic0" => &log.topics.iter().next().unwrap_or(&H256::zero()).to_string(),
+                        "transaction" => log.transaction_hash.unwrap_or(H256::zero()).to_string(),
+                    );
+                }
+
                 // Process the log in each host in the same order the corresponding data
                 // sources appear in the subgraph manifest
                 let transaction = Arc::new(transaction);
-                for host in matching_hosts {
+                for (i, host) in matching_hosts.enumerate() {
+                    let host_context = format!("{}/{}", i + 1, hosts_count);
+                    let logger = logger.new(o!("runtime_host" => host_context));
                     state = host
                         .process_log(
-                            logger,
+                            &logger,
                             block,
                             &transaction,
                             &log,
@@ -192,11 +206,24 @@ where
                     .context("Found no transaction for call")?;
                 let transaction = Arc::new(transaction);
                 let matching_hosts = hosts.iter().filter(|host| host.matches_call(&call));
+                let hosts_count = matching_hosts.clone().count();
 
-                for host in matching_hosts {
+                if hosts_count > 1 {
+                    info!(
+                        logger,
+                        "{} matching runtime hosts found for call trigger.", hosts_count;
+                        "from" => &call.from.to_string(),
+                        "to" => &call.to.to_string(),
+                        "transaction" => &call.transaction_hash.map(|hash| hash.to_string()).unwrap_or("unkown".to_string()),
+                    );
+                }
+
+                for (i, host) in matching_hosts.enumerate() {
+                    let host_context = format!("{}/{}", i + 1, hosts_count);
+                    let logger = logger.new(o!("runtime_host" => host_context));
                     state = host
                         .process_call(
-                            logger,
+                            &logger,
                             block,
                             &transaction,
                             &call,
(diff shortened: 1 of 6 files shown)


core/src/subgraph/instance.rs
```diff
@@ -168,13 +168,27 @@ where
                     .map(Arc::new)
                     .context("Found no transaction for event")?;
                 let matching_hosts = hosts.iter().filter(|host| host.matches_log(&log));
+                let hosts_count = matching_hosts.clone().count();
```
Jannis, line 171: Iterators are a good candidate for CheapClone. This reads like we're coming the runtime hosts, but really, we're just cleaning the iterator. CheapClone would make this more obvious. Our a comment.

core/src/subgraph/instance.rs
```diff
@@ -192,11 +206,24 @@ where
                     .context("Found no transaction for call")?;
                 let transaction = Arc::new(transaction);
                 let matching_hosts = hosts.iter().filter(|host| host.matches_call(&call));
+                let hosts_count = matching_hosts.clone().count();
+
+                if hosts_count > 1 {
+                    info!(
```
Jannis, line 212: Feels more like a debug message.

core/src/subgraph/instance.rs
```diff
@@ -168,13 +168,27 @@ where
                     .map(Arc::new)
                     .context("Found no transaction for event")?;
                 let matching_hosts = hosts.iter().filter(|host| host.matches_log(&log));
+                let hosts_count = matching_hosts.clone().count();
+
+                if hosts_count > 1 {
+                    info!(
```
Jannis, line 174: I'd prefer debug here.

core/src/subgraph/instance.rs
```diff
@@ -210,10 +237,23 @@ where
                 let matching_hosts = hosts
                     .iter()
                     .filter(|host| host.matches_block(&trigger_type, ptr.number));
-                for host in matching_hosts {
+                let hosts_count = matching_hosts.clone().count();
+
+                if hosts_count > 1 {
+                    info!(
```
Jannis, line 243: debug

core/src/subgraph/instance.rs
```diff
@@ -168,13 +168,27 @@ where
                     .map(Arc::new)
                     .context("Found no transaction for event")?;
                 let matching_hosts = hosts.iter().filter(|host| host.matches_log(&log));
+                let hosts_count = matching_hosts.clone().count();
+
+                if hosts_count > 1 {
+                    info!(
+                        logger,
+                        "{} matching runtime hosts found for log trigger.", hosts_count;
```
Jannis, line 176: Remove the full stop at the end of the message

core/src/subgraph/instance.rs
```diff
@@ -192,11 +206,24 @@ where
                     .context("Found no transaction for call")?;
                 let transaction = Arc::new(transaction);
                 let matching_hosts = hosts.iter().filter(|host| host.matches_call(&call));
+                let hosts_count = matching_hosts.clone().count();
+
+                if hosts_count > 1 {
+                    info!(
+                        logger,
+                        "{} matching runtime hosts found for call trigger.", hosts_count;
```
Jannis, line 214: Remove full stop

core/src/subgraph/instance.rs
```diff
@@ -210,10 +237,23 @@ where
                 let matching_hosts = hosts
                     .iter()
                     .filter(|host| host.matches_block(&trigger_type, ptr.number));
-                for host in matching_hosts {
+                let hosts_count = matching_hosts.clone().count();
+
+                if hosts_count > 1 {
+                    info!(
+                        logger,
+                        "{} matching runtime hosts found for block trigger.", hosts_count;
```
Jannis, line 245: Remove full stop

core/src/subgraph/instance_manager.rs
```diff
@@ -941,8 +941,22 @@ where
...
-        data_sources.push(data_source);
-        runtime_hosts.push(host);
+        match host {
+            Some(host) => {
+                data_sources.push(data_source);
+                runtime_hosts.push(host);
+            }
+            None => warn!(
+                logger,
+                "no runtime hosted created, there is already a runtime host instantiated for \
```
Jannis, line 951: Log messages start with a capital letter

runtime/wasm/src/host.rs
```diff
@@ -718,3 +719,16 @@ impl RuntimeHostTrait for RuntimeHost {
         .await
     }
 }
+
+impl PartialEq for RuntimeHost {
+    fn eq(&self, other: &Self) -> bool {
+        // mapping_request_sender, host_exports and host_metrics are operational structs not needed
+        // to define uniqueness; each runtime host should be for a unique data source.
+        self.data_source_name == other.data_source_name
```
Jannis, line 727: I wonder if it isn't already enough to just compare the data source names. I believe those are guaranteed to be unique. Probably good to not rely on that though.

Jannis (changes requested):

runtime/wasm/src/host.rs
fordN: Based on my sleuthing around the database, I don't think name uniqueness is enforced for dynamic data sources (standard data source names _are_ unique.)

core/src/subgraph/instance.rs
fordN: Yeah, it's just cloning the iterator and then it's dropped immediately after counting.  I'll cheap clone and add a comment to make it clear.

core/src/subgraph/instance_manager.rs
```diff
@@ -970,6 +984,13 @@ where
     // Add entity operations to the block state in order to persist
     // the dynamic data sources
     for data_source in data_sources.iter() {
+        debug!(
+            logger,
+            "Persisting data_source";
+            "name" => &data_source.name,
+            "address" => &data_source.source.address.map(|address| address.to_string()).unwrap_or("none".to_string()),
+            "abi" => &data_source.source.abi,
```
leoyvens, line 992: Does this print the entire ABI? That seems very verbose.

runtime/wasm/src/host.rs
```diff
@@ -718,3 +719,16 @@ impl RuntimeHostTrait for RuntimeHost {
...
+            && self.data_source_contract == other.data_source_contract
+            && self.data_source_contract_abi == other.data_source_contract_abi
+            && self.data_source_event_handlers == other.data_source_event_handlers
+            && self.data_source_call_handlers == other.data_source_call_handlers
+            && self.data_source_block_handlers == other.data_source_block_handlers
```
leoyvens, line 732: Destructure `self` before comparing, to prevent this becoming incorrect when new fields are added.

leoyvens (changes requested):

leoyvens:
This is not taking the data source context into account. Two data sources that are otherwise identical can differ in behavior due to having a different context. It's questionable whether this is an useful pattern, but if in doubt I think we should allow it.

core/src/subgraph/instance_manager.rs
fordN: Great point; I'll remove "abi" from this log message.

fordN:
@leoyvens I see your point and in principle would like to allow a subgraph developer to instantiate duplicate data sources if they for some reason need to. However, I think including data source context in the definition of uniqueness here is more likely to shoot someone in the foot than provide any useful functionality.  I'd rather prohibit duplicate runtime host functionality altogether right now to be safe  

I can't think of a reason one would need two data sources triggering from the same event, running the same mapping handlers, and updating the same entities.  On the other hand, data source contexts are likely to be used to simply add useful labels to a data sources without intending for this to be a potential loophole to the duplicate functionality check.

leoyvens:
@fordN I might be partial to your argument if I hadn't just come across Livepeer using exactly this pattern https://github.com/livepeer/livepeerjs/blob/as/shares/packages/subgraph/src/mappings/bondingManager_LIP11.ts#L54.

leoyvens (approved):

Jannis (approved):

PR #1905, opened 2020-09-18
Eth call by hash

leoyvens:
This makes calls by block hash instead of block number. The tricky part is that contract calls may no longer be retried indefinitely, so if the eth node returns an error after 10 retries, we consider that a possible reorg. Getting that information returned to the instance manager requires some plumbing. The instance manager will then reset the block stream so we may handle the reorg.

There is one edge case which is when re-processing a block for new data sources. At this point we have a dirty indexing context due to the new data sources, and reverting that isn't a straightforward task, so if we fail to make a contract call that is treated as an ordinary error and not retried. Ideally we'd be able to retry this case as well, but that's left as future work.

I could not test this handling an actual reorg because that's not easy to reproduce. But I did test that when forcing an invalid hash on the call, the block stream would be reset.

### DIFF
--- a/chain/ethereum/src/ethereum_adapter.rs
+++ b/chain/ethereum/src/ethereum_adapter.rs
@@ -24,6 +24,7 @@ pub struct EthereumAdapter<T: web3::Transport> {
     url_hostname: Arc<String>,
     web3: Arc<Web3<T>>,
     metrics: Arc<ProviderEthRpcMetrics>,
+    is_ganache: bool,
 }
 
 lazy_static! {
@@ -70,6 +71,7 @@ impl<T: web3::Transport> CheapClone for EthereumAdapter<T> {
             url_hostname: self.url_hostname.cheap_clone(),
             web3: self.web3.cheap_clone(),
             metrics: self.metrics.cheap_clone(),
+            is_ganache: self.is_ganache,
         }
     }
 }
@@ -80,17 +82,31 @@ where
     T::Batch: Send,
     T::Out: Send,
 {
-    pub fn new(url: &str, transport: T, provider_metrics: Arc<ProviderEthRpcMetrics>) -> Self {
+    pub async fn new(
+        url: &str,
+        transport: T,
+        provider_metrics: Arc<ProviderEthRpcMetrics>,
+    ) -> Self {
         // Unwrap: The transport was constructed with this url, so it is valid and has a host.
         let hostname = graph::url::Url::parse(url)
             .unwrap()
             .host_str()
             .unwrap()
             .to_string();
+
+        let web3 = Arc::new(Web3::new(transport));
+        let is_ganache = web3
+            .web3()
+            .client_version()
+            .compat()
+            .await
+            .unwrap()
+            .contains("TestRPC");
         EthereumAdapter {
             url_hostname: Arc::new(hostname),
-            web3: Arc::new(Web3::new(transport)),
+            web3,
             metrics: provider_metrics,
+            is_ganache,
         }
     }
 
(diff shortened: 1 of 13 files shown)


chain/ethereum/src/ethereum_adapter.rs
```diff
@@ -374,171 +390,148 @@ where
...
-            .when(|result: &Result<Bytes, _>| {
-                match result {
-                    // Retry only if zero-length response received
-                    Ok(bytes) => bytes.0.is_empty(),
-
-                    // Errors are retried in the inner retry
-                    Err(_) => false,
-                }
+        // Ganache does not support calls by block hash.
+        // See https://github.com/trufflesuite/ganache-cli/issues/745
```
Jannis, line 401: Perfect. Thanks for including a link to the issue. I've verified `eth_clientVersion` returns `...TestRPC...` so the distinction we're making here looks correct.

chain/ethereum/src/ethereum_adapter.rs
```diff
@@ -374,171 +390,148 @@ where
...
+
+        retry("eth_call RPC call", &logger)
+            .when(|result| match result {
+                Ok(_) | Err(EthereumContractCallError::Revert(_)) => false,
+                Err(_) => true,
             })
-            .limit(16)
-            .no_logging()
-            .no_timeout()
+            .limit(10)
```
Jannis, line 413: What impact does this have in situations where the Ethereum node is temporarily down? Do we still keep retrying indefinitely on e.g. network errors, or do subgraphs fail eventually, after 10 attempts?

core/src/subgraph/instance_manager.rs
```diff
@@ -653,18 +648,39 @@ where
...
+
+            // In case of a possible reorg, we want this function to do nothing and restart the
+            // block stream so it has a chance to detect the reorg.
+            //
+            // The `ctx` is unchanged at this point, except for having cleared the entity cache.
+            // Losing the cache is a bit annoying but not an issue for correctness.
+            //
+            // See also b21fa73b-6453-4340-99fb-1a78ec62efb1.
+            return Ok((ctx, true));
+        }
```
Jannis, line 682: Ok, this answers my previous question. We're bubbling the problem up and retrying the stream as a whole because we may not want to stick to the hash that keeps failing. Makes perfect sense.

core/src/subgraph/instance_manager.rs
```diff
@@ -739,6 +755,15 @@ where
                 proof_of_indexing.cheap_clone(),
             )
             .await
+            .map_err(|e| {
+                // This treats a `PossibleReorg` as an ordinary error which will fail the subgraph.
+                // This can cause an unecessary subgraph failure, to fix it we need to figure out a
+                // way to revert the effect of `create_dynamic_data_sources` so we may return a
+                // clean context as in b21fa73b-6453-4340-99fb-1a78ec62efb1.
```
Jannis, line 762: This feels like a tradeoff worth making to get the feature in. It also seems like something we should prioritize immediately after merging. I'm pretty sure the problem will happen, and it'll surface in a similar way to the current issues caused by calling by block.

runtime/wasm/src/host_exports.rs
```diff
@@ -304,12 +316,23 @@ impl HostExports {
                 info!(logger, "Contract call reverted"; "reason" => reason);
                 Ok(None)
             }
-            Err(e) => Err(anyhow::anyhow!(
+
+            // Any error reported by the Ethereum node could be due to the block no longer being on
+            // the main chain. This is very inespecific but we don't want to risk failing a
```
Jannis, line 321: Typo: inespecific -> unspecific?

leoyvens: Thanks!

core/src/subgraph/instance_manager.rs
leoyvens: Considering how these triggers are much less common, the chance of us hitting the bug here is proportionally lower. And this bug already doesn't come up everyday, so I wouldn't count on us seeing this happen. But I think the fix for this and for the duplicate trigger processing will be similar, so we can try to fix them together.

Jannis (approved):

PR #1910, opened 2020-09-23
Optional full eth_call logging

leoyvens:
This is useful for debugging, especially when suspecting that the issue is with the Ethereum provider, since it allows fully reproducing the eth_call request.


lutter (approved):
Could we get that information out of the `eth_call_cache` in the database? If not, I am fine with this, though we have to be careful to not run into the logging limit in hosted.

leoyvens:
@lutter This is not included in the cache, it's hashed into the key. I don't intend for this to be always turned on, only when some specific issue needs to be investigated.

PR #1913, opened 2020-09-24
Facility for background maintenance jobs

lutter:
This creates a facility for background maintenance jobs that the block ingestor performs. It is meant for things that need to be done fairly infrequently, and that can tolerate a somewhat loose timeline.

For now, the job runner only has one very simple job, vacuuming `subgraph_deployment` frequently in each shard.

There's one mystery: the unit test in `graph::util::jobs` fails, even though running a `graph-node` locally with the vacuum job works, and I can see the `vacuum` statement in Postgres logs. I would highly appreciate any help with why the unit test does not run the job, even though from debug printing, it does call `CounterJob.run`


graph/src/util/jobs.rs
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    // Can't figure out what is wrong with this test: the test runner logs
+    // "Running job", but this test never sees the updates to `count`.
+    // Running graph-node, I can see that the store's VacuumDeploymentJob
+    // does get run with the configured interval, and the corresponding SQL
+    // statement turns up in the Postgres logs
+    #[ignore]
+    #[tokio::test]
+    async fn jobs_run() {
```
leoyvens, line 125: Don't know exactly what's going on, but try `#[tokio::test(threaded_scheduler)]`.

lutter: Nice. That did the trick. I updated the PR

node/src/main.rs
```diff
@@ -317,6 +317,11 @@ async fn main() {
                     network_store.block_store(),
                     &logger_factory,
                 );
+
+                // Start a task runner
+                let mut job_runner = graph::util::jobs::Runner::new(&logger);
+                register_store_jobs(&mut job_runner, network_store.clone());
+                graph::spawn_blocking(job_runner.start());
```
leoyvens, line 324: The use of `graph::spawn_blocking` should be avoided. If the runner is for futures, and therefore should not be blocking, then this should be just `spawn`.

leoyvens (changes requested):
Fancy stuff!

node/src/main.rs
lutter: This is intentional: these jobs will all (or almost all) interact with the database. Just throwing everything wholesale into the blocking pool makes it easier to write those jobs.

leoyvens: I see so even though the vacuum uses `with_conn` you don't want to promise they all will.

leoyvens (approved):

PR #1923, opened 2020-10-01
Skip trace_filter on empty block

### LINKED ISSUE #1920: Empty blocks on xDai cause non-fatal error on trace request
Bug

Empty blocks on xDai throw a non-fatal error upon trace request:
image
Example block: https://blockscout.com/poa/xdai/blocks/12275783/transactions
This halts processing, which means that my trace-based subgraph will never fully sync (as it waits ~20 minutes to retry) https://thegraph.com/explorer/subgraph/azf20/nifty-ink

See debug logs on the pending version: https://thegraph.com/explorer/subgraph/azf20/nifty-ink?version=pending
image

The absence of trace does not cause a problem, the subgraph is able to update every block and sync with the xDaichain

leoyvens:
Hopefully resolves #1920.

### DIFF
--- a/chain/ethereum/src/block_stream.rs
+++ b/chain/ethereum/src/block_stream.rs
@@ -491,7 +491,9 @@ where
                         // Note that head_ancestor is a child of subgraph_ptr.
                         let eth_adapter = self.eth_adapter.clone();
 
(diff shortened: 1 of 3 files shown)


fordN (approved):
👍

Issue #767, opened 2019-02-25
Develop convenient way to decode transaction inputs


Jannis:
Perhaps ethabi provides something here that we can expose to mappings. Its CLI provides convenient decoding of byte arrays: https://github.com/paritytech/ethabi#examples

If we know the parameter types to a function, we should be able to decode its inputs with e.g. https://docs.rs/ethabi/6.1.0/ethabi/fn.decode.html

Jannis:
@fordN This sounds good. I'd change the the generated code a little bit, not 100% decided what would be best though. But the approach feels solid.

leoyvens:
Thanks to @otaviopace we have this now!

PR #1933, opened 2020-10-07
runtime/wasm: add hosted abi encode/decode functions

jdkanani:
Related to https://github.com/graphprotocol/graph-node/issues/767

This PR will add support for `abi.encode` and `abi.decode` in hosted mapping functions. This can also be used to decode transaction data as mentioned by @fordN at https://github.com/graphprotocol/graph-node/issues/767 

Usage:

```ts
let encoded = abi.encode([
  ethereum.Value.fromUnsignedBigInt(BigInt.fromI32(64)),
  ethereum.Value.fromAddress(Address.fromString("0x0000000000000000000000000000000000000420"))
])

// log.info("Encoded data {}", [encoded.toHexString()])

let decoded = abi.decode(["uint256", "address"], encoded)

// log.info("Decoded data {}", [decoded[0].toBigInt().toString()])
// log.info("Decoded data {}", [decoded[1].toAddress().toHexString()])
```


Jannis:
@jdkanani I've been dreaming about someone contributing this for a long time! Thanks for the pull request 😄 

Regarding `abi.decode`: Isn't it limting to have only one flat array as the first argument? Considering that there are now deeply nested values (thanks to structs/tuples), I'm wondering if it wouldn't be better to have have something like

```ts
abi.decode("(uint256,address,(bytes32,int8,address))", [...values...])
```
where the types are passed in as a single string that is then parsed in Rust.

jdkanani:
@Jannis For nested tuples/structs, it will be like this as per https://github.com/openethereum/ethabi:

```js
// For tuple (tested)
abi.decode(["uint256", "address", "(bytes32,uint256,address)"], [...values....])

// For tuple array (tested)
abi.decode(["uint256", "address", "(bytes32,uint256,address)[]"], [...values....])
```

Jannis:
@jdkanani And that works with an arbitrary depth of tuples and arrays?

jdkanani:
@Jannis Seems like `ethabi` or https://github.com/graphprotocol/ethabi doesn't support it for now.

Check here: https://github.com/openethereum/ethabi/pull/186 -- It could be fully-fledged PR to ethabi repo. 

It shouldn't affect this PR and this PR can be merged.

jdkanani:
@Jannis quick question - does graph-node supports nested dynamic array/tuples?

Jannis:
> @Jannis quick question - does graph-node support nested dynamic array/tuples?

Yes it does!

Jannis:
I'm pretty sure our ethabi version supports arbitrary nesting.

hilmarx:
hey, any updates on supporting `abi.encode` and `abi.decode`? Would be really helpful for me :)

leoyvens:
@jdkanani are you still willing to work on this? We have an ethabi fork in which nested tuples should be supported, so I agree with @Jannis that the argument first argument should be a single type represented as a string. So instead of

```
abi.decode(["uint256", "address", "(bytes32,uint256,address)"], [...values....])
```

you'd have
```
abi.decode("(uint256,address,(bytes32,uint256,address))"], value)
```

We also need to add tests for this, ping me on Discord and I can give you pointers on how to add a test.

leoyvens:
We're closing in favor of #2348 which was based on this, thanks!

PR #1938, opened 2020-10-07
Consider empty list when constructing tsvectors

### LINKED ISSUE #1937: SQL syntax error when setting empty list to TSVector value

:bug: 

When the node attempts to add an empty list (which occurs when an optional field is being referenced but no value for that field is given) for a fulltext query, the subgraph fails, with the SQL reporting `2020-10-07 19:25:45.015 UTC [161] ERROR:  syntax error at or near ")" at character 2341`

With the following `schema.graphql`:

```graphql
type Thing @entity {
  id: ID!
  description: String
}

type _Schema_
  @fulltext(
    name: "search",
    language: en
    algorithm: rank,
    include: [
      {
        entity: "Thing",
        fields: [
          { name: "description" }
        ]
      }
    ]
  )
```

Make a Thing but don't set the `description` field.

The query constructed should not contain syntax errors.

cag:
Should resolve #1937

### DIFF
--- a/store/postgres/src/relational_queries.rs
+++ b/store/postgres/src/relational_queries.rs
@@ -492,14 +492,22 @@ impl<'a> QueryFragment<Pg> for QueryValue<'a> {
                     }
                     // TSVector will only be in a Value::List() for inserts so "to_tsvector" can always be used here
                     ColumnType::TSVector(config) => {
-                        out.push_sql("''::tsvector");
-                        for (i, value) in sql_values.iter().enumerate() {
-                            out.push_sql(" || to_tsvector(");
-                            out.push_bind_param::<Text, _>(&config.language.as_str().to_string())?;
-                            out.push_sql("::regconfig, ");
-                            out.push_bind_param::<Text, _>(&value)?;
-                            out.push_sql(")");
+                        if sql_values.is_empty() {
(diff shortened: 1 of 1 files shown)


cag:
Oh yeah I don't really know SQL so I hope that stuff works... testing it out on my local setup rn :crossed_fingers:

store/postgres/src/relational_queries.rs
```diff
@@ -492,17 +492,14 @@ impl<'a> QueryFragment<Pg> for QueryValue<'a> {
                     }
                     // TSVector will only be in a Value::List() for inserts so "to_tsvector" can always be used here
                     ColumnType::TSVector(config) => {
-                        out.push_sql("(");
+                        out.push_sql("''::tsvector");
```
lutter, line 495: Two quick comments:

1. Since this can be used in a larger query, can we keep the whole construct enclosed in parentheses?
2. I assume this is to deal with the case where `sql_values` is empty, right? It would be clearer if the code was structured as
```
if sql_values.is_empty() {
  out.push_sql("''::tsvector");
} else {
  .. what we did before this PR ..
}
```

lutter (approved):
Nice! Thanks for the contribution. I am happy to merge this once my comment has been addressd

PR #1947, opened 2020-10-13
Remove optional dependencies of `config`

leoyvens:
This removes a dependency on an old version of `linked-hash-map` which doesn't build on the latest Rust beta.


lutter (approved):
Nice!

PR #1948, opened 2020-10-14
Don't reenter the load manager semaphore

leoyvens:
That happened when the indexing status resolver called the store resolver, and could potentially deadlock.

In passing also fix getting `panic processing query: internal error: entered unreachable code` when querying the indexing status of a non-existing subgraph.

### DIFF
--- a/graph/src/components/graphql.rs
+++ b/graph/src/components/graphql.rs
@@ -20,7 +20,12 @@ pub type SubscriptionResultFuture =
 #[async_trait]
 pub trait GraphQlRunner: Send + Sync + 'static {
     /// Runs a GraphQL query and returns its result.
-    async fn run_query(self: Arc<Self>, query: Query, state: DeploymentState) -> Arc<QueryResult>;
+    async fn run_query(
+        self: Arc<Self>,
+        query: Query,
+        state: DeploymentState,
+        nested_resolver: bool,
+    ) -> Arc<QueryResult>;
 
     /// Runs a GraphqL query up to the given complexity. Overrides the global complexity limit.
     async fn run_query_with_complexity(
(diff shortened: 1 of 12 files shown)


graph/src/components/graphql.rs
```diff
@@ -19,7 +19,7 @@ pub type SubscriptionResultFuture =
 #[async_trait]
 pub trait GraphQlRunner: Send + Sync + 'static {
     /// Runs a GraphQL query and returns its result.
-    async fn run_query(self: Arc<Self>, query: Query) -> Arc<QueryResult>;
+    async fn run_query(self: Arc<Self>, query: Query, nested_resolver: bool) -> Arc<QueryResult>;
```
lutter, line 22: Is there any way to avoid forcing callers to keep track of whether we hold the semaphore or not? It feels like forcing callers to track what's essentially internal state of the runner is brittle and will be easy to get wrong. For example, could the runner contain mutable state that indicates whether it's holding the semaphore?

leoyvens: I don't think this is very robust either, but I couldn't come up with something better. The runner and the load manager are global resources, so it's normal for there to be many simultaneous calls runner holding permits. What we don't want is for that to happen in the same call stack. The question is how to mark a call stack, tokio has support for "task-local variables" but that doesn't work so well because queries are run in a blocking task, not in a future.

lutter: One option would be to split the `GraphQLRunner` functionality into two structs, one that just runs GraphQL queries, and one that has the extra functionality we need for user-supplied queries. Another would be to not use GraphQL queries internally (what the index-node resolver does could be easily written as an `EntityQuery`)

But neither of these need to be in the scope of this PR.

lutter (approved):
LGTM

PR #1950, opened 2020-10-14
graphql: Log details about subscription execution

lutter:
The logging is done in a way that `qlog` will ignore these log messages for now. But we need some indication of how slow subscription queries are in the logs


leoyvens (approved):

lutter:
This isn't needed anymore.

PR #1968, opened 2020-10-21
graphql: Use the same QueryStore for the entire query

lutter:
Make queries more consistent when replication is in use. We were using a
different QueryStore for each block constraint, and were also resolving
block pointers against the primary. This could lead to us mixing data from
different states of indexing a subgraph, without us even knowing.


leoyvens (approved):

lutter:
Rebased to latest master. You might want to have another look that this still makes sense.

leoyvens (approved):

PR #1974, opened 2020-10-27
Add a _meta field

lutter:
This PR makes it possible to include a `_meta` field at the toplevel of queries that reports the block against which the query was run and the deployment id. A query like
```
query {
  _meta {
    block {
      number
      hash
    }
    deployment
  }
  things(..) { }
}
```

will return the block number and hash of the latest block that the subgraph has indexed, as well as the deployment ID for the subgraph.

### DIFF
--- a/graphql/src/execution/execution.rs
+++ b/graphql/src/execution/execution.rs
@@ -537,7 +537,13 @@ pub async fn execute_root_selection_set<R: Resolver>(
         let mut cache = QUERY_BLOCK_CACHE.write().unwrap();
 
         // Get or insert the cache for this network.
-        if cache.insert(network, block_ptr, key, result.cheap_clone(), weight) {
+        if cache.insert(
+            network,
+            block_ptr.clone(),
+            key,
+            result.cheap_clone(),
+            weight,
+        ) {
             ctx.cache_status.store(CacheStatus::Insert);
         } else {
             // Results that are too old for the QUERY_BLOCK_CACHE go into the QUERY_LFU_CACHE
--- a/graphql/src/execution/execution.rs
+++ b/graphql/src/execution/execution.rs
@@ -1,6 +1,6 @@
 use super::cache::QueryCache;
 use crossbeam::atomic::AtomicCell;
-use graph::prelude::CheapClone;
+use graph::{data::schema::META_FIELD_NAME, prelude::CheapClone};
 use graphql_parser::query as q;
 use graphql_parser::schema as s;
 use indexmap::IndexMap;
@@ -380,6 +380,7 @@ pub fn execute_root_selection_set_uncached(
         span: selection_set.span.clone(),
         items: Vec::new(),
     };
+    let mut meta_items = Vec::new();
 
     for (_, fields) in collect_fields(ctx, root_type, iter::once(selection_set)) {
         let name = fields[0].name.clone();
@@ -389,16 +390,19 @@ pub fn execute_root_selection_set_uncached(
         // the data_set SelectionSet
         if is_introspection_field(&name) {
             intro_set.items.extend(selections)
+        } else if &name == META_FIELD_NAME {
+            meta_items.extend(selections)
         } else {
             data_set.items.extend(selections)
         }
     }
 
     // If we are getting regular data, prefetch it from the database
-    let mut values = if data_set.items.is_empty() {
+    let mut values = if data_set.items.is_empty() && meta_items.is_empty() {
         BTreeMap::default()
     } else {
-        let initial_data = ctx.resolver.prefetch(&ctx, selection_set)?;
+        let initial_data = ctx.resolver.prefetch(&ctx, &data_set)?;
+        data_set.items.extend(meta_items);
         execute_selection_set_to_map(&ctx, iter::once(&data_set), root_type, initial_data)?
     };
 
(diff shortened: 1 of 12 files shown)


graphql/src/store/resolver.rs
```diff
@@ -141,6 +155,45 @@ impl StoreResolver {
...
+
+    fn handle_meta(
+        &self,
+        prefetched_object: Option<q::Value>,
+        object_type: &ObjectOrInterface<'_>,
+    ) -> Result<(Option<q::Value>, Option<q::Value>), QueryExecutionError> {
+        // Pretend that the whole `_meta` field was loaded by prefetch. Eager
+        // loading this is ok until we add more information to this field
+        // that would force us to query the database; when that happens, we
+        // need to switch to loading on demand
```
leoyvens, line 167: Very soon I'll need to query the DB for subgraph errors for the meta field, where in the code would I do that? Do sneak a DB query into this function? Or does it go in the real prefetch code?

graphql/src/store/resolver.rs
```diff
@@ -141,6 +155,45 @@ impl StoreResolver {
...
+        if object_type.is_meta() {
+            let hash = self
+                .block_ptr
+                .and_then(|ptr| {
+                    if ptr.hash == web3::types::H256::zero() {
+                        None
```
leoyvens, line 173: Worth a comment and a `See also` linked by uuid to the place where we set the hash to zero.

leoyvens (changes requested):
Minor comments. Regarding the schema type names, I think we've set a precedent with `_Schema_`, so imo we should name them `_Meta_` and `_Block_`. Maybe @Jannis has an opinion.

graphql/src/execution/execution.rs
```diff
@@ -537,7 +541,13 @@ pub async fn execute_root_selection_set<R: Resolver>(
         let mut cache = QUERY_BLOCK_CACHE.write().unwrap();
 
         // Get or insert the cache for this network.
-        if cache.insert(network, block_ptr, key, result.cheap_clone(), weight) {
+        if cache.insert(
+            network,
+            block_ptr.clone(),
```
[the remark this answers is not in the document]
lutter: `EthereumBlockPointer` does not implement `CheapClone`, but I could add that; not sure if that is still within the intent of `CheapClone`

graphql/src/schema/api.rs
```diff
@@ -627,6 +651,27 @@ fn query_fields_for_type(schema: &Document, type_name: &Name) -> Vec<Field> {
     ]
 }
 
+fn meta_field() -> Field {
```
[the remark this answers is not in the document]
lutter: Moved that to a `lazy_static`

graphql/src/store/resolver.rs
lutter: That's the question I was trying to sidestep here - the thing is that we should avoid a database query when the user does not ask for fields that need database access. But it seems heinously hard to track robustly what fields are being accessed just based on the arguments that `resolve_object` gets - maybe for what you need you can have a special GraphQL type that you can use to trigger the query, similar to how the `MetaType` triggers this piece of code. I wouldn't bother putting it into prefetch, and just do the query straight from the resolver.

graphql/src/store/resolver.rs
lutter: Done

graphql/src/store/resolver.rs
leoyvens: I see, it needs to be a resolver on the field so not something we do here.

graphql/src/execution/execution.rs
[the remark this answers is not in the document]
leoyvens: It would be since the hash is fixed length, but it's optional.

leoyvens:
@lutter what's your opinion on the type names in the schema? See my comment above.

lutter:
> @lutter what's your opinion on the type names in the schema? See my comment above.

I'll defer to @Jannis on that - I am fine with either.

Jannis:
I'm also in favor of `_Meta_` and `_Block_`.

Jannis:
@lutter What's the behavior of `_meta { block { number } }` if there is a query that involves multiple blocks?

lutter:
> @lutter What's the behavior of `_meta { block { number } }` if there is a query that involves multiple blocks?

`_meta` behaves like any other toplevel field: it gives you the latest block for the subgraph. If you want info for an older block, you can add a block constraint like `_meta(block: { hash: "0xdeadbeef" }) {  .. }` The one caveat is that `_meta(block: {number: 17})` will not resolve to a hash since we don't have a reliable and fast way to translate a block number to a block hash for a subgraph

Jannis:
> > @lutter What's the behavior of `_meta { block { number } }` if there is a query that involves multiple blocks?
> 
> `_meta` behaves like any other toplevel field: it gives you the latest block for the subgraph. If you want info for an older block, you can add a block constraint like `_meta(block: { hash: "0xdeadbeef" }) { .. }` The one caveat is that `_meta(block: {number: 17})` will not resolve to a hash since we don't have a reliable and fast way to translate a block number to a block hash for a subgraph

Ah! That's good.

lutter:
Renamed the types - should be all good to go now.

leoyvens (approved):

lutter:
Rebased to latest master

leoyvens (approved):

PR #1975, opened 2020-10-27
docker: Make postgres_port configurable

adzialocha:
The current start script allows custom ports in postgres_host but still uses the default `5432` for checking if the database is available. This small fix introduces a new variable `postgres_port` which is `5432` by default, causing no breaking changes for previous docker setups.

Background: We're running a self-hosted graph node on a managed Digital Ocean database cluster which has a custom post.


lutter (approved):
Nice addition. Thanks!

PR #1976, opened 2020-10-27
Check trigger boundaries

leoyvens:
We've seen reports of subgraphs jumping to the future and past blocks being processed again, having these checks helps narrow the cause. This also asyncifies `blocks_with_triggers`.

### DIFF
--- a/graph/src/components/ethereum/adapter.rs
+++ b/graph/src/components/ethereum/adapter.rs
@@ -843,7 +843,6 @@ pub async fn triggers_in_block(
                 call_filter,
                 block_filter,
             )
-            .compat()
             .await?;
             assert!(blocks.len() <= 1);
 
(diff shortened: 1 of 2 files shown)


graph/src/components/ethereum/adapter.rs
```diff
@@ -943,63 +942,80 @@ pub fn blocks_with_triggers(
...
+        .await?;
+
+    blocks.sort_by_key(|block| block.ethereum_block.number());
+
+    // Sanity check that the returned blocks are in the correct range.
+    // Unwrap: `blocks` always includes at least `to`.
+    let first = blocks.first().unwrap().ethereum_block.number();
+    let last = blocks.last().unwrap().ethereum_block.number();
+    if first < from {
+        return Err(format_err!(
```
lutter, line 1004: I am not all that familiar with this code - what happens when we return an `Err` here? Does the subgraph fail? It might be better to trim `blocks` to the range we asked for and continue with that, and log an error about what we had to trim because it was outside the requested range.

leoyvens: I think it this case we should fail loudly, since it's an inconsistency in the Ethereum node response. Trimming would make sense but then we're likely to miss it if this ever happens.

lutter (approved):

PR #1977, opened 2020-10-27
Include network capabilities description for Ethereum network startup arguments

fordN:
An optional network capabilities specification has been supported for awhile now.   This PR updates the README and the arg descriptions in main to reflect this feature.


That3Percent (approved):
I see that `full` is the default, but is there a list of other values?

azf20:
Is this PR still valid?

leoyvens:
Closing because old and we prefer that this be configured in the config file.

PR #1980, opened 2020-10-28
node: Move argument parsing to StructOpt

lutter:


leoyvens (approved):
Awesome! Tsted this with my usual cli args and it worked fine.

lutter:
I tested by running `graph-node --help` and diffing the output before/after, and there were no real differences

PR #1981, opened 2020-10-28
Expose which types are affected by metadata operations to the store

lutter:


graph/src/data/subgraph/schema.rs
```diff
@@ -45,9 +46,48 @@ lazy_static! {
 pub const POI_TABLE: &str = "poi2$";
 pub const POI_OBJECT: &str = "Poi$";
 
+#[derive(Debug, Clone, IntoStaticStr)]
+pub enum MetadataType {
```
leoyvens, line 50: Niiiiiiice.

leoyvens (approved):
I'm in favor of everything that improves our metadata situation!

PR #1988, opened 2020-11-03
Reset block range size on error

leoyvens:
In case the Ethereum node is returning errors due to overly expensive requests, resetting the range size might help the subgraph recover from a retry loop. If we can infer that resetting the block range size worked, reduce the maximum block range size in an attempt to prevent the error from happening again.


lutter (approved):
This all looks good; I am a little concerned that we could get into a situation where subgraphs index slowly because the max range size has been reduced, but we have no good way to tell that that's what happened other than searching the logs.

leoyvens:
@lutter That's possible though we also have lower level retries which should get past spurious errors such as rate limiting. But we should watch how things go on the hosted service.

PR #1989, opened 2020-11-03
Deterministic errors

leoyvens:
This marks as determinstic few common reasons for subgraph failures, for example:
- An `unreachable` wasm trap which is caused by an array access out of bounds.
- An `abort` which AssemblyScript uses for multiple reasons including failed assert.
- Some uncontroversial host export erros such as division by zero.

There may be more cases we'd like to cover, but those can be added as long as we consider it a breaking change and bump the `specVersion` (or `apiVersion`?). Deterministic errors are flagged by a boolean in the metadata which can be accessed through the indexing status API.

### DIFF
--- a/core/src/subgraph/instance_manager.rs
+++ b/core/src/subgraph/instance_manager.rs
@@ -597,7 +597,7 @@ where
 }
 
 #[derive(thiserror::Error, Debug)]
-enum ProcessBlockError {
+enum BlockProcessingError {
     #[error("{0:#}")]
     Unknown(anyhow::Error),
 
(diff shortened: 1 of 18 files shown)


core/src/subgraph/instance_manager.rs
```diff
@@ -595,6 +596,30 @@ where
     }
 }
 
+#[derive(thiserror::Error, Debug)]
+enum ProcessBlockError {
```
lutter, line 600: To me, `BlockProcessingError` or even `IndexingError` would be clearer than `ProcessBlockError`

core/src/subgraph/instance_manager.rs
```diff
@@ -663,7 +688,12 @@ where
     .await
     {
         Ok(block_state) => block_state,
-        Err(MappingError::Unknown(e)) => return Err(e.compat_err().into()),
+        Err(MappingError::Unknown(e)) => {
+            return Err(CancelableError::Error(ProcessBlockError::Unknown(e)))
```
lutter, line 692: Is there any advantage to have separate `MappingError` and `ProcessBlockError`? Maybe we should just elevate `MappingError` to be used throughout?

lutter (approved):

core/src/subgraph/instance_manager.rs
leoyvens: Renamed to `BlockProcessingError`.

core/src/subgraph/instance_manager.rs
leoyvens: `ProcessBlockError` doesn't have the `PossibleReorg` variant.

lutter: But for that one, we know that it's not deterministic for sure.

leoyvens: I just see it as an advantage to have a fine grained types with less variants, and I don't see what code would be greatly simplified by having a single type.

PR #1991, opened 2020-11-04
metadata: Dont duplicate errors

leoyvens:
Right now on every graph node restart a new error entity is created for every failed subgraph, which is an unnecessary amount of bloat. Use the hash of the contents as the error id so that they are no longer duplicated.


lutter (approved):
Nice! Using the semantics of `Set` for good :)

PR #1994, opened 2020-11-06
Remove JSONB support

lutter:
The PR we've all been waiting for: remove support for JSONB storage and simplify the code because we do not need to accommodate two different storage schemes.

Note that this PR applies on top of #1992


store/postgres/src/entities.rs
```diff
@@ -666,7 +456,7 @@ impl Connection<'_> {
         // 5 minutes
         const MIGRATION_CHECK_FREQ: u64 = 20;
 
-        if self.storage.needs_migrating() {
+        if false {
```
leoyvens, line 459: `if false`?

store/postgres/src/entities.rs
```diff
@@ -700,7 +490,7 @@ impl Connection<'_> {
     ) -> Result<bool, Error> {
```
leoyvens, line 490: The return type can be changed to `()` I believe.

leoyvens (approved):
This is the most beautiful PR I've ever reviewed!

store/postgres/src/entities.rs
lutter: I didn't quite want to do away with the code for subgraph migrations when I wrote this; but commit 6e5424f3 in `lutter/shard` will do away with this nonsense

store/postgres/src/entities.rs
lutter: Likewise, that same commit will unceremoniously yank all that code.

lutter:
> This is the most beautiful PR I've ever reviewed!

Haha .. yes, nothing better than deleting 2k lines of code

PR #2001, opened 2020-11-10
Remove note to self

leoyvens:
Don't ask


lutter (approved):
I am somewhat sad to see this go.

PR #2002, opened 2020-11-10
store: Fix migration to add unique constraint on subgraph(name)

lutter:


leoyvens (approved):
Ah now I get it, thanks for the comment!

PR #2008, opened 2020-11-14
Cherry pick unfail logic

leoyvens:
This is from lutter/shard, but it's also very useful for the non-fatal errors work so it's worth merging these separately.


lutter (approved):
LGTM ;)

PR #2009, opened 2020-11-14
Cargo.lock: Revert ipfs-api update

leoyvens:
It's causing problems with the IPFS version we currently use with graph-node.


lutter (approved):
LGTM. Why does this only touch `Cargo.lock` though?

leoyvens:
@lutter good point this would easily get reverted if I left it only on the lock file, I've frozen the version on the tomls as well.

PR #2013, opened 2020-11-18
store: Do not unnecessarily allocate a txid in deployment_synced

lutter:


leoyvens (approved):

lutter:
@leoyvens I added another commit to update the `wasmtime` dependency so Travis will hopefully not fail. Could you let me know that that is the right thing to do?

lutter:
Urgh .. just realized that a fix was already committed for `wasmtime`; just got rid of the additional commit again.

PR #2021, opened 2020-11-23
Load dynamic data sources without using GraphQL

lutter:
With this change, we load dynamic data sources by directly querying key data from the database and using the manifest we already have in memory to fill in the remainder of the data source.

This avoids using GraphQL, and also does not resolve data sources against IPFS redundantly, which should lead to a considerable speed up when starting subgraphs with a lot of dynamic data sources.

This branch sits on top of the `lutter/indexing` branch; since it touches parts of the system that are complex and very seinsitive to maintain correctness, I'd appreciate feedback on whether something about this approach is obviously wrong.

### DIFF
--- a/core/src/subgraph/loader.rs
+++ b/core/src/subgraph/loader.rs
@@ -4,207 +4,71 @@ use std::time::Instant;
 
 use async_trait::async_trait;
 
-use graph::data::subgraph::schema::*;
-use graph::data::subgraph::UnresolvedDataSource;
-use graph::prelude::{DataSourceLoader as DataSourceLoaderTrait, GraphQlRunner, *};
-use graph_graphql::graphql_parser::{parse_query, query as q};
+use graph::components::store::StoredDynamicDataSource;
+use graph::prelude::{DataSourceLoader as DataSourceLoaderTrait, *};
 
-pub struct DataSourceLoader<L, Q, S> {
+pub struct DataSourceLoader<S> {
     store: Arc<S>,
-    link_resolver: Arc<L>,
-    graphql_runner: Arc<Q>,
 }
 
-impl<L, Q, S> DataSourceLoader<L, Q, S>
+impl<S> DataSourceLoader<S>
 where
-    L: LinkResolver,
     S: Store + SubgraphDeploymentStore,
-    Q: GraphQlRunner,
 {
-    pub fn new(store: Arc<S>, link_resolver: Arc<L>, graphql_runner: Arc<Q>) -> Self {
-        Self {
-            store,
-            link_resolver,
-            graphql_runner,
-        }
-    }
-
-    fn dynamic_data_sources_query(
-        &self,
-        deployment: &SubgraphDeploymentId,
-        skip: i32,
-    ) -> Result<Query, Error> {
-        // Obtain the "subgraphs" schema
-        let schema = self.store.api_schema(&SUBGRAPHS_ID)?;
-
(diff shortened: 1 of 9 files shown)


lutter:
No idea why this was closed. It should now target `master`

store/postgres/src/dynds.rs
```diff
@@ -0,0 +1,131 @@
...
+    dynamic_ethereum_contract_data_source,
+    ethereum_contract_source
+);
+
+fn to_source(
+    deployment: &str,
+    ds_id: &str,
+    (address, abi, start_block): (Option<Vec<u8>>, String, Option<BigDecimal>),
+) -> Result<Source, StoreError> {
+    // Treat a missing address as an error. TODO: Is that correct?
```
leoyvens, line 60: This is correct for dynamic data sources.

core/src/subgraph/loader.rs
```diff
@@ -4,207 +4,71 @@ use std::time::Instant;
...
-            let next_data_sources = self
-                .resolve_data_sources(unresolved_data_sources, &logger)
-                .await?;
-
-            if next_data_sources.is_empty() {
-                break;
-            }
-
-            data_sources.extend(next_data_sources);
+        for stored in self.store.load_dynamic_data_sources(&deployment_id)? {
```
leoyvens, line 44: Not great that this calls a blocking function from an `async` fn. Could we leverage `with_entity_conn` to make this an async store fn?

core/src/subgraph/loader.rs
```diff
@@ -4,207 +4,71 @@ use std::time::Instant;
...
+                .transpose()?;
+
+            let ds = DataSource {
+                kind: template.kind.clone(),
+                network: template.network.clone(),
+                name,
+                source,
+                mapping: template.mapping.clone(),
+                context,
+                templates: Vec::new(),
```
leoyvens, line 69: If we're doing this we should go ahead and remove the `templates` field from `DataSource`. Some old subgraphs might break, but now is our last opportunity to do this sort of breakage, so I'm in favor of going ahead.

leoyvens (changes requested):

store/postgres/src/dynds.rs
lutter: Updated comment

core/src/subgraph/loader.rs
lutter: I just tried, and ran into the issue of going back and forth between `anyhow::Error` and `StoreError`. Can we do an overhaul of errors from the store and asyncifying this after we are done with the sharding PR I am going to open in a sec?

core/src/subgraph/loader.rs
lutter: There are 5 subgraphs that currently have templates in `ethereum_contract_data_source`:
```
name         | raisehq/raisekovan
name         | gvillarc91/gaston
name         | perfectmak/marketprotocol
name         | perfectmak/marketprotocol-kovan
name         | kohshiba/forecastory-mainnet
```
We probably need to talk to their owners to get them to redeploy.

leoyvens: This PR already breaks them, which I'm fine with. If you don't want to remove the templates field right now, could you put a comment above the field declaration it should be removed?

core/src/subgraph/loader.rs
leoyvens: Fine by me.

PR #2037, opened 2020-11-30
Continue indexing on deterministic error

leoyvens:
This implements the non-fatal errors proposal, by which a subgraph which has synced will continue indexing when encountering a deterministic error, skipping the handler and discarding any changes done while processing it. The errors will be registered when transacting operations for a block, and will be reverted if necessary. To enable non-fatal errors the manifest must include the section:
```
features:
  - nonFatalErrors
```

By default data is hidden from queries in the presence of indexing errors (except for the `_meta` field). To receive data in a query the `subgraphError: allow` argument must be passed in. See the `non_fatal_errors` test for examples of what the responses look like. The `_meta` field gains an `hasIndexingErorrs: Boolean!` field as a way of checking for errors. Additionally if indexing errors are present a graphql error with the message `"indexing_error"` will always be added.

This is ready for review. The `nonFatalErrors` field of the indexing status API will be implemented in a follow up. I'd also like to add an integration test before merging this.

### DIFF
--- a/core/src/subgraph/instance_manager.rs
+++ b/core/src/subgraph/instance_manager.rs
@@ -1,7 +1,7 @@
 use atomic_refcell::AtomicRefCell;
 use futures01::sync::mpsc::{channel, Receiver, Sender};
 use lazy_static::lazy_static;
-use std::collections::HashMap;
+use std::collections::{BTreeSet, HashMap};
 use std::sync::{Arc, RwLock};
 use std::time::Instant;
 
@@ -31,6 +31,7 @@ type SharedInstanceKeepAliveMap = Arc<RwLock<HashMap<SubgraphDeploymentId, Cance
 
 struct IndexingInputs<B, S> {
     deployment_id: SubgraphDeploymentId,
+    features: BTreeSet<SubgraphFeatures>,
     network_name: String,
     start_blocks: Vec<u64>,
     store: Arc<S>,
@@ -359,6 +360,7 @@ impl SubgraphInstanceManager {
             &deployment_id,
             stopwatch_metrics,
         ));
+        let features = manifest.features.clone();
         let instance =
             SubgraphInstance::from_manifest(&logger, manifest, host_builder, host_metrics.clone())?;
 
@@ -366,6 +368,7 @@ impl SubgraphInstanceManager {
         let ctx = IndexingContext {
             inputs: IndexingInputs {
                 deployment_id: deployment_id.clone(),
+                features,
                 network_name,
                 start_blocks,
                 store,
@@ -662,7 +665,25 @@ where
     )
     .await
     {
+        // The triggers were processed but some were skipped due to deterministic errors.
+        Ok(block_state) if block_state.has_errors() => {
+            use SubgraphFeatures::*;
+
+            // TODO: || subgraph still syncing
+            if !ctx.inputs.features.contains(&nonFatalErrors) {
+                // Take just the first error to report.
+                return Err(CancelableError::Error(BlockProcessingError::Deterministic(
+                    block_state.deterministic_errors.into_iter().next().unwrap(),
+                )));
+            }
+
+            block_state
+        }
+
+        // Triggers processed with no errors.
         Ok(block_state) => block_state,
+
+        // Some form of unknown or non-deterministic error ocurred.
         Err(MappingError::Unknown(e)) => {
             return Err(CancelableError::Error(BlockProcessingError::Unknown(e)))
         }
--- a/core/src/subgraph/instance_manager.rs
+++ b/core/src/subgraph/instance_manager.rs
@@ -732,7 +732,7 @@ where
 
     // If new data sources have been created, restart the subgraph after this block.
     // This is necessary to re-create the block stream.
-    let needs_restart = !block_state.created_data_sources.is_empty();
+    let needs_restart = block_state.has_created_data_sources();
     let host_metrics = ctx.host_metrics.clone();
 
     // This loop will:
@@ -743,13 +743,13 @@ where
     // Note that this algorithm processes data sources spawned on the same block _breadth
     // first_ on the tree implied by the parent-child relationship between data sources. Only a
     // very contrived subgraph would be able to observe this.
-    while !block_state.created_data_sources.is_empty() {
+    while block_state.has_created_data_sources() {
         // Instantiate dynamic data sources, removing them from the block state.
         let (data_sources, runtime_hosts) = create_dynamic_data_sources(
             logger.clone(),
             &mut ctx,
             host_metrics.clone(),
-            block_state.created_data_sources.split_off(0).into_iter(),
+            block_state.drain_created_data_sources(),
         )
         .compat_err()?;
 
@@ -789,7 +789,7 @@ where
             &mut block_state.entity_cache,
             data_sources,
             block_ptr_for_new_data_sources,
-        )?;
+        );
 
         // Process the triggers in each host in the same order the
         // corresponding data sources have been created.
@@ -933,7 +933,7 @@ async fn update_proof_of_indexing(
             digest: updated_proof_of_indexing,
         };
 
-        entity_cache.set(entity_key, new_poi_entity)?;
+        entity_cache.set(entity_key, new_poi_entity);
     }
 
     Ok(())
@@ -989,7 +989,7 @@ fn create_dynamic_data_sources<B, T: RuntimeHostBuilder, S>(
     logger: Logger,
     ctx: &mut IndexingContext<B, T, S>,
     host_metrics: Arc<HostMetrics>,
-    created_data_sources: impl Iterator<Item = DataSourceTemplateInfo>,
+    created_data_sources: Vec<DataSourceTemplateInfo>,
 ) -> Result<(Vec<DataSource>, Vec<Arc<T::Host>>), anyhow::Error>
 where
     B: BlockStreamBuilder,
@@ -1037,8 +1037,7 @@ fn persist_dynamic_data_sources<B, T: RuntimeHostBuilder, S>(
     entity_cache: &mut EntityCache,
     data_sources: Vec<DataSource>,
     block_ptr: EthereumBlockPointer,
-) -> Result<(), Error>
-where
+) where
     B: BlockStreamBuilder,
     S: ChainStore + Store,
 {
@@ -1066,7 +1065,7 @@ where
         ));
         let id = DynamicEthereumContractDataSourceEntity::make_id();
         let operations = entity.write_entity_operations(id.as_ref());
-        entity_cache.append(operations)?;
+        entity_cache.append(operations);
     }
 
     // Merge log filters from data sources into the block stream builder
@@ -1083,6 +1082,4 @@ where
     ctx.state
         .block_filter
         .extend(EthereumBlockFilter::from_data_sources(&data_sources));
-
-    Ok(())
 }
--- a/core/src/subgraph/instance_manager.rs
+++ b/core/src/subgraph/instance_manager.rs
@@ -583,8 +583,8 @@ enum BlockProcessingError {
 
     // The error had a determinstic cause but, for a possibly non-deterministic reason, we chose to
     // halt processing due to the error.
-    #[error("{0:#}")]
-    Deterministic(anyhow::Error),
+    #[error("{0}")]
+    Deterministic(SubgraphError),
 
     #[error("subgraph stopped while processing triggers")]
     Canceled,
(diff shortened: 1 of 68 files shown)


store/postgres/src/metadata.rs
```diff
@@ -883,3 +886,21 @@ fn check_health(conn: &PgConnection, id: &SubgraphDeploymentId) -> Result<(), an
...
+    conn: &PgConnection,
+    id: &SubgraphDeploymentId,
+    reverted_block: BlockNumber,
+) -> Result<(), anyhow::Error> {
+    use subgraph_error as e;
+
+    delete(
+        e::table
+            .filter(e::subgraph_id.eq(id.as_str()))
+            .filter(sql("lower(block_range) >= $1").bind::<Integer, _>(reverted_block)),
```
lutter: this should use `block_range::BLOCK_RANGE_COLUMN` as the name of the block range column

graph/src/components/store.rs
```diff
@@ -1282,6 +1282,7 @@ pub trait EthereumCallCache: Send + Sync + 'static {
 }
 
 /// Store operations used when serving queries
+#[async_trait]
```
lutter, line 1285: Similarly, let's please hold off on asyncifying store operations.

graph/src/data/query/error.rs
```diff
@@ -252,6 +252,12 @@ impl From<StoreError> for QueryExecutionError {
     }
 }
 
+impl From<anyhow::Error> for QueryExecutionError {
```
lutter, line 255: It seems weird that any `anyhow::Error` gets translated into a `QueryExecutionError::StoreError`

graphql/src/query/ext.rs
```diff
@@ -68,64 +69,22 @@ impl Default for BlockConstraint {
...
-                    Err(invalid_argument("block", self, value))
-                }
-            } else {
-                Err(invalid_argument("block", self, value))
-            }
+        if let Some(hash) = map.get("hash") {
+            Ok(BlockConstraint::Hash(TryFromValue::try_from_value(hash)?))
+        } else if let Some(number_value) = map.get("number") {
+            let number: u64 = TryFromValue::try_from_value(number_value)?;
+            Ok(BlockConstraint::Number(TryFrom::try_from(number)?))
```
lutter, line 85: Nice! No idea what drugs I was on when I initially wrote that code.

graphql/src/schema/api.rs
```diff
@@ -174,6 +216,26 @@ fn add_block_height_type(schema: &mut Document) {
     schema.definitions.push(def);
 }
 
+fn add_subgraph_error_policy_type(schema: &mut Document) {
```
lutter, line 219: Not sure if you noticed, but I added a `meta.graphql` file a while ago. Would be cleaner to add this type definition to that file, too, since a GraphQL schema definition is much easier to read than Rust code that builds the AST.

graphql/tests/query.rs
```diff
@@ -1637,3 +1640,27 @@ fn can_query_meta() {
...
+            deterministic: true,
+        };
+
+        transact_errors(&*STORE, id.clone(), BLOCK_TWO.block_ptr(), vec![err]).unwrap();
+
+        let query = "query { musician(id: \"m1\") { id } }";
+        let query = graphql_parser::parse_query(query).unwrap();
+
+        dbg!(execute_query_document(&id, query).await);
+        assert!(false);
```
lutter: Looks like this test needs just a little more work ;)

graph/src/data/query/result.rs
```diff
@@ -97,54 +172,65 @@ impl QueryResult {
...
+        serialize_with = "serialize_data"
+    )]
+    data: Option<Data>,
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    errors: Vec<QueryError>,
+    #[serde(
+        skip_serializing_if = "Option::is_none",
+        serialize_with = "serialize_extensions"
+    )]
+    extensions: Option<q::Value>,
```
lutter, line 191: We should really nuke `extensions` since it is always `None`

lutter (changes requested):

store/postgres/src/metadata.rs
```diff
@@ -751,6 +753,62 @@ pub fn reassign_subgraph(
...
+            e::subgraph_id.eq(subgraph_id.as_str()),
+            e::message.eq(message),
+            e::handler.eq(handler),
+            e::deterministic.eq(deterministic),
+            e::block_hash.eq(block_ptr.as_ref().map(|ptr| ptr.hash.as_bytes())),
+            e::block_number.eq(sql(&format!("{}::numeric", block_num))),
+            e::block_range.eq((Bound::Included(block_num), Bound::Unbounded)),
+        ))
+        .on_conflict_do_nothing()
+        .execute(conn)?;
```
lutter, line 795: You also need to clamp the range of precious errors, otherwise, if you insert an error with range `[1,)`, you will never be able to insert another error for that subgraph.

And since `block_number = lower(block_range)`, do we even need that column?

lutter (changes requested):

store/postgres/src/metadata.rs
lutter: Since `on_conflict_do_nothing()` can mask problems silently, we need tests that check the various combinations of storing errors and that they work; at a minimum, it should check
* storing the same error twice stores only one error
* storing the same error at two different blocks stores two errors
* reverting correctly reverts errors

And it would be really nice if errors had a `created_at` column to make it easier to find new errors.

graph/src/components/store.rs
```diff
@@ -1448,6 +1451,19 @@ impl EntityCache {
         Ok(())
     }
 
+    /// A copy of the pending updates. Those are kept in a persistent data structure so this is a
+    /// cheap operation.
+    pub(crate) fn updates_snapshot(&self) -> im::HashMap<EntityKey, Option<Entity>> {
+        self.updates.clone()
+    }
+
+    pub(crate) fn restore_updates_snapshot(
```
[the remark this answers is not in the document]
leoyvens: Good news is I implemented this without the persistent data structure, which really was bad for performance. Bad news is the code touched some tricky logic in the entity cache and is not super robust since it requires explicitly entering and exiting the handler. But I added and end-to-end test and plenty of asserts.

core/src/subgraph/instance_manager.rs
```diff
@@ -662,7 +665,25 @@ where
     )
     .await
     {
+        // The triggers were processed but some were skipped due to deterministic errors.
+        Ok(block_state) if block_state.has_errors() => {
+            use SubgraphFeatures::*;
+
+            // TODO: || subgraph still syncing
+            if !ctx.inputs.features.contains(&nonFatalErrors) {
```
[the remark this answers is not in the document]
leoyvens: Changed.

runtime/wasm/src/module/mod.rs
```diff
@@ -245,9 +245,9 @@ impl WasmInstance {
         };
 
         if let Some(deterministic_error) = deterministic_error {
-            // Log the error and restore the updates snaphsot, effectively skipping the handler.
+            // Log the error and restore the updates snaphsot, effectively reverting the handler.
             error!(&self.instance_ctx().ctx.logger,
-                "Handler skipped due to execution failure";
+                "Handler reverted";
```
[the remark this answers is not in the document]
leoyvens: Changed to the previous message.

core/src/subgraph/instance_manager.rs
```diff
@@ -25,6 +25,10 @@ lazy_static! {
             .unwrap_or("10000".into())
             .parse::<usize>()
             .expect("invalid GRAPH_ENTITY_CACHE_SIZE");
+
+    // Keep deterministic errors non-fatal even if the subgraph is pending. For test purposes.
+    pub static ref DISABLE_FAIL_FAST: bool =
+        std::env::var("GRAPH_DISABLE_FAIL_FAST").is_ok();
```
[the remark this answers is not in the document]
leoyvens: Yhea graph node itself, updated the comment.

store/postgres/src/network_store.rs
```diff
@@ -48,6 +51,7 @@ impl NetworkStore {
     }
 }
 
+#[async_trait::async_trait]
```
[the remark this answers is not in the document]
leoyvens: We agreed that this is ok.

store/postgres/src/network_store.rs
```diff
@@ -169,6 +173,14 @@ impl StoreTrait for NetworkStore {
         self.store.deployment_state_from_id(id)
     }
 
+    async fn fail_subgraph(
+        &self,
+        id: SubgraphDeploymentId,
+        error: SubgraphError,
+    ) -> Result<(), anyhow::Error> {
```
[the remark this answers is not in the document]
leoyvens: Yes it wasn't a great idea to use anyhow, I switched this and a bunch of other places to StoreError.

store/postgres/src/metadata.rs
leoyvens: Done.

graph/src/data/query/error.rs
leoyvens: Super weird, refactored this away.

graphql/src/schema/api.rs
leoyvens: That is nice! Moved it there.

graph/src/data/query/result.rs
leoyvens: Yay done

store/postgres/src/metadata.rs
leoyvens: Dropped `block_number`, added `created_at`, added some tests that the conflict do nothing works.

leoyvens:
@lutter I believe I've addressed all your comments, happy reviewing and rebasing!

graph/src/data/query/error.rs
```diff
@@ -252,11 +252,11 @@ impl From<StoreError> for QueryExecutionError {
...
-impl From<anyhow::Error> for QueryExecutionError {
-    fn from(e: anyhow::Error) -> Self {
-        QueryExecutionError::StoreError(CloneableFailureError(Arc::new(e.compat_err())))
-    }
-}
+// impl From<anyhow::Error> for QueryExecutionError {
+//     fn from(e: anyhow::Error) -> Self {
+//         QueryExecutionError::StoreError(CloneableFailureError(Arc::new(e.compat_err())))
+//     }
+// }
```
lutter: Leftover from trying things out? Should probably be removed

store/postgres/migrations/2020-12-7-190800_subgraph_error_changes/up.sql
```diff
@@ -0,0 +1,9 @@
+alter table
+    subgraphs.subgraph_error
+drop
+    column block_number;
+
+alter table
+    subgraphs.subgraph_error
+add
+    column created_at timestamp NOT NULL DEFAULT (now() at time zone 'utc');
```
lutter, line 9: The data type should be `timstamptz` so that [the timezone gets stored](https://www.postgresql.org/docs/9.6/datatype-datetime.html), too. Just `timestamp` defaults to `without time zone`. And I think you can just say `default now()` if we use a type with time zone.

store/postgres/src/metadata.rs
```diff
@@ -914,3 +914,66 @@ pub(crate) fn revert_subgraph_errors(
 
     check_health(conn, id)
 }
+
+#[test]
```
lutter, line 918: I am not against this, but I thought we put tests that require external components like the store into `tests/`?

store/postgres/src/entities.rs
```diff
@@ -773,6 +773,7 @@ pub fn delete_all_entities_for_test_use_only(
         delete from subgraphs.ethereum_contract_data_source_template;
         delete from subgraphs.ethereum_contract_data_source_template_source;
         delete from subgraphs.ethereum_contract_event_handler;
+        delete from subgraphs.subgraph_error;
```
lutter, line 776: Is this necessary? There's an `on delete cascade` on the foreign key `subgraph_error.subgraph_id`

store/postgres/src/metadata.rs
lutter: Actually, leave that here (fear of merge conflicts) - I can always move it later

lutter (approved):
Great! Once the tests pass, this is ready to merge.

store/postgres/src/entities.rs
leoyvens: Oh that's why we could get away without it. Removed it.

store/postgres/migrations/2020-12-7-190800_subgraph_error_changes/up.sql
leoyvens: Done.

PR #2038, opened 2020-12-01
WIP: Gas, error determinism, -failure, graphql-parser 0.3

That3Percent:
Introduces Gas. There's a low signal to noise ratio in the diff because basic changes required editing hundreds of disparate lines. I hope the PR boils down to these talking points:

**Gas**
   * All rates of consumption are in gas.rs for easy reference instead of at the consume_gas callsite 
   * Most gas consumption is intentionally very low.
   * Gas is an upper bound on what we are willing to verify attestations for, not what Indexers would allow on a normal basis.
   * Gas price for each individual element is the answer to the question "when does consumption of this resource become unreasonable"? Eg: Allocating bytes on the heap caps out at 12GiB.
   * The existing timeout handles the normal "this is too costly for normal operations" case, and because the timeout is a non-deterministic error this is OK. Other non-deterministic gas consumption modifiers are allowed if they create non-deterministic errors
   * Note that some future optimizations may be breaking changes if they consume less gas! This should be handled with careful versioning
   * Comparing gas consumption across updates before deploying should be a part of our automated determinism story in the future.
   * Now is a good time to bikeshed on values.

**Error Determinism**
   * Introduced determinism in error types
   * Before this would set deterministic_host_trap, but access to this value may not be available in many cases where the error originates or may use shared references forcing atomics
   * anyhow errors are necessarily considered non-deterministic until proven otherwise.
   * Anyhow is dangerous for our use-case because it hides information by design. (This extends to any use-case which handles errors) But it's too late to do anything about this.
   * Lots of anyhow errors have been converted to deterministic ones (eg: json parsing, type conversions, math, and most other host exports are deterministic)
   * Note that like anyhow, panics do not carry determinism information. These must be treated as non-deterministic and should be avoided for error conditions arising out of the subgraph or the data it acts upon.

**Non-panicky errors**
   * BigDecimal % 0 no longer panics but has `DeterministicHostError` instead
   * Before the wasm was trusted to interact with the heap in-bounds, leading to panics - which with this change is a DeterministicHostError instead
   * Same story for assertions on binary representations for AscValue which also are untrusted inputs

**failure -> thiserror & anyhow**
   * Part of a security audit because failure is deprecated
   * Made some of the deterministic error migration easier
   * _Finally_

**graphql-parser 0.3.0+patch**
   * Protection from stackoverflows for security
   * impl of Error instead of Fail to match anyhow (Fail doesn't impl Error)
   * Eventually better performance using &str as the Text type (Additional generics work to enable)

May GitHub have mercy on whoever has to review this.


lutter:
Can we hold off on this? This is such a vast change that rebasing what I have in the `lutter/shard` branch will be an absolute nightmare (and keeping that branch going has been a rebase nightmare already). I wish I would have known that we are planning on such an invasive change.

That3Percent:
> Can we hold off on this? This is such a vast change that rebasing what I have in the `lutter/shard` branch will be an absolute nightmare (and keeping that branch going has been a rebase nightmare already). I wish I would have known that we are planning on such an invasive change.

@lutter:
Yes, I can put this on hold and work on something else. Sorry for the surprise in the scope. None of the extras were planned, but came out of problems run into when adding gas that made a lot of previously infallible functions fallible since `consume_gas` could trigger an error and that needs to be called by everything.

Holding off a merge just moves the rebase nightmare from you to me, but it's probably better to do the rebase in this PR because many of the kinds of changes this PR makes can just be re-made easily after "accept theirs".

leoyvens:
@That3Percent does it make sense to keep this open?

That3Percent:
> @That3Percent does it make sense to keep this open?

We still have to finish and merge the last part - the actual gas changes. I've kept this open so that I can use the review changes tool to keep track of migration status on a per file level.

leoyvens:
Closing in favor of #2414.

PR #2040, opened 2020-12-04
Shard storage of subgraphs

lutter:
This PR makes it possible to store subgraph data in several database shards. The shards, their replicas and the rules for how subgraphs get deployed are configured in a TOML configuration file.

This changes the meaning of the `node_id` parameter of the `subgraph_deploy` JSON-RPC call: the `node_id` is only used as a fallback, and the rules in the configuration are consulted first to determine which node a subgraph should be assigned to.

The basic idea is that metadata is divided into store-wide metadata and per-subgraph metadata: store-wide metadata (`subgraph`, `subgraph_version`, `subgraph_deployment_assignment`, and `deployment_schemas`) is always stored in the primary, and deployment-specific metadata goes into the shard that holds the deployment data. The `deployment_schemas` table serves as a directory of deployments, and ensures that the internal id (`sgdNNNN`) of deployments is unique across all shards.

I am also subsuming PR #2021 into this one. This PR is on top of #2037 as it is expected that that will be committed before this PR. But the PR is ready for review in its current form.


store/postgres/src/entities.rs
```diff
@@ -191,105 +70,127 @@ pub(crate) fn make_storage_cache() -> StorageCache {
...
+        let data = ids_for_type
+            .iter()
+            .filter(|(typ, _)| typ.is_data_type())
+            .map(|(typ, ids)| (typ.as_str(), ids))
+            .collect();
+        let metadata = ids_for_type
+            .iter()
+            .filter(|(typ, _)| !typ.is_data_type())
+            .map(|(typ, ids)| (typ.as_str(), ids))
+            .collect();
```
leoyvens, line 168: Looks like you want https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.partition.

graph/src/data/query/result.rs
```diff
@@ -174,13 +181,16 @@ pub struct QueryResult {
     data: Option<Data>,
     #[serde(skip_serializing_if = "Vec::is_empty")]
     errors: Vec<QueryError>,
+    #[serde(skip_serializing)]
+    pub deployment: Option<SubgraphDeploymentId>,
```
leoyvens, line 185: Not very nice that this ended up here, I haven't looked closely but maybe it will be better located in `QueryResults`?

leoyvens (approved):
This looks good to me! Feel free to merge after addressing the comments.

store/postgres/src/dynds.rs
```diff
@@ -0,0 +1,135 @@
...
+        .unwrap_or(0);
+
+    Ok(Source {
+        address,
+        abi,
+        start_block,
+    })
+}
+
+#[allow(dead_code)]
```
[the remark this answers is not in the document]
lutter: Fixed

store/postgres/src/sharded_store.rs
```diff
@@ -0,0 +1,736 @@
...
+            primary,
+            stores,
+            sites,
+            placer,
+        }
+    }
+
+    // Only needed for tests
+    #[cfg(debug_assertions)]
+    #[allow(dead_code)]
```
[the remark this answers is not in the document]
lutter: Fixed

graphql/src/store/resolver.rs
```diff
@@ -95,40 +95,39 @@ impl StoreResolver {
...
-                    .and_then(|ptr| {
-                        let ptr =
-                            ptr.expect("we should have already checked that the subgraph exists");
-                        if ptr.number < number as u64 {
-                            Err(QueryExecutionError::ValueParseError(
-                                "block.number".to_owned(),
-                                format!(
-                                    "subgraph {} has only indexed up to block number {} \
+        // Relational storage (most subgraphs); block constraints fully
+        // supported
```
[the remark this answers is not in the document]
lutter: Fixed

store/postgres/src/entities.rs
lutter: `partition` only lets you partition by a boolean predicate - I need to partition by the type name (i.e., potentially more than two partitions)

leoyvens: I meant partitioning on `typ.is_data_type()`.

graph/src/data/query/result.rs
lutter: I tried that very briefly, but couldn't get it right; I totally agree that it's weird that that is there (it's only needed to report the query execution time for the right deployment in `server/http/service.rs`) I'll try and think of a better way for a followup PR.

store/postgres/src/entities.rs
leoyvens: It does seem tricky that you can't partition and map at the same time. It's fine to leave this as is.

leoyvens (approved):

store/postgres/src/entities.rs
lutter: Aah, yes, I totally misunderstood; next time we need to touch this code ;)

lutter:
Rebase onto latest master (after merging 2 commits from the original PR)

lutter:
Rename all migrations so that they come after any migrations currently on master

PR #2051, opened 2020-12-11
Add a special API to support the explorer

lutter:
This adds new endpoints under `/index-node/explorer` to support the explorer. These API endpoints are not meant for general consumption, and can and will change with no regard to backwards compatibility.

Responses for this API are cached for `GRAPH_EXPLORER_TTL` seconds, by default 10s, to protect `graph-node` from too much load from the explorer.


graph/src/data/subgraph/schema.rs
```diff
@@ -163,6 +163,14 @@ impl SubgraphHealth {
             SubgraphHealth::Failed => "failed",
         }
     }
+
+    pub fn is_failed(&self) -> bool {
+        match self {
+            SubgraphHealth::Healthy => false,
+            SubgraphHealth::Unhealthy => true,
```
leoyvens, line 170: While we don't have a proper UI design for the three health states, it seems preferable for the unhealthy state to be displayed as "Failed". So we can take the opportunity to change this behavior.

server/index-node/src/explorer.rs
```diff
@@ -0,0 +1,222 @@
...
+
+/// Caching of values for a specified amount of time
+#[derive(Debug)]
+struct CacheEntry<T> {
+    value: Arc<T>,
+    expires: Instant,
+}
+
+#[derive(Debug)]
+struct Cache<T> {
```
leoyvens, line 188: Worth a comment that this does not ever remove entries, but starts returning `None` after the TTL has elapsed.

server/index-node/src/explorer.rs
```diff
@@ -0,0 +1,222 @@
...
+    fn get(&self, key: &str) -> Option<Arc<T>> {
+        match self.entries.read().unwrap().get(key) {
+            Some(CacheEntry { value, expires }) => {
+                let now = Instant::now();
+                if *expires < now {
+                    Some(value.clone())
+                } else {
+                    None
+                }
+            }
```
leoyvens, line 210: I think this match could be written as:
```
Some(CacheEntry { value, expires }) if *expires < Instant::now() => {
  Some(value.clone())
}
_ => None
```

store/postgres/src/sharded_store.rs
```diff
@@ -649,6 +652,57 @@ impl StoreTrait for ShardedStore {
         let info = store.subgraph_info(id)?;
         Ok(info.network)
     }
+
+    fn version_info(&self, version: &str) -> Result<VersionInfo, StoreError> {
+        let primary = self.primary_conn()?;
```
leoyvens, line 657: Should all of the store operations in this fn be done atomically within a single transaction?

leoyvens (changes requested):

graph/src/data/subgraph/schema.rs
lutter: Changed

server/index-node/src/explorer.rs
lutter: Good point, added comments

server/index-node/src/explorer.rs
lutter: Yep, changed that

store/postgres/src/sharded_store.rs
lutter: There's not much we could gain there: we need to first consult the primary for details about the subgraph, then look up information about the deployment in the deployment's shard. The only other thing we get from that shard is `store.subgraph_info`, which is immutable info that is read once and cached for the lifetime of the store instance. To finish, we look up the chain head pointer - right now, that lives in the primary, but will soon move to (possibly) it's own shard.

So unfortunately, there's not much we can gain in terms of atomicity

lutter:

leoyvens (approved):

leoyvens (approved):

lutter:
This was merged at 839947af but somehow Github didn't realize that

PR #2054, opened 2020-12-11
docker: Fix README instructions for Linux

bingen:
I had two problems while trying to run docker image on Linux:

- The docker image didn’t have `iproute2` installed, so `ip` command was not working.
- The escaped quotes were not working for me.


leoyvens:
@bingen looks like this needs to be rebased.

bingen:
Rebased onto master.

PR #2056, opened 2020-12-11
Add `subgraphError` argument only when the feature is present

leoyvens:
We saw an issue in a graphql client that is avoided by doing this. And it is generally more correct to only include it if the subgraph opts in. This requires storing the `features` field in the metadata, which we should be doing anyways.


lutter (approved):
Nice!

PR #2057, opened 2020-12-11
store: Don't trip over old errors with invalid block hash/range combinations

lutter:

### DIFF
--- a/store/postgres/src/detail.rs
+++ b/store/postgres/src/detail.rs
@@ -6,7 +6,9 @@ use diesel::prelude::{
 };
 use graph::{
     data::subgraph::schema::SubgraphError,
-    prelude::{bigdecimal::ToPrimitive, BigDecimal, StoreError, SubgraphDeploymentId},
+    prelude::{
+        bigdecimal::ToPrimitive, BigDecimal, EthereumBlockPointer, StoreError, SubgraphDeploymentId,
+    },
 };
 use graph::{
     data::subgraph::{schema::SubgraphHealth, status},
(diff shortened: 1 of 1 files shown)


leoyvens (changes requested):
We should also be able to deal with the block hash being null, as far as I know it's very much possible for a fatal error to not be associated with a block.

lutter:
Take 2

store/postgres/src/detail.rs
```diff
@@ -140,8 +140,14 @@ impl TryFrom<ErrorDetail> for SubgraphError {
             block_range,
         } = value;
         let block_number = crate::block_range::first_block_in_range(&block_range).map(|n| n.into());
-        let block_ptr = block(&subgraph_id, "fatal_error", block_hash, block_number)?
-            .map(|block| block.to_ptr());
+        // In existing databases, we have errors that have a `block_range` of
+        // `UNVERSIONED_RANGE`, which leads to `None` as the block number, but
+        // has a hash. In that case, just return a block pointer of `None`
+        let block_ptr = match (&block_number, &block_hash) {
+            (Some(_), Some(_)) => block(&subgraph_id, "fatal_error", block_hash, block_number)?
```
leoyvens, line 147: Going through `block` function is complicating this code, we should not convert to `BigDecimal` and construct the `EthereumBlock` directly in this branch.

leoyvens (changes requested):

fordN (approved):
LGTM 👍

lutter:
Take 3

leoyvens (approved):

PR #2059, opened 2020-12-14
Added flag to reject any operation using non-deterministic fulltext search

That3Percent:


leoyvens (approved):
Good to merge after resolving conflicts.

PR #2068, opened 2020-12-16
Remove templates nested in data sources

leoyvens:
We've already dropped support for it, this is doing a cleanup removing related code. specVersion 0.0.1 is no longer accepted. There are only 11 subgraphs on the hosted service that are still on that, and they do not get any queries.


lutter (approved):
Nice! PR's that delete code are the best PR's

PR #2070, opened 2020-12-16
No failure

That3Percent:
`failure` is not an `Option` (migrate from failure to anyhow + thiserror)
Also fixed an error message about store configuration that was wrong.

Failure came up in a security review because it was deprecated. This also makes progress toward the deterministic error changes required for the gas PR.


leoyvens (approved):
Yay! No more fixing all those compiler errors from bridging between anyhow and failure.

PR #2074, opened 2020-12-17
server: Instrument getting entity_count

lutter:


server/index-node/src/explorer.rs
```diff
@@ -121,14 +153,32 @@ where
...
 
-    fn handle_entity_count(&self, deployment: &str) -> Result<Response<Body>, GraphQLServerError> {
+    fn handle_entity_count(
+        &self,
+        logger: &Logger,
+        deployment: &str,
+    ) -> Result<Response<Body>, GraphQLServerError> {
+        let start = Instant::now();
         if let Some(value) = self.entity_counts.get(deployment) {
+            if start.elapsed() > *LOCK_THRESHOLD {
```
That3Percent, line 163: Lets move the check outside of the if statement so that we check the elapsed duration even on misses.

That3Percent (changes requested):

That3Percent (approved):

PR #2076, opened 2020-12-18
Instrument RwLock cache

leoyvens:
On Linux, readers will by [default starve writers](https://man7.org/linux/man-pages/man3/pthread_rwlockattr_setkind_np.3.html).
> As long as there are readers, the writer will be starved.

The query cache is a very busy lock and it runs in a non-blocking task, so we want to log any contention on it.

### DIFF
--- a/graph/src/util/timed_rw_lock.rs
+++ b/graph/src/util/timed_rw_lock.rs
@@ -1,0 +1,46 @@
+use slog::{warn, Logger};
+use std::{
+    sync::RwLock,
+    time::{Duration, Instant},
+};
+
+/// Adds instrumentation for timing the performance of the lock.
+pub struct TimedRwLock<T> {
+    id: String,
+    lock: RwLock<T>,
+}
+
+impl<T> TimedRwLock<T> {
+    pub fn new(id: impl Into<String>, x: T) -> Self {
+        TimedRwLock {
+            id: id.into(),
+            lock: RwLock::new(x),
+        }
+    }
+
+    pub fn write(&self, logger: &Logger) -> std::sync::RwLockWriteGuard<T> {
+        let start = Instant::now();
+        let guard = self.lock.write().unwrap();
+        let elapsed = start.elapsed();
+        if elapsed > Duration::from_millis(100) {
+            warn!(logger, "Write lock took a long time to acquire";
+                          "id" => &self.id,
+                          "wait_ms" => elapsed.as_millis(),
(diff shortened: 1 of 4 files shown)


graph/src/util/timed_rw_lock.rs
```diff
@@ -0,0 +1,46 @@
...
+            id: id.into(),
+            lock: RwLock::new(x),
+        }
+    }
+
+    pub fn write(&self, logger: &Logger) -> std::sync::RwLockWriteGuard<T> {
+        let start = Instant::now();
+        let guard = self.lock.write().unwrap();
+        let elapsed = start.elapsed();
+        if elapsed > Duration::from_millis(100) {
```
lutter, line 25: It would be better to make the threshold an environment variable - that way, if it turns out that this is spamming the logs, we can raise the threshold and disable logging with a simple config change.

graphql/src/execution/execution.rs
```diff
@@ -201,7 +202,8 @@ lazy_static! {
 
     // Cache query results for recent blocks by network.
     // The `VecDeque` works as a ring buffer with a capacity of `QUERY_CACHE_BLOCKS`.
-    static ref QUERY_BLOCK_CACHE: RwLock<QueryBlockCache> = RwLock::new(QueryBlockCache(vec![]));
+    static ref QUERY_BLOCK_CACHE: TimedRwLock<QueryBlockCache> =
+                                    TimedRwLock::new("query_block_cache", QueryBlockCache(vec![]));
     static ref QUERY_HERD_CACHE: QueryCache<Arc<QueryResult>> = QueryCache::new();
     static ref QUERY_LFU_CACHE: Mutex<LfuCache<QueryHash, WeightedResult>> = Mutex::new(LfuCache::new());
```
lutter, line 208: Should we do something similar for this `Mutex`? We see quite a few historical queries, and we might actually be contending on this lock.

And similar, can we be sure that the herd cache isn't heavily contended, too?

With logging for additional caches, make sure it's easy to search for the log messages for lock contention and to connect them to the cache we are contending on.

lutter (changes requested):

graphql/src/execution/execution.rs
leoyvens: Instrumented the mutexes, I tested the log messages and they are searchable and informative.

lutter (approved):

PR #2077, opened 2020-12-18
WIP: Use lifetimes to protect store from deadlock

That3Percent:


leoyvens:
Deadlocks are not considered `unsafe` in Rust. We could make our own conventions on what unsafe means but imo we should just stick with the general convention that `unsafe` is for memory safety. We can still call this "deadlock safety", just not use the `unsafe` keyword.

That3Percent:
I think @lutter got what he needed from this. Closing.

PR #2080, opened 2020-12-19
Swap RwLock for Mutex on query cache

leoyvens:
We don't have that many more writers than readers, and the read section is shorter than the write section, so I don't believe the overhead of an RwLock is justified. Empirically we see a ton of contention on the RwLock cache but none on the Mutex cache.


lutter (approved):

lutter:
Need to try it in production

PR #2082, opened 2020-12-19
Shard query block cache

leoyvens:
This should cut the queue on a lock proportionally to `QUERY_BLOCK_CACHE_SHARDS` which defaults to 10.

### DIFF
--- a/graphql/src/execution/execution.rs
+++ b/graphql/src/execution/execution.rs
@@ -202,19 +202,24 @@ lazy_static! {
     };
 
     /// In how many shards (mutexes) the query block cache is split.
+    /// Ideally this should divide 256 so that the distribution of queries to shards is even.
     static ref QUERY_BLOCK_CACHE_SHARDS: u8 = {
         std::env::var("GRAPH_QUERY_BLOCK_CACHE_SHARDS")
-        .unwrap_or("100".to_string())
+        .unwrap_or("128".to_string())
         .parse::<u8>()
         .expect("Invalid value for GRAPH_QUERY_BLOCK_CACHE_SHARDS environment variable, max is 255")
     };
 
 
-    // Cache query results for recent blocks by network.
+    // Sharded query results cache for recent blocks by network.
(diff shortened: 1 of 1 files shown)


lutter (approved):

graphql/src/execution/execution.rs
```diff
@@ -199,11 +201,20 @@ lazy_static! {
...
+        .expect("Invalid value for GRAPH_QUERY_BLOCK_CACHE_SHARDS environment variable, max is 255")
+    };
+
 
     // Cache query results for recent blocks by network.
     // The `VecDeque` works as a ring buffer with a capacity of `QUERY_CACHE_BLOCKS`.
-    static ref QUERY_BLOCK_CACHE: TimedMutex<QueryBlockCache> =
-                                    TimedMutex::new(QueryBlockCache(vec![]), "query_block_cache");
+    static ref QUERY_BLOCK_CACHE: Vec<TimedMutex<QueryBlockCache>> =
+                                    std::iter::repeat_with(|| TimedMutex::new(QueryBlockCache(vec![]), "query_block_cache"))
```
lutter, line 216: What do you think of including the number of the shard in the identifier for the mutex, something like `query_block_cache_42`?

graphql/src/execution/execution.rs
```diff
@@ -199,11 +201,20 @@ lazy_static! {
         .expect("Invalid value for GRAPH_QUERY_CACHE_STALE_PERIOD environment variable")
     };
 
+    /// In how many shards (mutexes) the query block cache is split.
+    static ref QUERY_BLOCK_CACHE_SHARDS: u8 = {
+        std::env::var("GRAPH_QUERY_BLOCK_CACHE_SHARDS")
+        .unwrap_or("100".to_string())
```
lutter, line 207: To get even distribution across the shards, maybe the default value should divide 256, i.e., either 64 or 128

lutter (approved):

PR #2084, opened 2020-12-19
Allow type named 'List'

leoyvens:
An amusing oddity that we are considering the string `"List"` as representing a list type, but that's never the case. This was discovered because a user tried using a `type List @entity { ... }` and this caused an error.


lutter (approved):
Nice find!

PR #2085, opened 2020-12-21
Update link to geth docs

pulkitsethi:
geth docs have moved from Github wiki to its own site. If you click on current link, it will show you a wiki page with a link to the new docs.


leoyvens:
Thank you

PR #2091, opened 2020-12-22
Query cache stats

leoyvens:
This logs details about the generational query cache operating conditions. The second commit moves it into the `cache.rs` module and refactors it a bit.

### DIFF
--- a/graphql/src/execution/cache.rs
+++ b/graphql/src/execution/cache.rs
@@ -6,12 +6,15 @@ use graph::{
 };
 use stable_hash::crypto::SetHasher;
 use stable_hash::prelude::*;
-use std::collections::HashMap;
-use std::collections::{hash_map::Entry, VecDeque};
 use std::future::Future;
 use std::pin::Pin;
 use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::Arc;
+use std::{collections::HashMap, time::Duration};
+use std::{
+    collections::{hash_map::Entry, VecDeque},
+    time::Instant,
+};
 
 use super::QueryHash;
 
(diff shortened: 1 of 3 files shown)


graphql/src/execution/cache.rs
```diff
@@ -75,3 +78,162 @@ impl<R: CheapClone> QueryCache<R> {
...
+            graph::spawn(async move {
+                let mut dead_inserts = 0;
+                let mut total_hits = 0;
+                for (_, hits) in block.cache.values() {
+                    let hits = hits.load(Ordering::SeqCst);
+                    total_hits += hits;
+                    if hits == 0 {
+                        dead_inserts += 1;
+                    }
+                }
```
lutter, line 199: If we want to switch to caching on the second cache access, it would be good to know how many entries we have that are accessed only once, since those would go from hits to misses.

graphql/src/execution/cache.rs
```diff
@@ -75,3 +78,162 @@ impl<R: CheapClone> QueryCache<R> {
...
+    }
+
+    /// Returns `true` if the insert was successful or `false` if the cache was full.
+    fn insert(&mut self, key: QueryHash, value: Arc<QueryResult>, weight: usize) -> bool {
+        // We never try to insert errors into this cache, and always resolve some value.
+        assert!(!value.has_errors());
+        let fits_in_cache = self.weight + weight <= self.max_weight;
+        if fits_in_cache {
+            self.weight += weight;
+            self.cache.insert(key, (value, AtomicU64::new(0)));
```
lutter, line 115: Should we also track how long inserts take for each `CacheByBlock`? Since it might be related to allocations, it would be good to instrument the `CacheByBlock` to keep a `total_insertion_time`

lutter (approved):
LGTM!

graphql/src/execution/cache.rs
leoyvens: Those would have a new status, `PreInsert` or something, we'll figure it out when we do that.

lutter (approved):

PR #2095, opened 2020-12-23
Make subscription queries cacheable

lutter:
The overall effect will be that subscriptions will be deduplicated


lutter:
Leo looked at this and gave his thumbs up.

PR #2100, opened 2021-01-07
*: Require a block number in `proofOfIndexing` queries

Jannis:
This is to avoid `null` POIs if the block hash isn't found in the block store (which could be the case on one indexer but not on another). This can lead to indexers losing rewards despite indexing a subgraph correctly.

By requiring e.g. the indexer agent to always provide block number and hash in `proofOfIndexing` queries, we can compute the POI on any indexer, regardless of what is in their block store. This is easier than resolving the block hash -> number on-demand in the `proofOfIndexing` resolver. The agent already knows both values anyway.

Part of fixing https://github.com/graphprotocol/indexer/issues/186.

_Note: This is a breaking change but I didn't want to make the block number argument optional, as that just adds unnecessary complexity._

### DIFF
--- a/graph/src/components/store.rs
+++ b/graph/src/components/store.rs
@@ -91,9 +91,7 @@ impl EntityType {
     /// This method will panic if `self` is a `Data` type
     pub fn expect_metadata(&self) -> &str {
         match self {
-            Self::Data(_) => {
-                unreachable!("callers check that this is never called for data")
-            }
+            Self::Data(_) => unreachable!("callers check that this is never called for data"),
             Self::Metadata(typ) => typ.as_str(),
         }
     }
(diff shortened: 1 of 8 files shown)


graph/src/components/store.rs
```diff
@@ -1179,7 +1177,7 @@ impl Store for MockStore {
         self: Arc<Self>,
         _subgraph_id: &'a SubgraphDeploymentId,
         _indexer: &'a Option<Address>,
-        _block_hash: H256,
+        _block: &'a EthereumBlockPointer,
```
leoyvens, line 1180: `EthereumBlockPointer` is `Copy` so there is no need to use a reference.

That3Percent (approved):

graph/src/components/store.rs
Jannis: Cool, updated to avoid the reference and things still compile and work. Thanks!

PR #2104, opened 2021-01-09
Startup failure should not fail the subgraph

leoyvens:
It's problematic to register subgraph start errors as a subgraph failure, since we could be processing a request to start a subgraph that is already running.

Generally, I think the ideal situation would be that only deterministic failures are recorded as subgraph errors, and transient failures can either be retried or just logged for the operation to be retried when the index node is reset.


Jannis (approved):
LGTM

PR #2112, opened 2021-01-14
*: Deterministic error work in preparation for gas

That3Percent:
**Error Determinism**
  -  Introduced determinism in error types
   - Before this would set deterministic_host_trap, but access to this value may not be available in many cases where the error originates or may use shared references forcing atomics
   - anyhow errors are necessarily considered non-deterministic until proven otherwise.
   - Anyhow is dangerous for our use-case because it hides information by design. (This extends to any use-case which handles errors) There's not much that can be done here in the short term.
   - Lots of anyhow errors have been converted to deterministic ones (eg: json parsing, type conversions, math, and most other host exports are deterministic)
   - Note that like anyhow, panics do not carry determinism information. These must be treated as non-deterministic and should be avoided for error conditions arising out of the subgraph or the data it acts upon.

**Non-Panicky Errors**

- BigDecimal % 0 no longer panics but has DeterministicHostError instead
- Before the wasm was trusted to interact with the heap in-bounds, leading to panics - which with this change is a DeterministicHostError instead
- Same story for assertions on binary representations for AscValue which also are untrusted inputs
There's probably a lot more that can be done along these lines...

### DIFF
--- a/runtime/derive/src/lib.rs
+++ b/runtime/derive/src/lib.rs
@@ -67,29 +67,34 @@ fn asc_type_derive_struct(item_struct: ItemStruct) -> TokenStream {
 
     TokenStream::from(quote! {
         impl#impl_generics AscType for #struct_name#ty_generics #where_clause {
-            fn to_asc_bytes(&self) -> Vec<u8> {
+            fn to_asc_bytes(&self) -> Result<Vec<u8>, DeterministicHostError> {
                let mut bytes = Vec::new();
-                #(bytes.extend_from_slice(&self.#field_names.to_asc_bytes());)*
+                #(bytes.extend_from_slice(&self.#field_names.to_asc_bytes()?);)*
 
                 // Assert that the struct has no padding.
                 assert_eq!(bytes.len(), size_of::<Self>());
-                bytes
+                Ok(bytes)
             }
 
             #[allow(unused_variables)]
-            fn from_asc_bytes(asc_obj: &[u8]) -> Self {
-                assert_eq!(asc_obj.len(), size_of::<Self>());
+            fn from_asc_bytes(asc_obj: &[u8]) -> Result<Self, DeterministicHostError> {
+                if asc_obj.len() != size_of::<Self>() {
+                    return Err(DeterministicHostError(anyhow::anyhow!("Size does not match")));
+                }
                 let mut offset = 0;
 
                 #(
                 let field_size = std::mem::size_of::<#field_types>();
-                let #field_names2 = AscType::from_asc_bytes(&asc_obj[offset..(offset + field_size)]);
+                let field_data = asc_obj.get(offset..(offset + field_size)).ok_or_else(|| {
+                    DeterministicHostError(anyhow!("Attempted to read past end of array"))
+                })?;
+                let #field_names2 = AscType::from_asc_bytes(&field_data)?;
                 offset += field_size;
                 )*
 
-                Self {
+                Ok(Self {
                     #(#field_names3,)*
-                }
+                })
             }
         }
     })
@@ -109,7 +114,7 @@ fn asc_type_derive_struct(item_struct: ItemStruct) -> TokenStream {
 //
 // Example output:
 // impl AscType for JsonValueKind {
-//     fn to_asc_bytes(&self) -> Vec<u8> {
+//     fn to_asc_bytes(&self) -> Result<Vec<u8>, DeterministicHostError> {
 //         let discriminant: u32 = match *self {
 //             JsonValueKind::Null => 0u32,
 //             JsonValueKind::Bool => 1u32,
@@ -118,11 +123,14 @@ fn asc_type_derive_struct(item_struct: ItemStruct) -> TokenStream {
 //             JsonValueKind::Array => 4u32,
 //             JsonValueKind::Object => 5u32,
 //         };
-//         discriminant.to_asc_bytes()
+//         Ok(discriminant.to_asc_bytes())
 //     }
 //
-//     fn from_asc_bytes(asc_obj: &[u8]) -> Self {
+//     fn from_asc_bytes(asc_obj: &[u8]) -> Result<Self, DeterministicHostError> {
 //         let mut u32_bytes: [u8; size_of::<u32>()] = [0; size_of::<u32>()];
+//         if std::mem::size_of_val(&u32_bytes) != std::mem::size_of_val(&as_obj) {
+//             return Err(DeterministicHostError(anyhow::anyhow!("Invalid asc bytes size")));
+//         }
 //         u32_bytes.copy_from_slice(&asc_obj);
 //         let discr = u32::from_le_bytes(u32_bytes);
 //         match discr {
@@ -132,7 +140,7 @@ fn asc_type_derive_struct(item_struct: ItemStruct) -> TokenStream {
 //             3u32 => JsonValueKind::String,
 //             4u32 => JsonValueKind::Array,
 //             5u32 => JsonValueKind::Object,
-//             _ => panic!("value {} is out of range for {}", discr, "JsonValueKind"),
+//             _ => Err(DeterministicHostError(anyhow::anyhow!("value {} is out of range for {}", discr, "JsonValueKind")),
 //         }
 //     }
 // }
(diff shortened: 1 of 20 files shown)


runtime/wasm/src/asc_abi/class.rs
```diff
@@ -176,15 +187,24 @@ impl AscType for AscString {
...
     }
 
     /// The Rust representation of an Asc object as layed out in Asc memory.
-    fn from_asc_bytes(asc_obj: &[u8]) -> Self {
+    fn from_asc_bytes(asc_obj: &[u8]) -> Result<Self, DeterministicHostError> {
         // Pointer for our current position within `asc_obj`,
         // initially at the start of the content skipping `length`.
         let mut offset = size_of::<u32>();
 
+        // TODO: Verify that the length specified matches the byte length
```
leoyvens, line 199: What's the situation with this TODO?

runtime/wasm/src/asc_abi/class.rs
```diff
@@ -197,11 +217,15 @@ impl AscType for AscString {
...
 
-    fn asc_size<H: AscHeap>(ptr: AscPtr<Self>, heap: &H) -> u32 {
-        let length = ptr.read_u32(heap);
+    fn asc_size<H: AscHeap>(ptr: AscPtr<Self>, heap: &H) -> Result<u32, DeterministicHostError> {
+        let length = ptr.read_u32(heap)?;
         let length_size = size_of::<u32>() as u32;
         let code_point_size = size_of::<u16>() as u32;
-        length_size + code_point_size * length
+        let data_size = code_point_size.checked_mul(length);
+        let total_size = data_size.and_then(|d| d.checked_add(length_size));
```
leoyvens, line 225: Checked arithmetic, nice.

runtime/wasm/src/asc_abi/mod.rs
```diff
@@ -88,14 +93,14 @@ pub trait TryFromAscObj<C: AscType>: Sized {
...
+    fn to_asc_bytes(&self) -> Result<Vec<u8>, DeterministicHostError>;
 
     /// The Rust representation of an Asc object as layed out in Asc memory.
-    fn from_asc_bytes(asc_obj: &[u8]) -> Self;
+    fn from_asc_bytes(asc_obj: &[u8]) -> Result<Self, DeterministicHostError>;
 
     /// Size of the corresponding Asc instance in bytes.
-    fn asc_size<H: AscHeap>(_ptr: AscPtr<Self>, _heap: &H) -> u32 {
-        size_of::<Self>() as u32
+    fn asc_size<H: AscHeap>(_ptr: AscPtr<Self>, _heap: &H) -> Result<u32, DeterministicHostError> {
```
leoyvens, line 102: Thanks for doing this lift to make AS memory operations not panic, this was the most daunting requirement for deterministic errors.

runtime/wasm/src/error.rs
```diff
@@ -0,0 +1,50 @@
...
+
+impl error::Error for DeterministicHostError {}
+
+impl From<DeterministicHostError> for HostExportError {
+    fn from(value: DeterministicHostError) -> Self {
+        HostExportError::Deterministic(value.0)
+    }
+}
+
+impl From<DeterministicHostError> for MappingError {
```
leoyvens, line 37: This impl seems to be used only for tests, lets mark it as `#[cfg(debug_assertions)]`.

runtime/wasm/src/module/mod.rs
```diff
@@ -591,17 +618,38 @@ impl AscHeap for WasmInstanceContext {
...
         // Safety:
         // This unsafe block has been checked to not cause unsoundness by itself.
         // See 2155cdca-dfaa-4fba-86e4-289e7683c1bf for why this is sufficient.
-        unsafe { self.memory.data_unchecked()[offset..(offset + size)].to_vec() }
+        let data = unsafe {
+            self.memory
+                .data_unchecked()
+                .get(offset..end)
+                .map(|s| s.to_vec())
+        };
```
leoyvens, line 644: We should be able to get rid of these `unsafe`s soon https://github.com/bytecodealliance/wasmtime/pull/2528.

runtime/wasm/src/module/mod.rs
```diff
@@ -878,30 +935,29 @@ impl WasmInstanceContext {
...
     /// function ipfs.cat(link: String): Bytes
-    fn ipfs_cat(&mut self, link_ptr: AscPtr<AscString>) -> Result<AscPtr<Uint8Array>, Trap> {
-        if !self.allow_non_determinstic_ipfs {
-            return Err(anyhow::anyhow!(
+    fn ipfs_cat(
+        &mut self,
+        link_ptr: AscPtr<AscString>,
+    ) -> Result<AscPtr<Uint8Array>, HostExportError> {
+        if !self.experimental_features.allow_non_deterministic_ipfs {
+            return Err(HostExportError::Deterministic(anyhow!(
```
leoyvens, line 947: I wouldn't consider this deterministic because it depends on an env var set by the indexer.

runtime/wasm/src/module/mod.rs
```diff
@@ -915,19 +971,18 @@ impl WasmInstanceContext {
         callback: AscPtr<AscString>,
         user_data: AscPtr<AscEnum<StoreValueKind>>,
         flags: AscPtr<Array<AscPtr<AscString>>>,
-    ) -> Result<(), Trap> {
-        if !self.allow_non_determinstic_ipfs {
-            return Err(anyhow::anyhow!(
+    ) -> Result<(), HostExportError> {
+        if !self.experimental_features.allow_non_deterministic_ipfs {
+            return Err(HostExportError::Deterministic(anyhow!(
```
leoyvens, line 976: Same.

runtime/wasm/src/module/mod.rs
```diff
@@ -1252,58 +1284,76 @@ impl WasmInstanceContext {
...
     /// function box.profile(address: string): JSONValue | null
-    fn box_profile(&mut self, address: AscPtr<AscString>) -> Result<AscPtr<AscJson>, Trap> {
-        let address: String = self.asc_get(address);
+    fn box_profile(
+        &mut self,
+        address: AscPtr<AscString>,
+    ) -> Result<AscPtr<AscJson>, HostExportError> {
+        // TODO: 3box data is mutable and the best solution here is probably to remove support
+        if !self.experimental_features.allow_non_deterministic_3box {
+            return Err(HostExportError::Deterministic(anyhow!(
```
leoyvens, line 1349: Same.

runtime/wasm/src/module/mod.rs
```diff
@@ -1252,58 +1284,76 @@ impl WasmInstanceContext {
...
 
     /// function arweave.transactionData(txId: string): Bytes | null
     fn arweave_transaction_data(
         &mut self,
         tx_id: AscPtr<AscString>,
-    ) -> Result<AscPtr<Uint8Array>, Trap> {
-        let tx_id: String = self.asc_get(tx_id);
+    ) -> Result<AscPtr<Uint8Array>, HostExportError> {
+        if !self.experimental_features.allow_non_deterministic_arweave {
+            return Err(HostExportError::Deterministic(anyhow!(
```
leoyvens, line 1332: Same.

leoyvens (commented):
Very happy to see this lift done!

runtime/derive/src/lib.rs
```diff
@@ -155,20 +163,20 @@ fn asc_type_derive_enum(item_enum: ItemEnum) -> TokenStream {
 
     TokenStream::from(quote! {
         impl#impl_generics AscType for #enum_name#ty_generics #where_clause {
-            fn to_asc_bytes(&self) -> Vec<u8> {
+            fn to_asc_bytes(&self) -> Result<Vec<u8>, DeterministicHostError> {
                 let discriminant: u32 = match *self {
                     #(#enum_name_iter::#variant_paths => #variant_discriminant,)*
                 };
-                discriminant.to_asc_bytes()
+                Ok(discriminant.to_asc_bytes()?)
```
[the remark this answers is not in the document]
That3Percent: Fixing

runtime/wasm/src/asc_abi/class.rs
That3Percent: The code as it was before (and as it is now) does not verify that the value of the first 4 bytes specifying the string length coincides with the length of `asc_obj`.

leoyvens: Ah got it

runtime/wasm/src/module/mod.rs
That3Percent: I spent a while waffling about how to handle this. We've used the word "determinism" to actually mean multiple conflated things in different contexts. In this case, the error is in fact _attestable_. Within the context of The Network this subgraph should always fail with this error, deterministically. I'd be interested in a longer chat on slack about whether or not we should tease these apart somehow or what.

runtime/wasm/src/error.rs
That3Percent: Fixing.

runtime/wasm/src/module/mod.rs
leoyvens: I see your point that in the network the this should never be allowed so it is deterministic. Ok we can keep this as deterministic.

leoyvens (approved):

enzojavier-desimone:
I think that the issue that I am facing is related to this issue, 

I am running in local the [ens-subgraph ](https://github.com/ensdomains/ens-subgraph) when reaching to the [first block to index](https://etherscan.io/block/3330220) I have the error

```
Mar 08 18:28:07.005 ERRO Subgraph error 1/1, code: SubgraphSyncingFailure, error: `ens_name_by_hash` is deprecated	wasm backtrace:	    0: 0x2e77 - <unknown>!src/ensRegistry/handleNewOwner	 in handler `handleNewOwnerOldRegistry` at block #3330220 (4ea8feee666e509db9d574b097efaa47b23de6b06fb33222dbed1484b1ca4748), block_hash: 0x4ea8feee666e509db9d574b097efaa47b23de6b06fb33222dbed1484b1ca4748, block_number: 3330220, sgd: 1, subgraph_id: QmYbUu3ZfmrK8ugavLFEnz2hHHiEUx22ya3SwqwBe8HuKQ, component: SubgraphInstanceManager
Mar 08 18:28:07.041 ERRO Subgraph instance failed to run: `ens_name_by_hash` is deprecated	wasm backtrace:	    0: 0x2e77 - <unknown>!src/ensRegistry/handleNewOwner	 in handler `handleNewOwnerOldRegistry` at block #3330220 (4ea8feee666e509db9d574b097efaa47b23de6b06fb33222dbed1484b1ca4748), code: SubgraphSyncingFailure, sgd: 1, subgraph_id: QmYbUu3ZfmrK8ugavLFEnz2hHHiEUx22ya3SwqwBe8HuKQ, component: SubgraphInstanceManager
```

doing some research on it I see that it comes for this implementation here https://github.com/graphprotocol/graph-node/blob/master/runtime/wasm/src/module/mod.rs#L1576 

then my question is, how it is possible to set the `experimental_features`? I would like to change the value of `allow_non_deterministic_ipfs` but I could not find a command-line param or a env variable for it

enzojavier-desimone:
UPDATE, I found the undocumented env variable for set the experimental feature `GRAPH_ALLOW_NON_DETERMINISTIC_IPFS`

PR #2114, opened 2021-01-15
Fix unused variable warning

leoyvens:
Depending if this is built with the `failpoints` feature turned on or not, it can trigger a warning.


chain/ethereum/src/block_stream.rs
```diff
@@ -833,6 +833,7 @@ where
 
 // This always returns `false` in a normal build. A test may configure reorg by enabling
 // "test_reorg" fail point with the number of the block that should be reorged.
+#[warn(unused_variables)]
```
lutter, line 836: Shouldn't that be `allow`?

lutter (approved):

chain/ethereum/src/block_stream.rs
leoyvens: Of course, this does nothing. Amended the commit.

Issue #2106, opened 2021-01-12
_meta fields in schema do not resolve __typename


jeffywu:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
ApolloClient automatically attaches the __typename field to queries for caching purposes, these do not resolve when querying the _meta fields exposed in v0.20.0.

Unfortunately ApolloClient does not have a good way of disabling __typename for select queries, so to get _meta queries to resolve you have to monkey patch it using an ApolloLink.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem.**
Run the following query:
```
{
  _meta {
    block {
      number
      __typename
   }
  __typename
  hasIndexingErrors
}
```

You should receive errors such as:
```
{
  "errors": [
    {
      "locations": [
        {
          "line": 5,
          "column": 9
        }
      ],
      "message": "Null value resolved for non-null field `__typename`"
    },
    {
      "locations": [
        {
          "line": 8,
          "column": 7
        }
      ],
      "message": "Null value resolved for non-null field `__typename`"
    }
  ]
}
```

**What is the expected behavior?**
It would be nice if these _meta fields would expose __typename so that developers don't have to patch anything if they're using ApolloClient.

fubhy:
I just ran into this too. I came to the same conclusion to add a ApolloLink middleware for now to modify the query document AST but this is definitely a bug in the graph-node query resolver.

fubhy:
In case anyone else runs into this issue, this sufficiently resolves this for us for now (it's still a hack obviously):

```typescript
new ApolloLink((operation, forward) => {
        operation.query = visit(operation.query, {
          SelectionSet: {
            enter: (node, _, parent) => {
              if ((parent as FieldNode).kind === 'Field' && (parent as FieldNode).name.value === '_meta') {
                return visit(node, {
                  Field: {
                    enter: (node) => (node.name.value === '__typename' ? null : undefined),
                  },
                });
              }
            },
          },
        });

        return forward(operation).map((result) => {
          if (result.data?._meta) {
            result.data._meta.__typename = '_Meta_';
          }

          if (result.data?._meta?.block) {
            result.data._meta.block.__typename = '_Block_';
          }

          return result;
        });
      }),
```

PR #2118, opened 2021-01-18
graphql: Resolve __typename in _meta field

lutter:
Fixes https://github.com/graphprotocol/graph-node/issues/2106


leoyvens (approved):

leoyvens:
Thanks! Well remembered to also add it to the inner `block` field.

PR #2121, opened 2021-01-20
Fix has_non_fatal_errors

### LINKED ISSUE #2093: indexing_error is returned for subgraphs that don't have the feature enabled
If a subgraph fails with a fatal but deteministic error, we return the `indexing_error` graphql error. Since fatal errors should not alter deterministic responses, this is a bug.

leoyvens:
This is a bit subtle: When the instance manager calls `fail_subgraph` at block N, that block has not been transacted into the store, and is therefore N is beyond the subgraph head block.  But right now because the error is deterministic we'd consider it non-fatal. To fix this, `non_fatal_error` will not check if any determinstic error exists, but if one exists up to the subgraph head block and therefore is an error that has been transacted. The test demonstrates this situation.

Resolves #2093.

### DIFF
--- a/store/postgres/src/deployment.rs
+++ b/store/postgres/src/deployment.rs
@@ -7,6 +7,7 @@ use diesel::{
     dsl::{delete, insert_into, select, sql, update},
     sql_types::Integer,
 };
+use graph::data::subgraph::schema::SubgraphError;
 use graph::data::subgraph::{
     schema::{MetadataType, SubgraphManifestEntity},
     SubgraphFeature,
@@ -16,7 +17,6 @@ use graph::prelude::{
     DeploymentState, EntityChange, EntityChangeOperation, EthereumBlockPointer, Schema, StoreError,
     StoreEvent, SubgraphDeploymentId,
 };
-use graph::{data::subgraph::schema::SubgraphError, prelude::BLOCK_NUMBER_MAX};
 use stable_hash::crypto::SetHasher;
 use std::str::FromStr;
 use std::{collections::BTreeSet, convert::TryFrom, ops::Bound};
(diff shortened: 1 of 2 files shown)


Jannis:
> This is a bit subtle: When the instance manager calls `fail_subgraph` at block N, that block has not been transacted into the store, and is therefore N is beyond the subgraph head block. But right now because the error is deterministic we'd consider it non-fatal. To fix this, `non_fatal_error` will not check if any deterministic error exists, but if one exists up to the subgraph head block and therefore is an error that has been transacted. The test demonstrates this situation.

The subgraphs I noticed had this problem had synced beyond the error block already, and there were errors in `subgraphs.subgraph_error` for them. I may not be following the above explanation 100% and am wondering if this really fixes the problem.

Doesn't the fix assume that the query is made between the error being written to the db and the block (from which the error came) being transacted? I could reliably reproduce this query error regardless of what block the subgraphs were on.

leoyvens:
> The subgraphs I noticed had this problem had synced beyond the error block already, and there were errors in subgraphs.subgraph_error for them.

A subgraph that does not have the feature enabled should not sync past an error block, it would be a separate bug that we'd need to reproduce.

Jannis:
> > The subgraphs I noticed had this problem had synced beyond the error block already, and there were errors in subgraphs.subgraph_error for them.
> 
> A subgraph that does not have the feature enabled should not sync past an error block, it would be a separate bug that we'd need to reproduce.

That's true, and they didn't initially. Here's what happened:

1. The subgraphs failed with that "Attempted to read past end of array" error.
2. I rolled back graph-node to an older commit.
3. The subgraphs got past the previously failing block fine after that.
4. But the "Attempted to read past end of array" error would remain in the database and cause the `indexing_errors` results.

So at point 4, the subgraphs made it further than the previously failing block but the previously fatal error was still lingering around and be included in the `indexing_error` check apparently.

I wonder if it could be necessary to not just reset the health of a subgraph on startup, but also clear out any errors from the previously failing block, since that block will be re-attempted?

leoyvens:
@Jannis I see, yes that case I hadn't considered. Clearing out the error should do it, I'll add that to this PR.

Jannis:
@leoyvens Nice, thank you 😄

store/postgres/src/deployment.rs
```diff
@@ -502,14 +502,26 @@ pub(crate) fn has_non_fatal_errors(
...
+    };
+
     select(diesel::dsl::exists(
         e::table
             .filter(e::subgraph_id.eq(id.as_str()))
             .filter(e::deterministic)
-            .filter(sql("block_range @> ").bind::<Integer, _>(block)),
+            .filter(
+                sql("block_range @> ")
+                    .bind(block.single_value())
```
lutter, line 522: Wow .. a subselect with Diesel? Nice!

store/postgres/tests/subgraph.rs
```diff
@@ -503,3 +504,44 @@ fn subgraph_error() {
...
+        let id = SubgraphDeploymentId::new("failUnfail").unwrap();
+        remove_subgraphs();
+        create_test_subgraph(&id, SUBGRAPH_GQL);
+        id
+    }
+
+    run_test_sequentially(setup, |store, id| async move {
+        let query_store = store
+            .query_store(
+                SubgraphDeploymentId::new("failUnfail").unwrap().into(),
```
lutter, line 520: That could be `id`

lutter (approved):

store/postgres/src/deployment.rs
leoyvens: Yhea I had fun with this.

store/postgres/tests/subgraph.rs
leoyvens: I refactored this and rebased.

leoyvens:
Since this already got a review and is an improvement in itself, I'll go ahead with it and do a follow up to remove errors when unfailing.

PR #2120, opened 2021-01-19
Make `deployment_schemas.network` mandatory

lutter:
This is the second part of the work mentioned in PR #2117

PR #2134, opened 2021-01-23
Sharding of the block store

lutter:
This PR makes it possible to store chain data (`ethereum_blocks`, `eth_call_cache`, `eth_call_meta`) in different database shards. Placement of the blocks for a chain/network into a shard is controlled through the configuration file that is also used for sharding subgraph data. 

The chain data tables might still contain data for several chains, which is only a problem if it is ever necessary to completely remove a chain, as the blocks and call cache table can easily contain tens of millions of rows and become very large; removing a chain requires deleting some of the data in those tables, which is much slower than dropping a dedicated per-chain table would be, and can wreak havoc on the organisation of the data on disk and in indexes. Essentially, with the current schema, deleting a chain is not advisable.

This PR is on top of PR #2120 since it requires some of the changes made there, but is otherwise ready for review.


store/postgres/src/chain_store.rs
```diff
@@ -206,6 +213,69 @@ impl ChainStore {
             },
         )
     }
+
+    /// Find the first block that is missing from the database needed to
+    /// complete the chain from block `hash` to the block with number
+    /// `first_block`. We return the hash of the missing block as an
+    /// array because the remaining code expects that, but the array will only
+    /// ever have at most one element.
```
leoyvens: The original sql I believe could in fact return an array, if you had a state in the DB such as:
```
*missing* <- B2 <- *missing* <- B4
```

Then iirc it would return both B1 and B3.

store/postgres/src/chain_store.rs
```diff
@@ -1177,18 +1424,10 @@ impl EthereumCallCache for ChainStore {
...
-                .optional()?
-            {
-                use public::eth_call_cache::dsl;
-
+            if let Some(return_value) = self.storage.get_call(conn, &old_id)? {
                 // Migrate to the new format by re-inserting the call and deleting the old entry.
                 self.set_call(contract_address, encoded_call, block, &return_value)?;
-                diesel::delete(eth_call_cache::table.filter(dsl::id.eq(old_id.as_ref())))
-                    .execute(conn)?;
+                self.storage.clear_cache(conn, old_id.as_ref())?;
```
leoyvens, line 1430: Now you've already gone through the trouble, but it is fine to drop this migration code by now.

store/postgres/src/chain_store.rs
lutter: I don't think that can happen since the recursive query will stop as soon as `b.parent_hash is null` - that sets `chain.last` to `true` and stops the loop.

store/postgres/src/chain_store.rs
lutter: Just pushed one more commit to get rid of the migration code.

leoyvens: Nice thanks

leoyvens (approved):
The psql to Rust migration of queries is correct for all I can tell. Approved, congrats on the massive refactor!

lutter:
Rebased to latest master

PR #2138, opened 2021-01-26
Fix max memory of the lfu cache

leoyvens:
`QUERY_CACHE_MAX_MEM` is the total memory budget, not per shard.


lutter:
Can we also add an environment variable that allows turning the `LfuCache` sharding off that's separate from the other sharded locks?

leoyvens:
@lutter Done

lutter (approved):

PR #2139, opened 2021-01-26
docker: Allow custom ethereum poll interval

cyrilledec:
This PR will allow to optionally specify ethereum poll interval when using graph-node's docker image

### DIFF
--- a/docker/start
+++ b/docker/start
@@ -51,10 +51,6 @@ wait_for_ipfs() {
 
 start_query_node() {
     export DISABLE_BLOCK_INGESTOR=true
-    # Set custom poll interval
-    if [ -n "$ethereum_polling_interval" ]; then
-        export ETHEREUM_POLLING_INTERVAL=$ethereum_polling_interval
-    fi
     graph-node \
         --postgres-url "$postgres_url" \
         --ethereum-rpc $ethereum \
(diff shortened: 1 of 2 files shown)


docker/start
```diff
@@ -51,6 +51,10 @@ wait_for_ipfs() {
 
 start_query_node() {
     export DISABLE_BLOCK_INGESTOR=true
+    # Set custom poll interval
+    if [ -n "$ethereum_polling_interval" ]; then
+        export ETHEREUM_POLLING_INTERVAL=$ethereum_polling_interval
+    fi
```
lutter, line 57: It would be better to have this block in the 'main' part of the script so that we don't need it in three different places; it should go either before or after setting `PGAPPNAME`

lutter (changes requested):

docker/start
cyrilledec: @lutter yes true, I've updated the script

lutter (approved):

cyrilledec:
@lutter clould you please re-run the following CI job : `Run tests (stable)` ?
It looks like previous run has failed due to a lack of space on the github action runner.

lutter:
> @lutter clould you please re-run the following CI job : `Run tests (stable)` ?
> It looks like previous run has failed due to a lack of space on the github action runner.

I just kicked off another run - we are still in the process of migrating to GH actions, and for now the thing I pay attention to is the Travis result. Let's see whether the additional run will fix it, if not, I'll merge since Travis had already passed.

PR #2148, opened 2021-01-29
Replace panic with error in the store

leoyvens:
Somehow a subgraph is triggering these panics and crashing nodes.


Jannis (approved):
LGTM

PR #2149, opened 2021-01-29
Delete fatal error on unfail

leoyvens:
This fixes the `has_non_fatal_errors` check in the case described here https://github.com/graphprotocol/graph-node/pull/2121#issuecomment-767538602.

But unfortunately it means that errors will be recreated upon restart and it will again be hard to tell which errors are new.

### DIFF
--- a/core/src/subgraph/instance_manager.rs
+++ b/core/src/subgraph/instance_manager.rs
@@ -455,6 +455,7 @@ where
     let store_for_err = ctx.inputs.store.cheap_clone();
     let logger = ctx.state.logger.cheap_clone();
     let id_for_err = ctx.inputs.deployment_id.clone();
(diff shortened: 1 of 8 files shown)


lutter:
I don't like the idea of going back to errors showing up as new on every node restart, especially since this is triggered by a pathological situation - being able to tell what's truly new is a huge improvement. 

What about deferring unfailing the deployment until we do the first successful `transact_block_operations` after a subgraph start? We'd have to be careful so we don't have to check whether the subgraph is failed every time we transact, but that would avoid recreating errors on every restart.

leoyvens:
@lutter I like that idea, it complements deleting the error on unfail. I've implemented it in the second commit.

lutter (approved):
Nice!

Issue #2145, opened 2021-01-28
Add missing BigInt host exports


leoyvens:
BigInt is currently missing some functionality, ideally we'd implement it in AS directly but for expediency lets add some more host functions for this since the Rust library already supports it. They are:
- Constructor from string: `BigInt.fromString`
- `@operator("|") BigInt.bitOr`
- `@operator("&") BigInt.bitAnd`
- `@operator("<<") BigInt.leftShift`
- `@operator(">>") BigInt.rightShift`

This involves the usual steps for a host function:
- Adding it to [graph-ts](https://github.com/graphprotocol/graph-ts), both the host function declaration and the BigInt APIs.
- Linking it into the [runtime](https://github.com/graphprotocol/graph-node/blob/master/runtime/wasm/src/module/mod.rs) and adding a Rust implementation.
- Testing it, I'd add it to our existing [big decimal test](https://github.com/graphprotocol/graph-node/tree/master/tests/integration-tests/big-decimal).

leoyvens:
Done

PR #2151, opened 2021-01-30
runtime: Add missing BigInt host exports

otaviopace:
Hello! This is my first Pull Request here, I am trying to solve this issue https://github.com/graphprotocol/graph-node/issues/2145.

I don't know much about the project, so correct me if I am doing anything wrong :slightly_smiling_face:

### DIFF
--- a/runtime/wasm/src/host_exports.rs
+++ b/runtime/wasm/src/host_exports.rs
@@ -560,10 +560,7 @@ impl HostExports {
         Ok(x.pow(exponent))
     }
 
-    pub(crate) fn big_int_from_string(
-        &self,
-        s: String,
-    ) -> Result<BigInt, DeterministicHostError> {
+    pub(crate) fn big_int_from_string(&self, s: String) -> Result<BigInt, DeterministicHostError> {
         BigInt::from_str(&s)
             .with_context(|| format!("string  is not a BigInt: '{}'", s))
             .map_err(DeterministicHostError)
--- a/runtime/wasm/src/host_exports.rs
+++ b/runtime/wasm/src/host_exports.rs
@@ -562,7 +562,7 @@ impl HostExports {
 
     pub(crate) fn big_int_from_string(&self, s: String) -> Result<BigInt, DeterministicHostError> {
         BigInt::from_str(&s)
-            .with_context(|| format!("string  is not a BigInt: '{}'", s))
+            .with_context(|| format!("string is not a BigInt: `{}`", s))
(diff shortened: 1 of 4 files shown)


runtime/wasm/src/host_exports.rs
```diff
@@ -560,6 +560,44 @@ impl HostExports {
         Ok(x.pow(exponent))
     }
 
+    pub(crate) fn big_int_from_string(&self, s: String) -> Result<BigInt, DeterministicHostError> {
+        BigInt::from_str(&s)
+            .with_context(|| format!("string  is not a BigInt: '{}'", s))
```
Jannis, line 565: There's a double space after "string". And I think our convention is to wrap `{}` in backticks rather than single quotes, like here: https://github.com/graphprotocol/graph-node/pull/2151/files#diff-9f6f2b4b1457abc5d61a58aa2ba3f2b3a2a4f2010def91f76dca50d4780c2f38L547

tests/integration-tests/big-decimal/src/mapping.ts
```diff
@@ -7,7 +7,7 @@ let one = BigDecimal.fromString("1");
 export function handleTrigger(event: Trigger): void {
   // There are 35 digits after the dot.
   // big_decimal exponent will be: -35 - 6109 = -6144.
-  // Minmum exponent is: -6143.
+  // Minimum exponent is: -6143.
```
Jannis, line 10: Good catch 😄

Jannis (changes requested):
Looks good. If you could update the error message, I'd say this is good to go in!

runtime/wasm/src/host_exports.rs
otaviopace: Fixed!

otaviopace:
@Jannis so, in the issue @leoyvens wrote that some tests should be added [here](https://github.com/graphprotocol/graph-node/blob/master/tests/integration-tests/big-decimal/src/mapping.ts#L45).

I should just call the new `BigInt` functions there with some sample assertions?

leoyvens:
@otaviopace That's my suggestion. Tests are actually failing on master right now, that will be fixed shortly.

otaviopace:
@Jannis @leoyvens I've added the tests, but I don't know if I am doing something wrong, the pipeline is failing.

otaviopace:
Yaaaay :tada: :confetti_ball: 
Thanks for the help :blush:

Issue #2168, opened 2021-02-07
Postgres schema errors in log


zmanian:
We are deploying a local graph node.

We are running postgres 13.1. We don't see any postgres version recommendations in the documentation. 

We see errors like below after deploying the uniswap subgraph.
```
Feb 07 16:50:30.334 ERRO call cache get error, error: missing FROM-clause entry for table "eth_call_meta", runtime_host: 1/1, block_hash: 0x9c8330580e8a2c8841d244ac455cd57205b2a76106bef1d5a85c2a0d4364cf4e, block_number: 3481503, subgraph_id: QmaeD5E3U7uqg3DR61ZgzpzPoXwxPh16UU4SL71SrnjWa2, component: SubgraphInstanceManager
Feb 07 16:50:30.342 ERRO call cache set error, error: ON CONFLICT DO UPDATE requires inference specification or constraint name, runtime_host: 1/1, block_hash: 0x9c8330580e8a2c8841d244ac455cd57205b2a76106bef1d5a85c2a0d4364cf4e, block_number: 3481503, subgraph_id: QmaeD5E3U7uqg3DR61ZgzpzPoXwxPh16UU4SL71SrnjWa2, component: SubgraphInstanceManager
Feb 07 16:50:30.346 ERRO call cache get error, error: missing FROM-clause entry for table "eth_call_meta", runtime_host: 1/1, block_hash: 0x9c8330580e8a2c8841d244ac455cd57205b2a76106bef1d5a85c2a0d4364cf4e, block_number: 3481503, subgraph_id: QmaeD5E3U7uqg3DR61ZgzpzPoXwxPh16UU4SL71SrnjWa2, component: SubgraphInstanceManager
```

Any suggestions for further root cause analysis?

zmanian:
We are running GCP Cloud SQL instances.

Just tried Postgres 11 and got the same thing.

lutter:
It's not an issue with the PG version - we run PG 13 in production w/o issue. You stumbled into a bug in how we generate a certain SQL query that was recently introduced. To get around that you can either go back to an official release, or apply PR #2169 to your checkout and try with that.

zmanian:
thanks!

PR #2169, opened 2021-02-07
store: Fix call_cache queries when the chain is in a separate namespace

lutter:
Fixes https://github.com/graphprotocol/graph-node/issues/2168

### DIFF
--- a/store/postgres/src/chain_store.rs
+++ b/store/postgres/src/chain_store.rs
@@ -179,6 +179,7 @@ mod data {
 
     impl CallMetaTable {
         const TABLE_NAME: &'static str = "call_meta";
+        const ACCESSED_AT: &'static str = "accessed_at";
 
         fn new(namespace: &str) -> Self {
             CallMetaTable {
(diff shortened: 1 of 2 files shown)


leoyvens:
For my understanding, how does this change make the code better handle a separate namespace?

lutter:
> For my understanding, how does this change make the code better handle a separate namespace?

It doesn't improve anything, it just makes it work ;) Before the change, the queries were just incorrect when a chain had its own namespace.

leoyvens:
@lutter I just wanted to understand how `"CURRENT_DATE > eth_call_meta.accessed_at"` is functionally different from

```
sql(&format!(
                            "CURRENT_DATE > {}.{}",
                            CallMetaTable::TABLE_NAME,
                            CallMetaTable::ACCESSED_AT
                        )),
```

lutter:
> @lutter I just wanted to understand how `"CURRENT_DATE > eth_call_meta.accessed_at"` is functionally different from
> 
> ```rust
> sql(&format!(
>                             "CURRENT_DATE > {}.{}",
>                             CallMetaTable::TABLE_NAME,
>                             CallMetaTable::ACCESSED_AT
> )),
> ```

Aah .. now I understand the confusion: the original query had the wrong table name. `CallMetaTable::TABLE_NAME` boils down to `call_meta`, which is actually used in the query whereas `eth_call_meta` is not. Chains that live in their own namespace have tables `blocks`, `call_cache` and `call_meta` in their namespace that mirror `ethereum_block`, `eth_call_cache` and `eth_call_meta` in the public schema.

leoyvens (approved):

leoyvens:
@lutter ah so we don't need to explicitly handle the namespace. Thanks for the clarification.

PR #2187, opened 2021-02-12
Address a couple issues with the sharded block store

lutter:


node/src/main.rs
```diff
@@ -184,10 +185,15 @@ async fn main() {
...
         PrometheusMetricsServer::new(&logger_factory, prometheus_registry.clone());
 
-    // Ethereum clients
-    let eth_networks = create_ethereum_networks(logger.clone(), metrics_registry.clone(), &config)
-        .await
-        .expect("Failed to parse Ethereum networks");
+    // Ethereum clients; query nodes ignore all ethereum clients and never
+    // connect to them directly
+    let eth_networks = if query_only {
+        EthereumNetworks::new()
```
fordN, line 191: Nice! Finally ridding the query nodes of their unnecessary Ethereum connection burden.

store/postgres/src/block_store.rs
```diff
@@ -125,52 +171,119 @@ impl BlockStore {
...
+        let sender = ChainHeadUpdateSender::new(self.primary.clone(), chain.name.clone());
+        let store = ChainStore::new(
+            chain.name.clone(),
+            chain.storage.clone(),
+            chain.network_identifier()?,
+            self.chain_head_update_listener.clone(),
+            sender,
+            pool,
+        );
+        let store = Arc::new(store);
```
fordN, line 215: Any reason to keep setting the store and putting it in an Arc on separate lines?  Not a big deal either way but I'm curious.

fordN (approved):
Looks good! Thanks for the quick turn around today.

store/postgres/src/block_store.rs
lutter: This is also called from the `store()` method which needs to return that, too - I could have also done another lookup in the map, but this seemed easier.

PR #2195, opened 2021-02-17
node: ignore non-deterministic test

tilacog:
The test function `node_fails_to_start_with_invalid_ipfs` relies on
async and results in variable results, so we are ignoring it for now.

### DIFF
--- a/node/tests/cli.rs
+++ b/node/tests/cli.rs
@@ -29,6 +29,7 @@ fn node_fails_to_start_with_unnamed_ethereum_network() {
 }
 
 #[test]
+#[ignore = "non deterministic"]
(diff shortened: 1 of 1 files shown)


tilacog:
More non-deterministic tests could also be included in this same PR, if desired.

node/tests/cli.rs
```diff
@@ -29,6 +29,7 @@ fn node_fails_to_start_with_unnamed_ethereum_network() {
 }
 
 #[test]
+#[ignore = "non deterministic"]
```
leoyvens, line 32: I didn't know you could give a reason to ignore, nice!

leoyvens (approved):

PR #2196, opened 2021-02-17
Add CORS for all HTTP responses

### LINKED ISSUE #2193: Make sure all http responses allow cors
Right now we don't always set the allow cors headers that we should. In the some places we do allow them correctly like [here](https://github.com/graphprotocol/graph-node/blob/master/graph/src/data/query/result.rs#L159) but in others like [here](https://github.com/graphprotocol/graph-node/blob/master/server/http/src/service.rs#L355) we don't. We should always allow it.

tilacog:
Resolves #2193

This PR inserts `Access-Control-Alllow-Origin: *` on all response headers.

Also, all header names were refactored to constants so the type checker can protect us from possible typing mistakes.

### DIFF
--- a/server/http/src/test_utils.rs
+++ b/server/http/src/test_utils.rs
@@ -60,6 +60,7 @@ pub fn assert_error_response(
         .clone()
 }
 
+#[track_caller]
 pub fn assert_expected_headers(response: &Response<Body>) {
     assert_eq!(
         response
--- a/server/http/src/test_utils.rs
+++ b/server/http/src/test_utils.rs
@@ -1,14 +1,14 @@
 use graph::prelude::serde_json;
 use graph::prelude::*;
 use http::StatusCode;
-use hyper::{Body, Response};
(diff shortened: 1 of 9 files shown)


server/http/src/test_utils.rs
```diff
@@ -59,3 +59,13 @@ pub fn assert_error_response(
         .expect("GraphQL \"errors\" field must be a vector")
         .clone()
 }
+
+pub fn assert_expected_headers(response: &Response<Body>) {
```
lutter, line 63: If you annotate this function with `#[track_caller]`, a failing test will point to where this function is called, rather than at the `assert_eq!` in here. Not a huge deal, but can make life a little easier when things go wrong

lutter (approved):
Nice! I am a big fan of using constants instead of bare strings as much as possible, too.

server/http/src/test_utils.rs
tilacog: That's awesome! I didn't know that.
I'm sending a new commit with that addition.

PR #2205, opened 2021-02-19
Let one deployment rule match multiple networks

lutter:


node/src/config.rs
```diff
@@ -655,6 +657,31 @@ impl Default for Predicate {
     }
 }
 
+#[derive(Clone, Debug, Deserialize, Serialize)]
+#[serde(untagged)]
+enum NetworkPredicate {
+    Single(String),
+    Many(Vec<String>),
```
leoyvens, line 664: The code would be a bit nicer if we use a `BTreeSet` here instead of `Vec`. Not that it should make much of a difference so feel free to merge as is.

leoyvens (approved):

PR #2206, opened 2021-02-20
node: Add `graphman query`

lutter:
This makes it possible to execute a GraphQL query with

```
graphman query sub/graph '{ .. GraphQL .. }' var1=val1 ...
```

By passing `GRAPH_LOG_QUERY_TIMING=gql,sql GRAPH_LOG=debug` in the environment, the command will log a lot of info about query execution.

For now, it's probably best to not run this against production directly though. Amongst other things, the code sets up notification listeners as a side-effect of creating the store.

### DIFF
--- a/node/src/bin/manager.rs
+++ b/node/src/bin/manager.rs
@@ -2,8 +2,9 @@ use std::{collections::HashMap, env, sync::Arc};
 
 use config::PoolSize;
 use git_testament::{git_testament, render_testament};
-use graph::prometheus::Registry;
+use graph::{data::graphql::effort::LoadManager, prometheus::Registry};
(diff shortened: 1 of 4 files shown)


leoyvens:
Should this print a warning about the notification listener and not running it against prod?

lutter:
> Should this print a warning about the notification listener and not running it against prod?

I think the real solution is to change the code to not set up notification listeners by default; I think all that is needed is getting the `ChainHeadUpdateListener` our of the `ChainStore` - it really has no business being in there anyway, and is just used as a convenience to make it easier to pass both the listener and the store around.

leoyvens (approved):
Got it, if we want to merge as is and improve later, fine by me.

PR #2210, opened 2021-02-22
Fix CI

tilacog:


leoyvens:
Looks like there are redundant semicolon to hunt down.

leoyvens (approved):

PR #2213, opened 2021-02-23
Enable the `order by` optimization in entity queries in a few more queries

lutter:


lutter:
Somethign went wrong with my branch. Will try again.

PR #2218, opened 2021-02-24
Better error messages for `SubgraphManifestValidationError`

### LINKED ISSUE #1763: Update error text on SourceAddressRequired
More of a documentation update

I was seeing this error and it took me awhile to figure out the cause, though in retrospect this is kinda obvious and makes sense:
`subgraph validation error: [SourceAddressRequired]`

The issue is that I had call handlers in my manifest but the docs online do not mention that these mean I need to specify a source.

Make a subgraph with call or block handlers and leave the dataSource.source field empty. The relevant code is here:
https://github.com/graphprotocol/graph-node/blob/ecb61ca75b388cc8c92de3fe0f5a0ebd499c3b7c/graph/src/data/subgraph/mod.rs#L968

Perhaps update the error text to mention that call and block handlers cannot have a blank source address. It's already in the code comment right there. Would save others so time perhaps.

tilacog:
Resolves #1763

### DIFF
--- a/graph/src/data/subgraph/mod.rs
+++ b/graph/src/data/subgraph/mod.rs
@@ -304,7 +304,7 @@ pub enum SubgraphRegistrarError {
     QueryExecutionError(QueryExecutionError),
     #[error("subgraph registrar error with store: {0}")]
     StoreError(StoreError),
-    #[error("subgraph validation error: {0:?}")]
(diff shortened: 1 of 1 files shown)


leoyvens (approved):
Feel free to merge once CI passes.

PR #2219, opened 2021-02-24
Add Content Type headers

### LINKED ISSUE #1859: /graphql endpoints should return a Content-Type header
Bug

`/graphql` endpoints do not return a Content-Type header; this leads to some proxies (i.e. Traefik) to 'auto-detect' the MIME type as text/plain causing the browser to not render the page.

The `/graphql` endpoint emitting a correct Content-Type header would mean that all content is rendered as expected, and not left to the behest of proxies.

N/A

Header is emitted `Content-Type: text/html`

tilacog:
Resolves #1859

### DIFF
--- a/server/http/src/service.rs
+++ b/server/http/src/service.rs
@@ -124,7 +124,7 @@ where
         Ok(Response::builder()
             .status(200)
             .header(ACCESS_CONTROL_ALLOW_ORIGIN, "*")
-            .header(CONTENT_TYPE, "text/html")
+            .header(CONTENT_TYPE, "text/plain")
             .body(Body::from(String::from(
                 "Access deployed subgraphs by deployment ID at \
                 /subgraphs/id/<ID> or by name at /subgraphs/name/<NAME>",
@@ -244,7 +244,7 @@ where
                     .status(StatusCode::FOUND)
                     .header(ACCESS_CONTROL_ALLOW_ORIGIN, "*")
                     .header(LOCATION, loc_header_val)
-                    .header(CONTENT_TYPE, "text/html")
+                    .header(CONTENT_TYPE, "text/plain")
                     .body(Body::from("Redirecting..."))
                     .unwrap()
             })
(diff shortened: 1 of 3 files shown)


server/http/src/service.rs
```diff
@@ -124,6 +124,7 @@ where
         Ok(Response::builder()
             .status(200)
             .header(ACCESS_CONTROL_ALLOW_ORIGIN, "*")
+            .header(CONTENT_TYPE, "text/html")
```
lutter, line 127: That one needs to be `text/plain` since we aren't actually sending HTML but plain text.

server/http/src/service.rs
```diff
@@ -240,6 +244,7 @@ where
                     .status(StatusCode::FOUND)
                     .header(ACCESS_CONTROL_ALLOW_ORIGIN, "*")
                     .header(LOCATION, loc_header_val)
+                    .header(CONTENT_TYPE, "text/html")
```
lutter, line 247: This should also be `text/plain`

server/http/src/service.rs
```diff
@@ -250,6 +255,7 @@ where
         async {
             Ok(Response::builder()
                 .status(StatusCode::NOT_FOUND)
+                .header(CONTENT_TYPE, "text/html")
```
lutter, line 258: And another `text/plain`

server/http/src/service.rs
```diff
@@ -346,7 +352,7 @@ where
                 Ok(response) => Ok(response),
                 Err(err @ GraphQLServerError::ClientError(_)) => Ok(Response::builder()
                     .status(400)
-                    .header(CONTENT_TYPE, "text/plain")
+                    .header(CONTENT_TYPE, "text/html")
```
lutter, line 355: `text/plain` here was correct, I think. The `err` just contains random text, it's not marked up as HTML

server/http/src/service.rs
```diff
@@ -355,7 +361,7 @@ where
 
                     Ok(Response::builder()
                         .status(400)
-                        .header(CONTENT_TYPE, "text/plain")
+                        .header(CONTENT_TYPE, "text/html")
```
lutter, line 364: That was also fine

server/http/src/service.rs
```diff
@@ -365,7 +371,7 @@ where
 
                     Ok(Response::builder()
                         .status(500)
-                        .header(CONTENT_TYPE, "text/plain")
+                        .header(CONTENT_TYPE, "text/html")
```
lutter, line 374: That was also fine.

lutter (approved):
Looks good. Just fix up the few content types i commented on, and this is good to go.

PR #2225, opened 2021-02-26
Document implementation details of the store architecture

lutter:


leoyvens:
To improve the discoverability of these docs, it might help duplicate the `docs/implementation/README.md` into maybe a `store/README.md`.

leoyvens (approved):
Great to bring these in-tree!

PR #2239, opened 2021-03-01
Make block pointers less Ethereum specific

lutter:
This PR changes `EthereumBlockPointer` so that the hash is `Bytes`(`Vec<u8>`) and the block number is a `BlockNumber` (`i32`). The PR does not rename the struct yet, but that should be a fairly easy change.

### DIFF
--- a/graph/src/components/ethereum/types.rs
+++ b/graph/src/components/ethereum/types.rs
@@ -11,12 +11,12 @@ use web3::types::{
     U128, U256, U64,
 };
 
-use crate::prelude::{EntityKey, SubgraphDeploymentId, ToEntityKey};
+use crate::prelude::{BlockNumber, EntityKey, SubgraphDeploymentId, ToEntityKey};
 
 pub type LightEthereumBlock = Block<Transaction>;
 
 pub trait LightEthereumBlockExt {
-    fn number(&self) -> u64;
+    fn number(&self) -> BlockNumber;
     fn transaction_for_log(&self, log: &Log) -> Option<Transaction>;
     fn transaction_for_call(&self, call: &EthereumCall) -> Option<Transaction>;
     fn parent_ptr(&self) -> Option<EthereumBlockPointer>;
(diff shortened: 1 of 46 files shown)


graph/src/components/ethereum/types.rs
```diff
@@ -416,25 +414,26 @@ impl Clone for EthereumCallData {
 /// Maximum block number supported: 2^63 - 1
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
 pub struct EthereumBlockPointer {
-    pub hash: H256,
+    pub hash: Bytes,
```
leoyvens, line 417: Lets make this just a `Vec<u8>` so it is not tied to a type from the web3 crate.

graph/src/components/ethereum/types.rs
```diff
@@ -414,7 +414,7 @@ impl Clone for EthereumCallData {
 /// A block hash and block number from a specific Ethereum block.
 ///
 /// Maximum block number supported: 2^63 - 1
-#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
```
leoyvens, line 415: Lets make it `CheapClone`.

leoyvens (changes requested):
The `i32` block numbers seems like the way to go for now, I'll update the RFC to reflect the implementation choices.

graph/src/components/ethereum/types.rs
lutter: I actually made it `Box<[u8]>` to save a word of storage; I didn't use `scalar::Bytes`, since that has a scary `to_string()` that prefixes with `0x`, and using that seems too dangerous.

graph/src/components/ethereum/types.rs
lutter: Done

lutter:
> The `i32` block numbers seems like the way to go for now, I'll update the RFC to reflect the implementation choices.

Nice. A maximum of 2G block numbers is really not that big a deal: that's 68 years worth of seconds. We'll probably have to bump it to `i64` at some point, but that point is several years out, and it's somewhat tricky since we'll need to be able to deal with subgraphs that use `i32` and subgraphs that use `i64` simultaneously. Keeping our representation of block numbers consistent internally, and only converting when we take block numbers from external sources is a first step in making that migration a bit more feasible.

graph/src/components/ethereum/types.rs
leoyvens: Nice!

leoyvens (approved):

graph/src/components/ethereum/types.rs
leoyvens: It's funny how we use `Box<[u8]>` but nobody ever thinks of using `Box<str>`.

PR #2242, opened 2021-03-04
Log non-fatal errors

leoyvens:
We don't want any errors to be silent in the logs.

### DIFF
--- a/core/src/subgraph/instance_manager.rs
+++ b/core/src/subgraph/instance_manager.rs
@@ -929,6 +929,14 @@ where
         info!(&logger, "Applying {} entity operation(s)", mods.len());
     }
 
+    let err_count = block_state.deterministic_errors.len();
+    for (i, e) in block_state.deterministic_errors.iter().enumerate() {
+        error!(&logger, "Subgraph error {}/{}", i + 1, err_count;
+            "error" => e.to_string(),
+            "code" => LogCode::SubgraphSyncingFailure
+        );
+    }
+
     // Transact entity operations into the store and update the
     // subgraph's block stream pointer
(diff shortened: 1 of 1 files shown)


core/src/subgraph/instance_manager.rs
```diff
@@ -929,6 +929,22 @@ where
         info!(&logger, "Applying {} entity operation(s)", mods.len());
     }
 
+    if !block_state.deterministic_errors.is_empty() {
+        info!(
```
lutter, line 933: Does it make sense to log this at info? Maybe when you log the individual errors, have something like `Subgraph error (3/17)` in the log message.

core/src/subgraph/instance_manager.rs
```diff
@@ -929,6 +929,22 @@ where
...
+            "Applying {} error(s)",
+            block_state.deterministic_errors.len()
+        );
+    }
+
+    for e in &block_state.deterministic_errors {
+        error!(&logger, "Subgraph error";
+            "error" => e.to_string(),
+            "id" => ctx.inputs.deployment_id.to_string(),
+            "code" => LogCode::SubgraphSyncingFailure
```
lutter, line 944: Nice; I didn't know we had `LogCode` - we should probably use it more pervasively

lutter (approved):

core/src/subgraph/instance_manager.rs
leoyvens: Yhea Jorge introduced it

core/src/subgraph/instance_manager.rs
leoyvens: That is better I've updated the PR. Also removed the `id`, it's implicit in the logger context.

PR #2246, opened 2021-03-05
server: parameterize content type in `serve_file` function

### LINKED ISSUE #2245: GraphQL playground is broken
Hello,

I'm doing a new installation and faced this issue when trying to run some queries to check my subgraph. I've realized that's happening since the latest pushed graph-node images.
For instance, the image `graphprotocol/graph-node:ebc5f79` hasn't this problem.

tilacog:
Resolves #2245

### DIFF
--- a/server/http/src/service.rs
+++ b/server/http/src/service.rs
@@ -133,12 +133,16 @@ where
     }
 
     /// Serves a static file.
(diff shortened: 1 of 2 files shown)


lutter (approved):
Nice, thanks!

PR #2247, opened 2021-03-06
docker: Make it possible to use an (experimental) configuration file

lutter:


tilacog (approved):

tilacog:
LGTM.

As a minor improvement, we could use a `PathBuf` instead of a `String` in the `structopt` file fields, so the value arrives in our code already validated/converted. It is probably being validated/converted down the line, but It might come in handy to do it during the argument parsing phase.

PR #2259, opened 2021-03-09
Improvements to start block handling

### LINKED ISSUE #2007: Call handlers with a start block higher than other start blocks make subgraph indexing unbearably slow

This is a bug.

This [Keep Project](https://thegraph.com/explorer/subgraph/miracle2k/all-the-keeps) subgraph takes about 4 hours to build.

Then I add the following to `subgraph.yaml`:

```
  - kind: ethereum/contract
    name: ECDSARewards
    network: mainnet
    source:
      startBlock: 11193245
      address: "0xc5aC5A8892230E0A3e1c473881A2de7353fFcA88"
      abi: ECDSARewards
    mapping:
      kind: ethereum/events
      apiVersion: 0.0.4
      language: wasm/assemblyscript
      entities:
        - StakedropInterval
      abis:
        - name: ECDSARewards
          file: ./abis/ECDSARewards.json
      eventHandlers:
        - event: RewardReceived(bytes32,uint256)
          handler: handleECDSARewardReceivedEvent
      callHandlers:
        - function: allocateRewards(uint256)
          handler: handleAllocateECDSARewards
      file: ./src/mappingStakedrop.ts
```

The start block here is a pretty recent block, and there should only be a couple transactions that match. Yet the graph now takes almost 24 hours to index. Right from the start, there are many message such as:

``` Received 18478 traces for blocks [10021376, 10021425], component: BlockStream```

That's a lot of traces! Looking behind the scenes, I can see that the eth node is being called with:

```
{"jsonrpc":"2.0","method":"trace_filter","params":[{"fromBlock":"0x9faab9","toBlock":"0x9faae0"}],"id":624632}
```

Notably, this does not include any address filter, which explains why so many traces are returned.

I believe the issue may be in this line:

https://github.com/graphprotocol/graph-node/blob/d1ed689e94c5179637826faf622746f0473d96a6/chain/ethereum/src/ethereum_adapter.rs#L1174

The problem is that the earliest block the graph needs to look at is 9958365, due to a different contract specification in `subgraph.yaml`. Then, when processing the call handlers for this contract (with a start block of 11193245), all the addresses are filtered out, leaving an empty list, and thus a call to `trace_filter` that is not restricted to any address.

Consequently, changing the `startBlock` value which is affecting the call handlers to the lowest value used anywhere in subgraph.yaml brings the indexing times back to the expected level.

Another note is that this does not seem to affect callHandlers used in templates, as those were already used in this subgraph and did not cause issues.

Similar, but likely separate issues:

https://github.com/graphprotocol/graph-node/issues/1216
https://github.com/graphprotocol/graph-node/issues/1395

leoyvens:
Three independent commits with improvements related to start blocks. Resolves #2007.

### DIFF
--- a/chain/ethereum/src/block_stream.rs
+++ b/chain/ethereum/src/block_stream.rs
@@ -362,17 +362,21 @@ where
                             // then we start with the genesis block
                             let from = subgraph_ptr.map_or(0, |ptr| ptr.number + 1);
 
-                            // Get the next subsequent data source start block to ensure the block range
-                            // is aligned with data source.
+                            // Get the next subsequent data source start block to ensure the block
+                            // range is aligned with data source. This is not necessary for
+                            // correctness, but it avoids an ineffecient situation such as the range
+                            // being 0..100 and the start block for a data source being 99, then
+                            // `calls_in_block_range` would request unecessary traces for the blocks
+                            // 0 to 98 because the start block is within the range.
                             let next_start_block: BlockNumber = start_blocks
                                 .into_iter()
                                 .filter(|block_num| block_num > &from)
                                 .min()
                                 .unwrap_or(BLOCK_NUMBER_MAX);
 
-                            // End either just before the the next data source start_block or
-                            // just prior to the reorg threshold. It isn't safe to go any farther
-                            // due to race conditions.
+                            // End either just before the the next data source start_block or just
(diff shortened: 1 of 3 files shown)


leoyvens:
@fordN thanks for the review!

PR #2263, opened 2021-03-11
Log subgraph errors in a single line

leoyvens:
Stacktraces look nicer on separate lines, but that messes with how we sort logs based on the id tag. Not sure why this only became a problem now, maybe a slog dependency bump changed something.

Also log block hashes as hex strings.

### DIFF
--- a/core/src/subgraph/instance_manager.rs
+++ b/core/src/subgraph/instance_manager.rs
@@ -494,7 +494,7 @@ where
                         logger,
                         "Reverting block to get back to main chain";
                         "block_number" => format!("{}", subgraph_ptr.number),
-                        "block_hash" => format!("{}", subgraph_ptr.hash_as_h256())
+                        "block_hash" => format!("{}", subgraph_ptr.hash)
                     );
 
                     // We would like to revert the DB state to the parent of the current block.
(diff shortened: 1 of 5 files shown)


core/src/subgraph/instance_manager.rs
```diff
@@ -627,16 +627,17 @@ where
 
                 // Handle unexpected stream errors by marking the subgraph as failed.
                 Err(e) => {
+                    let message = format!("{:#}", e).replace("\n", " ");
```
lutter, line 630: It might be nicer to replace `\n` with `\t` because then it is possible to get the nice formatting back by reversing the substitution (e.g., when you downloaded a log and want to study the backtrace in more detail)

core/src/subgraph/instance_manager.rs
```diff
@@ -923,8 +924,9 @@ where
 
     let err_count = block_state.deterministic_errors.len();
     for (i, e) in block_state.deterministic_errors.iter().enumerate() {
+        let message = format!("{:#}", e).replace("\n", " ");
```
lutter, line 927: Same here

runtime/wasm/src/module/mod.rs
```diff
@@ -242,15 +242,17 @@ impl WasmInstance {
         };
 
         if let Some(deterministic_error) = deterministic_error {
+            let message = format!("{:#}", deterministic_error).replace("\n", " ");
```
lutter, line 245: And here

lutter (approved):

leoyvens:
@lutter done, nice suggestion

PR #2267, opened 2021-03-12
Change how subgraph_deployment is organized

lutter:
This PR is a steptowards using the same `id` as the primary key as `deployment_schemas`. It sits on top of #2266 since the former can easily be rolled back in the hosted service if anything goes wrong. This PR contains an irreversible migration and should therefore be rolled out by itself.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -54,9 +54,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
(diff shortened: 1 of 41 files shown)


tilacog (approved):

lutter:
This is merged, but github has trouble tracking this because of the different base branch

PR #2285, opened 2021-03-16
store: Allow enforcing a timeout for SQL queries

lutter:
This isn't great since the timeout is set through an environment variable - it would be better to derive it from the overall timeout for GraphQL queries, but that's a little more work.

### DIFF
--- a/store/postgres/src/relational.rs
+++ b/store/postgres/src/relational.rs
@@ -6,7 +6,7 @@
 //!
 //! The pivotal struct in this module is the `Layout` which handles all the
 //! information about mapping a GraphQL schema to database tables
-use diesel::connection::SimpleConnection;
+use diesel::{connection::SimpleConnection, Connection};
 use diesel::{debug_query, OptionalExtension, PgConnection, RunQueryDsl};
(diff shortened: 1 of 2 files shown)


store/postgres/src/relational.rs
```diff
@@ -64,6 +64,20 @@ lazy_static! {
             .map(|v| v.split(",").map(|s| s.to_owned()).collect())
             .unwrap_or(HashSet::new())
     };
+
+    /// `GRAPH_SQL_STATEMENT_TIMEOUT` is the timeout for queries in seconds.
+    /// If it is not set, no statement timeout will be enforced. The statement
+    /// timeout is local, i.e., can only be used within a transaction and
+    /// will be cleared at the end of the transaction
+    static ref STATEMENT_TIMEOUT: Option<String> = {
+        env::var("GRAPH_SQL_STATEMENT_TIMEOUT")
```
leoyvens, line 73: This is important, lets document it in `environment_variables.md`.

leoyvens (changes requested):

leoyvens (approved):

PR #2286, opened 2021-03-16
Optimize entity modifications queries

### LINKED ISSUE #2223: Change `transact_block_operations` in the store to doing bulk operations
When we apply changes for a block, we issue individual SQL queries for each entity. Instead, we should change these operations to do bulk updates, for example, to clamp all entities that need it in one query. Ideally, transacting operations for a block will only issue 2 queries per block and entity type, one to update the block range of updated/deleted entities, and one to insert new entity versions.

tilacog:
Resolves #2223

### DIFF
--- a/store/postgres/src/deployment_store.rs
+++ b/store/postgres/src/deployment_store.rs
@@ -382,7 +382,7 @@ impl DeploymentStore {
                     removals
                         .entry(key.entity_type.clone())
                         .or_insert_with(Vec::new)
-                        .push(key);
+                        .push(key.entity_id);
                 }
             }
         }
@@ -392,11 +392,11 @@ impl DeploymentStore {
             count +=
                 self.insert_entities(entity_type, entities, conn, layout, ptr, &stopwatch)? as i32
         }
-        for (entity_type, entities) in overwrites.iter_mut() {
+        for (entity_type, entities) in overwrites.drain() {
             // not updating count since the number of entities remains the same
             self.overwrite_entities(entity_type, entities, conn, layout, ptr, &stopwatch)?;
         }
-        for (entity_type, entity_keys) in removals.into_iter() {
+        for (entity_type, entity_keys) in removals.iter() {
             count -=
                 self.remove_entities(entity_type, entity_keys, conn, layout, ptr, &stopwatch)?
                     as i32;
@@ -426,8 +426,8 @@ impl DeploymentStore {
 
     fn overwrite_entities(
         &self,
-        entity_type: &EntityType,
-        data: &mut Vec<(EntityKey, Entity)>,
+        entity_type: EntityType,
+        data: Vec<(EntityKey, Entity)>,
         conn: &PgConnection,
         layout: &Layout,
         ptr: &EthereumBlockPointer,
@@ -446,8 +446,8 @@ impl DeploymentStore {
 
     fn remove_entities(
         &self,
-        entity_type: EntityType,
-        entity_keys: Vec<EntityKey>,
+        entity_type: &EntityType,
+        entity_keys: &Vec<String>,
         conn: &PgConnection,
         layout: &Layout,
         ptr: &EthereumBlockPointer,
@@ -455,13 +455,7 @@ impl DeploymentStore {
     ) -> Result<usize, StoreError> {
         let _section = stopwatch.start_section("apply_entity_modifications_delete");
         layout
-            .delete(
-                conn,
-                entity_type,
-                &entity_keys,
-                block_number(ptr),
-                stopwatch,
-            )
+            .delete(conn, entity_type, entity_keys, block_number(ptr), stopwatch)
             .map_err(|_error| anyhow!("Failed to remove entities: {:?}", entity_keys).into())
     }
 
--- a/store/postgres/src/deployment_store.rs
+++ b/store/postgres/src/deployment_store.rs
@@ -386,17 +386,20 @@ impl DeploymentStore {
                 }
             }
         }
+
         // apply modification groups
         for (entity_type, entities) in inserts.iter_mut() {
-            count += self.insert_entities(entity_type, entities, conn, layout, ptr, &stopwatch)?;
+            count +=
+                self.insert_entities(entity_type, entities, conn, layout, ptr, &stopwatch)? as i32
         }
         for (entity_type, entities) in overwrites.iter_mut() {
-            count +=
-                self.overwrite_entities(entity_type, entities, conn, layout, ptr, &stopwatch)?;
+            // not updating count since the number of entities remains the same
+            self.overwrite_entities(entity_type, entities, conn, layout, ptr, &stopwatch)?;
         }
         for (entity_type, entity_keys) in removals.into_iter() {
-            count +=
-                self.remove_entities(entity_type, entity_keys, conn, layout, ptr, &stopwatch)?;
+            count -=
+                self.remove_entities(entity_type, entity_keys, conn, layout, ptr, &stopwatch)?
+                    as i32;
         }
         Ok(count)
     }
@@ -409,7 +412,7 @@ impl DeploymentStore {
         layout: &Layout,
         ptr: &EthereumBlockPointer,
         stopwatch: &StopwatchMetrics,
-    ) -> Result<i32, StoreError> {
+    ) -> Result<usize, StoreError> {
         let section = stopwatch.start_section("check_interface_entity_uniqueness");
         for (key, _) in data.iter() {
             // WARNING: This will potentially execute 2 queries for each entity key.
@@ -418,9 +421,7 @@ impl DeploymentStore {
         section.end();
 
         let _section = stopwatch.start_section("apply_entity_modifications_insert");
-        layout
-            .insert(conn, entity_type, data, block_number(ptr))
-            .map(|_| 1)
+        layout.insert(conn, entity_type, data, block_number(ptr))
     }
 
     fn overwrite_entities(
@@ -431,7 +432,7 @@ impl DeploymentStore {
         layout: &Layout,
         ptr: &EthereumBlockPointer,
         stopwatch: &StopwatchMetrics,
-    ) -> Result<i32, StoreError> {
+    ) -> Result<usize, StoreError> {
         let section = stopwatch.start_section("check_interface_entity_uniqueness");
         for (key, _) in data.iter() {
             // WARNING: This will potentially execute 2 queries for each entity key.
@@ -440,9 +441,7 @@ impl DeploymentStore {
         section.end();
 
         let _section = stopwatch.start_section("apply_entity_modifications_update");
-        layout
-            .update(conn, entity_type, data, block_number(ptr))
-            .map(|_| 0)
+        layout.update(conn, entity_type, data, block_number(ptr))
     }
 
     fn remove_entities(
@@ -453,12 +452,10 @@ impl DeploymentStore {
         layout: &Layout,
         ptr: &EthereumBlockPointer,
         stopwatch: &StopwatchMetrics,
-    ) -> Result<i32, StoreError> {
+    ) -> Result<usize, StoreError> {
         let _section = stopwatch.start_section("apply_entity_modifications_delete");
         layout
             .delete(conn, entity_type, &entity_keys, block_number(ptr))
-            // This conversion is ok since n will only be 0 or 1
-            .map(|n| -(n as i32))
             .map_err(|_error| anyhow!("Failed to remove entities: {:?}", entity_keys).into())
     }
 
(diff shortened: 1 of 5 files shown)


store/postgres/src/deployment_store.rs
```diff
@@ -357,52 +357,111 @@ impl DeploymentStore {
...
+        entity_type: &EntityType,
+        data: &mut Vec<(EntityKey, Entity)>,
+        conn: &PgConnection,
+        layout: &Layout,
+        ptr: &EthereumBlockPointer,
+        stopwatch: &StopwatchMetrics,
+    ) -> Result<i32, StoreError> {
+        let section = stopwatch.start_section("check_interface_entity_uniqueness");
+        for (key, _) in data.iter() {
+            // WARNING: This will potentially execute 2 queries for each entity key.
```
lutter, line 415: Good point - we should figure out how to batch this check, too, but that can wait for another day. Maybe file an issue for just that?

store/postgres/src/deployment_store.rs
```diff
@@ -357,52 +357,111 @@ impl DeploymentStore {
...
+            self.check_interface_entity_uniqueness(conn, layout, key)?;
+        }
+        section.end();
+
+        let _section = stopwatch.start_section("apply_entity_modifications_insert");
+        layout
+            .insert(conn, entity_type, data, block_number(ptr))
+            .map(|_| 1)
```
lutter, line 423: This needs to return the number of entities that were inserted, not 1

store/postgres/src/deployment_store.rs
```diff
@@ -357,52 +357,111 @@ impl DeploymentStore {
...
+        conn: &PgConnection,
+        layout: &Layout,
+        ptr: &EthereumBlockPointer,
+        stopwatch: &StopwatchMetrics,
+    ) -> Result<i32, StoreError> {
+        let _section = stopwatch.start_section("apply_entity_modifications_delete");
+        layout
+            .delete(conn, entity_type, &entity_keys, block_number(ptr))
+            // This conversion is ok since n will only be 0 or 1
+            .map(|n| -(n as i32))
```
lutter, line 461: Similarly, this needs to return the negative of the number of entities deleted to make sure the overall entity count is correct.

store/postgres/src/relational.rs
```diff
@@ -546,13 +546,12 @@ impl Layout {
...
-        key: &EntityKey,
-        entity: Entity,
+        entity_type: &EntityType,
+        entities: &mut Vec<(EntityKey, Entity)>,
         block: BlockNumber,
     ) -> Result<(), StoreError> {
-        let table = self.table_for_entity(&key.entity_type)?;
-        let query = InsertQuery::new(table, key, entity, block)?;
-        query.execute(conn)?;
+        let table = self.table_for_entity(&entity_type)?;
```
lutter, line 553: Small nit: the `&` is not necessary

store/postgres/src/relational_queries.rs
```diff
@@ -1230,50 +1230,52 @@ impl<'a> LoadQuery<PgConnection, EntityData> for FindManyQuery<'a> {
...
+    entities: &'a mut Vec<(EntityKey, Entity)>,
     block: BlockNumber,
 }
 
 impl<'a> InsertQuery<'a> {
     pub fn new(
         table: &'a Table,
-        key: &'a EntityKey,
-        entity: Entity,
+        _entity_type: &'a EntityType,
```
lutter, line 1243: If we are not using this argument, just get rid of it (it's available as `table.object` anyway)

store/postgres/src/relational_queries.rs
```diff
@@ -1291,24 +1293,50 @@ impl<'a> QueryFragment<Pg> for InsertQuery<'a> {
...
 
         out.push_sql("(");
-        for column in self.table.columns.iter() {
-            if self.entity.contains_key(&column.field) {
-                out.push_identifier(column.name.as_str())?;
-                out.push_sql(", ");
+
+        // Build the column name list using the subset of all keys among our entities.
+        let unique_column_names: Vec<&Column> = {
+            let mut btreemap = BTreeMap::new();
```
lutter, line 1299: This should just be a set; I even think that a `HashSet` is good enough here since we don't care about the order of columns.

store/postgres/src/relational_queries.rs
```diff
@@ -1291,24 +1293,50 @@ impl<'a> QueryFragment<Pg> for InsertQuery<'a> {
...
+            for (_key, entity) in self.entities.iter() {
+                for column in &self.table.columns {
+                    if entity.get(&column.field).is_some() {
+                        btreemap.entry(column.name.as_str()).or_insert(column);
+                    }
+                }
```
lutter, line 1305: The computation of `unique_columns` should go into `new` since `walk_ast` can be called multiple times by Diesel, and we'd redundantly repeat the calculation.

store/postgres/src/relational_queries.rs
```diff
@@ -1291,24 +1293,50 @@ impl<'a> QueryFragment<Pg> for InsertQuery<'a> {
...
         out.push_identifier(BLOCK_RANGE_COLUMN)?;
 
-        out.push_sql(")\nvalues(");
-        for column in self.table.columns.iter() {
-            if let Some(value) = self.entity.get(&column.field) {
-                QueryValue(value, &column.column_type).walk_ast(out.reborrow())?;
+        out.push_sql(") values\n");
+
+        // Use a `Peekable` iterator to help us decide how to finalize each line.
+        let mut iter = self.entities.iter().map(|(_key, entity)| entity).peekable();
```
lutter, line 1318: Nice! I used to do this with an explicit `first` boolean flag, but a peekable iterator seems nicer.

store/postgres/src/relational_queries.rs
```diff
@@ -1291,24 +1293,50 @@ impl<'a> QueryFragment<Pg> for InsertQuery<'a> {
...
+        while let Some(entity) = iter.next() {
```
lutter, line 1319: Couldn't that just be `for entity in iter {` ?

store/postgres/src/deployment_store.rs
```diff
@@ -357,52 +357,111 @@ impl DeploymentStore {
...
-                            e
-                        )
-                        .into()
-                    }),
-            }?;
-            count += n;
+            }
+        }
+        // apply modification groups
+        for (entity_type, entities) in inserts.iter_mut() {
```
lutter, line 390: I haven't tried it, but would some of the complications around `Vec<&EntityKey>` vs `&Vec<EntityKey>` in `Layout.update` and `Layout.delete` etc. go away if you used `inserts.drain()` here instead of `iter_mut()`? If it does help, maybe we can get avoid building an auxiliary `Vec` in those methods.

store/postgres/tests/relational.rs
```diff
@@ -187,8 +187,13 @@ fn insert_entity(conn: &PgConnection, layout: &Layout, entity_type: &str, entity
...
     );
-    let errmsg = format!("Failed to insert entity {}[{}]", entity_type, key.entity_id);
-    layout.insert(&conn, &key, entity, 0).expect(&errmsg);
+    let entity_id = key.entity_id.clone();
+    let entity_type = EntityType::from(entity_type);
+    let mut entities = vec![(key, entity)];
+    let errmsg = format!("Failed to insert entity {}[{}]", entity_type, entity_id);
+    layout
+        .insert(&conn, &entity_type, &mut entities, 0)
+        .expect(&errmsg);
```
lutter, line 196: It would be good to also test that inserting 2 or 3 entities work, just to have a test that our SQL generation doesn't break in the future.

store/postgres/tests/relational.rs
```diff
@@ -197,8 +202,13 @@ fn update_entity(conn: &PgConnection, layout: &Layout, entity_type: &str, entity
...
     );
-    let errmsg = format!("Failed to update entity {}[{}]", entity_type, key.entity_id);
-    layout.update(&conn, &key, entity, 1).expect(&errmsg);
+    let entity_id = key.entity_id.clone();
+    let entity_type = EntityType::from(entity_type);
+    let mut entities = vec![(key, entity)];
+    let errmsg = format!("Failed to insert entity {}[{}]", entity_type, entity_id);
+    layout
+        .update(&conn, &entity_type, &mut entities, 0)
+        .expect(&errmsg);
```
lutter, line 211: Similarly here - it would also be good to test that with more than one entity. Since creating entities is annoying from Rust code, there's an `entity!` macro that can ease notation a bit.

store/postgres/tests/relational.rs
```diff
@@ -546,18 +562,28 @@ fn delete() {
...
             THINGS_SUBGRAPH_ID.clone(),
             "Scalar".to_owned(),
             "no such entity".to_owned(),
         );
-        let count = layout.delete(&conn, &key, 1).expect("Failed to delete");
+        let entity_type = EntityType::from("Scalar");
+        let mut entity_keys = vec![key];
+        let count = layout
+            .delete(&conn, entity_type.clone(), &entity_keys, 1)
+            .expect("Failed to delete");
```
lutter, line 574: We should also have a test for deletion of more than one entity - this will also help test that we update the entity count correctly in that case. Getting that wrong is very hard to detect, so we need to have good tests for that.

lutter (changes requested):
Very nice! This looks great! Can't wait to try this out in production.

store/postgres/src/relational_queries.rs
tilacog: I ran into some weird ordering errors when using HashSet, which I got rid of when I sorted the resulting vector, so I ended up using a BTreeMap.

The errors were caught by (flaky) results on the `conflicting_implementors_id`. I'm not sure why, though, but I can dig into that.

store/postgres/src/relational_queries.rs
tilacog: Unfortunately, no, because the iterator gets borrowed when using for loops, and the call to `peek()` also requires a mutable borrow.

store/postgres/tests/relational.rs
tilacog: Additionally, I think we could use an `assert` right into this helper function, so we are also testing that the expected number of entities were inserted, updated or deleted (according to the specific function).

store/postgres/src/relational_queries.rs
tilacog: Moving the unique column scanning outside of the `walk_ast` did the trick, but I still have to use `HashMap<&str, &Column>` since the `Column` type does not implement the `Hash` or `Eq` traits.

store/postgres/src/deployment_store.rs
tilacog: Let's change the ClampQuery signature so it accepts the same type of vector as the InsertQuery.

tilacog: On second thought, this might not be possible since the `Remove` variant does not have an `Entity` value, but only an `EntityKey`

tilacog: I couldn't figure out how to avoid creating the vector you've mentioned, but I was able to avoid creating a new vector inside `ClampRangeQuery` implementation of `walk_ast` on commit 59f1606c28d72cedc4f6e3b6fa8fecb151d7ba40 by generalizing the `ForeignKeyClauses::is_in` function to accept a vector of either `&str` or `String` So we can use that very same vector.

store/postgres/src/relational_queries.rs
lutter: Aah .. maybe that's why I usually use the explicit `first` flag. Not a big issue though

store/postgres/src/deployment_store.rs
lutter: Nice! Yes, it looks like we have to allocate the extra vector no matter what.

store/postgres/src/relational.rs
```diff
@@ -650,7 +650,9 @@ impl Layout {
         let table = self.table_for_entity(&entity_type)?;
         let entity_keys: Vec<&EntityKey> = entities.iter().map(|(key, _)| key).collect();
         ClampRangeQuery::new(table, &entity_type, &entity_keys, block).execute(conn)?;
-        Ok(InsertQuery::new(table, entity_type, entities, block)?.execute(conn)?)
+        Ok(InsertQuery::new(table, entity_type, entities, block)?
+            .get_results(conn)
+            .map(|ids| ids.len())?)
```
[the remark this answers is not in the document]
tilacog: You're absolutely right.

At first, I thought that `execute` was returning 0 because of our custom queries, but after close inspection, I noticed that that zero was coming from a mistake I'd made when obtaining the length of a drained vector... 

I'll revert those changes and remove the `LoadQuery` trait implementation for `ClampRangeQuery`.

lutter (approved):

tilacog:
I'll squash those extra commits and rebase.
Thanks! :+1:

PR #2293, opened 2021-03-19
Copy subgraph data in batches when grafting

lutter:
These changes make it so that the data copy for grafting is done in batches to avoid long-running transactions. It is now also permissible to graft subgraphs across shards.

Note that this PR sits on top of #2268 since it relies on its changes


store/postgres/src/relational_queries.rs
```diff
@@ -2818,14 +2832,26 @@ impl<'a> QueryFragment<Pg> for CopyEntityDataQuery<'a> {
         }
         out.push_sql("block_range from ");
         out.push_sql(self.src.qualified_name.as_str());
+        out.push_sql(" where vid >= ");
+        out.push_bind_param::<BigInt, _>(&self.first_vid)?;
+        out.push_sql(" and vid < ");
+        out.push_bind_param::<BigInt, _>(&self.last_vid)?;
```
leoyvens, line 2838: `last_vid` isn't really the last vid but the one more than the last, I think this would be more readable if this where `<= last_vid` and `last_vid` was passed in as the actual last vid.

store/postgres/src/copy.rs
```diff
@@ -0,0 +1,540 @@
...
+
+        let target_vid = sql_query(&format!(
+            "select coalesce(max(vid), 0) as max_vid from {} where lower(block_range) <= $1",
+            src.qualified_name.as_str()
+        ))
+        .bind::<Integer, _>(&target_block.number)
+        .load::<MaxVid>(conn)?
+        .first()
+        .map(|v| v.max_vid)
+        .unwrap_or(0);
```
leoyvens, line 266: If the table is empty, I believe the `target_vid` should be `-1`.

store/postgres/src/copy.rs
```diff
@@ -0,0 +1,540 @@
...
+            src,
+            dst,
+            next_vid: 0,
+            target_vid,
+            batch_size: INITIAL_BATCH_SIZE,
+        })
+    }
+
+    fn finished(&self) -> bool {
+        self.next_vid >= self.target_vid
```
leoyvens, line 279: My understanding is that `next_vid` has not yet been copied, and we want to copy up to and including `target_vid`, so we're only finished after `next_vid > target_vid`.

store/postgres/src/copy.rs
```diff
@@ -0,0 +1,540 @@
...
+    }
+
+    fn copy_batch(&mut self, conn: &PgConnection) -> Result<(), StoreError> {
+        let start = Instant::now();
+
+        rq::CopyEntityBatchQuery::new(
+            self.dst.as_ref(),
+            &self.src,
+            self.next_vid,
+            self.next_vid + self.batch_size,
```
leoyvens, line 386: We're not limiting by `target_vid` here, I think that'd be good to make the behavior more predictable.

store/postgres/src/deployment_store.rs
```diff
@@ -1025,32 +1038,57 @@ impl DeploymentStore {
...
-                    &base.deployment,
-                    block.clone(),
-                )?;
+                let block_to_revert: BlockNumber = (block.number + 1)
+                    .try_into()
+                    .expect("block numbers fit into an i32");
+                dst.revert_block(&conn, &dst.site.deployment, block_to_revert)?;
+                Layout::revert_metadata(&conn, &dst.site.deployment, block_to_revert)?;
+                info!(logger, "Rewound subgraph to block {}", block.number;
+                      "time_ms" => start.elapsed().as_millis());
```
leoyvens, line 1081: It'd be good to see more justification for this step in the comment.

store/postgres/src/subgraph_store.rs
```diff
@@ -772,12 +776,13 @@ impl SubgraphStoreTrait for SubgraphStore {
...
-                let site = self.primary_conn()?.find_existing_site(&base_id)?;
-                Some((Arc::new(site), base_ptr))
+                let src = self.layout(&base_id)?;
+                Some((src, base_ptr))
             }
             None => None,
         };
-        store.start_subgraph(logger, site, graft_base)
+        store.start_subgraph(logger, site.clone(), graft_base)?;
+        self.primary_conn()?.copy_finished(site.as_ref())
```
leoyvens, line 785: It reads a bit weird to execute this unconditionally even though most subgraphs won't be copied when deployed, maybe this `copy_finished` call can be moved into `start_subgraph`?

leoyvens (changes requested):
This foreign schema thing is some dark sorcery!

store/postgres/migrations/2021-03-12-070453_add_copy_state/up.sql
```diff
@@ -0,0 +1,32 @@
...
+   dst  int primary key references deployment_schemas(id) on delete cascade,
+   unique(src, dst)
+);
+
+-- These two tables live in each shard; entries are in the same
+-- shard as the subgraph_deployment we are copying to ('dst')
+create table subgraphs.copy_state(
+   src                 int not null,
+   dst                 int primary key
+                           unique references subgraphs.subgraph_deployment(id)
```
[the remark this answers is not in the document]
lutter: Totally, must have happened in one of the iterations of working on this.

store/postgres/src/catalog.rs
```diff
@@ -103,3 +124,33 @@ pub fn current_servers(conn: &PgConnection) -> Result<Vec<String>, StoreError> {
...
 }
+
+pub fn has_namespace(conn: &PgConnection, namespace: &Namespace) -> Result<bool, StoreError> {
+    use pg_namespace as nsp;
+
+    Ok(nsp::table
+        .filter(nsp::nspname.eq(namespace.as_str()))
+        .count()
+        .get_result::<i64>(conn)?
+        > 0)
```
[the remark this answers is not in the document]
lutter: Neat! Changed.

store/postgres/src/catalog.rs
```diff
@@ -103,3 +124,33 @@ pub fn current_servers(conn: &PgConnection) -> Result<Vec<String>, StoreError> {
...
+/// schema, do nothing. This crucially depends on the fact that we never mix
+/// foreign and local tables in the same schema.
+pub fn drop_foreign_schema(conn: &PgConnection, src: &Site) -> Result<(), StoreError> {
+    use foreign_tables as ft;
+
+    let is_foreign = ft::table
+        .filter(ft::foreign_table_schema.eq(src.namespace.as_str()))
+        .count()
+        .get_result::<i64>(conn)?
+        > 0;
```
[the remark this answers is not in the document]
lutter: Changed too

store/postgres/src/copy.rs
```diff
@@ -0,0 +1,540 @@
...
+            .execute(conn)?;
+
+        // If we imported the schema for `src`, and no other in-progress
+        // copy is using it, get rid of it again
+        if self.crosses_shards() {
+            let active_copies = cs::table
+                .filter(cs::src.eq(self.src.site.id))
+                .filter(cs::finished_at.is_null())
+                .count()
+                .get_result::<i64>(conn)?;
```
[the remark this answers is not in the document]
lutter: Changed

store/postgres/src/relational_queries.rs
lutter: True; I changed that.

store/postgres/src/copy.rs
lutter: Tru; that saves an unnecessary attempt to copy `vid = 0`

store/postgres/src/copy.rs
lutter: Yes, I made changes to that logic yesterday - I went from `current_vid` to `next_vid` because I thought it would be clearer, but didn't catch all the places that affected. Changed

store/postgres/src/copy.rs
lutter: Yes, that's actually a pretty serious bug since we would end up copying data we should not be copying.

store/postgres/src/deployment_store.rs
lutter: I've actually been meaning to get rid of this and simply forgot. It was left over from when we copied everything and then got rid of the excess because copying the complex metadata structure up to just some block was too cumbersome. I've removed that in favor of copying just what we need.

store/postgres/src/subgraph_store.rs
lutter: Unfortunately, this can't go into `start_subgraph` since it manipulates a different database.

store/postgres/src/deployment_store.rs
lutter: As it turns out, reverting the data is important since it needs to unclamp versions that are deleted in the future. In theory, `CopyEntityBatchQuery` could do that during copying, but the SQL to do that becomes pretty unwieldy. Maybe as an optimization in the future.

lutter:
Added more commits to address all the review comments.

store/postgres/src/deployment_store.rs
leoyvens: Ooh tricky

store/postgres/src/deployment_store.rs
```diff
@@ -1063,20 +1063,21 @@ impl DeploymentStore {
...
+                // clamped in the future (beyond `block`) become valid for
+                // all blocks after `block`. `revert_block` gets rid of
+                // everything including the block passed to it. We want to
+                // preserve `block` and therefore revert `block+1`
                 let start = Instant::now();
                 let block_to_revert: BlockNumber = (block.number + 1)
                     .try_into()
                     .expect("block numbers fit into an i32");
                 dst.revert_block(&conn, &dst.site.deployment, block_to_revert)?;
-                Layout::revert_metadata(&conn, &dst.site.deployment, block_to_revert)?;
```
leoyvens: Note to self: Remember this interaction if we ever add support for removing data sources.

leoyvens (approved):

lutter:
I made one very small change to this PR: moving `active_copies` to the public database schema, and adding a `cancelled_at` column so that ongoing copy processes can be signalled to stop copying.

lutter:
Fixed a mistake in the `copy_table_state` table where `next_vid` and `target_vid` were `int` instead of `int8`

Issue #2325, opened 2021-03-31
Subgraph built on Mac M1 chip fails: 'registering new sigaltstack failed'


azf20:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?*
Indexing a subgraph built on a [Mac M1 Chip](https://www.apple.com/uk/mac/m1/) fails after deployment with the following error:
```
graph-node_1  | Mar 30 17:44:30.544 INFO Syncing 1 blocks from Ethereum., code: BlockIngestionStatus, blocks_needed: 1, blocks_behind: 1, latest_block_head: 2, current_block_head: 1, provider: localhost-rpc-0, component: BlockIngestor
graph-node_1  | Mar 30 17:44:40.634 INFO Syncing 1 blocks from Ethereum., code: BlockIngestionStatus, blocks_needed: 1, blocks_behind: 1, latest_block_head: 3, current_block_head: 2, provider: localhost-rpc-0, component: BlockIngestor
graph-node_1  | Mar 30 17:44:41.150 INFO 1 trigger found in this block for this subgraph, block_hash: 0x591ae02eb53ff160e33abe0cc1fcbf0813c380e897523f3ab1aed294cff56159, block_number: 3, subgraph_id: QmRvEgBm8fknmKVcBC9SJECFg1hJZFSsE5SzJ3VSuji8hQ, component: SubgraphInstanceManager
graph-node_1  | thread 'mapping-QmRvEgBm8fknmKVcBC9SJECFg1hJZFSsE5SzJ3VSuji8hQ-635e0c5d-65f3-4468-abf8-5bcfdfff157d' panicked at 'assertion failed: `(left == right)`
graph-node_1  |   left: `-1`,
graph-node_1  |  right: `0`: registering new sigaltstack failed', /usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/wasmtime-runtime-0.21.0/src/traphandlers.rs:774:9
graph-node_1  | note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
[...]

PR #2326, opened 2021-03-31
Update wasmtime

leoyvens:
Update wasmtime to 0.25. The API to call wasm functions changed. There is now a safe API to access memory, which allows us to remove our only uses of `unsafe`.
 
Might help with #2325.


That3Percent (approved):
LGTM

PR #2327, opened 2021-03-31
note about supported ipfs version

mktcode:
Would have saved me a lot of time.


leoyvens (approved):
Thanks, hopefully we'll soon get rid of this limitation!

mktcode:
That was faster than expected! :rocket:

PR #2328, opened 2021-04-01
Allow network to opt out of eip-1898

leoyvens:
This is necessary support fuse until their OpenEthereum version is updated.

I removed the previous global escape hatch `GRAPH_ETH_CALL_BY_NUMBER` since with OpenEthereum 3.2 released, indexers should now all have access to an EIP-1898 compatible node.

### DIFF
--- a/node/src/main.rs
+++ b/node/src/main.rs
@@ -56,18 +56,6 @@ lazy_static! {
         .map(|s| BlockNumber::from_str(&s)
              .unwrap_or_else(|_| panic!("failed to parse env var ETHEREUM_ANCESTOR_COUNT")))
         .unwrap_or(50);
-
-    // Comma separated network names that do not support EIP-1898 (calls by block hash). The only
-    // network known to not support it is fuse, once all known networks support eip 1898 this env
-    // var can be removed.
-    static ref NO_EIP_1898_SUPPORT: Vec<String> = {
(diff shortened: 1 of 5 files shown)


node/src/main.rs
```diff
@@ -56,6 +56,18 @@ lazy_static! {
         .map(|s| BlockNumber::from_str(&s)
              .unwrap_or_else(|_| panic!("failed to parse env var ETHEREUM_ANCESTOR_COUNT")))
         .unwrap_or(50);
+
+    // Comma separated network names that do not support EIP-1898 (calls by block hash). The only
+    // network known to not support it is fuse, once all known networks support eip 1898 this env
+    // var can be removed.
+    static ref NO_EIP_1898_SUPPORT: Vec<String> = {
```
lutter, line 63: As a rule, I don't want to add more environment variables to configure things; can we make that something that you add to the `features` in `graph-node.toml`? It's a little funky that it's a feature that is one by default, but I would much rather have a feature `no-eip1898` than another environment variable.

leoyvens:
@lutter I've added this as a provider feature, and did some refactorings around `NodeCapabilities` as well, even though it turns out this doesn't need to be added there.

lutter (approved):
Nice. Also, good refactor

PR #2338, opened 2021-04-04
Change log level to `debug`

leoyvens:
This isn't really `info!` worthy.


lutter (approved):
Does it even need its own log? We log the range size later on when we use it, too, don't we?

leoyvens:
@lutter fair point, I've removed the log

PR #2344, opened 2021-04-05
Remove some unwraps in the block stream

leoyvens:
We've seen these panic in production. Also this is why nobody misses futures 0.1.


lutter (approved):
Nice! Even got rid of a gratuitous unwrap

PR #2346, opened 2021-04-05
Drop support for old Geth revert format, detect stack limit

leoyvens:
The old format is not returned by Geth since the Berlin fork. A few subgraphs have hit the `stack limit reached` error, which is deterministic.


lutter (approved):

shahafan:
Still getting this, do I have to redeploy the subgraph or wait for the indexer to update their side?

PR #2352, opened 2021-04-06
Avoid over-fetching queries

tilacog:
This PR adds logic to replace `SELECT *` SQL statements to use specific column names as declared in GraphQL queries.

A new enum, `ColumnNames`, is used to transport either a vector of column names (as strings) or a variant to select all columns in cases where we want to keep previous behaviour.


leoyvens (commented):
Not a full review, but I have some comments

graph/src/components/store.rs
```diff
@@ -1716,3 +1743,47 @@ impl LfuCache<EntityKey, Option<Entity>> {
         }
     }
 }
+
+/// Determines which columns should be selected in a table.
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub enum ColumnNames {
```
lutter, line 1749: I know this is going to be an icky change, but this should really be called `AttributeNames` or `Attributes` - we are operating at the GraphQL level here, which deals in (entity) types and attributes; it's only the relational store level that operates at the level of tables and columns.

graph/src/components/store.rs
```diff
@@ -1716,3 +1743,47 @@ impl LfuCache<EntityKey, Option<Entity>> {
...
+        self.insert(&field.name)
+    }
+
+    pub fn extend(&mut self, other: Self) {
+        use ColumnNames::*;
+        match (self, other) {
+            (All, All) => {}
+            (self_ @ All, other @ Select(_)) => *self_ = other,
+            (Select(_), All) => {
+                unreachable!()
```
lutter, line 1784: Is it safe to leave this here? And even then, shouldn't this be `*self = All`?

graphql/src/store/prefetch.rs
```diff
@@ -825,14 +875,119 @@ fn fetch(
...
     }
-
     store
         .find_query_values(query)
         .map(|entities| entities.into_iter().map(|entity| entity.into()).collect())
 }
+
+/// Represents a finished column collection operation, mapping each object type to the final set of
+/// selected SQL columns.
+type ColumnNamesByObjectType<'a> = BTreeMap<ObjectCondition<'a>, ColumnNames>;
```
lutter, line 891: Similarly, this should be `AttributesByObjectType`

graphql/src/store/prefetch.rs
```diff
@@ -825,14 +875,119 @@ fn fetch(
...
+
+#[derive(Debug, Default, Clone)]
+struct CollectedColumnNames<'a>(HashMap<ObjectOrInterface<'a>, ColumnNames>);
```
lutter, line 894: .. and `CollectedAttributes`

lutter (commented):
Nice progress, really looking forward to having this running in the wild.

store/postgres/src/deployment_store.rs
```diff
@@ -733,7 +733,11 @@ impl DeploymentStore {
                         let query = EntityQuery::new(
                             site4.deployment.clone(),
                             block.number.try_into().unwrap(),
-                            EntityCollection::All(vec![POI_OBJECT.cheap_clone()]),
+                            EntityCollection::All(vec![(
+                                POI_OBJECT.cheap_clone(),
+                                // TODO: discover how we should handle the ColumnNames in this case
```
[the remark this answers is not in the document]
tilacog: Yes. I'm not 100% sure if we can specify what columns were interested in or if it is safe to use `Columns::All` in this specific case. If the selected columns never vary,  we could hardcode the `Columns::Select` value here.

graph/src/components/store.rs
tilacog: I agree. Leo had the same impression, too.

graphql/src/store/prefetch.rs
```diff
@@ -822,17 +833,171 @@ fn fetch(
...
+            }
+            Entry::Vacant(entry) => {
+                entry.insert(column_names);
+            }
+        }
+    }
+
+    /// Creates a new, combined `CollectedColumnNames` using drained values from `CollectedColumnNames`
+    /// scattered across different `CollectedResponseKey`s.
+    fn consolidate_column_names<'schema, 'collection>(
```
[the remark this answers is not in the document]
leoyvens: I now understand this.

graphql/src/store/prefetch.rs
```diff
@@ -825,14 +875,119 @@ fn fetch(
...
+fn filter_derived_fields(column_names_type: ColumnNames, object: &s::ObjectType) -> ColumnNames {
+    match column_names_type {
+        ColumnNames::All => column_names_type,
+        ColumnNames::Select(sql_column_names) => {
+            let mut filtered = ColumnNames::All;
+            sql_column_names
+                .into_iter()
+                .filter_map(|column_name| {
+                    if let Some(schema_field) = sast::get_field(object, &column_name) {
+                        if schema_field.find_directive("derivedFrom").is_none() {
```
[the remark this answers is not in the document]
tilacog: Totally agree. 
I created a new issue for us to track that. #2464

graph/src/components/store.rs
tilacog: Agree.

It may sound counterintuitive, but since `self` is more specific, I believe it should keep its original value in that case.
However, this might lead to "under-fetching," which undesirable. 

In conclusion, I think that using `All` in this case is a safe assumption.

tilacog: I agree, this `unreachable!()` should go out.

As the `All` variant encompasses all the table's columns, this might be counterintuitive, but I believe `self` should keep the original value in this case since it is more specific.

On the other hand, that might lead to unforeseen "under-fetching" scenarios, even though the `All` variant is currently used only in cases where we are uncertain of which columns to select and all of our tests, for that matter, are passing. 

I don't mean to digress, but we could consider the creation of a new variant to represent that state:
```enum ColumnNames {
    All,
   Select(BTreeSet<String>),
   Unspecified,
}```

Comment (leoyvens) [reply graph/src/components/store.rs:1784]: In theory I agree that `*self = All` seems correct. But since this is never used anywhere I'd keep the `unreachable!()`, though I'm fine with either.

@tilacog I see what you mean that `All` is a bit ambiguous, bu that's fine imo.

**Comment by leoyvens** [comment]
@tilacog Lets add an env var to revert everything back to `All`, in case we uncover bugs when this is deployed to the hosted service and need a revert switch.

File: store/postgres/src/relational_queries.rs [hunk 8]
```diff
@@ -78,7 +77,7 @@ lazy_static! {
 
     /// Those are columns that we always want to fetch from the database.
     static ref BASE_SQL_COLUMNS: BTreeSet<String> =
-        ["id", "vid"].iter().map(ToString::to_string).collect();
+        ["id"].iter().map(ToString::to_string).collect();
 }
```
tilacog: Had to remove `vid` because it doesn't exist in `Table.columns`.

store/postgres/src/relational_queries.rs
```diff
@@ -2881,17 +2882,22 @@ pub struct CopyVid {
...
+                        .expect("failed to find column for field")
+                        .name
+                })
                 .peekable();
             while let Some(column_name) = iterator.next() {
-                out.push_sql(&column_name);
+                out.push_sql(&column_name.as_str());
                 if iterator.peek().is_some() {
                     out.push_sql(", ");
                 }
```
tilacog: Obtaining SQL column name directly from `Table` now.

store/postgres/src/deployment_store.rs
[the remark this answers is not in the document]
lutter: I don't see a problem using `ColumnNames::All` here; the `poi2$` table only has two 'real' columns. Or am I missing something?

graph/src/components/store.rs
lutter: My concern with this is that it's very unintuitive that the operation is not symmetric, i.e. that `a.extend(b)` and `b.extend(a)` do not yield the same result. It makes sense for how this is currently used, but if a future code change changes the order in which entries are traversed, it could lead to incorrect query execution (I _think_ the GraphQL layer would blow up if we do not fetch all required columns) But I now think that if we leave the `unreachable!` there, we should be safe since things will blow up loudly if we change the order in which attribute names are traversed and collected.

lutter (changes requested):

graph/src/components/store.rs
```diff
@@ -1780,9 +1780,7 @@ impl AttributeNames {
         match (self, other) {
             (All, All) => {}
             (self_ @ All, other @ Select(_)) => *self_ = other,
-            (Select(_), All) => {
-                unreachable!()
-            }
+            (Select(_), All) => {} // no-op, since we want to keep selected columns
```
lutter, line 1783: I thought having the `unreachable` there was better, since we do think this is unreachable, and it's better to have this fail loudly if we ever change the code by mistake so that this becomes reachable.

lutter (changes requested):

lutter (approved):

PR #2354, opened 2021-04-07
Fix hanging when starting over 512 subgraphs

leoyvens:
This issue was introduced when upgrading to tokio 1.0, the `#[tokio::main]` syntax changed and the number of blocking threads was inadvertently lowered from 2000 to the default 512. I was able to reproduce the issue locally. This fixes the root cause which is the incorrect use of the blocking thread pool. At least in the critical locations, we still use it incorrectly in others, #905 tracks getting this right in all places.

There are at least two ways we where using the blocking thread pool incorrectly:
- The code path to start a subgraph had a few nested calls to `spawn_blocking`. This leads to a "double dip" deadlock, the nested calls are waiting on the callers to release a blocking threads, but the callers are of course waiting on the nested call to complete.
- We were using `spawn_blocking` for the task that calls `run_subgraph`. Tasks that run indeterminately should not be put on the blocking thread pool, because that could exhaust the available threads. Ideally we'd make it so that `run_subgraph` doesn't do blocking calls, but the quick fix is to use an OS thread. I added `graph::spawn_thread` to make this convenient.

### DIFF
--- a/core/src/subgraph/instance_manager.rs
+++ b/core/src/subgraph/instance_manager.rs
@@ -206,8 +206,7 @@ where
     ) {
         let logger = self.logger_factory.subgraph_logger(&id);
 
-        // Blocking due to store interactions. Won't be blocking after #905.
-        match graph::spawn_blocking(Self::start_subgraph_inner(
+        match Self::start_subgraph_inner(
             logger.clone(),
             self.instances.clone(),
             self.host_builder.clone(),
@@ -219,10 +218,8 @@ where
             manifest,
             self.metrics_registry.cheap_clone(),
             self.link_resolver.cheap_clone(),
-        ))
+        )
         .await
-        .map_err(Error::from)
-        .and_then(|e| e)
         {
             Ok(()) => self.manager_metrics.subgraph_count.inc(),
             Err(err) => error!(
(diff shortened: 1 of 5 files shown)


core/src/subgraph/instance_manager.rs
```diff
@@ -355,7 +352,21 @@ where
                 &network,
                 &required_capabilities, e))?.clone();
 
-        store.start_subgraph_deployment(&logger, &manifest.id)?;
+        {
```
lutter, line 355: I think this is ok. In [this commit](https://github.com/graphprotocol/graph-node/pull/2313/commits/55642e78d82c124703c1109fa8bacc774f34b284) I had to move `start_subgraph_deployment` to the same thread that does `run_subgraph` since, with copying, `start_subgraph_deployment` can take hours and would otherwise block the instance manager. (There's nothing to do for this PR, just an FYI)

lutter (approved):

core/src/subgraph/instance_manager.rs
leoyvens: Good thing it is in a blocking task then! Ideally it would use something like `with_conn` but it didn't seem so straightforward to refactor it, so that's left for another day.

PR #2355, opened 2021-04-07
Do not retry `ipfs.cat` indefinitely

leoyvens:
I believe #2324 accidentally applied infinite retries to the link resolver used by the runtime, this was meant to apply only to the assignment provider.


lutter (approved):
Yep, I had a bad feeling that my fix was a little too easy

PR #2364, opened 2021-04-09
node: Have `graphman` override configured pool sizes

lutter:
In general, we don't want graphman to use the configured pool size to
reduce the risk of running out of database handles. Instead, `graphman`
uses a pool size of 3 by default for everything which should be plenty for
its purposes

I also tacked a small logging change that should make it easier to understand when we skip creating a block ingestor onto the PR


lutter:
By now, this PR should have probably been two. It now also avoids a panic during startup if we fail to connect to one of the ethereum clients; instead, it just ignores such clients.

leoyvens (approved):

lutter:
Rebased to latest master

PR #2365, opened 2021-04-10
chain head listener: Use separate watchers per chain

leoyvens:
As I've been able to reproduce locally, under a sufficent load of concurrent subgraphs and with many fast networks connected, there is a high probability of the mainnet chain head update being overwritten with that of another network before all subgraphs can read the update, to the point where the subgraphs can seem deadlocked. So this may very well be the cause for the subgraph deadlocks observed in prod.

The important change is in `chain_head_listener.rs`, which now demultiplexes the DB notifications into one watcher per network. This also does a bit of modernizing in `store_events.rs` and `notification_listener.rs` to use more of tokio 1.0 and futures 0.3.

### DIFF
--- a/store/postgres/src/chain_head_listener.rs
+++ b/store/postgres/src/chain_head_listener.rs
@@ -1,14 +1,15 @@
+use std::collections::BTreeMap;
+
 use diesel::RunQueryDsl;
 use lazy_static::lazy_static;
-use tokio::sync::watch;
-use web3::types::H256;
+use tokio::sync::{mpsc::Receiver, watch, RwLock};
 
 use crate::{
     connection_pool::ConnectionPool,
-    notification_listener::{NotificationListener, SafeChannelName},
+    notification_listener::{JsonNotification, NotificationListener, SafeChannelName},
 };
 use graph::prelude::serde_json::{self, json};
-use graph::prelude::{ChainHeadUpdateListener as ChainHeadUpdateListenerTrait, *};
+use graph::prelude::*;
 use graph::tokio_stream::wrappers::WatchStream;
 use graph_chain_ethereum::BlockIngestorMetrics;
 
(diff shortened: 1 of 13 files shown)


store/postgres/src/chain_head_listener.rs
```diff
@@ -65,56 +77,65 @@ impl ChainHeadUpdateListener {
...
+        graph::spawn(async move {
+            while let Some(notification) = receiver.recv().await {
+                // Create ChainHeadUpdate from JSON
+                let update: ChainHeadUpdate =
+                    match serde_json::from_value(notification.payload.clone()) {
+                        Ok(update) => update,
+                        Err(e) => {
+                            crit!(
+                                logger,
+                                "invalid chain head update received from database";
```
lutter, line 96: Can you change the message to `Ignoring invalid ...` - that makes it clearer in the log that there shouldn't be any additional action required. Not sure if that really is `crit!` - this is probably only going to happen during updates if the notification format changes. A `warn!` might be enough, especially if successful updates go through later on.

store/postgres/src/chain_head_listener.rs
```diff
@@ -65,56 +77,65 @@ impl ChainHeadUpdateListener {
...
-                futures03::future::ready(Some(()))
+    pub async fn subscribe(&self, network_name: String) -> ChainHeadUpdateStream {
+        let update_receiver = {
+            let existing = {
+                let watchers = self.watchers.read().await;
+                watchers.get(&network_name).map(|w| w.receiver.clone())
+            };
+
+            if let Some(watcher) = existing {
+                watcher
```
lutter, line 127: I tripped over this for about 15s: might be better to call this variable `receiver`

lutter (approved):

leoyvens:
@lutter thank you for the review, I've amended the commit with your suggestions.

PR #2366, opened 2021-04-11
chain head listener: Fix race condition

leoyvens:
This is a correction to #2365. There is a race condition in the `else` branch where the watcher could be overwritten, and a subscriber might end up with a receiver that has no sender and would error.

This keeps it simple by using a Mutex, and switching to a sync Mutex since that's what tokio recommends for short critical sections.


otaviopace:
Why wasn't `RwLock` working before? 🤔

otaviopace (approved):

PR #2367, opened 2021-04-11
Chunk application of entity modifications

### LINKED ISSUE #2330: Large inserts fail
With the changes from #2223, we can hit the Postgres limit on the number of bind variables that can be used in one statement. That limit is [65535 parameters](https://stackoverflow.com/a/49379324). We need to change the code that generates the insert/update/delete statements to break such large changes into multiple statements.

tilacog:
This PR chunks the application of entity modifications as an attempt to fix #2330.

Given that the maximum number of PostgreSQL bind parameters per query is 65535, it makes sense to use N chunks where:
    
    chunk size =  65535 / number of fields per entity

Nonetheless, we must remain wary that parameters other than entity fields are also bound in each query, such as block numbers and auxiliary data.

Therefore, it would be reasonable for us to discuss ways to accommodate those extra bindings, probably by reducing `chunk size` by an arbitrary amount.

### DIFF
--- a/store/postgres/src/deployment_store.rs
+++ b/store/postgres/src/deployment_store.rs
@@ -38,6 +38,8 @@ use crate::relational_queries::FromEntityData;
 use crate::{connection_pool::ConnectionPool, detail};
 use crate::{dynds, primary::Site};
 
+const POSTGRES_MAX_BINDINGS_PER_STATEMENT: usize = u16::MAX as usize; // 65535
+
 /// When connected to read replicas, this allows choosing which DB server to use for an operation.
 #[derive(Copy, Clone, Debug, PartialEq, Eq)]
 pub enum ReplicaId {
@@ -311,19 +313,42 @@ impl DeploymentStore {
             }
         }
 
-        // apply modification groups
-        for (entity_type, entities) in inserts.iter_mut() {
-            count +=
-                self.insert_entities(entity_type, entities, conn, layout, ptr, &stopwatch)? as i32
+        // Apply modification groups:
+        // Each operation must respect the maximum number of bindings allowed in PostgreSQL queries,
+        // so we need to act in chunks whose size is defined by the number of entities times the
+        // number of attributes each entity type has.
+
+        // Inserts:
+        for (entity_type, mut entities) in inserts.into_iter() {
+            let chunk_size = {
+                let num_fields = (*entities[0].1).len();
+                POSTGRES_MAX_BINDINGS_PER_STATEMENT / num_fields
+            };
+            for chunk in entities.chunks_mut(chunk_size) {
+                count +=
+                    self.insert_entities(&entity_type, chunk, conn, layout, ptr, &stopwatch)? as i32
+            }
         }
-        for (entity_type, entities) in overwrites.drain() {
-            // not updating count since the number of entities remains the same
-            self.overwrite_entities(entity_type, entities, conn, layout, ptr, &stopwatch)?;
+
+        // Overwrites:
+        for (entity_type, mut entities) in overwrites.into_iter() {
+            let chunk_size = {
+                let num_fields = (*entities[0].1).len();
+                POSTGRES_MAX_BINDINGS_PER_STATEMENT / num_fields
+            };
+            for chunk in entities.chunks_mut(chunk_size) {
+                // we do not update the count since the number of entities remains the same
+                self.overwrite_entities(&entity_type, chunk, conn, layout, ptr, &stopwatch)?;
+            }
         }
-        for (entity_type, entity_keys) in removals.iter() {
-            count -=
-                self.remove_entities(entity_type, entity_keys, conn, layout, ptr, &stopwatch)?
+
+        // Removals
+        for (entity_type, entity_keys) in removals.into_iter() {
+            let chunk_size = POSTGRES_MAX_BINDINGS_PER_STATEMENT / entity_keys.len(); // only ID is bound
+            for chunk in entity_keys.chunks(chunk_size) {
+                count -= self.remove_entities(&entity_type, chunk, conn, layout, ptr, &stopwatch)?
                     as i32;
+            }
         }
         Ok(count)
     }
@@ -331,7 +356,7 @@ impl DeploymentStore {
     fn insert_entities(
         &self,
         entity_type: &EntityType,
-        data: &mut Vec<(EntityKey, Entity)>,
+        data: &mut [(EntityKey, Entity)],
         conn: &PgConnection,
         layout: &Layout,
         ptr: &EthereumBlockPointer,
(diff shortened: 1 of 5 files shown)


store/postgres/src/deployment_store.rs
```diff
@@ -38,6 +38,8 @@ use crate::relational_queries::FromEntityData;
 use crate::{connection_pool::ConnectionPool, detail};
 use crate::{dynds, primary::Site};
 
+const POSTGRES_MAX_BINDINGS_PER_STATEMENT: usize = u16::MAX as usize; // 65535
```
lutter, line 41: It might be better to call this `POSTGRES_MAX_PARAMETERS` since that's what the Postgres error message complains about.

store/postgres/src/deployment_store.rs
```diff
@@ -311,27 +313,50 @@ impl DeploymentStore {
...
+        // Apply modification groups:
+        // Each operation must respect the maximum number of bindings allowed in PostgreSQL queries,
+        // so we need to act in chunks whose size is defined by the number of entities times the
+        // number of attributes each entity type has.
+
+        // Inserts:
+        for (entity_type, mut entities) in inserts.into_iter() {
+            let chunk_size = {
+                let num_fields = (*entities[0].1).len();
+                POSTGRES_MAX_BINDINGS_PER_STATEMENT / num_fields
```
lutter, line 325: If you did the chunking inside `Layout.insert_entities`, you know how many parameters each entity needs by looking at `Table.columns.len()` That doesn't account for `block_range` and `vid`, so you'd have to add 2 to that number to be on the safe side (though I don't think we ever write `vid`)

lutter (changes requested):

tilacog:
I've moved the batch size logic so it can access the `table` struct.
I hope I got the numbers right:

- `ClampRangeQuery` always uses 2 binds: one for the `block_number` and another for an `entity_id` array, so we shouldn't worry about it – assuming that an array still counts as one bind.

- `InsertQuery` uses `table.columns.len() + block_id` binds. _(I couldn't find any bindings for `vid, indeed)_.

tilacog:
I hope I've got the math right:

- `ClampRangeQuery` always uses 2 binds: one for `block_range`and another for the entities id array. Assuming that an array still counts as one bind, I believe there's nothing to do here.
- `InsertQuery` uses one bind for `block_range` and one bind for every column of it's table, so it got batched.

tilacog:
I've moved the batch logic to the `relational.rs` module, where we can use `table` information.

I hope I got the math right:
- `InsertQuery` uses one bind for `block_range` and one bind for each column in the given table.
- `ClampRangeQuery` always uses 2 binds: one for `block_range` and another for the entity ids array. Assuming that one array uses only one bind, I understand we shouldn't batch it.

store/postgres/src/relational.rs
```diff
@@ -565,15 +567,21 @@ impl Layout {
...
         stopwatch: &StopwatchMetrics,
     ) -> Result<usize, StoreError> {
         let table = self.table_for_entity(entity_type)?;
         let _section = stopwatch.start_section("insert_modification_insert_query");
-        Ok(InsertQuery::new(table, entities, block)?
-            .get_results(conn)
-            .map(|ids| ids.len())?)
+        let mut count = 0;
+        // we add 1 to account for `block_range` bind parameter
+        let chunk_size = (POSTGRES_MAX_PARAMETERS / table.columns.len()) + 1;
```
lutter, line 578: That should be `(POSTGRES_MAX_PARAMETERS / (table.columns.len() + 1))`, right?

store/postgres/src/relational.rs
```diff
@@ -680,18 +688,26 @@ impl Layout {
...
+
         let section = stopwatch.start_section("update_modification_clamp_range_query");
         ClampRangeQuery::new(table, &entity_type, &entity_keys, block).execute(conn)?;
         section.end();
+
         let _section = stopwatch.start_section("update_modification_insert_query");
-        Ok(InsertQuery::new(table, &mut entities, block)?.execute(conn)?)
+        let mut count = 0;
+        // we add 1 to account for `block_range` bind parameter
+        let chunk_size = (POSTGRES_MAX_PARAMETERS / table.columns.len()) + 1;
```
lutter, line 699: Same change in the math here

store/postgres/src/deployment_store.rs
```diff
@@ -311,18 +311,28 @@ impl DeploymentStore {
             }
         }
 
-        // apply modification groups
-        for (entity_type, entities) in inserts.iter_mut() {
+        // Apply modification groups:
+        // Each operation must respect the maximum number of bindings allowed in PostgreSQL queries,
+        // so we need to act in chunks whose size is defined by the number of entities times the
+        // number of attributes each entity type has.
```
lutter, line 317: This comment would now make more sense in the `Layout` methods somewhere

lutter (changes requested):
Yes, I agree on what you say about `ClampRangeQuery` - the one thing to check is if there is some other limit on array size. IIRC, in other contexts it was actually advantageous to break queries with large arrays into smaller ones because you get O(n^2) behavior from scanning these large arrays, so that, for example 10 queries with an array of length 1000 was faster than 1 query with an array of length 10,000

tilacog:
> Yes, I agree on what you say about `ClampRangeQuery` - the one thing to check is if there is some other limit on array size. IIRC, in other contexts it was actually advantageous to break queries with large arrays into smaller ones because you get O(n^2) behavior from scanning these large arrays, so that, for example 10 queries with an array of length 1000 was faster than 1 query with an array of length 10,000

The [docs state that array size limits are ignored](https://www.postgresql.org/docs/current/arrays.html), but Postgres will complain if the [field size exceeds 1GB](https://www.postgresql.org/docs/current/arrays.html). 
From what I have (quickly) researched, I couldn't any info on what that optimal array size for inserting values would be. 

Do you believe this could be better addressed and reviewed in a new issue/PR?
If so, I can create and assign myself to it.

tilacog:
> Yes, I agree on what you say about `ClampRangeQuery` - the one thing to check is if there is some other limit on array size. IIRC, in other contexts it was actually advantageous to break queries with large arrays into smaller ones because you get O(n^2) behavior from scanning these large arrays, so that, for example 10 queries with an array of length 1000 was faster than 1 query with an array of length 10,000

The [docs state that array size limits are ignored](https://www.postgresql.org/docs/current/arrays.html), but Postgres will complain if the [field size exceeds 1GB](https://www.postgresql.org/docs/current/arrays.html). 
From what I have (quickly) researched, I couldn't any info on what that optimal array size for inserting values would be. 

Do you believe this could be better addressed and reviewed in a new issue/PR?
If so, I can create and assign myself to it.

lutter (approved):

v0idpwn (approved):

PR #2372, opened 2021-04-12
Log if the notification queue is large

leoyvens:
This way we can inspect any notification queue buildup.

### DIFF
--- a/store/postgres/src/notification_listener.rs
+++ b/store/postgres/src/notification_listener.rs
@@ -137,8 +137,19 @@ impl NotificationListener {
                 // Wait until the listener has been started
                 barrier.wait();
 
+                let mut max_queue_size_seen = 0;
+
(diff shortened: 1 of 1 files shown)


store/postgres/src/notification_listener.rs
```diff
@@ -137,8 +137,19 @@ impl NotificationListener {
                 // Wait until the listener has been started
                 barrier.wait();
 
+                let mut max_queue_size_seen = 0;
+
                 // Read notifications until the thread is to be terminated
                 while !terminate.load(Ordering::SeqCst) {
+                    let queue_size = conn.notifications().len();
+                    if queue_size > 1000 && queue_size > max_queue_size_seen {
```
lutter, line 145: It might be good to limit how often we log this (e.g., every 5s) just so that this doesn't spam the logs too much.

lutter (approved):

PR #2373, opened 2021-04-12
Add `graphman rewind`

lutter:


store/postgres/src/deployment_store.rs
```diff
@@ -954,6 +947,33 @@ impl DeploymentStore {
...
+        block_ptr_to: EthereumBlockPointer,
+    ) -> Result<StoreEvent, StoreError> {
+        let conn = self.get_conn()?;
+        // Unwrap: If we are reverting then the block ptr is not `None`.
+        let block_ptr_from = Self::block_ptr_with_conn(&site.deployment, &conn)?.unwrap();
+
+        // Sanity check on block numbers
+        if block_ptr_from.number != block_ptr_to.number + 1 {
+            panic!("revert_block_operations must revert a single block only");
+        }
```
Jannis, line 972: Is this just here because we revert block by block usually? It's not here because reverting many blocks at once breaks things?

Jannis (commented):
There's one thing in the changed code that I'm not sure about. Otherwise the changes look correct.

store/postgres/src/deployment_store.rs
lutter: It probably used to be that we had to go block-by-block; but for quite some time now, it's been safe to jump the block pointer back (assuming you know that you are jumping _to_ the correct block pointer, which this PR slyly puts on the user) All the reversion queries just look at things that happened at/after the given block.

Jannis (approved):

PR #2374, opened 2021-04-12
Downgrade to tokio 0.2

leoyvens:
We are having prod issues that are related to a recent code change, and the tokio 1.0 upgrade is the prime suspect. This reverts to tokio 0.2, and fully reverts #2362 since we cannot update web3 if we remain in tokio 0.2. Note that this is not a total revert of #2334 but only the necessary changes to get back on 0.2.


lutter (approved):

leoyvens:
Thankfully this turned out to be unecessary.

PR #2376, opened 2021-04-13
Revert web3 update

leoyvens:
Both this and #2374 were tested in the community nodes and worked ok, so it seems that reverting the web3 update is necessary but reverting the tokio update is not, so we should prefer this over #2374. Also adds an unrelated log to `link_resolver.rs`.


lutter (approved):

leoyvens:
Not sure what exactly what's the issue with the latest web3, maybe it's that it doesn't restrict the number of parallel request anymore. But it does seem to be the cause of issues on the community nodes that have a ton of subgraphs making parallel calls.

PR #2378, opened 2021-04-14
Have a fallback timeout for chain head updates

leoyvens:
We believe chain head updates not going through are the cause for some issues seen in production, so introduce a fallback so that things can still work while we debug.


leoyvens:
There are other issues so tokio is getting reverted.

PR #2381, opened 2021-04-14
Downgrade noisy logs to trace

leoyvens:
Don't know what others think, but the 'start' and 'done' logs are basically redundant, but the 'done' includes the timing so I downgraded the 'start' to trace level. Also the contract call log isn't very useful so I downgraded it to 'trace' as well. The goal is to make logs a bit more scrollable and readable at debug level, which is what the hosted service uses.


lutter (approved):
I am all for that. I think we generate way too many logs, and a lot of times it's about showing that something is happening, for which logging might not be the best way to achieve that.

PR #2384, opened 2021-04-14
Parallel trace_filter calls

leoyvens:
This reduces the default range size to 50, so each request is less likely to timeout, but makes requests in parallel for a potential performance improvement.

### DIFF
--- a/chain/ethereum/src/ethereum_adapter.rs
+++ b/chain/ethereum/src/ethereum_adapter.rs
@@ -42,7 +42,7 @@ pub struct EthereumAdapter<T: web3::Transport> {
 
 lazy_static! {
     static ref TRACE_STREAM_STEP_SIZE: BlockNumber = std::env::var("ETHEREUM_TRACE_STREAM_STEP_SIZE")
-        .unwrap_or("200".into())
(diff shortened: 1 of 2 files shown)


chain/ethereum/src/ethereum_adapter.rs
```diff
@@ -306,17 +301,18 @@ where
...
                     end,
                     addresses.clone(),
-                )
-                .map(move |traces| (traces, new_start)),
-            )
+                ),
+                new_start,
+            )))
         })
+        .buffered(*BLOCK_BATCH_SIZE)
```
lutter, line 315: We should probably also update the documentation in `docs/environment-variables.md` to indicate that `ETHEREUM_BLOCK_BATCH_SIZE` is used to control the degree of parallelism of various operations, not just requests for blocks (and that it defaults to 10)

lutter (approved):

chain/ethereum/src/ethereum_adapter.rs
leoyvens: I had forgotten to do this, thanks!

PR #2396, opened 2021-04-21
store: Tolerate removing a deployment whose metadata is missing

lutter:
This fixes an issue where, if removing an unused subgraph gets interrupted after deleting the data but before deleting the subgraph from the primary, we would not be able to finish the removal.


lutter:
This is not needed anymore since PR #2313 fixes it a different way.

PR #2405, opened 2021-04-22
store: Make sure SubgraphStore.locators returns all deployment instances

lutter:
We were filtering out inactive instances because that's what we look at for
deployment status, but for locators, we need to consider active and
inactive instances of a deployment.


leoyvens (approved):

lutter:
The clever commit to force a recount of entities breaks a lot of tests.

lutter:
Fixed the problems with a more explicit way to do a recount.

leoyvens (approved):

PR #2411, opened 2021-04-26
Make the BlockIngestor independent from Ethereum (was: Simplify ingestion logic)

lutter:
This PR started out as a simple refactor of some ingestion logic, but now has the code to make block ingestion independent of Ethereum by introducing an `IngestorAdapter`.

This is what I had to say on the initial PR:

The code looked like we could try to find multiple missing ancestors for a  chain head candidate. But the logic of a chain dictates that we can only    ever find one block that we are missing when checking whether we have a    long enough chain in our database. Therefore, the vecs of missing parents    that we were passing around could only ever contain 0 or 1 element.
    
The code now clearly spells out that we will only ever know about one    missing parent at a time.

### DIFF
--- a/chain/ethereum/src/chain.rs
+++ b/chain/ethereum/src/chain.rs
@@ -1,0 +1,188 @@
+use std::{pin::Pin, sync::Arc, task::Context};
+
+use anyhow::Error;
+use graph::{
+    blockchain::{
+        block_stream::{
+            BlockStream, BlockStreamEvent, BlockWithTriggers, ScanTriggersError, TriggersAdapter,
+        },
+        Block, Blockchain, DataSource, IngestorAdapter, Manifest, TriggerFilter,
+    },
+    prelude::{async_trait, serde_yaml, BlockPtr, DeploymentHash, LinkResolver, Logger},
+    runtime::{AscType, DeterministicHostError},
+    tokio_stream::Stream,
+};
+
+struct Chain;
+
+impl Blockchain for Chain {
+    type Block = DummyBlock;
+
+    type DataSource = DummyDataSource;
+
+    type DataSourceTemplate = DummyDataSourceTemplate;
+
+    type Manifest = DummyManifest;
+
+    type TriggersAdapter = DummyTriggerAdapter;
+
+    type BlockStream = DummyBlockStream;
+
+    type TriggerData = DummyTriggerData;
+
+    type MappingTrigger = DummyMappingTrigger;
+
+    type TriggerFilter = DummyTriggerFilter;
+
+    type NodeCapabilities = DummyNodeCapabilities;
+
+    type IngestorAdapter = DummyIngestorAdapter;
+
+    fn reorg_threshold() -> u32 {
+        todo!()
+    }
+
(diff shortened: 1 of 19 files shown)


store/postgres/src/chain_store.rs
```diff
@@ -1159,67 +1168,83 @@ impl ChainStoreTrait for ChainStore {
...
-                        n::head_block_number.eq(number),
-                    ))
-                    .execute(&conn)?;
-                Ok((missing, Some((hash, number))))
-            })
-        }?;
+                    let hash = ptr.hash_hex();
+                    let number = ptr.number as i64;
+
+                    conn.transaction(
```
leoyvens, line 1233: Why is an explicit transaction necessary here if a single query is executed?

chain/ethereum/src/chain.rs
```diff
@@ -0,0 +1,263 @@
...
+            .eth_adapter
+            .load_full_block(&self.logger, block)
+            .compat()
+            .await?;
+
+        // Store it in the database and try to advance the chain head pointer
+        self.chain_store.upsert_block(block).await?;
+
+        self.chain_store
+            .clone()
```
leoyvens, line 246: `cheap_clone`

store/postgres/src/chain_store.rs
```diff
@@ -607,7 +610,7 @@ mod data {
             first_block: i64,
             hash: H256,
             genesis: H256,
-        ) -> Result<Vec<H256>, Error> {
+        ) -> Result<Option<H256>, Error> {
```
leoyvens, line 613: Should `missing_parents` be renamed to `missing_parent`?

leoyvens (approved):
This is great! Chains of the world, The Graph is coming!

node/src/main.rs
```diff
@@ -719,12 +721,22 @@ fn start_block_ingestor(
...
+                "BlockIngestor",
+                Some(ComponentLoggerConfig {
+                    elastic: Some(ElasticComponentLoggerConfig {
+                        index: String::from("block-ingestor-logs"),
+                    }),
+                }),
+            )
+            .new(o!("provider" => eth_adapter.provider().to_string()));
+
+            let chain = ethereum::Chain::new(chain_logger.clone(), chain_store, eth_adapter.clone(), *ANCESTOR_COUNT);
```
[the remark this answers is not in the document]
lutter: I am going light on how exactly setup/early startup works, since I know that that will need to be revisited in future PR's. Totally agree that it's a bit weird

store/postgres/src/chain_store.rs
lutter: It's not - it's more to make sure that if we ever add more write operations here we have it in a txn.

chain/ethereum/src/chain.rs
lutter: Changed

store/postgres/src/chain_store.rs
lutter: Totally. Fixed

PR #2416, opened 2021-04-28
Serialize entity fields in defined order

leoyvens:
Currently they are passed in the random order given by the hash map. Remove `Deref` and `DerefMut` from `Entity` to prevent bugs like this.

### DIFF
--- a/core/tests/interfaces.rs
+++ b/core/tests/interfaces.rs
@@ -23,7 +23,7 @@ fn insert_and_query(
             key: EntityKey {
                 subgraph_id: subgraph_id.clone(),
                 entity_type: EntityType::new(entity_type.to_owned()),
-                entity_id: data["id"].clone().as_string().unwrap(),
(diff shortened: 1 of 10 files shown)


lutter (approved):
This looks fine. My concern is that

* having to remember to call `sorted()` makes it easy to introduce subtle bugs by forgetting that call in the future
* a minor concern about introducing sorting (and therefore data copying) in lots of places influencing performance

We should figure out a way to just make `Entity` a `BTreeMap`.

leoyvens:
@lutter If you don't call `sorted`, you simply cannot iterate over the entity fields. Surprisingly to me, those two places that call `sorted()` in the runtime are the only ones that iterated over the fields without converting to a BTreeMap first, so I didn't bother with adding a better iteration API since it wouldn't be used. On performance, the only two places that call `sorted()` don't need a clone, so this isn't making perf any worse other than the cost of sorting. But generally I do think turning it into a BTreeMap has more performance gains than losses because we need to convert it into a BTreeMap for query results.

PR #2424, opened 2021-04-30
Consider providers that timeout as broken

leoyvens:
The thing is that `net_identifiers` will retry forever. So even if it tries 15 times and returns an error every time, it will continue retrying until the outer timeout is hit, so we will try to use the provider even if it is evidently defective. @lutter interested to know if you have a better solution for this.

### DIFF
--- a/node/src/main.rs
+++ b/node/src/main.rs
@@ -539,7 +539,6 @@ async fn connect_networks(
             network: String,
             provider: String,
         },
-        NoVersion,
         Version {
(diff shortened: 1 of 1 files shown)


lutter (approved):
I don't have a better idea either that we could implement quickly; we'd need to do a lot more error analysis in `net_identifiers` to distinguish between temporary and permanent failures, which is a little too much to take on right now.

PR #2426, opened 2021-05-01
Replace &String by &str

vrmiguel:
This PR replaces graph-node's usage of `&String` with `&str`

`&str` is preferable since it allows for more generic code, removing the need of the argument to be allocated on the heap (e.g., can remove the several instances of `&"some-term".to_string()` from the code, probably increasing performance).

This new change also removes a layer of indirection since you no longer need to deref `&String` to get a `String` and then deref that to get a `&str`

[Related Clippy lint](https://rust-lang.github.io/rust-clippy/master/index.html#ptr_arg)

P.S.: this ended up being divided in many little commits. I could squash all of them into a single one if so wished

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3664,9 +3664,9 @@ dependencies = [
 
 [[package]]
 name = "regex"
-version = "1.4.5"
+version = "1.4.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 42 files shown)


tilacog:
LGTM.
Can you please squash them into a single commit, please? 
Thanks! =)

vrmiguel:
> LGTM.
> Can you please squash them into a single commit, please?
> Thanks! =)

Done :D

vrmiguel:
This has turned into somewhat of a git hell so I feel like it'd be easier to redo the PR

PR #2435, opened 2021-05-05
graph: Replace some functions with lazily evaluated counterparts

vrmiguel:


graph/src/components/ethereum/adapter.rs
```diff
@@ -955,7 +955,7 @@ pub async fn blocks_with_triggers(
 
     // Make sure `to` is included, even if empty.
     block_hashes.insert(to_hash);
-    triggers_by_block.entry(to).or_insert(Vec::new());
```
leoyvens, line 958: `Vec::new` is a const fn, so I wouldn't change this.

vrmiguel: Makes sense! 

Reversed it

leoyvens (approved):

PR #2437, opened 2021-05-05
runtime: truncate some logs

leoyvens:
Otherwise this can print excessively long log lines.


lutter (approved):

leoyvens:
using `[..1024]` was actually not good because it would panic if the slice is does not have a length of a least 1024, I've fixed that.

PR #2438, opened 2021-05-06
Genericize the BlockStream with the Blockchain trait

lutter:
This PR separates the Ethereum-specific pieces out of the generic `BlockStream`, and makes it therefore reusable with other chains.


chain/ethereum/src/chain.rs
```diff
@@ -68,50 +119,141 @@ impl Blockchain for Chain {
...
     }
 
     fn parent_ptr(&self) -> Option<BlockPtr> {
-        todo!()
+        self.0.parent_ptr()
     }
 }
 
-pub struct DummyDataSource;
+pub struct WrappedDataSource(DataSource);
```
leoyvens, line 230: `WrappedDataSource` looks like dead code to me.

graph/src/blockchain/mod.rs
```diff
@@ -79,29 +64,39 @@ pub trait Blockchain: Sized + Send + Sync + 'static {
...
 
-    type NodeCapabilities;
+    type NodeCapabilities: std::fmt::Display;
 
     type IngestorAdapter: IngestorAdapter<Self>;
 
     // type RuntimeAdapter: RuntimeAdapter;
     // ...WIP
 
     fn reorg_threshold() -> u32;
```
leoyvens, line 74: We can just remove this, it's not needed here.

node/src/main.rs
```diff
@@ -685,12 +679,57 @@ fn create_ipfs_clients(logger: &Logger, ipfs_addresses: &Vec<String>) -> Vec<Ipf
...
     logger: &Logger,
-    block_polling_interval: Duration,
+    node_id: NodeId,
+    registry: Arc<MetricsRegistry>,
     eth_networks: &EthereumNetworks,
-    block_store: Arc<DieselBlockStore>,
+    store: &Store,
+    chain_head_update_listener: Arc<ChainHeadUpdateListener>,
     logger_factory: &LoggerFactory,
+) -> HashMap<String, Arc<ethereum::Chain>> {
```
leoyvens, line 690: How networks are going to be handled is something I was out of ideas on. Having separate `Chain` instances for different networks seems like a good solution! We will have to see how to adapt the config file format to handle different chains.

graph/src/blockchain/block_stream.rs
```diff
@@ -1,21 +1,57 @@
...
+use crate::components::store::BlockNumber;
+use crate::components::store::WritableStore;
+use crate::{prelude::*, prometheus::labels};
+
+use super::{Block, BlockPtr, Blockchain};
+
+#[cfg(debug_assertions)]
+use fail::fail_point;
 
 pub struct BlockWithTriggers<C: Blockchain> {
```
leoyvens, line 18: I'd really like to see `EthereumBlockWithTriggers` gone before merging, but I might be missing why that is a large lift.

leoyvens (changes requested):

graph/src/blockchain/mod.rs
```diff
@@ -68,7 +54,6 @@ pub trait Blockchain: Sized + Send + Sync + 'static {
     type Manifest: Manifest<Self>;
 
     type TriggersAdapter: TriggersAdapter<Self>;
-    type BlockStream: BlockStream<Self>;
```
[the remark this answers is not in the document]
lutter: I agree that the concrete block stream might not be right for all chains, but would define that interface and what it should look like once we have to deal with multiple block stream implementations.

graph/src/blockchain/mod.rs
```diff
@@ -79,29 +64,39 @@ pub trait Blockchain: Sized + Send + Sync + 'static {
...
+    ) -> Result<Arc<Self::TriggersAdapter>, Error>;
 
     fn new_block_stream(
         &self,
-        current_head: BlockPtr,
+        deployment: DeploymentLocator,
+        start_blocks: Vec<BlockNumber>,
         filter: Self::TriggerFilter,
-    ) -> Result<Self::BlockStream, Error>;
+        metrics: Arc<BlockStreamMetrics>,
```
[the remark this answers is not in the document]
lutter: It's a pass-through right now, and keeps the chain from having to understand what `metrics` means. Looking at `Blockchain.new_block_stream`, we could actually factor that back into a new `BlockStreamBuilder` since everything it needs from the chain is available as  public methods (mostly the chain store and the triggers adapter). But that's the other side of how generic the block stream needs to be: if we can reuse the block stream more widely, it makes sense to move the responsibility for creating one out of the chain and into a helper struct. If it's just Ethereum that uses that block stream, there's no point in splitting it out.

graph/src/blockchain/block_stream.rs
```diff
@@ -24,22 +60,729 @@ pub trait TriggersAdapter<C: Blockchain> {
     // `step_size` is the suggested number blocks to be scanned.
     async fn scan_triggers(
         &self,
-        chain_base: BlockPtr,
-        step_size: u32,
+        from: BlockNumber,
+        to: BlockNumber,
         filter: C::TriggerFilter,
-    ) -> Result<Vec<BlockWithTriggers<C>>, ScanTriggersError>;
+    ) -> Result<Vec<BlockWithTriggers<C>>, Error>;
```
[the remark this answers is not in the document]
lutter: But that interface would be wrong (or expensive) for Ethereum: if the caller makes no guarantee about finality, the adapter has to check that themselves. Right now, this is only called for blocks that are known to be final, and I should have added a comment to that effect. I just think that getting this right will be a lot easier with another example in hand. I think what you are getting at is who needs to know about finality: only the chain itself, or the generic framework, too. Right now, that knowledge is split, since the block stream depends on it. If we want to remove this knowledge from the framework, we'd have to consider the `BlockStream` chain-specific, which AFAICT would obviate the need for things like a `TriggerAdapter`.

graph/src/blockchain/block_stream.rs
lutter: The trouble here is that the runtime still wants an `EthereumBlockWithTriggers`; going from the generic block stream to the concrete runtime requires turning a `C::Block` into an `EthereumBlockWithTriggers`, which right now is done through the awful `TriggerFilter.convert_block`. Once the runtime also accepts `C::Block`, we can pull all of `graph::components::ethereum::types` into `chain::ethereum`.

node/src/main.rs
lutter: Yes, parsing the config and the subgraph manifest will require some amount of `serde` trickery; for configs there will need to be some sort of `enum` that lists all possible chains; it's not clear to me how that could become generic code.

graph/src/blockchain/mod.rs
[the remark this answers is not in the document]
leoyvens: Ok good point that this is more reusable.

graph/src/blockchain/mod.rs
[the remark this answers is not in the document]
leoyvens: Ok

graph/src/blockchain/block_stream.rs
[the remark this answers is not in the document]
leoyvens: Fair point that the Ethereum implementation wouldn't really be correct, so we can't have a single interface that's correct and performant for both cases. Agreed that we can figure this out if we start to work on a second block stream implementation.

graph/src/blockchain/block_stream.rs
leoyvens: Ah I see now. Maybe the `convert_block` could be done on `C::Block` then, but since this is temporary it's fine.

graph/src/blockchain/mod.rs
```diff
@@ -122,6 +117,12 @@ impl From<Error> for IngestorError {
 
 #[async_trait]
 pub trait IngestorAdapter<C: Blockchain> {
+    fn logger(&self) -> &Logger;
+
+    /// How long a chain from the current chain head back to blocks that are
+    /// considered final should be
```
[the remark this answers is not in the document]
lutter: I expanded the comment.

chain/ethereum/src/chain.rs
lutter: Removed.

graph/src/blockchain/mod.rs
lutter: Sadly, it is as an argument to `Blockchain.triggers_adapter`.

leoyvens (approved):
:shipit:

PR #2444, opened 2021-05-07
.github: Do not run CI on beta versions of the toolchain

lutter:


leoyvens:
Can we not disable this for integration tests since they are a bit flaky still?

lutter:
Added a commit. Our integration tests are so nice, you gotta run them twice.

leoyvens:
@tilacog the tests are getting diss'ed, we have to either implement those Docker Hub retries or come back with a better rhyme.

leoyvens (approved):

PR #2446, opened 2021-05-07
Write to the call cache concurrently to handler execution

leoyvens:
The call cache table can get very large, and therefore inserts can get slow. And when you think about it this is the only code path that writes to the DB during handler execution. So this might be a relevant optimization.


lutter (approved):
Nice!

On a related note, I've always wanted to add a background job that removes unneeded entries from that cache.

PR #2463, opened 2021-05-13
Move the `DataSource` type from `graph` to `ethereum`

leoyvens:
This does the minimal necessary to move the module `graph/src/data/subgraph/data_source.rs` to `chain/ethereum/src/data_source.rs`, which unblocks other refactorings. There are still a lot of Ethereum dependencies in the manifest data structures.

### DIFF
--- a/graph/Cargo.toml
+++ b/graph/Cargo.toml
@@ -65,3 +65,4 @@ web3 = { git = "https://github.com/graphprotocol/rust-web3", branch = "master" }
 test-store = { path = "../store/test-store" }
 maplit = "1.0.2"
 structopt = { version = "0.3" }
+graph-chain-ethereum = { path = "../chain/ethereum" }
--- a/graph/src/blockchain/mod.rs
+++ b/graph/src/blockchain/mod.rs
@@ -7,7 +7,12 @@ pub mod block_stream;
 mod types;
 
 // Try to reexport most of the necessary types
-use crate::{components::store::DeploymentLocator, runtime::AscType};
+use crate::{
+    components::store::DeploymentLocator,
+    data::subgraph::{Mapping, Source},
+    prelude::DataSourceContext,
+    runtime::AscType,
+};
 use crate::{
     components::store::{BlockNumber, ChainStore},
     prelude::{thiserror::Error, DeploymentHash, LinkResolver},
@@ -17,6 +22,7 @@ use async_trait::async_trait;
 use slog;
 use slog::Logger;
 use std::collections::HashMap;
+use std::fmt::Debug;
 use std::sync::Arc;
 use web3::types::H256;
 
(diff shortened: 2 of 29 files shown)


graph/Cargo.toml
```diff
@@ -65,3 +65,4 @@ web3 = { git = "https://github.com/graphprotocol/rust-web3", branch = "master" }
 test-store = { path = "../store/test-store" }
 maplit = "1.0.2"
 structopt = { version = "0.3" }
+graph-chain-ethereum = { path = "../chain/ethereum" }
```
lutter, line 68: I've been trying very hard to avoid this; totally fine as a temporary thing, but should have an `ETHDEP` comment with it that it needs to be removed before we can consider the refactor finished.

graph/src/blockchain/mod.rs
```diff
@@ -156,15 +162,38 @@ pub trait TriggerFilter<C: Blockchain>: Default + Clone + Send + Sync {
     fn node_capabilities(&self) -> C::NodeCapabilities;
 }
 
-pub trait DataSource<C: Blockchain>: 'static {
+// ETHDEP: `Source` and `Mapping`, at least, are Ethereum-specific.
+pub trait DataSource: 'static + Sized + Send + Sync + Debug {
```
lutter, line 166: Out of curiosity: why the switch from type parameter to associated type?

graph/src/components/store.rs
```diff
@@ -821,17 +824,13 @@ pub struct StoredDynamicDataSource {
     pub creation_block: Option<BlockNumber>,
 }
 
-impl From<&DataSource> for StoredDynamicDataSource {
-    fn from(ds: &DataSource) -> Self {
+impl<DS: DataSource> From<&DS> for StoredDynamicDataSource {
+    fn from(ds: &DS) -> Self {
```
lutter, line 828: Longer term, we want the specific blockchain to provide this conversion, and probably need to put the appropriate constraint on `blockchain::DataSource`. It's fine for now, but maybe worth an `ETHDEP` comment

lutter (approved):
Nice! This is a huge step forward.

graph/src/blockchain/mod.rs
leoyvens: I started getting `type parameter is unconstrained` errors when making types generic over `DataSource`.

graph/Cargo.toml
leoyvens: This is a `dev-dependencies`. To test things that live in `graph` we might need _some_ concrete implementation of the traits, I don't see harm in testing with Ethereum. This of course could never be a regular dependency since those cannot be circular.

graph/src/components/store.rs
leoyvens: Giving blockchains flexibility here is about more than just the conversion but also about the fields of `StoredDynamicDataSource`. `name`, `context` and `creation_block` can all be common, but what to do with the `Source` type is an open question. The solution here will fall out of how we solve the `Source` type.

graph/src/blockchain/mod.rs
leoyvens: Also just noticed this `Debug` bound ended up being unnecessary, I'm removing it.

PR #2465, opened 2021-05-14
Remove generic type parameter from ArrayBuffer

otaviopace:
I've done some simple fixes as well as removing the generic type parameter.

One of them is to rollback this PR: https://github.com/graphprotocol/graph-node/pull/2457

This is being done so that its a bit easier to do the AssemblyScript update (that I'm working on).

### DIFF
--- a/graph/src/runtime/mod.rs
+++ b/graph/src/runtime/mod.rs
@@ -25,7 +25,7 @@ use std::mem::size_of;
 pub trait AscType: Sized {
     /// Transform the Rust representation of this instance into an sequence of
     /// bytes that is precisely the memory layout of a corresponding Asc instance.
-    fn to_asc_bytes(self) -> Result<Vec<u8>, DeterministicHostError>;
+    fn to_asc_bytes(&self) -> Result<Vec<u8>, DeterministicHostError>;
 
     /// The Rust representation of an Asc object as layed out in Asc memory.
     fn from_asc_bytes(asc_obj: &[u8]) -> Result<Self, DeterministicHostError>;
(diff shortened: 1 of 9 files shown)


graph/src/runtime/mod.rs
```diff
@@ -36,6 +36,18 @@ pub trait AscType: Sized {
     }
 }
 
+impl<T> AscType for std::marker::PhantomData<T> {
```
otaviopace, line 39: I don't know if this is the best place to put this implementation though.

graph/src/runtime/mod.rs
```diff
@@ -36,6 +36,18 @@ pub trait AscType: Sized {
...
+    fn to_asc_bytes(&self) -> Result<Vec<u8>, DeterministicHostError> {
+        Ok(vec![])
+    }
+
+    fn from_asc_bytes(asc_obj: &[u8]) -> Result<Self, DeterministicHostError> {
+        assert!(asc_obj.len() == 0);
```
otaviopace, line 45: Should this be a `DeterministicHostError`?

runtime/derive/Cargo.toml
```diff
@@ -7,5 +7,5 @@ edition = "2018"
 proc-macro = true
 
 [dependencies]
-syn = { version = "0.15", features = ["full"] }
-quote = "0.6"
+syn = { version = "1.0", features = ["full"] }
```
otaviopace, line 10: I just updated this because I was trying some things on our derive crate and I couldn't do it because of the version.

I may use this updates later. Even if I don't, I think its good to keep up to date.

graph/src/runtime/mod.rs
leoyvens: I agree that this is an assert.

graph/src/runtime/mod.rs
leoyvens: Here is fine, but please add a comment that this is only to `derive(AscType)` for structs that contain `PhantomData`, and it is not really an asc type.

runtime/derive/Cargo.toml
leoyvens: If tests are happy, I'm happy.

leoyvens (approved):

graph/src/runtime/mod.rs
otaviopace: Done ✅

PR #2477, opened 2021-05-18
Remove a couple clones of trigger filters

lutter:


chain/ethereum/src/data_source.rs
```diff
@@ -187,7 +187,7 @@ impl DataSource {
                 .iter()
                 .find(move |handler| {
                     handler.filter.is_some()
-                        && handler.filter.clone().unwrap() == BlockHandlerFilter::Call
+                        && handler.filter.as_ref().unwrap() == &BlockHandlerFilter::Call
```
leoyvens, line 190: Lets refactor this whole condition to `handler.filter == Some(BlockHandlerFilter::Call)`

leoyvens (approved):
lgtm, just have a bonus refactor suggestion.

chain/ethereum/src/data_source.rs
lutter: Good point. Not sure why I didn't write it like that right away. Fixed now

PR #2478, opened 2021-05-20
chain: Fix error in EthereumAdapter.get_calls

lutter:
This fixes a logic error introduced in commit fad1679b


leoyvens (approved):
Lets wait for integration tests to pass :wink:

PR #2482, opened 2021-05-21
Better abstract over `TriggerData`, `MappingTrigger` and `DataSource`

leoyvens:
This brings us close to dropping the runtime depedency on the `ethereum` crate.

- `EthereumTrigger` and `MappingTrigger` are moved from `graph` to `ethereum`.
- `DataSource` starts being used through the trait, removing the `todo!()`.
- A few signatures start using the associated types instead of concrete types.


core/src/subgraph/instance_manager.rs
```diff
@@ -179,10 +162,8 @@ impl SubgraphInstanceMetrics {
         }
     }
 
-    pub fn observe_trigger_processing_duration(&self, duration: f64, trigger: TriggerType) {
-        self.trigger_processing_duration
-            .with_label_values(vec![trigger.label_value()].as_slice())
-            .observe(duration);
```
leoyvens, line 185: This was a functional change, I don't think this label is very useful so it wasn't worth abstracting it, instead it was removed.

lutter: I wonder if we could just get rid of the entire metric - I think the data we get out of the various stopwatch sections is much more useful.

leoyvens: Could be, but for the purpose of this PR removing the label is enough.

graph/src/components/store.rs
```diff
@@ -829,7 +829,11 @@ impl<DS: DataSource> From<&DS> for StoredDynamicDataSource {
         Self {
             name: ds.name().to_owned(),
             source: ds.source().clone(),
-            context: ds.context().map(|ctx| serde_json::to_string(&ctx).unwrap()),
+            context: ds
```
lutter, line 832: Ultimately, the data model for dds is still pretty eth dependent; we might have to change it to something that just stores the dds as a JSONB blob to get the flexibility we need; that should be fine since we never query for individual dds attributes.

runtime/wasm/src/host.rs
```diff
@@ -174,7 +172,7 @@ pub struct RuntimeHost<C: Blockchain> {
 
 impl<C> RuntimeHost<C>
 where
-    C: Blockchain<DataSource = graph_chain_ethereum::DataSource>,
+    C: Blockchain,
```
lutter, line 175: :tada:

lutter (approved):
Nice! Big step forward. Just a few idle thoughts as I was reading the code, nothing that blocks merging

graph/src/components/store.rs
leoyvens: Yes there are decisions to be made here, storing as jsonb is an interesting idea.

graph/src/blockchain/mod.rs
```diff
@@ -51,18 +50,18 @@ pub trait Block: Send + Sync {
 }
 
 pub trait Blockchain: Sized + Send + Sync + 'static {
-    type Block: Block;
+    type Block: Block + Clone;
```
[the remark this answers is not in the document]
leoyvens: I've added a comment. In terms of performance I believe the impact is small because this is only cloned when reprocessing a block, but I agree that it's unfortunate to require this bound.

PR #2483, opened 2021-05-21
Add timestamps for deployment_schemas and unused_deployments tables

### LINKED ISSUE #2224: Record when a subgraph was created
The `deployment_schemas` table should have a column `created_at` that records when a deployment was created. The information could be gleaned from `subgraph_version`, but that's a bit too cumbersome, and longer term, `subgraph_version` should go away.

The `created_at` timestamp should also be copied to `unused_deployments` when a deployment is recorded as unused.

tilacog:
Resolves  #2224

- [x] add timestamp to `deployment_schemas` table
- [x] copy that column data  to `unused_deployments` when a deployment is recorded as unused  
- [x] fix compilation errors

### DIFF
--- a/store/postgres/migrations/2021-05-21-165219_add_created_at_columns_for_deployment_schemas/up.sql
+++ b/store/postgres/migrations/2021-05-21-165219_add_created_at_columns_for_deployment_schemas/up.sql
@@ -2,4 +2,13 @@ alter table public.deployment_schemas
     add column created_at timestamptz not null default now();
 
 alter table public.unused_deployments
-    add column created_at timestamptz not null;
+    add column created_at timestamptz;
+
+-- use a predefined default date that doesn't have a meaning.
+update
+    public.unused_deployments
+set
+    created_at = '2000-01-01';
+
+alter table public.unused_deployments
+    alter column created_at set not null;
--- a/store/postgres/migrations/2021-05-21-165219_add_created_at_columns_for_deployment_schemas/up.sql
+++ b/store/postgres/migrations/2021-05-21-165219_add_created_at_columns_for_deployment_schemas/up.sql
@@ -1,0 +1,5 @@
+alter table public.deployment_schemas
+    add column created_at timestamptz not null default now();
+
+alter table public.unused_deployments
+    add column deployment_created_at timestamptz not null;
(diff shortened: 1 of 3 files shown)


store/postgres/migrations/2021-05-21-165219_add_created_at_columns_for_deployment_schemas/up.sql
```diff
@@ -0,0 +1,5 @@
+alter table public.deployment_schemas
+    add column created_at timestamptz not null default now();
+
+alter table public.unused_deployments
+    add column deployment_created_at timestamptz not null;
```
tilacog, line 5: @lutter I'm not sure how should we handle default values in this case.

For migrations applied on a running instance, this migration will probably fail due to missing values.
Should we allow this column to contain null values?

store/postgres/migrations/2021-05-21-165219_add_created_at_columns_for_deployment_schemas/up.sql
```diff
@@ -0,0 +1,5 @@
+alter table public.deployment_schemas
+    add column created_at timestamptz not null default now();
+
+alter table public.unused_deployments
+    add column created_at timestamptz not null;
```
lutter, line 5: Since databases will have entries in `unused_deployments`, you'll need to populate the new column before setting it `not null`.  You'll have to do something like
```sql
alter table unused_deployments
        add column created_at timestamptz;
-- use a dummy date that is clearly made up
update unused_deployments set created_at = '2000-01-01';
alter table unused_deployments
        alter column created_at set not null;
```

tilacog: Got it.
It would be OK to use a placeholder value, like the one in your example?

store/postgres/migrations/2021-05-21-165219_add_created_at_columns_for_deployment_schemas/up.sql
lutter: Yes, that was my intent - use a value that is clearly made up in case we ever need to tell real values from fake ones.

lutter (approved):
Looks good. Can you remove the draft label on the PR?

PR #2486, opened 2021-05-23
Dont try to insert 0 ddses

leoyvens:
This is called on every `apply_entity_modifications`, and I believe even if empty diesel will still generate trivial SQL for the DB to execute.

### DIFF
--- a/store/postgres/src/dynds.rs
+++ b/store/postgres/src/dynds.rs
@@ -118,6 +118,11 @@ pub(crate) fn insert(
 ) -> Result<usize, StoreError> {
     use dynamic_ethereum_contract_data_source as decds;
 
+    if data_sources.is_empty() {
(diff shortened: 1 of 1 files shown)


store/postgres/src/dynds.rs
```diff
@@ -154,6 +154,11 @@ pub(crate) fn insert(
         })
         .collect::<Result<_, _>>()?;
 
+    if dds.is_empty() {
```
lutter, line 157: It would be better to check `data_sources` before even constructing `dds`.

lutter (approved):

store/postgres/src/dynds.rs
leoyvens: Collecting an empty iterator is a no-op anyways, but I suppose it's more readable if the check is at the start of the function.

Issue #1804, opened 2020-07-23
Fulltext directive breaks: Error: _Schema__orderBy values must be an object with value names as keys.


cbrzn:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**

Trying to develop the fulltext search query in a subgraph and I am getting an error, this is what's happening:

I have added the fulltext search directive and had to update graph packages in order to make this functionality available (since it's a new one). Right now everything compiles, the deployment works, but when I go to `http://localhost:8000/subgraphs/name/protofire/omen/graphql`, I am getting an error:
```Error: _Schema__orderBy values must be an object with value names as keys.```

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem.**

- Download the repository https://github.com/dOrgTech/omen-subgraph
- Checkout to branch `issue-28-fulltext_search_support`
- Make sure you have docker installed and running
- Run in a terminal (inside of project's folder): `docker-compose up`, wait until everything is downloaded and running
- In another terminal, do: `npm i && npm run migrate && npm run codegen && npm run create-local && npm run deploy-local`
- Go to browser and go to the following url:  `http://localhost:8000/subgraphs/name/protofire/omen`
- You should see the error

<img width="1277" alt="Screenshot 2020-07-23 at 21 54 38" src="https://user-images.githubusercontent.com/21031138/88332612-1e13ea00-cd2f-11ea-8fb3-8f3f623426e4.png">

**What is the expected behavior?**

I should be able to do queries and test this new directive

edit: you can check the source code here https://github.com/protofire/omen-subgraph/pull/32/files

cag:
Hey so I tested this out, and you can still do the queries actually. The issue is that the autocomplete/doc generator is broken for some reason, which is not nearly as bad.

cbrzn:
Awesome, I will close this then sir! @cag Thank you for checking it

fubhy:
This is actually a valid issue report and definitely a bug that should be fixed. Introspection is important for quite a few useful tools in the graphql space.

Issue #2507, opened 2021-05-31
Invalid GraphQL Schema at https://gateway.testnet.thegraph.com/network


fubhy:
**Do you want to request a *feature* or report a *bug*?**
Bug report

**What is the current behavior?**
The schema hosted at https://gateway.testnet.thegraph.com/network is invalid

The invalid part of the schema seems to be an empty enum type.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem.**
Regenerating a schema from an introspection query on https://gateway.testnet.thegraph.com/network currently yields this error:

```Error: _Schema__orderBy values must be an object with value names as keys```

This can easily be reproduced by using a graphql query client like graphiql or graphql playground (yoga) or postmark, insomnia, etc. or by attempting to use code generation tooling (e.g. graphql-codegen).

**What is the expected behavior?**

Enums aren't generated if the enum would end up not having any key-value pairs.

fubhy:
https://github.com/graphprotocol/graph-node/blob/ed7510674993b4d3436e1967a923b16cf385e04a/graphql/src/schema/api.rs#L283

fubhy:
Related: https://github.com/graphprotocol/graph-node/issues/1804

fubhy:
There also seem to be other schema issues as I'm also now getting this error: `Null value resolved for non-null field __typename`

Maybe it would be good to run the introspection query alongside the queries sent by the existing test cases that run queries against subgraphs and the other (e.g. status) graphql endpoints.

Jannis:
@fubhy Is still a problem? I can't reproduce it using graphiql-online.com.

fubhy:
@Jannis Looks good now (also just verified that on https://studio.apollographql.com/).

I think #2508 might've fixed this.

PR #2508, opened 2021-05-31
fix: do not generate empty enums

fubhy:


fubhy:
Fixes https://github.com/graphprotocol/graph-node/issues/1804 and https://github.com/graphprotocol/graph-node/issues/2507

graphql/src/schema/api.rs
```diff
@@ -263,6 +263,11 @@ fn add_order_by_type(
 
     match ast::get_named_type(schema, &type_name) {
         None => {
+            // Don't generate empty enums.
+            if fields.is_empty() {
+                return Ok(());
+            }
+
```
Jannis, line 270: If I remember the API generation code correctly, there's another place where the `_orderBy` type is referenced (where the top-level query fields are generated). That will now refer to a non-existent `_orderBy` type, which will probably break schema introspection as well.

Even if I don't like special-casing, we may have to add a static/hard-coded list of types for which to do things a little differently.

fubhy: I've changed the approach and went a bit further to completely remove the `_Schema_` type from the generated schema after all associated information from attached directives has been extracted. I think that's the saner approach anyways considering that the `_Schema_` type seems to only be used for metadata / configuration purposes.

The introspection works now also when there are full-text search fields.

graphql/src/schema/api.rs
```diff
@@ -296,15 +308,6 @@ fn add_filter_type(
...
         None => {
-            let input_values = field_input_values(schema, fields)?;
-
-            // Don't generate an input object with no fields, this makes the JS
-            // graphql library, which graphiql uses, very confused and graphiql
-            // is unable to load the schema. This happens for example with the
-            // definition `interface Foo { x: OtherEntity }`.
-            if input_values.is_empty() {
-                return Ok(());
-            }
```
fubhy, line 307: There is no such case anymore as far as I can tell and the example given here doesn't seem correct because it then still lets you filter by the reference entity id? Hence, I removed this bit and also simplified the logic accordingly in the other places where this was then taken care of.

leoyvens (approved):

PR #2511, opened 2021-06-01
Prevent handling triggers for failed transactions

### LINKED ISSUE #2409: Out of gas tx triggers subgraph call handler
[This transaction](https://etherscan.io/tx/0x2cd6ef64770b12e29006fd3e4a9cd008e98f62d51f9f7b91e58d7d670219b8da) ran out of gas. The purpose of this transaction was to execute 3 different transactions through a transaction batcher contract. You can see that the first of those transactions is calling executeRequest() of [this contract](https://etherscan.io/address/0xC5E9dDebb09Cd64DfaCab4011A0D5cEDaf7c9BDb).

My understanding is that this reverted transaction should not have triggered any call in the [subgraph](https://thegraph.com/explorer/subgraph/kleros/proof-of-humanity-mainnet). However, it did trigger executeRequest() call handler. Is this a bug or am I missing something?

tilacog:
Resolves #2409

- [x] Handle Non-Final blocks
- [x] Handle Final blocks
  - [x] Fetch existing transaction receipts from database
  - [x] Fetch existing transaction receipts from ethereum client
- [x] Add stopwatch metrics
- [x] Add manifest version switch

This PR adds a verification step to call-handler execution that filters off traces with failed transactions.

For **non-final blocks**, we determine failure by inspecting each transaction receipt and gas usage, which are already present in program memory, so it should be fast. 

For **final blocks**, we search the database once for each block. If that data is absent, we call an external API once for each transaction receipt.

### DIFF
--- a/chain/ethereum/Cargo.toml
+++ b/chain/ethereum/Cargo.toml
@@ -28,6 +28,7 @@ ethabi = { git = "https://github.com/graphprotocol/ethabi.git", branch = "master
 
 # We have a couple custom patches to web3.
 web3 = { git = "https://github.com/graphprotocol/rust-web3", branch = "master" }
+itertools = "0.10.0"
 
 graph-runtime-wasm = { path = "../../runtime/wasm" }
 graph-runtime-derive = { path = "../../runtime/derive" }
--- a/chain/ethereum/src/adapter.rs
+++ b/chain/ethereum/src/adapter.rs
@@ -24,6 +24,7 @@ use graph::{components::ethereum::EthereumNetworkIdentifier, prelude::*};
 use crate::{data_source::DataSource, Chain};
 
 pub type EventSignature = H256;
+pub type FunctionSelector = [u8; 4];
 
 #[derive(Clone, Debug)]
 pub struct EthereumContractCall {
@@ -270,7 +271,8 @@ impl EthereumLogFilter {
 pub(crate) struct EthereumCallFilter {
     // Each call filter has a map of filters keyed by address, each containing a tuple with
     // start_block and the set of function signatures
-    pub contract_addresses_function_signatures: HashMap<Address, (BlockNumber, HashSet<[u8; 4]>)>,
+    pub contract_addresses_function_signatures:
+        HashMap<Address, (BlockNumber, HashSet<FunctionSelector>)>,
 }
 
 impl EthereumCallFilter {
@@ -355,12 +357,12 @@ impl EthereumCallFilter {
     }
 }
 
-impl FromIterator<(BlockNumber, Address, [u8; 4])> for EthereumCallFilter {
+impl FromIterator<(BlockNumber, Address, FunctionSelector)> for EthereumCallFilter {
     fn from_iter<I>(iter: I) -> Self
     where
-        I: IntoIterator<Item = (BlockNumber, Address, [u8; 4])>,
+        I: IntoIterator<Item = (BlockNumber, Address, FunctionSelector)>,
     {
-        let mut lookup: HashMap<Address, (BlockNumber, HashSet<[u8; 4]>)> = HashMap::new();
+        let mut lookup: HashMap<Address, (BlockNumber, HashSet<FunctionSelector>)> = HashMap::new();
         iter.into_iter()
             .for_each(|(start_block, address, function_signature)| {
                 if !lookup.contains_key(&address) {
@@ -387,7 +389,7 @@ impl From<EthereumBlockFilter> for EthereumCallFilter {
                 .contract_addresses
                 .into_iter()
                 .map(|(start_block_opt, address)| (address, (start_block_opt, HashSet::default())))
-                .collect::<HashMap<Address, (BlockNumber, HashSet<[u8; 4]>)>>(),
+                .collect::<HashMap<Address, (BlockNumber, HashSet<FunctionSelector>)>>(),
         }
     }
 }
--- a/chain/ethereum/src/chain.rs
+++ b/chain/ethereum/src/chain.rs
@@ -1,10 +1,8 @@
-use std::collections::HashSet;
-use std::iter::FromIterator;
-use std::sync::Arc;
-
 use anyhow::{Context, Error};
 use graph::data::subgraph::UnifiedMappingApiVersion;
-use graph::prelude::{EthereumCallCache, LightEthereumBlock, LightEthereumBlockExt};
+use graph::prelude::{
+    EthereumCallCache, LightEthereumBlock, LightEthereumBlockExt, StopwatchMetrics,
+};
 use graph::{
     blockchain::{
         block_stream::{
@@ -23,6 +21,9 @@ use graph::{
         LoggerFactory, MetricsRegistry, NodeId, SubgraphStore,
     },
 };
+use std::collections::HashSet;
+use std::iter::FromIterator;
+use std::sync::Arc;
 
 use crate::data_source::DataSourceTemplate;
 use crate::data_source::UnresolvedDataSourceTemplate;
@@ -142,6 +143,7 @@ impl Blockchain for Chain {
         loc: &DeploymentLocator,
         capabilities: &Self::NodeCapabilities,
         unified_api_version: UnifiedMappingApiVersion,
+        stopwatch_metrics: StopwatchMetrics,
     ) -> Result<Arc<Self::TriggersAdapter>, Error> {
         let eth_adapter = self.eth_adapters.cheapest_with(capabilities)?.clone();
         let logger = self
@@ -154,8 +156,9 @@ impl Blockchain for Chain {
             logger,
             ethrpc_metrics,
             eth_adapter,
+            stopwatch_metrics,
             chain_store: self.chain_store.cheap_clone(),
-            _unified_api_version: unified_api_version,
+            unified_api_version,
         };
         Ok(Arc::new(adapter))
     }
@@ -182,7 +185,12 @@ impl Blockchain for Chain {
         let requirements = filter.node_capabilities();
 
         let triggers_adapter = self
-            .triggers_adapter(&deployment, &requirements, unified_api_version.clone())
+            .triggers_adapter(
+                &deployment,
+                &requirements,
+                unified_api_version.clone(),
+                metrics.stopwatch.clone(),
+            )
             .expect(&format!(
                 "no adapter for network {} with capabilities {}",
                 self.name, requirements
@@ -330,9 +338,10 @@ impl Manifest<Chain> for DummyManifest {
 pub struct TriggersAdapter {
     logger: Logger,
     ethrpc_metrics: Arc<SubgraphEthRpcMetrics>,
+    stopwatch_metrics: StopwatchMetrics,
     chain_store: Arc<dyn ChainStore>,
     eth_adapter: Arc<EthereumAdapter>,
-    _unified_api_version: UnifiedMappingApiVersion,
+    unified_api_version: UnifiedMappingApiVersion,
 }
 
 #[async_trait]
@@ -348,9 +357,11 @@ impl TriggersAdapterTrait<Chain> for TriggersAdapter {
             self.logger.clone(),
             self.chain_store.clone(),
             self.ethrpc_metrics.clone(),
+            self.stopwatch_metrics.clone(),
             from,
             to,
             filter,
+            self.unified_api_version.clone(),
         )
         .await
     }
@@ -378,9 +389,11 @@ impl TriggersAdapterTrait<Chain> for TriggersAdapter {
                     logger.clone(),
                     self.chain_store.clone(),
                     self.ethrpc_metrics.clone(),
+                    self.stopwatch_metrics.clone(),
                     block_number,
                     block_number,
                     filter,
+                    self.unified_api_version.clone(),
                 )
                 .await?;
                 assert!(blocks.len() == 1);
@@ -392,7 +405,7 @@ impl TriggersAdapterTrait<Chain> for TriggersAdapter {
                     &filter.log,
                     &full_block.ethereum_block,
                 ));
-                triggers.append(&mut parse_call_triggers(&filter.call, &full_block));
+                triggers.append(&mut parse_call_triggers(&filter.call, &full_block)?);
                 triggers.append(&mut parse_block_triggers(filter.block.clone(), &full_block));
                 Ok(BlockWithTriggers::new(block, triggers))
             }
--- a/chain/ethereum/src/ethereum_adapter.rs
+++ b/chain/ethereum/src/ethereum_adapter.rs
@@ -1,25 +1,22 @@
+use ethabi::ParamType;
 use ethabi::Token;
 use futures::future;
 use futures::prelude::*;
-use lazy_static::lazy_static;
-use std::collections::{HashMap, HashSet};
-use std::convert::TryFrom;
-use std::iter::FromIterator;
-use std::sync::Arc;
-use std::time::Instant;
-
-use ethabi::ParamType;
+use graph::components::transaction_receipt::LightTransactionReceipt;
+use graph::data::subgraph::UnifiedMappingApiVersion;
+use graph::prelude::StopwatchMetrics;
 use graph::{
     blockchain::{block_stream::BlockWithTriggers, BlockPtr, IngestorError},
     prelude::{
-        anyhow, async_trait, debug, error, ethabi,
+        anyhow::{self, anyhow, bail},
+        async_trait, debug, error, ethabi,
         futures03::{self, compat::Future01CompatExt, FutureExt, StreamExt, TryStreamExt},
-        hex, retry, stream, tiny_keccak, trace, warn,
+        hex, info, retry, stream, tiny_keccak, trace, warn,
         web3::{
             self,
             types::{
                 Address, Block, BlockId, BlockNumber as Web3BlockNumber, Bytes, CallRequest,
-                FilterBuilder, Log, H256,
+                Filter, FilterBuilder, Log, Transaction, TransactionReceipt, H256,
             },
         },
         BlockNumber, ChainStore, CheapClone, DynTryFuture, Error, EthereumCallCache, Logger,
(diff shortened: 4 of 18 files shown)


graph/src/components/ethereum/types.rs
```diff
@@ -111,6 +111,35 @@ pub struct EthereumBlockWithCalls {
...
+        };
+
+        // TODO: should we handle None differently than Some(0)?
+        match receipt.status {
+            Some(x) if x == web3::types::U64::from(1) => true,
+            Some(_) | None => false,
+        }
+    }
+}
+
```
tilacog, line 142: I'm not sure should we handle the `None` branches in this function, as stated in the `TODO` comments.

chain/ethereum/src/ethereum_adapter.rs
```diff
@@ -1588,14 +1588,21 @@ pub(crate) fn parse_log_triggers(
...
             .filter(move |call| call_filter.matches(call))
-            .map(move |call| EthereumTrigger::Call(Arc::new(call.clone())))
+            .map(
+                move |call| match block.transaction_for_call_succeeded(call) {
+                    Ok(true) => Ok(Some(EthereumTrigger::Call(Arc::new(call.clone())))),
+                    Ok(false) => Ok(None),
+                    Err(e) => Err(e),
+                },
+            )
+            .filter_map_ok(|some_trigger| some_trigger)
```
tilacog, line 1603: This ended up a bit convoluted, but the point is that we need to keep `Err`s  and drop `Ok(None)`s for the conversion to `Result<Vec<_>>` to work.

leoyvens: At this point I'd just use a `for` so that you can use `?`, but this also works.

graph/src/components/ethereum/types.rs
```diff
@@ -11,6 +11,8 @@ use crate::{
     prelude::{BlockNumber, DeploymentHash, EntityKey, ToEntityKey},
 };
 
+const BYZANTIUM_FORK_BLOCK_NUMBER: u64 = 4_370_000;
```
leoyvens, line 14: You can't do this, the block number will be different for different networks such as testnets. Simply assume that if all gas was used, that is a failed transaction regardless of the block.

chain/ethereum/src/ethereum_adapter.rs
```diff
@@ -707,8 +712,27 @@ impl EthereumAdapter {
...
         }
 
+        // Hit the database and fetch all failed transactions in the block range.
+        // TODO: handle this Result/expect
+        let mut transaction_statuses_in_block_range = {
+            stopwatch_metrics.start_section("calls_in_block_range__fetch_transaction_statuses");
+            chain_store
+                .transaction_statuses_in_block_range(&(from..to))
+                .expect("failed to fetch failed transactions in the database")
+        };
```
leoyvens, line 722: You can do an early return with a `once` stream that emits an error https://docs.rs/futures/0.1.31/futures/stream/fn.once.html.

chain/ethereum/src/ethereum_adapter.rs
```diff
@@ -707,8 +712,27 @@ impl EthereumAdapter {
...
+                .filter(move |trace| {
+                    // TODO: handle this Result/expect
+                    trace_transaction_succeeded(
+                        &trace,
+                        &eth2,
+                        &mut transaction_statuses_in_block_range,
+                        &stopwatch_metrics,
+                    )
+                    .expect("failed to determine if trace's transaction has failed")
+                })
```
leoyvens, line 735: You might need to `filter_map` into a Result and then do an `and_then(|x| x)`.

store/postgres/src/chain_store.rs
```diff
@@ -1413,6 +1450,74 @@ impl ChainStoreTrait for ChainStore {
...
+            &self.chain,
+            &block_range,
+        )?;
+
+        let mut statuses = HashMap::new();
+        let mut pending = HashMap::new();
+        for receipt in receipts.into_iter() {
+            match (receipt.status, receipt.gas_used) {
+                (Some(_status), _) => {
+                    statuses.insert(receipt.transaction_hash, receipt.is_sucessful());
```
leoyvens, line 1479: We could replace `is_successful()` with just `!status.is_zero()`

store/postgres/src/chain_store.rs
```diff
@@ -1413,6 +1450,74 @@ impl ChainStoreTrait for ChainStore {
...
+        let receipts = self.storage.find_transaction_receipts_for_block_range(
+            &conn,
+            &self.chain,
+            &block_range,
+        )?;
+
+        let mut statuses = HashMap::new();
+        let mut pending = HashMap::new();
+        for receipt in receipts.into_iter() {
+            match (receipt.status, receipt.gas_used) {
```
leoyvens, line 1477: If `gas_used` is `None` this should return an error.

chain/ethereum/src/ethereum_adapter.rs
leoyvens: Alternatively, we can log the error and move on as if we had nothing cached.

chain/ethereum/src/adapter.rs
```diff
@@ -24,6 +25,7 @@ use graph::{components::ethereum::EthereumNetworkIdentifier, prelude::*};
 use crate::{data_source::DataSource, Chain};
 
 pub type EventSignature = H256;
+pub type FunctionSelector = [u8; 4];
```
tilacog, line 28: I created this new type before I knew I wouldn't come back to this anymore.
I can revert those changes if necessary.

chain/ethereum/src/adapter.rs
```diff
@@ -653,6 +656,8 @@ pub trait EthereumAdapter: Send + Sync + 'static {
         call: EthereumContractCall,
         cache: Arc<dyn EthereumCallCache>,
     ) -> Box<dyn Future<Item = Vec<Token>, Error = EthereumContractCallError> + Send>;
+
+    fn api_version(&self) -> Version;
```
tilacog, line 660: This is necessary to access `apiVersion` from inside an adapter.

chain/ethereum/src/chain.rs
```diff
@@ -140,6 +142,8 @@ impl Blockchain for Chain {
         &self,
         loc: &DeploymentLocator,
         capabilities: &Self::NodeCapabilities,
+        unified_api_version: UnifiedMappingApiVersion,
+        stopwatch_metrics: Arc<StopwatchMetrics>,
```
tilacog, line 146: All edits in this file are related to passing those types around.

chain/ethereum/src/ethereum_adapter.rs
```diff
@@ -1499,6 +1504,22 @@ pub(crate) async fn blocks_with_triggers(
...
+            let futures = blocks.into_iter().map(|block| {
+                filter_call_triggers_from_unsuccessful_transactions(block, &eth, &chain_store)
+            });
+            let blocks = futures03::future::try_join_all(futures).await?;
+            section.end();
+            blocks
+        } else {
+            blocks
+        };
+
```
tilacog, line 1522: Not sure if this is the most performant way to do `async/await`, but I opted to do it the simplest way without making calls wait for each other.  

Also, this `let/if/else` is necessary due to `filter_call_triggers_from_unsuccessful_transactions` consuming and returning the `blocks` value. 

I tried to go with `&mut` but I found a lot of borrow-checking problems. Maybe there is a better way to do it.

chain/ethereum/src/ethereum_adapter.rs
```diff
@@ -1624,3 +1652,149 @@ pub(crate) fn parse_block_triggers(
...
+    match eth
+        .web3
+        .eth()
+        .transaction_receipt(*transaction_hash)
+        .compat()
+        .await
+    {
+        Ok(Some(receipt)) => Ok(receipt),
+        Ok(None) => bail!("Could not find transaction receipt"),
+        Err(error) => bail!("Failed to fetch transaction receipt: {}", error),
```
tilacog, line 1669: I'm considering the absence of a receipt to be an error here.

chain/ethereum/src/ethereum_adapter.rs
```diff
@@ -1624,3 +1652,149 @@ pub(crate) fn parse_block_triggers(
...
+                    .await
+                    .map(|receipt| (transaction, receipt))
+            });
+        futures03::future::try_join_all(futures.into_iter())
+            .await?
+            .into_iter()
+            .for_each(|(transaction, receipt)| {
+                receipts_and_transactions.push((transaction, receipt.into()))
+            });
+    }
```
tilacog, line 1750: I'm using the same `async/await` strategy from before.

chain/ethereum/src/ethereum_adapter.rs
```diff
@@ -1624,3 +1652,149 @@ pub(crate) fn parse_block_triggers(
...
+    block.trigger_data.retain(|trigger| {
+        if let EthereumTrigger::Call(call_trigger) = trigger {
+            // Unwrap: We already checked that those values exist
+            transaction_success[&call_trigger.transaction_hash.unwrap()]
+        } else {
+            // We are not filtering other types of triggers
+            true
+        }
+    });
+    Ok(block)
```
tilacog, line 1779: It's important to note that a block with empty triggers might be returned from this.
I'm not sure if I should drop the whole block in those cases.

chain/ethereum/src/ethereum_adapter.rs
```diff
@@ -1624,3 +1652,149 @@ pub(crate) fn parse_block_triggers(
...
+        .collect();
+
+    // Then we get the transaction hash for each call
+    let transaction_hashes = calls
+        .iter()
+        .map(|call| call.transaction_hash)
+        .collect::<Option<BTreeSet<H256>>>()
+        .ok_or(anyhow!(
+            "failed to obtain transaction hash from call triggers"
+        ))?;
```
tilacog, line 1695: It's not clear to me when an `EthereumCall` might have a `None` for its `transaction_hash` field.

Since that information is required for the rest of the function, I'm considering its absence to be an error.

graph/src/components/ethereum/types.rs
```diff
@@ -67,6 +67,42 @@ pub struct EthereumBlockWithCalls {
...
+            .gas_used
+            .ok_or(anyhow::anyhow!("Running in light client mode)"))?
+            >= transaction.gas
+        {
+            return Ok(false);
+        }
+        Ok(matches!(receipt.status, Some(x) if x == web3::types::U64::from(1)))
+    }
+}
+
```
tilacog, line 105: This is for non-final blocks.

graph/src/components/transaction_receipt.rs
```diff
@@ -0,0 +1,216 @@
...
+    Ok(output)
+}
+
+#[test]
+fn test_drain_vector() {
+    let input = vec![191, 153, 17];
+    let expected_output = [0, 0, 0, 0, 0, 191, 153, 17];
+    let result = drain_vector(input).expect("failed to drain vector into array");
+    assert_eq!(result, expected_output);
+}
```
tilacog, line 139: Helper function to convert Vectors into Arrays while ensuring matching lengths.

graph/src/components/transaction_receipt.rs
```diff
@@ -0,0 +1,216 @@
...
+
+/// Queries the database for all the transaction receipts in a given block range.
+pub fn find_transaction_receipts_for_block(
+    conn: &PgConnection,
+    chain_name: &str,
+    block_hash: &H256,
+) -> anyhow::Result<Vec<LightTransactionReceipt>> {
+    let query = TransactionReceiptQuery {
+        schema_name: chain_name,
```
tilacog, line 179: I'm still not sure if `chain_name` maps strictly to `schema_name`
I'll have that checked.

graph/src/components/transaction_receipt.rs
```diff
@@ -0,0 +1,216 @@
...
+        block_hash: block_hash.as_bytes(),
+    };
```
tilacog, line 181: I'm still not sure if `chain_name` maps strictly to `schema_name`
I'll have that checked.

graph/src/data/subgraph/mod.rs
```diff
@@ -666,6 +673,59 @@ impl UnresolvedMapping {
...
+    pub fn equal_or_greater_than(&self, other_version: &Version) -> bool {
+        assert!(
+            other_version >= &API_VERSION_0_0_5,
+            "api versions before 0.0.5 should not be used for comparison"
+        );
+        match &self.0 {
+            Some(version) => version >= other_version,
+            None => false,
+        }
+    }
```
tilacog, line 689: I was having some broad lifetimes issues when using:

```rust
other_version: &'static Version
``` 

So I removed the `'static` lifetime bound.

graph/src/data/subgraph/mod.rs
```diff
@@ -246,36 +251,6 @@ impl<'de> de::Deserialize<'de> for SubgraphName {
...
-    assert!(SubgraphName::new("a/a_").is_err());
-    assert!(SubgraphName::new("a/_a").is_err());
-    assert!(SubgraphName::new("aaaa aaaaa").is_err());
-    assert!(SubgraphName::new("aaaa!aaaaa").is_err());
-    assert!(SubgraphName::new("aaaa+aaaaa").is_err());
-    assert!(SubgraphName::new("a/graphql").is_err());
-    assert!(SubgraphName::new("graphql/a").is_err());
-    assert!(SubgraphName::new("this-component-is-longer-than-the-length-limit").is_err());
-}
-
```
tilacog, line 278: Moved to the bottom of the file.

chain/ethereum/src/adapter.rs
leoyvens: I like it, worth a comment linking to Solidity docs on this.

chain/ethereum/Cargo.toml
```diff
@@ -27,6 +27,8 @@ ethabi = { git = "https://github.com/graphprotocol/ethabi.git", branch = "master
 
 # We have a couple custom patches to web3.
 web3 = { git = "https://github.com/graphprotocol/rust-web3", branch = "master" }
+itertools = "0.10.0"
+diesel = { version = "1.4.6", features = ["postgres", "serde_json", "numeric", "r2d2"] }
```
leoyvens, line 31: What is this `diesel` dep doing?

chain/ethereum/src/chain.rs
```diff
@@ -142,6 +143,7 @@ impl Blockchain for Chain {
         loc: &DeploymentLocator,
         capabilities: &Self::NodeCapabilities,
         unified_api_version: UnifiedMappingApiVersion,
+        stopwatch_metrics: Arc<StopwatchMetrics>,
```
leoyvens, line 146: `StopwatchMetrics` is already a reference, it should not be wrapped in `Arc`.

chain/ethereum/src/ethereum_adapter.rs
```diff
@@ -1627,3 +1655,151 @@ pub(crate) fn parse_block_triggers(
...
+) -> anyhow::Result<BlockWithTriggers<crate::Chain>> {
+    // First, we separate call triggers from other trigger types
+    let calls: Vec<&Arc<EthereumCall>> = block
+        .trigger_data
+        .iter()
+        .filter_map(|trigger| match trigger {
+            EthereumTrigger::Call(call_trigger) => Some(call_trigger),
+            _ => None,
+        })
+        .collect();
```
leoyvens, line 1689: I don't see a need to collect this.

chain/ethereum/src/ethereum_adapter.rs
```diff
@@ -1627,3 +1655,151 @@ pub(crate) fn parse_block_triggers(
...
+    mut block: BlockWithTriggers<crate::Chain>,
+    eth: &EthereumAdapter,
+    chain_store: &Arc<dyn ChainStore>,
+) -> anyhow::Result<BlockWithTriggers<crate::Chain>> {
+    // First, we separate call triggers from other trigger types
+    let calls: Vec<&Arc<EthereumCall>> = block
+        .trigger_data
+        .iter()
+        .filter_map(|trigger| match trigger {
+            EthereumTrigger::Call(call_trigger) => Some(call_trigger),
```
leoyvens, line 1686: You can map straight to `call_trigger.transaction_hash`.

leoyvens, line 1686: Is it possible to map straight into `call.transaction_hash`, effectively merging this with the block below?

store/postgres/src/chain_store.rs
```diff
@@ -1413,6 +1432,16 @@ impl ChainStoreTrait for ChainStore {
             .block_number(&conn, hash)?
             .map(|number| (self.chain.clone(), number)))
     }
+
+    fn transaction_receipts_in_block(
```
leoyvens, line 1436: Store operations are blocking. Since this is called in an async context, it should be made async. You can accomplish this by using `with_conn` instead of `get_conn`.

graph/src/components/ethereum/types.rs
```diff
@@ -67,6 +67,42 @@ pub struct EthereumBlockWithCalls {
...
+        // assume the transaction failed if all gas was used
+        if receipt
+            .gas_used
+            .ok_or(anyhow::anyhow!("Running in light client mode)"))?
+            >= transaction.gas
+        {
+            return Ok(false);
+        }
+        Ok(matches!(receipt.status, Some(x) if x == web3::types::U64::from(1)))
```
leoyvens, line 102: This logic seems duplicated with `fn is_transaction_successful`, is there a way to de-duplicate it?

tilacog: Placed the common logic in the new `evaluate_transaction_status` function.

tilacog:
Please don't merge yet.
I've to check this w/ a live subgraph first.

graph/src/data/subgraph/mod.rs
```diff
@@ -656,7 +656,7 @@ impl UnresolvedMapping {
 pub struct UnifiedMappingApiVersion(Option<Version>);
 
 impl UnifiedMappingApiVersion {
-    pub fn equal_or_greater_than(&self, other_version: &'static Version) -> bool {
+    pub fn equal_or_greater_than(&self, other_version: &Version) -> bool {
```
leoyvens, line 659: the `'static` here doesn't work anymore?

leoyvens (approved):

PR #2512, opened 2021-06-01
Make `AscHeap` dyn-safe

leoyvens:
By moving `try_asc_get`, `asc_get` and `asc_new` out of the trait.

The only real change is in `graph/src/runtime/asc_heap.rs`, the rest is just mechanical refactoring from the fallout.

### DIFF
--- a/graph/src/runtime/asc_heap.rs
+++ b/graph/src/runtime/asc_heap.rs
@@ -5,73 +5,88 @@ use super::{AscPtr, AscType, DeterministicHostError};
 /// for reading and writing Rust structs from and to Asc.
 ///
 /// The implementor must provide the direct Asc interface with `raw_new` and `get`.
-pub trait AscHeap: Sized {
+pub trait AscHeap {
     /// Allocate new space and write `bytes`, return the allocated address.
     fn raw_new(&mut self, bytes: &[u8]) -> Result<u32, DeterministicHostError>;
 
     fn get(&self, offset: u32, size: u32) -> Result<Vec<u8>, DeterministicHostError>;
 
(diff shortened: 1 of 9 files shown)


otaviopace:
The code seems correct, I just want to know more about the change itself.

Wasn't `AscHeap` already dyn safe? I mean, it had `Sized` as super trait.

This PR probably is to help on something else you're working on, how does it?

leoyvens:
@otaviopace A trait that has `Sized` as a supertrait is not object safe, because `dyn Foo: !Sized`. It helps on abstracting the runtime for multiblockchain, so that I can write `Box<dyn Fn(&mut dyn AscHeap, u32) -> u32>` as an abstraction for a host function that receives a pointer and returns another.

otaviopace (approved):

otaviopace (changes requested):
Changed my mind, I was researching on "sizeness" in Rust and I've found you can remove the `Sized` bound in `trait AscHeap: Sized` and add it to the methods you want like this:

```rust
    fn asc_new<C, T: ?Sized>(&mut self, rust_obj: &T) -> Result<AscPtr<C>, DeterministicHostError>
    where
        Self: Sized,
        C: AscType,
        T: ToAscObj<C>,
    {
        let obj = rust_obj.to_asc_obj(self)?;
        AscPtr::alloc_obj(obj, self)
    }
```

I just added `Self: Sized,` above on the `where` clause.

It will work as long as any of these marked methods don't get called on unsized objects 😉

leoyvens:
@otaviopace But I do want to call `asc_new` from a `dyn AscHeap`.

otaviopace (approved):

PR #2514, opened 2021-06-02
Control which tables are 'account like' without query node restarts

lutter:
With these changes, which tables are 'account like' is set through a database table, and the corresponding information is refreshed every 5 minutes, making it possible to change query behavior without restarting query nodes.

A future enhancement would be to gather information about which tables are 'account like', i.e., with a high ratio of entity versions to entities automatically, and base the decision on that data. Gathering this data isn't entirely trivial, and is therefore left out of this PR.

### DIFF
--- a/store/postgres/src/relational.rs
+++ b/store/postgres/src/relational.rs
@@ -323,7 +323,7 @@ impl Layout {
             .map(|table| {
                 format!(
                     "select count(*) from \"{}\".\"{}\" where block_range @> {}",
-                    &catalog.namespace, table.name, BLOCK_NUMBER_MAX
+                    &catalog.site.namespace, table.name, BLOCK_NUMBER_MAX
                 )
             })
             .collect::<Vec<_>>()
@@ -350,7 +350,7 @@ impl Layout {
         let table_name = SqlName::verbatim(POI_TABLE.to_owned());
         Table {
             object: POI_OBJECT.to_owned(),
-            qualified_name: SqlName::qualified_name(&catalog.namespace, &table_name),
+            qualified_name: SqlName::qualified_name(&catalog.site.namespace, &table_name),
             name: table_name,
             columns: vec![
                 Column {
@@ -391,7 +391,7 @@ impl Layout {
         site: Arc<Site>,
         schema: &Schema,
     ) -> Result<Layout, StoreError> {
-        let catalog = Catalog::new(conn, site.namespace.clone())?;
+        let catalog = Catalog::new(conn, site.clone())?;
         let layout = Self::new(site, schema, catalog, true)?;
         let sql = layout
             .as_ddl()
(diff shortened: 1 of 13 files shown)


docker/tag.sh
```diff
@@ -20,9 +20,9 @@ set -ex
 
 tag_and_push "$SHORT_SHA"
 
-# Builds on the master branch become the 'latest'
-[ "$BRANCH_NAME" = master ] && tag_and_push latest
 # Builds of tags set the tag in Docker Hub, too
 [ -n "$TAG_NAME" ] && tag_and_push "$TAG_NAME"
+# Builds for tags vN.N.N become the 'latest'
+[[ "$TAG_NAME" =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]] && tag_and_push latest
```
leoyvens, line 26: This change doesn't belong in this branch. Though it's a great idea!

store/postgres/src/relational.rs
```diff
@@ -1331,6 +1338,130 @@ fn is_object_type(field_type: &q::Type, enums: &EnumMap) -> bool {
...
+                if now <= expires {
+                    // Entry is not expired; use it
+                    return Ok(value);
+                } else {
+                    // Only do a cache refresh once; we don't want to have
+                    // multiple threads refreshing the same layout
+                    // simultaneously. It's easiest to refresh at most one
+                    // layout globally
+                    let refresh = self.refresh.try_lock();
+                    if let Err(_) = refresh {
```
leoyvens, line 1427: I usually just unwrap poison errors. Or use the `parking_lot` mutex which doesn't poison.

store/postgres/src/relational.rs
```diff
@@ -1331,6 +1338,130 @@ fn is_object_type(field_type: &q::Type, enums: &EnumMap) -> bool {
...
+            let entry = CacheEntry {
+                expires: Instant::now() + self.ttl,
+                value: layout,
+            };
+            &self.entries.lock().unwrap().insert(deployment, entry);
+        }
+    }
+
+    /// Return the corresponding layout if we have one in cache already, and
+    /// ignore expiration infromation
```
leoyvens, line 1393: s/infromation/information

leoyvens (approved):
Looks good! @lutter the indexbender.

docker/tag.sh
lutter: I merged and then didn't push master

store/postgres/src/relational.rs
lutter: This isn't about poisoning, but about whether we would block if we tried to lock, in which case we just return the old layout

leoyvens: oh of course, that's what happens when I try to do reviews at the end of the day.

PR #2518, opened 2021-06-02
docker: Do not tag master builds as `latest`

lutter:
Only builds whose tag matches ^vN.N.N$ will change the `latest` tag from now on.


Jannis (approved):
Perfect!

PR #2530, opened 2021-06-03
Change how tests ensure that store tests run sequentially

lutter:
We used the `STORE_RUNTIME` to both force tests to run sequentially, and to run async tasks. That created a chicken-and-egg problem where setting up the store in `test-store` would deadlock during a test. Tests sidestepped that by causing store setup to run before locking the `STORE_RUNTIME`, but that meant that each test could hold a connection while waiting to get the `STORE_RUNTIME`, exhausting the connection pool and ultimately deadlocking the tests; an unnerving experience since it's never clear whether new code or the test harness causes such deadlocks.

This PR separates using the runtime and running tests sequentially by using a standalone lock; for this to work, it was necessary to switch to the threaded scheduler.

It is now possible to run the tests with a pool as small as 5 connections.

### DIFF
--- a/chain/ethereum/tests/manifest.rs
+++ b/chain/ethereum/tests/manifest.rs
@@ -136,56 +136,53 @@ graft:
 specVersion: 0.0.2
 ";
 
-    test_store::run_test_sequentially(
-        || (),
-        |store, ()| async move {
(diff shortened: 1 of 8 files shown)


leoyvens (approved):
Nice!

PR #2538, opened 2021-06-04
store: Avoid using subgraph_deployment when looking for non fatal errors

lutter:
In the common case where we know at which block we are looking for non-fatal errors, do not involve subgraph_deployment since it's a very busy table, we do this lookup a lot, and there seem to be lock waits because of it


leoyvens (approved):
Why does a SELECT need to wait for locks?

lutter:
> Why does a SELECT need to wait for locks?

You are right - it shouldn't cause any lock waits. But it's still better to not make PG access that table.

PR #2558, opened 2021-06-10
multiblockchain: Flip dependency between the runtime and ethereum crates

leoyvens:
This PR removes the depedency of `runtime` on `ethereum`, in fact now `ethereum` depends on `runtime`.

- Adds `Blockchain::RuntimeAdapter`.
- Move ethereum-specific `AscType`s into `ethereum`.
- Does a first pass at abstracting host funtions with the `RuntimeAdapter` trait and the `HostFn` and `HostFnCtx` data structures.
- Move all the `ethereum.call` related functionality into the ethereum `RuntimeAdapter`.
- The runtime is no longer aware of specific trigger types, the `handle_*` functions are replaced with the generic `handle_trigger`, this is enabled by adding `MappingTrigger::to_asc`, to which the logic that serializes the trigger to AS was moved.
- `EthereumCallError` is merged into `HostExportError` by adding `HostExportError::PossibleReorg`.
- The `core` crate now depends directly on the runtime so that it can instantiate the concrete `RuntimeHostBuilder`, which now depends on the blockchain making it hard to instantiate it on `main.rs`.

The second commit fixes up runtime tests and moves them to a dedicated `runtime/test` crate, so that the tests can depend on `ethereum` without creating a circular `dev-dependency`.


graph/src/blockchain/mod.rs
```diff
@@ -73,7 +75,7 @@ pub trait Blockchain: Debug + Sized + Send + Sync + 'static {
     type TriggerData: TriggerData + Ord;
 
     /// Decoded trigger ready to be processed by the mapping.
-    type MappingTrigger: MappingTrigger;
+    type MappingTrigger: MappingTrigger + Debug;
```
otaviopace, line 78: Silly question: is this the best place to put the `Debug`? I just don't know really 😅

graph/src/blockchain/mod.rs
```diff
@@ -264,6 +267,42 @@ pub trait TriggerData {
     fn error_context(&self) -> String;
 }
 
-pub trait MappingTrigger {
+pub trait MappingTrigger: Send + Sync {
     fn handler_name(&self) -> &str;
+
+    /// A flexible interface for writing a type to AS memory, any pointer can be returned.
+    /// Use `AscPtr::erased` to convert `AscPtr<T>` into `AscPtr<()>`.
+    fn to_asc<H: AscHeap>(self, heap: &mut H) -> Result<AscPtr<()>, DeterministicHostError>;
```
otaviopace, line 275: What do you think about naming `to_asc_ptr` instead?

We have `to_asc_obj` and `to_asc_bytes`, by following the pattern I think this one should be specific as well.

runtime/derive/src/lib.rs
```diff
@@ -72,21 +72,21 @@ fn asc_type_derive_struct(item_struct: ItemStruct) -> TokenStream {
                 #(bytes.extend_from_slice(&self.#field_names.to_asc_bytes()?);)*
 
                 // Assert that the struct has no padding.
-                assert_eq!(bytes.len(), size_of::<Self>());
+                assert_eq!(bytes.len(), std::mem::size_of::<Self>());
```
otaviopace, line 75: Nice! I actually wanted to change our derive crate to use more full names, it's a good practice in proc macros.

Reference: https://github.com/dtolnay/proc-macro-workshop/blob/master/builder/tests/09-redefined-prelude-types.rs

I mean, only our code uses it, but I guess it's probably better to avoid conflicting names errors.

otaviopace:
Conceptual question, have you moved the `runtime` tests to the same level as the crates `runtime/derive` and `runtime/wasm` because they test `derive` as well indirectly? Like this:

Screen Shot 2021-06-15 at 17 04 35

chain/ethereum/src/runtime/runtime_adapter.rs
```diff
@@ -0,0 +1,206 @@
...
+
+use super::abi::{AscUnresolvedContractCall, AscUnresolvedContractCall_0_0_4};
+
+pub struct RuntimeAdapter {
+    pub(crate) eth_adapters: Arc<EthereumNetworkAdapters>,
+    pub(crate) call_cache: Arc<dyn EthereumCallCache>,
+}
+
+impl blockchain::RuntimeAdapter<Chain> for RuntimeAdapter {
+    fn host_fns(&self, ds: &DataSource) -> Result<Vec<HostFn>, Error> {
```
otaviopace, line 29: Would it make sense for the `HostFn` vector to be a `HashMap` using the `function` names as keys?

chain/ethereum/src/runtime/abi.rs
```diff
@@ -0,0 +1,290 @@
...
+};
+
+use super::runtime_adapter::UnresolvedContractCall;
+
+type AscH256 = Uint8Array;
+type AscLogParamArray = Array<AscPtr<AscLogParam>>;
+
+#[repr(C)]
+#[derive(AscType)]
+pub(crate) struct AscUnresolvedContractCall_0_0_4 {
```
lutter, line 20: What's the significance of the `0_0_4` suffix?

lutter (approved):
LGTM

chain/ethereum/src/runtime/abi.rs
otaviopace: Subgraph's apiVersion

graph/src/blockchain/mod.rs
leoyvens: It could also be a super trait on `MappingTrigger`. Unsure of what's best.

runtime/derive/src/lib.rs
leoyvens: Yes definitely bad practice to use names that are not fully qualified.

chain/ethereum/src/runtime/runtime_adapter.rs
leoyvens: I guess that would guarantee that the function names are unique, but besides that it wouldn't gain us much. I like the simplicity of good ol' `Vec`.

And `HashMap` iteration order is not deterministic, so I'd use `BTreeMap`.

graph/src/blockchain/mod.rs
leoyvens: Renamed.

leoyvens:
@otaviopace not directly related to derive, but since everything other directory in `runtime/` is also a crate, I figured to stick to the convention. Thanks for your review, I've addressed your comments.

PR #2559, opened 2021-06-10
store: Enclose column names in double quotes

tilacog:


store/postgres/src/relational_queries.rs
```diff
@@ -2903,7 +2903,9 @@ fn write_column_names(column_names: &AttributeNames, table: &Table, out: &mut As
                 })
                 .peekable();
             while let Some(column_name) = iterator.next() {
+                out.push_sql("\"");
```
lutter, line 2906: It would be better to do these as `out.push_identifier(column_name.as_str())?`

leoyvens:
Are we able to test the problematic case?

lutter (approved):
Just change to `push_identifier` before committing.

tilacog:
> Are we able to test the problematic case?

Sure. I'll write a test for a subgraph that has entities with fields named after SQL reserved words.

PR #2562, opened 2021-06-11
runtime: Fix json_to_bytes error not null on success bug

otaviopace:
Fix the bug introduced in here: https://github.com/graphprotocol/graph-node/pull/2513, where the error wasn't null on success.


runtime/wasm/src/module/test.rs
```diff
@@ -314,7 +314,7 @@ async fn json_parsing() {
     let bytes_ptr = asc_new(&mut module, bytes).unwrap();
     let return_value: AscPtr<AscString> = module.invoke_export("handleJsonError", bytes_ptr);
     let output: String = asc_get(&module, return_value).unwrap();
-    assert_eq!(output, "OK: foo, ERROR: false");
+    assert_eq!(output, "OK: foo, ERROR: null? true");
```
leoyvens, line 317: Why did `ERROR` go from `false` to `true`?

otaviopace: I inverted the logic, because now it calls `isError`, and that should be `false`.

leoyvens: But if `isError` is false, why should this print `ERROR: true`?

otaviopace: Because of this: https://github.com/graphprotocol/graph-node/pull/2562/files#diff-d71d37021eff538d2206947314070971b299fc6d355def1b4ad1359846b2fb23R74

otaviopace: I can invert something if it's confusing.

leoyvens: It just doesn't seem logical to print `ERROR: true` when there wasn't an error.

otaviopace: That's why I've put the `null?`. I'll leave as it was then.

leoyvens: Aaaah, now I get it, this is saying "yes this is null".

leoyvens (approved):

PR #2569, opened 2021-06-15
Reduce max delay

leoyvens:
30 seconds can be a long time to be stuck in a retry loop that might just be failing repeatedly and waiting for the retries to run out. This reduces the max delay between retries to 5 seconds.


leoyvens:
Closing this because spamming eth_call with non-existing block hash can put Matic nodes further behind. I was targeting reducing the wait for block ingestor retries, but we should probably have a more specific solution for that.

PR #2576, opened 2021-06-17
ethereum: Remove support for old config

leoyvens:
This code supported a config format that I believe nobody is using, and it has been superseded by the new config format. This drops the `lexical-core` transitive dependency that is not building on Rust 1.53.


lutter (approved):
I _think_ you are right that nobody is using this. If this functionality is still needed, we might want to incorporate it into `graph-node.toml` at some point. But for now, we can just get rid of it.

Jannis:
Before we remove this:

- Do we have and data points to back up the theory that nobody is using this?
- Do we have a full replacement where you can define custom http headers for ethereum nodes?

leoyvens:
@Jannis I tried searching the Discord and didn't find any discussion of it. I've amended the commit to add the headers functionality to the new config format, so we don't lose that.

Jannis:
@leoyvens Excellent, that works. Could we also remove `docs/ethereum-config.md`, maybe rename `sharding.md` to `config.md` and add an example on how to define HTTP headers in that file?

leoyvens:
@Jannis I've renamed `sharding.md` and did a few updates to it, including documenting the headers feature.

Jannis (approved):
Perfect, thanks for the changes @leoyvens!

PR #2578, opened 2021-06-21
Prohibit subgraphs from using different api_versions

tilacog:
This PR adds the following:
- [X] Manifest validation logic to prohibit subgraphs from using different `apiVersions` across mappings
- [X] Passes the unified mapping API version value to `BlockStream` and `TriggerAdapter` instances.

### DIFF
--- a/chain/ethereum/src/chain.rs
+++ b/chain/ethereum/src/chain.rs
@@ -155,7 +155,7 @@ impl Blockchain for Chain {
             ethrpc_metrics,
             eth_adapter,
             chain_store: self.chain_store.cheap_clone(),
-            unified_api_version,
+            _unified_api_version: unified_api_version,
         };
         Ok(Arc::new(adapter))
     }
@@ -332,7 +332,7 @@ pub struct TriggersAdapter {
     ethrpc_metrics: Arc<SubgraphEthRpcMetrics>,
     chain_store: Arc<dyn ChainStore>,
     eth_adapter: Arc<EthereumAdapter>,
-    unified_api_version: UnifiedMappingApiVersion,
+    _unified_api_version: UnifiedMappingApiVersion,
 }
 
 #[async_trait]
--- a/chain/ethereum/src/chain.rs
+++ b/chain/ethereum/src/chain.rs
@@ -141,6 +141,7 @@ impl Blockchain for Chain {
         &self,
         loc: &DeploymentLocator,
         capabilities: &Self::NodeCapabilities,
+        unified_api_version: UnifiedMappingApiVersion,
     ) -> Result<Arc<Self::TriggersAdapter>, Error> {
         let eth_adapter = self.eth_adapters.cheapest_with(capabilities)?.clone();
         let logger = self
@@ -154,6 +155,7 @@ impl Blockchain for Chain {
             ethrpc_metrics,
             eth_adapter,
             chain_store: self.chain_store.cheap_clone(),
+            unified_api_version,
         };
         Ok(Arc::new(adapter))
     }
@@ -180,7 +182,7 @@ impl Blockchain for Chain {
         let requirements = filter.node_capabilities();
 
         let triggers_adapter = self
-            .triggers_adapter(&deployment, &requirements)
+            .triggers_adapter(&deployment, &requirements, unified_api_version.clone())
             .expect(&format!(
                 "no adapter for network {} with capabilities {}",
                 self.name, requirements
@@ -330,6 +332,7 @@ pub struct TriggersAdapter {
     ethrpc_metrics: Arc<SubgraphEthRpcMetrics>,
     chain_store: Arc<dyn ChainStore>,
     eth_adapter: Arc<EthereumAdapter>,
+    unified_api_version: UnifiedMappingApiVersion,
 }
 
 #[async_trait]
--- a/chain/ethereum/src/chain.rs
+++ b/chain/ethereum/src/chain.rs
@@ -3,6 +3,7 @@ use std::iter::FromIterator;
 use std::sync::Arc;
 
 use anyhow::{Context, Error};
+use graph::data::subgraph::UnifiedMappingApiVersion;
 use graph::prelude::{EthereumCallCache, LightEthereumBlock, LightEthereumBlockExt};
 use graph::{
     blockchain::{
@@ -163,6 +164,7 @@ impl Blockchain for Chain {
         start_blocks: Vec<BlockNumber>,
         filter: TriggerFilter,
         metrics: Arc<BlockStreamMetrics>,
+        unified_api_version: UnifiedMappingApiVersion,
     ) -> Result<BlockStream<Self>, Error> {
         let logger = self
             .logger_factory
@@ -198,6 +200,7 @@ impl Blockchain for Chain {
             metrics,
             *MAX_BLOCK_RANGE_SIZE,
             *TARGET_TRIGGERS_PER_BLOCK_RANGE,
+            unified_api_version,
         ))
     }
 
--- a/core/src/subgraph/instance_manager.rs
+++ b/core/src/subgraph/instance_manager.rs
@@ -1,8 +1,5 @@
 use super::loader::load_dynamic_data_sources;
-use super::loader::load_dynamic_data_sources;
 use super::SubgraphInstance;
-use super::SubgraphInstance;
-use crate::subgraph::registrar::IPFS_SUBGRAPH_LOADING_TIMEOUT;
 use atomic_refcell::AtomicRefCell;
 use fail::fail_point;
 use graph::components::arweave::ArweaveAdapter;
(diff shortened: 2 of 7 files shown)


graph/src/data/subgraph/mod.rs
```diff
@@ -814,6 +819,23 @@ impl<C: Blockchain> UnvalidatedSubgraphManifest<C> {
             errors.push(SubgraphManifestValidationError::DataSourceBlockHandlerLimitExceeded)
         }
 
+        // For API versions newer than 0.5, validate that all mappings uses the same api_version
+        let referential_api_version = Version::new(0, 5, 0);
```
leoyvens, line 823: 0.0.5, not 0.5!

core/src/subgraph/instance_manager.rs
```diff
@@ -323,7 +325,7 @@ where
             .and_then(|x| x)?;
         }
 
-        let manifest = {
+        let manifest: SubgraphManifest<C> = {
             info!(logger, "Resolve subgraph files using IPFS");
```
tilacog, line 329: Not 100% sure if this is desirable, but I had to specify the concrete type here. Otherwise, it would use the `BaseSubgraphManifest` type, which has no methods.

graph/src/data/subgraph/mod.rs
```diff
@@ -814,6 +859,27 @@ impl<C: Blockchain> UnvalidatedSubgraphManifest<C> {
...
+        {
+            let error_msg = unique_api_versions
+                .iter()
+                .map(ToString::to_string)
+                .join(", ");
+            errors.push(SubgraphManifestValidationError::DifferentApiVersions(
+                error_msg,
+            ));
+        }
+
```
tilacog, line 882: I could not reuse the `UnifiedMappingApiVersion` logic in this spot because it would silently resolve into a `None` for invalid cases.  

Also, we're now iterating over `Manifest.mappings(), which yields values both from `.data_sources` and `.templates`.

chain/ethereum/src/chain.rs
```diff
@@ -153,6 +155,7 @@ impl Blockchain for Chain {
             ethrpc_metrics,
             eth_adapter,
             chain_store: self.chain_store.cheap_clone(),
+            _unified_api_version: unified_api_version,
```
tilacog, line 158: Using underscores here because this value is not used yet.

graph/src/data/subgraph/mod.rs
```diff
@@ -956,6 +1022,10 @@ impl<C: Blockchain> SubgraphManifest<C> {
             }),
         }
     }
+
+    pub fn unified_mapping_api_version(&self) -> UnifiedMappingApiVersion {
+        self.mappings().iter().collect()
+    }
```
tilacog, line 1028: Iterating over `mappings()` so we get all values from both `data_sources` and `templates`

core/src/subgraph/instance_manager.rs
leoyvens: This is fine, it's just a type hint.

graph/src/data/subgraph/mod.rs
leoyvens: >I could not reuse the UnifiedMappingApiVersion logic in this spot because it would silently resolve into a None for invalid cases.

It doesn't have to be that way, you can make that return a `Result`. We _really_ should avoid duplicating the logic.

leoyvens (changes requested):

graph/src/data/subgraph/mod.rs
```diff
@@ -666,6 +675,42 @@ impl UnresolvedMapping {
...
+impl UnifiedMappingApiVersion {
+    pub fn equal_or_greater_than(&self, other_version: &'static Version) -> bool {
+        match &self.0 {
+            Some(version) => version >= other_version,
+            None => false,
+        }
+    }
+}
+
+impl<'a> FromIterator<&'a Mapping> for UnifiedMappingApiVersion {
```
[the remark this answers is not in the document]
tilacog: I can see the error now, thanks! 
I'll rewrite this using Results and add some tests

graph/src/data/subgraph/mod.rs
```diff
@@ -666,6 +676,68 @@ impl UnresolvedMapping {
...
+        match &self.0 {
+            Some(version) => version >= other_version,
+            None => false,
+        }
+    }
+
+    pub fn try_from_versions<'a>(
+        versions: impl Iterator<Item = &'a Version>,
+    ) -> Result<Self, HashSet<&'a Version>> {
+        let unique_versions: HashSet<&Version> = versions.into_iter().collect();
```
leoyvens, line 697: Lets use a `BTreeSet` so that the error message is deterministic.

graph/src/data/subgraph/mod.rs
```diff
@@ -666,6 +676,68 @@ impl UnresolvedMapping {
...
+
+        let all_below_referential_version = unique_versions.iter().all(|v| *v < &API_VERSION_0_0_5);
+        let all_the_same = unique_versions.len() == 1;
+
+        let unified_version: Option<Version> = match (all_below_referential_version, all_the_same) {
+            (false, false) => return Err(unique_versions),
+            (false, true) => Some(unique_versions.iter().nth(0).unwrap().deref().clone()),
+            (true, true) => None,
+            (true, false) => None,
```
leoyvens, line 706: This match is nice, lets just collapse the last two cases as `(true, _) => None`.

graph/src/data/subgraph/mod.rs
leoyvens: @tilacog this logic is still duplicated.

leoyvens (changes requested):

graph/src/data/subgraph/mod.rs
```diff
@@ -666,6 +673,95 @@ impl UnresolvedMapping {
...
+            (Ok(a), Ok(b)) => assert_eq!(a, *b),
+            (Err(a), Err(b)) => assert_eq!(a, *b),
+            _ => panic!(),
+        }
+    }
+}
+
+#[derive(Error, Debug, PartialEq)]
+#[error("Expected a single apiVersion for mapings. Found: {}.", format_versions(.0))]
+pub struct DifferentMappingApiVersions(BTreeSet<Version>);
```
tilacog, line 747: Added this new error just for improving ergonomy when handling it.

graph/src/data/subgraph/mod.rs
```diff
@@ -666,6 +673,95 @@ impl UnresolvedMapping {
...
+        Self(versions)
+    }
+}
+
+impl From<DifferentMappingApiVersions> for SubgraphManifestValidationError {
+    fn from(versions: DifferentMappingApiVersions) -> Self {
+        SubgraphManifestValidationError::DifferentApiVersions(versions.0)
+    }
+}
+
```
tilacog, line 764: Added this new error so it is ergonomic to handling in other places.

graph/src/data/subgraph/mod.rs
```diff
@@ -666,6 +673,95 @@ impl UnresolvedMapping {
...
+        match (unified, expected_unified_version) {
+            (Ok(a), Ok(b)) => assert_eq!(a, *b),
+            (Err(a), Err(b)) => assert_eq!(a, *b),
+            _ => panic!(),
+        }
+    }
+}
+
+#[derive(Error, Debug, PartialEq)]
+#[error("Expected a single apiVersion for mapings. Found: {}.", format_versions(.0))]
```
leoyvens, line 746: s/mapings/mappings

graph/src/data/subgraph/mod.rs
leoyvens: Nice!

graph/src/data/subgraph/mod.rs
```diff
@@ -666,6 +673,95 @@ impl UnresolvedMapping {
...
+            (false, true) => Some(unique_versions.iter().nth(0).unwrap().deref().clone()),
+            (true, _) => None,
+        };
+
+        Ok(UnifiedMappingApiVersion(unified_version))
+    }
+}
+
+#[test]
+fn unified_mapping_api_version_from_iterator() {
```
leoyvens, line 710: I see you tried to keep the test close to the tested code, which makes sense but I think we usually just keep tests at the bottom of the file.

leoyvens (changes requested):
Small comments and CI is not passing.

graph/src/data/subgraph/mod.rs
```diff
@@ -814,6 +844,15 @@ impl<C: Blockchain> UnvalidatedSubgraphManifest<C> {
             errors.push(SubgraphManifestValidationError::DataSourceBlockHandlerLimitExceeded)
         }
 
+        // For API versions newer than 0.0.5, validate that all mappings uses the same api_version
+        let mappings = self.0.mappings();
+        let versions_iterator = mappings.iter().map(|mapping| &mapping.api_version);
+        if let Err(different_api_versions) =
+            UnifiedMappingApiVersion::try_from_versions(versions_iterator)
```
leoyvens, line 851: Last comment, couldn't you use `if let Err(different_api_versions) = self.0.unified_mapping_api_version()` here?

leoyvens (changes requested):

graph/src/data/subgraph/mod.rs
tilacog: I think I can. 
I'll try and use it if it works =)

tilacog: It worked! :tada:

leoyvens (approved):
Took a while to perfect this but it's great now, thanks!

PR #2585, opened 2021-06-22
block stream: Clarify comment

leoyvens:
Even I was confused about this, so very much worth documenting what's going on.


lutter (approved):
Can you copy that comment also to `NextBlocks::Revert`? That comment is also confusing

leoyvens:
@lutter done, thanks for the review.

PR #2590, opened 2021-06-24
Asyncify is_deployment_synced

leoyvens:
We noticed that if one DB shard is overloaded, that causes contention in the query nodes' non-blocking tokio threadpool. This is likely caused by a blocking DB operation being done in an async context in the query execution path. The only case I could find of this was in `throttle_while_syncing` for subscriptions, the solution was to asyncify the `is_deployment_synced` store operation.

To facilitate this change, the first commit checks the synced status for subscription throttling only when the subscription is first created. Hopefully this is ok.

### DIFF
--- a/graph/src/components/store.rs
+++ b/graph/src/components/store.rs
@@ -660,7 +660,7 @@ where
     /// on the returned stream as a single `StoreEvent`; the events are
     /// combined by using the maximum of all sources and the concatenation
     /// of the changes of the `StoreEvents` received during the interval.
-    pub fn throttle_while_syncing(
+    pub async fn throttle_while_syncing(
         self,
         logger: &Logger,
         store: Arc<dyn QueryStore>,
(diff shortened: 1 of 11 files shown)


graph/src/components/store.rs
```diff
@@ -660,23 +660,16 @@ where
...
-        static SYNC_REFRESH_FREQ: u32 = 4;
-
-        // Check whether a deployment is marked as synced in the store
-        let mut synced = store.is_deployment_synced().unwrap_or(false);
-        let synced_check_interval = interval.checked_mul(SYNC_REFRESH_FREQ).unwrap();
-        let mut synced_last_refreshed = Instant::now();
+        // Check whether a deployment is marked as synced in the store. Note that in the moment a
+        // subgraph becomes synced any existing subscriptions will continue to be throttled since
+        // this is not re-checked.
+        let synced = store.is_deployment_synced().await.unwrap_or(false);
```
lutter, line 672: We close subscriptions after a while, right? So the effect would just be that subscriptions are throttled for a bit longer than they are now, which seems fine.

lutter (approved):
Nice!

graph/src/components/store.rs
leoyvens: Also, I find querying or subscribing to a subgraph that's still syncing a very questionable thing to do in any sort of prod environment anyways.

PR #2595, opened 2021-06-25
Improve query validation robustness

leoyvens:
A few changes to improve guarding against overflows:
- Explicitly validate against cyclical fragments.
- Validate complexity before fields.
- Validate query depth even if the complexity is unlimited.
- Add a test that previously caused a stack overflow.

### DIFF
--- a/core/tests/interfaces.rs
+++ b/core/tests/interfaces.rs
@@ -1250,3 +1250,63 @@ fn enums() {
         ]}
     );
 }
+
+#[test]
+fn recursive_fragment() {
+    let subgraph_id = "RecursiveFragment";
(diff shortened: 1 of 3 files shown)


core/tests/interfaces.rs
```diff
@@ -1250,3 +1250,63 @@ fn enums() {
...
+            ...FooFrag
+            }
+        }
+
+        fragment FooFrag on Foo {
+        id
+        bar {
+            ...BarFrag
+        }
+        }
```
Jannis, line 1307: The indentation here is a little off.

Jannis (approved):
Looks good!

PR #2600, opened 2021-06-29
Revamp query semaphore to be by shard

leoyvens:
This moves the query semaphore implementation from the `LoadManager` to the per-shard (or replica) `ConnectionPool`s, with a capacity equal to the shard's (or replica's) connection pool size. It is accessed through the resolver. The indexing status queries are now also guarded by the primary shard semaphore.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,10 @@
 # NEWS
 
+## Unreleased
+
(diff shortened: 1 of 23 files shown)


lutter (approved):
Nice! Long overdue

PR #2607, opened 2021-07-01
ethereum: Fix accessing call handler transaction input

leoyvens:
For clarity this renames `AscEthereumTransaction` to `AscEthereumTransaction_0_0_1`. The actual fix is to switch `AscEthereumCall_0_0_3` to use `AscEthereumTransaction_0_0_2`.

### DIFF
--- a/chain/ethereum/src/runtime/abi.rs
+++ b/chain/ethereum/src/runtime/abi.rs
@@ -85,7 +85,7 @@ pub(crate) struct AscEthereumBlock {
 
 #[repr(C)]
 #[derive(AscType)]
-pub(crate) struct AscEthereumTransaction {
+pub(crate) struct AscEthereumTransaction_0_0_1 {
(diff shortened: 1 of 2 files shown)


chain/ethereum/src/runtime/abi.rs
```diff
@@ -146,7 +146,7 @@ pub(crate) struct AscEthereumCall_0_0_3 {
     pub to: AscPtr<AscAddress>,
     pub from: AscPtr<AscAddress>,
     pub block: AscPtr<AscEthereumBlock>,
-    pub transaction: AscPtr<AscEthereumTransaction>,
+    pub transaction: AscPtr<AscEthereumTransaction_0_0_2>,
```
otaviopace, line 149: What a catch 🦅 👀

otaviopace (approved):

PR #2617, opened 2021-07-08
ethereum: Downgrade spammy log to trace

leoyvens:
This is a nicely written log, but for some subgraphs it logs too often during normal execution. In the hosted service, this logs thousands of times per second!

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+### Logs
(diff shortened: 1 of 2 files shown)


otaviopace (approved):

otaviopace:
Should we create some kind of notice to indexers about this so they don't search the log incorrectly?

leoyvens:
@otaviopace good point, I've added a note to NEWS.md

PR #2620, opened 2021-07-09
Rename gas used to gas limit

leoyvens:
It was never the gas used, which is in the receipt not the transaction.


otaviopace (approved):

otaviopace:
Since this is a breaking change, I think we should have it documented on `NEWS.md` as well.

leoyvens:
The graph node side of this doesn't break anything.

PR #2629, opened 2021-07-15
multichain: Abstract NodeCapabilities

leoyvens:
This removes the last case in which the instance manager was constraining an associated type of `Blockchain` to something ethereum specific. Also removes `Blockchain::Manifest` since I think we'll keep the manifest as a shared structure.

### DIFF
--- a/graph/src/data/subgraph/mod.rs
+++ b/graph/src/data/subgraph/mod.rs
@@ -26,7 +26,6 @@ use crate::{blockchain::DataSourceTemplate as _, data::query::QueryExecutionErro
 use crate::{
     blockchain::{Blockchain, UnresolvedDataSource as _, UnresolvedDataSourceTemplate as _},
     components::{
-        ethereum::NodeCapabilities,
         link_resolver::LinkResolver,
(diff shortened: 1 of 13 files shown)


graph/src/data/subgraph/mod.rs
```diff
@@ -268,8 +267,6 @@ pub enum SubgraphRegistrarError {
     NameNotFound(String),
     #[error("Ethereum network not supported by registrar: {0}")]
     NetworkNotSupported(String),
-    #[error("Ethereum nodes for network {0} are missing the following capabilities: {1}")]
```
otaviopace, line 271: I understood that you moved the capabilities code to the chain/ethereum crate, but this won't happen anymore?

I'm a bit confused.

leoyvens: It really is confusing, because this code was just already dead.

otaviopace: Oh! okay 🙂

otaviopace (approved):

PR #2630, opened 2021-07-15
blockchain: Remove `DataSource::source`

leoyvens:
`Source` is an Ethereum-specific type. It's still used in more places than we'd like, but this is a start.

### DIFF
--- a/runtime/wasm/src/host_exports.rs
+++ b/runtime/wasm/src/host_exports.rs
@@ -1,6 +1,6 @@
 use crate::{error::DeterminismLevel, module::IntoTrap};
 use ethabi::param_type::Reader;
-use ethabi::{decode, encode, Address, Token};
+use ethabi::{decode, encode, Token};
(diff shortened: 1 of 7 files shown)


runtime/wasm/src/host_exports.rs
```diff
@@ -78,7 +78,7 @@ impl<C: Blockchain> HostExports<C> {
             subgraph_id,
             api_version: data_source.mapping().api_version.clone(),
             data_source_name: data_source.name().to_owned(),
-            data_source_address: data_source.source().address.clone(),
+            data_source_address: data_source.address().unwrap_or_default().to_owned(),
```
otaviopace, line 81: Why is it fine to have an empty `Vec` right here?

leoyvens: This is only used when being returned from the `data_source_address` host export, which used `unwrap_or_default`. So I figured we might as well do that earlier. But it is not supposed to be a functional change.

otaviopace (approved):

PR #2634, opened 2021-07-16
Add `is_derived` to `DirectiveFinder`

### LINKED ISSUE #2464: Create a dedicated method to check if a field is derived.
We must create a `schema_field.is_derived()` method to `DirectiveFinder` to cut down on the number of places where we hardcode the string "derivedFrom" across the codebase to detect if a field is derived or not.

_Originally posted by @lutter in https://github.com/graphprotocol/graph-node/pull/2352#discussion_r631329917_

vrmiguel:
Closes #2464 

Thanks @Xunjin for the help with this PR 😊

### DIFF
--- a/graph/src/data/graphql/ext.rs
+++ b/graph/src/data/graphql/ext.rs
@@ -250,6 +250,7 @@ impl ValueExt for Value {
 
 pub trait DirectiveFinder {
     fn find_directive(&self, name: &str) -> Option<&Directive>;
+    fn is_derived(&self) -> bool;
 }
 
 impl DirectiveFinder for ObjectType {
@@ -257,6 +258,12 @@ impl DirectiveFinder for ObjectType {
         self.directives
             .iter()
             .find(|directive| directive.name.eq(&name))
+    }
+
+    fn is_derived(&self) -> bool {
+        let is_derived = |directive: &Directive| directive.name.eq("derivedFrom");
+
+        self.directives.iter().any(is_derived)
     }
 }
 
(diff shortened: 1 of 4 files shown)


vrmiguel:
I think it'd be nice to implement some unit tests for things relating to `DirectiveFinder` but I'm having difficulties working with `ObjectType`, `Field` and `Directive`

Any help with that would be very appreciated :-)

tilacog:
I believe that an interesting test approach would involve [parsing](https://docs.rs/graphql-parser/0.3.0/graphql_parser/index.html) a simple graphql schema that contains some directives and then call the new functions over that structure, asserting that the given directives were found and associated with the expected fields.

tilacog:
LGTM

graph/src/data/graphql/ext.rs
```diff
@@ -266,10 +273,78 @@ impl DirectiveFinder for Field {
...
+mod directive_finder_tests {
+    use graphql_parser::parse_schema;
+
+    use super::*;
+
+    const SCHEMA: &str = "
+    type BuyEvent implements Event @derivedFrom(field: \"buyEvent\") {
+        id: ID!,
+        transaction: Transaction! @derivedFrom(field: \"buyEvent\")
+    }";
```
tilacog, line 306: Just minor note:
As far as I can tell, the `derivedFrom` directive is applied to fields, not to types. 
Thus, I'd suggest removing the directive from the type and focusing the test only on fields marked by it.

tilacog (approved):

PR #2645, opened 2021-07-21
Remove 3box and arweave support

leoyvens:
Nobody uses these and they are not how we want them to be.


tests/tests/parallel_tests.rs
```diff
@@ -26,7 +26,6 @@ lazy_static::lazy_static! {
 /// All integration tests subdirectories to run
 pub const INTEGRATION_TESTS_DIRECTORIES: [&str; 10] = [
     "api-version-v0-0-4",
-    // "arweave-and-3box",
```
otaviopace, line 29: Finally! 🎉 🙌

runtime/wasm/src/host.rs
```diff
@@ -26,18 +24,12 @@ lazy_static! {
         .map(Duration::from_secs);
     static ref ALLOW_NON_DETERMINISTIC_IPFS: bool =
         std::env::var("GRAPH_ALLOW_NON_DETERMINISTIC_IPFS").is_ok();
-    static ref ALLOW_NON_DETERMINISTIC_3BOX: bool =
-        std::env::var("GRAPH_ALLOW_NON_DETERMINISTIC_3BOX").is_ok();
-    static ref ALLOW_NON_DETERMINISTIC_ARWEAVE: bool =
-        std::env::var("GRAPH_ALLOW_NON_DETERMINISTIC_ARWEAVE").is_ok();
```
otaviopace, line 32: These features weren't used?

otaviopace, line 32: Shouldn't we keep this feature flags and check them alongside the `apiVersion` <= `0.0.4`?

runtime/wasm/src/module/mod.rs
```diff
@@ -506,9 +504,11 @@ impl<C: Blockchain> WasmInstance<C> {
 
         link!("log.log", log_log, level, msg_ptr);
 
-        link!("arweave.transactionData", arweave_transaction_data, ptr);
-
-        link!("box.profile", box_profile, ptr);
+        // `arweave and `box` functionality was removed, but apiVersion <= 0.0.4 must link it.
+        if api_version <= Version::new(0, 0, 4) {
```
otaviopace, line 508: Here: https://github.com/graphprotocol/graph-node/pull/2645#discussion_r674075137

runtime/wasm/src/module/mod.rs
```diff
@@ -1424,34 +1424,20 @@ impl<C: Blockchain> WasmInstanceContext<C> {
...
-        if !self.experimental_features.allow_non_deterministic_arweave {
-            return Err(HostExportError::Deterministic(anyhow!(
-                "`arweave.transactionData` is deprecated. Improved support for arweave may be added in the future"
-            )));
-        }
-        let tx_id: String = asc_get(self, tx_id)?;
-        let data = self.ctx.host_exports.arweave_transaction_data(&tx_id);
-        data.map(|data| asc_new(self, &*data).map_err(Into::into))
-            .unwrap_or(Ok(AscPtr::null()))
+        Err(HostExportError::Deterministic(anyhow!(
```
otaviopace, line 1429: We don't need to check for the features anymore?

runtime/wasm/src/module/mod.rs
```diff
@@ -1424,34 +1424,20 @@ impl<C: Blockchain> WasmInstanceContext<C> {
...
-            return Err(HostExportError::Deterministic(anyhow!(
-                "`box.profile` is deprecated. Improved support for 3box may be added in the future"
-            )));
-        }
-        let address: String = asc_get(self, address)?;
-        let profile = self.ctx.host_exports.box_profile(&address);
-        profile
-            .map(|profile| asc_new(self, &profile).map_err(|e| e.into()))
-            .unwrap_or(Ok(AscPtr::null()))
+        Err(HostExportError::Deterministic(anyhow!(
```
otaviopace, line 1439: Same as https://github.com/graphprotocol/graph-node/pull/2645/files#r674076283

runtime/wasm/src/host.rs
leoyvens: Maybe if this were widely used it would be nice to catch that, but it's not necessary.

runtime/wasm/src/module/mod.rs
leoyvens: No, the error is unconditional.

jsoneaday:
So then we can no longer create subgraphs for arweave?

azf20:
Hi @jsoneaday this was a piece of functionality that was not fully supported. We are looking at how we can properly support in future

jsoneaday:
Thank you. Arweave usage and popularity continues to grow. I hope you will support it soon.

PR #2646, opened 2021-07-21
Added tracing logs when a giving deployment is assigned/not assigned

maoueh:
While doing some development, I had a misconfiguration where my `node_id` was not in sync with my configuration file.

This led to a deployment to be immediately removed on the receiving node leading to the deployment not starting.

To alleviate that problem a bit, I added some log lines that should make it clearer about what is happening and more especially why it's happening.

I choose to go with `debug` level but willing to switch to a `trace` level if you think it would be more appropriate.

### DIFF
--- a/core/src/subgraph/registrar.rs
+++ b/core/src/subgraph/registrar.rs
@@ -171,12 +171,14 @@ where
                                     if let Some(assigned) = assigned {
                                         if assigned == node_id {
                                             // Start subgraph on this node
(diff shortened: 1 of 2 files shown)


leoyvens (approved):

leoyvens:
Looks like the build is failing.

maoueh:
I fixed the compilation error, let's see.

PR #2655, opened 2021-07-23
Added a BlockStream trait and turned existing one into PollingBlockStream impl

maoueh:
This is another small PR for upcoming Firehose feature. The BlockStream trait will enable using a PollingBlockStream (actual implementation) and an upcoming FirehoseBlockStream (next or so PR in line).

We decided on submitting a PR with just that for now to have few focused PRs around.


graph/src/blockchain/mod.rs
```diff
@@ -104,7 +106,7 @@ pub trait Blockchain: Debug + Sized + Send + Sync + 'static {
         filter: Self::TriggerFilter,
         metrics: Arc<BlockStreamMetrics>,
         unified_api_version: UnifiedMappingApiVersion,
-    ) -> Result<BlockStream<Self>, Error>;
+    ) -> Result<Box<dyn BlockStream<Self>>, Error>;
```
leoyvens, line 109: For chain-specific types we've so far chosen to use associated types rather than the trait objects, so to be consistent a `type BlockStream` should be added, and this should be `Self::BlockStream`.

maoueh: `BlockStream` is completely chain agnostic, hence why I did not create an associated type. Both of the implementation are chain agnostic (`PollingBlockStream` and the upcoming `FirehoseBlockStream`).

Personally, I do not think that chain agnostic stuff should be associated type on the chain. Tell me what you think.

leoyvens: Ok, that convinces me to go with `dyn`, at least for as long as it covers our needs for this trait.

maoueh:
Rebased and ready for approval, thanks @leoyvens!

leoyvens (approved):

Issue #2649, opened 2021-07-22
Investigate slowness during writing to the database


azf20:
Stopwatch indicates that the stage taking time is transact_block, which should not be slow.
Impact TBD.

leoyvens:
This has been identified as hosted service issue where a DB shard is overcrowded so there is contention to acquire a DB connection.

PR #2658, opened 2021-07-26
Optimize duplicate block processing error check

leoyvens:
This reduces one roundtrip to the DB when transacting a block by leveraging the `forward_block_ptr` where clause to check for duplicate block processing. Part of #2649.


lutter (approved):
Nice! I like having this check in the query much better

PR #2664, opened 2021-07-27
Update tokio 1.9

jubeless:
Attempt to upgrade tokio to 1.9 By doing so it would give us the flexibility to upgrade `ethabi` library as well as `tokio-retry` library.


leoyvens:
Closing in favor of #2679.

PR #2665, opened 2021-07-27
ethereum: Don't use cached block on `is_on_main_chain`

leoyvens:
As it turns out, the optimization in https://github.com/graphprotocol/graph-node/pull/1477 was incorrect wrt reverts.

The block stream relies on this function to check with the Ethereum node if a block is on the main chain, but it is assuming that the block being checked is final if it is past the reorg threshold. If the current subgraph head is in the cache when passed into `is_on_main_chain`, as long as it is present in the block cache it will be assumed to be final, which can lead to it not being reverted when it should be.

The most likely way for this bug to affect a subgraph is if a large reorg happens, such that by the time the last blocks are having the reorg processed by graph node they have fallen out of the threshold and are passed into `is_on_main_chain` and considered final. Since mainnet does not see large reorgs relative to the threshold this should not affect subgraphs currently on the network. In the hosted service, I have seen a Ropsten subgraph affected during a 200 block reorg, which was how I became aware of this bug.

The fix is to basically revert https://github.com/graphprotocol/graph-node/pull/1477.

### DIFF
--- a/chain/ethereum/src/ethereum_adapter.rs
+++ b/chain/ethereum/src/ethereum_adapter.rs
@@ -648,11 +648,10 @@ impl EthereumAdapter {
     pub(crate) async fn is_on_main_chain(
         &self,
         logger: &Logger,
-        chain_store: Arc<dyn ChainStore>,
         block_ptr: BlockPtr,
     ) -> Result<bool, Error> {
         let block_hash = self
-            .block_hash_by_block_number(&logger, chain_store, block_ptr.number, true)
+            .block_hash_by_block_number(&logger, block_ptr.number)
             .compat()
             .await?;
         block_hash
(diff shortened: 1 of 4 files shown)


server/http/tests/server.rs
```diff
@@ -107,7 +107,7 @@ mod test {
                         // Send an empty JSON POST request
                         let client = Client::new();
                         let request =
-                            Request::post(format!("http://localhost:8001/subgraphs/id/{}", id))
+                            Request::post(format!("http://localhost:8007/subgraphs/id/{}", id))
```
leoyvens, line 110: This is unrelated, just picked a port that would stop conflicting with graph-node defaults.

chain/ethereum/src/ethereum_adapter.rs
```diff
@@ -1123,71 +1118,29 @@ impl EthereumAdapterTrait for EthereumAdapter {
...
-        };
-        let num_hashes = hashes.len();
-        let logger1 = logger.clone();
-        let confirm_block_hash = move |hash: &Option<H256>| {
-            // If there was more than one hash, now that we know what the
-            // 'right' one is, get rid of all the others
-            if let Some(hash) = hash {
-                if block_is_final && num_hashes > 1 {
-                    chain_store
-                        .confirm_block_hash(block_number, hash)
```
lutter, line 1144: It's probably not a big deal in practice, but by never calling this, we'll hang on to forked blocks forever; duplicate blocks IIRC can lead to trouble resolving block numbers to block hashes. I don't think it's a big deal, since we don't make changes for nonfinal blocks anyway, but something to keep in mind.

lutter (approved):

chain/ethereum/src/ethereum_adapter.rs
leoyvens: I kept the `fn confirm_block_hash` around, in case we can find a different spot for it. But I suppose an incoming firehose integration means we won't have to think to hard about this.

PR #2671, opened 2021-07-30
ethereum: Update rev of the ethereum fork

leoyvens:
Updating for https://github.com/graphprotocol/rust-web3/pull/6.

### DIFF
--- a/chain/ethereum/src/network_indexer/mod.rs
+++ b/chain/ethereum/src/network_indexer/mod.rs
@@ -52,6 +52,7 @@ impl From<LightEthereumBlock> for Ommer {
             size: block.size,
             mix_hash: block.mix_hash,
             nonce: block.nonce,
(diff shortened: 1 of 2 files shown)


chain/ethereum/src/network_indexer/mod.rs
```diff
@@ -52,6 +52,7 @@ impl From<LightEthereumBlock> for Ommer {
             size: block.size,
             mix_hash: block.mix_hash,
             nonce: block.nonce,
+            base_fee_per_gas: block.base_fee_per_gas,
```
otaviopace, line 55: Dumb question: adding this doesn't affect somehow POI and/or determinism? (I imagine it doesn't)

leoyvens: it doesn't, we don't hash web3 structures directly into the PoI, only entities.

otaviopace (approved):

PR #2672, opened 2021-07-30
Make a few fields and functions public

arthas168:
We need these fields, functions and the whole test crate to be public in order to reuse as much as possible in the [unit testing framework](https://github.com/LimeChain/subtest).

Notes:
`#![allow(unused)]` is used in the tests file, because otherwise the compile treats the tokio test functions as unused. This didn't show before because the module was marked as a test module, but now it's just a normal one.

The bit below is being added to the root Cargo.toml because `cargo build --release` won't work without it. I suspect it's because now the compiler is treating the test module as a normal module.
```
[profile.release]
debug-assertions = true
```

UPDATE: It's because `graph-node` requires the user to start up the DB before running the tests. Not an issue IMO.


leoyvens:
@arthas168 please squash the commits and revert the unnecessary changes to Cargo.lock

PR #2675, opened 2021-08-02
Convert test crate to a normal module and make a few other things public

arthas168:
We need these fields, functions and the whole test crate to be public in order to reuse as much as possible in the [unit testing framework](https://github.com/LimeChain/subtest).

### DIFF
--- a/runtime/test/src/test.rs
+++ b/runtime/test/src/test.rs
@@ -29,7 +29,7 @@ use std::collections::{BTreeMap, HashMap};
 #[cfg(test)]
 use test_store::STORE;
 #[cfg(test)]
-use web3::typesH160;
+use web3::types::H160;
 
 mod abi;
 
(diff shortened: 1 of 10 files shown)


runtime/test/src/test.rs
```diff
@@ -602,11 +636,12 @@ async fn ipfs_fail_v0_0_4() {
     test_ipfs_fail(API_VERSION_0_0_4).await;
 }
 
-#[tokio::test(threaded_scheduler)]
```
leoyvens, line 605: It seems like you removed this test by removing the annotation.

leoyvens:
One alternative way to organize this is to put the tests and things that are exclusively for tests in a separate, then you can use `#[cfg(test)]` only once. Or keep the current `tests` module and put the things you need in a `common` module.

PR #2676, opened 2021-08-02
Integration-tests: Update yarn.lock to rebased branches

otaviopace:
Graph-cli and graph-ts had new commits in master but the `otavio/update-assembly-script` branch was not up to date.

This PR fixed this by pointing to the new commits on a branch with the `2` suffix, this was necessary because if the commits were overwritten in the original branch, the integration tests would start to fail, because they would point to an unexisting commit of a branch.

### DIFF
--- a/tests/integration-tests/yarn.lock
+++ b/tests/integration-tests/yarn.lock
@@ -781,7 +781,7 @@
   optionalDependencies:
     keytar "^7.4.0"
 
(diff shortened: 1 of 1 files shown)


otaviopace:
This is being closed in favor of [this PR](https://github.com/graphprotocol/graph-node/pull/2692)

PR #2678, opened 2021-08-02
Create public common module in test crate

arthas168:
We need these fields, functions and the whole test crate to be public in order to reuse as much as possible in the [unit testing framework](https://github.com/LimeChain/matchstick).

PS I know there's lots of commits 😢  If you guys can do "Squash and merge" upon merging that'll be awesome.

### DIFF
--- a/runtime/test/src/common.rs
+++ b/runtime/test/src/common.rs
@@ -1,13 +1,13 @@
 use ethabi::Contract;
+use graph::components::store::DeploymentLocator;
 use graph::data::subgraph::*;
+use graph::ipfs_client::IpfsClient;
 use graph::prelude::*;
-use graph_chain_ethereum::{DataSource, Chain, DataSourceTemplate};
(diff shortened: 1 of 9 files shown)


runtime/test/src/common.rs
```diff
@@ -0,0 +1,62 @@
+use ethabi::Contract;
+use graph::data::subgraph::*;
+use graph::prelude::*;
+use graph_chain_ethereum::DataSource;
+use semver::Version;
+use std::str::FromStr;
+use web3::types::Address;
+
+#[allow(unused)]
```
leoyvens, line 9: Is this necessary? If the function is `pub` then it can never be considered unused.

arthas168: Yep, the compiler complains that it's unused when it doesn't have that annotation

leoyvens (approved):

leoyvens:
Still failing it seems.

leoyvens (approved):

PR #2682, opened 2021-08-04
Feature Management and Validation

tilacog:
This PR adds:
1. a new subgraph manifest validation step which checks that no undeclared feature is in use;
2. a new graphQL endpoint that receives a subgraph Qm-hash and returns  a list of its detected features or a list of undeclared features.


tilacog:
@leoyvens, two things:

1. I forgot to condition feature validation to a specific `specVersion`, I'll do that now.
2. I'm not sure if we should remove/refactor any code that dealt with feature management before, such as [this](https://github.com/graphprotocol/graph-node/blob/master/graph/src/data/graphql/ext.rs#L139-L142).

leoyvens:
@tilacog Is this missing tests? What will be the testing strategy for this? I think unit tests might work fine.

tilacog:
Yes, there should be tests.

I think we could reuse the subgraphs we already have defined for integration tests here... and even writing new subgraphs to test new features as well.

leoyvens:
@tilacog What about the error message returned to graph-cli? We should make sure an informative message is being returned by the deploy json-rpc and is being printed by graph-cli.

tilacog:
> @tilacog What about the error message returned to graph-cli? We should make sure an informative message is being returned by the deploy json-rpc and is being printed by graph-cli.

I'll link a companion PR for `graph-cli` in this PR's description, and vice-versa.

chain/ethereum/src/capabilities.rs
```diff
@@ -73,7 +73,11 @@ impl_slog_value!(NodeCapabilities, "{}");
 impl graph::blockchain::NodeCapabilities<crate::Chain> for NodeCapabilities {
     fn from_data_sources(data_sources: &[DataSource]) -> Self {
         NodeCapabilities {
-            archive: data_sources.iter().any(|ds| ds.mapping.requires_archive()),
+            archive: data_sources.iter().any(|ds| {
+                ds.mapping
+                    .requires_archive()
+                    .expect("failed to parse mappings")
```
otaviopace, line 79: Is this a breaking change? Where does this error go to?

server/index-node/src/resolver.rs
```diff
@@ -149,24 +165,135 @@ where
...
+    fn resolve_subgraph_features(
+        &self,
+        arguments: &HashMap<&str, q::Value>,
+    ) -> Result<q::Value, QueryExecutionError> {
+        // We can safely unwrap because the argument is non-nullable and has been validated.
+        let subgraph_id = arguments.get_required::<String>("subgraphId").unwrap();
+
+        // TODO:
+        //
+        // An intersting optimization would involve trying to get the subgraph manifest from the
```
otaviopace, line 178: minor typo, from `intersting` to `interesting`.

Also, great comment here for the future! 👌

server/index-node/src/resolver.rs
```diff
@@ -149,24 +165,135 @@ where
...
+                        .iter()
+                        .map(ToString::to_string)
+                        .map(q::Value::String)
+                        .collect(),
+                );
+                let errors = q::Value::List(vec![]);
+                (features, errors)
+            }
+            Either::Right(errors) => {
+                assert!(!errors.is_empty());
```
otaviopace, line 252: What happens when confidence check fails? The GraphQL request receives what?

graph/src/data/subgraph/features.rs
```diff
@@ -0,0 +1,173 @@
...
+
+use crate::{
+    blockchain::Blockchain,
+    data::{graphql::DocumentExt, schema::Schema, subgraph::SubgraphManifest},
+    prelude::{Deserialize, Serialize},
+};
+use itertools::Itertools;
+use std::{collections::BTreeSet, fmt, str::FromStr};
+
+const IPFS_ON_ETHEREUM_CONTRACTS_FUNCTION_NAMES: [&'static str; 2] = ["ipfs.cat", "ipfs.map"];
```
otaviopace, line 22: I think it's useful to have a comment above this with an UUID that should be [here](https://github.com/graphprotocol/graph-node/blob/704fefc870ec2ed93a48467dab601999a0029334/runtime/wasm/src/module/mod.rs#L441) as well. 

Also informing that whenever a new `ipfs` function is added to the `host-exports`, we should add here as well.

What do you think?

graph/src/data/subgraph/features.rs
```diff
@@ -0,0 +1,173 @@
...
+fn fmt_subgraph_features(subgraph_features: &BTreeSet<SubgraphFeature>) -> String {
+    subgraph_features.iter().join(", ")
+}
+
+pub fn validate_subgraph_features<C: Blockchain>(
+    manifest: &SubgraphManifest<C>,
+) -> Result<BTreeSet<SubgraphFeature>, SubgraphFeatureValidationError> {
+    let declared: &BTreeSet<SubgraphFeature> = &manifest.features;
+    let used = detect_features(&manifest)?;
+    let undeclared: BTreeSet<SubgraphFeature> = used.difference(&declared).cloned().collect();
```
otaviopace, line 70: Great idea to use the `BTreeSet` 👏 
`difference` came to be very handy 🙂

graph/src/data/subgraph/features.rs
```diff
@@ -0,0 +1,173 @@
...
+fn detect_grafting<C: Blockchain>(manifest: &SubgraphManifest<C>) -> Option<SubgraphFeature> {
+    manifest.graft.as_ref().map(|_| SubgraphFeature::Grafting)
+}
+
+fn detect_full_text_search(schema: &Schema) -> Option<SubgraphFeature> {
+    match schema.document.get_fulltext_directives() {
+        Ok(directives) => (!directives.is_empty()).then(|| SubgraphFeature::FullTextSearch),
+
+        Err(_) => {
+            // Currently we retun an error from `get_fulltext_directives` function if the
```
otaviopace, line 112: Minor typo `retun` -> `return`

chain/ethereum/tests/manifest.rs
```diff
@@ -222,3 +228,317 @@ specVersion: 0.0.2
...
+graft:
+  base: Qmbase
+  block: 1
+";
+    test_store::run_test_sequentially(|store| async move {
+        let store = store.subgraph_store();
+        let unvalidated = resolve_unvalidated(YAML).await;
+        assert!(unvalidated
+            .validate(store.clone())
+            .expect_err("Validation must fail")
```
otaviopace, line 285: I'm a bit confused, this should return a validation error as well? 🤔

chain/ethereum/src/capabilities.rs
tilacog: This panicking behaviour already happened deep down the `requires_archive` method, we're just bubbling up the error handling here.

server/index-node/src/resolver.rs
tilacog: The `IndexNodeServer` component panics, but since it is managed by a superior process, the error message gets both logged as an `ERROR` and returned in the response, like this:

```json
{
  "errors": [
    {
      "message": "panic processing query: Something went wrong [bla bla bla]..."
    }
  ]
}
```

tilacog: With the refactor, we don't need this confidence check anymore. Thanks @otaviopace!

chain/ethereum/tests/manifest.rs
tilacog: Yes, in this case, we should get a vector with an expected GraftBaseInvalid error here, but we're really interested in discovering if there will be a FeatureValidationError alongside it.

otaviopace (approved):

PR #2696, opened 2021-08-11
Revert node: Refactor config to accept Firehose providers

leoyvens:
This reverts commit 1415d5b4742942da6a546cd9321d494d95d82500 due to backwards compatibility issues. Namely existing configurations were crashing with:
```
configuration error: missing field `transport` for key `chains`
```

### DIFF
--- a/node/src/config.rs
+++ b/node/src/config.rs
@@ -8,14 +8,11 @@ use graph::{
 use graph_chain_ethereum::NodeCapabilities;
 use graph_store_postgres::{DeploymentPlacer, Shard as ShardName, PRIMARY_SHARD};
 
-use http::{HeaderMap, Uri};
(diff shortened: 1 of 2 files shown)


lutter (approved):

maoueh:
Do we have the configuration at hand?

I'll make the necessary update, unit test it. Do you have other configs at hand, might I could check them all.

lutter:
> Do we have the configuration that failed at hand?
> 
> I'll make the necessary update, unit test it. Do you have other configs at hand, might I could check them all.

The production config is scarily big at this point, but there are two config files in `store/test-store` that would be a good starting point for such a unit test (just cut&paste/adapt them in the code). The main thing that is missing from `config.sharded.toml` is having replicas for one of the database.

PR #2697, opened 2021-08-11
store: optimize `check_interface_entity_uniqueness`

### LINKED ISSUE #2290: Optimize queries for `check_interface_entity_uniqueness`
We should figure out how to batch this check

_Originally posted by @lutter in https://github.com/graphprotocol/graph-node/pull/2286#r596227885_

vrmiguel:
Closes #2290


store/postgres/src/deployment_store.rs
```diff
@@ -288,9 +299,11 @@ impl DeploymentStore {
                 .filter(|type_name| type_name != &key.entity_type),
         );
 
+        let entity_ids: Vec<_> = keys.map(|key| &*key.entity_id).collect();
+
         if !types_with_shared_interface.is_empty() {
             if let Some(conflicting_entity) =
```
tilacog, line 305: In theory, we should be careful not to surpass PostgreSQL binding limit (which is ~64k bindings per query).
Since this query would spend only 2 bindings per conflicting ObjectType (one for the `id` array and the other for the table name), it's unlikely that this error will happen. 

If we wanted do be extra careful, we could always iterate in chunks of ~32,000 pair of values.

store/postgres/src/deployment_store.rs
```diff
@@ -256,13 +256,24 @@ impl DeploymentStore {
...
+        keys: impl Iterator<Item = &'a EntityKey> + Clone,
     ) -> Result<(), StoreError> {
+        let mut keys = keys.peekable();
+
+        // We must receive at least one key
+        let key = match keys.peek() {
+            Some(&key) => key.to_owned(),
+            None => panic!("check_interface_entity_uniqueness received no keys"),
+        };
+
```
tilacog, line 272: This function always receives at least one value, so we can expect this to panic only in case of a development mistake, almost like a debug assertion.

store/postgres/src/deployment_store.rs
```diff
@@ -256,13 +256,24 @@ impl DeploymentStore {
...
         assert_eq!(&key.subgraph_id, &layout.site.deployment);
+        debug_assert!(keys
```
tilacog, line 274: All `keys` received by this function came from a previous `group_by` operation, so we can expect them to be of the same type. Nonetheless, it's always good to keep that premise under an assertion.

store/postgres/src/deployment_store.rs
```diff
@@ -256,13 +256,24 @@ impl DeploymentStore {
...
+            .clone()
+            .all(|other_key| other_key.entity_type == key.entity_type));
```
tilacog, line 276: All `keys` received by this function came from a previous `group_by` operation, so we can expect them to be of the same type. Nonetheless, it's always good to keep that premise under an assertion.

lutter: Wouldn't it be better to check each key for `other_key.subgraph_id == layout.site.deployment`? That would also make pulling out one key unnecessary.

lutter (changes requested):

azf20:
hey @vrmiguel thanks for this - were you able to address the suggested changes?

vrmiguel:
> hey @vrmiguel thanks for this - were you able to address the suggested changes?

Hello Adam!

I'm waiting for some input to [this comment](https://github.com/graphprotocol/graph-node/pull/2697#discussion_r697737600) before taking any further action

vrmiguel:
@tilacog @azf20 @lutter 

Could you guys please have another look on this PR? :smiley:

leoyvens:
Closing because this hasn't seen progress. If work continues on this branch feel free to open a new PR.

PR #2702, opened 2021-08-13
graphman commands for interacting with POI dispute service

goaaron:
`graphman` addendum to allow an indexer to sync their data for arbitration. 

Build
---
`cargo build --bin graphman`

How to run: 
---
1. `graphman` requires that you have a valid `GRAPH_NODE_CONFIG` environment variable pointing to a configuration. You can find an example one in `graph-node/store/test-store/config.simple.toml`
> export `GRAPH_NODE_CONFIG=/path_to/graph-node/store/test-store/config.simple.toml`
2. Submit your POI to the service:
> `graphman poi dump -d $DISPUTE_ID -i $INDEXER_ID -s $SUBGRAPH_DEPLOYMENT  -h $DISPUTE_SERVICE_HOST -k`

This will only work if the service recognizes the dispute as accepting POI and that the indexer can contribute to said dispute. 

3. Once analysis has been completed on the POI you will be able to sync your entities
> `graphman poi sync-d $DISPUTE_ID -i $INDEXER_ID -s $SUBGRAPH_DEPLOYMENT  -h $DISPUTE_SERVICE_HOST   -k`

This will only work if the service recognizes the dispute as accepting entities and that the indexer can contribute to said dispute. 

Enhancements:
---

- [ ] Remove flag for `indexer_id` and auto-populate by signing with graph-node
- [ ] Expose an endpoint in the graph-node server that can invoke the functions in this PR when called by an arbitrator. 
- [ ] Allow graph-node to process tabular data without need of the service by using  something like https://github.com/pola-rs/polars
- [ ] Instead of uploading the whole POI table, use a binary search on block range to compare to other indexers to find the earliest mismatch. Requires exposing an endpoint that indexers can use to query other indexers who are indexing a subgraph deployment.


lutter:
I think this is missing files in `node/src/manager/commands`.

For the command line, it would be better to call this subcommand just `poi` so that you can write `graphman poi dump ...` and `graphman poi sync ...`. Since what the commands do isn't entirely obvious, it would be good to have more code comments/help text that explains to users what these things do.

node/src/manager/commands/poi.rs
```diff
@@ -0,0 +1,655 @@
...
+    #[sql_type = "Text"]
+    data_type: String,
+}
+
+/// Gathers entries from entity tables that contain the divergent block if in closed interval
+/// Diesel isn't of any use here because this is dynamic.
+/// Also need to catch the columns for passing to csv.
+///
+/// Note: there may be no return values for a table filtered on a block. In this case, eagerly return
+/// Note: map null values to empty string;
```
lutter, line 557: This replicates a lot of the work that things like the `Layout` and `relational_queries` in the store do already and in a way that's very dependent on the details of how data is stored which shouldn't leak outside of the `store` crate.

I would much prefer if we added a new variant to `EntityFilter`, say `ChangedAtBlock(BlockNumber)` so that you can just use `QueryStore::find_query_values` or `SubgraphStore::find` to get the entities you want. Having an official way to get the entities that were modified at a specific block would be generally useful.

You can find all the entity types by getting the subgraph's schema with `SubgraphStore::input_schema` and then using `DocumentExt::get_object_type_definitions()` on it. You'd then construct one `EntityQuery` for each object type and get back a list of entities.

node/src/manager/commands/poi.rs
```diff
@@ -0,0 +1,655 @@
...
+       ORDER by ordinal_position",
+        schema_name, table_name
+    );
+    // let columns = raw_client.query(column_query, &[&schema_name, &table_name])?;
+
+    let columns = sql_query(column_query)
+        .get_results::<ColumnNameAndType>(conn)
+        .unwrap();
+
+    // Reserved keywords in SQL appear as subgraph tables unfortunately. I circumvent this with appending underscores
```
lutter, line 581: The right way to get around that is to enclose names in double quotes, like `select 1 from deployment_schemas as "table"`

node/src/manager/commands/poi.rs
```diff
@@ -0,0 +1,655 @@
...
+    let entity_query = format!("select * from {}.{} as {}__ WHERE ({}::int4  <@ {}__.block_range AND upper_inc({}__.block_range))\
+OR ({}::int4 = lower({}__.block_range))",schema_name, table_name, table_name,divergent_block, table_name, table_name,divergent_block,table_name);
```
lutter, line 583: `block_range` is never inclusive. This therefore boils down to `lower(block_range) = $n` which will get you all entities that were created or updated at block `$n`. If you also need to find which entities were deleted at block `$n`, you'd also need to query with `upper(block_range) = $n`. I would make that something like `EntityFilter::ClampedAtBlock(BlockNumber)` (we refer to restricting the range of an entity version as 'clamping')

node/src/manager/commands/poi.rs
```diff
@@ -0,0 +1,655 @@
...
+}
+
+#[derive(Queryable, QueryableByName)]
+pub struct PoiRecord {
+    #[sql_type = "Text"]
+    digest: String,
+    #[sql_type = "Text"]
+    id: String,
+    #[sql_type = "BigInt"]
+    vid: i64,
```
lutter, line 325: The `vid` is generally meaningless outside of a specific `graph-node` installations, and can depend on things like how many/which reverts that `graph-node` has seen. It's not useful to report that externally or for comparison.

lutter (changes requested):
I think this will be a really useful tool for indexers. It does need some bigger revisions in how it uses the store to fit better with the rest of the code and to make it less vulnerable to changes in how exactly data is stored.

PR #2708, opened 2021-08-16
Use fully qualified table names when fetching transaction receipts

tilacog:
This PR:
- refiles all `transaction_receipt` items that interfaces with PostgreSQL under the `store` crate;
- uses the `public.ethereum_blocks` instead of `public.blocks` when dealing with shared storage

Fixes: #2712


store/postgres/src/chain_store.rs
```diff
@@ -364,6 +363,14 @@ mod data {
...
         }
 
+        /// Returns a fully qualified table name to the blocks table
+        pub(super) fn blocks_table(&self) -> String {
+            match self {
+                Storage::Shared => String::from("public.ethereum_blocks"),
+                Storage::Private(Schema { name, .. }) => format!("{}.blocks", name),
+            }
+        }
+
```
tilacog, line 373: This is the proposed change to fix the improper access to a `public.blocks` table.

store/postgres/src/chain_store.rs
```diff
@@ -1439,10 +1446,11 @@ impl ChainStoreTrait for ChainStore {
     ) -> Result<Vec<LightTransactionReceipt>, StoreError> {
         let pool = self.pool.clone();
         let storage = self.storage.clone();
-        let block_hash = block_hash.clone();
+        let blocks_table_path: String = storage.blocks_table();
```
lutter, line 1449: Since you are passing this to a function on `storage`, it doesn't need to be an argument on `storage.find_transaction_receipts_in_block` and that method can call it on `self`. That way, `blocks_table()` can be a private method on `Storage`

store/postgres/src/chain_store.rs
```diff
@@ -364,6 +363,14 @@ mod data {
             }
         }
 
+        /// Returns a fully qualified table name to the blocks table
+        pub(super) fn blocks_table(&self) -> String {
+            match self {
+                Storage::Shared => String::from("public.ethereum_blocks"),
+                Storage::Private(Schema { name, .. }) => format!("{}.blocks", name),
```
lutter, line 370: Make this
```
        fn blocks_table(&self) -> &str {
            match self {
                Storage::Shared => "public.ethereum_blocks",
                Storage::Private(Schema { blocks, .. }) => &blocks.qname,
            }
        }
```

store/postgres/src/transaction_receipt.rs
```diff
@@ -0,0 +1,166 @@
...
+        .into_iter()
+        .map(LightTransactionReceipt::try_from)
+        .collect()
+}
+
+/// Parameters for querying for all transaction receipts of a given block.
+struct TransactionReceiptQuery<'a> {
+    block_hash: &'a [u8],
+    blocks_table_name: &'a str,
+}
```
lutter, line 42: Since this query is very simple, it would have been enough to just use `sql_query` for this; but since it's there, it's fine to leave it as is. My rule of thumb is that if a query can be easily written as one string (maybe with using `format!` to get a table name or similar in there) I use `sql_query`, and only when query generation gets more complex than that do I resort to the whole `QueryFragment` dance.

lutter (changes requested):

store/postgres/Cargo.toml
```diff
@@ -34,6 +34,9 @@ anyhow = "1.0.41"
 git-testament = "0.2.0"
 itertools = "0.10.1"
 
+# Our fork contains a small but hacky patch.
+web3 = { git = "https://github.com/graphprotocol/rust-web3", branch = "master" }
```
lutter, line 38: I missed that the first time around - we try to import common dependencies from the `graph` crate so that we can update them in lockstep across crates. If you `use graph::prelude::web3::H256` etc. in your code, you don't need this additional dependency in the `store` crate.

lutter (approved):
Just fix the dependency on `web3` and then this is ready to go.

PR #2716, opened 2021-08-17
core,chain,graph,node,store: added `FirehoseBlockStream` implementation of `BlockStream`

maoueh:
This PR adds a new `BlockStream` implementation called `FirehoseBlockStream` that is a built-in alternative for the existing `PollingBlockStream`. If your config has a provider of `type = firehose` (needs to use a new format for the provider definitions) **and** your started the `graph-node` with the environment variable `FIREHOSE_PRIORITY=true`, when a subgraph is syncing, a `FirehoseBlockStream` will be used to feed blocks to the subgraph deployed.

Left:
- [X] Ensure best practices around gRPC connection sharing (should be good but if someone has experience with it, you be great to get feedback on that part)
- [X] ~More integration tests and behaviour validation (dynamic data sources is one of them)~ Full testing will be an on going effort
- [X] ~Decide where to host generated code for Protobuf & gRPC services (we have dedicated package(s) in Golang, I think it could be a similar approach here, others could depend on that and not pull the full blown `graph-node` or have the need to regenerate codes, will see)~ Decision was to make a copy of the protocol buffer definitions for now and sync when needed from the canonical repositories (github.com/streamingfast/proto & github.com/streamingfast/proto-ethereum).
- [X] ~Check how `FirehoseBlockStream` should be enabled in the future (by default if present, keep and env var, etc.)~ For now, as soon as Firehose is defined, it takes precedences. 
- [x] Review & addressing feedback (*In progress*)
- [X] ~Documentation~ Will come later when NEAR is integrated


chain/ethereum/src/chain.rs
```diff
@@ -169,57 +292,33 @@ impl Blockchain for Chain {
...
-        // Special case: Detect Celo and set the threshold to 0, so that eth_getLogs is always used.
-        // This is ok because Celo blocks are always final. And we _need_ to do this because
-        // some events appear only in eth_getLogs but not in transaction receipts.
-        // See also ca0edc58-0ec5-4c89-a7dd-2241797f5e50.
-        let chain_id = self.eth_adapters.cheapest().unwrap().chain_id().await?;
-        let reorg_threshold = match CELO_CHAIN_IDS.contains(&chain_id) {
-            false => self.reorg_threshold,
-            true => 0,
-        };
+        let favor_firehose = env::var("FIREHOSE_PRIORITY");
```
lutter, line 295: It would be better to map to a boolean right here to make the next line a little clearer.

chain/ethereum/src/chain.rs
```diff
@@ -169,57 +292,33 @@ impl Blockchain for Chain {
...
+            return self
+                .new_firehose_block_stream(
+                    deployment,
+                    start_blocks,
+                    filter,
+                    metrics,
+                    unified_api_version,
+                )
+                .await;
+        }
```
lutter, line 306: This would be clearer as a `if .. { .. } else { .. }` without the `return` statement

chain/ethereum/src/chain.rs
```diff
@@ -443,6 +542,86 @@ impl TriggersAdapterTrait<Chain> for TriggersAdapter {
...
+            bstream::ForkStep::StepUndo => Ok(BlockStreamEvent::Revert(
+                BlockPtr {
+                    hash: BlockHash::from(block.hash),
+                    number: block.number as i32,
+                },
+                Cursor::Some(response.cursor.clone()),
+            )),
+
+            bstream::ForkStep::StepIrreversible => {
+                panic!("irreversible step is not handled and should not be requested in the Firehose request")
```
lutter, line 615: We usually use `unreachable!` for 'should not happen' type errors.

chain/ethereum/src/lib.rs
```diff
@@ -6,6 +6,9 @@ pub mod network_indexer;
 pub mod runtime;
 mod transport;
 
+// StreamingFast components
+pub mod sf;
```
lutter, line 10: Feel free to put your code right into the corresponding crates/modules. No need to namespace it like this from my POV.

chain/ethereum/src/sf/mod.rs
```diff
@@ -0,0 +1,277 @@
+pub mod pb {
```
lutter, line 1: I am a big fan of short names, but this might be a little too short; maybe call it `protobuf`?

graph/src/sf/endpoints.rs
```diff
@@ -0,0 +1,157 @@
...
+        url: S,
+        token: Option<String>,
+    ) -> Result<Self, anyhow::Error> {
+        let uri = url
+            .as_ref()
+            .parse::<Uri>()
+            .expect("the url should have been validated by now, so it is a valid Uri");
+
+        let endpoint;
+        match uri.scheme().unwrap().as_str() {
```
lutter, line 43: This should be `let endpoint = match uri.scheme.unwrap().as_str() {  .. }`

graph/src/components/store.rs
```diff
@@ -962,6 +962,9 @@ pub trait SubgraphStore: Send + Sync + 'static {
         deployment: &DeploymentLocator,
     ) -> Result<Arc<dyn WritableStore>, StoreError>;
 
+    /// Return a `CursorStore` that is used
+    fn cursor(&self, deployment: &DeploymentLocator) -> Result<Arc<dyn CursorStore>, StoreError>;
```
lutter, line 966: I'd be fine with just returning the cursor from this without going through a separate `CursorStore` unless  you think that `CursorStore` will end up getting passed around a lot and we should therefore not expose all of the subgraph store to those places. But if it's only used in places that already have a `SubgraphStore`, there's no need for the extra indirection.

graph/src/blockchain/firehose_block_stream.rs
```diff
@@ -0,0 +1,252 @@
...
+        tokio::spawn(async move {
+            tokio::time::sleep(wait_duration).await;
+            waker.wake();
+        });
+
+        Poll::Pending
+    }
+}
+
+fn wait_duration_secs(attempt_number: u64) -> u64 {
```
lutter, line 246: It's better to return a `Duration` here, then there's no confusion what unit is used for the return value

graph/src/blockchain/firehose_block_stream.rs
```diff
@@ -0,0 +1,252 @@
...
+                        "Blockstream disconnected, (re-)connecting"; "endpoint uri" => format_args!("{}", self.endpoint),
+                    );
+
+                    let start_block_num: BlockNumber = self
+                        .ctx
+                        .start_blocks
+                        .clone()
+                        .into_iter()
+                        .min()
+                        .unwrap_or(BLOCK_NUMBER_MAX);
```
lutter, line 111: Is this really meant to be `BLOCK_NUMBER_MAX`? It seems it should default to `0`. If that's intended, might be good to comment why `BLOCK_NUMBER_MAX` is used here.

lutter (commented):
Not a full review, but I really like where this is going. I just left some mostly nitpicky/style comments.

As for the todos you list:
* where do you anticipate the protobuf definitions will be needed outside of `graph-node`? It might be enough to figure out how to make them more shareable when there is a need to use them somewhere else than `graph-node`
* I think it would be fine to use a firehose provider by default when it is configured without the additional `FIREHOSE_PRIORITY` environment variable

core/src/subgraph/instance_manager.rs
```diff
@@ -480,69 +479,69 @@ where
...
-                Some(Ok(BlockStreamEvent::ProcessBlock(block))) => block,
-                Some(Ok(BlockStreamEvent::Revert(subgraph_ptr))) => {
-                    info!(
-                        logger,
-                        "Reverting block to get back to main chain";
-                        "block_number" => format!("{}", subgraph_ptr.number),
-                        "block_hash" => format!("{}", subgraph_ptr.hash)
-                    );
+            let (block, cursor) = match block_stream.next().await {
+                Some(res) => match res {
```
leoyvens, line 483: I see there was a change from `Some(Ok(...))` to `Some(res) => match res {`. Regardless of the style preference, mixing that into this PR makes the diff harder to read, so I'd prefer if we kept the previous style to minimize the diff.

core/src/subgraph/instance_manager.rs
```diff
@@ -480,69 +479,69 @@ where
...
-                    // will be broader than necessary. This is not ideal for performance, but is not
-                    // incorrect since we will discard triggers that match the filters but do not
-                    // match any data sources.
-                    ctx.state.instance.revert_data_sources(subgraph_ptr.number);
-                    ctx.state.entity_lfu_cache = LfuCache::new();
-                    continue;
-                }
-                // Log and drop the errors from the block_stream
-                // The block stream will continue attempting to produce blocks
-                Some(Err(e)) => {
```
leoyvens, line 538: What was wrong with this comment? Could it be updated?

graph/Cargo.toml
```diff
@@ -66,3 +69,6 @@ web3 = { git = "https://github.com/graphprotocol/rust-web3", branch = "master" }
 test-store = { path = "../store/test-store" }
 maplit = "1.0.2"
 structopt = { version = "0.3" }
+
+[build-dependencies]
+tonic-build = "0.5.1"
```
leoyvens, line 74: Add newline at the end of the file.

graph/src/blockchain/block_stream.rs
```diff
@@ -1,16 +1,22 @@
...
 use futures03::Stream;
+use std::sync::Arc;
+use thiserror::Error;
 
 use super::{Block, BlockPtr, Blockchain};
 use crate::components::store::BlockNumber;
+use crate::sf::bstream;
 use crate::{prelude::*, prometheus::labels};
 
+#[cfg(debug_assertions)]
```
leoyvens, line 11: How come `trait BlockStream` isn't used in release builds?

graph/src/blockchain/firehose_block_stream.rs
```diff
@@ -0,0 +1,252 @@
...
+        }
+    }
+}
+
+impl<C: Blockchain> BlockStream<C> for FirehoseBlockStream<C> {}
+
+impl<C: Blockchain> Stream for FirehoseBlockStream<C> {
+    type Item = Result<BlockStreamEvent<C>, Error>;
+
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
```
leoyvens, line 96: I find that reasoning about `poll` implementations is harder than reasoning about async/await and stream combinators, and the manual `Stream` implementation is one thing I dislike about the current block stream. I see you already hit some complications here such as in `schedule_error_retry`.

I'd suggest trying to write this stream without a manual implementation. [try_unfold](https://docs.rs/futures/0.3.17/futures/stream/fn.try_unfold.html) might be a good way to implement a state machine with async/await instead of messing with `Poll`.

graph/src/blockchain/polling_block_stream.rs
```diff
@@ -580,15 +588,18 @@ impl<C: Blockchain> Stream for PollingBlockStream<C> {
...
+                BlockStreamState::RetryAfterDelay(ref mut delay) => {
+                    match delay.as_mut().poll(cx) {
+                        Poll::Ready(Ok(..)) | Poll::Ready(Err(_)) => {
+                            self.state = BlockStreamState::BeginReconciliation;
+                        }
 
-                    Poll::Pending => {
-                        break Poll::Pending;
+                        Poll::Pending => {
+                            // self.state = BlockStreamState::RetryAfterDelay(delay);
```
leoyvens, line 598: commented out code

leoyvens (changes requested):

graph/src/blockchain/firehose_block_stream.rs
leoyvens: This can be left as a to-do for future PRs.

chain/ethereum/proto/codec.proto
```diff
@@ -0,0 +1,423 @@
...
+}
+
+// BlockWithRefs is a lightweight block, with traces and transactions
+// purged from the `block` within, and only.  It is used in transports
+// to pass block data around.
+message BlockWithRefs {
+  string id = 1;
+  Block block = 2;
+  TransactionRefs transaction_trace_refs = 3;
+  bool irreversible = 4;
```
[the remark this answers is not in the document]
maoueh: Internally, StreamingFast systems for Ethereum uses -200 as the irreversible block. We added some arguments that enable tweaking the irreversible block (it cannot be passed 200 since in the backend, block after the 200 delta are not tracked anymore).

Not sure neither, that will need a live discussion. NEAR will be different has the chain has the concept of final block.

chain/ethereum/Cargo.toml
```diff
@@ -17,6 +17,14 @@ dirs-next = "2.0"
...
 tiny-keccak = "1.5.0"
 hex = "0.4.3"
+tonic = "0.5.1"
+prost = "0.8.0"
+prost-types = "0.8.0"
+
+futures03 = { version = "0.3.1", package = "futures", features = ["compat"] }
+tokio = { version = "1.9.0", features = ["full"] }
+tokio-stream = { version = "0.1.6", features = ["sync"] }
+tokio-retry = "0.3.0"
```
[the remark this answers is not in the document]
maoueh: Is there a good benefit in compile time by doing this? Does it solve some versioning inconsistencies issue?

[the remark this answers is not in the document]
maoueh: Protobuf generated code use fully qualified `::prost` statements, it's not possible to have it from prelude (at least with my knowledge of Rust modules, is there a trick?).

I've refactored all other but `prost` is still in the crate Cargo file.

chain/ethereum/src/chain.rs
```diff
@@ -443,6 +542,86 @@ impl TriggersAdapterTrait<Chain> for TriggersAdapter {
     }
 }
 
+pub struct FirehoseMapper {
+    logger: Logger,
+}
+
+impl FirehoseMapper {
+    fn triggers_in_block(
```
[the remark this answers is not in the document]
maoueh: Refactored, removed this `FirehoseMapper` and the new one simply convert to `C::Block` than use the standard `C::TriggersAdapter` to extract the triggers from.

graph/src/blockchain/firehose_block_stream.rs
maoueh: I totally agree, we found it really hard also and were trying to look at alternative. The `try_unfold` seems pretty good, I'll take a look and revisit later.

maoueh: How one can use the combinator like `try_unfold` and "implement" the `Stream<Item = ...>` trait?

Like I have my struct that right now implement the `poll` method, what is the recipe so that I still implement the trait but use `try_unfold` as the implementation?

```
impl<C: Blockchain, F: FirehoseMapper<C>> Stream for FirehoseBlockStream<C, F> {
   ...?
}
```

Is the trick to actual create a stream "object" using `try_unfold` and then use that to return the `poll` value?

chain/ethereum/src/chain.rs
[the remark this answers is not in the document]
maoueh: I had to back off a bit. `C::TriggersAdapter#triggers_in_block` and I would have used it in my `poll_next` implementation. It was really hard to overcome and I didn't want to loose too much time on that for now, as I think a stream combinator could make the code much easier. I left a FIXME comment

maoueh:
All comments addressed, let's see how it going to look like in the CI! @lutter if all pass, it would be ready for another quick round to see if it can be approved.

leoyvens:
Lets make sure 0.24.1 is released before this is merged.

maoueh:
@leoyvens Perfect I'll wait the release.

Jannis:
With 0.24.1 out the door, are we ready to merge this or are there any blockers?

leoyvens (approved):

chain/ethereum/examples/firehose.rs
```diff
@@ -0,0 +1,74 @@
...
+use std::sync::Arc;
+use tonic::Streaming;
+
+#[tokio::main]
+async fn main() -> Result<(), Error> {
+    let mut cursor: Option<String> = None;
+
+    let logger = logger(true);
+    let firehose = Arc::new(
+        FirehoseEndpoint::new(logger, "firehose", "https://bsc.streamingfast.io:443", None).await?,
```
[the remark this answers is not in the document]
maoueh: For an example, I feel like some sane defaults is important, should be easy and fast to run. I'm all good adding a way to override via an args.

I'm even thinking about moving this helper/example to graphman to easily stream out a range of blocks from the firehose, would be useful for debugging purposes. What do you think? If you agree, I would leave it like that and move the code later into `graphman` directly.

store/postgres/src/deployment.rs
```diff
@@ -264,6 +265,35 @@ pub fn forward_block_ptr(
...
+}
+
+pub fn update_firehose_cursor(
+    conn: &PgConnection,
+    id: &DeploymentHash,
+    cursor: Option<String>,
+) -> Result<(), StoreError> {
+    use subgraph_deployment as d;
+
+    let cursor = cursor.unwrap_or(String::from(""));
```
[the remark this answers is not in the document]
maoueh: I agree, will change.

store/postgres/src/deployment_store.rs
```diff
@@ -680,6 +690,11 @@ impl DeploymentStore {
         Self::block_ptr_with_conn(&site.deployment, &conn)
     }
 
+    pub(crate) fn block_cursor(&self, site: &Site) -> Result<Option<String>, StoreError> {
+        let conn = self.get_conn()?;
+        Self::block_cursor_with_conn(&site.deployment, &conn)
```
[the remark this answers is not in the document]
maoueh: Perfect, thanks for the info.

maoueh:
Rebase and resolved last comments, once the CI passes, I'll merge.

leoyvens:
@maoueh Try disabling lld to see if that helps CI.

leoyvens:
Or you can try setting `CARGO_INCREMENTAL=0` in `.github/workflows/ci.yml`

maoueh:
I'll test first with lld disabled to see and try again with disabled incremental build if it does not work.

leoyvens:
@maoueh didn't work, lets try keeping lld and disabling incremental compilation.

maoueh:
Incremental fixed it, hurray! Merging.

PR #2719, opened 2021-08-18
runtime: Add env var for maximum stack size

otaviopace:
Just an environment variable to set the maximum stack size for the WASM runtime.
If not set, the default is 512KiB, half of what it was before (default from `wasmtime` crate was 1MiB).

### DIFF
--- a/runtime/wasm/src/mapping.rs
+++ b/runtime/wasm/src/mapping.rs
@@ -8,9 +8,18 @@ use std::collections::BTreeMap;
 use std::sync::Arc;
 use std::thread;
 
+const ONE_MIB: usize = 1 << 20; // 1_048_576
+
 lazy_static! {
     /// Verbose logging of mapping inputs
     pub static ref LOG_TRIGGER_DATA: bool = std::env::var("GRAPH_LOG_TRIGGER_DATA").is_ok();
+
+    /// Maximum stack size for the WASM runtime
+    pub static ref MAX_STACK_SIZE: usize = std::env::var("GRAPH_RUNTIME_MAX_STACK_SIZE")
+        .ok()
+        .and_then(|max_stack_size| max_stack_size.parse().ok())
+        // 512KiB
+        .unwrap_or(ONE_MIB / 2);
 }
 
 /// Spawn a wasm module in its own thread.
@@ -140,6 +149,8 @@ impl ValidModule {
         config.interruptable(true); // For timeouts.
         config.cranelift_nan_canonicalization(true); // For NaN determinism.
         config.cranelift_opt_level(wasmtime::OptLevel::None);
+        config.max_wasm_stack(*MAX_STACK_SIZE).unwrap(); // Safe because this only panics if size passed is 0.
+
         let engine = &wasmtime::Engine::new(&config)?;
         let module = wasmtime::Module::from_binary(&engine, raw_module)?;
 
--- a/docs/environment-variables.md
+++ b/docs/environment-variables.md
@@ -75,6 +75,10 @@ those.
    corresponds to 1GB.
 - `GRAPH_QUERY_CACHE_STALE_PERIOD`: Number of queries after which a cache
   entry can be considered stale. Defaults to 100.
+- `GRAPH_MAX_API_VERSION`: Maximum `apiVersion` supported, if a developer tries to create a subgraph
+  with a higher `apiVersion` than this in their mappings, they'll receive an error. Defaults to `0.0.5`.
+- `GRAPH_RUNTIME_MAX_STACK_SIZE`: Maximum stack size for the WASM runtime, if exceeded the execution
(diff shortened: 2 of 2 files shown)


runtime/wasm/src/mapping.rs
```diff
@@ -8,9 +8,20 @@ use std::collections::BTreeMap;
 use std::sync::Arc;
 use std::thread;
 
+const ONE_MEGA_BYTE: usize = 1 << 20; // 1_048_576
```
leoyvens, line 11: Lets call this `ONE_MIB`.

runtime/wasm/src/mapping.rs
```diff
@@ -8,9 +8,20 @@ use std::collections::BTreeMap;
...
     pub static ref LOG_TRIGGER_DATA: bool = std::env::var("GRAPH_LOG_TRIGGER_DATA").is_ok();
+
+    /// Maximum stack size for the WASM runtime
+    pub static ref MAX_STACK_SIZE: usize = std::env::var("GRAPH_RUNTIME_MAX_STACK_SIZE")
+        .ok()
+        .and_then(|max_stack_size| max_stack_size.parse().ok())
+        // 1 MiB is wasmtime default
+        // https://docs.rs/wasmtime/0.27.0/wasmtime/struct.Config.html#method.max_wasm_stack
+        // https://docs.rs/wasmtime/0.27.0/src/wasmtime/config.rs.html#419
+        .unwrap_or(ONE_MEGA_BYTE);
```
leoyvens, line 24: Lets default to 512 MiB since that's what we intend to try on hosted.

leoyvens (changes requested):
This should be documented in https://github.com/graphprotocol/graph-node/blob/master/docs/environment-variables.md.

runtime/wasm/src/mapping.rs
otaviopace: Is it safe for us to change the default for **everyone**?

otaviopace:
@leoyvens should I put it in the [running mapping handlers](https://github.com/graphprotocol/graph-node/blob/master/docs/environment-variables.md#running-mapping-handlers) section, or create another one about the `runtime`?

runtime/wasm/src/mapping.rs
otaviopace: Done!

runtime/wasm/src/mapping.rs
```diff
@@ -140,6 +149,8 @@ impl ValidModule {
         config.interruptable(true); // For timeouts.
         config.cranelift_nan_canonicalization(true); // For NaN determinism.
         config.cranelift_opt_level(wasmtime::OptLevel::None);
+        config.max_wasm_stack(*MAX_STACK_SIZE)?;
```
otaviopace, line 152: Do you think it makes a difference to `unwrap` right here? The only error this can return is [this one](https://docs.rs/wasmtime/0.27.0/src/wasmtime/config.rs.html#637), which is when the `size` parameter is `0`.

leoyvens: This could be unwrapped.

runtime/wasm/src/mapping.rs
```diff
@@ -8,9 +8,18 @@ use std::collections::BTreeMap;
...
 lazy_static! {
     /// Verbose logging of mapping inputs
     pub static ref LOG_TRIGGER_DATA: bool = std::env::var("GRAPH_LOG_TRIGGER_DATA").is_ok();
+
+    /// Maximum stack size for the WASM runtime
+    pub static ref MAX_STACK_SIZE: usize = std::env::var("GRAPH_RUNTIME_MAX_STACK_SIZE")
+        .ok()
+        .and_then(|max_stack_size| max_stack_size.parse().ok())
+        // 512KiB
+        .unwrap_or((ONE_MIB as f32 * 0.5) as usize);
```
leoyvens, line 22: What's wrong with division by 2?

otaviopace: Nothing wrong, for some reason I felt like `0.5` was easier to associate with `512KiB`.

Also I think it's easier to change if we want something like `800KiB`.

leoyvens: Not that it matters here, but integer arithmetic is much easier to reason about than float arithmetic, if we need to change this to something else we can introduce new constants.

otaviopace: Fixed!

runtime/wasm/src/mapping.rs
otaviopace: Unwrapped!

runtime/wasm/src/mapping.rs
otaviopace: We discussed on slack that we'll test on the hosted service first (before doing another release).

leoyvens (approved):

PR #2723, opened 2021-08-19
Chain head watcher fallback

### LINKED ISSUE #2714: Subgraphs don't receive notifications of chain head updates so stop indexing
Bug

Sometimes the block stream of synced subgraphs stop getting notifications of chain head updates, so stop indexing. This has been operationally resolved by restarting the indexing nodes. This may be related to the recent tokio upgrade.

The block stream should continue to get notifications of chain head updates.
As an interim fix, subgraphs could continue to poll for new blocks, even in the absence of a notification

leoyvens:
A re-implementation of #2378, since we've again seen the chain head update stream have issues in production in heavily loaded nodes, post the tokio upgrade. Generally it seems like a good idea to be robust to the listener going down, though this will likely work differently after the block stream is pipelined. Resolves #2714.

### DIFF
--- a/graph/src/blockchain/mod.rs
+++ b/graph/src/blockchain/mod.rs
@@ -36,10 +36,7 @@ use std::{collections::BTreeMap, fmt::Debug};
 use std::{collections::HashMap, convert::TryFrom};
 use web3::types::H256;
 
-pub use block_stream::{
-    BlockStream, BlockStreamMetrics, ChainHeadUpdateListener, ChainHeadUpdateStream,
-    TriggersAdapter,
-};
+pub use block_stream::{BlockStream, BlockStreamMetrics, ChainHeadUpdateListener, TriggersAdapter};
 pub use polling_block_stream::PollingBlockStream;
 pub use types::{BlockHash, BlockPtr};
 
--- a/graph/src/blockchain/polling_block_stream.rs
+++ b/graph/src/blockchain/polling_block_stream.rs
@@ -1,5 +1,4 @@
 use super::block_stream::{BlockStreamEvent, BlockWithTriggers, TriggersAdapter};
-use super::ChainHeadUpdateStream;
 use super::{BlockPtr, Blockchain};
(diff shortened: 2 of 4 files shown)


graph/src/blockchain/mod.rs
```diff
@@ -36,10 +36,7 @@ use std::{collections::BTreeMap, fmt::Debug};
 use std::{collections::HashMap, convert::TryFrom};
 use web3::types::H256;
 
-pub use block_stream::{
-    BlockStream, BlockStreamMetrics, ChainHeadUpdateListener, ChainHeadUpdateStream,
-    TriggersAdapter,
-};
+pub use block_stream::{BlockStream, BlockStreamMetrics, ChainHeadUpdateListener, TriggersAdapter};
```
maoueh, line 39: Do I have some ide/rustfmt misconfiguration?

leoyvens: This wasn't just a reformat, the `ChainHeadUpdateStream` import was removed. And rustfmt runs on CI so you never have to worry about that.

graph/src/blockchain/polling_block_stream.rs
```diff
@@ -531,11 +544,35 @@ impl<C: Blockchain> Stream for PollingBlockStream<C> {
...
-                            self.state = BlockStreamState::Idle;
-
-                            // Poll for chain head update
-                            continue;
+                            // Switch to idle. When idle, wait for a chain head update notification,
+                            // or as a fallback poll for an update after a timeout.
+                            let mut watcher = self.chain_head_update_watcher.clone();
+                            let watcher_fut: Pin<Box<dyn Future<Output = _> + Send>> = async move {
+                                watcher
+                                    .changed()
```
leoyvens, line 552: On further testing, because this watcher will always have an unseen change since the original `chain_head_update_watcher` wasn't checked for the change. This is a deadly flaw in the PR since I don't see a good way to solve this, I'll rework this fallback with a different approach.

PR #2725, opened 2021-08-20
chain head listener: Ensure periodic updates

### LINKED ISSUE #2714: Subgraphs don't receive notifications of chain head updates so stop indexing

leoyvens:
Chain head listener fallback, take 3. #2723 had a serious problem and this approach seems cleaner, since it doesn't touch the block stream. This also enables the `parking_lot` feature in tokio, so that we use fair locks in more places in an attempt to prevent deadlocks under load.

Resolves #2714.


otaviopace (approved):
💯

tilacog (approved):

PR #2733, opened 2021-08-24
[WIP] Index only a prefix of byte fields

### LINKED ISSUE #2287: Bytes field of over a certain size won't be indexable

image

image

vrmiguel:
Closes #2287

Changes:

* Similar to what was done in #1138, byte fields are now indexed by their prefixes instead of their full value as to avoid `btree`'s limitation of index size.
    * Truncation done [here](https://github.com/graphprotocol/graph-node/blob/d6afa12d80f6f2a7482e5fc0f75ba33959262e18/store/postgres/src/relational.rs#L1324)
- [WIP] Use `PrefixComparison` for bytes as well

### DIFF
--- a/store/postgres/src/relational.rs
+++ b/store/postgres/src/relational.rs
@@ -54,6 +54,7 @@ const DELETE_OPERATION_CHUNK_SIZE: usize = 1_000;
 /// This also makes sure that we do not put strings into a BTree index that's
 /// bigger than Postgres' limit on such strings which is about 2k
 pub const STRING_PREFIX_SIZE: usize = 256;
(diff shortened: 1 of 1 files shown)


store/postgres/src/relational.rs
```diff
@@ -1067,6 +1068,11 @@ impl Column {
         named_type(&self.field_type) == "String" && !self.is_list()
     }
 
+    // Checks if this column's type is a binary string (Posgres' `bytea`)
+    pub fn is_binary_string(&self) -> bool {
```
vrmiguel, line 1072: I'm not fully confident that this is the right way of checking this but it seems to work

@lutter @leoyvens

PR #2736, opened 2021-08-24
Add a 'deployment_failed' metric

lutter:
That makes it possible to distinguish between failed and not-failed subgraphs in Prometheus


leoyvens (approved):
We can add this to NEWS.md for the release.

PR #2747, opened 2021-08-27
[WIP] store: index only a prefix of byte fields

### LINKED ISSUE #2287: Bytes field of over a certain size won't be indexable

vrmiguel:
Closes #2287

Changes:

* Similar to what was done in #1138, byte fields are now indexed by their prefixes instead of their full value as to avoid `btree`'s limitation of index size.
    * Truncation done [here](https://github.com/graphprotocol/graph-node/blob/d6afa12d80f6f2a7482e5fc0f75ba33959262e18/store/postgres/src/relational.rs#L1324)
- [WIP] Use `PrefixComparison` for bytes as well

### DIFF
--- a/store/postgres/src/relational.rs
+++ b/store/postgres/src/relational.rs
@@ -54,6 +54,7 @@ const DELETE_OPERATION_CHUNK_SIZE: usize = 1_000;
 /// This also makes sure that we do not put strings into a BTree index that's
 /// bigger than Postgres' limit on such strings which is about 2k
 pub const STRING_PREFIX_SIZE: usize = 256;
+pub const BYTE_ARRAY_PREFIX_SIZE: usize = 256;
 
 lazy_static! {
     /// Deprecated; use 'graphman stats account-like' instead. A list of
@@ -1067,6 +1068,11 @@ impl Column {
         named_type(&self.field_type) == "String" && !self.is_list()
     }
 
+    /// Checks if this column's type is a binary string (Postgres' `bytea`)
+    pub fn is_binary_string(&self) -> bool {
+        named_type(&self.field_type) == "Bytes" && !self.is_list()
+    }
+
     pub fn is_assignable_from(&self, source: &Self, object: &EntityType) -> Option<String> {
         if !self.is_nullable() && source.is_nullable() {
             Some(format!(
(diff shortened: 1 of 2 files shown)


lutter:
This looks good so far, but as you mentioned in your message, it still requires changes to `PrefixComparison` so that Postgres will actually use the index that this PR creates. I _think_ you can get away with just adapting `PrefixComparison.push_column_prefix` and `PrefixColumn.push_value_prefix` to generate the right thing for `bytea`. Please verify with some manual queries that that index actually gets used.

vrmiguel:
@lutter Do I also need to add support for byte arrays to `DeleteByPrefixQuery`?

store/postgres/src/relational_queries.rs
```diff
@@ -636,8 +636,21 @@ struct PrefixComparison<'a> {
...
+        } else {
+            "left"
+        }
+    }
+
     fn push_column_prefix(column: &Column, mut out: AstPass<Pg>) -> QueryResult<()> {
-        out.push_sql("left(");
+        // If not a binary string (bytea), then we're prefixing text.
+        out.push_sql(Self::prefixing_function(column));
+        out.push_sql("(");
```
lutter, line 653: This won't work; `select substring('\xdeadbeef'::bytea, 3)` returns the byte string starting at position 3. To get the equivalent of `left(s, n)` you need to use `substring(b, 1, n)`

You also introduced a constant `BYTE_ARRAY_PREFIX_SIZE` so that the equivalent of `left(s, STRING_ARRAY_PREFIX_SIZE)` should be `substring(b, 1, BYTEA_ARRAY_PREFIX_SIZE)`.

store/postgres/src/relational.rs
```diff
@@ -54,6 +54,7 @@ const DELETE_OPERATION_CHUNK_SIZE: usize = 1_000;
 /// This also makes sure that we do not put strings into a BTree index that's
 /// bigger than Postgres' limit on such strings which is about 2k
 pub const STRING_PREFIX_SIZE: usize = 256;
+pub const BYTE_ARRAY_PREFIX_SIZE: usize = 256;
```
lutter, line 57: I wonder if we could make that one smaller, say 64, in the interest of keeping indexes smaller since `bytea` should have more entropy than `text`.

lutter (changes requested):
Almost good to go.

lutter:
> @lutter Do I also need to add support for byte arrays to `DeleteByPrefixQuery`?

Thanks for the reminder - as it turns out, that query is dead code. I'll remove it in a separate PR. So no need to worry about that.

lutter:
#3340 is a better, more maintainable implementation of this idea.

Issue #2677, opened 2021-08-02
Return PoI for failed subgraphs


azf20:
*At or after* the deterministic error has occurred

- Ensure that we save the state
- Allow an indexer to close an allocation even after a deterministic error, at the normal block that they would close this allocation
- Add a new event that an error has occurred and hash that into the PoI state
- Return the PoI value
- Are there backwards compatibility considerations?

otaviopace:
I've done some research and to return PoI for failing subgraphs we'll have to actually apply database operations even if deterministic errors happened.

Whenever we [process a block](https://github.com/graphprotocol/graph-node/blob/505a7254f4edcb537b584794bc8bb04fe8a5775d/core/src/subgraph/instance_manager.rs#L656), we do an [early return](https://github.com/graphprotocol/graph-node/blob/505a7254f4edcb537b584794bc8bb04fe8a5775d/core/src/subgraph/instance_manager.rs#L819) that happens before we actually save entities in the database and the **PoI as well**.

The idea for the fix is to change this `if` to be after this [function call](https://github.com/graphprotocol/graph-node/blob/505a7254f4edcb537b584794bc8bb04fe8a5775d/core/src/subgraph/instance_manager.rs#L892)(`store.transact_block_operations`), so the PoI actually gets saved, and **can be returned via GraphQL**.

There are two points I'm not exactly sure (still need to talk with Leo and Zac about this) is related to "unfailing" subgraphs. After a subgraph recovered from a failure:

1. Should the `subgraphs.subgraph_error` table be empty again?
2. Should the `health` of the `subgraphs.subgraph_deployment` be `healthy` or `unhealthy`?

There's a function call that does this: https://github.com/graphprotocol/graph-node/blob/505a7254f4edcb537b584794bc8bb04fe8a5775d/store/postgres/src/deployment.rs#L517

I **may** have to call this function in another place, or change the conditions for it to be called [here](https://github.com/graphprotocol/graph-node/blob/505a7254f4edcb537b584794bc8bb04fe8a5775d/core/src/subgraph/instance_manager.rs#L579).

Also, I'm not quite sure yet on automated testing (and integration testing). Once I solve those questions and get more into the coding part, I'll post here, if necessary.

For now I've done [this PR](https://github.com/graphprotocol/graph-node/pull/2700) that is in progress just to see if the CI fails with the simple `if` move.

otaviopace:
Sample GraphQL query to get PoI:

```shell
[...]

PR #2748, opened 2021-08-27
PoI for failed subgraphs

evaporei:
This PR aims to allow failed subgraphs (by deterministic reasons) to have (and return) the PoI.

This can be reviewed commit by commit, and the main changes are:

- Make the instance manager save only the PoI when a deterministic error happens;
- Add a new variant to the ProofOfIndexingEvent, to represent the deterministic error case;
- Make the runtime stop writing ProofOfIndexingEvents when the BlockState already has a deterministic error;
- Add integration tests reproducing this case;
- Make store return PoI for future blocks.

Issue: https://github.com/graphprotocol/graph-node/issues/2677

### DIFF
--- a/core/src/subgraph/instance_manager.rs
+++ b/core/src/subgraph/instance_manager.rs
@@ -714,6 +714,9 @@ async fn process_block<T: RuntimeHostBuilder<C>, C: Blockchain>(
         None
     };
 
+    // There are currently no other causality regions since offchain data is not supported.
+    let causality_region = CausalityRegion::from_network(ctx.state.instance.network());
+
     // Process events one after the other, passing in entity operations
     // collected previously to every new event being processed
     let mut block_state = match process_triggers(
@@ -727,6 +730,7 @@ async fn process_block<T: RuntimeHostBuilder<C>, C: Blockchain>(
         &ctx.state.instance,
         &block,
         triggers,
+        &causality_region,
     )
     .await
     {
@@ -806,20 +810,9 @@ async fn process_block<T: RuntimeHostBuilder<C>, C: Blockchain>(
             data_sources,
         );
 
-        // There are currently no other causality regions since offchain data is not supported.
-        let causality_region = CausalityRegion::from_network(ctx.state.instance.network());
-
         // Process the triggers in each host in the same order the
         // corresponding data sources have been created.
         for trigger in triggers {
-            let error_count = block_state.deterministic_errors.len();
-
-            if let Some(proof_of_indexing) = &proof_of_indexing {
-                proof_of_indexing
-                    .borrow_mut()
-                    .start_handler(&causality_region);
-            }
-
             block_state = SubgraphInstance::<C, T>::process_trigger_in_runtime_hosts(
                 &logger,
                 &runtime_hosts,
@@ -827,6 +820,7 @@ async fn process_block<T: RuntimeHostBuilder<C>, C: Blockchain>(
                 &trigger,
                 block_state,
                 proof_of_indexing.cheap_clone(),
+                &causality_region,
             )
             .await
             .map_err(|e| {
@@ -840,18 +834,6 @@ async fn process_block<T: RuntimeHostBuilder<C>, C: Blockchain>(
                     }
                 }
             })?;
-
-            if let Some(proof_of_indexing) = &proof_of_indexing {
-                if block_state.deterministic_errors.len() != error_count {
-                    assert!(block_state.deterministic_errors.len() == error_count + 1);
-
-                    // If a deterministic error has happened, write a new
-                    // ProofOfIndexingEvent::DeterministicError to the SharedProofOfIndexing.
-                    proof_of_indexing
-                        .borrow_mut()
-                        .write_deterministic_error(&logger, &causality_region);
-                }
-            }
         }
     }
 
@@ -1038,6 +1020,7 @@ async fn process_triggers<C: Blockchain>(
     instance: &SubgraphInstance<C, impl RuntimeHostBuilder<C>>,
     block: &Arc<C::Block>,
     triggers: Vec<C::TriggerData>,
+    causality_region: &str,
 ) -> Result<BlockState<C>, MappingError> {
     use graph::blockchain::TriggerData;
 
@@ -1050,6 +1033,7 @@ async fn process_triggers<C: Blockchain>(
                 &trigger,
                 block_state,
                 proof_of_indexing.cheap_clone(),
+                causality_region,
             )
             .await
             .map_err(move |mut e| {
--- a/core/src/subgraph/instance.rs
+++ b/core/src/subgraph/instance.rs
@@ -112,6 +112,7 @@ where
         trigger: &C::TriggerData,
         state: BlockState<C>,
         proof_of_indexing: SharedProofOfIndexing,
+        causality_region: &str,
     ) -> Result<BlockState<C>, MappingError> {
         Self::process_trigger_in_runtime_hosts(
             logger,
@@ -120,6 +121,7 @@ where
             trigger,
             state,
             proof_of_indexing,
+            causality_region,
         )
         .await
     }
@@ -131,7 +133,16 @@ where
         trigger: &C::TriggerData,
         mut state: BlockState<C>,
         proof_of_indexing: SharedProofOfIndexing,
+        causality_region: &str,
     ) -> Result<BlockState<C>, MappingError> {
+        let error_count = state.deterministic_errors.len();
+
+        if let Some(proof_of_indexing) = &proof_of_indexing {
+            proof_of_indexing
+                .borrow_mut()
+                .start_handler(causality_region);
+        }
+
         for host in hosts {
             let mapping_trigger =
                 match host.match_and_decode(trigger, block.cheap_clone(), logger)? {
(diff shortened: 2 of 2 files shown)


core/src/subgraph/instance_manager.rs
```diff
@@ -841,6 +836,19 @@ async fn process_block<T: RuntimeHostBuilder<C>, C: Blockchain>(
     }
 
     if let Some(proof_of_indexing) = proof_of_indexing {
+        // If a deterministic error has happened, write a new
+        // ProofOfIndexingEvent::DeterministicError to the SharedProofOfIndexing.
+        if has_errors && !is_non_fatal_errors_active {
+            // TODO: get the correct causality region
```
leoyvens, line 842: What's the status on this TODO?

store/postgres/src/deployment_store.rs
```diff
@@ -729,14 +729,16 @@ impl DeploymentStore {
...
+
                         // FIXME: (Determinism)
                         //
                         // It is vital to ensure that the block hash given in the query
                         // is a parent of the latest block indexed for the subgraph.
                         // Unfortunately the machinery needed to do this is not yet in place.
                         // The best we can do right now is just to make sure that the block number
                         // is high enough.
-                        if latest_block_ptr.number < block.number {
+                        if latest_block_ptr.number < block.number && subgraph_error_count != 0 {
```
That3Percent, line 741: We should be returning a PoI when there is an error. I though that was the whole point of this PR. What needs to be verified here is whether the error is a deterministic one.

Also let's fix this determinism issue (above). Determinism is non-negotiable.

That3Percent (changes requested):

runtime/wasm/src/host_exports.rs
```diff
@@ -24,6 +24,34 @@ use wasmtime::Trap;
...
+        //
+        // Let's say up until now we've written 3 events,
+        // and we can imagine the PoI something like this vector:
+        // [setEvent, removeEvent, setEvent]
+        //
+        // Then, a Deterministic error happened. At this point
+        // we'll stop writting to this "list", and we'll have the last
+        // event added by the 'instance_manager' module, like this:
+        //
+        // [setEvent, removeEvent, setEvent, deterministicErrorEvent]
```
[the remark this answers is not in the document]
leoyvens: I think we're just missing an exact specification of how these error events should be written to the PoI. The alternative implemented here does address the concern that if the code is optimized to fail early on the first error, this doesn't introduce a bug to the PoI. But it does have the issue you mention of not including events after the error.

graph/src/components/subgraph/proof_of_indexing/event.rs
```diff
@@ -62,6 +66,7 @@ impl fmt::Debug for ProofOfIndexingEvent<'_> {
                 builder.field("id", id);
                 builder.field("data", &data.iter().collect::<BTreeMap<_, _>>());
             }
+            Self::DeterministicError => {} // NOOP
```
leoyvens, line 69: Is something still written to the PoI, even though this says 'noop'?

That3Percent: Yes, the discriminant is written above.

runtime/wasm/src/host_exports.rs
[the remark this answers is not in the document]
That3Percent: Let's Zoom. I think we could figure it out together quickly.

core/src/subgraph/instance_manager.rs
otaviopace: Do you have a suggestion of where that source of truth would be? 🤔

core/src/subgraph/instance_manager.rs
```diff
@@ -841,6 +836,19 @@ async fn process_block<T: RuntimeHostBuilder<C>, C: Blockchain>(
     }
 
     if let Some(proof_of_indexing) = proof_of_indexing {
+        // If a deterministic error has happened, write a new
+        // ProofOfIndexingEvent::DeterministicError to the SharedProofOfIndexing.
+        if has_errors && !is_non_fatal_errors_active {
```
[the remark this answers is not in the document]
otaviopace: Oh okay, I can change that 🙂 
I think @leoyvens suggested this at some point in the development phase.

graph/src/components/subgraph/proof_of_indexing/event.rs
```diff
@@ -7,15 +7,18 @@ use strum_macros::AsStaticStr;
...
+    /// For when an entity is removed from the store.
+    RemoveEntity { entity_type: &'a str, id: &'a str },
+    /// For when an entity is set into the store.
     SetEntity {
         entity_type: &'a str,
         id: &'a str,
         data: &'a HashMap<String, Value>,
     },
+    /// For when a deterministic error has happened.
+    /// This will be the last event written to the SharedProofOfIndexing until the subgraph
```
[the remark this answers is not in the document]
otaviopace: I think I got it wrong here because to fix a subgraph with a deterministic error, you have to deploy a new one. I'll fix this comment, thanks!

core/src/subgraph/instance_manager.rs
That3Percent: Right now this value is constructed (IIRC) in the host export constructor, which I assume is where you copied it from. It would be nice if there was just one place. I don't care where it ends up, but having two places is going to be error prone when we start introducing new regions.

BTW: It turns out there is a nifty way to ensure that only 2 entities need to be queried to construct a PoI once we have causality regions for off-chain data. Not relevant right now, but I just want to mention this so that when the time comes we can talk about it.

store/postgres/src/deployment_store.rs
otaviopace: The change to the `instance_manager` so that the PoI entity get's saved to the database is enough for the GraphQL query to return the PoI for the failed block.

I think I made the `error_count` change because of some discussions I had at the time about fetching the PoI for future blocks. I think this can be removed.

About the determinism issue above, I'll don't know if we have the "machinery" to fix it right now, but I'll definitely investigate 🙂

That3Percent: I don't think the check can be removed entirely, we still need to verify that the errors are deterministic before producing a PoI... unless non-deterministic errors do not `increment latest_block_ptr.number`?

core/src/subgraph/instance_manager.rs
otaviopace: Oh okay no problem, I'll centralize this then to avoid duplication.

store/postgres/src/deployment_store.rs
otaviopace: The `store.transact_block_operations` that happens in the `instance_manager` does exactly that. This PR makes the call for that function happen so:

- The PoI gets saved to the database even when deterministic errors happen
- The latest_block_ptr gets updated (makes the GraphQL PoI fetch work)

From my understanding this `if` here should work like a confidence check. So (if possible) we should make it stop comparing block number values, and actually check if the block number refers to the correct parent hash.

core/src/subgraph/instance_manager.rs
otaviopace: I've done it here: https://github.com/graphprotocol/graph-node/pull/2748/commits/34428ac06753ff797d012e18e40969a294c6236b#diff-b3d0b8016fff8abd89acad30bfd8e246bd28b54fc340dd4b98c3a19174da7134R45

I would like an opinion if this is a bad place, since I put it there more to use `CausalityRegion` as a "namespace". Instead of following the Rust standard of methods with the `from_` prefix, that implies a "constructor" of some sort.

core/src/subgraph/instance_manager.rs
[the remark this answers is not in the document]
otaviopace: Fixed! https://github.com/graphprotocol/graph-node/pull/2748/commits/532b277dfd799870876327208759d8f74f1d7078

graph/src/components/subgraph/proof_of_indexing/event.rs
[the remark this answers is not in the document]
otaviopace: Fixed! https://github.com/graphprotocol/graph-node/pull/2748/commits/532b277dfd799870876327208759d8f74f1d7078

core/src/subgraph/instance_manager.rs
```diff
@@ -894,14 +902,42 @@ async fn process_block<T: RuntimeHostBuilder<C>, C: Blockchain>(
...
+            "There should be only one PoI EntityModification"
+        );
+    }
+
+    let BlockState {
+        deterministic_errors,
+        ..
+    } = block_state;
+
+    let first_error = deterministic_errors.iter().next().map(|e| e.clone());
```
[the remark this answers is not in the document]
otaviopace: Fixed! https://github.com/graphprotocol/graph-node/pull/2748/commits/8dc1e704321af2b23c98bbc41736e820c7c29356

store/postgres/src/deployment_store.rs
otaviopace: Removed the `error_count` part: https://github.com/graphprotocol/graph-node/pull/2748/commits/7c06e32bdaede435d77f1d92ef74a6881eed4dcd

I'll create an issue for this and talk with David about it.

runtime/wasm/src/host_exports.rs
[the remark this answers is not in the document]
otaviopace: We've done a call to discuss this and this `match` got removed here https://github.com/graphprotocol/graph-node/pull/2748/commits/6ba70d7ad67817b336d2becd7944e37b83bc00f9

otaviopace:
@That3Percent @leoyvens I've moved the `poi.write(deterministic_err_event)` to each handler here: https://github.com/graphprotocol/graph-node/pull/2748/commits/d9b37277f780ce77f016a03fc6bfc4a0ec3cc27e

That3Percent (approved):

PR #2751, opened 2021-08-28
Assorted minor code improvements

vrmiguel:


otaviopace (approved):
LGTM :shipit: 

Also, how did you find these? `clippy`? Maybe we should add a step in our CI that at least list the warnings/suggestions, we can let it fail so it doesn't block PRs.

vrmiguel:
Yup, I used Clippy!  :v

PR #2753, opened 2021-08-28
store, graph, chain: make `WritableStore::get` receive borrowed keys

### LINKED ISSUE #2752: `WritableStore::get` needing owned keys leads to unneeded clones
We could make `WritableStore::get` receive borrowed values, similar to the `get` methods of `std`'s collections, in order to avoid clones like this:

https://github.com/graphprotocol/graph-node/blob/ff93613a3519f1abb9810aaef406fdc99d43033a/graph/src/components/store.rs#L1729

vrmiguel:
Closes #2752


tilacog:
Awesome, thanks!

tilacog (approved):

Issue #2732, opened 2021-08-24
[meta] NEAR Support


maoueh:
This is a meta issue listing the actual work required to fully support NEAR inside the The Graph.

#### Depends 

- #2716 

##### Tasks

- [ ] Ensure `graph-node` is really multi-chain ready (data sources and manifest parsing/instantiation were discussed as potential missing area)
- [ ] Add basic `chain/near` package fulfilling `Blockchain` trait (will probably start with POC mappings for now, maybe just block header or even just logs as `String`).
- [ ] Ensure NEAR on StreamingFast works when block(s) are skipped.
- [ ] Discuss, finalized and implement correct mappings (stuff is still in the discussion right now about what will be available as a mapping)
- [ ] Code  generation adjustments for NEAR on [graph-cli](https://github.com/graphprotocol/graph-cli) (Does `graph-cli` requires some multi-chain love?)
- [ ] Alpha preview testing for early tester(s)
- [ ] Documentation modifications for NEAR support
- [ ] Fix all `FIXME (NEAR)` in the codebase
- [ ] Discuss and decide how hosted NEAR beta version will be operationalized
- [ ] Beta testing on hosted environment

PR #2755, opened 2021-08-30
blockchain: Introduce `BlockchainMap` and `BlockchainKind`

leoyvens:
This introduces the `BlockchainMap` in which `Blockchain` instances are stored, keyed by their `BlockchainKind` and network. This is used in the entry points that need to call functions which are generic on `C: Blockchain`, which are: `InstanceManager::start_subgraph`, `SubgraphRegistrar::create_subgraph` and `IndexNodeResolver::resolve_subgraph_features`.

Part of #2732.

### DIFF
--- a/core/src/subgraph/registrar.rs
+++ b/core/src/subgraph/registrar.rs
@@ -3,34 +3,35 @@ use std::time::Instant;
 
 use async_trait::async_trait;
 use graph::blockchain::Blockchain;
+use graph::blockchain::BlockchainKind;
 use graph::blockchain::BlockchainMap;
 use graph::components::store::{DeploymentId, DeploymentLocator, SubscriptionManager};
 use graph::data::subgraph::schema::SubgraphDeploymentEntity;
+use graph::data::subgraph::MAX_SPEC_VERSION;
 use graph::prelude::{
     CreateSubgraphResult, SubgraphAssignmentProvider as SubgraphAssignmentProviderTrait,
     SubgraphRegistrar as SubgraphRegistrarTrait, *,
 };
 
-pub struct SubgraphRegistrar<C, L, P, S, SM> {
+pub struct SubgraphRegistrar<L, P, S, SM> {
     logger: Logger,
     logger_factory: LoggerFactory,
     resolver: Arc<L>,
     provider: Arc<P>,
     store: Arc<S>,
     subscription_manager: Arc<SM>,
-    chains: Arc<BlockchainMap<C>>,
+    chains: Arc<BlockchainMap>,
     node_id: NodeId,
     version_switching_mode: SubgraphVersionSwitchingMode,
     assignment_event_stream_cancel_guard: CancelGuard, // cancels on drop
 }
 
-impl<C, L, P, S, SM> SubgraphRegistrar<C, L, P, S, SM>
+impl<L, P, S, SM> SubgraphRegistrar<L, P, S, SM>
 where
     L: LinkResolver + Clone,
     P: SubgraphAssignmentProviderTrait,
     S: SubgraphStore,
     SM: SubscriptionManager,
-    C: Blockchain,
 {
     pub fn new(
         logger_factory: &LoggerFactory,
@@ -38,7 +39,7 @@ where
         provider: Arc<P>,
         store: Arc<S>,
         subscription_manager: Arc<SM>,
-        chains: Arc<BlockchainMap<C>>,
+        chains: Arc<BlockchainMap>,
         node_id: NodeId,
         version_switching_mode: SubgraphVersionSwitchingMode,
     ) -> Self {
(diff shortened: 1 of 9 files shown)


core/src/subgraph/registrar.rs
```diff
@@ -271,46 +271,47 @@ where
...
+                    )
+                })?;
 
-        let chain = self
-            .chains
-            .get(&network_name)
-            .ok_or(SubgraphRegistrarError::NetworkNotSupported(
-                network_name.clone(),
-            ))?
-            .cheap_clone();
```
leoyvens, line 294: This block of code was moved to the `create_subgraph_version` free function.

graph/src/data/subgraph/mod.rs
```diff
@@ -795,26 +795,22 @@ impl<C: Blockchain> UnvalidatedSubgraphManifest<C> {
...
         store: Arc<S>,
-    ) -> Result<
-        (SubgraphManifest<C>, Vec<SubgraphManifestValidationWarning>),
-        Vec<SubgraphManifestValidationError>,
-    > {
-        let (schemas, import_errors) = self.0.schema.resolve_schema_references(store.clone());
-        let validation_warnings = import_errors
-            .into_iter()
-            .map(SubgraphManifestValidationWarning::SchemaValidationWarning)
-            .collect();
```
leoyvens, line 817: These `import_errors` are not used anyways, so they are now ignored.

graph/src/data/subgraph/mod.rs
```diff
@@ -921,40 +917,12 @@ impl<C: Blockchain> UnvalidatedSubgraphManifest<C> {
 
 impl<C: Blockchain> SubgraphManifest<C> {
     /// Entry point for resolving a subgraph definition.
-    /// Right now the only supported links are of the form:
-    /// `/ipfs/QmUmg7BZC1YP1ca66rRtWKxpXp77WgVHrnv263JtDuvs2k`
-    pub async fn resolve(
```
leoyvens, line 926: Removed because `resolve_from_raw` is now used in all cases.

node/src/main.rs
```diff
@@ -205,15 +205,21 @@ async fn main() {
         let chain_head_update_listener = store_builder.chain_head_update_listener();
         let network_store = store_builder.network_store(idents);
 
-        let chains = Arc::new(networks_as_chains(
+        // To support the ethereum block ingestor, ethereum networks are referenced both by the
+        // `blockchain_map` and `ethereum_chains`. Future chains should be referred to only in
+        // `blockchain_map`.
+        let mut blockchain_map = BlockchainMap::new();
+        let ethereum_chains = networks_as_chains(
+            &mut blockchain_map,
```
leoyvens, line 213: This hack is because I preferred to not refactor the block ingestor to use `BlockchainMap` since it is unlikely to be used for new chains, so `ethereum_chains` continues to exists.

graph/src/data/subgraph/mod.rs
```diff
@@ -1024,6 +992,7 @@ impl<C: Blockchain> UnresolvedSubgraphManifest<C> {
         self,
         resolver: &impl LinkResolver,
         logger: &Logger,
+        max_spec_version: semver::Version,
```
leoyvens, line 995: Adding this is actually unrelated to this PR, it's just improving the tests to no need the GRAPH_MAX_SPEC_VERSION env var set which was introduced in https://github.com/graphprotocol/graph-node/pull/2682.

graph/src/blockchain/mod.rs
```diff
@@ -305,3 +310,73 @@ pub trait RuntimeAdapter<C: Blockchain>: Send + Sync {
...
+impl FromStr for BlockchainKind {
+    type Err = Error;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "ethereum" => Ok(BlockchainKind::Ethereum),
+            _ => Err(anyhow!("unknown blockchain kind {}", s)),
+        }
+    }
+}
```
tilacog, line 339: How do you feel about delegating this to `serde::plain`, just like we did for [subgraph features](https://github.com/graphprotocol/graph-node/blob/master/graph/src/data/subgraph/features.rs#L36-L51)?

While the current implementation works flawlessly, if we use `serde` here we wouldn't need to write any more code when introducing new variants.

server/index-node/src/resolver.rs
```diff
@@ -379,7 +396,9 @@ where
             }
 
             // The top-level `indexingStatusForPendingVersion` field
-            (None, "subgraphFeatures") => self.resolve_subgraph_features(arguments),
+            (None, "subgraphFeatures") => {
+                graph::block_on(self.resolve_subgraph_features(arguments))
+            }
```
tilacog, line 401: Way better! :clap:

tilacog (approved):

graph/src/blockchain/mod.rs
leoyvens: I don't have much of a preference so I'll keep it as it is.

PR #2768, opened 2021-09-01
store: Allow setting min_idle and idle_timeout

lutter:
This introduces two new environment variables that allow setting the
minimum number of idle connections and the timeout when a connection is
considered idle. If the variables are not set, the behavior defaults to the
current behavior of keeping all connections and never reaping idle ones,
and an idle timeout of 10 minutes.

### DIFF
--- a/store/postgres/src/connection_pool.rs
+++ b/store/postgres/src/connection_pool.rs
@@ -46,10 +46,25 @@ lazy_static::lazy_static! {
             })
             .unwrap_or(0)
     };
+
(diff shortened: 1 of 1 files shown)


leoyvens (approved):
Are these worthy of documenting?

lutter:
> Are these worthy of documenting?

I didn't document them on purpose. I think eventually we want them to be configurable per pool in the config file, but this change should at least let us try out how closing idle connections more aggressively works at scale.

PR #2769, opened 2021-09-01
Fix transaction receipts queries

tilacog:
This PR introduces specific column handling for differentiating queries in the tables:
- `public.ethereum_blocks`
- `chain*.blocks`

Also, a basic test is implemented for checking SQL correctness.

### DIFF
--- a/graph/src/components/transaction_receipt.rs
+++ b/graph/src/components/transaction_receipt.rs
@@ -6,6 +6,7 @@
 use web3::types::{TransactionReceipt, H256, U256, U64};
 
 /// Like web3::types::Receipt, but with fewer fields.
+#[derive(Debug, PartialEq)]
 pub struct LightTransactionReceipt {
     pub transaction_hash: H256,
     pub transaction_index: U64,
--- a/store/postgres/src/chain_store.rs
+++ b/store/postgres/src/chain_store.rs
@@ -80,7 +80,8 @@ mod data {
     };
 
     use crate::transaction_receipt::RawTransactionReceipt;
-    use crate::ETHEREUM_BLOCKS_TABLE_NAME;
+
+    pub(crate) const ETHEREUM_BLOCKS_TABLE_NAME: &'static str = "public.ethereum_blocks";
 
     mod public {
         pub(super) use super::super::public::ethereum_networks;
@@ -1136,7 +1137,6 @@ from (
                         .get_results(conn),
                 }
             };
-
             query_results
                 .map_err(|error| {
                     anyhow::anyhow!(
--- a/store/postgres/src/chain_store.rs
+++ b/store/postgres/src/chain_store.rs
@@ -49,6 +49,7 @@ pub use data::Storage;
 /// Encapuslate access to the blocks table for a chain.
 mod data {
 
+    use diesel::sql_types::Binary;
     use diesel::{connection::SimpleConnection, insert_into};
     use diesel::{delete, prelude::*, sql_query};
     use diesel::{dsl::sql, pg::PgConnection};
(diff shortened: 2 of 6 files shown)


graph/src/components/transaction_receipt.rs
```diff
@@ -6,6 +6,7 @@
 use web3::types::{TransactionReceipt, H256, U256, U64};
 
 /// Like web3::types::Receipt, but with fewer fields.
+#[derive(Debug, PartialEq)]
```
tilacog, line 9: This is only for tests

store/postgres/src/chain_store.rs
```diff
@@ -366,7 +368,7 @@ mod data {
         /// Returns a fully qualified table name to the blocks table
         fn blocks_table(&self) -> &str {
             match self {
-                Storage::Shared => "public.ethereum_blocks",
+                Storage::Shared => ETHEREUM_BLOCKS_TABLE_NAME,
                 Storage::Private(Schema { blocks, .. }) => &blocks.qname,
             }
         }
```
tilacog, line 374: Since I'm reusing this table name, I'm storing it inside a constant.

store/postgres/src/transaction_receipt.rs
```diff
@@ -1,72 +1,26 @@
...
 use graph::prelude::transaction_receipt::LightTransactionReceipt;
 
+use crate::ETHEREUM_BLOCKS_TABLE_NAME;
+
 /// Queries the database for all the transaction receipts in a given block range.
 pub fn find_transaction_receipts_in_block(
     conn: &PgConnection,
     blocks_table_name: &str,
     block_hash: H256,
 ) -> anyhow::Result<Vec<LightTransactionReceipt>> {
```
tilacog, line 21: Diffing got weird here, but I remove all the custom impls for `QueryFragment` and used  the`sql_query` funciton instead.

store/postgres/tests/relational.rs
```diff
@@ -1614,3 +1615,14 @@ fn text_not_in() {
...
 }
+
+#[test]
+fn test_find_transaction_receipts_in_block_function() {
+    use graph_store_postgres::transaction_receipt::find_transaction_receipts_in_block;
+    run_test_with_conn(|conn| {
+        let receipts =
+            find_transaction_receipts_in_block(conn, "public.ethereum_blocks", H256::zero())
+                .unwrap();
+        assert_eq!(receipts, vec![])
```
tilacog, line 1626: In essence, this is a smoke test that is only concerned about the SQL roundtrip being error free.

store/postgres/tests/relational.rs
```diff
@@ -1614,3 +1615,14 @@ fn text_not_in() {
...
+    })
+}
```
tilacog, line 1628: In essence, this is a smoke test that is only concerned about the SQL roundtrip being error-free.

lutter: This test should go in `chain_head.rs` - the `run_test` there will actually run your test against both a chain that stores things in the public schema and in a private schema, i.e., runs your test twice for both setups.

store/postgres/src/transaction_receipt.rs
```diff
@@ -75,32 +29,44 @@ select
...
-        Nullable<Binary>,
-        Nullable<Binary>,
-        Nullable<Binary>,
-    );
-}
+    let query_results: Result<Vec<RawTransactionReceipt>, diesel::result::Error> = {
+        // The `hash` column has different types between the `public.ethereum_blocks` and the
+        // `chain*.blocks` tables, so we must check which one is being queried to bind the
+        // `block_hash` parameter to the correct type
+        if blocks_table_name == ETHEREUM_BLOCKS_TABLE_NAME {
```
lutter, line 45: If this whole function was a method on `Storage` you could make the decision based on what `Storage` variant you have rather than relying on specific table names (see e.g. how `Storage.ancestor_block` works)

lutter (changes requested):

lutter (approved):

PR #2774, opened 2021-09-03
node, store: Do not run pool setup from `graphman`

lutter:
We don't graphman to ever cause migrations or other wholesale database
changes, plus skipping setup makes graphman invocations much faster

### DIFF
--- a/node/src/bin/manager.rs
+++ b/node/src/bin/manager.rs
@@ -361,18 +361,20 @@ impl Context {
 
     fn primary_pool(self) -> ConnectionPool {
         let primary = self.config.primary_store();
-        StoreBuilder::main_pool(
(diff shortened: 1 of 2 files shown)


otaviopace:
@lutter why we don't want `graphman` to run migrations?

lutter:
> @lutter why we don't want `graphman` to run migrations?

Two reasons:
1. It's dangerous - if you run the wrong `graphman` version against an existing database, you might end up migrating your schema which would force you to update `graph-node`
2. It's slow because of all the checking and recreating of fdw tables, each `graphman` invocation becomes painfully slow

If we ever want `graphman` to do schema migrations, it should be an explicit command, and not the silent side effect of using it to look up some information

otaviopace (approved):

PR #2780, opened 2021-09-08
runtime: Add null pointer exception for read

otaviopace:
This can be reviewed commit by commit 🙂 

Before this PR users would get the `Subtract overflow` error in apiVersion `0.0.5`, by using operator overloading from types like `BigInt`:

image

This PR aims at making the error message more clear. Also there are tests for that case as well, the AS code for it should be self explanatory.

Also, this message will be avoided for operator overloading in `graph-ts` just like the `SafeBigInt` did on the tests. PR for this: https://github.com/graphprotocol/graph-ts/pull/204


vrmiguel (approved):

graph/src/runtime/asc_ptr.rs
```diff
@@ -150,12 +150,13 @@ impl<C: AscType> AscPtr<C> {
...
+        // not null before using it.
+        self.assert_is_not_null()?;
+
         let size_of_rt_size = 4;
         let start_of_rt_size = self.0.checked_sub(size_of_rt_size).ok_or_else(|| {
-            DeterministicHostError(anyhow::anyhow!(
-                "Subtract overflow on pointer: {}", // Usually when pointer is zero because of null in AssemblyScript
-                self.0
-            ))
+            DeterministicHostError(anyhow::anyhow!("Subtract overflow on pointer: {}", self.0))
```
lutter, line 159: Could we say something more helpful here (like what that means in terms of pointers)? Is this always an internal error? And it seems that `size_of_rt_size` could be `const`.

graph/src/runtime/asc_ptr.rs
```diff
@@ -173,6 +174,21 @@ impl<C: AscType> AscPtr<C> {
         self.0 == 0
     }
 
+    /// There's no problem in an AscPtr being 'null' (see above AscPtr::is_null function).
+    /// However if one tries to read that pointer, it should fail with a helpful error message,
+    /// this function does this error handling.
+    ///
+    /// Summary: ALWAYS call this before reading an AscPtr.
+    pub fn assert_is_not_null(&self) -> Result<(), DeterministicHostError> {
```
lutter, line 182: Minor, but it would be better to call this `check_is_not_null`, when I see `assert` I expect that the code could panic.

lutter (approved):
Nice!

graph/src/runtime/asc_ptr.rs
otaviopace: Makes sense! I'll change this rn :)

graph/src/runtime/asc_ptr.rs
otaviopace: You're correct, the size of `rt_size` can be changed to a `const`, I'll change that.

This `Subtract overflow` shouldn't happen, it's related to the way every type is layed out in memory.

All class instances in AssemblyScript have a header of 20 bytes, of which the last 4 hold the size of the following data.

AS header memory layout docs: https://www.assemblyscript.org/memory.html#common-header-layout.

Our `AscPtr` structure holds a number which is the address in WebAssembly memory of where that class instance starts their data (right after the header).

So this part of the code tries to do a pointer arithmetic to get the size of the structure, so it can read it fully in a correct fashion.

Maybe what I can write instead of this message is that it failed to find the size of the following structure.

graph/src/runtime/asc_ptr.rs
otaviopace: Fixed!

graph/src/runtime/asc_ptr.rs
otaviopace: I added the constant but I'll let these (there are more) "Subtract overflow" to fix later. I want to discuss more with @leoyvens on this.

PR #2781, opened 2021-09-08
graph, store: bump `num-bigint` and `bigdecimal`

vrmiguel:
This PR bumps
* `num-bigint` from `^0.2.6` to `^0.4.7`
* `bigdecimal`from `0.1.0` to `0.3.0`

This change has pros and cons, however.

### Pros

* These types have had many optimizations since the versions we're currently using were released:
    * https://github.com/rust-num/num-bigint/pull/216
    * https://github.com/rust-num/num-bigint/pull/199
    * https://github.com/rust-num/num-bigint/pull/171
    * https://github.com/rust-num/num-bigint/pull/142
    * https://github.com/rust-num/num-bigint/pull/141
    * https://github.com/rust-num/num-bigint/pull/62
* We can now use upstream's implementation for `BigDecimal::normalized`.
* `bigdecimal` no longer implements `From<f64>` for `BigDecimal`, which is something that could panic before.
This means that the graph-node-defined `scalar::BigDecimal`  struct now implements `TryFrom<f64>` instead of `From<f64>`

### Cons

* The latest Diesel release still uses `num-bigint 0.2.6` and `bigdecimal 0.1.0`.
This means that we're no longer able to use the Diesel-defined `ToSql<diesel::sql_types::Numeric, _>` and `FromSql<diesel::sql_types::Numeric, _> for BigDecimal` implementations for version 0.3.0 of `BigDecimal` until Diesel upgrades their dependencies. 
In order to maintain support for these traits, I implemented the [BigDecimalRetrocompatibility](https://github.com/graphprotocol/graph-node/blob/6d3eb16a5b28a19e66ce7e3d56b9c6c3ce63faa6/graph/src/data/store/scalar.rs#L156) struct, which has functions for converting to and from versions 0.1 and 0.3 of `BigDecimal`.
API limitations make this operation make heap allocations, although I believe those allocations should be small most of the time.

@tilacog  @lutter @otaviopace @leoyvens 
If you guys believe that this PR is the way to go, I'll add tests for `BigDecimalRetrocompatibility` and further docs on things I changed

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -190,7 +190,18 @@ version = "0.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1374191e2dd25f9ae02e3aa95041ed5d747fc77b3c102b49fe2dd9a8117a6244"
 dependencies = [
- "num-bigint",
+ "num-bigint 0.2.6",
+ "num-integer",
+ "num-traits",
+]
+
+[[package]]
+name = "bigdecimal"
+version = "0.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6aaf33151a6429fe9211d1b276eafdf70cdff28b071e76c0b0e1503221ea3744"
+dependencies = [
+ "num-bigint 0.4.2",
  "num-integer",
  "num-traits",
  "serde",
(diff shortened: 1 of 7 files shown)


leoyvens:
Diesel 2.0 will upgrade these dependencies https://github.com/diesel-rs/diesel/pull/2883 so we can choose to wait for that so we don't need the compatibility tricks.

otaviopace:
@vrmiguel @leoyvens will this be continued? Otherwise I think we should close it.

leoyvens:
Closing, we should upgrade this along with Diesel 2.0 which will be released soon.

PR #2791, opened 2021-09-13
chain, graph, runtime: stop cloning `semver::Version` in `from_asc_bytes` impls

vrmiguel:


otaviopace:
Since the `Version` is a struct that implements `Copy` cause it's just 3 numbers, I think passing it by reference is actually more expensive. I mean, it's not like this will do a big performance change, but we did think about doing this earlier, but we didn't because it's "copyable".

vrmiguel:
[`semver::Version`](https://docs.rs/semver/1.0.4/semver/struct.Version.html) isn't actually Copy :C 

It is a quite small  big struct, containing three u64s and two `NonZeroU64`s, so you're right in that cloning it should be really fast.

I got curious as to how both options would perform, however, so I made up a small benchmark using Criterion

Run no. 1
```
do_work_ref             time:   [568.24 us 569.82 us 571.51 us]
do_work_owned           time:   [926.84 us 931.59 us 936.67 us] 
```

Run no. 2
```
do_work_ref             time:   [552.72 us 553.80 us 555.09 us]                        
                        change: [-3.1502% -2.6461% -2.1674%] (p = 0.00 < 0.05)
                        Performance has improved.

do_work_owned           time:   [872.11 us 873.24 us 874.65 us]                          
                        change: [-6.3545% -5.9167% -5.4835%] (p = 0.00 < 0.05)
                        Performance has improved.
```

[Actual benchmark code here](https://gist.github.com/vrmiguel/e3dac2498649fd26232ad38f196fe728)

So this PR _should_ yield us a kind of marginal decrease in latency, specially in functions like `ArrayBuffer::get`, where the version is cloned for every object being read

otaviopace (approved):
Nice job at doing the benchmark 👏 😊

PR #2794, opened 2021-09-14
Changing access modifiers to make unit testing possible and moving test-store to dev dependencies.

georg-getz:
Test store adds a lot of dependencies which aren't really necessary.
In order to mock results and entry data we need access to the changed method and field.


leoyvens (approved):

leoyvens:
thanks!

PR #2795, opened 2021-09-14
Optionally validate subgraph graft base

tilacog:
-------

This PR adds a flag parameter to the `UnvalidatedSubgraphManifest::validate` function to skip graft base validation. At the moment, this will only be used by the feature detection GraphqQL endpoint to statically validate a manifest so it can properly inspect its features.

### DIFF
--- a/server/index-node/src/resolver.rs
+++ b/server/index-node/src/resolver.rs
@@ -222,7 +222,7 @@ where
         // `validate` also validates subgraph features), so we must filter them out to build our
         // response.
         let subgraph_validation: Either<_, _> =
-            match unvalidated_subgraph_manifest.validate(self.subgraph_store.clone()) {
+            match unvalidated_subgraph_manifest.validate(self.subgraph_store.clone(), false) {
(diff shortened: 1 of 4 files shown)


server/index-node/src/resolver.rs
```diff
@@ -222,7 +222,7 @@ where
         // `validate` also validates subgraph features), so we must filter them out to build our
         // response.
         let subgraph_validation: Either<_, _> =
-            match unvalidated_subgraph_manifest.validate(self.subgraph_store.clone()) {
+            match unvalidated_subgraph_manifest.validate(self.subgraph_store.clone(), false) {
                 Ok(subgraph_manifest) => Either::Left(subgraph_manifest),
                 Err(validation_errors) => {
                     // We must ensure that all the errors are of the `FeatureValidationError`
```
tilacog, line 228: This is the only part of the code that need to skip the graft base check.

vrmiguel (approved):

leoyvens (approved):

PR #2813, opened 2021-09-22
build: Fix issues post-firehose

leoyvens:
Fixes a couple of issues falling out from https://github.com/graphprotocol/graph-node/pull/2716. One is the build failing with incremental compilation, which is fixed by disabling that. The other is having rustfmt as a build-time dependency, which is failing the docker build, this is worked around by formatting the tonic generated file only in debug builds.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,6 @@ env:
   CARGO_TERM_COLOR: always
   RUST_BACKTRACE: full
   THEGRAPH_STORE_POSTGRES_DIESEL_URL: "postgresql://postgres:postgres@localhost:5432/graph_node_test"
-  GRAPH_MAX_SPEC_VERSION: "0.0.4"
 
 jobs:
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -10,7 +10,6 @@ env:
   CARGO_TERM_COLOR: always
   RUST_BACKTRACE: full
   THEGRAPH_STORE_POSTGRES_DIESEL_URL: "postgresql://postgres:postgres@localhost:5432/graph_node_test"
-  GRAPH_MAX_SPEC_VERSION: "0.0.4"
```
leoyvens, line 13: This is an unrelated cleanup of an unnecessary setting.

otaviopace:
Do we need to do any change for `release` mode?

lutter (approved):

leoyvens:
@otaviopace release mode has incremental compilation off by default.

PR #2814, opened 2021-09-22
Small fixes around protobuf generation

lutter:
This fixes two small things:
* only regenerate protobuf definitions if the `.proto` files have changed
* install `rustfmt` in the docker build container so docker builds work again

### DIFF
--- a/chain/ethereum/build.rs
+++ b/chain/ethereum/build.rs
@@ -1,4 +1,5 @@
 fn main() {
+    println!("cargo:rerun-if-changed=proto");
     tonic_build::configure()
         .out_dir("src/protobuf")
         .format(cfg!(debug_assertions)) // Release build environments might not have rustfmt installed
--- a/chain/ethereum/build.rs
+++ b/chain/ethereum/build.rs
@@ -2,7 +2,7 @@ fn main() {
     println!("cargo:rerun-if-changed=proto");
     tonic_build::configure()
         .out_dir("src/protobuf")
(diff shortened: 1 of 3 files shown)


chain/ethereum/build.rs
```diff
@@ -1,4 +1,5 @@
 fn main() {
+    println!("cargo:rerun-if-changed=proto");
```
leoyvens, line 2: Witchcraft! For my education, where are the docs on this?

graph/build.rs
```diff
@@ -1,4 +1,5 @@
 fn main() {
+    println!("cargo:rerun-if-changed=proto");
     tonic_build::configure()
         .out_dir("src/firehose")
         .format(true)
```
leoyvens, line 5: This must have been why rustfmt was still being required, I inserted the `cfg!(debug_assertions)` only in the `ethereum` build script and not here. Lets add it here as well for consistency, or remove it in both places if you think that's better.

chain/ethereum/build.rs
otaviopace: https://doc.rust-lang.org/cargo/reference/build-scripts.html

graph/build.rs
lutter: I think requiring `rustfmt` for building is better - I already ran into the problem locally that `cargo check --release` modified the generated files because of the different format settings.

leoyvens (approved):

PR #2815, opened 2021-09-23
Make graph-node more robust in the face of shards being down or restarting

lutter:
This PR is a continuation of PR #2727 and rounds out the story of making `graph-node` handle shards being down during startup and during normal operation more gracefully, including coming back to a sane state when a failed shard comes back up. Even so, it's probably a good idea to restart `graph-node` after a shard restarted.

The basic ideas behind these changes are:
- queries should fail fast if they involve a failed shard
- queries should only fail if they involve a failed shard
- when the primary fails, queries against other shards should still work (achieved with home-grown table replication)
- indexing operations in a failed shard retry and block indefinitely
- indexing operations in other shards are not affected by a shard failure, even when the primary failed
- indexing will not make progress if the failed shard contains the block store for the subgraph's network, but will resume when the block store becomes available again

Testing these things is unfortunately a very manual process. Testing was done under very little load, and it's possible that this did not reveal issues that only appear under high loads. Here are my notes on what I tested:

## Setup
- three shards: `primary`, `sharda`, and `shardb`
- block store also in `primary`
- one simple subgraph (`dice2win`) indexing in each shard

## Notes
- indexing service API returns errors when any shard is down if that shard is needed to form response
- not clear how all this works under high load

### `sharda` is down on startup
- [X] graph-node starts up
- [X] queries against `primary` and `shardb` work
- [X] queries against `sharda` fail fast
- [X] indexing in `primary` and `shardb` continues
- [X] queries against `sharda` work after it comes back up
- [X] indexing in `sharda` resumes when `sharda` comes back up

### `sharda` goes down during operation
- [X] queries against `primary` and `shardb` work
- [X] queries against `sharda` fail fast
- [X] indexing in `primary` and `shardb` continues
- [X] queries against `sharda` work after it comes back up
- [X] indexing in `sharda` resumes when `sharda` comes back up

### `primary` is down on startup
- [X] graph-node starts up
- [X] queries against `sharda` and `shardb` work
- [X] queries against `primary` fail fast
- [X] queries against `primary` work after it comes back up
- [X] indexing resumes when `primary` comes back up

### `primary` goes down during operation
- [X] queries against `sharda` and `shardb` work
- [X] queries against `primary` fail fast
- [X] queries against `primary` work after it comes back up
- [X] indexing resumes when `primary` comes back up

### DIFF
--- a/store/postgres/src/connection_pool.rs
+++ b/store/postgres/src/connection_pool.rs
@@ -232,6 +232,8 @@ impl ForeignServer {
             "dynamic_ethereum_contract_data_source",
             "table_stats",
             "subgraph_deployment_assignment",
+            "subgraph",
+            "subgraph_version",
         ] {
             let create_stmt =
                 catalog::create_foreign_table(conn, "subgraphs", table_name, &nsp, &self.name)?;
--- a/store/postgres/src/connection_pool.rs
+++ b/store/postgres/src/connection_pool.rs
@@ -231,6 +231,7 @@ impl ForeignServer {
             "subgraph_error",
             "dynamic_ethereum_contract_data_source",
             "table_stats",
+            "subgraph_deployment_assignment",
         ] {
             let create_stmt =
                 catalog::create_foreign_table(conn, "subgraphs", table_name, &nsp, &self.name)?;
--- a/store/postgres/src/connection_pool.rs
+++ b/store/postgres/src/connection_pool.rs
@@ -479,6 +479,21 @@ impl ConnectionPool {
     }
 }
 
+fn brief_error_msg(error: &dyn std::error::Error) -> String {
+    // For 'Connection refused' errors, Postgres includes the IP and
+    // port number in the error message. We want to suppress that and
+    // only use the first line from the error message. For more detailed
+    // analysis, 'Connection refused' manifests as a
+    // `ConnectionError(BadConnection("could not connect to server:
+    // Connection refused.."))`
+    error
+        .to_string()
+        .split("\n")
+        .next()
+        .unwrap_or("no error details provided")
+        .to_string()
+}
+
 #[derive(Clone)]
 struct ErrorHandler(Logger, Counter);
 
@@ -491,7 +506,7 @@ impl std::fmt::Debug for ErrorHandler {
 impl r2d2::HandleError<r2d2::Error> for ErrorHandler {
     fn handle_error(&self, error: r2d2::Error) {
         self.1.inc();
-        error!(self.0, "Postgres connection error"; "error" => error.to_string());
+        error!(self.0, "Postgres connection error"; "error" => brief_error_msg(&error));
     }
 }
 
@@ -777,7 +792,7 @@ impl PoolInner {
             match self.pool.get_timeout(*CONNECTION_TIMEOUT) {
                 Ok(conn) => return Ok(conn),
                 Err(e) => error!(logger, "Error checking out connection, retrying";
-                   "error" => e.to_string(),
+                   "error" => brief_error_msg(&e),
                 ),
             }
         }
--- a/store/postgres/src/connection_pool.rs
+++ b/store/postgres/src/connection_pool.rs
@@ -468,13 +468,6 @@ impl ConnectionPool {
         self.get_ready()?.get()
     }
 
-    pub fn get_with_timeout_warning(
-        &self,
-        logger: &Logger,
-    ) -> Result<PooledConnection<ConnectionManager<PgConnection>>, StoreError> {
-        self.get_ready()?.get_with_timeout_warning(logger)
-    }
-
     /// Get a connection from the pool for foreign data wrapper access;
     /// since that pool can be very contended, periodically log that we are
     /// still waiting for a connection
--- a/store/postgres/src/connection_pool.rs
+++ b/store/postgres/src/connection_pool.rs
@@ -24,6 +24,7 @@ use graph::{
 
 use std::fmt::{self, Write};
 use std::str::FromStr;
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
 use std::time::Duration;
 use std::{collections::HashMap, sync::RwLock};
@@ -248,18 +249,20 @@ impl ForeignServer {
 const FDW_IDLE_TIMEOUT: Duration = Duration::from_secs(60);
 
 /// A pool goes through several states, and this enum tracks what state we
-/// are in. When first created, the pool is in state `Created`; once we
-/// successfully called `setup` on it, it moves to state `Ready`. If we
-/// detect that the database is not available, the pool goes into state
-/// `Unavailable` until we can confirm that we can connect to the database,
-/// at which point it goes back to `Ready`
-// `Unavailable` is not used in the code yet
-#[allow(dead_code)]
+/// are in, together with the `state_tracker` field on `ConnectionPool`.
+/// When first created, the pool is in state `Created`; once we successfully
+/// called `setup` on it, it moves to state `Ready`. During use, we use the
+/// r2d2 callbacks to determine if the database is available or not, and set
+/// the `available` field accordingly. Tracking that allows us to fail fast
+/// and avoids having to wait for a connection timeout every time we need a
+/// database connection. That avoids overall undesirable states like buildup
+/// of queries; instead of queueing them until the database is available,
+/// they return almost immediately with an error
 enum PoolState {
     /// A connection pool, and all the servers for which we need to
     /// establish fdw mappings when we call `setup` on the pool
     Created(Arc<PoolInner>, Arc<Vec<ForeignServer>>),
-    Unavailable(Arc<PoolInner>),
+    /// The pool has been successfully set up
     Ready(Arc<PoolInner>),
 }
 
@@ -268,6 +271,15 @@ pub struct ConnectionPool {
     inner: Arc<TimedMutex<PoolState>>,
     logger: Logger,
     pub shard: Shard,
+    state_tracker: PoolStateTracker,
+}
+
+impl fmt::Debug for ConnectionPool {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("ConnectionPool")
+            .field("shard", &self.shard)
+            .finish()
+    }
 }
 
 /// The name of the pool, mostly for logging, and what purpose it serves.
@@ -295,6 +307,31 @@ impl PoolName {
     }
 }
 
+#[derive(Clone)]
+struct PoolStateTracker {
+    available: Arc<AtomicBool>,
+}
+
+impl PoolStateTracker {
+    fn new() -> Self {
+        Self {
+            available: Arc::new(AtomicBool::new(true)),
+        }
+    }
+
+    fn mark_available(&self) {
+        self.available.store(true, Ordering::Relaxed);
+    }
+
+    fn mark_unavailable(&self) {
+        self.available.store(false, Ordering::Relaxed);
+    }
+
+    fn is_available(&self) -> bool {
+        self.available.load(Ordering::Relaxed)
+    }
+}
+
 impl ConnectionPool {
     pub fn create(
         shard_name: &str,
@@ -306,6 +343,7 @@ impl ConnectionPool {
         registry: Arc<dyn MetricsRegistry>,
         servers: Arc<Vec<ForeignServer>>,
     ) -> ConnectionPool {
+        let state_tracker = PoolStateTracker::new();
         let pool = PoolInner::create(
             shard_name,
             pool_name.as_str(),
@@ -314,6 +352,7 @@ impl ConnectionPool {
             fdw_pool_size,
             logger,
             registry,
+            state_tracker.clone(),
         );
         let shard = pool.shard.clone();
         let pool_state = if pool_name.is_replica() {
@@ -325,6 +364,7 @@ impl ConnectionPool {
             inner: Arc::new(TimedMutex::new(pool_state, format!("pool-{}", shard_name))),
             logger: logger.clone(),
             shard,
+            state_tracker,
         }
     }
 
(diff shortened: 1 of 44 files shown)


store/postgres/src/connection_pool.rs
```diff
@@ -773,7 +773,7 @@ impl PoolInner {
             // closure failed.
             let conn = pool
                 .get()
-                .map_err(|e| CancelableError::Error(StoreError::Unknown(e.into())))?;
```
otaviopace, line 744: Do we still use the `Unknown` variant for other `StoreError` scenarios?

store/postgres/src/connection_pool.rs
```diff
@@ -522,8 +568,11 @@ impl std::fmt::Debug for ErrorHandler {
 
 impl r2d2::HandleError<r2d2::Error> for ErrorHandler {
     fn handle_error(&self, error: r2d2::Error) {
-        self.1.inc();
-        error!(self.0, "Postgres connection error"; "error" => brief_error_msg(&error));
+        self.counter.inc();
+        if self.state_tracker.is_available() {
```
otaviopace, line 574: This `handle_error` function, it really was never called when the database was unavailable?

I ask this because we would always log the error message from PostgreSQL it seems, so the connection should be working I beleive.

store/postgres/src/subgraph_store.rs
```diff
@@ -1056,61 +1060,118 @@ impl WritableStore {
             site,
         })
     }
+
+    fn retry<T, F>(&self, op: &str, f: F) -> Result<T, StoreError>
```
otaviopace, line 1074: Is there a possibility where all shards are down and we keep multiple loops like this one running trying to execute queries, until the database goes back up and we overload it with too many queries?

Also I imagine the number of connections requested at the same time might be huge, but I'm not completely sure, maybe I understood something wrong conceptually.

**Edit**: I've seen you removed `get_with_timeout_warning`, so connection might not be an issue here.

store/postgres/src/subgraph_store.rs
```diff
@@ -1111,6 +1111,22 @@ impl WritableStore {
...
+    /// Try to send a `StoreEvent`; if sending fails, log the error but
+    /// return `Ok(())`
+    fn try_send_store_event(&self, event: StoreEvent) -> Result<(), StoreError> {
+        if *SEND_SUBSCRIPTION_NOTIFICATIONS {
+            self.store
+                .send_store_event(&event)
+                .map_err(
+                    |e| error!(self.logger, "Could not send store event"; "error" => e.to_string()),
+                )
+                .ok();
```
otaviopace, line 1118: Why there is this `.ok` here?

The way I usually ignore `Result`s using `let _ = ...;`, but I'm not sure if that's what you're trying to do here, enlighten me 🙂

otaviopace (commented):
First of all, great PR 👏  It took me some time to review haha.

Well, I've posted multiple questions and suggestions, some are more style related though.

Some questions I have about the PR description are:

> Even so, it's probably a good idea to restart graph-node after a shard restarted.

I'm curious, why is that?

> queries should only fail if they involve a failed shard

What part of this scenario is not covered today by https://github.com/graphprotocol/graph-node/pull/2727? Also, after reading through your code, it seems that the behavior is to try to get a connection from a shard and if that fails, we try on the next one. How would a failed shard get a query invoked? 🤔  (maybe I just misunderstood your statement above)

> indexing operations in a failed shard retry and block indefinitely

What happens today on the indexing part? I think I'm more aware on the query side.

Awesome test cases by the way, that must've been a lot of work 😅

lutter:
> First of all, great PR clap It took me some time to review haha.

Thanks!

> Some questions I have about the PR description are:
> 
> > Even so, it's probably a good idea to restart graph-node after a shard restarted.
> 
> I'm curious, why is that?

There's at least two things that I am worried about: (1) I am not 100% confident that this catches all ways in which indexing could fail when a shard goes down (especially when the shard with the block store goes down at the wrong moment) and (2) because listeners reconnect after some delay, it's possible that they miss assignment events, and, e.g., a subgraph deployed during that time doesn't start syncing.

> > queries should only fail if they involve a failed shard
> 
> What part of this scenario is not covered today by #2727? Also, after reading through your code, it seems that the behavior is to try to get a connection from a shard and if that fails, we try on the next one. How would a failed shard get a query invoked? thinking (maybe I just misunderstood your statement above)

#2727 does not cover the scenario where the primary fails - without this PR, when the primary fails, most queries will fail, since the primary is needed to determine where a subgraph is stored. There's a bit of [caching in memory](https://github.com/graphprotocol/graph-node/blob/master/store/postgres/src/subgraph_store.rs#L246) of this data, but it's with a 120s TTL and doesn't cover the resolution of subgraph names to hashes.

The only data that we try getting from different shards is whatever is mentioned in `primary::Mirror`, but that's the crucial stuff that helps in resolving subgraph names to deployments, and finding which shard has the data for a deployment.


> > indexing operations in a failed shard retry and block indefinitely
> 
> What happens today on the indexing part? I think I'm more aware on the query side.

The subgraph will just fail and needs to be restarted.

store/postgres/src/notification_listener.rs
```diff
@@ -102,6 +107,34 @@ impl NotificationListener {
...
+                        let delay = Duration::from_secs(1 << attempt);
+                        error!(logger, "Failed to connect notification listener: {}", e;
+                                       "attempt" => attempt,
+                                       "retry_delay_s" => delay.as_secs());
+                        if attempt < 5 {
+                            attempt = attempt + 1;
+                        }
+                        thread::sleep(delay);
+                    }
+                    Ok(conn) => {
```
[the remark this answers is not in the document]
lutter: The `return` is needed to get out of the outer `loop` - that also makes using `map_err` etc. awkward, specially since we just log the error and forget about it.

store/postgres/src/notification_listener.rs
```diff
@@ -102,6 +107,34 @@ impl NotificationListener {
...
+                    Ok(conn)
+                });
+                match res {
+                    Err(e) => {
+                        let delay = Duration::from_secs(1 << attempt);
+                        error!(logger, "Failed to connect notification listener: {}", e;
+                                       "attempt" => attempt,
+                                       "retry_delay_s" => delay.as_secs());
+                        if attempt < 5 {
+                            attempt = attempt + 1;
```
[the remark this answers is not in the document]
lutter: Heh .. looks like I did that when I factored this into `ExponentialBackoff`

store/postgres/src/notification_listener.rs
```diff
@@ -102,6 +107,34 @@ impl NotificationListener {
         Arc<AtomicBool>,
         Arc<Barrier>,
     ) {
+        /// Connect to the database at `postgres_url` and do a `LISTEN
+        /// {channel_name}`. If that fails, retry with exponential backoff
+        /// with a delay between 1s and 32s
+        fn connect_and_listen(logger: &Logger, postgres_url: &str, channel_name: &str) -> Client {
```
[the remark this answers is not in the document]
lutter: I think if we ever have tests for this, we should pull it out ;)

store/postgres/src/notification_listener.rs
```diff
@@ -158,27 +192,28 @@ impl NotificationListener {
...
+                                if connected {
+                                    let msg = format!("{}", e);
+                                    crit!(logger, "Error receiving message"; "error" => &msg);
+                                }
+                                connected = false;
+                                None
                             }
                         })
                         .filter(|notification| notification.channel() == channel_name.0)
-                        .take(64)
```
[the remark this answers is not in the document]
lutter: No, it's fairly random; we just can't wait until we got all messages in the queue since we might never finish getting messages on a busy system. That number is just made up, but seems to work fine.

store/postgres/src/notification_listener.rs
```diff
@@ -63,6 +63,11 @@ impl NotificationListener {
     /// channel.
     ///
     /// Must call `.start()` to begin receiving notifications on the returned receiver.
+    //
+    /// The listener will handle dropping the database connection by
+    /// indefinitely trying to reconnect to the database. Users of the
+    /// listener have now way to find out whether the connection had been
```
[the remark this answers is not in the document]
lutter: That should read 'have no way'. Fixed

store/postgres/src/connection_pool.rs
```diff
@@ -479,6 +479,21 @@ impl ConnectionPool {
     }
 }
 
+fn brief_error_msg(error: &dyn std::error::Error) -> String {
```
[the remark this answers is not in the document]
lutter: Because those logs can show up in public (e.g., epxlorer) and I didn't want that information in public logs. We still know which shard is affected and could then look the IP up in the config.

store/postgres/src/connection_pool.rs
```diff
@@ -220,6 +222,45 @@ impl ForeignServer {
...
 
+    fn mirror_primary_tables(conn: &PgConnection) -> Result<(), StoreError> {
+        // `chains` needs to be mirrored before `deployment_schemas` because
+        // of the fk constraint on `deployment_schemas.network`. We don't
+        // care much about mirroring `active_copies` but it has a fk
+        // constraint on `deployment_schemas` and is tiny, therefore it's
+        // easiest to just mirror it
+        const PUBLIC_TABLES: [&str; 3] = ["chains", "deployment_schemas", "active_copies"];
+
+        fn copy_table(
```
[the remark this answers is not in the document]
lutter: No, we don't have that elsewhere; and it really only works for relatively small tables. For large tables, the resulting txn would run way too slow. That's one reason why I wrote this as a nested method - it's really just a convenience to construct a very specific query.

store/postgres/src/connection_pool.rs
```diff
@@ -877,6 +931,11 @@ impl PoolInner {
             die(&pool.logger, "failed to release migration lock", &err);
         });
         result.unwrap_or_else(|err| die(&pool.logger, "migrations failed", &err));
+        pool.mirror_primary_tables().unwrap_or_else(|e| {
```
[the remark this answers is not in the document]
lutter: I tried it at some point with hosted service data, and IIRC, this took less than a minute (could have been even faster)

store/postgres/src/primary.rs
```diff
@@ -1365,3 +1366,107 @@ impl<'a> Connection<'a> {
...
+        Mirror { pools }
+    }
+
+    /// Execute the function `f` with connections from each of our pools in
+    /// order until for one of them we get any result other than
+    /// `Err(StoreError::DatabaseUnavailable)`. In other words, we try to
+    /// execute `f` against our pools in order until we can be sure that we
+    /// talked to a database that is up. The function `f` must only access
+    /// tables that are mirrored through `refresh_tables`
+    pub(crate) fn read<'a, T>(
```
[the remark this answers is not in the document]
lutter: Thanks!

store/postgres/src/connection_pool.rs
lutter: Yeah, there are a ton of uses of it all over the code.

store/postgres/src/connection_pool.rs
lutter: Yes, it does get called when the database is unavailable. The only thing this change does is only log the error the first time the db becomes unavailable rather than spamming the logs with the same error message.

otaviopace: Ohhh okay then.

store/postgres/src/subgraph_store.rs
lutter: Requesting too many connections isn't an issue since most of them will just block on the pool until one becomes available.

Not sure what happens if the shard has been down for a long time - there is a chance of a thundering herd, we'll have to see. Since the backoff is 10s after a while, I'd hope that the retries are splayed reasonably.

store/postgres/src/subgraph_store.rs
lutter: That's a much better way (and all I was going for here) Changed.

lutter:
Addressed all review comments (I think)

otaviopace (approved):
I'm approving because it seems correct and I think I understood what most of your code does. However I'm not sure if you want anyone else to take a look just to to be more sure that none of the new code can have unexpected behavior on the database connection/query handling/execution.

lutter:
Rebased to latest master

store/postgres/src/notification_listener.rs
```diff
@@ -81,6 +82,11 @@ impl NotificationListener {
     /// channel.
     ///
     /// Must call `.start()` to begin receiving notifications on the returned receiver.
+    //
```
otaviopace, line 85: Small thing, but I think there should be another `/` here.

otaviopace:
Also, one thing I forgot to ask on the first review is about the `Cow`s.
I read your [commit description](https://github.com/graphprotocol/graph-node/pull/2815/commits/4512235fffdc68696ac5930af7a184c1d86d829d) on it and it makes sense, but I didn't actually found where this is used on the [retry](https://github.com/graphprotocol/graph-node/pull/2815/commits/d37c805531c25ec1de9ca4f95fa16c96d7a157fc#diff-fbebe8c1d9a66d5a40732d736208ae22810af9d9f5c321c0d7da9f23d3d8385eR1079).
I'm probably looking at the wrong place. Can you give me a hint on that please?

lutter:
> Also, one thing I forgot to ask on the first review is about the `Cow`s. I read your [commit description](https://github.com/graphprotocol/graph-node/pull/2815/commits/4512235fffdc68696ac5930af7a184c1d86d829d) on it and it makes sense, but I didn't actually found where this is used on the [retry](https://github.com/graphprotocol/graph-node/pull/2815/commits/d37c805531c25ec1de9ca4f95fa16c96d7a157fc#diff-fbebe8c1d9a66d5a40732d736208ae22810af9d9f5c321c0d7da9f23d3d8385eR1079). I'm probably looking at the wrong place. Can you give me a hint on that please?

What changed with retrying is that we need two ways to access the entities we are about to write to the store: one for forming the actual `relational_queries::InsertQuery` and one if the insert fails and we need to retry. Without retrying, we just passed ownership of the entities into the `InsertQuery`, where the entities can get mutated if there are fulltext search fields. Because we still need access to the entities for a possible retry, we can't pass ownership to `InsertQuery` any more. At the same time, mutating the entities in `InsertQuery` is pretty rare, so we don't want to just clone entities on every insert. That's where the `CoW` comes into play since it allows us to pass references into the `InsertQuery`: in the common case, the insert will just use references, and only when there are fulltext search fields will a clone happen.

lutter:
I made a mistake when merging this (not pushing the updated `master`) This is fixed now, and the PR has been merged at commit 042cf0ac

PR #2824, opened 2021-09-27
store: DB notification timeout

leoyvens:
This will prevent the notification listener from blocking forever and letting the PG queue fill up. The timeout is large so the queue might still fill up, but at least we have a log to inform further debugging.

### DIFF
--- a/store/postgres/src/notification_listener.rs
+++ b/store/postgres/src/notification_listener.rs
@@ -1,6 +1,7 @@
 use crate::functions::pg_notify;
 use diesel::pg::PgConnection;
 use diesel::select;
+use graph::prelude::tokio::sync::mpsc::error::SendTimeoutError;
 use lazy_static::lazy_static;
 use postgres::Notification;
 use postgres::{fallible_iterator::FallibleIterator, Client, NoTls};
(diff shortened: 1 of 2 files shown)


otaviopace (approved):
LGTM

store/postgres/src/notification_listener.rs
```diff
@@ -25,6 +26,15 @@ lazy_static! {
                 )))
             )
             .unwrap_or(Duration::from_secs(300));
+    static ref DB_NOTIFICATION_BROADCAST_TIMEOUT: Duration =
+        env::var("GRAPH_DB_NOTIFICATION_BROADCAST_TIMEOUT")
```
lutter, line 30: Can we just call this `GRAPH_NOTIFICATION_BROADCAST_TIMEOUT`?

lutter (approved):

store/postgres/src/notification_listener.rs
```diff
@@ -25,6 +26,15 @@ lazy_static! {
...
+            .ok()
+            .map(
+                |s| Duration::from_secs(u64::from_str(&s).unwrap_or_else(|_| panic!(
+                    "failed to parse env var GRAPH_DB_NOTIFICATION_BROADCAST_TIMEOUT"
+                )))
+            )
+            .unwrap_or(Duration::from_secs(600));
```
lutter, line 37: I actually think the default should be reduced to 60s - for chain head updates, we already start polling after 30s, and for subscriptions :shrug:

store/postgres/src/notification_listener.rs
leoyvens: Done

store/postgres/src/notification_listener.rs
leoyvens: Makes sense, done

PR #2826, opened 2021-09-28
Include network name in feature detection graphql endpoint

tilacog:


server/index-node/src/resolver.rs
```diff
@@ -286,6 +287,9 @@ where
         let mut response: BTreeMap<String, q::Value> = BTreeMap::new();
         response.insert("features".to_string(), features);
         response.insert("errors".to_string(), errors);
+        if let Some(network_name) = network {
+            response.insert("network".to_string(), q::Value::String(network_name));
```
otaviopace, line 291: Maybe this is a picky comment but I think it would make more sense to move the `q::Value::String` construction to the `Ether::Left` match branch, instead of here, since `features` and `errors` are already created with `q::Value::List` there.

server/index-node/src/resolver.rs
```diff
@@ -286,6 +287,9 @@ where
         let mut response: BTreeMap<String, q::Value> = BTreeMap::new();
         response.insert("features".to_string(), features);
         response.insert("errors".to_string(), errors);
+        if let Some(network_name) = network {
```
otaviopace, line 290: Does it make sense to always return a `network` field, but `null` when `errors` is filled? 🤔 

From my understanding the way it is written in this PR, if the validation worked, the field exists, otherwise it isn't returned.

otaviopace (approved):
💯

PR #2827, opened 2021-09-28
Add Prometheus metrics to monitor the notification queue

lutter:
This adds the following metrics to help monitor the Postgres notification queue:

* `notification_queue_usage`: gauge between 0 and 1 that shows the result of `pg_notification_queue_usage()`, updated every minute
* `notification_queue_sent`: counter (labeled with channel and network) for the number of messages sent; for the `store_events` channel, the network is `none`
* `notification_queue_recvd`: counter (labeled with channel and network) for the number of messages received; for the `store_events` channel, the network is `none`

### DIFF
--- a/store/postgres/src/notification_listener.rs
+++ b/store/postgres/src/notification_listener.rs
@@ -1,6 +1,6 @@
-use crate::functions::pg_notify;
 use diesel::pg::PgConnection;
 use diesel::select;
+use diesel::sql_types::Text;
 use graph::prelude::tokio::sync::mpsc::error::SendTimeoutError;
 use lazy_static::lazy_static;
 use postgres::Notification;
@@ -378,6 +378,10 @@ impl NotificationSender {
         use diesel::ExpressionMethods;
         use diesel::RunQueryDsl;
         use public::large_notifications::dsl::*;
+
+        sql_function! {
+            fn pg_notify(channel: Text, msg: Text)
+        }
 
         let msg = data.to_string();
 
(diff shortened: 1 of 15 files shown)


store/postgres/src/notification_listener.rs
```diff
@@ -342,16 +350,39 @@ impl JsonNotification {
             _ => Err(anyhow!("JSON notifications must be numbers or objects"))?,
         }
     }
+}
+
+pub struct NotificationSender {
```
leoyvens, line 355: could this have `Metrics` in the name? The name implies it actually contains a channel sender.

leoyvens (approved):
Clever to leverage the job runner for the queue metric! I just have a comment on naming.

store/postgres/src/notification_listener.rs
lutter: I kinda didn't want to make metrics gathering the name of this struct, it's meant for sending stuff through PG; the naming problem comes more from the ambiguity of `Sender` where it's not clear whether it refers to something about sending with `tokio` or about sending through PG `NOTIFY`. I don't have a great idea for an alternative name. Maybe just a comment on what this struct is for might help?

leoyvens: It does seem strange for the `self` in the notify function to be something that just holds the metrics, at that point it could just be a free function with the metrics as a normal argument. But if you feel this refactor will be useful as the sender might grow more fields, I've no objection.

PR #2830, opened 2021-10-01
ethereum: Consider 'out of gas' as a deterministic eth_call error

leoyvens:
This would make setting the `GRAPH_ETH_CALL_GAS` env var a major footgun, but still it makes sense to keep it around for debugging purposes. So this introduces the `--unsafe-config` flag, which is required to be able to set `GRAPH_ETH_CALL_GAS` to anything different from the default value.

@otaviopace @azf20

### DIFF
--- a/chain/ethereum/src/ethereum_adapter.rs
+++ b/chain/ethereum/src/ethereum_adapter.rs
@@ -125,14 +125,8 @@ const GETH_ETH_CALL_ERRORS: &[&str] = &[
     "invalid opcode",
     // Ethereum says 1024 is the stack sizes limit, so this is deterministic.
     "stack limit reached 1024",
-    // "out of gas" is commented out because Erigon has not yet bumped the default gas limit to 50
-    // million. It can be added through `GETH_ETH_CALL_ERRORS_ENV` if not using Erigon. Once
-    // https://github.com/ledgerwatch/erigon/pull/2572 has been released and indexers have updated,
-    // this can be uncommented.
-    //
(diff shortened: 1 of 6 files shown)


otaviopace (approved):
LGTM, but maybe we want to merge only after @Jannis looks at this since he mentioned he would be writing a GIP about the two graph-node modes.

That3Percent:
> I think we can get away with an `--unsafe-config` mode for self-hosted and hosted service environments and expect indexers to run in safe mode. @That3Percent do you agree that this would be an ok way to go about standardizing/controlling some of the configuration options to avoid/discourage non-determinism on the network?

In theory there are no problems with config flags. In practice there are a whole bunch of configs and various methods of running `graph-node`. Every single one of them is an opportunity for bugs not found in testing or opportunities for bugs caused by not updating multiple locations in the code. This is being discussed in the File Data Sources GIP [as well](https://github.com/edgeandnode/graph-improvement-proposals/pull/10#discussion_r722570406).

My fear is that developers develop a feature and update one code branch then test using the dev config which hits that branch and then deploy a radically different thing to The Network which is barely tested.

Every change we make should be removing these kinds of footguns but I feel like we're just getting farther away from the goal. There needs to be really good justification for this.

I don't understand why this env variable is necessary and it doesn't seem useful enough to warrant an exception. If you really need to get the gas limit down for testing (presumably to make it a shorter time before testing what happens if you hit the limit) we could introduce a `burnGas(uint)` host export that is deterministic and would accomplish the same. This is a function that could be deployed to The Network without causing any problems and has the advantage of only affecting the current test instead of the entire process.

PR #2831, opened 2021-10-01
Add `subgraphs-for-deployment` explorer route

tilacog:
This PR adds a new explorer HTTP endpoint for querying deployments for a given subgraph Qm-hash.  

The route is: `/explorer/deployments-for-subgraph/<Qm-hash`.

For example, the following request
```
HTTP GET host:8030/explorer/deployments-for-subgraph/QmTKXLEdMD6Vq7Nwxo8XAfnHpG6H1TzL1AGwiqLpoae3Pb
```
would be responded with a list of all subgraph names and their versions _(`current`, `pending` or `unused`)_ for that subgraph identifier:

```json
[
    {
        "name": "subgraph/name",
        "version": "current"
    }, 
    {
        "name": "another-subgraph/name",
        "version": "pending"
    }
]
```

### DIFF
--- a/graph/src/components/store.rs
+++ b/graph/src/components/store.rs
@@ -1398,10 +1398,10 @@ pub trait StatusStore: Send + Sync + 'static {
     ) -> Result<(Option<String>, Option<String>), StoreError>;
 
     /// Support for the explorer-specific API. Returns a vector of (name, version) of all
-    /// deployments for a given subgraph hash.
-    fn deployments_for_subgraph_id(
+    /// subgraphs for a given deployment hash.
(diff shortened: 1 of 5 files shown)


lutter (approved):
Nice!

lutter:
One minor thing: the name of the route is backwards - it should be `subgraphs-for-deployment`: we list the subgraph (names) for a given deployment (hash)

lutter:
The naming should be fixed before merging it; otherwise, it's very confusing and hard to fix afterwards

lutter (approved):
Nice! Looks good.

PR #2835, opened 2021-10-04
Expose transaction.nonce field

### LINKED ISSUE #2796: ability to access event.transaction.nonce
Feature

current behaviour does not allow the mapping to access the transaction nonce

I expect to be able to fetch the transaction nonce from the event object

tilacog:
Resolves #2796


chain/ethereum/src/trigger.rs
```diff
@@ -145,32 +146,32 @@ impl blockchain::MappingTrigger for MappingTrigger {
                 params,
                 handler: _,
             } => {
-                if heap.api_version() >= Version::new(0, 0, 2) {
+                let ethereum_data = EthereumEventData {
```
otaviopace, line 149: As we talked, maybe it's worth changing this to `ethereum_event_data` or just `event_data` 🙂

PR #2836, opened 2021-10-04
Fix how NEAR networks are extracted from the config

maoueh:
NEAR chains were extracted from the config just by using all providers of type Firehose. This is of course quite wrong as Firehose can provide multiple different networks. The net effect was unique constraint violations since NEAR networks that were not actually real NEAR networks were appended to Ethereum networks leading to duplicate (because the NEAR networks should have been empty).

We are introducing a `kind` field on `Chain` in the `ChainSection` config. This `kind` field can then be used to discriminate between Ethereum and NEAR networks and avoid the duplication and the unique constraints issue.


node/src/main.rs
```diff
@@ -220,21 +221,22 @@ async fn main() {
         let mut blockchain_map = BlockchainMap::new();
 
         let (eth_networks, ethereum_idents) = connect_networks(&logger, eth_networks).await;
-        let near_idents = compute_near_network_identifiers(&firehose_networks);
+        let near_idents =
+            compute_near_network_identifiers(firehose_networks_by_kind.get(&BlockchainKind::Near));
 
-        let network_identifiers = ethereum_idents
+        let network_identifiers = dbg!(ethereum_idents)
```
maoueh, line 227: Oups leftover, will remove those

PR #2841, opened 2021-10-05
feat: extend enum filters to allow _in and _not_in filters.

### LINKED ISSUE #2668: Allow "field_in" and "field_not_in" filters for enums
At Balancer we use an [enum to differentiate between different pool types](https://github.com/balancer-labs/balancer-subgraph-v2/blob/53d9974c610a6145c8e8ee5bd854c71d8fab3ba0/schema.graphql#L12-L17) however this means that we can only filter on `poolType: Weighted` or `poolType_not: Stable`, etc. as shown in the snippet below:

https://github.com/graphprotocol/graph-node/blob/383150ef37e36ccecd8d00d5b4671f0b371457b4/graphql/src/schema/api.rs#L385-L425

Ideally we'd also be able to make use of `poolType_in` or `poolType_not_in` for cases in which we want to pull a subset of all the types of pools. The most obvious example being in situations where we may add a new pool type which is not fully supported in a particular service yet.

I don't see any obvious reason that would prevent these filters being applied to enums so is there a reason why this feature doesn't exist?

fubhy:
Fixes #2668


leoyvens:
Nice! We'd also need some tests that the queries return the expected results. `interfaces.rs` might be a good place in which to add that test.

fubhy:
Sure, can do. There are currently no other tests for filters in `interfaces.rs` but there are some in `graphql/tests/query.rs`. Are you sure you want them in `interfaces.rs`?

fubhy:
Added the integration test to `interfaces.rs` now as you suggested.

PR #2843, opened 2021-10-05
graph, firehose: fix panick because of re-use of an already completed stream

maoueh:
The state machine of the FirehoseBlockStream was not properly resetting the state to disconnected on error leading to the usage of an already completed future leading to a panic.

All code paths are not properly returning the state to Disconnected so the stream can be re-created to contact the remote Firehose endpoint.

### DIFF
--- a/graph/src/blockchain/firehose_block_stream.rs
+++ b/graph/src/blockchain/firehose_block_stream.rs
@@ -39,7 +39,7 @@ impl<C: Blockchain, F: FirehoseMapper<C>> Clone for FirehoseBlockStreamContext<C
 }
 
 enum BlockStreamState {
-    Disconneted,
+    Disconnected,
     Connecting(
         Pin<
             Box<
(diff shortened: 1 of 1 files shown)


graph/src/blockchain/firehose_block_stream.rs
```diff
@@ -144,6 +144,8 @@ impl<C: Blockchain, F: FirehoseMapper<C>> Stream for FirehoseBlockStream<C, F> {
 
                         Poll::Ready(Err(e)) => {
                             error!(self.ctx.logger, "Unable to connect to endpoint {}", e);
+                            self.state = BlockStreamState::Disconneted;
```
lukanus, line 147: I think you have a typo in the state ;)

maoueh:
> Manual stream implementations are tricky, exhibit 1 

Of course, it was a pain to implement it, still a pain to maintain it :D Just adding yet another future to wait for in the cycle is a pain (hence the post-poned refactoring to directly use `triggers_in_blocks` and not duplicate it for Firehose).

I've extracted the `FirehoseBlockStream` into a playground code so I can actually play with alternative solution. I started fiddling a bit with https://github.com/tokio-rs/async-stream, will report my feeling about it when I'm more advanced.

PR #2868, opened 2021-10-11
Add support for minimum block constraint

Theodus:
This PR adds support for a minimum block constraint on queries via the `number_gte` block argument. The minimum block constraint will be used by the gateway to reduce the likelihood that a client receives query results from blocks prior to those from past queries from the same client.

### DIFF
--- a/graphql/src/query/ext.rs
+++ b/graphql/src/query/ext.rs
@@ -58,6 +58,8 @@ impl ValueExt for q::Value {
 pub enum BlockConstraint {
     Hash(H256),
     Number(BlockNumber),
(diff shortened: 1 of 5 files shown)


That3Percent (approved):

fubhy:
This introduced a likely unintended change to the `_meta.block` output type. I'd propose to introduce a `_BlockFilter_` or `_BlockInput_` type instead of re-using the `_Block_` for both input and output.

Ping @Theodus

Jannis:
Good catch! I agree with your suggestion.

Theodus:
Thanks @fubhy. See #3006

fubhy:
Thanks for the quick fix.

PR #2878, opened 2021-10-14
Near: Parse receipt triggers and validate data sources

leoyvens:
Also validate the `kind` field for ethereum.


leoyvens:
@otaviopace I'll assume this is approved.

PR #2898, opened 2021-10-20
graph: add enum variant discriminant to `ObjectOrInterface`'s Hash impl

vrmiguel:
Avoids letting `ObjectOrInterface`s of different variants but same underlying name have the same hash


otaviopace:
LGTM, but just a question, do you know which places we use the `Hash` implementation for `ObjectOrInterface`?

I would like to know that to be sure about what this change actually impacts.

PR #2916, opened 2021-10-23
Speed up entity updates and deletions during indexing by remembering the vid of the previous version

lutter:


graph/src/data/store/mod.rs
```diff
@@ -609,6 +609,33 @@ impl CacheWeight for Entity {
...
+    fn from(ev: EntityVersion) -> Self {
+        ev.data
+    }
+}
+
+impl From<Entity> for EntityVersion {
+    fn from(data: Entity) -> Self {
+        Self { data, vid: None }
+    }
+}
```
leoyvens, line 637: Is this impl necessary or useful? It could be error prone to implicitly set the vid to None.

graph/src/components/store.rs
```diff
@@ -1444,16 +1444,20 @@ pub enum EntityModification {
...
     /// Update the entity by overwriting it
-    Overwrite { key: EntityKey, data: Entity },
+    Overwrite {
+        key: EntityKey,
+        data: Entity,
+        prev_vid: Vid,
+    },
     /// Remove the entity
-    Remove { key: EntityKey },
+    Remove { key: EntityKey, prev_vid: Vid },
```
leoyvens, line 1453: The vid type is optional, but deep in the clamp query code it is treated as a required value. Is there a way we could already treat it as required here?

lutter: Yes, I am not happy about that either; the issue is that when mappings create entities, we don't have a `vid` yet; I'll think about how to address that by refining the `EntityVersion` further (or, less excited about that, have the store handle `None` `vid`'s properly)

leoyvens: Got it.

leoyvens (approved):

lutter:
Rebased to latest master

graph/src/components/store.rs
lutter: I looked at this some more, and we'll have to deal with `vid` possibly being `None` in some form until we transact the changes. We can express the transition from '`None` is permissible' to '`None` is not allowed' differently, but I think it will always be a dynamic property and can't be enforced statically.

But AFAICT, what the store does is safe in the sense that a `None` in the wrong place will lead to a subgraph failure, not to corrupted data.

lutter:
Ran this in the integration cluster for ~ 12 hours against `integer` at commit 2345ddf5. Only observed differences were in failed subgraphs reporting one block later than on `integer` which is expected behavior.

PR #2935, opened 2021-10-27
Improve our CacheWeight estimates

lutter:
This PR does the following:

* fix a bug in `CacheWeight.weight` that caused us to assume any struct has size 8
* improve the estimate for `BTreeMap` by following the structure of a B-Tree much more closely; that is now precise for maps up to size about 20
* improve the `stress` example to make it a little more versatile and representative of real-life data that we cache

### DIFF
--- a/graph/examples/stress.rs
+++ b/graph/examples/stress.rs
@@ -145,6 +145,29 @@ impl Template<ValueMap> for ValueMap {
     }
 }
 
+type UsizeMap = BTreeMap<usize, usize>;
+
+/// Template for testing roughly a GraphQL response, i.e., a `BTreeMap<String, Value>`
+impl Template<UsizeMap> for UsizeMap {
+    type Item = UsizeMap;
+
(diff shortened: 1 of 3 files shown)


graph/examples/stress.rs
```diff
@@ -500,35 +500,32 @@ pub fn main() {
...
-        //    3      |     16.51
-        //    5      |     12.07
-        //   10      |      4.64
-        //   50      |      3.07
-        //  100      |      2.94
+        //    3      |      1.3
+        //    5      |      1.5
+        //   10      |      1.5
+        //   50      |      1.2
+        //  100      |      0.9
```
leoyvens, line 525: Yay!

leoyvens (approved):

leoyvens:
I suppose that if we manage to switch our representation of a graphql response to a `Vec`, that will help with the troubles of estimating the memory usage of a BTree.

PR #2946, opened 2021-10-28
Report nondeterministic errors as unattestable

Theodus:
This change sets the Graph-Attestable header on query results to signal
to the indexer-service whether it should include an attestation for the
client. Unattestable responses are those that are nondeterministic,
according to [GIP 20](https://hackmd.io/@CGmldHLrTRaYRnAva-HoPQ/S1W2vMjEt).

### DIFF
--- a/graph/src/data/query/error.rs
+++ b/graph/src/data/query/error.rs
@@ -83,6 +83,28 @@ pub enum QueryExecutionError {
     ResultTooBig(usize, usize),
 }
 
+impl QueryExecutionError {
+    pub fn is_attestable(&self) -> bool {
+        use self::QueryExecutionError::*;
+        match self {
+            ResolveEntitiesError(_)
+            | Timeout
+            | TooComplex(_, _)
+            | TooDeep(_)
+            | IncorrectPrefetchResult { .. }
+            | Panic(_)
+            | EventStreamError
+            | TooExpensive
+            | Throttled
+            | DeploymentReverted
+            | SubgraphManifestResolveError(_)
+            | InvalidSubgraphManifest
+            | ResultTooBig(_, _) => false,
+            _ => true,
+        }
+    }
+}
+
 impl Error for QueryExecutionError {
     fn description(&self) -> &str {
         "Query execution error"
@@ -217,7 +239,7 @@ impl fmt::Display for QueryExecutionError {
             EventStreamError => write!(f, "error in the subscription event stream"),
             FulltextQueryRequiresFilter => write!(f, "fulltext search queries can only use EntityFilter::Equal"),
             TooExpensive => write!(f, "query is too expensive"),
-            Throttled=> write!(f, "service is overloaded and can not run the query right now. Please try again in a few minutes"),
+            Throttled => write!(f, "service is overloaded and can not run the query right now. Please try again in a few minutes"),
             DeploymentReverted => write!(f, "the chain was reorganized while executing the query"),
             SubgraphManifestResolveError(e) => write!(f, "failed to resolve subgraph manifest: {}", e),
             InvalidSubgraphManifest => write!(f, "invalid subgraph manifest file"),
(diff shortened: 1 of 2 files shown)


graph/src/data/query/error.rs
```diff
@@ -83,6 +83,63 @@ pub enum QueryExecutionError {
...
 
+impl QueryExecutionError {
+    pub fn is_attestable(&self) -> bool {
+        use self::QueryExecutionError::*;
+        match self {
+            OperationNameRequired
+            | OperationNotFound(_)
+            | NotSupported(_)
+            | NoRootSubscriptionObjectType
+            | NonNullError(_, _)
```
Jannis, line 94: You would expect this to be consistent across indexers but let's say one of them wipes one of the entity tables of a subgraph and suddenly all references that are supposed to be non-null are `null`... then this _could_ be non-deterministic. Therefor I'd move it into the unattestable category.

graph/src/data/query/error.rs
```diff
@@ -83,6 +83,63 @@ pub enum QueryExecutionError {
...
+            | ListValueError(_, _)
```
Jannis, line 95: I _think_ a similar argument could apply to this error type, so out of caution I'd make it non-deterministic for now.

graph/src/data/query/error.rs
```diff
@@ -83,6 +83,63 @@ pub enum QueryExecutionError {
...
+            | InvalidVariableTypeError(_, _)
+            | MissingVariableError(_, _)
+            | OrderByNotSupportedError(_, _)
+            | OrderByNotSupportedForType(_)
+            | FilterNotSupportedError(_, _)
+            | UnknownField(_, _, _)
+            | EmptyQuery
+            | MultipleSubscriptionFields
+            | SubgraphDeploymentIdError(_)
+            | RangeArgumentsError(_, _, _)
```
Jannis, line 109: At least right now it's possible for indexers to configure different limits for pagination arguments (`first` and `skip`). Until these are guaranteed to be identical everywhere, we have to make this unattestable.

graph/src/data/query/error.rs
```diff
@@ -83,6 +83,63 @@ pub enum QueryExecutionError {
...
+            | InvalidFilterError
+            | EntityFieldError(_, _)
+            | ListTypesError(_, _)
+            | ListFilterError(_)
+            | ValueParseError(_, _)
```
Jannis, line 114: This one can happen when there are bad values in the db. _Should_ be deterministic but isn't guaranteed to be.

graph/src/data/query/error.rs
```diff
@@ -83,6 +83,63 @@ pub enum QueryExecutionError {
...
+            | AttributeTypeError(_, _)
+            | EntityParseError(_)
```
Jannis, line 116: This can happen if there is bad data in the db, e.g. corrupt entities. Should be unattestable.

graph/src/data/query/error.rs
```diff
@@ -83,6 +83,63 @@ pub enum QueryExecutionError {
...
+            | EmptySelectionSet(_)
+            | AmbiguousDerivedFromResult(_, _, _, _)
```
Jannis, line 118: Also based on data in the db and not guaranteed to always be identical. Unattestable.

graph/src/data/query/error.rs
```diff
@@ -83,6 +83,63 @@ pub enum QueryExecutionError {
...
+            | Unimplemented(_)
+            | EnumCoercionError(_, _, _, _, _)
```
Jannis, line 120: Based on data in the db and not guaranteed to always be identical. Unattestable.

graph/src/data/query/error.rs
```diff
@@ -83,6 +83,63 @@ pub enum QueryExecutionError {
...
+            | ScalarCoercionError(_, _, _, _)
```
Jannis, line 121: Based on data in the db and not guaranteed to always be identical. Unattestable.

Jannis (changes requested):
Looking great!

Jannis (approved):
LGTM!

PR #2956, opened 2021-11-06
store: Guard inserts into the call cache with a check

lutter:
This check mimics the check for inserts into call_meta and also aims to
reduce lock wait times for busy chains


leoyvens:
Looks like there is a test that needs fixing.

lutter:
I actually don't think this is needed anymore; closing it for now. If it should seem like it would be useful again, I'll revisit.

PR #2963, opened 2021-11-08
Serialize the BigInt bytes to the correct endianess expected by BigInt

### LINKED ISSUE #2955: gasPrice for NEAR blocks is incorrect
The gasPrice entity is a lot higher in the subgraph than what docs + NEAR’s explorer db shows:
image

It should be 100000 by default and for most of the blocks (e.g. [this block](https://explorer.near.org/blocks/HyWy4uJyaU9yg5oYMhQ5pKrmBoXhcQtLKVmfZhhChynk))

Replicated here: https://thegraph.com/hosted-service/subgraph/azf20/near-blocks

maoueh:
The received bytes from the Firehose NEAR blocks are not in the endianess expected directly by the `BigInt` within AssemblyScript. To fix the issue, we simply reverse the bytes prior turning them into the `Uint8Array` AscObj which ultimately get written to the WASM memory.

While the original issue listed `gasPrice` a being problematic, the bug actually affected all `BigInt` types coming from the Firehose NEAR blocks which means quite a lot of fields.

All NEAR subgraphs using such affected fields will need to be re-synced from their `startBlock`.

Fixes #2955

### DIFF
--- a/chain/near/src/runtime/abi.rs
+++ b/chain/near/src/runtime/abi.rs
@@ -618,8 +618,9 @@ impl ToAscObj<Uint8Array> for codec::BigInt {
         &self,
         heap: &mut H,
     ) -> Result<Uint8Array, DeterministicHostError> {
-        // FIXME (NEAR): Hopefully the bytes here fits the BigInt format expected in `graph-node`,
-        //               will need to validate that in the subgraph directly.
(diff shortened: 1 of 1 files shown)


leoyvens (approved):
Good catch!

PR #2970, opened 2021-11-11
Update `db_version` to version 3 at startup

tilacog:
Adds a startup operation to check if `db_version` is less than `3`.

If not, performs full cleanup on all block cache tables _(using `TRUNCATE TABLE`)_ and then sets the `db_version` to `3`.

### DIFF
--- a/store/postgres/src/block_store.rs
+++ b/store/postgres/src/block_store.rs
@@ -463,7 +463,7 @@ impl BlockStore {
         Ok(())
     }
 
-    pub fn truncate_block_caches(&self) -> Result<(), StoreError> {
+    fn truncate_block_caches(&self) -> Result<(), StoreError> {
         for (_chain, store) in &*self.stores.read().unwrap() {
             store.truncate_block_cache()?
         }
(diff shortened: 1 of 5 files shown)


tilacog:
Looking into failed tests

store/postgres/src/block_store.rs
```diff
@@ -462,6 +462,28 @@ impl BlockStore {
...
+            store.truncate_block_cache()?
+        }
+        Ok(())
+    }
+
+    pub fn update_db_version(&self) -> Result<(), StoreError> {
+        use crate::primary::db_version as dbv;
+        use diesel::prelude::*;
+
+        let connection = self.mirror.primary().get()?;
```
lutter, line 477: You really want the primary here, not the mirror. The mirror reads from other shards if the primary is down, but `db_version` is not copied to other shards; so something like `self.pools.get(&*PRIMARY_SHARD).unwrap().get()?`

lutter (approved):
Looks great! Just make the two small changes before committing. Also, add something to `NEWS` telling people that their block cache will be truncated and that that is ok.

Also, before committing, make sure to bump the `db_version` in the hosted service (and probably in staging, too) so we don't do the wrong thing there. Bump it in the primary for sure, but to be extra safe, maybe in all the shards, too

lutter:
> Also, before committing, make sure to bump the `db_version` in the hosted service (and probably in staging, too)

On second thought, maybe do not change staging so that this truncates and we see how the system reacts

PR #2972, opened 2021-11-12
docker: Wait for IPFS to be up even with a config file

lutter:


docker/start
```diff
@@ -54,6 +54,8 @@ run_graph_node() {
     then
         # Start with a configuration file; we don't do a check whether
         # postgres is up in this case, though we probably should
+        wait_for_ipfs "$ipfs"
```
otaviopace, line 57: It's not clear for me why this is being added, and if we should add for postgres as well (given the comment above).

Could you please explain/elaborate on the subject? 🙂

lutter: This mimics what we do without a config file so that `graph-node` doesn't blow up when IPFS isn't up yet (e.g., in docker compose) We should also check that for PG, but that would require parsing the config file. This is a bandaid that will make things a bit better, but a real fix would require to pull all this wait functionality into `graph-node` which is a bigger expedition.

otaviopace (approved):

PR #2974, opened 2021-11-15
chain head listener: Used timed rwlock

leoyvens:
To rule out the possibility of a deadlock, though I don't believe this is the case. This also changes the TimedRwLock to not block forever on read/write, so that it logs even in the case of deadlock.


leoyvens:
I've also added a log around `watcher.send();` so we can confirm whether it's deadlocking, which is my real suspicion.

otaviopace (approved):
LGTM

PR #2976, opened 2021-11-16
Elasticsearch retries

### LINKED ISSUE #2880: Elasticsearch retries and configurable flush interval
Retries for 429 rate limiting and the possibility of increasing the flush interval through the environment would help dealing with high load on ES.

leoyvens:
Resolves #2880.


graph/src/log/factory.rs
```diff
@@ -11,6 +11,12 @@ lazy_static::lazy_static! {
         .unwrap_or("5".into())
         .parse::<u64>()
         .expect("invalid GRAPH_ELASTIC_SEARCH_FLUSH_INTERVAL_SECS"));
+
+    static ref ES_MAX_RETRIES: usize =
+        std::env::var("GRAPH_ELASTIC_SEARCH_MAX_RETRIES")
+        .unwrap_or("5".into())
+        .parse::<usize>()
+        .expect("invalid GRAPH_ELASTIC_SEARCH_FLUSH_INTERVAL_SECS");
```
otaviopace, line 19: It should be `GRAPH_ELASTIC_SEARCH_MAX_RETRIES` here

graph/src/log/elastic.rs
```diff
@@ -272,19 +276,24 @@ impl ElasticDrain {
...
-                    .and_then(|response| async { response.error_for_status() })
-                    .map_ok(|_| ())
-                    .unwrap_or_else(move |e| {
+
+                retry("send logs to elasticsearch", &flush_logger)
+                    .limit(max_retries)
+                    .timeout_secs(30)
+                    .run(move || {
+                        header
+                            .try_clone()
```
otaviopace, line 285: I don't understand why there's the need for the `clone` here

graph/src/log/elastic.rs
```diff
@@ -272,19 +276,24 @@ impl ElasticDrain {
...
+                            .unwrap() // Unwrap: Request body not yet set
```
otaviopace, line 286: I don't understand this comment.

What's the relation between the `unwrap` and not having the `body` set yet?

I remember one of the times I've used `reqwest` there was some kind of "catch" when unwraping the body, but I think it was on the `response` part. Is this related to it?

I took a quick look through the [docs](https://docs.rs/reqwest/0.11.6/reqwest/) and some old code I've made using the lib and I still don't understand.

graph/src/log/elastic.rs
leoyvens: This closure may be called multiple times, one for each retry, so it cannot move a captured variable.

graph/src/log/elastic.rs
leoyvens: See https://docs.rs/reqwest/0.11.6/reqwest/struct.RequestBuilder.html#method.try_clone

leoyvens:
@otaviopace addressed the comments.

PR #2979, opened 2021-11-17
fix firehose UNDO case on old blocks

sduchesneau:
When firehose sends an undo event from an older block, the
instance-manager would try to fetch the block from RPC to
retrieve its parent block info.

Now, the firehose will include the parent's hash and num in
the Revert event to remove that extraneous RPC call

### DIFF
--- a/chain/ethereum/src/chain.rs
+++ b/chain/ethereum/src/chain.rs
@@ -597,13 +597,10 @@ impl FirehoseMapperTrait<Chain> for FirehoseMapper {
                     number: block.number as i32,
                 },
                 FirehoseCursor::Some(response.cursor.clone()),
-                match block.header {
-                    Some(header) => Some(BlockPtr {
-                        hash: BlockHash::from(header.parent_hash),
-                        number: (block.number as i32) - 1,
-                    }),
-                    None => None,
-                },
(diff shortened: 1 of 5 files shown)


chain/ethereum/src/chain.rs
```diff
@@ -597,6 +597,13 @@ impl FirehoseMapperTrait<Chain> for FirehoseMapper {
                     number: block.number as i32,
                 },
                 FirehoseCursor::Some(response.cursor.clone()),
+                match block.header {
+                    Some(header) => Some(BlockPtr {
+                        hash: BlockHash::from(header.parent_hash),
+                        number: (block.number as i32) - 1,
```
leoyvens, line 603: The ideal way of dealing with overflow is to use [checked_sub](https://doc.rust-lang.org/std/primitive.u32.html#method.checked_sub), and if you're sure it's impossible for `block.number` to be 0, then you can use unwrap and comment on why it is safe.

sduchesneau: fixed, thanks! 
also @maoueh mentionned that the unwrap() on header is also safe (block.header is always populated)

leoyvens (approved):

PR #2982, opened 2021-11-18
chain head listener: workaround watcher deadlock

leoyvens:
With the logs we were able to confirm that this deadlocks within the tokio watcher implementation, this has been reported upstream https://github.com/tokio-rs/tokio/issues/4246. This PR hacks around that until it is fixed by detecting a deadlocked watcher and continuing to pull notifications, if only to discard them.

### DIFF
--- a/store/postgres/src/chain_head_listener.rs
+++ b/store/postgres/src/chain_head_listener.rs
@@ -7,9 +7,9 @@ use graph::{
     prometheus::{CounterVec, GaugeVec},
     util::timed_rw_lock::TimedRwLock,
 };
-use std::str::FromStr;
-use std::sync::Arc;
+use std::sync::{atomic::AtomicBool, Arc};
 use std::{collections::BTreeMap, time::Duration};
+use std::{str::FromStr, sync::atomic};
 
 use lazy_static::lazy_static;
 
@@ -38,16 +38,20 @@ lazy_static! {
 }
 
 struct Watcher {
-    sender: watch::Sender<()>,
+    sender: Arc<watch::Sender<()>>,
     receiver: watch::Receiver<()>,
 }
 
 impl Watcher {
     fn new() -> Self {
         let (sender, receiver) = watch::channel(());
-        Watcher { sender, receiver }
+        Watcher {
+            sender: Arc::new(sender),
+            receiver,
+        }
     }
 
+    #[allow(dead_code)]
     fn send(&self) {
         // Unwrap: `self` holds a receiver.
         self.sender.send(()).unwrap()
@@ -147,6 +151,7 @@ impl ChainHeadUpdateListener {
     ) {
         // Process chain head updates in a dedicated task
         graph::spawn(async move {
+            let sending_to_watcher = Arc::new(AtomicBool::new(false));
             while let Some(notification) = receiver.recv().await {
                 // Create ChainHeadUpdate from JSON
                 let update: ChainHeadUpdate =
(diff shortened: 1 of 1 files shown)


store/postgres/src/chain_head_listener.rs
```diff
@@ -173,9 +178,19 @@ impl ChainHeadUpdateListener {
...
                 if let Some(watcher) = watchers.read(&logger).get(&update.network_name) {
-                    debug!(logger, "sending chain head update"; "network" => &update.network_name);
-                    watcher.send();
-                    debug!(logger, "chain head update sent"; "network" => &update.network_name);
+                    // Due to a tokio bug, we must assume that the watcher can deadlock, see
+                    // https://github.com/tokio-rs/tokio/issues/4246.
+                    if !sending_to_watcher.load(atomic::Ordering::SeqCst) {
+                        let sending_to_watcher = sending_to_watcher.cheap_clone();
+                        let sender = watcher.sender.cheap_clone();
+                        tokio::task::spawn_blocking(move || {
```
lutter, line 186: Can that lead to a large number of dead threads hanging around? Maybe we need another `watcher_is_live: Arc<AtomicBool>` that is set to `false` just before the call to send and set to `true` right afterwards where we do not try to send if `watcher_is_live` is false.

lutter (changes requested):

store/postgres/src/chain_head_listener.rs
```diff
@@ -173,9 +178,19 @@ impl ChainHeadUpdateListener {
 
                 // If there are subscriptions for this network, notify them.
                 if let Some(watcher) = watchers.read(&logger).get(&update.network_name) {
-                    debug!(logger, "sending chain head update"; "network" => &update.network_name);
-                    watcher.send();
```
otaviopace, line 177: We didn't `unwrap` this call before?

store/postgres/src/chain_head_listener.rs
```diff
@@ -173,9 +178,19 @@ impl ChainHeadUpdateListener {
...
-                    debug!(logger, "chain head update sent"; "network" => &update.network_name);
+                    // Due to a tokio bug, we must assume that the watcher can deadlock, see
+                    // https://github.com/tokio-rs/tokio/issues/4246.
+                    if !sending_to_watcher.load(atomic::Ordering::SeqCst) {
+                        let sending_to_watcher = sending_to_watcher.cheap_clone();
+                        let sender = watcher.sender.cheap_clone();
+                        tokio::task::spawn_blocking(move || {
+                            sending_to_watcher.store(true, atomic::Ordering::SeqCst);
+                            sender.send(()).unwrap();
```
otaviopace, line 188: I see that `watcher.send()` didn't receive any arguments before, why is it receiving a unit `()` value now?

leoyvens: Before we were using the `send` on our watcher struct, which is now dead code until we can revert this hack.

store/postgres/src/chain_head_listener.rs
leoyvens: I believe the `sending_to_watcher` is already serving as the check you propose.

leoyvens:
@lutter @otaviopace thanks for the review, comments addressed.

store/postgres/src/chain_head_listener.rs
lutter: Yes, of course, you are right.

lutter (approved):

Issue #2945, opened 2021-10-28
Add retry loop to Graph Node for failed subgraphs


azf20:
Improvement over: https://github.com/graphprotocol/graph-node/issues/2932

This applies for non-deterministic failures.

otaviopace:
This will help a lot: https://github.com/graphprotocol/graph-node/blob/master/graph/src/util/backoff.rs

@lutter made this a month ago.

azf20:
Solved in https://github.com/graphprotocol/graph-node/pull/2988

PR #2988, opened 2021-11-19
Add exponential backoff retry for non deterministic errors

otaviopace:
After extensive testing, I think this is finally correct 🙌 

This PR adds an exponential backoff to keep retrying non-deterministic errors until they get solved and the subgraph unfails.

Main changes:

- Re add block range confidence check for non-determinstic errors on unfailure. Aka reversal and fix of this PR: https://github.com/graphprotocol/graph-node/pull/3015
- Use the `ExponentialBackoff` structure to keep retrying non-deterministic errors

Solves #2945

### DIFF
--- a/core/src/subgraph/instance_manager.rs
+++ b/core/src/subgraph/instance_manager.rs
@@ -8,7 +8,7 @@ use graph::data::store::EntityVersion;
 use graph::data::subgraph::{UnifiedMappingApiVersion, MAX_SPEC_VERSION};
 use graph::prelude::TryStreamExt;
 use graph::prelude::{SubgraphInstanceManager as SubgraphInstanceManagerTrait, *};
-use graph::util::lfu_cache::LfuCache;
+use graph::util::{backoff::ExponentialBackoff, lfu_cache::LfuCache};
 use graph::{blockchain::block_stream::BlockStreamMetrics, components::store::WritableStore};
 use graph::{blockchain::block_stream::BlockWithTriggers, data::subgraph::SubgraphFeature};
 use graph::{
@@ -27,8 +27,10 @@ use graph::{
 use lazy_static::lazy_static;
 use std::collections::{BTreeSet, HashMap};
 use std::sync::{Arc, RwLock};
-use std::time::Instant;
+use std::time::{Duration, Instant};
 use tokio::task;
+
+const MINUTE: Duration = Duration::from_secs(60);
 
 lazy_static! {
     /// Size limit of the entity LFU cache, in bytes.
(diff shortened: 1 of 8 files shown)


core/src/subgraph/instance_manager.rs
```diff
@@ -460,6 +462,10 @@ where
     let id_for_err = ctx.inputs.deployment.hash.clone();
     let mut first_run = true;
 
+    // Exponential backoff that starts with two minutes and keeps
+    // increasing its timeout exponentially until it reaches 48 hours total, aka two days.
+    let mut backoff = ExponentialBackoff::new(MINUTE * 2, MINUTE * 60 * 24 * 2);
```
leoyvens, line 467: The intention is that the subgraph self-heals _soon_ after the underlying issue is fixed, even if it takes hours or days for the issue to be fixed. 48h is too long to be a plausible definition of _soon_, I'd make the default maximum maybe 30 minutes.  Either way this maximum should be configurable.

core/src/subgraph/instance_manager.rs
```diff
@@ -675,6 +680,24 @@ where
...
+                    if !deterministic {
+                        // Cancel the stream for real
+                        ctx.state
+                            .instances
+                            .write()
+                            .unwrap()
+                            .remove(&ctx.inputs.deployment.id);
+
+                        // Sleep before restarting
+                        error!(logger, "Subgraph failed for non-deterministic reasons: {}", e;
```
leoyvens, line 692: Lets use `error` instead of `reasons`.

leoyvens (changes requested):

core/src/subgraph/instance_manager.rs
otaviopace: Fixed!

core/src/subgraph/instance_manager.rs
```diff
@@ -42,6 +44,14 @@ lazy_static! {
     // Used for testing Graph Node itself.
     pub static ref DISABLE_FAIL_FAST: bool =
         std::env::var("GRAPH_DISABLE_FAIL_FAST").is_ok();
+
+    /// Ceiling for the backoff retry of non-deterministic errors, in milliseconds.
+    pub static ref BACKOFF_RETRY_CEILING_FOR_NON_DETERMINISTIC_ERRORS: Duration =
```
otaviopace, line 49: @leoyvens also if you have any suggestion that's smaller than this one, I would love to hear it 😅

core/src/subgraph/instance_manager.rs
leoyvens: Retrying forever is the desirable behavior for the subgraph to be able to self-heal even if it takes days for the underlying issue to be fixed.

core/src/subgraph/instance_manager.rs
leoyvens: Maybe `SUBGRAPH_ERROR_RETRY_CEIL`, and the env var should reflect the unit. I don't think millisecond is a useful granularity, I'd prefer seconds so `MAX_SUBGRAPH_ERROR_RETRY__CEIL_SECS`.

store/postgres/src/deployment_store.rs
```diff
@@ -1352,4 +1347,12 @@ impl DeploymentStore {
                   "shard" => self.pool.shard.as_str())
         });
     }
+
+    pub(crate) fn health(
+        &self,
+        id: &DeploymentHash,
+    ) -> Result<deployment::SubgraphHealth, StoreError> {
+        let conn = self.get_conn()?;
+        deployment::health(&conn, id)
```
leoyvens, line 1356: This is a blocking function called from an async context, we should make it async by using `with_conn`.

leoyvens (changes requested):

store/postgres/src/deployment_store.rs
leoyvens: Code looks fine, thanks for addressing this!

leoyvens (approved):

PR #2995, opened 2021-11-23
Local Graph node debug tool using a remote GraphQL endpoint (per LimeChain)

VIVelev:
This PR implements the Debug Tool discussed with LimeChain.

## What is it?
A way to debug your failed subgraph at block *X* without needing to wait to sync up to block *X*.

## How it is done?
It uses a remote Graph node GraphQL endpoint that is guaranteed to have the subgraph indexed up to block *X* in order to provide the locally-deployed subgraph being debugged a synced-up store.

## What changes have been made?
 - [store_get](https://github.com/LimeChain/graph-node/blob/2c5ca559dad2ec9253305dd0bfbd67a1cb23baf1/runtime/wasm/src/module/mod.rs#L821) in `runtime/wasm/src/module/mod.rs` has been changed to resort to the endpoint when needed.
 - There is a new file: [runtime/wasm/src/module/debug_tool.rs](https://github.com/LimeChain/graph-node/blob/master/runtime/wasm/src/module/debug_tool.rs), that's were the main logic is.
 - The `subgraphs.subgraph_manifest` PostgreSQL table has been altered to include a `debug_endpoint` string, nullable column to store the endpoint's address.
 - The `SubgraphStore` trait and related have been modified to enable easy retrieval of `debug_endpoint` when needed. (most prominently a new method [get_debug_endpoint](https://github.com/LimeChain/graph-node/blob/2c5ca559dad2ec9253305dd0bfbd67a1cb23baf1/graph/src/components/store.rs#L912) has been added)
 - New JSON-RPC handler ([debug_handler](https://github.com/LimeChain/graph-node/blob/2c5ca559dad2ec9253305dd0bfbd67a1cb23baf1/server/json-rpc/src/lib.rs#L126)) to communicate with graph-cli's `graph debug`
 - `SubgraphRegistrar`'s [create_subgraph_version](https://github.com/LimeChain/graph-node/blob/2c5ca559dad2ec9253305dd0bfbd67a1cb23baf1/graph/src/components/subgraph/registrar.rs#L29) has been altered to accommodate the 2 new params: `debug_endpoint` and `debug_block_number`.

## Points to discuss
 - Should `create_subgraph_version` be split into 2 separate functions - one for deploy only and one for debug only?


otaviopace:
I see there are a lot of changes in the imports (ordering, position, etc), is this being by your IDE somehow?

VIVelev:
Actually those changes were not made by me, so I am not sure. They are coming from https://github.com/graphprotocol/graph-node/pull/2995/commits/9b0cdcfb716100519fc339277d7a52298ea21012. Judging by the commit message, I suppose they are caused by `Rustfmt`?

cc: @axiomatic-aardvark

otaviopace:
@VIVelev I don't think it's `rustfmt` because we have it running on CI, if the `master` branch has incorrect formatting, the CI wouldn't pass.

otaviopace (changes requested):
Nice PR! I commented on suggestions and required changes 🙂

runtime/wasm/src/module/debug_tool.rs
```diff
@@ -0,0 +1,121 @@
...
+
+    let query = Query {
+        query: get_query_string(&entity_type.to_lowercase(), &fields),
+        variables: Variables { id: id.to_string() },
+    };
+    return Ok((query, fields));
+}
+
+fn send(query: &Query, endpoint: &str) -> Result<String> {
+    let client = reqwest::blocking::Client::new();
```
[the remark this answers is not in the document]
VIVelev: Yes, although I just thought that we might eliminate the dependency on the `blocking` feature of `reqwest` if we simply use `block_on` in `fetch_entity` and make `send` async. What do you think?

Regarding performance improvements, these post requests can actually be cached, persistently, so if the debugging session is long, we don't spam the GraphQL endpoint.

runtime/wasm/Cargo.toml
```diff
@@ -23,3 +23,5 @@ anyhow = "1.0"
 wasmtime = "0.27.0"
 defer = "0.1"
 never = "0.1"
+serde = "1.0.114"
+serde_json = "1.0.56"
```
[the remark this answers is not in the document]
VIVelev: So, I think we can go without `serde_json`, but when it comes to `serde` the `#[derive(Serialize)]` macro is causing problems, since in its expansion there is the statement: `extern crate serde as _serde;` which to my knowledge requires `serde` to be a dependency for the crate (although my knowledge is limited, since I am pretty new to rust hah). Do you have any idea?

A work around might be to implement the `serde::Serialize` trait by hand for `Query` and `Variables` in *debug_tool.rs*, which will make the code a bit more verbose, but it will eliminate the overlapping dependency on serde.

Yet another solution will be to change the macro implementation in `serde_derive`, but I'm not sure if that's an option.

core/src/subgraph/registrar.rs
```diff
@@ -530,8 +543,21 @@ async fn create_subgraph_version<C: Blockchain, S: SubgraphStore, L: LinkResolve
...
+
+    // Overwrite the start block if the debug block number is set
+    if let Some(number) = debug_block_number {
+        start_block = chain
+            .block_pointer_from_number(&logger, number - 1)
+            .await
+            .map(Some)
+            .map_err(move |_| {
+                SubgraphRegistrarError::ManifestValidationError(vec![
+                    SubgraphManifestValidationError::BlockNotFound(number.to_string()),
```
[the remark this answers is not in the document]
VIVelev: I followed how you guys have done it in [core/src/subgraph/registrari.rs](https://github.com/graphprotocol/graph-node/blob/master/core/src/subgraph/registrar.rs#L454). I was also wondering why do you need to start indexing from one block behind the min `startBlock`?

runtime/wasm/src/module/debug_tool.rs
```diff
@@ -0,0 +1,121 @@
...
+    let entity = Entity::from(map);
+    return Ok(entity);
+}
+
+pub fn fetch_entity(
+    schema: &Arc<Schema>,
+    endpoint: &str,
+    entity_type: &str,
+    id: &str,
+) -> Result<Option<Entity>> {
```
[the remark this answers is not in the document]
VIVelev: It did indeed :D

runtime/wasm/src/module/mod.rs
```diff
@@ -825,12 +827,41 @@ impl<C: Blockchain> WasmInstanceContext<C> {
...
-        let entity_ptr = asc_get(self, entity_ptr)?;
-        let id_ptr = asc_get(self, id_ptr)?;
-        let entity_option =
-            self.ctx
-                .host_exports
-                .store_get(&mut self.ctx.state, entity_ptr, id_ptr)?;
+
+        let entity_type: String = asc_get(self, entity_ptr)?;
+        let id: String = asc_get(self, id_ptr)?;
+        let mut entity_option = self.ctx.host_exports.store_get(
```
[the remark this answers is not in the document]
VIVelev: Yes! I like that! Thanks :D

runtime/wasm/src/module/debug_tool.rs
```diff
@@ -0,0 +1,121 @@
...
+    let entity: Option<&ObjectType> = schema.document.definitions.iter().find_map(|def| {
+        if let Definition::TypeDefinition(TypeDefinition::Object(o)) = def {
+            if o.name == entity_type {
+                Some(o)
+            } else {
+                None
+            }
+        } else {
+            None
+        }
```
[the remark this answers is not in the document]
VIVelev: Wow, nice! Didn't even know you could use if statements like that.

runtime/wasm/src/module/mod.rs
```diff
@@ -840,7 +843,28 @@ impl<C: Blockchain> WasmInstanceContext<C> {
...
+                        // TODO: Better error handling.
+                        HostExportError::Unknown(anyhow!("Failed to fetch entity from fork: {}", e))
+                    })?;
+                    match entity {
+                        Some(entity) => {
+                            // TODO: Probably change the name of the section?
+                            let _section = self
+                                .host_metrics
+                                .stopwatch
+                                .start_section("store_get_asc_new");
```
otaviopace, line 858: Personally I would change this section to `store_get_debug_fork__asc_new`, perhaps @leoyvens or @Jannis can give a suggestion.

store/postgres/src/fork.rs
```diff
@@ -0,0 +1,347 @@
...
+/// associated with a GraphQL endpoint at `fork_url`.
+///
+/// Since this mechanism is used for debug forks, entities are
+/// fetched only once per id in order to avoid fetching an entity
+/// that was deleted from the local store and thus causing inconsistencies.
+pub(crate) struct SubgraphFork {
+    client: reqwest::Client,
+    endpoint: Url,
+    schema: Arc<Schema>,
+    fetched_ids: Mutex<HashSet<String>>,
```
otaviopace, line 42: The implementation of the fetching mechanism seems correct, @Jannis do you see any problem with using a `Mutex` here? 🤔

graph/src/data/value.rs
```diff
@@ -170,6 +170,32 @@ impl TryFrom<q::Value> for Value {
     }
 }
 
+impl From<serde_json::Value> for Value {
```
otaviopace, line 173: I'm confused why there's the need for two `Value`s in the `fork.rs` file, is this really needed or could we just do one JSON transformation?

store/postgres/src/fork.rs
```diff
@@ -0,0 +1,347 @@
...
+            return Ok(None);
+        }
+        fids.insert(id.clone());
+
+        info!(self.logger, "Fetching entity from {}", &self.endpoint; "entity_type" => &entity_type, "id" => &id);
+
+        // NOTE: Subgraph fork compatability checking (similar to the grafting compatability checks)
+        // will be added in the future (in a separate PR).
+        // Currently, forking incompatible subgraphs is allowed, but, for example, storing the
+        // incompatible fetched entities in the local store results in an error.
```
otaviopace, line 65: Typo `compatability` -> `compatibility`

otaviopace (changes requested):
Hey @VIVelev thanks for fixing the issues and restructuring your PR ❤️ 
I'm really sorry it took so long to look over it again.

The only two points I'm not 100% sure about are:

1. If the `Mutex` is the best way to handle the already saved/fetched ids
2. I see that you just pass a `DeploymentHash` through the `graph-cli` (`graph deploy`, [CLI PR for reference](https://github.com/graphprotocol/graph-cli/pull/785/files)) and no block number like grafting. Even though it makes sense in the scenario of a failing subgraph because the block number will be the last one (the one that failed), if someone wants to debug-fork for some reason in a different point in time, this PR doesn't allow that. I think you could add this for better flexibility of the feature.

But before changing any of the two points above, I would like to know @Jannis opinion 😊

runtime/wasm/src/module/mod.rs
leoyvens: imo we can keep the same section name, probably nobody is going to benchmark graph node with a fork anyways.

graph/src/data/value.rs
VIVelev: The field types supported by `serde_json::Value` and the `Value` used in the subgraph store are different, so there needs to be a mechanism in which to convert from one to the other. The above `Value` and the `From<serde_json::Value>` provide this mechanism.

VIVelev:
Regarding `2.`, I decided to leave the block number out of the CLI, since the user can pass it through `dataSources.source.startBlock` in the subgraph manifest. It seems like adding `startBlock` to the CLI duplicates functionality and makes things confusing imo.

leoyvens:
https://github.com/graphprotocol/graph-node/pull/2414 likely generated conflicts for this, lmk if you have issues when rebasing.

VIVelev:
> #2414 likely generated conflicts for this, lmk if you have issues when rebasing.

It's all good now :)

store/postgres/src/fork.rs
```diff
@@ -0,0 +1,347 @@
...
+            StoreError::ForkFailure(format!(
+                "attempt to acquire lock on `fetched_ids` failed with {}",
+                e,
+            ))
+        })?;
+        if fids.contains(&id) {
+            info!(self.logger, "Already fetched entity! Abort!"; "entity_type" => entity_type, "id" => id);
+            return Ok(None);
+        }
+        fids.insert(id.clone());
```
leoyvens, line 58: The lock should be dropped after this line. A good way to do this would be to encapsulate this mutex operation into a separate fn.

leoyvens (changes requested):

store/postgres/src/fork.rs
VIVelev: What about just introducing a new scope?

leoyvens: that also works

otaviopace (approved):

store/postgres/src/fork.rs
```diff
@@ -94,11 +94,19 @@ impl SubgraphFork {
...
+            None => return Err(StoreError::ForkFailure(format!("invalid base URL: {}", base))),
+        };
+
         Ok(Self {
             client: reqwest::Client::new(),
-            endpoint: base
-                .join(id.as_str())
-                .map_err(|e| StoreError::ForkFailure(format!("failed to join fork base: {}", e)))?,
+            endpoint: Url::parse(
+                format!("https://{}{}/{}", domain, base.path(), id).as_str(),
```
otaviopace, line 105: I don't think we should go this path (pun intended), I think we should error when someone uses a URL that doesn't end with `/`.

And if ideally this error should happen on `graph-node` startup, so that the error gets caught early.

I would even add this to the `docs` PR.

What do you think?

store/postgres/src/fork.rs
```diff
@@ -140,49 +144,79 @@ impl SubgraphFork {
...
+            return Ok(None);
+        }
 
         let map: HashMap<Attribute, Value> = {
             let mut map = HashMap::new();
             for f in fields {
-                let value = json.get(&f.name).unwrap().clone();
-                let value = RValue::from(value);
+                if f.is_derived() {
+                    // Derived fields cannot be store.set'ed.
```
otaviopace, line 208: I'm confused about this comment, can you clarify it for me please?

I know that derived fields cannot be `store.set`'ed, but doesn't the debug fork fetching only happens when **getting** data?

I'm probably missing something, and it might be the case to actually return an error here? (like [this one](https://github.com/graphprotocol/graph-node/blob/399ee444d1023a95f82404d14bb5e4534d827678/graph/src/data/store/mod.rs#L677) we already have)

store/postgres/src/subgraph_store.rs
```diff
@@ -997,10 +1005,25 @@ impl SubgraphStoreTrait for SubgraphStore {
...
+    ) -> Result<Option<Arc<dyn SubgraphFork>>, StoreError> {
+        let (store, site) = self.store(id)?;
         let info = store.subgraph_info(&site)?;
-        Ok(info.debug_fork)
+        let fork_id = info.debug_fork;
+        let schema = self.input_schema(id)?;
+
+        match (self.fork_base.as_ref(), fork_id) {
+            (Some(b), Some(f)) => Ok(Some(Arc::new(fork::SubgraphFork::new(
+                b.join(&format!("id/{}", f))
```
otaviopace: The reason this used to fail with URLs that ended up with no `/` is because `.join` **replaces** the "file" from the path if it's not a folder, eg:

```rust
// Without `/`
let base = Url::parse("https://api.thegraph.com/subgraphs/id")?;
let url = base.join("QmSH3hhL9vAxQbE6ydM6eYN9ULtf6WS7hCHgQqxAhMJwjL")?;
// Removed the `/id`
assert_eq!(url.as_str(), "https://api.thegraph.com/subgraphs/QmSH3hhL9vAxQbE6ydM6eYN9ULtf6WS7hCHgQqxAhMJwjL");

// With `/`
et base = Url::parse("https://api.thegraph.com/subgraphs/id/")?;
let url = base.join("QmSH3hhL9vAxQbE6ydM6eYN9ULtf6WS7hCHgQqxAhMJwjL")?;
// Kept `/id`
assert_eq!(url.as_str(), "https://api.thegraph.com/subgraphs/id/QmSH3hhL9vAxQbE6ydM6eYN9ULtf6WS7hCHgQqxAhMJwjL");
```

`Url::join` docs: https://docs.rs/url/latest/url/struct.Url.html#method.join

store/postgres/src/fork.rs
otaviopace: Related to the discussion: https://github.com/graphprotocol/graph-node/pull/2995#discussion_r795717925

If you agree to early return the error, a simple implementation that I can think of is to assert that the `path` is a `folder`.

store/postgres/src/fork.rs
VIVelev: The fetched data must be stored locally, so that the local store is synced once and moves on, rather than constantly fetching new data from the endpoint, which at some point will be outdated and cause unexpected behavior.

That's why fetched entities need to also be `store.set`'ed, but the derived fields can be ignored, since they are never `store.set`'ed.

Now to think of it, I believe I should nevertheless return the derived fields as part of the entity - not completely ignore them. `store.get` resolves derived fields, right?

store/postgres/src/fork.rs
VIVelev: `FORK_BASE_URL` env? It is not intended to set the fork base using this environment variable. Although I might be wrong. Maybe some rust macro I used to define the `--fork_base` arg does this behind my back?

Anyway, now that I tested it - yes, it has this issue. Huh, I remember testing this scenario...

But imo, missing a `/` at the end should not cause an error. Imo, this is an expected input. In a web browser when you dial a URL and miss the `/` it doesn't complain that the URL is invalid or smth. It seems an intuitive behavior for me.

store/postgres/src/fork.rs
otaviopace: Oh that makes sense. Hmm about the resolution of derived fields, I can't recall if they show up or not on a `store.get`.

One way to be sure is to add a simple dummy `log` to a regular subgraph to be sure if they are resolved or not.

store/postgres/src/fork.rs
otaviopace: Sorry the env doesn't exist, I forgot I'm running `graph-node` like this 😅 

```
export FORK_BASE_URL=https://api.thegraph.com/subgraphs/id
cargo run -p graph-node -- \
        --postgres-url $THEGRAPH_STORE_POSTGRES_DIESEL_URL \
        --ethereum-rpc $XDAI_RPC_URL \
        --ipfs 127.0.0.1:5001 \
        --fork-base $FORK_BASE_URL
```

My point about the error is because we know before hand we'll concatenate the base URL with a deployment id. Since we know that for sure, I think `graph-node` should validate it before hand (and panic) so that someone doesn't discover that only later when trying to debug the subgraph.

otaviopace: But yeah, I don't think this is a must to finish the PR, we can merge with the old way (preferable IMO) where the `file` is replaced.

The error message can be improved for it later.

And I would recommend add a note about it in the docs (optional as well), so people don't shoot themselves in the foot.

VIVelev: Ok, sounds good. I will revert the last commit and update the doc with a note.
Afterwards, I will work on a small PR to address the `/` in the fork base url.

store/postgres/src/fork.rs
VIVelev: Ok, I've played around with [CryptoPunks](https://github.com/itsjerryokolo/CryptoPunks). `dbg!` printing all entities resulting from `store.get`, no one contains derived fields. Thus, I think we can conclude that store.get does not resolve derived fields.

So, we can safely ignore the derived fields fetched from the endpoint. It seems like derived fields resolutions happens at the GraphQL endpoint level, rather than WASM runtime level.

Please correct me if something is wrong in my reasoning.

PR #3005, opened 2021-11-24
Refactor GraphQL execution

### LINKED ISSUE #2961: GraphQL simplification & refactor
This refactor will solve #2720 & #2943 , and simplify future iteration on the graphQL API

lutter:
Refactor how we execute GraphQL queries by transforming the AST we get from `graphql_parser` into a form that is easier to execute. The preprocessing step that translates from `graphql_parser`'s AST to our AST does the following:

* interpolate variables
* coerce argument values to their correct type
* expand fragments so that our selection sets only contain fields
* resolve interfaces (and unions) into their constituent object types

This removes a lot of hard-to-follow code in query execution, especially in `prefetch`.

Besides these simplifications, this PR also fixes a few bugs, and changes the order in which fields appear in the response to what the GraphQL spec mandates. This is a breaking change for the network, as it changes the hash for semantically equivalent query responses (cc @That3Percent on rolling this out on the network)

The refactored code also makes a few weaknesses of the current query execution visible, in particular, field merges (from users specifying the same field multiple times, or from fragment expansion) continue to use a yolo strategy that is not entirely standard conformant. This needs to be addressed in a separate PR.

Fixes #2961.

### DIFF
--- a/graphql/src/execution/ast.rs
+++ b/graphql/src/execution/ast.rs
@@ -2,7 +2,7 @@ use std::{collections::HashSet, ops::Deref};
 
 use graph::{
     components::store::EntityType,
-    data::graphql::{DocumentExt, ObjectOrInterface},
+    data::graphql::ObjectOrInterface,
     prelude::{anyhow, q, r, s, ApiSchema, QueryExecutionError, ValueMap},
 };
 use graphql_parser::Pos;
@@ -318,8 +318,6 @@ pub(crate) fn resolve_object_types(
 ) -> Result<HashSet<ObjectType>, QueryExecutionError> {
     let mut set = HashSet::new();
     match schema
-        .schema
-        .document
         .get_named_type(name)
         .ok_or_else(|| QueryExecutionError::AbstractTypeError(name.to_string()))?
     {
--- a/graphql/src/execution/ast.rs
+++ b/graphql/src/execution/ast.rs
@@ -22,7 +22,10 @@ use crate::schema::ast::ObjectType;
 #[derive(Debug, Clone, PartialEq)]
 pub struct SelectionSet {
     // Map object types to the list of fields that should be selected for
-    // them
+    // them. In most cases, this will have a single entry. If the
+    // `SelectionSet` is attached to a field with an interface or union
+    // type, it will have an entry for each object type implementing that
+    // interface or being part of the union
     items: Vec<(ObjectType, Vec<Field>)>,
 }
 
@@ -96,6 +99,10 @@ impl SelectionSet {
     }
 
     /// Iterate over all fields for the given object type
+    ///
+    /// # Panics
+    /// If this `SelectionSet` does not have an entry for `obj_type`, this
+    /// method will panic
     pub fn fields_for(&self, obj_type: &ObjectType) -> impl Iterator<Item = &Field> {
         let item = self
             .items
@@ -254,7 +261,7 @@ impl ValueMap for Field {
 
 /// A set of object types, generated from resolving interfaces into the
 /// object types that implement them, and possibly narrowing further when
-/// expanding fragments with type conitions
+/// expanding fragments with type conditions
 #[derive(Debug, Clone, PartialEq)]
 pub(crate) enum ObjectTypeSet {
     Any,
--- a/graphql/src/execution/ast.rs
+++ b/graphql/src/execution/ast.rs
@@ -3,7 +3,7 @@ use std::{collections::HashSet, ops::Deref};
 use graph::{
     components::store::EntityType,
     data::graphql::{DocumentExt, ObjectOrInterface},
-    prelude::{q, r, s, QueryExecutionError, Schema},
+    prelude::{anyhow, q, r, s, QueryExecutionError, Schema, ValueMap},
 };
 use graphql_parser::Pos;
 
@@ -228,6 +228,25 @@ impl Field {
 
     fn is_leaf(&self) -> bool {
         self.selection_set.is_empty()
+    }
+}
+
+impl ValueMap for Field {
+    fn get_required<T: graph::prelude::TryFromValue>(&self, key: &str) -> Result<T, anyhow::Error> {
+        self.argument_value(key)
+            .ok_or_else(|| anyhow!("Required field `{}` not set", key))
+            .and_then(|value| T::try_from_value(value).map_err(|e| e.into()))
+    }
+
+    fn get_optional<T: graph::prelude::TryFromValue>(
+        &self,
+        key: &str,
+    ) -> Result<Option<T>, anyhow::Error> {
+        self.argument_value(key)
+            .map_or(Ok(None), |value| match value {
+                r::Value::Null => Ok(None),
+                _ => T::try_from_value(value).map(Some).map_err(Into::into),
+            })
     }
 }
 
--- a/graphql/src/execution/ast.rs
+++ b/graphql/src/execution/ast.rs
@@ -7,7 +7,7 @@ use graph::{
 };
 use graphql_parser::Pos;
 
-use crate::schema::ast::ObjectCondition;
+use crate::schema::ast::ObjectType;
 
 /// A selection set is a table that maps object types to the fields that
 /// should be selected for objects of that type. The types are always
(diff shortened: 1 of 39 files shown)


graphql/src/execution/ast.rs
```diff
@@ -0,0 +1,349 @@
...
+            .map_or(Ok(None), |value| match value {
+                r::Value::Null => Ok(None),
+                _ => T::try_from_value(value).map(Some).map_err(Into::into),
+            })
+    }
+}
+
+/// A set of object types, generated from resolving interfaces into the
+/// object types that implement them, and possibly narrowing further when
+/// expanding fragments with type conitions
```
leoyvens, line 255:
```suggestion
/// expanding fragments with type conditions
```

graphql/src/execution/query.rs
```diff
@@ -464,189 +632,299 @@ impl Query {
...
+    /// Expand fragments and interpolate variables in a selection set
+    fn expand_selection_set(
+        &self,
+        set: q::SelectionSet,
+        type_set: &a::ObjectTypeSet,
+        ty: ObjectOrInterface<'_>,
+    ) -> Result<a::SelectionSet, Vec<QueryExecutionError>> {
+        let q::SelectionSet { span: _, items } = set;
+        // check_complexity already checked for cycles in fragment
+        // expansion, i.e. situations where a named fragment includs itself
```
leoyvens, line 838:
```suggestion
        // expansion, i.e. situations where a named fragment includes itself
```

graphql/src/execution/ast.rs
```diff
@@ -0,0 +1,349 @@
...
+
+    /// Return `true` if this selection set does not select any fields for
+    /// its types
+    pub fn is_empty(&self) -> bool {
+        self.items.iter().all(|(_, fields)| fields.is_empty())
+    }
+
+    /// If the selection set contains a single field across all its types,
+    /// return it. Otherwise, return `None`
+    pub fn single_field(&self) -> Option<&Field> {
```
leoyvens, line 55: Looking at the one place where this is used, this definition seems broader than necessary, I think it could be simplified to check for a single type with a single field.

kamilkisiela:
@lutter Hi, I started a Pull Request for graphql_parser that introduces Document Transformer https://github.com/graphql-rust/graphql-parser/pull/56

Not sure it's useful but with such trait, we could have one API for transformations and split the transformation logic into multiple pieces. What do you think?

graphql/src/execution/ast.rs
```diff
@@ -0,0 +1,349 @@
...
+///
+/// The set of types that a `SelectionSet` can accommodate must be set at
+/// the time the `SelectionSet` is constructed. It is not possible to add
+/// more types to it, but it is possible to add fields for all known types
+/// or only some of them
+#[derive(Debug, Clone, PartialEq)]
+pub struct SelectionSet {
+    // Map object types to the list of fields that should be selected for
+    // them
+    items: Vec<(String, Vec<Field>)>,
```
leoyvens, line 26: Is `items.len() > 1` only when the selection is over a union or interface? If so it would be clarifying to note that in the comment.

graphql/src/execution/ast.rs
```diff
@@ -0,0 +1,349 @@
...
+        self.items.iter().map(|(name, fields)| {
+            (
+                name.as_str(),
+                fields.iter().filter(|field| !field.is_leaf()),
+            )
+        })
+    }
+
+    /// Iterate over all fields for the given object type
+    pub fn fields_for(&self, obj_type: &s::ObjectType) -> impl Iterator<Item = &Field> {
```
leoyvens, line 97: This should document the panic condition.

graphql/src/execution/ast.rs
leoyvens: It would be nice to use a `type` for strings that represent object names (or type names in general, whatever makes more sense).

leoyvens (commented):
Small comments from reading `ast.rs`.

That3Percent:
Our arbitration charter allows for flexibility around bugfixes, so we are good to go to push this to The Network @lutter.

lutter:
> @lutter Hi, I started a Pull Request for graphql_parser that introduces Document Transformer [graphql-rust/graphql-parser#56](https://github.com/graphql-rust/graphql-parser/pull/56)
> 
> Not sure it's useful but with such trait, we could have one API for transformations and split the transformation logic into multiple pieces. What do you think?

I saw that PR and was sad that it was just a PR (I even went as far as trying to implement a visitor, but then decided it's too much work for what we need) Now that I've been through this refactor, I actually think it would be better if `graphql_parser` was more opinionated than just offering transformers, and instead had something along the lines of what this PR does, i.e., a preprocessing step for GraphQL queries. A lot of this PR is very generic, and useful for anyone who wants to execute GraphQL queries, but by necessity closely tied to our codebase. The things that I think should live upstream are

* the `graph::data::graphql::ext` module which is just an attempt to make the `graphql_parser` API more ergonomic (in `graphql_parser`, these could just be methods on its types, rather than the awkward dance with extension traits)
* the `graphql::execution::ast` module and the transformation to turn a `graphql_parser` query into that form in `graphql::execution::query`
* the `IntrospectionResolver` and the code that merges the introspection schema into the user's schema (`APISchema.schema` in our case) Keeping the introspection schema and the user schema separate lead to a lot of awkward code which is now gone, but even that you have to switch resolvers is a bit awkward.

The biggest question around doing that to me is how values should be handled, since they are user-specific, and the `graphql_parser` API for the above would need to be templated in some form around the concrete notion of `Value`, and, for example, how to coerce a `q::Value` to the user's notion of `Value`.

Besides these, one big feature request I'd have for `graphql_parser` is to use more `Arc`s in the schema AST, especially for types, since the current data structures make it hard to keep pointers to types around, and you end up referencing them with strings that need to be looked up in the schema all the time.

I have no idea what the upstream devs would think about these changes, but it would make our lives a lot easier.

graphql/src/execution/ast.rs
lutter: I've wanted to use something like `&s::ObjectType` here, but that's not possible for various reasons. I've now resolved to do the next best thing, which is keeping a copy of all object types in `ApiSchema`, and wrapping them in an `Arc`. With that, a `SelectionSet` now has basically `items: Vec<(Arc<s::ObjectType>, Vec<Field>)>` (in reality a little more complicated to work around other shortcomings of `s::ObjectType`)

I am curios what you think of that approach.

leoyvens: Looks good! I see that saved some calls to `get_object_type_definition`.

graphql/src/execution/ast.rs
lutter: The difficulty is that if the selection is for an interface, the `SelectionSet` will contain entries for multiple types. In that case, `single_field` checks that they are all the same.

graphql/src/execution/ast.rs
lutter: Added a comment to clarify when `items.len() > 1`

graphql/src/execution/ast.rs
lutter: Added

lutter:
Addressed all the comments so far

lutter:
Rebased to latest master and a dded a few commits to replicate the behavior of the current GraphQL execution for some queries that are really invalid

lutter:
This PR is now ready for review (and hopefully merging/rollout soon)

I've tested this branch against all the queries we get in the hosted service, and the only difference in responses comes from treating the (invalid) construct of not having a selection on a field with an object type slightly differently: the current implementation includes an empty object (or list of empty objects, depending on the field type) in the response, with this PR such fields do not appear in the response.

In other words, for a type like
```graphql
type Person @entity {
  id ID!
  spouse Person!
  friends [Person!]!
}
```

and the query
```graphql
{ persons { id spouse friends } }
```

with current `master`, the response has `{ "id": "..", "spouse": {}, friends: [ {}, {}, {}]}` for each person in it (with one empty object in the list for each friend). With this PR, the response would contain neither a `spouse` nor a `friends` field. The impact on apps should be negligible since they've been getting no real data all along.

tilacog (approved):

PR #3006, opened 2021-11-24
Remove `number_gte` constraint from `_Meta_`

Theodus:
This field should only be present in the generated `Block_height` type.


Jannis (approved):

fubhy:
Oh. I didn't realize that Block_height was actually already it's own type. Whoops :-)

PR #3007, opened 2021-11-24
ethereum: Env var for getLogs batch, lower default

leoyvens:
Hosted service getLogs calls more than double at the value of 1000, so a default of 2000 might be a better balance.

### DIFF
--- a/graph/src/env.rs
+++ b/graph/src/env.rs
@@ -36,3 +36,20 @@ pub fn unsafe_env_var<E: std::error::Error + Send + Sync, T: FromStr<Err = E> + 
 
     value
 }
+
+/// Panics if:
+/// - The value is not UTF8.
+/// - The value cannot be parsed as T..
+pub fn env_var<E: std::error::Error + Send + Sync, T: FromStr<Err = E> + Eq>(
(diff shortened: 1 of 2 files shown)


graph/src/env.rs
```diff
@@ -36,3 +36,21 @@ pub fn unsafe_env_var<E: std::error::Error + Send + Sync, T: FromStr<Err = E> +
 
     value
 }
+
+/// Panics if:
+/// - The value is not UTF8.
+/// - The value cannot be parsed as T.
+/// - The value differs from the default, and `--unsafe-config` flag is not set.
+pub fn env_var<E: std::error::Error + Send + Sync, T: FromStr<Err = E> + Eq>(
```
lutter, line 44: Finally!

lutter (approved):

graph/src/env.rs
leoyvens: Ha yes! That comment shouldn't be mentioning `--unsafe-config`, fixed that.

otaviopace: Wow I didn't see this before, this is awesome! 🎉

PR #3008, opened 2021-11-24
Fix unfail mechanism based of determinism

otaviopace:
Before the "PoI for failed subgraphs" feature, where we advance the
deployment head a block if the error is deterministic, we would do
the `unfail` logic after a block is successfully processed, like this:

```rust
match process_block() {
  Ok(needs_restart) => {
    if first_run {
      store.unfail();
    }
    // ...
  }
  // ...
}
```

With this feature the `unfail` mechanism had to change, we did that by
changing two things about it:

- It would run before `process_block`
- Making it revert to the parent block if the error was deterministic

This worked fine for deterministic cases, however for non-deterministic
ones we faced an issue.

We need to run `unfail` after `process_block` went successful (returned
`Ok`). This is necessary because we can only unfail the deployment if
the subgraph actually advanced past the error block range.

Example with the `unfail` previous to this commit:

- Subgraph state:
  - failed
  - deployment head at 1399
  - non-determinstic error happened at 1400
- Once the index-node gets restarted, `unfail` would run, but nothing
  would happen since the deployment head only advances after a
  successful `process_block` execution.
- Subgraph would continue to advance it's pointer but `unfail` wouldn't
  run anymore (only once at start). This would make the subgraph be in
  the `failed` state, but advancing it's pointer normally.

To fix the issue, we need to run `unfail` for non-deterministic errors
**after** the `process_block` went successful:

- Subgraph state:
  - failed
  - deployment head at 1399
  - non-determinstic error happened at 1400
- Once the index-node gets restarted, `unfail_determinstic_error` would
  run and be NOOP, then the next block would be processed (1400), the
  error can be fixed (non-deterministic) when `process_block` gets to
  run and returns `Ok`, then `unfail_non_determinstic_error` executes
  and **unfails** the subgraph finally 🙌
- Subgraph continues to advance it's pointer and the status is correct

### DIFF
--- a/core/src/subgraph/instance_manager.rs
+++ b/core/src/subgraph/instance_manager.rs
@@ -615,7 +615,7 @@ where
             let res = process_block(
                 &logger,
                 ctx.inputs.triggers_adapter.cheap_clone(),
-                ctx,
+                &mut ctx,
                 block_stream_cancel_handle.clone(),
                 block,
                 cursor.into(),
@@ -626,9 +626,7 @@ where
             subgraph_metrics.block_processing_duration.observe(elapsed);
 
             match res {
-                Ok((c, needs_restart)) => {
-                    ctx = c;
-
+                Ok(needs_restart) => {
                     deployment_failed.set(0.0);
 
                     // Notify the BlockStream implementation that a block was succesfully consumed
@@ -719,11 +717,11 @@ impl From<StoreError> for BlockProcessingError {
 async fn process_block<T: RuntimeHostBuilder<C>, C: Blockchain>(
     logger: &Logger,
     triggers_adapter: Arc<C::TriggersAdapter>,
-    mut ctx: IndexingContext<T, C>,
+    ctx: &mut IndexingContext<T, C>,
     block_stream_cancel_handle: CancelHandle,
     block: BlockWithTriggers<C>,
     firehose_cursor: Option<String>,
-) -> Result<(IndexingContext<T, C>, bool), BlockProcessingError> {
+) -> Result<bool, BlockProcessingError> {
     let triggers = block.trigger_data;
     let block = Arc::new(block.block);
     let block_ptr = block.ptr();
@@ -797,7 +795,7 @@ async fn process_block<T: RuntimeHostBuilder<C>, C: Blockchain>(
             // Losing the cache is a bit annoying but not an issue for correctness.
             //
             // See also b21fa73b-6453-4340-99fb-1a78ec62efb1.
-            return Ok((ctx, true));
+            return Ok(true);
         }
     };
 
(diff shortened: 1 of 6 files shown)


core/src/subgraph/instance_manager.rs
```diff
@@ -592,30 +593,33 @@ where
...
+            if should_try_unfail_deterministic {
+                should_try_unfail_deterministic = false;
+
+                if let Some(current_ptr) = ctx.inputs.store.block_ptr()? {
+                    if let Some(parent_ptr) =
+                        ctx.inputs.triggers_adapter.parent_ptr(&current_ptr).await?
+                    {
+                        // This reverts the deployment head to the parent_ptr if
+                        // deterministic errors happened.
+                        // There's no point in calling it if we have no current or parent block
```
leoyvens, line 610: The claim that there is no point in doing this if there are no current or parent ptrs is true but deserves a justification.

store/postgres/src/deployment_store.rs
```diff
@@ -1170,29 +1170,20 @@ impl DeploymentStore {
...
-
+    // Base for unfail logic, basically it early returns / does nothing if:
+    //
+    // - There's no fatal error for a subgraph
+    // - The error's determinism isn't the same as the one requested
+    //
+    // Otherwise it calls the provided closure/function with the subgraph error.
+    fn unfail_error<F>(&self, site: &Arc<Site>, deterministic: bool, f: F) -> Result<(), StoreError>
+    where
+        F: Fn(&PgConnection, &DeploymentHash, &ErrorDetail) -> Result<(), StoreError>,
```
leoyvens, line 1181: I dislike this callback style, I'd rather duplicate the function for the deterministic and non-deterministic cases, and then factor out code that can be reused between them.

leoyvens (changes requested):

store/postgres/src/deployment_store.rs
otaviopace: Do you have a suggestion for doing this without the callback style considering that we need early returns here?

I could do a common method, but that would have to return: `should_early_return: bool` + `maybe_subgraph_error: Option<ErrorDetail>` values, which I dislike more over the callback style.

Example:

```rust
fn unfail_error() -> (bool, Option<ErrorDetail>) {

}

fn unfail_deterministic() {
    let (should_early_return, maybe_subgraph_error) = unfail_error();

    if should_early_return {
        return Ok(());
    }

    let subgraph_error = maybe_subgraph_error.unwrap();

    // ... main logic
}

fn unfail_non_deterministic() {
    let (should_early_return, maybe_subgraph_error) = unfail_error();

    if should_early_return {
        return Ok(());
    }

    let subgraph_error = maybe_subgraph_error.unwrap();

    // ... main logic
}
```

core/src/subgraph/instance_manager.rs
otaviopace: Done! 🙂 

Please tell me if it's still not clear enough.

store/postgres/src/deployment_store.rs
otaviopace: IMO when we're talking about having to control flow like this (early returning), the callback style is more suitable.

And in the future if these two methods don't have anything in common anymore, the developer coding over this should remove the common function.

otaviopace: Abstraction removed, and `error::detail` one improved.

Fixed.

leoyvens (approved):
Nice! Please squash when merging.

PR #3010, opened 2021-11-26
Fix hard-coded retry limits being used instead of REQUEST_RETRIES

xJonathanLEI:
Currently the following retries use hard-coded retry limits that cannot be configured:

- `eth_call RPC call`: `10`
- `batch eth_getTransactionReceipt RPC call`: `16`

This PR fixes it by replacing the hard-coded values with `*REQUEST_RETRIES`.

### DIFF
--- a/chain/ethereum/src/ethereum_adapter.rs
+++ b/chain/ethereum/src/ethereum_adapter.rs
@@ -494,7 +494,7 @@ impl EthereumAdapter {
                 Ok(_) | Err(EthereumContractCallError::Revert(_)) => false,
                 Err(_) => true,
             })
-            .limit(10)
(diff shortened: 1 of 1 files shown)


tilacog:
LGTM. 
@lutter / @leoyvens, do you think we would need individual values for those two RPC calls that justify having different environment variables?

leoyvens:
Considering the default is 10, I think this is fine.

xJonathanLEI:
Hey guys any update on this? Thanks

otaviopace (approved):
Sorry @xJonathanLEI this should have been merged by us already.

Thanks for the contribution 🙂

Issue #2958, opened 2021-11-08
Add a _changeBlock attribute on all entities


azf20:

Jannis:
Closely related: https://github.com/graphprotocol/graph-node/issues/295 (confirmations API). The confirmations API _might_ be overkill but it's worth considering use cases before we add an API change that we then need to maintain forever.

azf20:
Yes there are definitely overlaps, though I think the confirmations API requires quite a lot more functionality (field level details, details on the changes, finalization) which is beyond the scope of this issue.

You do raise a good point on `an API change that we then need to maintain forever`. I think there is a question whether we want to maintain the `_meta` pattern that exists at the top level of the subgraph, where `_changeBlock` is part of `_meta` (and which leaves more scope for potential extension in the future).

dotansimha:
Does this a field a consumer might need as data, or metadata? (needed for app development, or while querying/testing the subgraph API?)

Because I think we can consider creating a spec for `thegraph-introspection` that exposes those? Just like GraphQL has it's own introspection that starts with `__`, we can introduce an internal introspection that can be part of every `type` / `schem`, and have it's own lifecycle? (so we can decide if we break/maintain it or not, and make a statement that it's an internal API that is subject to changes?)

azf20:
Interesting - if that is an established pattern that could work, and as you say it has the benefit of flexibility. To answer the first question, this would be useful for app development, for example when polling for new data, or finding recently changed entities - so that does make it a bit more difficult to break anything, once introduced

azf20:
OK. Revisiting this, I am tempted to opt for the simpler approach described in this issue: https://github.com/graphprotocol/graph-node/issues/1838
This adds this functionality purely as a new filter in the top-level `where`, rather than as an extension to the entity itself. This means that it is not available to Select, or Orderby.

azf20:
Closing for now as this was implemented as a filter, we can establish whether adding this information to the entity would be useful

Issue #1838, opened 2020-08-06
GraphQL: Add filter that only matches entities changed since a given block


lutter:
To support use cases where it is necessary to find out what has changed since a given block, we should offer a `changed_since_block: $block` filter that only matches entities that have been added or updated since `$block`. For completeness, it would also be good if that somehow included entities that have been deleted since `$block`, but it's not clear how 'this was deleted' could be indicated in the GraphQL response.

azf20:
Closing as a subset of https://github.com/graphprotocol/graph-node/issues/2958 (though maybe I should have elaborated this issue 🤔 )

azf20:
Re-opening this, as we may actually want to tackle this more constrained use-case first

azf20:
Proposed implementation:
- Extend the `where` top-level filter to include a `_change_block_gte` field (leading underscore in order to avoid collisions)

> The following would return all `things` as of block 1,234, which changed since block 1,000
```
things(block: { number: 1,234 }, where: { _change_block_gte: { number: 1,000 }) {
  id
}
```

This would not include deleted entities.

azf20:
@lutter I am interested in if you envisaged this functionality going here, or elsewhere (as a new top-level field?)
cc @dotansimha @kamilkisiela

dotansimha:
@azf20 I started working on a PR for this. Started from the GraphQL area of the codebase. I noticed that `input Block_height` is already exists, I guess I can use it as `where: { _block: Block_height }`? 
This way user can query with `where: { _block: { number: "XYZ" }}` or `where: { _block: { number_gte: "XYZ" }}`? 

Or should I use only `_change_block_gte` option under `where`?

azf20:
@dotansimha that is a very good point - I hadn't thought about the fact that `Block_height` already has `number_gte`, which fits perfectly with what we want to do. I think i would call it `_change_block` rather than `_block` to make it clear what the entities are being filtered on (the block when they last changed). What do you think @lutter?
You can probably take quite a lot from the existing `block` top-level filter (which is different, as it defines the state for the overall query, i.e. "query the data as of block 1000", but will use a lot of the same information that will be helpful here)

lutter:
[...]

Issue #295, opened 2018-08-25
Confirmations API


yanivtal:
This is from the Confirmations API spec by @Zerim that will soon be open sourced.

#### Confirmations API

In contrast to traditional SQL databases, data that is added to a blockchain be may reverted spontaneously due to [uncled](https://ethereum.stackexchange.com/questions/34/what-is-an-uncle-ommer-block) blocks and block reorganizations.

Apps built atop blockchains require the ability to communicate to users the likelihood that specific data is part of the permanent canonical blockchain. In Bitcoin and Ethereum (proof-of-work) blockchains this is a function of how many 'confirmations' a transaction has (how many blocks were created after this transaction). In pure proof-of-work blockchains a transaction can have many confirmations but is never technically final (i.e. there is always at least an infinitesimally small chance that it may be reverted). In proof-of-stake blockchains on the other hand, including mechanisms such as Casper FFG, blocks and transactions may achieve finality.

In our API we expose this additional information through a `_meta { fields }` field which returns a `_<Entity>_<Field>Meta` type for each entity attribute.

#### Example
Query the number of confirmations that the `owner` and `id` fields have on a `Token` entity, as well as whether the current value of the `owner` field is final and when it was last updated:
```graphql
query  {
  tokens {
    id
    owner
    _meta {
      fields {
        id {
          # The confirmations since the `id` was set is equivalent
          # to the amount of confirmations since the entity was created.
          confirmations
        }
        owner {
          # How many blocks were confirmed after the current value
          # of this field was set.
          confirmations
          # When this field was last modified
          updatedAt
          # Whether the value shown for this field has been finalized
          # Always `false` for pure proof-of-work blockchains.
          final
        }
      }
    }
  }
}

```
# 1.6 Entity and Field Changes
Even if a dApp communicates to users that some data has few confirmations, it would be jarring to see that data change suddenly without explanation. Without additional context, users would not understand whether data changed due to a chain reorganization or a legitimate transaction.

To solve this problem `change`, `latest_change` and `changes` fields are included in each entity entity's `_<Entity>Meta` type as well each field's `_<Entity>_<Field>Meta` type. It provides context as to how an entity arrived at its current value.

#### Example
Query for all `Token` entities. For each `Token` fetch the most recent operation type and the previous `owner` of that `Token`:
```graphql
query {
  tokens {
    id
    owner
    _meta {
      latest_change {
        id
        operation
        previousValue {
          owner
        }
      }
    }
  }
}
```

[...]

PR #3014, opened 2021-11-29
Add `_change_block` to entities GraphQL `input` type filter

dotansimha:
Related: https://github.com/graphprotocol/graph-node/issues/1838 

### Background

The initial discussion started at https://github.com/graphprotocol/graph-node/issues/2958 about adding a property to every type, so users can filter by it. It brought back https://github.com/graphprotocol/graph-node/issues/1838 where a user should be able to filter by block.

### Overview of the changes

* Added `EntityFilter::ChangeBlockGet` and implemented the SQL required by it.
* Added a new `input BlockChangedFilter` to allow filtering based on `number_gte` of the block change.
* No GraphQL breaking changes.
* Added special handling for filter key `_change_block` and translate that into the SQL query. 

### TODO:
- [x] Figure out the GraphQL schema structure 
- [x] Add field filter to the GraphQL `input` named `_change_block` 
- [x] Change HEAD to be based on @lutter 's refactor PR 
- [x] Introduce new input type for filtering
- [x] Modify resolvers/SQL/prefetch execution

image

image

### DIFF
--- a/graphql/src/schema/api.rs
+++ b/graphql/src/schema/api.rs
@@ -24,8 +24,9 @@ pub enum APISchemaError {
     FulltextSearchNonDeterministic,
 }
 
+// The followoing types are defined in meta.graphql
 const BLOCK_HEIGHT: &str = "Block_height";
-
+const CHANGE_BLOCK_FILTER_NAME: &str = "BlockChangedFilter";
 const ERROR_POLICY_TYPE: &str = "_SubgraphErrorPolicy_";
 
 #[derive(Debug, PartialEq, Eq, Copy, Clone)]
@@ -85,7 +86,6 @@ pub fn api_schema(input_schema: &Document) -> Result<Document, APISchemaError> {
     add_directives(&mut schema);
     add_builtin_scalar_types(&mut schema)?;
     add_order_direction_enum(&mut schema);
-    add_block_height_type(&mut schema);
     add_meta_field_type(&mut schema);
     add_types_for_object_types(&mut schema, &object_types)?;
     add_types_for_interface_types(&mut schema, &interface_types)?;
@@ -204,45 +204,6 @@ fn add_order_direction_enum(schema: &mut Document) {
     schema.definitions.push(def);
 }
 
-/// Adds a global `Block_height` type to the schema. The `block` argument
-/// accepts values of this type
-fn add_block_height_type(schema: &mut Document) {
-    let typedef = TypeDefinition::InputObject(InputObjectType {
-        position: Pos::default(),
-        description: None,
-        name: BLOCK_HEIGHT.to_string(),
-        directives: vec![],
-        fields: vec![
-            InputValue {
-                position: Pos::default(),
-                description: None,
-                name: "hash".to_owned(),
-                value_type: Type::NamedType("Bytes".to_owned()),
-                default_value: None,
-                directives: vec![],
-            },
-            InputValue {
-                position: Pos::default(),
-                description: None,
-                name: "number".to_owned(),
-                value_type: Type::NamedType("Int".to_owned()),
-                default_value: None,
-                directives: vec![],
-            },
-            InputValue {
-                position: Pos::default(),
-                description: None,
-                name: "number_gte".to_owned(),
-                value_type: Type::NamedType("Int".to_owned()),
-                default_value: None,
-                directives: vec![],
-            },
-        ],
-    });
-    let def = Definition::TypeDefinition(typedef);
-    schema.definitions.push(def);
-}
-
 /// Adds a global `_Meta_` type to the schema. The `_meta` field
 /// accepts values of this type
 fn add_meta_field_type(schema: &mut Document) {
@@ -326,12 +287,15 @@ fn add_filter_type(
     let filter_type_name = format!("{}_filter", type_name).to_string();
     match schema.get_named_type(&filter_type_name) {
         None => {
+            let mut generated_filter_fields = field_input_values(schema, fields)?;
+            generated_filter_fields.push(block_changed_filter_argument());
+
             let typedef = TypeDefinition::InputObject(InputObjectType {
                 position: Pos::default(),
                 description: None,
                 name: filter_type_name,
                 directives: vec![],
-                fields: field_input_values(schema, fields)?,
+                fields: generated_filter_fields,
             });
             let def = Definition::TypeDefinition(typedef);
             schema.definitions.push(def);
@@ -661,6 +625,17 @@ fn block_argument() -> InputValue {
         ),
         name: "block".to_string(),
         value_type: Type::NamedType(BLOCK_HEIGHT.to_owned()),
+        default_value: None,
+        directives: vec![],
+    }
+}
+
+fn block_changed_filter_argument() -> InputValue {
+    InputValue {
+        position: Pos::default(),
+        description: Some("Filter for the block changed event.".to_owned()),
+        name: "_change_block".to_string(),
+        value_type: Type::NamedType(CHANGE_BLOCK_FILTER_NAME.to_owned()),
         default_value: None,
         directives: vec![],
     }
@@ -1038,11 +1013,30 @@ mod tests {
                 "favoritePet_not_starts_with",
                 "favoritePet_ends_with",
                 "favoritePet_not_ends_with",
+                "_change_block"
             ]
             .iter()
             .map(ToString::to_string)
             .collect::<Vec<String>>()
         );
+
+        let change_block_filter = filter_type
+            .fields
+            .iter()
+            .find(move |p| match p.name.as_str() {
+                "_change_block" => true,
+                _ => false,
+            })
+            .expect("_change_block field is missing in User_filter");
+
+        match &change_block_filter.value_type {
+            Type::NamedType(name) => assert_eq!(name.as_str(), "BlockChangedFilter"),
+            _ => panic!("_change_block field is not a named type"),
+        }
+
+        schema
+            .get_named_type("BlockChangedFilter")
+            .expect("BlockChangedFilter type is missing in derived API schema");
     }
 
     #[test]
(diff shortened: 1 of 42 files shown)


yanivtal:
Awesome! I had pointed this old issue out to Adam for what we originally had in mind for what we called the [Confirmations API](https://github.com/graphprotocol/graph-node/issues/295). Have you had a chance to look at it and how would you compare the API proposed here vs the one off field in this design?

dotansimha:
> Awesome! Great to see this PR. I had pointed this old issue out to Adam for what we originally had in mind for what we called the [Confirmations API](https://github.com/graphprotocol/graph-node/issues/295). Have you had a chance to look at it and how would you compare the API proposed here vs the one off field in this design?

Thanks @yanivtal , this is super interesting. 

I think in terms of comparison of the two features, I can only say that Confirmations API seems more comprehensive, and exposes information about the actual changes, and not just the data itself. 
In this PR, I added only a filter for `_change_block` that translated into a simple `>=` operation in the database and doesn't expose any additional information (so it acts just like a basic filter). 

I went through the issue you mentioned, and it seems like that specific API that I added in this PR is not part of the examples? Or did I miss it? the actual change here allows you to fetch `tokens` that are changed after a specific block (at the level of `tokens` field), and not the information about the change. 
I guess Confirmations API can be an alternative for exposing metadata information as suggested in https://github.com/graphprotocol/graph-node/issues/2958 ? 

With some addions to Confirmations GraphQL API, I think we can introduce a similar filter on a higher level, something like: 

```graphql
schema {
  query: Query
}

type Query {
  _block(filter: BlockFilter!): EntitiesRootQuery!
  tokens(...): [Tokens]
}

input BlockFilter {
   gte: Int
}

type EntitiesRootQuery {
  tokens(...): [Tokens]
}
```

And then users can inspect the data from the point of view of a block (or, block filter), and fetch the data it needs - so the `tokens` filter remains "clear" and deals only with the actual entities, and not with the block information? 

```graphql
query {
  _block(gte: 10) {
     tokens(..) { 
        id
     }
  }
}
```

what do you think? (just a thought, I guess this might be more relevant to the discussion on GraphQL API v2?)

graphql/src/schema/api.rs
```diff
@@ -25,6 +25,7 @@ pub enum APISchemaError {
 }
 
 const BLOCK_HEIGHT: &str = "Block_height";
+const CHANGE_BLOCK_FILTER_NAME: &str = "BlockChangedFilter";
```
dotansimha, line 28: @lutter I saw that we don't have a clear naming convention standard for the built GraphQL schema, I used PascalCase because that's what we are using for inputs.

azf20:
> how would you compare the API proposed here vs the one off field in this design?

This PR is focused on filtering entities to support two specific use cases - fetching new entities as they are indexed, and debugging / inspecting when entities changed. The Confirmations API is much more feature rich, with a much bigger impact on the overall generated schema, surfacing field level changes, as well as entity level changes. That would require significant changes in SQL generation and perhaps database changes, while this was prioritised as a useful, but more incremental, feature

yanivtal:
Okay I'm fine with whatever you guys want to do here. Just want to make sure that as we consider API design, we're cognizant of the effect of adding many one off fields vs thinking through the structure of the underlying data model more holistically. If a one off field is more appropriate here I support.

graphql/src/schema/api.rs
lutter: You are right - it seems that most other type names use PascalCase; I might have screwed that up in the past when I added `Block_height` instead of `BlockHeight`; I think using that name is good and continues the more prevalent pattern.

graphql/src/schema/api.rs
```diff
@@ -240,6 +242,27 @@ fn add_block_height_type(schema: &mut Document) {
     schema.definitions.push(def);
 }
 
+/// Adds a global `BlockChangedFilter` type to the schema. The `block` argument
+/// accepts values of this type
+fn add_changed_block_filter_input_type(schema: &mut Document) {
```
lutter, line 247: This is not something you need to change, but directionally, I don't like how we create all these fixed types in code, just because it makes understanding what gets added much harder. With `meta.graphql` I started defining these types in GraphQL text, which I think is much easier to read than constructing an AST in code.

If we ever make bigger changes to the API schema in the future, I think it would be great to rename `meta.graphql` and add to that whatever can be added with fixed GraphQL text.

graphql/src/schema/api.rs
```diff
@@ -240,6 +242,27 @@ fn add_block_height_type(schema: &mut Document) {
...
+    let typedef = TypeDefinition::InputObject(InputObjectType {
+        position: Pos::default(),
+        description: None,
+        name: CHANGE_BLOCK_FILTER_NAME.to_string(),
+        directives: vec![],
+        fields: vec![InputValue {
+            position: Pos::default(),
+            description: None,
+            name: "number_gte".to_owned(),
+            value_type: Type::NamedType("Int".to_owned()),
```
lutter, line 257: That should be `number_gte: Int!`, not just `number_gte: Int`, right?

graphql/src/store/query.rs
```diff
@@ -141,6 +141,24 @@ fn build_fulltext_filter_from_object(
     )
 }
 
+fn parse_change_block_filter(
+    value: &graph::data::value::Value,
```
lutter, line 145: It's minor, but since we have so many `Value` structs floating around worth to be consistent: most of the code that deals with `graph::data::value::Value` actually does a `use graph::prelude::r` and then calls it `r::Value` (the `r` is because we call values from `graphql_parser` `q::Value` and `r` is the next letter ;) )

graphql/src/store/query.rs
```diff
@@ -141,6 +141,24 @@ fn build_fulltext_filter_from_object(
...
+) -> Result<BlockNumber, QueryExecutionError> {
+    match value {
+        graph::data::value::Value::Object(object) => {
+            let block_value = object
+                .get("number_gte")
+                .ok_or_else(|| QueryExecutionError::InvalidFilterError)?;
+
+            match block_value {
+                graph::data::value::Value::Int(n) => Ok(*n as i32),
```
lutter, line 154: There's a helper for this conversion: use `i32::try_from_value(block_value).map_err(|_| QueryExecutionError::InvalidFilterError)` instead of the `match` statement to do a conversion that checks bounds.

store/postgres/src/relational_queries.rs
```diff
@@ -1075,6 +1079,21 @@ impl<'a> QueryFilter<'a> {
...
 
+    fn filter_block_gte(
+        &self,
+        block_number_gte: &BlockNumber,
+        mut out: AstPass<Pg>,
+    ) -> QueryResult<()> {
+        // constructs a filter for lower(c.block_range) >= $block_number_gte
+        // to allow consumers to filter by a specific change block
+        out.push_sql("lower(c.");
+        out.push_identifier("block_range")?;
```
lutter, line 1090: You don't necessarily have to bother with `push_identifier` here since we know that `block_range` is SQL safe; but it would be nice to use the existing constant for that name and make it `out.push_sql(BLOCK_RANGE_COLUMN)`

lutter (changes requested):
Looks really good! Mostly just very small style comments, the only thing that I think needs to be addressed is the conversion to `i32`, though feel free to address other comments, too.  I think this would be good to merge then as soon as we merge #3005

graphql/src/schema/api.rs
dotansimha: Yeah, that's right :) will fix and move to `meta.graphql`.

graphql/src/schema/api.rs
dotansimha: Yeah, I agree that the SDL in `meta.graphql` is more readable. 
I will move the definition of `BlockChangedFilter` and `Block_height`.

graphql/src/store/query.rs
dotansimha: agree, will refactor :)

store/postgres/src/relational_queries.rs
dotansimha: oops, missed that constant, fixed!

dotansimha:
@lutter addressed all PR comments :) Thanks for the code review! 

Any idea why CI fails? the error is:

```
   Compiling wasmtime-obj v0.27.0
     Running `rustc --crate-name wasmtime_obj --edition=2018 /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/wasmtime-obj-0.27.0/src/lib.rs --error-format=json --json=diagnostic-rendered-ansi,artifacts --crate-type lib --emit=dep-info,metadata,link -C embed-bitcode=no -C debuginfo=2 -C metadata=23aca09821a9684b -C extra-filename=-23aca09821a9684b --out-dir /home/runner/work/graph-node/graph-node/target/debug/deps -L dependency=/home/runner/work/graph-node/graph-node/target/debug/deps --extern anyhow=/home/runner/work/graph-node/graph-node/target/debug/deps/libanyhow-7ce8467d8e955743.rmeta --extern more_asserts=/home/runner/work/graph-node/graph-node/target/debug/deps/libmore_asserts-bbe2ebcfe7cb672d.rmeta --extern object=/home/runner/work/graph-node/graph-node/target/debug/deps/libobject-233df90730997cd9.rmeta --extern target_lexicon=/home/runner/work/graph-node/graph-node/target/debug/deps/libtarget_lexicon-02e0409bd62baaaf.rmeta --extern wasmtime_debug=/home/runner/work/graph-node/graph-node/target/debug/deps/libwasmtime_debug-ffc768e88aabfe49.rmeta --extern wasmtime_environ=/home/runner/work/graph-node/graph-node/target/debug/deps/libwasmtime_environ-ce368bdfeb8a7ec1.rmeta --cap-lints allow -C link-arg=-fuse-ld=lld -D warnings`
error: field is never read: `logger`
  --> graph/src/firehose/endpoints.rs:21:5
   |
21 |     logger: Logger,
   |     ^^^^^^^^^^^^^^
   |
   = note: `-D dead-code` implied by `-D warnings`
```

but it seems unrelated to the changes I did? 🤔

lutter:
Yes, I think that failure was because of the new `rustc` version.

dotansimha:
> 

Fixed, and CI passes now, thanks @lutter  :) I think we are good to go with this PR?

PR #3019, opened 2021-11-30
node, store: Actually setup databases in parallel

lutter:
The previous code didn't parallelize setup, since setup makes blocking
calls, and so databases still started up one after the other.

### DIFF
--- a/store/postgres/src/connection_pool.rs
+++ b/store/postgres/src/connection_pool.rs
@@ -503,8 +503,14 @@ impl ConnectionPool {
     /// # Panics
     ///
     /// If any errors happen during the migration, the process panics
-    pub fn setup(&self) {
-        self.get_ready().ok();
+    pub async fn setup(&self) {
+        let pool = self.clone();
+        graph::spawn_blocking_allow_panic(move || {
+            pool.get_ready().ok();
+        })
+        .await
+        // propagate panics
+        .unwrap();
     }
 
     pub(crate) async fn query_permit(&self) -> tokio::sync::OwnedSemaphorePermit {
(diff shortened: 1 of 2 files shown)


store/postgres/src/connection_pool.rs
```diff
@@ -503,8 +503,16 @@ impl ConnectionPool {
     /// # Panics
     ///
     /// If any errors happen during the migration, the process panics
-    pub fn setup(&self) {
-        self.get_ready().ok();
+    pub async fn setup(&self) {
+        let pool = self.clone();
+        graph::spawn_blocking({
```
leoyvens, line 508: Lets use `spawn_blocking_allow_panic` here, `graph::spawn_blocking` is really an antipattern since futures should not block.

leoyvens (changes requested):

store/postgres/src/connection_pool.rs
lutter: Changed it to use `spawn_blocking_allow_panic` and panicking if `setup` panics

store/postgres/src/connection_pool.rs
```diff
@@ -503,8 +503,17 @@ impl ConnectionPool {
...
     /// If any errors happen during the migration, the process panics
-    pub fn setup(&self) {
-        self.get_ready().ok();
+    pub async fn setup(&self) {
+        let pool = self.clone();
+        let err = graph::spawn_blocking_allow_panic(move || {
+            pool.get_ready().ok();
+        })
+        .await
+        .unwrap_err();
```
leoyvens, line 512: Not sure on what you intended but you can propagate panics by just calling `.unwrap()`.

leoyvens (changes requested):

leoyvens (approved):

PR #3020, opened 2021-11-30
Update graphql-parser to v0.4.0

dotansimha:
v0.4 was released a few days ago, here's the changeset: https://github.com/graphql-rust/graphql-parser/releases/tag/v0.4.0 

It also includes a bugfix we needed (so we no longer need a patch in the root Cargo file). 

Also, there was a minor breaking change to Directive, I added a `repeatable: false` to all existing directives, to preserve to previous behavior.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1851,8 +1851,9 @@ dependencies = [
 
 [[package]]
 name = "graphql-parser"
-version = "0.3.0"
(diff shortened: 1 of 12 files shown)


lutter:
The integration tests are a bit flaky - I just restarted the build, maybe that will resolve it

lutter (approved):
Nice! Thanks for updating this!

Issue #2960, opened 2021-11-08
Filtering and sorting on the basis of child entities


azf20:
Subgraph consumers can currently only filter and sort on the basis of top-level entity fields. As a consumer, I want to be able to filter and sort on the basis of child entity fields.

Schema:
```
author @entity {
  id: ID!
  genre: Genre
  userProfile: User @derived(field: "id")
}

user @entity {
  id: ID!
  dateOfBirth: String
  country: String
}
```

Query:
```
authors(where: {userProfile: { dateOfBirth_eq: "1987-04-03" } }) {
  id
}
```

- Will there need to be a limit on how many layers deep a filter can be applied? _Restrict at one level, at least to begin with_
- Are there any considerations around arrays of child entities, vs. single child entities?
- How will this work with interfaces?
- Will there be a conflict between `userProfile: "12345"` and the nested filter? Will need to be a union of a String and an object, or make a breaking change

Meta-task: compare with syntax from other projects

Jannis:
From https://www.apollographql.com/docs/apollo-server/schema/unions-interfaces/#union-type:
> All of a union's included types must be object types (not scalars, input types, etc.). Included types do not need to share any fields.

So that means we can't support both
```
where: { userProfile: "foo" }
```
and
```
where: { userProfile: { id: "foo" } }
```
at the same time. :-/

dotansimha:
> All of a union's included types must be object types (not scalars, input types, etc.). Included types do not need to share any fields

Yeah, because union and interfaces need to be resolved using `resolveType` or `isTypeOf`, while scalars / input types doesn't have the ability to do that, so we can't do `UserFilter | String`. 

So ideally, we can introduce a new form of generating field arguments, that is more flexible and takes child entities into consideration, and gradually deprecate the existing one? 
We can also experiment with `@oneOf` implementation (https://github.com/graphql/graphql-spec/pull/825) to validate the input types.

kamilkisiela:
> Will there need to be a limit on how many layers deep a filter can be applied? Restrict at one level, at least to begin with
I guess it depends on how the data is fetched. All these filters could be applied as one or as many depending on the data fetching implementation. If some of them are super heavy for the database, we could define a score for them and calculate the complexity/heaviness of the query (score of 1200, where the limit is 250 will block the query from running).

> Are there any considerations around arrays of child entities, vs. single child entities?
I'm not sure what do you mean here.

> How will this work with interfaces?
[...]

PR #3022, opened 2021-12-01
Filter by child entity

kamilkisiela:
#2960

Given the schema:

```graphql
type Purpose @entity {
  id: ID!
  sender: Sender!
  purpose: String!
  createdAt: BigInt!
  transactionHash: String!
}

type Sender @entity {
  id: ID!
  address: Bytes!
  purposes: [Purpose!] @derivedFrom(field: "sender")
  createdAt: BigInt!
  purposeCount: BigInt!
}
```

The `sender_` is added to `Purpose_filter` input object.

```diff
input Purpose_filter {
    ...
+  sender_: Sender_filter
}
```

The `Sender_filter` input object is untouched.

## Questions

**How should I handle `@derivedFrom` directive?**
I see in code that the `where` argument is not generated for fields with `@derivedFrom`.

**Is it find that I hardcoded `c` for root table (the one we select columns from)?**
I noticed that aliases/prefixes for tables and columns are hard-coded so I guess it's fine to assume that `c` represents the root tabel.

**Filters can be deeply nested, is it a problem?**
The logic behind generating the joins is fairly easy and we won't end up with duplicates. It's table <-> table and it starts from "root" table (the one that we select from). 

Every table gets its own prefix/alias, the root table gets `c`. In case of a few level deep filter that adds a condition for `Foo` entity, it ends up being prefixed in the same way as the condition for `Foo` in the first or n-th level.

```sql
WHERE (c1.createdAt > 19283721 AND ( .... ( ... ( c1.purposeCount > 5 OR ... ) ) ) )
```

**Should a left join contain `vid = vid` condition?**
I used `left join` to join tables with `parent.id = child.column AND parent.vid = child.vid` condition. Is it fine? I guess so, because `vid` represents the version and they should match.

### DIFF
--- a/graph/src/data/value.rs
+++ b/graph/src/data/value.rs
@@ -3,6 +3,78 @@ use serde::ser::{SerializeMap, SerializeSeq, Serializer};
 use serde::Serialize;
 use std::collections::BTreeMap;
 use std::convert::TryFrom;
+use std::iter::FromIterator;
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct Object(BTreeMap<String, Value>);
+
+impl Object {
+    pub fn new() -> Self {
+        Self(BTreeMap::default())
+    }
+
+    pub fn get(&self, key: &str) -> Option<&Value> {
+        self.0.get(key)
+    }
+
+    pub fn remove(&mut self, key: &str) -> Option<Value> {
+        self.0.remove(key)
+    }
+
+    pub fn iter(&self) -> std::collections::btree_map::Iter<String, Value> {
+        self.into_iter()
+    }
+
+    fn len(&self) -> usize {
+        self.0.len()
+    }
+
+    pub fn extend(&mut self, other: Object) {
+        self.0.extend(other.0)
+    }
+
+    pub fn insert(&mut self, key: String, value: Value) -> Option<Value> {
+        self.0.insert(key, value)
+    }
+}
+
+impl FromIterator<(String, Value)> for Object {
+    fn from_iter<T: IntoIterator<Item = (String, Value)>>(iter: T) -> Self {
+        Object(BTreeMap::from_iter(iter))
+    }
+}
+
+impl IntoIterator for Object {
+    type Item = (String, Value);
+
+    type IntoIter = std::collections::btree_map::IntoIter<String, Value>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.0.into_iter()
+    }
+}
+
+impl<'a> IntoIterator for &'a Object {
+    type Item = (&'a String, &'a Value);
+
+    type IntoIter = std::collections::btree_map::Iter<'a, String, Value>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.0.iter()
+    }
+}
+
+impl CacheWeight for Object {
+    fn indirect_weight(&self) -> usize {
+        self.0.indirect_weight()
+    }
+}
+
+impl Default for Object {
+    fn default() -> Self {
+        Self(BTreeMap::default())
+    }
+}
 
 #[derive(Debug, Clone, PartialEq)]
 pub enum Value {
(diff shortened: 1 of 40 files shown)


graphql/src/store/query.rs
```diff
@@ -147,6 +147,7 @@ fn build_filter_from_object(
     entity: ObjectOrInterface,
     object: &BTreeMap<String, r::Value>,
 ) -> Result<Option<EntityFilter>, QueryExecutionError> {
+    // Kamil: we do iterate over the object's properties, but it's only 1 level deep.
```
kamilkisiela, line 150: @lutter 

Given the schema:

```graphql
type Note @entity {
  id: ID!
  text: String!
  author: User! @derived(field: "id")
}
              
type User @entity {
  id: ID!
  dateOfBirth: String
  country: String
  notes: [Note!]!
}
```

and this query:

```graphql
{
  users({
    where: {
      id_in: [1,2,3],
      notes_: {
        text_contains: "graphql"
      }
    }
  }) {
    id
    country
 }
}
```

The `build_filter_from_object` won't be able to get to `notes_ -> text_contains` because it goes 1 level deep and doesn't expect nested objects.

graph/src/data/value.rs
```diff
@@ -13,7 +13,7 @@ pub enum Value {
     Null,
     Enum(String),
     List(Vec<Value>),
-    Object(BTreeMap<String, Value>),
+    Object(BTreeMap<String, Value>), // Kamil: it does accept an object
```
kamilkisiela, line 16: @lutter `Value` supports nested objects

graph/src/data/store/mod.rs
```diff
@@ -246,6 +247,7 @@ impl Value {
             (r::Value::Int(i), _) => Value::Int(*i as i32),
             (r::Value::Boolean(b), _) => Value::Bool(b.to_owned()),
             (r::Value::Null, _) => Value::Null,
+            // Kamil: we should support r::Value::Object here if we want to add #2960
             _ => {
                 return Err(QueryExecutionError::AttributeTypeError(
```
kamilkisiela, line 252: @lutter  but we can't convert `Value::Object` to `r::Value` because it's not supported

graphql/src/store/query.rs
kamilkisiela: What I wanted to do is to allow `r::Value` to has `Object` and flatten the vector here (line 150-185 in this file).

```rust
Ok(Some(EntityFilter::And({
        object
            .iter()
            .map(|(key, value)| { ...
```

kamilkisiela: @lutter I need you guidance here

graphql/src/schema/ast.rs
```diff
@@ -26,6 +26,7 @@ pub(crate) enum FilterOp {
     EndsWith,
     NotEndsWith,
     Equal,
+    Child, // Kamil: This would represent a child-based filter
```
kamilkisiela, line 29: `FilterOp::Child`

graphql/src/store/query.rs
```diff
@@ -180,6 +181,7 @@ fn build_filter_from_object(
                     EndsWith => EntityFilter::EndsWith(field_name, store_value),
                     NotEndsWith => EntityFilter::NotEndsWith(field_name, store_value),
                     Equal => EntityFilter::Equal(field_name, store_value),
+                    // Child => EntityFilter::And(build_filter_from_object(..))
```
kamilkisiela, line 184: Or instead of flattening the list of EntityFilters, we could return `EntityFilter:And` built out of the 2nd level input object.

lutter:
@kamilkisiela First off, you should base your work off the `lutter/gql` branch which has the GraphQL refactor that touches quite a bit of code - you definitely don't want to have to rebase across that.

What I would do is define a new variant `EntityFilter::Child(String, Box<EntityFilter>)` (you have to box to avoid an endless recursive type) where the `String` contains the name of the field that we want to filter on. The `Object` that `build_filter_from_object` gets can contain other objects. The current code in `build_filter_from_object` will be the leaf of the recursion you'll now need, and you need to detect that recursion is needed before `Value::from_query_value` is called (~ line 165); if the field indicates that you are dealing with a child filter, you want to return something like `EntityFilter::Child(field_name, build_filter_from_object(object.get(field_name))` (there's a bit more to it to make sure that you are passing another `Object` around, the compiler will yell at you)

In your example, you want `build_filter_from_object` to return something like `vec![EntityFilter::In("id", $list), EntityFilter::Child("notes", EntityFilter::Contains("text", "graphql"))]` (I use `$list` to avoid the very verbose representation of `[1,2,3]` as a `Value::List`)

PR #3030, opened 2021-12-03
Update block ingestor to fetch receipts in parallel.

### LINKED ISSUE #3018: Update block ingestor to fetch receipts in parallel
The block ingestor currently makes batched RPC calls to fetch receipts for a given block ([link](https://github.com/graphprotocol/graph-node/blob/fda0b8606b80e11c4d6c33a67381e13a511468c1/chain/ethereum/src/ethereum_adapter.rs#L1064)). These requests are executed sequentially by the Ethereum client, which can lead to slow responses, errors & timeouts.

This could be updated to use parallel `eth_getTransactionReceipt` requests with a limited buffer ([source](https://docs.rs/futures/0.3.18/futures/prelude/stream/trait.StreamExt.html#method.buffered)), which may be more performant, in particular for chains with many transactions per block.

This functionality should be feature flagged (in case of regression), or could potentially be configured per provider.

This could also have retries at the inner-level (per receipt)

tilacog:
Replaces `Web3:Batch` with buffered, concurrent RPC calls when fetching transaction receipts for a block.

The `fn load_full_block` was updated to use `std::future::Future` and the actual handling of fetching transaction receipts was moved to a dedicated function called `fn resolve_transaction_receipts`, which uses a retry strategy for each request.

Fixes #3018

### DIFF
--- a/chain/ethereum/src/ethereum_adapter.rs
+++ b/chain/ethereum/src/ethereum_adapter.rs
@@ -1083,7 +1083,12 @@ impl EthereumAdapterTrait for EthereumAdapter {
         let hash_stream = graph::tokio_stream::iter(hashes);
 
         let receipt_stream = graph::tokio_stream::StreamExt::map(hash_stream, move |tx_hash| {
-            resolve_transaction_receipt(web3.clone(), tx_hash.clone(), block_hash, logger.clone())
+            resolve_transaction_receipt(
+                web3.cheap_clone(),
+                tx_hash,
+                block_hash,
+                logger.cheap_clone(),
+            )
         })
         .buffered(*MAX_CONCURRENT_JSON_RPC_CALLS);
 
--- a/chain/ethereum/src/ethereum_adapter.rs
+++ b/chain/ethereum/src/ethereum_adapter.rs
@@ -108,6 +108,12 @@ lazy_static! {
         .map(|s| s.split(';').filter(|s| s.len() > 0).map(ToOwned::to_owned).collect())
         .unwrap_or(Vec::new())
     };
+
+    static ref MAX_CONCURRENT_JSON_RPC_CALLS: usize = std::env::var("GRAPH_MAX_CONCURRENT_JSON_RPC_CALLS")
+            .unwrap_or("1000".into())
+            .parse::<usize>()
+            .expect("invalid GRAPH_MAX_CONCURRENT_JSON_RPC_CALLS env var");
+
 }
 
 /// Gas limit for `eth_call`. The value of 50_000_000 is a protocol-wide parameter so this
@@ -721,7 +727,7 @@ impl EthereumAdapter {
             )
         }))
         // Real limits on the number of parallel requests are imposed within the adapter.
-        .buffered(1000)
+        .buffered(*MAX_CONCURRENT_JSON_RPC_CALLS)
         .try_concat()
         .boxed()
     }
@@ -1081,7 +1087,7 @@ impl EthereumAdapterTrait for EthereumAdapter {
         let receipt_stream = graph::tokio_stream::StreamExt::map(hash_stream, move |tx_hash| {
             resolve_transaction_receipt(web3.clone(), tx_hash.clone(), block_hash, logger.clone())
         })
-        .buffered(10);
+        .buffered(*MAX_CONCURRENT_JSON_RPC_CALLS);
 
         let receipts_future = graph::tokio_stream::StreamExt::collect::<
             Result<Vec<TransactionReceipt>, IngestorError>,
--- a/chain/ethereum/src/ethereum_adapter.rs
+++ b/chain/ethereum/src/ethereum_adapter.rs
@@ -39,6 +39,7 @@ use std::convert::TryFrom;
 use std::iter::FromIterator;
 use std::sync::Arc;
 use std::time::Instant;
+use web3::api::Web3;
 
 use crate::chain::BlockFinality;
 use crate::{
@@ -1077,7 +1078,6 @@ impl EthereumAdapterTrait for EthereumAdapter {
                 .timeout_secs(*JSON_RPC_TIMEOUT)
                 .run(move || {
                     let block = block.clone();
-                    let batching_web3 = Web3::new(Batch::new(web3.transport().clone()));
 
                     let receipt_futures = block
                         .transactions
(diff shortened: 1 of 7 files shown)


leoyvens:
Were you able to observe any impact when ingesting blocks from BSC?

chain/ethereum/src/ethereum_adapter.rs
```diff
@@ -1052,115 +1058,48 @@ impl EthereumAdapterTrait for EthereumAdapter {
...
-                .no_logging()
-                .timeout_secs(*JSON_RPC_TIMEOUT)
-                .run(move || {
-                    let block = block.clone();
-                    let batching_web3 = Web3::new(Batch::new(web3.transport().clone()));
+        let hashes: Vec<_> = block
+            .transactions
+            .iter()
+            .map(|txn| txn.hash.clone())
+            .collect();
```
leoyvens, line 1081: Is this `collect` necessary?

chain/ethereum/src/ethereum_adapter.rs
```diff
@@ -1052,115 +1058,48 @@ impl EthereumAdapterTrait for EthereumAdapter {
...
-                            block_hash
-                        )
-                        .into()
-                    })
-                })
-                .boxed()
-                .compat(),
-        )
+        let receipt_stream = graph::tokio_stream::StreamExt::map(hash_stream, move |tx_hash| {
+            resolve_transaction_receipt(web3.clone(), tx_hash.clone(), block_hash, logger.clone())
```
leoyvens, line 1086: Is cloning `tx_hash` necessary? Lets use `cheap_clone` on web3 and logger.

chain/ethereum/src/ethereum_adapter.rs
tilacog: Using an iterator of owned/cloned transaction hashes caused two errors: 

```error[E0505]: cannot move out of `block` because it is borrowed
    --> chain/ethereum/src/ethereum_adapter.rs:1096:62
     |
1077 |         let hashes = block.transactions.iter().map(|txn| txn.hash.clone());
     |                      ------------------------- borrow of `block.transactions` occurs here
...
1096 |             futures03::TryFutureExt::map_ok(receipts_future, move |transaction_receipts| {
     |                                                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^ move out of `block` occurs here
1097 |                 EthereumBlock {
1098 |                     block: Arc::new(block),
     |                                     ----- move occurs due to use in closure
...
1103 |         Box::pin(block_future)
     |         ---------------------- returning this value requires that `block.transactions` is borrowed for `'static`
```
and
```
error[E0515]: cannot return value referencing local data `block.transactions`
    --> chain/ethereum/src/ethereum_adapter.rs:1103:9
     |
1077 |         let hashes = block.transactions.iter().map(|txn| txn.hash.clone());
     |                      ------------------------- `block.transactions` is borrowed here
...
1103 |         Box::pin(block_future)
     |         ^^^^^^^^^^^^^^^^^^^^^^ returns a value referencing data owned by the current function
```

It seems the borrow checker is complaining about the hashes being borrowed, even though we had them cloned.

leoyvens: Cool, thanks for checking.

docs/environment-variables.md
```diff
@@ -34,6 +34,9 @@ those.
   subgraph if the limit is reached, but will simply restart the syncing step,
   so it can be low. This limit guards against scenarios such as requesting a
   block hash that has been reorged. Defaults to 10.
+- `GRAPH_ETHEREUM_MAX _CONCURRENT_JSON_RPC_CALLS`: The maximum number of
```
otaviopace, line 37: There's an extra space here.

chain/ethereum/src/ethereum_adapter.rs
```diff
@@ -1846,3 +1790,64 @@ async fn filter_call_triggers_from_unsuccessful_transactions(
...
+    logger: Logger,
+) -> Result<TransactionReceipt, IngestorError> {
+    let retry_result = retry("batch eth_getTransactionReceipt RPC call", &logger)
+        .limit(*REQUEST_RETRIES)
+        .no_logging()
+        .timeout_secs(*JSON_RPC_TIMEOUT)
+        .run(move || web3.eth().transaction_receipt(transaction_hash))
+        .await;
+
+    match retry_result {
```
otaviopace, line 1807: Great comments down below 👏

tilacog: I agree, but they're not mine, though :sweat_smile:

tilacog:
Rebased on top of latest `master`

docs/environment-variables.md
```diff
@@ -34,6 +34,9 @@ those.
   subgraph if the limit is reached, but will simply restart the syncing step,
   so it can be low. This limit guards against scenarios such as requesting a
   block hash that has been reorged. Defaults to 10.
+- `GRAPH_ETHEREUM_MAX_CONCURRENT_JSON_RPC_CALLS`: The maximum number of
+   requests in-flight at once against Ethereum per batching context.
+   Defaults to 1,000.
```
leoyvens, line 39: This should mention that this specifically applies to the block ingestor, and not to any other json-roc calls. The env var name should also be more specific.

leoyvens (approved):
Code looks good! Just a comment on the env var.

chain/ethereum/src/ethereum_adapter.rs
```diff
@@ -1846,3 +1800,124 @@ async fn filter_call_triggers_from_unsuccessful_transactions(
...
+    retry("batch eth_getTransactionReceipt RPC call", &logger)
+        .limit(*REQUEST_RETRIES)
+        .no_logging()
+        .timeout_secs(*JSON_RPC_TIMEOUT)
+        .run(move || {
+            let web3 = web3.cheap_clone();
+            let hashes = hashes.clone();
+            let logger = logger.cheap_clone();
+            fetch_transaction_receipts_in_batch(web3, hashes, block_hash, logger).boxed()
+        })
```
tilacog, line 1820: Those clones are needed because those values are sent (possibly multiple times) to the `fetch_transaction_receipts_in_batch`, at line 1819.

leoyvens (approved):

chain/ethereum/src/ethereum_adapter.rs
```diff
@@ -1846,3 +1800,124 @@ async fn filter_call_triggers_from_unsuccessful_transactions(
...
+    logger: Logger,
+) -> Result<Vec<TransactionReceipt>, IngestorError> {
+    let batching_web3 = Web3::new(Batch::new(web3.transport().clone()));
+    let receipt_futures = hashes
+        .into_iter()
+        .map(|hash| {
+            let logger = logger.cheap_clone();
+            batching_web3
+                .eth()
+                .transaction_receipt(hash.clone())
```
tilacog, line 1839: Unecessary clone addressed in #3063.

Issue #1024, opened 2019-07-01
Automate Docker on Linux host IP address detection and injection


Jannis:
1. Write a script that
   * starts Docker Compose env
   * runs `docker inspect <graph node>` to get the host IP
   * stops Docker Compose env
   * puts the IP into `docker-compose.yml`
2. Document the script in README and on https://thegraph.com/docs/

j-mars:
Hello,

I tried running this in my Linux Virtual Machine (Ubuntu Mate) when doing a local quickstart.

`Note for Linux users: On Linux, host.docker.internal is not currently supported. Instead, you will have to replace it with the IP address of your Docker host (from the perspective of the Graph Node container). To do this, run:

$ CONTAINER_ID=$(docker container ls | grep graph-node | cut -d' ' -f1)
$ docker exec $CONTAINER_ID /bin/bash -c 'ip route | awk \'/^default via /{print $3}\''
This will print the host's IP address. Then, put it into docker-compose.yml:

$ sed -i -e 's/host.docker.internal/<IP ADDRESS>/g' docker-compose.yml`

And I couldn't get the IP Address - the graph-node gets stuck on INFO Connecting to Ethereum..., network: mainnet.

I tried following the solution depicted here: https://stackoverflow.com/questions/33102813/how-to-get-local-host-ip-address-in-docker-container

With no results as well

Jannis:
@j-mars Sorry for the delay, I was on vacation the last two weeks. Did you try running

```bash 
docker-compose up
```

followed by

```bash
CONTAINER_ID=$(docker container ls | grep graph-node | cut -d' ' -f1)
docker inspect $CONTAINER_ID
```

The inspection output should include a gateway IP address somewhere. That's the host IP address.

PR #3031, opened 2021-12-03
Fix docker-compose host IP for linux

lionseal:
This should fix issue https://github.com/docker/for-linux/issues/264

Also file `./docker/setup.sh` shouldn't be needed anymore

This step shouldn't be required anymore https://thegraph.academy/developers/local-development/
image

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -11,6 +11,8 @@ services:
     depends_on:
       - ipfs
       - postgres
+    extra_hosts:
+      - host.docker.internal:host-gateway
     environment:
       postgres_host: postgres
       postgres_user: graph-node


0xZoomEye:
the docker verion 20+ support this feature, but it still doesn`t work if the ganche-cli listen the **default** host. 
To solve this problem, the ganache should listen the **0.0.0.0** host, as below

`npx ganache-cli -h 0.0.0.0`

lionseal:
> the docker verion 20+ support this feature, but it still doesn`t work if the ganche-cli listen the **default** host. To solve this problem, the ganache should listen the **0.0.0.0** host, as below
> 
> `npx ganache-cli -h 0.0.0.0`

Yes, setting the host on the ganache node it's still necessary.

This PR only fixes the need to change the IP on the docker-compose file (or having to run the setup.sh script)

otaviopace (approved):
Seems to address https://github.com/graphprotocol/graph-node/issues/1024 well/correctly, what do you think @Jannis ?

PR #3041, opened 2021-12-07
Initial addition of FirehoseBlockIngestor for NEAR (and other Firehose chains)

maoueh:
The actual FirehoseBlockIngestor is generic enough to be re-usable on all chains that have Firehose support. This is the initial implementation that just follow chain head and update the head blocks. Rather naive in its approach right now.

Current limitations that I would like to discuss:
- ~Chain net_version and genesis_hash already exist with a default value. Now the value is fetched from Firehose correctly so there is a mismatch. How we should handle that, just delete the prior record "manually" before deploying? Automatically override if current value is XYZ? I would opt for the latter~. Going to fix it in the migrations that adds the `head_block_cursor` directly.
- ~Still need to discuss how we want to handler backfilling. There is different options available: backfill straight inside graph-node with a different cursor, backfill with graph-man in parallel fashion, backfill on demand when requestion block number to block hash~. Backfilling will be managed directly by `graph-node`. Two "stream" will fill concurrently, one from live tracking chain head, the other one from genesis up to joint point with live "stream". PR will be submitted independently for this feature.
- ~Currently doing 3 calls to database: upsert_block, attemp_chain_update and set_head_cursor, I would like to refactor that inside 1 efficient transaction instead (if possible, wanted to discuss that first before doing it).~ Merged in a single transaction with `undo` doing nothing and `new` block always updating `head_chain_block` without dealing with ancestor blocks and the like.

### DIFF
--- a/chain/ethereum/src/codec.rs
+++ b/chain/ethereum/src/codec.rs
@@ -1,12 +1,16 @@
 #[path = "protobuf/dfuse.ethereum.codec.v1.rs"]
 mod pbcodec;
 
-use graph::prelude::{
-    web3,
-    web3::types::TransactionReceipt as w3TransactionReceipt,
-    web3::types::{Bytes, H160, H2048, H256, H64, U256, U64},
-    EthereumBlock, EthereumBlockWithCalls, EthereumCall, LightEthereumBlock,
+use graph::{
+    blockchain::{Block as BlockchainBlock, BlockPtr},
+    prelude::{
+        web3,
+        web3::types::TransactionReceipt as w3TransactionReceipt,
+        web3::types::{Bytes, H160, H2048, H256, H64, U256, U64},
+        BlockNumber, EthereumBlock, EthereumBlockWithCalls, EthereumCall, LightEthereumBlock,
+    },
 };
+use std::convert::TryFrom;
 use std::sync::Arc;
 
 use crate::chain::BlockFinality;
@@ -292,6 +296,40 @@ impl Into<EthereumBlockWithCalls> for &Block {
                     })
                     .collect(),
             ),
+        }
+    }
+}
+
+impl From<Block> for BlockPtr {
+    fn from(b: Block) -> BlockPtr {
+        (&b).into()
+    }
+}
+
+impl<'a> From<&'a Block> for BlockPtr {
+    fn from(b: &'a Block) -> BlockPtr {
+        BlockPtr::from((H256::from_slice(b.hash.as_ref()), b.number))
+    }
+}
+
+impl BlockchainBlock for Block {
+    fn number(&self) -> i32 {
+        BlockNumber::try_from(self.number).unwrap()
+    }
+
+    fn ptr(&self) -> BlockPtr {
+        self.into()
+    }
+
+    fn parent_ptr(&self) -> Option<BlockPtr> {
+        let parent_hash = &self.header.as_ref().unwrap().parent_hash;
+
+        match parent_hash.len() {
+            0 => None,
+            _ => Some(BlockPtr::from((
+                H256::from_slice(parent_hash.as_ref()),
+                self.number - 1,
+            ))),
         }
     }
 }
--- a/graph/src/blockchain/firehose_block_ingestor.rs
+++ b/graph/src/blockchain/firehose_block_ingestor.rs
@@ -15,7 +15,6 @@ pub struct FirehoseBlockIngestor<M>
 where
     M: prost::Message + BlockchainBlock + Default + 'static,
 {
-    ancestor_count: i32,
     chain_store: Arc<dyn ChainStore>,
     endpoint: Arc<FirehoseEndpoint>,
     logger: Logger,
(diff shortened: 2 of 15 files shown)


maoueh:
Ready for review.

graph/src/blockchain/firehose_block_ingestor.rs
```diff
@@ -0,0 +1,156 @@
...
+        while let Some(message) = stream.next().await {
+            match message {
+                Ok(v) => {
+                    let step = bstream::ForkStep::from_i32(v.step)
+                        .expect("Fork step should always match to known value");
+
+                    let result = match step {
+                        bstream::ForkStep::StepNew => self.process_new_block(&v).await,
+                        bstream::ForkStep::StepUndo => {
+                            trace!(self.logger, "Received undo block to ingest, skipping");
```
leoyvens, line 110: Lets make this `debug!`.

graph/src/blockchain/firehose_block_ingestor.rs
```diff
@@ -0,0 +1,156 @@
...
+            "Stream blocks complete unexpectedly, expecting stream to always stream blocks"
+        );
+        latest_cursor
+    }
+
+    async fn process_new_block(&self, response: &bstream::BlockResponseV2) -> Result<(), Error> {
+        let block = decode_firehose_block::<M>(response)
+            .context("Mapping firehose block to blockchain::Block")?;
+
+        trace!(self.logger, "Received new block to ingest {}", block.ptr());
```
leoyvens, line 146: Trace is off by default and this isn't that frequent, lets make it `debug!`.

graph/src/components/store.rs
```diff
@@ -1317,6 +1317,23 @@ pub trait ChainStore: Send + Sync + 'static {
     /// The head block pointer will be None on initial set up.
     fn chain_head_ptr(&self) -> Result<Option<BlockPtr>, Error>;
 
+    /// Get the current head block cursor for this chain.
+    /// Any changes to the head block cursor will be to a block with a larger block number, never
+    /// to a block with a smaller or equal block number.
```
leoyvens, line 1322: Even in case of a revert?

node/src/main.rs
```diff
@@ -333,6 +336,14 @@ async fn main() {
                 start_block_ingestor(&logger, block_polling_interval, ethereum_chains);
             }
 
+            if near_chains.len() > 0 {
```
leoyvens, line 339: This check can be removed since `start_firehose_block_ingestor` is a noop if `near_chains` is empty. It will just log "starting 0 chains" but that's ok imo

graph/src/blockchain/firehose_block_ingestor.rs
maoueh: This is much more frequent on higher throughput chain like NEAR, printed each second. I would keep it at trace personally. What do you think?

leoyvens: Once a second isn't so bad, and in setups with multiple graph node processes the ingestor is its own process. That said, the ideal thing might be to keep this as trace but add a prometheus metric to track the latest block ingested, like the `ethereum_chain_head_number` metric in the current block ingestor.

chain/ethereum/src/codec.rs
```diff
@@ -295,3 +299,37 @@ impl Into<EthereumBlockWithCalls> for &Block {
...
+
+    fn ptr(&self) -> BlockPtr {
+        self.into()
+    }
+
+    fn parent_ptr(&self) -> Option<BlockPtr> {
+        let parent_hash = &self.header.as_ref().unwrap().parent_hash;
+
+        match parent_hash.len() {
+            0 => None,
```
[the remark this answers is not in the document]
maoueh: Actually the field is probably not populated at all in the protobuf serialization format at all, but since a `parent_hash` is backed in Rust by `::prost::alloc::vec::Vec<u8>`, `Vec` here is the empty vector.

graph/src/blockchain/firehose_block_ingestor.rs
```diff
@@ -0,0 +1,156 @@
...
+            info!(
+                self.logger,
+                "Blockstream disconnected, connecting"; "endpoint uri" => format_args!("{}", self.endpoint), "cursor" => format_args!("{}", latest_cursor),
+            );
+
+            let result = self
+                .endpoint
+                .clone()
+                .stream_blocks(bstream::BlocksRequestV2 {
+                    start_block_num: -1,
```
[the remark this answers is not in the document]
maoueh: Will add, note also the hovering the field(s) give some extra documentation (picked from the protobuf source files when they are generated).

For the cursor vs start block, the reasoning to make it easier to code instead of having two code path:

```
if cursor != <empty>
  StartStream(..., cursor: ...)
else
  StartStream(start_block, ...)
end
```

graph/src/blockchain/firehose_block_ingestor.rs
```diff
@@ -0,0 +1,156 @@
...
+                }
+            }
+        }
+    }
+
+    async fn process_blocks(
+        &self,
+        cursor: String,
+        mut stream: Streaming<bstream::BlockResponseV2>,
+    ) -> String {
```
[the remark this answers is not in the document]
maoueh: I'll keep this after #3054 and do it everywhere.

graph/src/blockchain/firehose_block_ingestor.rs
maoueh: Ok I'll create a new issue once we have this PR merged to do that (metric of `firehose_chain_head_number`).

graph/src/components/store.rs
maoueh: Forgot to adjust this comment after changing the behavior.

leoyvens (approved):

PR #3048, opened 2021-12-09
store: Fix fulltextsearch space handling

### LINKED ISSUE #3043: Fulltext search queries with spaces (but no operators) fail badly
On https://thegraph.com/hosted-service/subgraph/graphprotocol/everest, the following query fails pretty badly:

```graphql
{
  projectSearch(text:"using graphql") {
    name
  }
}
```

The error is:

```json
{
  "errors": [
    {
      "message": "Failed to get entities from store: syntax error in tsquery: \"using graphql\", query = \"/* qid: 5ff630b7821b817d-7bb418afea0e1b23 */\\nselect 'Project' as entity, c.vid, to_jsonb(c.*) as data from (select  * \\n  from \\\"sgd16672\\\".\\\"project\\\" c\\n where c.\\\"block_range\\\" @> $1 and \\\"project_search\\\" @@ to_tsquery($2)\\n\\n order by ts_rank(\\\"project_search\\\", to_tsquery($3)) asc, \\\"id\\\" asc\\n limit 100) c -- binds: [13764251, \\\"using graphql\\\", \\\"using graphql\\\"]\""
    }
  ]
}
```

1. This should probably work?
2. If it is expected to fail, the error certainly reveals way too many details about database internals.

mangas:
Fixes #3043

### DIFF
--- a/graph/src/data/query/error.rs
+++ b/graph/src/data/query/error.rs
@@ -77,6 +77,7 @@ pub enum QueryExecutionError {
     Panic(String),
     EventStreamError,
     FulltextQueryRequiresFilter,
+    FulltextQueryInvalidSyntax(String),
     DeploymentReverted,
     SubgraphManifestResolveError(Arc<SubgraphManifestResolveError>),
     InvalidSubgraphManifest,
@@ -113,6 +114,7 @@ impl QueryExecutionError {
             | Unimplemented(_)
             | CyclicalFragment(_)
             | UndefinedFragment(_)
+            | FulltextQueryInvalidSyntax(_)
             | FulltextQueryRequiresFilter => true,
             NonNullError(_, _)
             | ListValueError(_, _)
@@ -273,6 +275,7 @@ impl fmt::Display for QueryExecutionError {
             Panic(msg) => write!(f, "panic processing query: {}", msg),
             EventStreamError => write!(f, "error in the subscription event stream"),
             FulltextQueryRequiresFilter => write!(f, "fulltext search queries can only use EntityFilter::Equal"),
+            FulltextQueryInvalidSyntax(msg) => write!(f, "Invalid fulltext search query syntax. Error: {}. Hint: Search terms with spaces need to be enclosed in single quotes", msg),
             TooExpensive => write!(f, "query is too expensive"),
             Throttled => write!(f, "service is overloaded and can not run the query right now. Please try again in a few minutes"),
             DeploymentReverted => write!(f, "the chain was reorganized while executing the query"),
--- a/store/postgres/tests/relational.rs
+++ b/store/postgres/tests/relational.rs
@@ -13,6 +13,7 @@ use graph_store_postgres::layout_for_tests::SqlName;
 use hex_literal::hex;
 use lazy_static::lazy_static;
 use std::borrow::Cow;
+use std::panic;
 use std::str::FromStr;
 use std::sync::Arc;
 use std::thread::sleep;
@@ -889,6 +890,7 @@ impl<'a> QueryChecker<'a> {
     }
 
     fn check(self, expected_entity_ids: Vec<&'static str>, query: EntityQuery) -> Self {
+        let q = query.clone();
         let unordered = matches!(query.order, EntityOrder::Unordered);
         let entities = self
             .layout
@@ -921,7 +923,7 @@ impl<'a> QueryChecker<'a> {
             expected_entity_ids.sort();
         }
 
-        assert_eq!(entity_ids, expected_entity_ids);
+        assert_eq!(entity_ids, expected_entity_ids, "{:?}", q);
         self
     }
 }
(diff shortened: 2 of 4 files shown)


store/postgres/src/relational_queries.rs
```diff
@@ -529,7 +529,11 @@ impl<'a> QueryFragment<Pg> for QueryValue<'a> {
                 }
                 ColumnType::TSVector(_) => {
                     out.push_sql("to_tsquery(");
-                    out.push_bind_param::<Text, _>(s)?;
+                    let mut value = s.clone();
```
lutter, line 532: I don't think that clone is needed.

store/postgres/src/relational_queries.rs
```diff
@@ -529,7 +529,11 @@ impl<'a> QueryFragment<Pg> for QueryValue<'a> {
...
+                    if value.contains(" ") {
+                        value = format!("'{}'", value);
+                    }
+                    out.push_bind_param::<Text, _>(&value)?;
```
lutter, line 536: In this code, we try to avoid allocations as much as possible since diesel will call this code 3 or 4 times every time we issue a query, even though in this case it can't be avoided entirely. It would be better to structure this as
```rust
if s.contains(" ") {
  out.push_bind_param::<Text,_>(&format!("'{}'}, s)?;
} else {
  out.push_bind_param::<Text, _>(&s)?;
}
```

Also, is containing a space the only reason why this needs to be quoted? What happens if the value contains single quotes like `"John O'Leary"`?

store/postgres/src/relational_queries.rs
```diff
@@ -2182,7 +2186,11 @@ impl<'a> SortKey<'a> {
                 out.push_identifier(name)?;
                 out.push_sql(", to_tsquery(");
 
-                out.push_bind_param::<Text, _>(&value.unwrap())?;
+                let mut value = value.unwrap().to_string();
+                if value.contains(" ") {
+                    value = format!("'{}'", value);
+                }
+                out.push_bind_param::<Text, _>(&value.as_str())?;
```
lutter, line 2193: Since the exact same code is used twice, it would be better to factor that into a helper `fn to_tsquery(value: &str, out: &mut AstPass<Pg>) -> QueryResult<()> { .. }`

store/postgres/tests/relational.rs
```diff
@@ -1000,6 +1000,10 @@ fn check_find() {
                 vec!["3"],
                 user_query().filter(EntityFilter::Equal("userSearch".into(), "Shaq:*".into())),
             )
+            .check(
+                vec!["1"],
+                user_query().filter(EntityFilter::Equal("userSearch".into(), "Jono a".into())),
+            )
```
lutter, line 1006: Yes, I think this is exactly the right kind of test (I assume it was failing before the changes above)

We should try and see if there are other funky search terms that cause `to_tsquery` to fail. It's probably also worth checking with @fordN who wrote this initially because IIRC it is possible to put some of the text search operators into the search string.

store/postgres/src/relational_queries.rs
mangas: I'll add a test, we probably need to escape any additional single quotes before pushing it

store/postgres/src/relational_queries.rs
mangas: there are a few more edges like the List case which does a similar thing, I was thinking that this needs a little more work. I was wondering if it would be worth adding something like a quote fn for handling this in all the places as well as add the tests.

store/postgres/tests/relational.rs
mangas: cool, will do. There are definitely a few more use cases as well

mangas: And yep it was failing before the code changes :)

graph/src/data/query/error.rs
```diff
@@ -273,6 +275,7 @@ impl fmt::Display for QueryExecutionError {
             Panic(msg) => write!(f, "panic processing query: {}", msg),
             EventStreamError => write!(f, "error in the subscription event stream"),
             FulltextQueryRequiresFilter => write!(f, "fulltext search queries can only use EntityFilter::Equal"),
+            FulltextQueryInvalidSyntax(msg) => write!(f, "Invalid fulltext search query syntax. msg: {}. Tip: Spaces are only allowed around operators", msg),
```
Jannis, line 278: I would leave out the `msg: {}.` part because that's the cryptic in-db details that we probably shouldn't expose to the client (database schema names etc.).

Jannis (changes requested):

graph/src/data/query/error.rs
Jannis: Change to `Invalid fulltext search query syntax. Error: {}. Hint: Search terms with spaces need to be put in single quotes`?

mangas: updated

Jannis (approved):
LGTM!

Jannis (approved):

leoyvens (approved):
Didn't know you could have an expected message in `should_panic`, nice!

PR #3049, opened 2021-12-10
Allow placing new deployments onto one of several shards

lutter:
With this change, deployment rules in `graph-node.toml` accept multiple shards, e.g. with
```
[[deployment.rule]]
shards = [ "primary", "shard_a" ]
indexers = [ "index_node_1_a",
             "index_node_2_a",
             "index_node_3_a" ]
```

Similar to how multiple index nodes are used, when a rule specifies multiple shards, the deployment is placed into the least used shard.

### DIFF
--- a/store/postgres/src/primary.rs
+++ b/store/postgres/src/primary.rs
@@ -1232,7 +1232,7 @@ impl<'a> Connection<'a> {
             .iter()
             .map(|(node, count)| (node.as_str(), *count))
             .chain(missing)
-            .min_by(|(_, a), (_, b)| a.cmp(b))
+            .min_by_key(|(_, count)| *count)
             .map(|(node, _)| NodeId::new(node).map_err(|()| node))
             .transpose()
             // This can't really happen since we filtered by valid NodeId's
@@ -1260,6 +1260,8 @@ impl<'a> Connection<'a> {
             .order_by(sql::<i64>("count(*)"))
             .load::<(String, i64)>(self.conn.as_ref())?;
 
+        // Any shards that have no deployments in them will not be in
+        // 'used'; add them in with a count of 0
         let missing = shards
             .into_iter()
             .filter(|shard| !used.iter().any(|(s, _)| s == shard.as_str()))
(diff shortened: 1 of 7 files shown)


store/postgres/src/primary.rs
```diff
@@ -1241,6 +1241,40 @@ impl<'a> Connection<'a> {
...
+
+        let missing = shards
+            .into_iter()
+            .filter(|shard| !used.iter().any(|(s, _)| s == shard.as_str()))
+            .map(|shard| (shard.as_str(), 0));
+
+        used.iter()
+            .map(|(shard, count)| (shard.as_str(), *count))
+            .chain(missing)
+            .min_by(|(_, a), (_, b)| a.cmp(b))
```
leoyvens, line 1271: Does `min_by_key` work here?

store/postgres/src/subgraph_store.rs
```diff
@@ -117,7 +117,8 @@ impl ToSql<Text, Pg> for Shard {
 /// indexers that should index it. The deployment should then be assigned to
 /// one of the returned indexers.
```
leoyvens, line 118: This comment could use an update.

leoyvens (approved):
Nice! We should also update `docs/config.md`.

store/postgres/src/primary.rs
lutter: Yes, it does; updated.

store/postgres/src/primary.rs
```diff
@@ -1241,6 +1241,40 @@ impl<'a> Connection<'a> {
...
+            .filter(ds::shard.eq(any(shards)))
+            .select((ds::shard, sql("count(*)")))
+            .group_by(ds::shard)
+            .order_by(sql::<i64>("count(*)"))
+            .load::<(String, i64)>(self.conn.as_ref())?;
+
+        let missing = shards
+            .into_iter()
+            .filter(|shard| !used.iter().any(|(s, _)| s == shard.as_str()))
+            .map(|shard| (shard.as_str(), 0));
```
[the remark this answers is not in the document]
lutter: Good point; adding a comment

store/postgres/src/subgraph_store.rs
lutter: True. Updated that.

leoyvens (approved):

PR #3053, opened 2021-12-10
Prevent subscriptions from back-pressuring the notification queue

leoyvens:
By using a watcher instead of a channel, the back pressure is removed and notifications won't pile up in the DB due to a slow subscription. After #2922 this became easy to implement, however keeping compatibility with `throttle_while_syncing` became difficult. The watcher naturally collapses notifications which is part of what that did, but the actual throttling is no longer done. However this optimisation seems more critical at the moment, if the throttling again proves necessary we can rework it and use it again.


lutter (approved):
Nice!

PR #3054, opened 2021-12-10
Renamed of the Firehose service gRPC ids and packages

maoueh:
This will require a coordinate update for NEAR when Firehose service ids will change there, `graph-node` with this version will need to be deployed.

I suggest we merge this branch than quickly coordinate between us for the deployment of NEAR Mainnet, downtime will be minimal.


leoyvens (approved):
Code approved but lets coordinate before merging.

maoueh:
A backward compatibility layer has been pushed in the backed for this, so both version can co-exists. It was actually already been deployed to production on NEAR by mistake which caused a downtime, had to code the backward compatibility layer to fix the down time.

When the CI finished, I'll merge this.

Issue #3013, opened 2021-11-28
GraphQL: Improve documents validation before running execution flow


dotansimha:
Continue from: https://github.com/graphprotocol/graph-node/pull/3005 

> @lutter : "The refactored code also makes a few weaknesses of the current query execution visible, in particular, field merges (from users specifying the same field multiple times, or from fragment expansion) continue to use a yolo strategy that is not entirely standard conformant. This needs to be addressed in a separate PR."

I guess we can do that by adding unit tests? :)

lutter:
> Continue from: #3005
> 
> > @lutter : "The refactored code also makes a few weaknesses of the current query execution visible, in particular, field merges (from users specifying the same field multiple times, or from fragment expansion) continue to use a yolo strategy that is not entirely standard conformant. This needs to be addressed in a separate PR."
> 
> I guess we can do that by adding unit tests? :)

Yes, that should be possible. The big question is what these tests check for, but this should be testable by adding to `graphql/tests/query.rs`. On what to test: that's why I am wondering if there are some sort of GraphQL conformance tests; for example, I beleive the query `query { musicians(where: {name: "Tom" }) { id } musicians(where: {name: "Tom" }) { id } }` is legal and should respond with one `musicians` field, but if you change `Tom` to `Roy` in the second copy of `musicians`, the query should result in an error. I suspect there's lots of these cases to handle.

dotansimha:
> On what to test: that's why I am wondering if there are some sort of GraphQL conformance tests; for example, I beleive the query `query { musicians(where: {name: "Tom" }) { id } musicians(where: {name: "Tom" }) { id } }` is legal and should respond with one `musicians` field, but if you change `Tom` to `Roy` in the second copy of `musicians`, the query should result in an error. I suspect there's lots of these cases to handle.

I checked in `graphql-js` and it seems like execution doesn't really care about having multiple fields, and this is done in the `validate` phase.
In the JS ecosystem, there is a clear distinction between the request flow phases (parse, validate, execute) and what is the role of each one, and based on the spec, there are some checks that are happening only during `validate` and not checked during `execute`.

> @n1ru4l wrote a great article about the request flow: https://github.com/saihaj/graphql-ts-node-advanced-tutorial/blob/main/docs/2-customizing-the-graphql-request-flow.md 


I tested the following GraphQL schema: 

```graphql
type Query {
  a(b: String): String
}
```

And the following GraphQL query:

```query
query test {
  a(b: "1")
[...]

PR #3057, opened 2021-12-12
graph, graphql: introduce GraphQL spec-compliant `validation` phase and rules

dotansimha:
## Background

While @lutter worked on a refactor for the GraphQL execution (https://github.com/graphprotocol/graph-node/pull/3005), we needed a way to make sure some aspects of execution are valid after merging the refactor PR. 

We also noticed that the current implementation of the GraphQL engine does not include a validation phase (https://github.com/graphprotocol/graph-node/issues/3013). 

This PR uses `graphq-tools` crate (https://github.com/dotansimha/graphql-tools-rs, a set of utils on top of `graphql_parser` crate) for validations, using visitor pattern (similar to GraphQL-JS ecosystem). 

The following rules are now running: 

- `UniqueOperationNames` (https://spec.graphql.org/draft/#sec-Operation-Name-Uniqueness)
- `SingleFieldSubscriptions` (https://spec.graphql.org/draft/#sec-Operation-Name-Uniqueness)
- `KnownTypeNames` (https://spec.graphql.org/draft/#sec-Fragment-Spread-Type-Existence)
- `VariablesAreInputTypes` (https://spec.graphql.org/draft/#sec-Variables-Are-Input-Types)
- `FieldsOnCorrectType` (https://spec.graphql.org/draft/#sec-Field-Selections)
- `LoneAnonymousOperation` (https://spec.graphql.org/draft/#sec-Lone-Anonymous-Operation)
- `FragmentsOnCompositeTypes` (https://spec.graphql.org/draft/#sec-Fragments-On-Composite-Types)
- `OverlappingFieldsCanBeMerged` (https://spec.graphql.org/draft/#sec-Field-Selection-Merging)
- `NoUnusedFragments` (https://spec.graphql.org/draft/#sec-Fragments-Must-Be-Used)
- `KnownFragmentNamesRule` (https://spec.graphql.org/draft/#sec-Fragment-spread-target-defined)
- `LeafFieldSelections` (https://spec.graphql.org/draft/#sec-Leaf-Field-Selections)
- `UniqueFragmentNames` (https://spec.graphql.org/draft/#sec-Fragment-Name-Uniqueness)
- `NoFragmentsCycle` (https://spec.graphql.org/draft/#sec-Fragment-spreads-must-not-form-cycles)
- `PossibleFragmentSpreads` (https://spec.graphql.org/draft/#sec-Fragment-spread-is-possible)
- `NoUnusedVariables` (https://spec.graphql.org/draft/#sec-All-Variables-Used)
- `NoUndefinedVariables` (https://spec.graphql.org/draft/#sec-All-Variable-Uses-Defined)
- `KnownArgumentNames` (https://spec.graphql.org/draft/#sec-Argument-Names , https://spec.graphql.org/draft/#sec-Directives-Are-In-Valid-Locations)
- `UniqueArgumentNames` (https://spec.graphql.org/draft/#sec-Argument-Names)
- `UniqueVariableNames` (https://spec.graphql.org/draft/#sec-Variable-Uniqueness)
- `ProvidedRequiredArguments` (https://spec.graphql.org/draft/#sec-Required-Arguments)

The following spec rules are not part of this PR:
- `ExecutableDefinitions` (not actually needed, because of `graphql_parser` nature)
- `ValuesOfCorrectType` (WIP - not sure yet about requirement by `graph-node`)
- `VariablesInAllowedPosition` (WIP - might be needed by `graph-node`)
- `UniqueInputFieldNames`  (blocked by https://github.com/graphql-rust/graphql-parser/issues/59)
- `KnownDirectives` - not needed at the moment by `graph-node`.
- `UniqueDirectivesPerLocation` - not needed at the moment by `graph-node`.

## Changes in the PR

A `ValidationPlan` struct is now constructed in a static way, including all the rules, and can be disabled using the env var `DISABLE_GRAPHQL_VALIDATIONS=true`.

If there are any validation rules selected, the query fails with a standard error, based on the GraphQL spec:

image

image

image

image

image

## Future tasks

- We can consider moving complexity checks and a few more validations into an actual `ValidationRule` and use it there, it should make the GraphQL layer a bit more generic. 
- Drop `validate_fields` since it's covered by new validations. 

## TODO

- [x] Implement `validate` flow based on the spec
- [x] Support `graphql_parser` `Document` 
- [x] Implement initial implementation of `OverlappingFieldsCanBeMerged` 
- [x] Improve `OverlappingFieldsCanBeMerged` implementation
- [x] Add more crucial validation rules

### DIFF
--- a/core/tests/interfaces.rs
+++ b/core/tests/interfaces.rs
@@ -279,9 +279,11 @@ async fn follow_interface_reference_invalid() {
         .unwrap();
 
     match &res.to_result().unwrap_err()[0] {
-        QueryError::ExecutionError(QueryExecutionError::UnknownField(_, type_name, field_name)) => {
-            assert_eq!(type_name, "Legged");
-            assert_eq!(field_name, "parent");
+        QueryError::ExecutionError(QueryExecutionError::ValidationError(_, error_message)) => {
+            assert_eq!(
+                error_message,
+                "Cannot query field \"parent\" on type \"Legged\"."
+            );
         }
         e => panic!("error {} is not the expected one", e),
     }
@@ -1424,7 +1426,7 @@ async fn recursive_fragment() {
         .await
         .unwrap();
     let data = res.to_result().unwrap_err()[0].to_string();
-    assert_eq!(data, "query has fragment cycle including `FooFrag`");
+    assert_eq!(data, "Cannot spread fragment \"FooFrag\" within itself.");
 
     let co_recursive = "
         query {
@@ -1451,5 +1453,8 @@ async fn recursive_fragment() {
         .await
         .unwrap();
     let data = res.to_result().unwrap_err()[0].to_string();
-    assert_eq!(data, "query has fragment cycle including `BarFrag`");
+    assert_eq!(
+        data,
+        "Cannot spread fragment \"BarFrag\" within itself via \"FooFrag\"."
+    );
 }
--- a/graphql/src/execution/query.rs
+++ b/graphql/src/execution/query.rs
@@ -1,4 +1,7 @@
 use graphql_parser::Pos;
+use graphql_tools::validation::rules::*;
+use graphql_tools::validation::validate::{validate, ValidationPlan};
+use lazy_static::lazy_static;
 use std::collections::{HashMap, HashSet};
 use std::hash::{Hash, Hasher};
 use std::sync::Arc;
@@ -9,11 +12,10 @@ use graph::data::graphql::{
     ext::{DocumentExt, TypeExt},
     ObjectOrInterface,
 };
+use graph::data::query::QueryExecutionError;
 use graph::data::query::{Query as GraphDataQuery, QueryVariables};
 use graph::data::schema::ApiSchema;
-use graph::prelude::{
-    info, o, q, r, s, BlockNumber, CheapClone, Logger, QueryExecutionError, TryFromValue,
-};
+use graph::prelude::{info, o, q, r, s, BlockNumber, CheapClone, Logger, TryFromValue};
 
 use crate::introspection::introspection_schema;
 use crate::query::{ast as qast, ext::BlockConstraint};
@@ -22,6 +24,41 @@ use crate::{
     execution::{get_field, get_named_type, object_or_interface},
     schema::api::ErrorPolicy,
 };
+
+lazy_static! {
+    static ref GRAPHQL_VALIDATION_PLAN: ValidationPlan = ValidationPlan::from(
+        if std::env::var("DISABLE_GRAPHQL_VALIDATIONS")
+            .unwrap_or_else(|_| "false".into())
+            .parse::<bool>()
+            .unwrap_or_else(|_| false)
+        {
+            vec![]
+        } else {
+            vec![
+                Box::new(UniqueOperationNames {}),
+                Box::new(LoneAnonymousOperation {}),
+                Box::new(SingleFieldSubscriptions {}),
+                Box::new(KnownTypeNames {}),
+                Box::new(FragmentsOnCompositeTypes {}),
+                Box::new(VariablesAreInputTypes {}),
+                Box::new(LeafFieldSelections {}),
+                Box::new(FieldsOnCorrectType {}),
+                Box::new(UniqueFragmentNames {}),
+                Box::new(KnownFragmentNames {}),
+                Box::new(NoUnusedFragments {}),
+                Box::new(OverlappingFieldsCanBeMerged {}),
+                Box::new(NoFragmentsCycle {}),
+                Box::new(PossibleFragmentSpreads {}),
+                Box::new(NoUnusedVariables {}),
+                Box::new(NoUndefinedVariables {}),
+                Box::new(KnownArgumentNames {}),
+                Box::new(UniqueArgumentNames {}),
+                Box::new(UniqueVariableNames {}),
+                Box::new(ProvidedRequiredArguments {}),
+            ]
+        }
+    );
+}
 
 #[derive(Clone, Debug)]
 pub enum ComplexityError {
@@ -115,6 +152,21 @@ impl Query {
         max_complexity: Option<u64>,
         max_depth: u8,
     ) -> Result<Arc<Self>, Vec<QueryExecutionError>> {
+        let validation_errors = validate(
+            &schema.document(),
+            &query.document,
+            &GRAPHQL_VALIDATION_PLAN,
+        );
+
+        if validation_errors.len() > 0 {
+            return Err(validation_errors
+                .into_iter()
+                .map(|e| {
+                    QueryExecutionError::ValidationError(e.locations.first().cloned(), e.message)
+                })
+                .collect());
+        }
+
         let mut operation = None;
         let mut fragments = HashMap::new();
         for defn in query.document.definitions.into_iter() {
--- a/graphql/tests/query.rs
+++ b/graphql/tests/query.rs
@@ -895,17 +895,18 @@ fn query_complexity_subscriptions() {
                 "subscription {
                 musicians(orderBy: id) {
                     name
-                    bands(first: 100, orderBy: id) {
+                    t1: bands(first: 100, orderBy: id) {
                         name
                         members(first: 100, orderBy: id) {
                             name
                         }
                     }
-                }
-                __schema {
-                    types {
-                        name
-                    }
+                    t2: bands(first: 200, orderBy: id) {
+                      name
+                      members(first: 100, orderBy: id) {
+                          name
+                      }
+                  }
                 }
             }",
             )
@@ -932,12 +933,12 @@ fn query_complexity_subscriptions() {
             result_size: result_size_metrics(),
         };
 
-        // The extra introspection causes the complexity to go over.
         let result = execute_subscription(Subscription { query }, schema, options);
+
         match result {
-            Err(SubscriptionError::GraphQLError(e)) => match e[0] {
-                QueryExecutionError::TooComplex(1_010_200, _) => (), // Expected
-                _ => panic!("did not catch complexity"),
+            Err(SubscriptionError::GraphQLError(e)) => match &e[0] {
+                QueryExecutionError::TooComplex(3_030_100, _) => (), // Expected
+                e => panic!("did not catch complexity: {:?}", e),
             },
             _ => panic!("did not catch complexity"),
         }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1564,6 +1564,7 @@ dependencies = [
  "graph",
  "graph-chain-ethereum",
  "graphql-parser",
+ "graphql-tools",
  "indexmap",
  "lazy_static",
  "once_cell",
@@ -1794,6 +1795,15 @@ checksum = "d2ebc8013b4426d5b81a4364c419a95ed0b404af2b82e2457de52d9348f0e474"
 dependencies = [
  "combine",
  "thiserror",
+]
+
+[[package]]
+name = "graphql-tools"
+version = "0.0.15"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f47ccdd42a537f86a915d9ca4c5880005183d6eb544b86d9e924511942a00647"
+dependencies = [
+ "graphql-parser",
 ]
 
 [[package]]
--- a/graph/src/data/query/error.rs
+++ b/graph/src/data/query/error.rs
@@ -41,6 +41,7 @@ pub enum QueryExecutionError {
     AbstractTypeError(String),
     InvalidArgumentError(Pos, String, q::Value),
     MissingArgumentError(Pos, String),
+    ValidationError(Option<Pos>, String),
     InvalidVariableTypeError(Pos, String),
     MissingVariableError(Pos, String),
     ResolveEntitiesError(String),
@@ -137,6 +138,7 @@ impl QueryExecutionError {
             | DeploymentReverted
             | SubgraphManifestResolveError(_)
             | InvalidSubgraphManifest
+            | ValidationError(_, _)
             | ResultTooBig(_, _) => false,
         }
     }
@@ -160,6 +162,9 @@ impl fmt::Display for QueryExecutionError {
             OperationNameRequired => write!(f, "Operation name required"),
             OperationNotFound(s) => {
                 write!(f, "Operation name not found `{}`", s)
+            }
+            ValidationError(_pos, message) => {
+                write!(f, "{}", message)
             }
             NotSupported(s) => write!(f, "Not supported: {}", s),
             NoRootSubscriptionObjectType => {
--- a/graphql/Cargo.toml
+++ b/graphql/Cargo.toml
@@ -8,6 +8,7 @@ crossbeam = "0.8"
 futures01 = { package="futures", version="0.1.29" }
 graph = { path = "../graph" }
 graphql-parser = "0.4.0"
+graphql-tools = "0.0.15"
 indexmap = "1.7"
 Inflector = "0.11.3"
 lazy_static = "1.2.0"


graphql/src/runner.rs
```diff
@@ -135,12 +141,21 @@ where
...
         let logger = logger.new(o!("component" => "GraphQlRunner"));
         let result_size = Arc::new(ResultSizeMetrics::new(registry));
+        let mut graphql_validation_plan = ValidationPlan { rules: Vec::new() };
+
+        graphql_validation_plan.add_rule(Box::new(LoneAnonymousOperation {}));
+        graphql_validation_plan.add_rule(Box::new(FragmentsOnCompositeTypes {}));
+        graphql_validation_plan.add_rule(Box::new(OverlappingFieldsCanBeMerged {}));
+        graphql_validation_plan.add_rule(Box::new(KnownFragmentNamesRule {}));
+        graphql_validation_plan.add_rule(Box::new(NoUnusedFragments {}));
+
```
lutter, line 151: It would be better IMHO to move this and the call to validating the plan below to `graphql::execution::Query` either as part of `query.validate_fields` [here](https://github.com/graphprotocol/graph-node/blob/master/graphql/src/execution/query.rs#L379) or as a new method `validate_query` on that struct.

lutter (changes requested):
Nice progress! Two things:

1. With this approach, each validation requires a pass over the AST; probably not a big deal right now, but something to keep in mind as the number of validations grows.
2. It should be possible to turn all validations off with an environment variable - that's just so that when we roll this out in the hosted service, we can turn it off in case it breaks users' queries to give them some time to fix their queries. In the long run, that env variable will go away, but it's good to have that as we introduce new behavior.

Also, would it be possible to add a validation for [this](https://spec.graphql.org/draft/#sec-Leaf-Field-Selections) We actually have a bug right now where we run queries that violate this sentence:
```
If selectionType is an interface, union, or object

    The subselection set of that selection must NOT BE empty
```

graphql/src/runner.rs
dotansimha: I think this will change the nature of `validate` phase to be Query-specific, no? (some validation rules are useful for mutations, and some for subscriptions, so the `validate` phase is for every kind of operation, not only Query...)

I agree that we should find a better place to run it, I was thinking about doing it in `node/src/manager/commands/query.rs` in-between `parse` and `execute`, but we don't have the  `schema::Document` on that point. 

Also, I guess we can remove the `validate_fields` at some point, since it will be executed as validation rule (KnownFields), so it will also run on a separate layer from `Query` ?

dotansimha:
> Nice progress! Two things:

Thank you for the review @lutter ! 

> 1. With this approach, each validation requires a pass over the AST; probably not a big deal right now, but something to keep in mind as the number of validations grows.

I agree, that's why I also created this issue: https://github.com/graphprotocol/graph-node/issues/3077 . We can start with that and then add caching.

> 2. It should be possible to turn all validations off with an environment variable - that's just so that when we roll this out in the hosted service, we can turn it off in case it breaks users' queries to give them some time to fix their queries. In the long run, that env variable will go away, but it's good to have that as we introduce new behavior.

Sure, a flag like `DISABLE_GRAPHQL_VALIDATIONS` will do the trick here, I'll add that :)

> Also, would it be possible to add a validation for [this](https://spec.graphql.org/draft/#sec-Leaf-Field-Selections) We actually have a bug right now where we run queries that violate this sentence:
> 
> ```
> If selectionType is an interface, union, or object
> 
>     The subselection set of that selection must NOT BE empty
> ```

Sure (issue: https://github.com/dotansimha/graphql-tools-rs/issues/3), I'll work on that now.

graphql/src/runner.rs
lutter: > I think this will change the nature of `validate` phase to be Query-specific, no? (some validation rules are useful for mutations, and some for subscriptions, so the `validate` phase is for every kind of operation, not only Query...)

The `graphql::execution::query::Query` struct is actually used for both subscriptions and queries, so moving it there wouldn't bypass any logic. As for mutations, they are so far out for this part of the code that I don't think it's worth worrying about them.

> I agree that we should find a better place to run it, I was thinking about doing it in `node/src/manager/commands/query.rs` in-between `parse` and `execute`, but we don't have the `schema::Document` on that point.

The code in `node/src/manager` is only used by `graphman` commands, and not used in the `graph-node` service; putting validation there would have no effect for the service.

> Also, I guess we can remove the `validate_fields` at some point, since it will be executed as validation rule (KnownFields), so it will also run on a separate layer from `Query` ?

Yes, that's right; one important wrinkle is [this comment](https://github.com/graphprotocol/graph-node/blob/master/graphql/src/execution/query.rs#L178-L179): it should be a little clearer, but what that is talking about is mutually recursive fragments which can lead to infinite recursion if you just follow them. I haven't looked at the details with `graphql-tools`, but we have to either move any check that is following fragments after the call to `check_complexity` or make sure that `graphql-tools` checks that itself.

dotansimha: I see, let me check, the spec has a rule for recursive fragments, so maybe we can add that as well, and it will simplify some of the checks `graph-node` needs to to today. 

Regarding `Query`, I see what you mean - I'll move it there :) thanks!

dotansimha:
Ok so now we have more rules :) 

The `LeafFieldSelections` will now check the selection sets, based on the spec definition for that rule: 

```
{
  "errors": [
    {
      "message": "Field \"purpose\" must not have a selection since type \"String\" has no subfields."
    }
  ]
}
```

```
{
  "errors": [
    {
      "message": "Field \"sender\" of type \"Sender!\" must have a selection of subfields. Did you mean \"sender { ... }\"?"
    }
  ]
}
```

graphql/src/runner.rs
dotansimha: Done ;) thanks @lutter

dotansimha:
@lutter I updated to latest `graphql-tools` and now we have ~12 validation rules. The rules we have now cover everything that `validate_fields` was doing, so I think we can consider removing it?

graphql/src/runner.rs
```diff
@@ -135,12 +146,30 @@ where
...
+            graphql_validation_plan.add_rule(Box::new(KnownTypeNames {}));
+            graphql_validation_plan.add_rule(Box::new(FragmentsOnCompositeTypes {}));
+            graphql_validation_plan.add_rule(Box::new(VariablesAreInputTypes {}));
+            graphql_validation_plan.add_rule(Box::new(LeafFieldSelections {}));
+            graphql_validation_plan.add_rule(Box::new(FieldsOnCorrectType {}));
+            graphql_validation_plan.add_rule(Box::new(UniqueFragmentNames {}));
+            graphql_validation_plan.add_rule(Box::new(KnownFragmentNames {}));
+            graphql_validation_plan.add_rule(Box::new(NoUnusedFragments {}));
+            graphql_validation_plan.add_rule(Box::new(OverlappingFieldsCanBeMerged {}));
+        }
```
lutter, line 164: To me, the validation plan is an implementation detail of how we do validation, and it might be better to just make that a `lazy_static` in `query.rs` itself. I don't think there's much value in having it as a field in the GraphQL runner.

lutter (changes requested):
This looks great now! I really like how explicit the error messages are.

One small change to move the validation plan out of the runner, and this is good to merge. That change will get rid of a bunch of small modifications to pass the plan around.

Also, before merging, can you squash this? The PR now has a lot of small commits that reflect more this discussion than the sequence of steps needed to get to this point which in the long term don't make that much sense.

As for removing `validate_fields`: yes, I think we can remove this now. I would like to first roll this PR out so that if we have to disable the validations, we still have whatever `validate_fields` does right now, and then follow it with a PR in a week or so that gets rid of the env var to turn validations off and removes `validate_fields`.

graphql/src/runner.rs
dotansimha: agree, 100000%.

dotansimha:
> This looks great now! I really like how explicit the error messages are.

❤️ 

> One small change to move the validation plan out of the runner, and this is good to merge. That change will get rid of a bunch of small modifications to pass the plan around.

Done! it's now `lazy_static` ✅ 

> Also, before merging, can you squash this? The PR now has a lot of small commits that reflect more this discussion than the sequence of steps needed to get to this point which in the long term don't make that much sense.

Squashed ✅ 

> As for removing `validate_fields`: yes, I think we can remove this now. I would like to first roll this PR out so that if we have to disable the validations, we still have whatever `validate_fields` does right now, and then follow it with a PR in a week or so that gets rid of the env var to turn validations off and removes `validate_fields`.

I added `Future tasks` to the description, once it's merged, we can open issues for those and track it there?

graphql/src/execution/query.rs
```diff
@@ -23,6 +25,41 @@ use crate::{
...
+                Box::new(KnownTypeNames {}),
+                Box::new(FragmentsOnCompositeTypes {}),
+                Box::new(VariablesAreInputTypes {}),
+                Box::new(LeafFieldSelections {}),
+                Box::new(FieldsOnCorrectType {}),
+                Box::new(UniqueFragmentNames {}),
+                Box::new(KnownFragmentNames {}),
+                Box::new(NoUnusedFragments {}),
+                Box::new(OverlappingFieldsCanBeMerged {}),
+                Box::new(NoFragmentsCycle {}),
```
lutter, line 50: This check needs to be the first one that runs, otherwise other checks that follow fragments can recurse indefinitely.

Actually, looking at the code in `graphql-tools-rs`, `validate` always runs all the validations in the plan, but if fragments contain cycles, we must not ever do anything that recurses on fragments. One simple fix for now would be to call `validate` _after_ the call to `check_complexity` - that checks for cycles and returns an error.

lutter (changes requested):

graphql/src/execution/query.rs
lutter: With that change, the `NoFragmentsCycle` check is actually redundant and should be removed from the plan.

dotansimha: Fixed, and added integration tests to `graphql-tools` to make sure none of the rules are having infinite loops in any rule. 

The GraphQL validation rules are running all validations and do not short-circuit, since it needs to check all rules before returning a result.

The `QueryVisitor` only visits the tree and doesn't resolve `FragmentSpread` so this is something that each rule can do based on its needs.

dotansimha: Also, see `core/tests/interfaces.rs` changes - it's now capturing the cyclic fragments using the validation phase - so if the validation will have an infinite loop - those tests will fail :)

server/index-node/Cargo.toml
```diff
@@ -13,4 +13,4 @@ graphql-parser = "0.4.0"
 http = "0.2"
 hyper = "0.14"
 serde = "1.0"
-either = "1.6.1"
+either = "1.6.1"
```
leoyvens, line 16: What happened to the empty newline, some funkiness from your editor?

store/test-store/Cargo.toml
```diff
@@ -15,4 +15,4 @@ graph-store-postgres = { path = "../postgres" }
 lazy_static = "1.1"
 hex-literal = "0.3"
 diesel = { version = "1.4.8", features = ["postgres", "serde_json", "numeric", "r2d2"] }
-graph-chain-ethereum = { path = "../../chain/ethereum" }
+graph-chain-ethereum = { path = "../../chain/ethereum" }
```
leoyvens, line 18: same

graphql/tests/query.rs
```diff
@@ -932,12 +933,12 @@ fn query_complexity_subscriptions() {
...
 
-        // The extra introspection causes the complexity to go over.
         let result = execute_subscription(Subscription { query }, schema, options);
+
         match result {
-            Err(SubscriptionError::GraphQLError(e)) => match e[0] {
-                QueryExecutionError::TooComplex(1_010_200, _) => (), // Expected
-                _ => panic!("did not catch complexity"),
+            Err(SubscriptionError::GraphQLError(e)) => match &e[0] {
+                QueryExecutionError::TooComplex(3030100, _) => (), // Expected
```
otaviopace, line 940: I think it's good to keep the underscores in the number since it's pretty big.

graphql/src/execution/query.rs
```diff
@@ -115,6 +152,24 @@ impl Query {
...
+            &query.document,
+            &GRAPHQL_VALIDATION_PLAN,
+        );
+
+        if validation_errors.len() > 0 {
+            return Err(validation_errors
+                .into_iter()
+                .map(|e| {
+                    QueryExecutionError::ValidationError(
+                        e.locations.clone().into_iter().nth(0),
```
otaviopace, line 166: Perhaps here it's possible to just clone the first item instead of the whole `locations`.

lutter (approved):

graphql/tests/query.rs
```diff
@@ -895,17 +895,18 @@ fn query_complexity_subscriptions() {
...
-                __schema {
-                    types {
-                        name
-                    }
+                    t2: bands(first: 200, orderBy: id) {
+                      name
+                      members(first: 100, orderBy: id) {
+                          name
+                      }
+                  }
```
lutter, line 909: What happened here? Why does adding validations require changes to tests that check the complexity calculation?

lutter:
I just approved this, but then noticed that there are changes to tests that I don't understand why they are necessary. It would be good to remove those changes before merging.

Also, the commit message right now is just a log of squashed commits and is not very helpful. It should be changed to something like:
```
graph, graphql: Properly validate GraphQL queries before executing them

Uses `graphql-tools-rs` for the actual validation. Validation can be turned off by setting
`DISABLE_GRAPHQL_VALIDATIONS=true` in the environment
```

graphql/tests/query.rs
dotansimha: The subscription was failing now because of `SingleFieldSubscriptions` rule, since using root introspection fields along with a subscription is not valid. 
To achieve a similar complexity level, I changed the subscription to ask for more schema fields instead of introspection fields.

server/index-node/Cargo.toml
dotansimha: Yeah for sure, I'll fix that :)

core/tests/interfaces.rs
```diff
@@ -279,9 +279,11 @@ async fn follow_interface_reference_invalid() {
         .unwrap();
 
     match &res.to_result().unwrap_err()[0] {
-        QueryError::ExecutionError(QueryExecutionError::UnknownField(_, type_name, field_name)) => {
-            assert_eq!(type_name, "Legged");
-            assert_eq!(field_name, "parent");
+        QueryError::ExecutionError(QueryExecutionError::ValidationError(_, error_message)) => {
```
dotansimha, line 282: @lutter this one fails now in a different way because of the validations, the tests flow remains the same.

core/tests/interfaces.rs
```diff
@@ -1424,7 +1426,7 @@ async fn recursive_fragment() {
         .await
         .unwrap();
     let data = res.to_result().unwrap_err()[0].to_string();
-    assert_eq!(data, "query has fragment cycle including `FooFrag`");
+    assert_eq!(data, "Cannot spread fragment \"FooFrag\" within itself.");
```
dotansimha, line 1429: Captured by `NoFragmentCycle` rule

graphql/src/execution/query.rs
dotansimha: sure, fixed!

store/test-store/Cargo.toml
dotansimha: fixed :)

core/tests/interfaces.rs
```diff
@@ -1451,5 +1453,8 @@ async fn recursive_fragment() {
         .await
         .unwrap();
     let data = res.to_result().unwrap_err()[0].to_string();
-    assert_eq!(data, "query has fragment cycle including `BarFrag`");
+    assert_eq!(
+        data,
+        "Cannot spread fragment \"BarFrag\" within itself via \"FooFrag\"."
```
dotansimha, line 1458: Captured by NoFragmentCycle rule

graphql/tests/query.rs
dotansimha: fixed

dotansimha:
> I just approved this, but then noticed that there are changes to tests that I don't understand why they are necessary. It would be good to remove those changes before merging.

Sure. Please see my comments above - I explained on each test what's the reason for changing it. 

> Also, the commit message right now is just a log of squashed commits and is not very helpful. It should be changed to something like:
> 
> ```
> graph, graphql: Properly validate GraphQL queries before executing them
> 
> Uses `graphql-tools-rs` for the actual validation. Validation can be turned off by setting
> `DISABLE_GRAPHQL_VALIDATIONS=true` in the environment
> ```

✅

graphql/tests/query.rs
lutter: Ah .. that makes sense. Thanks for clarifying!

Issue #2581, opened 2021-06-21
Block stream pipelining


leoyvens:
The subgraph processing loop is currently entirely sequential. Pipelining is a way to parallelize it. The first target for this is the block stream, so that while a block is processed the next blocks are being scanned for triggers. This could reduce indexing time by up to 50%, though for must subgraphs that number is more like 20%. A few points to consider:

#### The block stream must be made independent from the subgraph store

Currently the block stream depends on a `subgraph_store`
https://github.com/graphprotocol/graph-node/blob/381f22e07607c66affa18b39a76d4c6003405c8e/graph/src/blockchain/block_stream.rs#L196-L196
this for for two reasons, first for keeping track of the current block pointer for the subgraph
https://github.com/graphprotocol/graph-node/blob/381f22e07607c66affa18b39a76d4c6003405c8e/graph/src/blockchain/block_stream.rs#L351-L351
which can be solved simply by the block stream keeping track of the last block pointer it emitted. Second for updating the synced status on the subgraph
https://github.com/graphprotocol/graph-node/blob/381f22e07607c66affa18b39a76d4c6003405c8e/graph/src/blockchain/block_stream.rs#L591-L591
this responsibility should be moved to the instance manager. Maybe `BlockStreamEvent` should gain a `Done` variant to signal that this should be checked.

#### The pipelining mechanism

An attractive solution is to implement this as an adapter to the existing block stream. This would be a tokio task that pulls blocks from the block stream and puts it in a channel. Buffering a single range will likely be enough for the optimization to be effective. However, the fact that the block stream scans blocks in batches (ranges) but emits them as individual blocks will require some async cleverness in this adapter. Possibly the block stream itself should emit batches (a stream of `Vec<BlockWithTriggers>`) and the adapter would expose it as a stream of individual blocks.
 
#### Stopwatch metrics

Right now the block stream uses only the `"scan_blocks"` stopwatch section. We could keep the same name for backwards compatibility with the dashboards, and just make `"scan_blocks"` mean time spent waiting for the block stream to emit a block.

azf20:
@leoyvens I know you discussed this previously with @maoueh - is this part of the current StreamingFast work or is it still waiting to be picked up by someone?

maoueh:
We worked on it a little bit, mainly pouring investigation and some time understanding. We have some half-baked implementation but we drifted towards updating Tokio so this stalled a bit.

[...]

PR #3058, opened 2021-12-13
graph: Move subgraph.block_ptr() check

mangas:


otaviopace:
@mangas do we have an issue for this? Can you provide a bit of description on why this is being done please?

If I understood correctly this PR doesn't remove the dependency the PollingBlockStream has on the deployment table (for getting the current block).

It mainly adds a `current_block_ptr` to the Context, and keeps track of that in the `Revert` events as well right?

mangas:
https://github.com/graphprotocol/graph-node/issues/2581 this is the main issue

mangas:
@otaviopace it partially removes that dependency after the first run where current_block ptr is None. 

By adding the current_block ptr, the stream no longer relies on `subgraph_store.block_ptr()` which means that further blocks can be consumed before the subgraph_store is updated. In case the process stops, it will resume from `subgraph_store.block_ptr()`

graph/src/blockchain/polling_block_stream.rs
```diff
@@ -229,7 +232,13 @@ where
 
         // Get pointers from database for comparison
         let head_ptr_opt = ctx.chain_store.chain_head_ptr()?;
-        let subgraph_ptr = ctx.subgraph_store.block_ptr()?;
+        let mut subgraph_ptr = self.current_block.clone();
+        // If current_block is None then use the subgraph block_ptr.
+        if self.current_block.is_none() {
+            if let Some(block_ptr) = self.subgraph_store.block_ptr()? {
```
leoyvens, line 238: This should not be fetched here but rather in `new_polling_block_stream`

graph/src/blockchain/polling_block_stream.rs
```diff
@@ -493,6 +518,7 @@ impl<C: Blockchain> Stream for PollingBlockStream<C> {
     type Item = Result<BlockStreamEvent<C>, Error>;
 
     fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        // let adapter = self.ctx.adapter.clone();
```
leoyvens, line 521: commented out code

leoyvens (changes requested):

graph/src/blockchain/polling_block_stream.rs
mangas: new just doesn't feel like a good place to return a Result. Could alternatively request a BlockPtr as an argument and force this to be retrieve between calling new?

graph/src/blockchain/polling_block_stream.rs
mangas: 😱

graph/src/blockchain/polling_block_stream.rs
leoyvens: `new_polling_block_stream` already returns a result.

mangas: ah yes you're right, I thought you were referring to "new" :)

mangas: I guess in this case I'll get rid of the Option completely since it's not really needed anymore 🤔

leoyvens: Perhaps ideally `new_block_stream` would receive this input, but to cover the firehose case it would need to contain both the block ptr and the firehose cursor.

leoyvens: The option is needed because `None` represents no block ever having been processed (if the subgraph has just been deployed for example).

leoyvens (approved):

PR #3078, opened 2021-12-16
Re-written FirehoseBlockStream using `async-stream`

maoueh:
The implementation of the never ending Firehose stream of blocks have been re-implemented using `async-stream` library. This will makes the implementation much more readable and maintainable in the future. This should enable re-using `TriggersAdapter::triggers_in_block` directly instead of having to defer to `firehose_triggers_in_block` (subsequent PR to come).

This also removes the `notify_block_consumed` callback on the `FirehoseBlockStream` and the latest active cursor to reconnect to is now saved inside the FirehoseBlockStream directly. This will make it easier to implement the `BufferedBlockStream` in the future.

*Caveats* Formatting of code inside `try_stream!` macro doesn't work, unclear if it's a general limitation in macros in general or this very specific macro.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1475,6 +1475,7 @@ version = "0.25.0"
 dependencies = [
  "Inflector",
  "anyhow",
+ "async-stream",
  "async-trait",
  "atomic_refcell",
  "bigdecimal",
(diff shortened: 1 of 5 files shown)


leoyvens (approved):
Amazing how much simpler this became

PR #3079, opened 2021-12-17
Feature/graphman test run (IN PROGRESS)

sduchesneau:
Work in progress -- do not merge yet

# graphman test-run
* add `graphman test-run` command (implements stop-block mechanism to stop when the subgraph's end block is reached)

# github firehose/rpc comparison workflow
* add a github workflow that runs firehose vs RPC runs
* runs in about 11 minutes
* currently only matches events, not calls or full blocks. additional runs will be added.
## requirements: 
* a github environment 'eth-mainnet' with STREAMING_FAST_API_TOKEN and STREAMING_FAST_ETH_MAINNET_RPC (ex: https://some-rpc-url-with-archive-and-trace/some-token/) 
* dependancy on example subgraph and config files in github.com/streamingfast/graph-node-dev
* action is currently configured to run on this branch only
## see it in action:
 https://github.com/streamingfast/graph-node/runs

### DIFF
--- a/.github/workflows/sf.yml
+++ b/.github/workflows/sf.yml
@@ -1,0 +1,130 @@
+name: Continuous Integration
+
+on:
+  push:
+    branches: [feature/graphman-test-run]
+
+env:
+  CARGO_TERM_COLOR: always
+  RUST_BACKTRACE: full
+  GRAPH_LOG: info
+  STREAMING_FAST_API_TOKEN: ${{ secrets.STREAMING_FAST_API_TOKEN }}
+  STREAMING_FAST_ETH_MAINNET_RPC: ${{ secrets.STREAMING_FAST_ETH_MAINNET_RPC }}
+  POSTGRES_PASSWORD: let-me-in
+  POSTGRES_DB: graph-node
+  POSTGRES_USER: graph-node
+  POSTGRES_HOSTPORT: localhost:5432
+
+jobs:
+  compare-dense:
+    name: Compare firehose with RPC-only on a dense segment
+    strategy:
+      fail-fast: false
+      matrix:
+        rust: ["stable"]
+    environment: eth-mainnet
+    runs-on: ubuntu-latest
+    services:
+      ipfs:
+        image: ipfs/go-ipfs:v0.4.23
+        ports:
+          - 5001:5001
+      postgres:
+        image: postgres
+        env:
+          POSTGRES_PASSWORD: let-me-in
+          POSTGRES_DB: graph-node
+          POSTGRES_USER: graph-node
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+        ports:
+          - 5432:5432
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v2
+
+      - name: Checkout sf config dev sources
+        uses: actions/checkout@v2
+        with:
+          repository: streamingfast/graph-node-dev
+          path: graph-node-dev
+
+      - name: Cache cargo registry
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+          key: test-cargo-${{ hashFiles('**/Cargo.toml') }}
+
+      - name: Install rust toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: ${{ matrix.rust }}
+          override: true
+
+      - name: Install lld
+        run: sudo apt-get install -y lld
+
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '17.2'
+
+      - run: |
+          echo "installing graph-cli"
+          npm install -g @graphprotocol/graph-cli
+          echo "building the mainnet-events subgraph"
+          set -o pipefail
+          pushd graph-node-dev
+          yarn
+          graph build  --ipfs http://localhost:5001 subgraphs/ethereum/transfer/mainnet-events.yaml |tee buildout
+          IPFS_HASH="$(awk '/Build.completed/ {print $3}' buildout)"
+          echo "exporting ipfs hash $IPFS_HASH"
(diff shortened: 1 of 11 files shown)


sduchesneau:
Requires some discussion about how the comparison tests should be run. Maybe they could be hooked to releases (not build graphman in the action to make it faster), maybe they could be in a different repository, maybe they could start a bigger matrix of runs, in parallel.

node/src/manager/commands/test_run.rs
```diff
@@ -0,0 +1,577 @@
...
+        }
+
+    }
+    info!(&logger, "Removing subgraph {}", name);
+    subgraph_store.clone().remove_subgraph(subgraph_name)?;
+
+    Ok(())
+}
+
+// Stuff copied directly moslty from `main.rs`
```
maoueh, line 216: Everything below that point has been copied from `main.rs`, we will re-use that in some shared code.

.github/workflows/sf.yml
```diff
@@ -0,0 +1,224 @@
+name: Continuous Integration
+
+on:
+  push:
+    branches: [feature/graphman-test-run,stepd/graphman-test-run-sandbox]
```
maoueh, line 5: @sduchesneau We need to update those.

sduchesneau:
@mangas ^ I've just removed the Github Actions part of this PR (by force-pushing rebased version over my feature-branch) 
This here is only the implementation of graphman 'test-run' command, allowing the following:

```bin/graphman --config eth-mainnet-firehose.toml test-run eth-mainnet {IPFS_HASH} {STOP_BLOCK}```

File: node/src/bin/manager.rs [hunk 3]
```diff
@@ -582,6 +605,30 @@ async fn main() {
                 sleep,
             )
         }
+        TestRun {
```
mangas, line 608: Any objection to calling this Run and having stop block as an optional flag?

mangas:
Adding @leoyvens for review as well as I don't really trust my reviews yet :D

chain/ethereum/src/codec.rs
```diff
@@ -134,7 +134,10 @@ impl Into<web3::types::U64> for TransactionTraceStatus {
 impl Into<Option<web3::types::U64>> for TransactionTraceStatus {
     fn into(self) -> Option<web3::types::U64> {
         match self {
-            Self::Unknown => None,
+            Self::Unknown => {
+                {}
```
mangas, line 138: is this needed?

maoueh: It's a separate fix that can go in another PR, we will extract it to it's own PR.

node/src/bin/manager.rs
maoueh: I'm with @mangas, `graphman run` feels better.

About your concerns @sduchesneau , `graphman` is a tool that is deemed for "management of graph-node" instances, so I don't think it will be confused as production means to run a graph node.

@lutter What do you think about this, do you think there could be some confusions if we use `graphman run`?

chain/ethereum/src/codec.rs
maoueh: See #3138, @sduchesneau need to rebase once my other PR is merged.

maoueh: Merged, ready for rebase

node/src/bin/manager.rs
leoyvens: I'm +1 on `run`

node/src/bin/manager.rs
```diff
@@ -411,6 +434,10 @@ impl Context {
         pools
     }
 
+    async fn store_builder(self) -> StoreBuilder {
+        StoreBuilder::new(&self.logger, &self.node_id, &self.config, self.registry).await
+    }
```
lutter, line 439: Heh .. the whole point of this `Context` struct was so that we wouldn't have to pass the `StoreBuilder` around because that obscures what exactly a command might be doing with the store. It would be better to add a utility method to the `Context` that returns the `NetworkStore`, the primary pool, and the chain head listener and pass those into `commands::run`

Also, `StoreBuilder::new` will run migrations - avoiding that is a big reason why there are all these other methods on `StoreBuilder` so that using the wrong `graphman` version won't screw up the database schema. 

I would change this method to call `Self::store_and_primary` and then copy the few lines that create a `ChainHeadUpdateListener` from `StoreBuilder::new` to here

maoueh: We changed it exactly to be able to run ... migrations :D I notice the differences indeed. Migrations was required first for CI since it always starts fresh but it's also quite good when using it for development purposes as I often starts from a fresh database.

The best will be to discuss that over a video chat to see what you be the best course of action here. Two things come to mind:

- Run only on pristine database
- Run only when specific flag is provided

core/src/subgraph/instance_manager.rs
```diff
@@ -620,6 +628,16 @@ where
...
+            match ctx.inputs.stop_block.clone() {
+                Some(stop_block) => {
+                    if block_ptr.number > stop_block {
+                        info!(&logger, "stop block reached for subgraph");
+                        return Ok(());
+                    }
+                }
+                _ => {}
+            }
+
```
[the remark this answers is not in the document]
sduchesneau: The instance manager stops after seeing (and before processing) the block HIGHER THAN the stop block:
```if block_ptr.number > stop_block```

So the behavior is to process up to stop_block INCLUSIVE, as you would expect (well, it Works on My Machine :tm:)

File: node/src/bin/manager.rs [hunk 7]
```diff
@@ -144,6 +144,17 @@ pub enum Command {
...
+    /// Runs a suite of tests around a specific subgraph
+    TestRun {
+        /// Network name (must fit one of the chain)
+        network_name: String,
+
+        /// Subgraph in the form `<IPFS Hash>` or `<name>:<IPFS Hash>`
+        subgraph: String,
+
+        /// Number of block to process
+        stop_block: i32,
```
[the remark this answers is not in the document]
sduchesneau: not needed if my previous claim is right 😃

PR #3083, opened 2021-12-17
Avoid parsing X triggers if the filter is empty

### LINKED ISSUE #3080: firehose_triggers_in_block should add conditional judgment before filter
The method **FirehoseMapper.firehose_triggers_in_block** should add conditional judgment before filter. 
```
        if !filter.log.is_empty() {
            triggers.append(&mut parse_log_triggers(&filter.log, &block.ethereum_block));
        }
        if !filter.call.is_empty() {
            triggers.append(&mut parse_call_triggers(&filter.call, &block)?);
        }
        if !filter.block.is_empty() {
            triggers.append(&mut parse_block_triggers(filter.block.clone(), &block));
        }
```

maoueh:
When a filter is empty, it means it cannot match anything. By guarding the parsing of the triggers upfront, we avoid looping through lot's of element in the blocks, this is especially true for full blocks that contain all the transactions.

Fixes #3080

### DIFF
--- a/chain/ethereum/src/adapter.rs
+++ b/chain/ethereum/src/adapter.rs
@@ -400,12 +400,6 @@ impl FromIterator<(BlockNumber, Address, FunctionSelector)> for EthereumCallFilt
     }
 }
 
-impl From<EthereumBlockFilter> for EthereumCallFilter {
-    fn from(ethereum_block_filter: EthereumBlockFilter) -> Self {
-        EthereumCallFilter::from(&ethereum_block_filter)
-    }
-}
-
 impl From<&EthereumBlockFilter> for EthereumCallFilter {
     fn from(ethereum_block_filter: &EthereumBlockFilter) -> Self {
         Self {
--- a/chain/ethereum/src/adapter.rs
+++ b/chain/ethereum/src/adapter.rs
@@ -483,6 +483,16 @@ impl EthereumBlockFilter {
     fn requires_traces(&self) -> bool {
         !self.contract_addresses.is_empty()
     }
+
+    /// An empty filter is one that never matches.
+    pub fn is_empty(&self) -> bool {
+        // If we are triggering every block, we are of course not empty
+        if self.trigger_every_block {
+            return false;
+        }
+
+        self.contract_addresses.is_empty()
+    }
 }
 
 #[derive(Clone)]
--- a/chain/ethereum/src/adapter.rs
+++ b/chain/ethereum/src/adapter.rs
@@ -402,11 +402,19 @@ impl FromIterator<(BlockNumber, Address, FunctionSelector)> for EthereumCallFilt
 
 impl From<EthereumBlockFilter> for EthereumCallFilter {
     fn from(ethereum_block_filter: EthereumBlockFilter) -> Self {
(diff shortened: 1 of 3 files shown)


chain/ethereum/src/chain.rs
```diff
@@ -522,12 +522,22 @@ impl TriggersAdapterTrait<Chain> for TriggersAdapter {
...
+                    ));
+                }
+
+                if !filter.call.is_empty() {
+                    triggers.append(&mut parse_call_triggers(&filter.call, &full_block)?);
+                }
+
+                if !filter.block.is_empty() {
+                    triggers.append(&mut parse_block_triggers(filter.block.clone(), &full_block));
+                }
```
leoyvens, line 539: My preferred style is to move these checks into each `parse_` fn, returning an empty Vec in that case.

leoyvens (changes requested):

chain/ethereum/src/chain.rs
maoueh: Addressed, I delayed the `filter.block.clone()` also to later so if there is no block handlers, the `clone` does not happen.

chain/ethereum/src/adapter.rs
```diff
@@ -402,11 +402,19 @@ impl FromIterator<(BlockNumber, Address, FunctionSelector)> for EthereumCallFilt
 
 impl From<EthereumBlockFilter> for EthereumCallFilter {
     fn from(ethereum_block_filter: EthereumBlockFilter) -> Self {
+        EthereumCallFilter::from(&ethereum_block_filter)
+    }
+}
+
+impl From<&EthereumBlockFilter> for EthereumCallFilter {
```
leoyvens, line 409: Generally if a `From<&T>` impl is effectively cloning `T` internally as this is doing, I prefer having only the `From<T>` impl and requiring the caller to clone if they want to use it on an `&T`, to make the potential cost of cloning explicit.

Feel free to merge after addressing this.

leoyvens (approved):

chain/ethereum/src/adapter.rs
maoueh: It's not exact that it's cloning `T` directly, it's actually cloning only what is required for the transformation to work. I  don't see why cloning `EthereumBlockFilter ` would be necessary when we only care about some parts of the actual `EthereumBlockFilter` struct. The overprice is not quite big here, but still present. We have other occurrences of this pattern in other location like `BlockPtr::From<EthereumBlock>`.

I do understand the concern about the implicit "copy", but any `From<>` calls does some kind of allocation, so it's already implied there is some allocations cost when dealing with `From<>` calls.

leoyvens: True that this can be implemented more efficiently on `&T` since it only clones the necessary bits, but it does pessimize the implementation for `T` since that now clones when previously it moved the addresses. Is `From< EthereumBlockFilter>` even used anywhere? If not we can just remove it.

maoueh: Ok, I see the point about the other version being now less efficient. I removed it and every places now uses the reference version.

PR #3085, opened 2021-12-17
refactor stream

mangas:


chain/near/src/chain.rs
```diff
@@ -151,6 +151,17 @@ impl Blockchain for Chain {
...
 
+    async fn new_polling_block_stream(
+        &self,
+        _deployment: DeploymentLocator,
+        _start_blocks: Vec<BlockNumber>,
+        _filter: Arc<Self::TriggerFilter>,
+        _metrics: Arc<BlockStreamMetrics>,
+        _unified_api_version: UnifiedMappingApiVersion,
+    ) -> Result<Box<dyn BlockStream<Self>>, Error> {
+        todo!()
```
leoyvens, line 162:
```suggestion
        panic!("NEAR does not support polling block stream")
```

chain/ethereum/src/chain.rs
```diff
@@ -300,20 +302,47 @@ impl Blockchain for Chain {
...
-                .await
-        } else {
-            self.new_polling_block_stream(
-                deployment,
-                start_blocks,
-                adapter,
-                filter,
-                metrics,
-                unified_api_version,
+        self.new_firehose_block_stream(
```
leoyvens, line 305: Seems like we don't need a separate two fns with the same name so this can be inlined here.

chain/ethereum/src/chain.rs
```diff
@@ -300,20 +302,47 @@ impl Blockchain for Chain {
...
+                metrics.stopwatch.clone(),
             )
-            .await
-        }
+            .expect(&format!(
+                "no adapter for network {} with capabilities {}",
+                self.name, requirements
+            ));
+
+        self.new_polling_block_stream(
```
leoyvens, line 337: Same here.

leoyvens (changes requested):

leoyvens:
Do you plan on making the firehose cursor and the block pointer inputs to the block stream in this PR?

mangas:
> Do you plan on making the firehose cursor and the block pointer inputs to the block stream in this PR?

Actually I'll do that as well in this one

chain/ethereum/src/chain.rs
mangas: done

chain/ethereum/src/chain.rs
mangas: done

leoyvens (changes requested):
Now lets just make `update_subgraph_synced_status` in the polling block stream use the in-memory block pointer rather than `subgraph_store.block_ptr()`, and we'll have successfully decoupled the block stream from the subgraph store!

mangas:
> Now lets just make `update_subgraph_synced_status` in the polling block stream use the in-memory block pointer rather than `subgraph_store.block_ptr()`, and we'll have successfully decoupled the block stream from the subgraph store!

What do we do about [this](https://github.com/graphprotocol/graph-node/blob/ab2d36e19c84485d824a820708a7c7b0a1225983/graph/src/blockchain/polling_block_stream.rs#L505)?

Sounds like this also needs to be moved to the instance_manager?

leoyvens:
@mangas True. @otaviopace is working on moving that to the instance manager. But for now we can at least remove the `block_ptr` call.

mangas:
> @mangas True. @otaviopace is working on moving that to the instance manager. But for now we can at least remove the `block_ptr` call.

Done!

leoyvens (approved):

leoyvens:
Approved!

For future PRs, when addressing review it's easier to review again if commits are added instead of amended. You can of course then squash before merging.

PR #3087, opened 2021-12-18
Fixed BlockHandler with Call filter

### LINKED ISSUE #2314: block handler filters not working: handleBlockWithCallToContract
bug
      blockHandlers:
        - handler: handleBlock
        - handler: handleBlockWithCallToContract
          filter:
            kind: call     

`handleBlockWithCallToContract` never got called
just deploy subgraph with block handlers
```
      blockHandlers:
        - handler: handleBlock
        - handler: handleBlockWithCallToContract
          filter:
            kind: call     

```
`handleBlock` get called for all blocks while `handleBlockWithCallToContract` only get called when the block contains txs to the contract.

maoueh:
The `EthereumBlockFilter.extend` was not properly extending the base structure with the extension struct.

The new code reduces drasticly the cloning behavior that was happening way too much in the previous implementation.

Fixes #2314

### DIFF
--- a/chain/ethereum/src/adapter.rs
+++ b/chain/ethereum/src/adapter.rs
@@ -692,24 +692,25 @@ impl EthereumBlockFilter {
         } = other;
 
         self.trigger_every_block = self.trigger_every_block || trigger_every_block;
-        self.contract_addresses = self.contract_addresses.iter().cloned().fold(
-            HashSet::new(),
-            |mut addresses, (start_block, address)| {
-                match contract_addresses
-                    .iter()
-                    .cloned()
-                    .find(|(_, other_address)| &address == other_address)
-                {
-                    Some((other_start_block, address)) => {
-                        addresses.insert((cmp::min(other_start_block, start_block), address));
-                    }
-                    None => {
-                        addresses.insert((start_block, address));
+
+        for other in contract_addresses {
+            let (other_start_block, other_address) = other;
+
+            match self.find_contract_address(&other.1) {
+                Some((current_start_block, current_address)) => {
+                    if other_start_block < current_start_block {
+                        self.contract_addresses
+                            .remove(&(current_start_block, current_address));
+                        self.contract_addresses
+                            .insert((other_start_block, other_address));
                     }
                 }
-                addresses
-            },
-        );
+                None => {
+                    self.contract_addresses
+                        .insert((other_start_block, other_address));
+                }
+            }
+        }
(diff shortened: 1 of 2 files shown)


leoyvens:
Nice work with the tests! Just to be complete, could you add a test that parses both a trigger on every block and a trigger with call filters?

maoueh:
@leoyvens Not sure I get you here, what test you want me t The code is this PR is on fixing and testing `extend` feature of independent XXXFilter. 

You would like me to add test for `parse_block_triggers` that contains both an `EthereumBlockFilter` and a `EthereumCallFilter`?

leoyvens:
> Or a extend test where one every_block: true and some block addresses to extend?

This

maoueh:
Done

leoyvens:
Because this would change the PoI of existing subgraphs that use this feature, we would need to implement a migration logic that detects and resync affected subgraphs. Given that difficulty, it seems simpler to implement this fix by bumping the api version.

azf20:
I think is only one live subgraph on the network which implements this feature.
[Network subgraph](https://thegraph.com/hosted-service/subgraph/graphprotocol/graph-network-mainnet)
```
{
  subgraphDeployments(where: { manifest_contains: "kind: call", subgraphCount_gt: 0, signalledTokens_gt: 0} ) {
    id
    ipfsHash
    subgraphCount
    signalledTokens
    versions {
      subgraph {
        id
      }
    }
    indexerAllocations(where: {activeForIndexer_not: null}) {
      id
      indexer {
        id
      }
    }
    manifest
    }
}
```
Given that it is only the [NFTX subgraph](https://thegraph.com/explorer/subgraph?id=4gZf3atMXjYDh4g48Zr83NFX3rkvZED86VqMNhgEXgLc&view=Overview), I think we could reach out to affected indexers specifically to resync (and this also limits the scope of any subsequent arbitration). Is there a simple way to resync, without degrading availability of the subgraph on the network during resyncing?

leoyvens:
That manual process seems reasonable. Ideally that could be achieved through a graphman command (or sequence of commands), but we'd need to implement that first.

leoyvens:
We've decided to merge this, and then as part of the release process have the affected subgraph on the network be resynced. @maoueh thanks for the patience on this one, could you rebase it?

leoyvens:
Rebased this.

leoyvens (approved):

PR #3093, opened 2021-12-20
graph-node: factorize some methods inside a shared component

maoueh:
This PR is a first step in merging #3079. In this upcoming, we had copied a bunch of methods from `main.rs`, mainly those around instantiating elements to connect to chains and prepare them.

In this PR, we moved
- `create_ipfs_clients`
- `create_ethereum_networks` (formely `create_networks`)
- `create_firehose_networks`
- `connect_ethereum_networks`
- `connect_firehose_networks`

To a common module `chain` under `graph-node` crate.

There is no functionality changed in this PR, only moving code around. Small differences are the renamed of `create_networks` and its `config` argument that is now received as a reference to `Config` (instead of a owned `Config`).


lutter (approved):
Nice! `main.rs` is way too big as it is, I am glad we are starting to move some of the functionality out of it.

PR #3094, opened 2021-12-21
store: Handle an unused deployment becoming used again

lutter:


store/postgres/src/primary.rs
```diff
@@ -1413,6 +1413,17 @@ impl<'a> Connection<'a> {
         Ok(())
     }
 
+    /// The deployment `site` that we marked as unused previously is in fact
+    /// no used again, e.g., because it was redeployed in between recording
```
leoyvens, line 1417:
```suggestion
    /// now used again, e.g., because it was redeployed in between recording
```

leoyvens (approved):

store/postgres/src/primary.rs
lutter: Fixed the tyop

PR #3095, opened 2021-12-22
stream pipelining

mangas:


graph/src/blockchain/block_stream.rs
```diff
@@ -6,10 +6,33 @@ use thiserror::Error;
...
 use crate::components::store::BlockNumber;
 use crate::firehose::bstream;
+use crate::prelude::tokio::sync::mpsc::Sender;
 use crate::{prelude::*, prometheus::labels};
 
+pub async fn stream_to_channel<C: Blockchain>(
+    stream: Box<dyn BlockStream<C>>,
+    guard: Arc<SharedCancelGuard>,
+    sender: Sender<Result<BlockStreamEvent<C>, CancelableError<Error>>>,
+) -> Result<(), Error> {
```
leoyvens, line 16: I think it would be more encapsulated and require less changes to the instance manager if this looks like
```rust
pub async fn spawn_buffered(stream: Box<dyn BlockStream<C>>, buffer_size: usize) -> Box<dyn BlockStream<C>>
```
This would internally spawn a task, create a channel, and wrap the receiver in a stream, creating a new block stream. This does not need to handle cancellation, the cancellation code in the instance manager can remain as it is today, which means the receiver will be canceled rather than the block stream directly, but that's fine since ultimately that should cause the block stream task to terminate.

leoyvens (commented):
A more encapsulated approach which I'd suggest is to introduce a `BufferedBlockStream`, which takes as input a block stream and a buffer size, and returns

core/src/subgraph/instance_manager.rs
```diff
@@ -406,18 +406,19 @@ where
...
+        let inputs = IndexingInputs {
+            deployment: deployment.clone(),
+            features,
+            start_blocks,
+            store,
+            triggers_adapter,
+            chain,
+            templates,
+            unified_api_version,
+        };
```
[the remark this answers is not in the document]
mangas: the ctx needs to be mut borrowed at some point and this doesn't allow to re-use it unless it implements Clone or something like that. I thought splitting it was the best option because it is read-only and this way we don't need to clone it everywhere.

graph/src/blockchain/block_stream.rs
mangas: generally speaking I think it makes sense to leave concurrency handling to caller because it improves testing and is more flexible. 

This also doesn't work very well because we need ways of cancelling which means we actually need to pass Cancellable everywhere. I had some issues getting those types to line up and play nice with async. 

My preference would be to prioritise the changes to be testable over being "concise to use"

leoyvens: On testability, I agree that a function which spawns a task is harder to test. But as we discussed we can have the best of both options by keeping `stream_to_channel`, for testability, and have `spawn_buffered` call it so we can have a cleaner interface and avoiding pushing this extra bit of complexity directly into the instance manager.

On cancellation, as we discussed cancelling the receiver should be fine.

otaviopace (commented):
LGTM, still I think it's worth to have @leoyvens take a last look

core/src/subgraph/instance_manager.rs
```diff
@@ -56,7 +57,7 @@ lazy_static! {
 
 type SharedInstanceKeepAliveMap = Arc<RwLock<HashMap<DeploymentId, CancelGuard>>>;
 
-struct IndexingInputs<C: Blockchain> {
+pub struct IndexingInputs<C: Blockchain> {
```
leoyvens, line 60: No need for `pub`.

graph/src/blockchain/block_stream.rs
```diff
@@ -1,15 +1,89 @@
...
+        stream: Box<dyn BlockStream<C>>,
+        size_hint: usize,
+    ) -> Box<dyn BlockStream<C>> {
+        let (sender, receiver) = mpsc::channel::<Result<BlockStreamEvent<C>, Error>>(size_hint);
+        crate::spawn(async move { BufferedBlockStream::stream_blocks(stream, sender).await });
+
+        Box::new(BufferedBlockStream::new(receiver))
+    }
+
+    pub fn new(mut receiver: Receiver<Result<BlockStreamEvent<C>, Error>>) -> Self {
```
leoyvens, line 28: Doesn't need to be `pub`.

graph/src/blockchain/block_stream.rs
```diff
@@ -1,15 +1,89 @@
...
+                yield event
+            }
+        };
+
+        Self {
+            inner: Box::pin(inner),
+        }
+    }
+
+    pub async fn stream_blocks(
```
leoyvens, line 45: Doesn't need to be `pub`.

graph/src/blockchain/block_stream.rs
```diff
@@ -169,3 +243,89 @@ pub trait ChainHeadUpdateListener: Send + Sync + 'static {
...
+
+                    break;
+                }
+                Some(Ok(BlockStreamEvent::ProcessBlock(block_triggers, _))) => {
+                    let block = block_triggers.block;
+                    blocks.insert(block.clone());
+                    count += 1;
+
+                    if block.number > initial_block + buffer_size as u64 {
+                        guard.cancel();
```
leoyvens, line 318: Testing cancellation, nice!

graph/src/blockchain/mock.rs
```diff
@@ -0,0 +1,371 @@
...
+
+    type TriggerData = MockTriggerData;
+
+    type MappingTrigger = MockMappingTrigger;
+
+    type TriggerFilter = MockTriggerFilter;
+
+    type NodeCapabilities = MockNodeCapabilities;
+
+    type IngestorAdapter = MockIngestorAdapter;
```
leoyvens, line 310: This is gone on master now, so one less thing here.

runtime/wasm/src/host_exports.rs
```diff
@@ -660,7 +660,7 @@ impl<C: Blockchain> HostExports<C> {
 
         // Remember that we need to create this data source
         state.push_created_data_source(DataSourceTemplateInfo {
-            template,
+            template: template.clone(),
```
leoyvens, line 663: Do you understand why moving the bound from the trait to the associated type resulted in this being necessary? This is confusing to me.

graph/src/blockchain/mod.rs
```diff
@@ -81,13 +82,13 @@ pub trait Blockchain: Debug + Sized + Send + Sync + Unpin + 'static {
...
 
-    type DataSourceTemplate: DataSourceTemplate<Self>;
+    type DataSourceTemplate: DataSourceTemplate<Self> + Clone;
     type UnresolvedDataSourceTemplate: UnresolvedDataSourceTemplate<Self>;
 
     type TriggersAdapter: TriggersAdapter<Self>;
 
     /// Trigger data as parsed from the triggers adapter.
-    type TriggerData: TriggerData + Ord;
+    type TriggerData: TriggerData + Ord + Send + Sync + Clone;
```
leoyvens, line 91: The `Clone` bound is not necessary.

leoyvens (changes requested):

core/src/subgraph/instance_manager.rs
```diff
@@ -480,51 +519,28 @@ where
...
-        }
-        .await?
-        .map_err(CancelableError::Error)
-        .cancelable(&block_stream_canceler, || Err(CancelableError::Cancel));
+
+        let metrics = ctx.block_stream_metrics.clone();
+        let filter = ctx.state.filter.clone();
+        let stream_inputs = inputs.clone();
+        let stream = new_block_stream(stream_inputs, filter, metrics).await?;
+        let mut stream = BufferedBlockStream::spawn_from_stream(stream, BUFFERED_BLOCK_STREAM_SIZE)
```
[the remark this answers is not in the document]
mangas: These are consumed one by one, so in the case of non-firehose as soon as a single slot is available in the worst case, a new fetch is triggered, don't think we really need to set it to MAX_SIZE for that reason, right?

core/src/subgraph/instance_manager.rs
```diff
@@ -480,51 +519,28 @@ where
...
             .unwrap()
-            .insert(ctx.inputs.deployment.id, block_stream_canceler);
+            .insert(inputs.deployment.id, block_stream_canceler);
 
         debug!(logger, "Starting block stream");
 
         // Process events from the stream as long as no restart is needed
         loop {
-            let (block, cursor) = match block_stream.next().await {
+            let (block, cursor) = match stream.next().await {
```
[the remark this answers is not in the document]
mangas: agreed, will rename

graph/src/blockchain/block_stream.rs
mangas: I disagree, I think the whole point is to expose the functions in a way they can be used separate but I can remove it

graph/src/blockchain/mod.rs
mangas: hmm it was needed for something but I can't remember what for. I'll remove it and see if everything still works 😛

runtime/wasm/src/host_exports.rs
mangas: This crate doesn't compile for me locally, it might not be necessary. Will look into it

graph/src/blockchain/block_stream.rs
leoyvens: Ok if you want to keep it.

graph/src/blockchain/mod.rs
mangas: indeed seems like it's not needed, was a leftover from the previous version

core/src/subgraph/instance_manager.rs
```diff
@@ -30,6 +30,8 @@ use std::time::{Duration, Instant};
 use tokio::task;
 
 const MINUTE: Duration = Duration::from_secs(60);
+const BUFFERED_BLOCK_STREAM_SIZE: usize = 100;
```
leoyvens, line 33: `BUFFERED_POLLING_BLOCK_STREAM_SIZE` would be more accurate.

leoyvens (approved):

leoyvens (approved):

PR #3096, opened 2021-12-23
graph,graphql,store: nested sorting

kamilkisiela:
#3737

Given this schema:

```graphql
type Country @entity {
  id: ID!
  capital: City!
  cities: [City!]!
}

type City @entity {
  id: ID!
  name: String
  population: Int
  landmarks: [String!]!
}
```

It adds fields of child entity to `*_orderBy` enum:

```diff
enum Country_orderBy {
   id
   capital
   cities
+  capital__id
+  capital__name
+  capital__population
}
```

- [x] sort by `type Child @entity { name: String }`
- [x] sort by `type Child @entity { id: ID }` 
- [x] can't sort by `type Child @entity { names: [String] }`
- [x] can't sort by `type Child @entity { secondLevelChild: AnotherEntity }`
- [x] can't sort by `type Child @entity { secondLevelChildren: [AnotherEntity] }`
- [x] support `@derivedFrom`
- [x] sort by `interface Child { name: String }`
- [x] sort by `interface Child { id: ID }`
- [ ] sort interface by `type Child @entity { name: String }` (done in https://github.com/graphprotocol/graph-node/pull/4058)

### DIFF
--- a/graph/src/components/store/mod.rs
+++ b/graph/src/components/store/mod.rs
@@ -280,6 +280,24 @@ impl EntityFilter {
     }
 }
 
+/// Holds the information needed to query a store.
+#[derive(Clone, Debug, PartialEq)]
+pub struct EntityOrderByChildInfo {
+    /// The attribute of the child entity that is used to order the results.
+    pub sort_by_attribute: Attribute,
+    /// The attribute that is used to join to the parent and child entity.
+    pub join_attribute: Attribute,
+    /// If true, the child entity is derived from the parent entity.
+    pub derived: bool,
+}
+
+/// Holds the information needed to order the results of a query based on nested entities.
+#[derive(Clone, Debug, PartialEq)]
+pub enum EntityOrderByChild {
+    Object(EntityOrderByChildInfo, EntityType),
+    Interface(EntityOrderByChildInfo, Vec<EntityType>),
+}
+
 /// The order in which entities should be restored from a store.
 #[derive(Clone, Debug, PartialEq)]
 pub enum EntityOrder {
@@ -287,6 +305,10 @@ pub enum EntityOrder {
     Ascending(String, ValueType),
     /// Order descending by the given attribute. Use `id` as a tie-breaker
     Descending(String, ValueType),
+    /// Order ascending by the given attribute of a child entity. Use `id` as a tie-breaker
+    ChildAscending(EntityOrderByChild),
+    /// Order descending by the given attribute of a child entity. Use `id` as a tie-breaker
+    ChildDescending(EntityOrderByChild),
     /// Order by the `id` of the entities
     Default,
     /// Do not order at all. This speeds up queries where we know that
--- a/graphql/src/schema/api.rs
+++ b/graphql/src/schema/api.rs
@@ -1,6 +1,6 @@
 use std::str::FromStr;
 
-use graphql_parser::Pos;
+use graphql_parser::{schema::TypeDefinition, Pos};
 use inflector::Inflector;
 use lazy_static::lazy_static;
 
@@ -157,16 +157,7 @@ fn add_order_by_type(
                 description: None,
                 name: type_name,
                 directives: vec![],
-                values: fields
-                    .iter()
-                    .map(|field| &field.name)
-                    .map(|name| EnumValue {
-                        position: Pos::default(),
-                        description: None,
-                        name: name.to_owned(),
-                        directives: vec![],
-                    })
-                    .collect(),
+                values: field_enum_values(schema, fields)?,
             });
             let def = Definition::TypeDefinition(typedef);
             schema.definitions.push(def);
@@ -174,6 +165,80 @@ fn add_order_by_type(
         Some(_) => return Err(APISchemaError::TypeExists(type_name)),
     }
     Ok(())
+}
+
+/// Generates enum values for the given set of fields.
+fn field_enum_values(
+    schema: &Document,
+    fields: &[Field],
+) -> Result<Vec<EnumValue>, APISchemaError> {
+    let mut enum_values = vec![];
+    for field in fields {
+        enum_values.push(EnumValue {
+            position: Pos::default(),
+            description: None,
+            name: field.name.to_owned(),
+            directives: vec![],
+        });
+        enum_values.extend(field_enum_values_from_child_entity(schema, field)?);
+    }
+    Ok(enum_values)
+}
+
+fn enum_value_from_child_entity_field(
+    schema: &Document,
+    parent_field_name: &str,
+    field: &Field,
+) -> Option<EnumValue> {
+    if ast::is_list_or_non_null_list_field(field) || ast::is_entity_type(schema, &field.field_type)
+    {
+        // Sorting on lists or entities is not supported.
+        None
+    } else {
+        Some(EnumValue {
+            position: Pos::default(),
+            description: None,
+            name: format!("{}__{}", parent_field_name, field.name),
+            directives: vec![],
+        })
+    }
+}
+
+fn field_enum_values_from_child_entity(
+    schema: &Document,
+    field: &Field,
+) -> Result<Vec<EnumValue>, APISchemaError> {
+    fn resolve_supported_type_name(field_type: &Type) -> Option<&String> {
+        match field_type {
+            Type::NamedType(name) => Some(name),
+            Type::ListType(_) => None,
+            Type::NonNullType(of_type) => resolve_supported_type_name(of_type),
+        }
+    }
+
+    let type_name = match ENV_VARS.graphql.disable_child_sorting {
+        true => None,
+        false => resolve_supported_type_name(&field.field_type),
+    };
+
+    Ok(match type_name {
+        Some(name) => {
+            let named_type = schema
+                .get_named_type(name)
+                .ok_or_else(|| APISchemaError::TypeNotFound(name.clone()))?;
+            match named_type {
+                TypeDefinition::Object(ObjectType { fields, .. })
+                | TypeDefinition::Interface(InterfaceType { fields, .. }) => fields
+                    .iter()
+                    .filter_map(|f| {
+                        enum_value_from_child_entity_field(schema, field.name.as_str(), f)
+                    })
+                    .collect(),
+                _ => vec![],
+            }
+        }
+        None => vec![],
+    })
 }
 
 /// Adds a `<type_name>_filter` enum type for the given fields to the schema.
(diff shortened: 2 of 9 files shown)


graphql/src/store/query.rs
```diff
@@ -297,14 +337,23 @@ fn build_order_by(
     }
 }
 
+/// Kamil: I don't like it... What if we have `_` in the field name?
+fn parse_field_as_order(field_name: &String) -> (&str, Option<&str>) {
```
kamilkisiela, line 341: In the following scenario we end up with `foo_bar` conflict.

```
Purpose {
  foo
  foo_bar
}

Sender {
  bar
}
```

graphql/src/store/query.rs
```diff
@@ -297,14 +337,23 @@ fn build_order_by(
     }
 }
 
+/// Kamil: I don't like it... What if we have `_` in the field name? For now I will use double underscore
+fn parse_field_as_order(field_name: &String) -> (&str, Option<&str>) {
+    let mut field_name_parts = field_name.split("__");
```
kamilkisiela, line 342: Used double underscore for now

store/postgres/src/relational_queries.rs
```diff
@@ -2560,7 +2595,9 @@ impl<'a> FilterQuery<'a> {
         write_column_names(&column_names, &table, &mut out)?;
         self.filtered_rows(table, filter, out.reborrow())?;
         out.push_sql("\n ");
-        self.sort_key.order_by(&mut out, Some("c."))?;
+        let join_table_collection = filter.as_ref().map(|f| &f.table_join_collection);
```
kamilkisiela, line 2598: - [ ] allow to extend the JoinTableCollection with SortKey (prefix won't be available if EntityFilter::Child is not used)

graphql/src/schema/api.rs
```diff
@@ -317,6 +308,69 @@ fn add_order_by_type(
...
+                                description: None,
+                                name: format!("{}__{}", field.name, f.name),
+                                directives: vec![],
+                            })
+                            .collect()
+                    }
+                }
+                // sorting for interfaces is not supported yet
+                // hard to pick correct table and pass the EntityType to EntityOrder::Ascending(EntityType)
+                _ => vec![],
```
kamilkisiela, line 364: Something to consider

graphql/src/store/query.rs
lutter: For now, double underscore is fine, but another option would be to use an input object: `orderBy: { child_: field }` but I think that runs into the same problem we had elsewhere that input types can't be unions (since we still need `orderBy: parentField`) A third option would be to have a separate `orderByChild: field`, i.e. for a `books` subfield you'd write `orderByBooks: title`

graphql/src/schema/api.rs
```diff
@@ -176,6 +167,77 @@ fn add_order_by_type(
...
+                TypeDefinition::Interface(i) => i
+                    .fields
+                    .iter()
+                    .filter_map(|f| enum_value_from_child_entity_field(schema, field, f))
+                    .collect(),
+                _ => vec![],
+            })
+        }
+        // Sort on lists is not supported.
+        Type::ListType(ref _t) => Ok(vec![]),
```
lutter, line 236: This should return an error; we assume that this is never called with a list type, if it is there's a programming error somewhere and we should call attention to it.

graphql/src/store/query.rs
```diff
@@ -360,27 +396,145 @@ fn list_values(value: Value, filter_type: &str) -> Result<Vec<Value>, QueryExecu
...
 }
 
+enum OrderByValue {
+    Direct(String),
+    Child(String, String),
+}
+
+fn parse_order_by(enum_value: &String) -> OrderByValue {
+    let mut parts = enum_value.split("__");
+    let first = parts.next().expect("some dark magic happened");
```
lutter, line 406: Since this is called from `build_order_by` which already returns `Result<_,_>`, make this return a `Result`, too, so we don't have a panic when something goes wrong, but rather an error that gets sent back to the user. (Though I also don't see how `next()` could be `None` here).

The general problem with `expect` and `unwrap` is that they cause panics which are much harder on the system to handle than an error, making it possible to cause trouble by sending really badly behaving queries.

graphql/src/store/query.rs
```diff
@@ -360,27 +396,145 @@ fn list_values(value: Value, filter_type: &str) -> Result<Vec<Value>, QueryExecu
...
+                    QueryExecutionError::EntityFieldError(entity.name().to_owned(), name.clone())
+                })?;
+                sast::get_field_value_type(&field.field_type)
+                    .map(|value_type| Some((name.to_owned(), value_type, None)))
+                    .map_err(|_| {
+                        QueryExecutionError::OrderByNotSupportedError(
+                            entity.name().to_owned(),
+                            name.clone(),
+                        )
+                    })
```
lutter, line 451: This is preexisting code, but it also makes me wonder what happens if you manage to use a list or derived field here. I am not super confident that our current validation will catch sorting by such fields, even if they were not in the schema.

graph/src/components/store/mod.rs
```diff
@@ -244,13 +244,39 @@ impl EntityFilter {
     }
 }
 
+#[derive(Clone, Debug, PartialEq)]
+pub struct EntityOrderByChildObject {
+    pub entity_type: EntityType,
+    pub attribute: Attribute,
+    pub join_attribute: Attribute,
+    pub derived: bool,
```
lutter, line 252: Do we really need the derived here? If the derived field is derived from the parent type, then we can just turn this into an ordinary sort. If it is derived from a grandchild, we shouldn't allow ordering by it because we'd need to join parent, child, and grandchild which we don't do currently.

lutter (commented):
I started reviewing this, but then realized that it also deals with sorting by interface fields, and now I am not certain how it relates to PR #4058. Should I continue reviewing this or is this superseded by that PR?

graphql/src/schema/api.rs
kamilkisiela: Are you sure? I think it's fine, only the GraphQL Schema is generated here, it's not related to runtime, or GraphQL request.
Returning an empty vector means there are no enum values, which makes sense.

graph/src/components/store/mod.rs
kamilkisiela: The `derived` value applies to child <-> parent relation. There's of course a case where sorting by `id` could be done without a left join as it "sits" in one of the columns of the parent table. I'm not sure what exactly you mean here, maybe I'm missing something, but sorting rows from `parent` by one of the fields of `child` table requires a left join.

graphql/src/store/query.rs
kamilkisiela: Solved in 4f1adae

graphql/src/schema/api.rs
```diff
@@ -176,6 +167,77 @@ fn add_order_by_type(
...
+                    .iter()
+                    .filter_map(|f| enum_value_from_child_entity_field(schema, field, f))
+                    .collect(),
+                TypeDefinition::Interface(i) => i
+                    .fields
+                    .iter()
+                    .filter_map(|f| enum_value_from_child_entity_field(schema, field, f))
+                    .collect(),
+                _ => vec![],
+            })
```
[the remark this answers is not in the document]
kamilkisiela: Solved in b6573ff

graphql/src/schema/api.rs
```diff
@@ -176,6 +167,77 @@ fn add_order_by_type(
...
+            name: format!("{}__{}", parent_field.name, field.name),
+            directives: vec![],
+        })
+    }
+}
+
+fn field_enum_values_from_child_entity(
+    schema: &Document,
+    field: &Field,
+    field_type: &Type,
```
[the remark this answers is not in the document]
kamilkisiela: Solved in 8042010

kamilkisiela:
@lutter 
> I started reviewing this, but then realized that it also deals with sorting by interface fields, and now I am not certain how it relates to PR https://github.com/graphprotocol/graph-node/pull/4058. Should I continue reviewing this or is this superseded by that PR?

It deals with sorting by interface fields, #4058 adds sorting an interface by children (as it's a bit more complex).

graph/src/components/store/mod.rs
```diff
@@ -246,13 +246,39 @@ impl EntityFilter {
     }
 }
 
+#[derive(Clone, Debug, PartialEq)]
+pub struct EntityOrderByChildObject {
```
lutter, line 250: It would be good to have comments describing what this and the other `EntityOrderChild*` structs are and what the fields mean. It's sorta clear from reading the code, but would be good to spell it out.

store/postgres/src/relational_queries.rs
```diff
@@ -2527,9 +2527,26 @@ impl<'a> FilterCollection<'a> {
     }
 }
 
+#[derive(Debug, Clone)]
+pub struct ChildKeyDetails<'a> {
+    pub parent_table: &'a Table,
+    pub parent_column: &'a Column,
+    pub child_table: &'a Table,
+    pub child_column: &'a Column,
+    pub column: &'a Column,
```
lutter, line 2536: This seemed to be clear until I saw `column` - this would be good to comment too, and explain what the fields are for.

store/postgres/src/relational_queries.rs
```diff
@@ -2617,6 +2659,108 @@ impl<'a> SortKey<'a> {
...
+        ) -> Result<SortKey<'a>, QueryExecutionError> {
+            let sort_by_column = child_table.column_for_field(&attribute)?;
+            if sort_by_column.is_fulltext() {
+                Err(QueryExecutionError::NotSupported(
+                    "Sorting by fulltext fields".to_string(),
+                ))
+            } else if sort_by_column.is_primary_key() {
+                Err(QueryExecutionError::NotSupported(
+                    "Sorting by id".to_string(),
+                ))
```
lutter, line 2678: Why is this not allowed? The example for the PR mentions sorting by `capital__id`

store/postgres/src/relational_queries.rs
```diff
@@ -2617,6 +2659,108 @@ impl<'a> SortKey<'a> {
...
+            } else {
+                let (parent_column, child_column) = match derived {
+                    true => (
+                        parent_table.primary_key(),
+                        child_table
+                            .column_for_field(&join_attribute)
+                            .expect("Column for a join attribute not found"),
```
lutter, line 2685: The `expect` here (and all `expects`) should be turned into a `constraint_violation!`, otherwise queries that somehow make it here causes panics which are much more expensive than returning errors.

lutter (changes requested):
I am still not completely through the PR, but figured these comments would be good to address to make some progress.

store/postgres/src/relational_queries.rs
kamilkisiela: 1cc9623d5 is it ok?

store/postgres/src/relational_queries.rs
kamilkisiela: 34d214e05

store/postgres/src/relational_queries.rs
```diff
@@ -2767,17 +3000,176 @@ impl<'a> SortKey<'a> {
...
+                // Type B: c.id = p.{parent_field}
+                out.push_sql(prefix);
+                out.push_sql(".");
+                out.push_identifier(child_column.name.as_str())?;
+                out.push_sql(" = ");
+                out.push_sql("c.");
+                out.push_identifier(parent_column.name.as_str())?;
+            }
+
+            out.push_sql(format!(" and {prefix}.").as_str());
```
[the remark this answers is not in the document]
kamilkisiela: 6565f9368

store/postgres/src/relational_queries.rs
lutter: Yes, that's much clearer!

store/postgres/src/relational_queries.rs
kamilkisiela: I fixed it, only sorting interfaces by id is not supported so far

kamilkisiela:
@lutter about the failing test, it works locally but I will try to use exact version on PostgreSQL docker image.

kamilkisiela:
@lutter 

Rows of `musicians` table are sorted by `main_band, block_range` columns. You can see that `b1` is used by two musicians.

Sometimes the unit test is 🟢, but sometimes it's 🔴 . 

Do you think that the order of row insertion affects the end result? These two rows are pretty much identical (`block_range` and `main_band`).

| entity | data |
| :--- | :--- |
| Musician | {"id": "m2", "vid": 1, "name": "Lisa", "bands": \["b1"\], "main\_band": "b1", "block\_range": "\[0,\)"} |
| Musician | {"id": "m1", "vid": 2, "name": "John", "bands": \["b1", "b2"\], "main\_band": "b1", "block\_range": "\[0,\)"} |
| Musician | {"id": "m3", "vid": 3, "name": "Tom", "bands": \["b1", "b2"\], "main\_band": "b2", "block\_range": "\[1,\)"} |
| Musician | {"id": "m4", "vid": 4, "name": "Valerie", "bands": \[\], "main\_band": null, "block\_range": "\[1,\)"} |

graph/src/components/store/mod.rs
kamilkisiela: e9544aa1c

lutter:
> ```sql
> select 'Musician' as entity, to_jsonb(c.*) as data from (select  c.*
>  from "sgd21"."musician" c left join "sgd21"."band" as cc on (cc."id" = c."main_band" and cc."block_range" @> 1)
> where c.block_range @> 1
> order by cc."id", cc.block_range
> limit 100) c
> ```

I am not entirely sure what is going on here; for me locally, `can_query_with_sorting_by_child_entity_id` fails consistently. Is that the test you are talking about? If so, it should say `order by cc.name, cc.id, cc.block_range` based on the GraphQL for `asc`.

graphql/src/store/query.rs
```diff
@@ -53,15 +53,57 @@ pub(crate) fn build_query<'a>(
...
     ) {
-        (Some((attr, value_type)), OrderDirection::Ascending) => {
+        (Some((attr, value_type, None)), OrderDirection::Ascending) => {
             EntityOrder::Ascending(attr, value_type)
         }
-        (Some((attr, value_type)), OrderDirection::Descending) => {
+        (Some((attr, value_type, None)), OrderDirection::Descending) => {
             EntityOrder::Descending(attr, value_type)
         }
+        (Some((attr, _, Some(child))), OrderDirection::Ascending) => match child {
```
lutter, line 65: Put a check just before the match that if `ENV_VARS.graphql.disable_child_sorting` is set, this returns `QueryExecutionError::NotSupported("Sorting by child attributes is not supported")`

graphql/src/schema/api.rs
```diff
@@ -176,6 +167,77 @@ fn add_order_by_type(
...
+        match field_type {
+            Type::NamedType(name) => Some(name),
+            Type::ListType(_) => None,
+            Type::NonNullType(of_type) => resolve_supported_type_name(of_type),
+        }
+    }
+
+    let type_name = resolve_supported_type_name(&field.field_type);
+
+    Ok(match type_name {
```
lutter, line 221: Only do this if `ENV_VARS.graphql.disable_child_sorting` is `false`

lutter (approved):
Nice work! Sorry it took so long to review.

One final thing: we should have an environment variable `GRAPH_GRAPHQL_DISABLE_CHILD_SORTING` which defaults to `false` that we can set to turn this off if it leads to performance issues. The flag should work just the same as `GRAPH_GRAPHQL_DISABLE_BOOL_FILTERS` does. I left comments where I think this needs to be checked to (a) suppress child sorting in the GraphQL schema and (b) return an error in queries.

Other than that, this should be good to go.

graphql/src/store/query.rs
kamilkisiela: I did :)

lutter (approved):

dotansimha:
@lutter merged with the integration tests failing, it seems like the same on `master`. Unrelated to this PR, and the only change was a rebase.

PR #3108, opened 2022-01-03
PollingBlockStream: Move sync status logic to instance manager

otaviopace:
🚨  In progress 🚨


graph/src/blockchain/polling_block_stream.rs
```diff
@@ -215,8 +215,7 @@ where
                     continue;
                 }
                 ReconciliationStep::Done => {
-                    // Reconciliation is complete, so try to mark subgraph as Synced
-                    ctx.update_subgraph_synced_status()?;
+                    ctx.handle_subgraph_synced().await?;
```
otaviopace, line 218: This fails, I might have to make `is_deployment_synced` from the store `sync`.

スクリーンショット 2022-01-03 13 32 39

graph/src/blockchain/polling_block_stream.rs
```diff
@@ -487,23 +485,15 @@ where
...
-            Ok(())
-        } else {
-            // Synced
-
+    /// Stop metrics if subgraph is synced.
+    /// Sync mechanism is now implemented in the instance manager.
+    async fn handle_subgraph_synced(&self) -> Result<(), StoreError> {
+        if self.subgraph_store.is_deployment_synced().await? {
             // Stop recording time-to-sync metrics.
             self.metrics.stopwatch.disable();
```
leoyvens, line 493: This should be move to the instance manager, which should allow us to remove the dependency on the `subgraph_store` entirely.

core/src/subgraph/instance_manager.rs
```diff
@@ -1268,3 +1296,125 @@ fn persist_dynamic_data_sources<T: RuntimeHostBuilder<C>, C: Blockchain>(
     // Merge filters from data sources into the block stream builder
     ctx.state.filter.extend(data_sources.iter());
 }
+
+/// Returns true if the timer has elapsed given duration/threshold.
+fn has_threshold_passed(timer: Instant, threshold: Duration) -> bool {
+    timer.elapsed() >= threshold
+}
```
leoyvens, line 1303: Isn't the timer ever reset?

leoyvens (changes requested):

core/src/subgraph/instance_manager.rs
```diff
@@ -665,6 +676,23 @@ where
 
             match res {
                 Ok(needs_restart) => {
+                    if should_try_update_sync_status(
+                        synced,
+                        sync_status_timer,
+                        *SYNC_STATUS_THRESHOLD,
+                        // This is the new deployment head since the block just has been processed.
+                        &block_ptr,
+                        || chain_store.chain_head_ptr(),
```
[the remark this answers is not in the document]
otaviopace: Passing the chain store unfortunately makes the test:

- depend on the database/store
- run slower

I ended up going for easier/faster testability because, as mentioned before, I really wanted to be sure the boolean part is correct, so we don't have to do patches on this.

If you think it really damages the readability, I can change this.

core/src/subgraph/instance_manager.rs
otaviopace: I had the reset, but I removed because it didn't seem needed, but now thinking again, the way it is right now it will try to check if it's synced for every processed block, which isn't ideal/efficient.

I'll change this, thanks for pointing out!

graph/src/blockchain/polling_block_stream.rs
otaviopace: I'll refactor this!

core/src/subgraph/instance_manager.rs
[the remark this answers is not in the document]
leoyvens: I agree testability is more important here. You could mock the `ChainStore` trait but that would be too much boilerplate unless you split this method to a separate trait. So ok, we can keep this as is.

core/src/subgraph/instance_manager.rs
```diff
@@ -1268,3 +1296,125 @@ fn persist_dynamic_data_sources<T: RuntimeHostBuilder<C>, C: Blockchain>(
...
+    let now = Instant::now();
+    assert!(!has_threshold_passed(now, MINUTE * 3));
+    assert!(!has_threshold_passed(now - MINUTE * 2, MINUTE * 3));
+    assert!(has_threshold_passed(now - MINUTE * 5, MINUTE * 3));
+}
+
+/// Checks if two BlockPtrs are the same.
+fn is_deployment_synced(deployment_head_ptr: &BlockPtr, chain_head_ptr: Option<BlockPtr>) -> bool {
+    matches!((deployment_head_ptr, &chain_head_ptr), (b1, Some(b2)) if b1 == b2)
+}
```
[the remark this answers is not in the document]
otaviopace: In this scenario, would it be safe to just check `BlockNumber`s? I mean in the sense that to compare `chain_head - 1` I would have to do another `store` call to get the `parent_ptr`.

core/src/subgraph/instance_manager.rs
otaviopace: I just reverted the commit that does this 😊

graph/src/blockchain/polling_block_stream.rs
otaviopace: Fixed! Also I've removed the unnecessary `+ Send` trait bound of `Blockchain::TriggerData` since it's not needed anymore.

If you think I should keep it, please inform me 😊

core/src/subgraph/instance_manager.rs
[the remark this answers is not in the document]
leoyvens: Yes, this can generally be just a block number comparison, the notion of 'synced' doesn't have to be exact.

graph/src/blockchain/polling_block_stream.rs
otaviopace: I removed the dependency completely!!!

core/src/subgraph/instance_manager.rs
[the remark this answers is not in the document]
otaviopace: Added the `sub.block >= (chain.block - 1)`

otaviopace:
@leoyvens I've addressed your comments and implemented the chain head block cache

core/src/subgraph/instance_manager.rs
```diff
@@ -721,6 +725,24 @@ where
 
             match res {
                 Ok(needs_restart) => {
+                    if should_try_update_sync_status(
```
leoyvens, line 728: I'd inline `should_try_update_sync_status` here, `!synced` is such a simple condition that it's not worth testing for imo. This way you don't need the closure anymore. But up to you.

leoyvens (commented):
This approach really is way simpler!

core/src/subgraph/instance_manager.rs
```diff
@@ -706,6 +709,23 @@ where
 
             match res {
                 Ok(needs_restart) => {
+                    // Once synced, no need to try to update the status again.
+                    if !synced &&
+                        // deployment.head == chain.head
```
leoyvens, line 714: I'd remove this comment, it looks like commented out code

leoyvens (approved):

core/src/subgraph/instance_manager.rs
otaviopace: Removed!

core/src/subgraph/instance_manager.rs
otaviopace: Done!

PR #3121, opened 2022-01-04
Include missing arguments to format string

tilacog:
Run into this when trying out an updated version of the `anyhow` crate.


maoueh:
Duplicate of #3107.

tilacog:
Thanks! I'll close this as it is a duplicate. Let's track it in the previous PR.

PR #3122, opened 2022-01-04
chain/near,chain/ethereum: improved revert code and fixed NEAR on revert to pass parent_ptr

maoueh:
This PR is a continuation of #2979 adding the required support for NEAR. 

At the same time, this PRs turns some no-op silent returns into panics in NEAR, all those around the `IngestorTrait` now that it's clear it's an Ethereum specific thing. We should probably actually move it out completely to an `EthereumBlockIngestor` and refactor the `Chain` implementation to get rid of it. I did not investigate the implications and might prove hard.

### DIFF
--- a/chain/ethereum/src/chain.rs
+++ b/chain/ethereum/src/chain.rs
@@ -602,17 +602,17 @@ impl FirehoseMapperTrait<Chain> for FirehoseMapper {
                 ))
             }
 
-            StepUndo => Ok(BlockStreamEvent::Revert(
-                BlockPtr {
-                    hash: BlockHash::from(block.hash),
-                    number: block.number as i32,
-                },
-                FirehoseCursor::Some(response.cursor.clone()),
-                Some(BlockPtr {
-                    hash: BlockHash::from(block.header.unwrap().parent_hash),
-                    number: (block.number.checked_sub(1).unwrap() as i32), // Will never receive undo on blocknum 0
-                }),
-            )),
+            StepUndo => {
+                let parent_ptr = block
+                    .parent_ptr()
+                    .expect("A reverted block should always have a parent");
+
+                Ok(BlockStreamEvent::Revert(
+                    block.ptr(),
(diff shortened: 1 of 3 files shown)


chain/ethereum/src/chain.rs
```diff
@@ -602,17 +602,17 @@ impl FirehoseMapperTrait<Chain> for FirehoseMapper {
...
-                FirehoseCursor::Some(response.cursor.clone()),
-                Some(BlockPtr {
-                    hash: BlockHash::from(block.header.unwrap().parent_hash),
-                    number: (block.number.checked_sub(1).unwrap() as i32), // Will never receive undo on blocknum 0
-                }),
-            )),
+            StepUndo => {
+                let parent_ptr = block
+                    .parent_ptr()
+                    .expect("A reverted block should always have a parent");
```
leoyvens, line 608: This message can be more specific by saying that block 0 is never reverted.

chain/near/src/chain.rs
```diff
@@ -321,14 +308,14 @@ impl FirehoseMapperTrait<Chain> for FirehoseMapper {
 
             StepUndo => {
                 let header = block.header();
+                let parent_ptr = header
+                    .parent_ptr()
+                    .expect("A reverted block should always have a parent");
```
leoyvens, line 313: Same.

leoyvens (approved):
Agreed that we can remove the BlockIngestor associated type out of the Blockchain trait, since as it is it's not really abstracting over anything and just getting a bit in the way for firehose chains. Don't know if @lutter agrees.

Doesn't seem too difficult but if you don't want to tackle that now you can just open an issue for it.

chain/ethereum/src/chain.rs
maoueh: I'll will use `genesis block` though, doing it right now.

maoueh:
@leoyvens Block ingestor moved in #3131

PR #3126, opened 2022-01-06
store: make compare functionality case-insensitive

### LINKED ISSUE #3116: Change `_contains` `_starts_with`, `_ends_with` functionality to be case insensitive
> Making `contains` and `starts_with` case insensitive is a trivial change in the backend (replacing `like` with `ilike` in queries) One small wrinkle is that case-sensitivity is locale-dependent, and the response will depend on the locale in which the database is running, but that's not a new thing, we have that with ordering too. More a reminder that we need to define the locale for all these operations for the network.

_Originally posted by @lutter in https://github.com/graphprotocol/graph-node/issues/1837#issuecomment-981943733_

saihaj:
Closes #3116

### DIFF
--- a/core/tests/interfaces.rs
+++ b/core/tests/interfaces.rs
@@ -67,7 +67,8 @@ async fn one_interface_zero_entities() {
         .unwrap();
 
     let data = extract_data!(res).unwrap();
-    assert_eq!(format!("{:?}", data), "Object({\"leggeds\": List([])})")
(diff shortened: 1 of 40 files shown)


lutter (approved):
Nice!

One note on the format of our commit messages: they should generally start with the name of the crate(s) that are affected, or, if a change touches too many crates, with `all: ..`. So, in this case the commit message should be `store: Make compare functionality case-insensitive`. No need to fix now, just something for future PR's

PR #3127, opened 2022-01-06
feat: add strict filter for case sensitive matching

### LINKED ISSUE #3117: Add new filters for case sensitive match `_contains_strict` , `_starts_with_strict`, `_ends_with_strict`

saihaj:
Adds the following filters
- `_contains_strict`
- `_not_contains_strict`
- `_start_with_strict`
- `_not_start_with_strict`
- `_ends_with_strict`
- `_not_ends_with_strict`

Closes #3117


store/postgres/src/relational_queries.rs
```diff
@@ -855,17 +861,18 @@ impl<'a> QueryFilter<'a> {
         attribute: &Attribute,
         value: &Value,
         negated: bool,
+        strict: bool,
         mut out: AstPass<Pg>,
     ) -> QueryResult<()> {
         let column = self.column(attribute);
-
+        let operation = if strict { "like" } else { "ilike" };
```
lutter, line 868: Better would be to add this only to the `Value::String(s)` match arm:
```
let operation = match (strict, negated) {
  (true, true) => " not like ",
  (true, false) => " like ",
 (false, true) => " not ilike ",
 (false, false) => " ilike ",
};
```
and then have `out.push_sql(operation)` instead of the `if negated { .. }` statement. That avoids allocations - this code is run several times for each query, and tries to be as careful with allocations as possible.

lutter (changes requested):

PR #3129, opened 2022-01-06
node: fix a potential crash when Firehose provider is faulty

maoueh:
When a provider is faulty, it's removed from the list of endpoints for this particular chain. This means later on, places where at least one valid provide must exists panicked, this is the case for the `FirehoseBlockIngestor`.

As well as removing the faulty provider in `connect_firehose_networks`, we also completely remove the chain itself if at the end the chain has 0 providers.

Alternative would be to keep the chain but make downstream users to deal with the fact that providers could be 0. I decided on removing at the source instead since a chain with 0 providers is like no chain at all since nothing could run.

### Renames

I took the opportunity to trimmed down the structs where firehose providers and chain id to providers are stored. I was always finding hard to reason about those struct where there is a lot of repition in the name and the single fields. Hopefully it's a bit better now. I will highlight the bug fix in the diff.

### DIFF
--- a/chain/ethereum/src/chain.rs
+++ b/chain/ethereum/src/chain.rs
@@ -1,7 +1,7 @@
 use anyhow::{Context, Error};
 use graph::blockchain::BlockchainKind;
 use graph::data::subgraph::UnifiedMappingApiVersion;
-use graph::firehose::FirehoseNetworkEndpoints;
+use graph::firehose::FirehoseEndpoints;
 use graph::prelude::{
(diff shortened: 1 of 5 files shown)


otaviopace (approved):
LGTM :shipit: 

Also, since most of this part of the codebase is data structure parsing, perhaps we could have more tests around it in the future, or have some step on the Hosted Service deployment to assert the config files (or code) are correct.

maoueh:
Yeah, the hardest part is mocking the transport and such networking stuff to simulate condition where for example a provider is not answering. Willing to take a look.

PR #3130, opened 2022-01-06
Default to fetching receipts concurrently

tilacog:
- Inverts the logic in `ethereum_adapter` module to make fetching receipts concurrently the default behaviour.
- An environment variable `GRAPH_ETHEREUM_FETCH_TXN_RECEIPTS_IN_BATCHES` is provided for reverting to the previous, batching strategy.

### DIFF
--- a/chain/ethereum/src/ethereum_adapter.rs
+++ b/chain/ethereum/src/ethereum_adapter.rs
@@ -115,8 +115,10 @@ lazy_static! {
         .parse::<usize>()
         .expect("invalid GRAPH_ETHEREUM_BLOCK_INGESTOR_MAX_CONCURRENT_JSON_RPC_CALLS_FOR_TXN_RECEIPTS env var");
 
-    static ref FETCH_RECEIPTS_CONCURRENTLY: bool = std::env::var("GRAPH_EXPERIMENTAL_FETCH_TXN_RECEIPTS_CONCURRENTLY")
-            .is_ok();
+    static ref FETCH_RECEIPTS_IN_BATCHES: bool = std::env::var("GRAPH_ETHEREUM_FETCH_TXN_RECEIPTS_IN_BATCHES")
+        .map_or(false, |var| var.parse::<bool>().unwrap_or_default());
+
+
 
 
 }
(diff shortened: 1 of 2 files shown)


chain/ethereum/src/ethereum_adapter.rs
```diff
@@ -115,8 +115,10 @@ lazy_static! {
         .parse::<usize>()
         .expect("invalid GRAPH_ETHEREUM_BLOCK_INGESTOR_MAX_CONCURRENT_JSON_RPC_CALLS_FOR_TXN_RECEIPTS env var");
 
-    static ref FETCH_RECEIPTS_CONCURRENTLY: bool = std::env::var("GRAPH_EXPERIMENTAL_FETCH_TXN_RECEIPTS_CONCURRENTLY")
-            .is_ok();
+    static ref FETCH_RECEIPTS_IN_BATCHES: bool = std::env::var("GRAPH_ETHEREUM_FETCH_TXN_RECEIPTS_IN_BATCHES")
+        .map_or(false, |var| var.parse::<bool>().unwrap_or_default());
+
+
```
otaviopace, line 121: Any reason for the two new lines?

otaviopace (approved):

wup-one:
Hi. I really like that this setting exists, but is there any way to limit the number of messages in a batch when it's enabled?

I tried both `ETHEREUM_BLOCK_BATCH_SIZE` and `GRAPH_ETHEREUM_MAX_BLOCK_RANGE_SIZE` without luck.

When my ingestor falls behind too far by RPC becomes unhappy and returns:

`Code: -32005, Message: batch size exceed limit 500`

If there was a way to limit the batch sizes, that would be fantastic.

azf20:
hi @wup-one there isn't a way to do that right now - I think from an Ethereum node perspective, concurrent receipt fetching should be more performant and reliable than batch fetching as batched calls are executed one at a time

wup-one:
> hi @wup-one there isn't a way to do that right now - I think from an Ethereum node perspective, concurrent receipt fetching should be more performant and reliable than batch fetching as batched calls are executed one at a time

Thanks. I find the concurrent requests tend to kill RPCs and it's more prone to getting timeouts and disconnects. After running my own erigon node for a bit and noticing these issues, I ended up enabling batching which helped a lot. I noticed it helps a lot on public RPCs too.

azf20:
OK thanks, that is helpful feedback - @leoyvens @tilacog I don't think there's currently a way to optimise the batching setup?

PR #3131, opened 2022-01-06
chain/ethereum,graph,node: moved BlockIngestor to chain/ethereum and removed IngestorAdapter

maoueh:
The `IngestorAdapter` was actually protocol specific and having the abstraction in the Blockchain trait was not used expect for Ethereum chain.

Removed `IngestorAdapter` from block chain and moved the chain agnostic `BlockIngestor` to `graph-chain-ethereum` crate.

### DIFF
--- a/chain/ethereum/src/chain.rs
+++ b/chain/ethereum/src/chain.rs
@@ -14,17 +14,15 @@ use graph::{
         },
         firehose_block_stream::FirehoseBlockStream,
         polling_block_stream::PollingBlockStream,
-        Block, BlockHash, BlockPtr, Blockchain, ChainHeadUpdateListener,
-        IngestorAdapter as IngestorAdapterTrait, IngestorError, TriggerFilter as _,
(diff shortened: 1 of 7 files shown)


leoyvens (approved):
Considering that any new chain will use the firehose ingestor, we can assume this block ingestor will always be EVM-specific and therefore not worth maintaining abstractions for.

leoyvens:
I'd like @lutter's opinion before merging since he originally introduced this in https://github.com/graphprotocol/graph-node/pull/2411.

lutter:
I agree that a generic `IngestorAdaptor` has been overtaken by events, and making it ethereum specific is the right move.

How does this mesh with the ingestor we need for NEAR to have black hash/number mappings in the database though?

maoueh:
Right now, they are completely separated. There is a `BlockIngestor` (in `chain/ethereum`) specific for `NEAR` and a more generic for Firehose ready chain in general called `FirehoseBlockIngestor` (in `graph/blockchain`).

In main entrypoint of `graph-node`, the logic remain the same, just the location of `BlockIngestor` changed.

lutter (approved):

PR #3132, opened 2022-01-07
core: subgraph might continue processing to next block when an error occurred on a revert

maoueh:
The `continue` in the `instance_manager` run loop are actually doing a `continue` that continue to the next block in the stream.

That meant to be `break`s instead so they actually go up to the outer loop which restart the subgraph.

### DIFF
--- a/core/src/subgraph/instance_manager.rs
+++ b/core/src/subgraph/instance_manager.rs
@@ -545,7 +545,9 @@ where
                                     "block_hash" => format!("{}", subgraph_ptr.hash),
                                     "error" => e.to_string(),
                                 );
-                                continue;
+
+                                // Exit inner block stream consumption loop and go up to loop that restarts subgraph
+                                break;
                             }
                         }
                         None => {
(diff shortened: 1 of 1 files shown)


core/src/subgraph/instance_manager.rs
```diff
@@ -606,10 +610,12 @@ where
...
                     debug!(
                         &logger,
-                        "Block stream produced a non-fatal error";
+                        "Block stream produced a non-fatal error. Retrying";
                         "error" => format!("{}", e),
                     );
-                    continue;
+
+                    // Exit inner block stream consumption loop and go up to loop that restarts subgraph
+                    break;
```
leoyvens, line 618: This case I think should be a continue, the polling block stream at least is designed to be able to recover from its errors.

leoyvens (changes requested):

core/src/subgraph/instance_manager.rs
maoueh: You are right, `FirehoseBlockStream` will also reconnect forever in case of connection drop, I've reverted.

leoyvens (approved):

leoyvens:
@mangas This is a case that we missed in https://github.com/graphprotocol/graph-node/pull/3085, what would happen if one of these continues were hit is that the block stream would believe that the revert for block number `n` was processed, when actually failed to process. It would then either:
1. Emit a revert for `n - 1`, which implicitly reverts `n` so it's fine.
2. Emit a new block `n`, which could lead to inconsistencies but thankfully in `fn forward_block_ptr` we would catch the attempt to process the same block number twice.

PR #3133, opened 2022-01-07
core,graph,store: on block revert, Firehose should save the cursor to the database

maoueh:
This PR ensures that the Firehose cursor is saved properly even on revert operations.

### DIFF
--- a/store/postgres/src/deployment_store.rs
+++ b/store/postgres/src/deployment_store.rs
@@ -1,3 +1,4 @@
+use anyhow::Context;
 use detail::DeploymentDetail;
 use diesel::connection::SimpleConnection;
 use diesel::pg::PgConnection;
@@ -905,7 +906,7 @@ impl DeploymentStore {
 
             if let Some(cursor) = firehose_cursor {
                 if cursor != "" {
-                    deployment::update_firehose_cursor(&conn, &site.deployment, &cursor)?;
+                    deployment::update_firehose_cursor(&conn, &site.deployment, cursor)?;
                 }
             }
 
@@ -920,6 +921,7 @@ impl DeploymentStore {
         conn: &PgConnection,
         site: Arc<Site>,
         block_ptr_to: BlockPtr,
+        firehose_cursor: Option<&str>,
     ) -> Result<StoreEvent, StoreError> {
         let event = conn.transaction(|| -> Result<_, StoreError> {
             // Don't revert past a graft point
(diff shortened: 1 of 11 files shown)


leoyvens (changes requested):
For all I can tell, we can refactor `BlockStreamEvent::Revert` to always require a parent block ptr, rather than taking an Option.

maoueh:
Better than that, it was actually already fetched once in `PollingBlockStream` and passed, so my change would have break that.

The code in instance manager is not required anymore actually.

maoueh:
Done.

leoyvens (changes requested):
All DB operations to process a revert should be done in the same DB transaction, so we can rely on ACID guarantees. So the `update_block_cursor` operation should be done in the same transaction as the other operations.

maoueh:
@leoyvens You are right, makes total sense. I've updated the PR.

While doing the operations, while updating where `revert_block_operations` is called, I noticed to other places where right now Firehose is not handled:
- `graphman rewind` command
- `unfail_deterministic_error` does some revert of block operations and should handle the firehose cursor also. I would like to discuss this in a chat with you today.

leoyvens:
For the purpose of this PR, we need to at least understand if it's safe to call `fn revert_block_operations` with a `None` cursor, as is being done in those two places.

maoueh:
As discussed last week, for `rewind` and `unfail_determinisc_error` we are good with just actually removing the cursor completely.

If an upcoming PR, I'm going to make it possible for firehose to start from a `block_ptr` and ensure it's the correct canonical block we seek.

I think the PR is good to merge in it's current shape and next improvement will come in a subsequent PR.

store/postgres/src/deployment_store.rs
```diff
@@ -920,6 +921,7 @@ impl DeploymentStore {
         conn: &PgConnection,
         site: Arc<Site>,
         block_ptr_to: BlockPtr,
+        firehose_cursor: Option<&str>,
```
leoyvens, line 924: Should this still be an `Option`, given this is now using the empty string?

maoueh: Well, I used `""` at some place to "clear" the cursor, because right now `None` means do nothing (which is the case when using Ethereum RPC).

So I'm of the opinion that we should keep it like this.

leoyvens (approved):

PR #3160, opened 2022-01-14
add github workflow for firehose POI tests

sduchesneau:
* new eth-transfers subgraph in integration-tests (previously in github.com/streamingfast/graph-node-dev)
* required environment vars GRAPH_FIREHOSE_API_TOKEN and GRAPH_ETH_MAINNET_RPC_URL for github tests
* tests will run along other integration tests on every push (8min to build graphman, 3min to run 2 tests in parallel)


mangas:
personally I don't think this is the right place for this type of test but I don't know what was agreed with regards to it

maoueh:
Nothing agreed, that was more showcasing some possibilities. If we have something better brewing, we are all happy to close this one.

maoueh:
I think we should close this one, what do you think @sduchesneau ?

PR #3161, opened 2022-01-14
chain/ethereum,graph,node: start backfill BlockIngestor process for unprocessed blocks

colindickson:
Since the Firehose Block Ingestor starts at the current HEAD block of the chain, the purpose of this change is to spawn a process which will backfill the blocks from the start of the chain all the way to the block where the ingestor first started processing its blocks.

### DIFF
--- a/graph/src/blockchain/firehose_block_ingestor.rs
+++ b/graph/src/blockchain/firehose_block_ingestor.rs
@@ -212,7 +212,19 @@ where
                     continue;
                 }
                 Some(block_num) => {
-                    self.set_backfill_target_block_num(block_num).await;
+                    match self
+                        .chain_store
+                        .clone()
+                        .set_chain_backfill_target_block_num(block_num)
+                        .await
+                    {
+                        Ok(_) => {},
+                        Err(e) => {
+                            error!(self.logger, "Setting chain backfill target failed: {:?}", e);
+                            backoff.sleep_async().await;
+                        }
+                    }
+
                     return block_num;
                 }
             }
@@ -250,26 +262,6 @@ where
                         Some(t) => Some(t),
                     }
                 }
-                Err(e) => {
-                    error!(self.logger, "Setting chain backfill target failed: {:?}", e);
-                    backoff.sleep_async().await;
-                }
-            }
-        }
-    }
-
-    async fn set_backfill_target_block_num(&self, block_num: BlockNumber) {
-        let mut backoff =
-            ExponentialBackoff::new(Duration::from_millis(250), Duration::from_secs(30));
-
-        loop {
-            match self
-                .chain_store
-                .clone()
-                .set_chain_backfill_target_block_num(block_num)
-                .await
-            {
-                Ok(_) => return,
                 Err(e) => {
                     error!(self.logger, "Setting chain backfill target failed: {:?}", e);
                     backoff.sleep_async().await;
--- a/graph/src/blockchain/firehose_block_ingestor.rs
+++ b/graph/src/blockchain/firehose_block_ingestor.rs
@@ -89,7 +89,11 @@ where
             ExponentialBackoff::new(Duration::from_millis(250), Duration::from_secs(30));
 
         let mut backfill_cursor = self.fetch_backfill_cursor().await;
-        let backfill_target_block_number = self.fetch_backfill_target_block_num().await;
+
+        let mut backfill_target_block_number = self.fetch_backfill_target_block_num().await;
+        if backfill_target_block_number.is_none() {
+                backfill_target_block_number =  Some(self.initialize_backfill_target_block_num().await);
+        }
 
         loop {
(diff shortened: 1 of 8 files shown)


graph/src/blockchain/firehose_block_ingestor.rs
```diff
@@ -82,6 +82,87 @@ where
...
+        let mut backfill_cursor = self.fetch_backfill_cursor().await;
+
+        loop {
+            let backfill_completed = self.fetch_backfill_is_completed().await;
+            if backfill_completed {
+                info!(self.logger, "Backfill completed");
+                break;
+            }
+
+            let backfill_target_block_number = self.fetch_backfill_target_block_num().await;
```
maoueh, line 100: This should be done outside the main loop, there is no need to fetch the backfill target block num on every loop, only once.

graph/src/blockchain/firehose_block_ingestor.rs
```diff
@@ -82,6 +82,87 @@ where
...
+            // If we reach this point, we must wait a bit before retrying
+            backoff.sleep_async().await;
+        }
+    }
+
+    async fn fetch_backfill_is_completed(&self) -> bool {
+        let mut backoff =
+            ExponentialBackoff::new(Duration::from_millis(250), Duration::from_secs(30));
+        loop {
+            match self.chain_store.clone().chain_backfill_is_completed() {
```
maoueh, line 140: I feel `chain_backfill_is_completed` should return `Result<bool, Error>`. What the case where it can be `Ok(None)` exactly? Wouldn't it better to return in `Error` if backfill completeness can't be determined somehow?

graph/src/blockchain/firehose_block_ingestor.rs
maoueh, line 100: I would like to have `fetch_backfill_target_block_num` split in two concerns, retrieve the target backfill block num alone and determine/set the backfill target block num if not set. Pseudo-code here would look like:
```suggestion
            let mut backfill_target_block_number = self.fetch_backfill_target_block_num().await;
            if backfill_target_block_number.is_none() {
                backfill_target_block_number =  Some(self.initialize_backfill_target_block_num().await);
            }
```

graph/src/blockchain/firehose_block_ingestor.rs
```diff
@@ -97,6 +178,99 @@ where
...
+                }
+                Err(e) => {
+                    error!(self.logger, "Setting chain backfill target failed: {:?}", e);
+                    backoff.sleep_async().await;
+                }
+            }
+        }
+    }
+
+    async fn set_backfill_target_block_num(&self, block_num: BlockNumber) {
```
maoueh, line 254: When splitting `fetch_backfill_target_block_num`, I don't think this will be necessary and could be put straight in `initialize_backfill_target_block_num` which would make the retry portion clearer and located in the `initialize_backfill_target_block_num` directly.

graph/src/blockchain/firehose_block_ingestor.rs
```diff
@@ -159,10 +333,87 @@ where
 
         self.chain_store
             .clone()
-            .set_chain_head(block, response.cursor.clone())
+            .set_chain_head(block.clone(), response.cursor.clone())
```
maoueh, line 336: Why is this required now? Isn't this in untouched portion of your PR?

graph/src/blockchain/firehose_block_ingestor.rs
```diff
@@ -159,10 +333,87 @@ where
...
+            .set_chain_backfill_cursor(response.cursor.clone())
+            .await
+            .context("Updating chain backfill cursor")?;
+
+        if block_number >= backfill_target_block_number.into() {
+            self.chain_store
+                .clone()
+                .set_chain_backfill_as_completed()
+                .await
+                .context("setting backfill completion")?;
```
maoueh, line 414:
```suggestion
                .context("Setting backfill completion")?;
```

maoueh (changes requested):

graph/src/blockchain/firehose_block_ingestor.rs
colindickson: done!

graph/src/blockchain/firehose_block_ingestor.rs
colindickson: done!

graph/src/blockchain/firehose_block_ingestor.rs
colindickson: done!

neysofu:
@maoueh should I close this?

maoueh:
I think it might not be important anymore. It was a way to backfill missing block from `FirehoseBlockIngestor` that start at "live" point of the chain.

Nobody ever reported any problem around such missing feature, so I imagine it's not needed.

lutter:

PR #3164, opened 2022-01-17
graphql: improvements for GraphQL operations validations

dotansimha:
### Changes in this PR 

- 100% coverage of the GraphQL spec validations
- Fixed a bug with `KnownDirectives` by adding `@skip` and `@include` definitions to the `meta.graphql` schema.
- New, improved version of `graphql-tools` that iterates nodes and validates faster. 
- Improved version of `OverlappingFieldsCanBeMerged` rule (covers all edge-cases now)
- New rule `ValuesOfCorrectType`: validates the literals of a query. 
- New rule `UniqueDirectivesPerLocation` for enforcing the uniqueness of directives.  
- New rule `VariablesInAllowedPosition`: validates that a variable matches the arguments that used it.
- New rule `KnownDirectives` to validate that directives used by the query are defined in the schema and on the correct location.  
- Update `SingleFieldSubscriptions` to the latest GraphQL spec (field `__typename` is no longer allowed to be queried on a subscription root) 

image

image

image

image


core/tests/interfaces.rs
```diff
@@ -601,9 +601,11 @@ async fn invalid_fragment() {
...
 
     match &res.to_result().unwrap_err()[0] {
-        QueryError::ExecutionError(QueryExecutionError::UnknownField(_, type_name, field_name)) => {
-            assert_eq!(type_name, "Legged");
-            assert_eq!(field_name, "name");
+        QueryError::ExecutionError(QueryExecutionError::ValidationError(_, error_message)) => {
+            assert_eq!(
+                error_message,
+                "Cannot query field \"name\" on type \"Legged\"."
+            );
```
lutter, line 608: IIRC, I addressed that on the `lutter/gql` branch by accepting either error message

lutter (approved):

core/tests/interfaces.rs
dotansimha: Cool, do you want me to drop this change? (because then this branch fails to pass unit tests...) or should i copy it from your branch?

dotansimha:
@lutter recent failure `thread 'cannot_filter_by_derved_relationship_fields' panicked at 'expected ResolveEntitiesError, got ExecutionError(InvalidArgumentError(Pos(0:0), "where", Object({"writtenSongs": List([String("s1")])})))', graphql/tests/query.rs:1294:18` seems to be unrelated? not sure...

core/tests/interfaces.rs
lutter: For now, I think it would be preferable if tests pass both when `ENABLE_GRAPHQL_VALIDATIONS` is set in the environment and when it is not.

lutter (approved):
Looks good. Other than making sure test pass with and without validations turned on, I think this is ready to go. I'll approve, but please make sure tests can be run in both modes before merging

dotansimha:
@lutter done, ran all unit tests with and without validations, pushed a few fixes, and now it works :)

PR #3166, opened 2022-01-17
Added an environment to control if Firehoe is preferred or not when present

maoueh:
If the environment is **not** set, the current behavior and Firehose is automatically picked up if there is endpoints configured with it.

If the environment set, firehose is preferred only if the variable is true and there is endpoints configured with it.

### DIFF
--- a/chain/ethereum/src/chain.rs
+++ b/chain/ethereum/src/chain.rs
@@ -2,6 +2,7 @@ use anyhow::{Context, Error};
 use graph::blockchain::BlockchainKind;
 use graph::components::store::WritableStore;
 use graph::data::subgraph::UnifiedMappingApiVersion;
+use graph::env::env_var;
 use graph::firehose::{FirehoseEndpoints, ForkStep};
 use graph::prelude::{
     EthereumBlock, EthereumCallCache, LightEthereumBlock, LightEthereumBlockExt, StopwatchMetrics,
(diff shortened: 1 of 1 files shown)


chain/ethereum/src/chain.rs
```diff
@@ -58,6 +58,11 @@ lazy_static! {
         .unwrap_or("100".into())
         .parse::<u64>()
         .expect("invalid GRAPH_ETHEREUM_TARGET_TRIGGERS_PER_BLOCK_RANGE");
+
+    /// Controls if firehose should be preferred over RPC if Firehose endpoints are present, if not set, the default behavior is
+    /// is kept which is to automatically favor Firehose.
+    static ref IS_FIREHOSE_PREFERRED: Option<bool> = std::env::var("GRAPH_ETHEREUM_IS_FIREHOSE_PREFERRED")
+        .map_or_else(|_| None, |input| Some(input.parse::<bool>().expect("invalid GRAPH_ETHEREUM_IS_FIREHOSE_PREFERRED")));
```
otaviopace, line 65: You can use `.ok().map()` here since you're not using the first closure passed to `map_or_else`.

otaviopace (approved):

leoyvens:
You can also use this helper
https://github.com/graphprotocol/graph-node/blob/8155ed5299e96a79a1554642486bcf88254b4a59/graph/src/env.rs#L43

I find it confusing that this is an `Option<bool>`, if the default is to prefer firehose then it should just default to `true`.

chain/ethereum/src/chain.rs
maoueh: Thanks, was indeed simpler with that (although it's been removed now :))

maoueh:
@leoyvens You are right, it's simpler without, it's was overcomplicated for the same net result, changed.

PR #3170, opened 2022-01-18
Graphman new command: `analyze`

### LINKED ISSUE #3152: graphman command to run analyze entity tables

tilacog:
Introduces a new Graphman `analyze` command that calls SQL `ANALYZE` for the given entity table.

Closes #3152


node/src/manager/commands/analyze.rs
```diff
@@ -0,0 +1,22 @@
...
+    prelude::{anyhow::anyhow, DeploymentHash, Error},
+};
+use graph_store_postgres::SubgraphStore;
+
+pub async fn analyze(
+    store: Arc<SubgraphStore>,
+    hash: String,
+    entity_name: String,
+) -> Result<(), Error> {
+    println!("Running ANALYZE for {entity_name} entity");
```
otaviopace, line 14: Dumb question, do we have a proper logger for `graphman`?

node/src/bin/manager.rs
```diff
@@ -179,6 +179,13 @@ pub enum Command {
     Chain(ChainCommand),
     /// Manipulate internal subgraph statistics
     Stats(StatsCommand),
+    /// Perform a SQL ANALYZE in a Entity table
+    Analyze {
```
otaviopace, line 183: Perhaps the command could accept a `verbose` argument?
I would even argue we want verbose as default.

node/src/manager/commands/analyze.rs
```diff
@@ -0,0 +1,22 @@
...
+    let entity_type = EntityType::new(entity_name);
+    let deployment_hash =
+        DeploymentHash::new(hash).expect("Subgraph hash must be a valid IPFS hash");
```
otaviopace, line 17: Could `.context()?` from `anyhow` provide better error info for `graphman` users? 🤔

store/postgres/src/deployment_store.rs
```diff
@@ -674,6 +674,24 @@ impl DeploymentStore {
...
+        &self,
+        site: Arc<Site>,
+        entity_type: EntityType,
+    ) -> Result<(), StoreError> {
+        let store = self.clone();
+        self.with_conn(move |conn, _| {
+            let layout = store.layout(conn, site)?;
+            let table = layout.table_for_entity(&entity_type)?;
+            let table_name = &table.qualified_name;
+            let sql = format!("analyze {table_name}");
```
otaviopace, line 689: It would be nice if `diesel` provided `analyze` in their [pg](https://docs.rs/diesel/1.4.8/diesel/pg/index.html) module, perhaps we can add there someday 🙂

node/src/manager/commands/analyze.rs
lutter: Yes, logging is hooked up for `graphman` and turned on by setting `GRAPH_LOG=debug` etc. But it's very annoying on the console.

Printing is the right way to produce information the user should see.

node/src/bin/manager.rs
lutter: I agree .. it might be a little tricky to get the command output out of diesel, but if it's not too much work, would be nice to just dump whatever Postgres says onto the tty.

lutter: I think this command should be `graphman stats analyze Qm.. Entity`, i.e., the `analyze` should be a subcommand of `stats`

lutter (approved):
This all looks good!

lutter:
As follow-on work (i.e., in a separate PR since I think having the command in its current form in use is more important than adding these bells and whistles) these things would be nice:

* allow specifying the deployment also with `sgdNNN` (have a look at how the `info` command resolves names and hashes, I think you can just reuse that code)
* allow specifying either the entity name or the table name on the command line

with that, `graphman stats analyze sgd321 pair_hour_data` should also be legal. That form is often more useful when you see a specific query take a long time.

And while we're at it, it would also be neat if the command could print when the table was last analyzed before running analyze. You can get that information with `select greatest(last_analyze, last_autoanalyze) from pg_stat_user_tables where relname = 'transaction' and schemaname = 'sgd52520';`

node/src/bin/manager.rs
tilacog: I couldn't find a way to capture PostgreSQL output from diesel. Not sure if this is supported at all.

PR #3172, opened 2022-01-19
Update comment on incremental compilation

leoyvens:
This doesn't seem to reproduce in 1.58

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ members = [
     "tests",
 ]
 
-# Incremental compilation on Rust 1.55 causes an ICE on build. As soon as graph node builds again, these can be removed.
+# Incremental compilation on Rust 1.58 causes an ICE on build. As soon as graph node builds again, these can be removed.
 [profile.dev]
 incremental = false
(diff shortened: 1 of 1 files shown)


maoueh:
Should we introduce the [rust-toolchain.toml](https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file) so that developers always pick the right Rust toolchain?

leoyvens:
Interesting, didn't know that was a thing. Generally we support building on the latest stable, I don't think we need to require something more specific in a toolchain file.

maoueh:
But if incremental is now active, devs not upgrading to 1.58 will likely face ICE on incremental build, hence why I suggested we introduced the toolchain file.

lutter (commented):
I am still getting an ICE with 1.58.0 on the `lutter/gql` branch. What I did was a clean build, then `touch graphql/src/execution/query.rs && cargo build`. So we can't reenable that just yet.

leoyvens:
@lutter Thanks for checking! I've changed this to just update the comment.

Issue #101, opened 2018-06-18
Export database API to Wasm runtime


fordN:
Use the Wasmi crate to expose database functions to the Wasm runtime

Issue #100, opened 2018-06-18
Initialize Wasm Module Instance in Rust


fordN:
Initialize the Wasm module using Parity Wasm and Wasmi.

PR #3174, opened 2022-01-19
chain,graph,store: added code to properly resume from a subgraph block ptr and no firehose cursor

maoueh:
Firehose is now using the subgraph ptr if no cursor is found. To ensure we do not create inconsistency in the data due to how Firehose resolves current block number, a sanity check has now been added. This check ensure that the first received block's parent from firehose is the same as the subgraph pointer block. If it's not the case, we revert to the last know final block so we are sure to start from a correct chain segment.

What final block to pick is decided by the chain itself. On Ethereum, we use `block.number - reorg_threshold` (guarded) while on NEAR, we use the last final block height defined by the block itself.

#### Missing/To Change

I pushed the PR to get an overall feeling of the idea and get early inputs.

Right now I'm using `revert_block_operations` but this is wrong as it's able to only revert a single block. I need in fact probably to `rewind` the subgraph but I would like to get advise here on how to do that properly and if there is any risks or stuff I missed with the approach.

Ping @lutter to get your input about the `rewind` thing above.

#### Draft

This is based on https://github.com/graphprotocol/graph-node/pull/3133 for now, so I would like to get the other merged first and rebased + finish this one after.


chain/ethereum/src/chain.rs
```diff
@@ -311,6 +313,20 @@ impl Blockchain for Chain {
             .await
     }
 
+    async fn final_block_pointer_for(
+        &self,
+        logger: &Logger,
+        block: &BlockFinality,
+    ) -> Result<BlockPtr, IngestorError> {
+        let final_block_number = match block.number() {
+            x if x >= self.reorg_threshold => x - self.reorg_threshold,
```
leoyvens, line 322: I haven't wrapped around the whole PR yet. But generally one of the advantages of the firehose is to not have to configure a reorg threshold at all. Maybe this could just revert one block a time until it gets a consistent response from the firehose?

maoueh: So all this PR is around the fact that we want to start a subgraph from a particular block_ptr. So the subgraph is at block `#100 abc`. 

Now, we ask Firehose to start at block `#101` but we cannot specify the block hash we want to start from. So the rough idea is the following:

- Start Firehose with block `#101`
- Retrieve the first block sent by Firehose
- Get the parent from it
- Ensure the `parent` is == `#100 abc`
- If it's not the case, rewind the subgraph to the last final block of the block `#100`

This is of course an edge case that will not happen often. 

I like your idea to be honest, it's probably easier to implement than doing the full rewind multiple blocks behind.

maoueh: Well, now that I think about it, ideally I would revert to the parent block of the current subgraph block ptr, but I don't think I have such info, is it available? 

Let's discuss this at the meeting in ~1h.

maoueh: I've moved all the logic in the `FirehoseBlockStream` now. I still require getting the last final block where to revert to. But it's good.

I still need to update `revert_block_...` to actually accept reverting more than one block, and I need to also test the revertion code.

maoueh:
This has been updated with the newest discussion we had, and some other improvements:
- `revert_block_operations` can now revert more than 1 block at a time (but still asserts that reverts goes backward)
- The chain continuity is now checked inside `FirehoseBlockStream` directly
- Final block from Firehose is actually is now faster and will actually be able to return a response much faster by levering the `stop_block_num`
- Cursor is now automatically deleted when starting an RPC subgraph (so that the case Firehose -> RPC -> Firehose works properly)
- NEAR has now a working `block_ptr_for_number`.

This was tested and worked (still need to validate the POI of a subgraph synced straight with RPC and when switching from RPC to Firehose and the like):
- RPC -> Firehose
- RPC -> Firehose -> RPC
- Firehose -> RPC

Ready for review and merge.

maoueh:
I have fully tested that resuming from RPC subgraph block then falling back to Firehose on a forked block produces the same POI for a few blocks after the forked block in question.

I've fixed also an edge case where I was allowing to revert back behind manifest's defined start block.

This is ready for a final review and merge.

chain/ethereum/src/ethereum_adapter.rs
```diff
@@ -1785,7 +1785,21 @@ async fn fetch_transaction_receipts_in_batch_with_retry(
             fetch_transaction_receipts_in_batch(web3, hashes, block_hash, logger).boxed()
         })
         .await
-        .map_err(|_timeout| anyhow!(block_hash).into())
+        .map_err(|timeout| {
```
maoueh, line 1788: Oups leftover from understanding an error I had. Will open a dedicated PR for those.

store/postgres/src/deployment_store.rs
```diff
@@ -1015,9 +1024,9 @@ impl DeploymentStore {
         // Unwrap: If we are reverting then the block ptr is not `None`.
         let block_ptr_from = Self::block_ptr_with_conn(&site.deployment, &conn)?.unwrap();
 
-        // Sanity check on block numbers
-        if block_ptr_from.number != block_ptr_to.number + 1 {
-            panic!("revert_block_operations must revert a single block only");
+        // Sanity check on revert to ensure we go backward only
+        if block_ptr_to.number >= block_ptr_from.number {
```
leoyvens, line 1028: I do find the assertion that a single block is reverted reassuring. Perhaps we could have two fns that share the implementation, but have different asserts for different uses?

core/src/subgraph/instance_manager.rs
```diff
@@ -806,6 +799,24 @@ where
     }
 }
 
+async fn delete_subgraph_firehose_cursor(logger: &Logger, store: &dyn WritableStore) {
+    debug!(logger, "Deleting any existing Firehose cursor");
+    let mut backoff = ExponentialBackoff::new(30 * SECOND, *SUBGRAPH_ERROR_RETRY_CEIL_SECS);
+
+    loop {
+        match store.delete_block_cursor() {
```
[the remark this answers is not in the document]
maoueh: This is a copy of other existing methods in the trait like `block_ptr`, `block_cursor` and many others. Is the new paradigm to always use `async` from now on?

I would like to perform that in another PR mainly because this is blocking the integration test workflow and it would be nice to unlock it.

core/src/subgraph/instance_manager.rs
```diff
@@ -806,6 +799,24 @@ where
     }
 }
 
+async fn delete_subgraph_firehose_cursor(logger: &Logger, store: &dyn WritableStore) {
+    debug!(logger, "Deleting any existing Firehose cursor");
+    let mut backoff = ExponentialBackoff::new(30 * SECOND, *SUBGRAPH_ERROR_RETRY_CEIL_SECS);
```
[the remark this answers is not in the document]
maoueh: I agree, in the `FirehoseBlockIngestor` there is a bunch of those also so it would be a good thing to have a re-usable method. Again, would like to postpone that to another PR.

store/postgres/src/deployment.rs
```diff
@@ -274,6 +274,19 @@ pub fn forward_block_ptr(
...
+    use subgraph_deployment as d;
+
+    update(d::table.filter(d::deployment.eq(id.as_str())))
+        .set(d::firehose_cursor.eq::<Option<&str>>(None))
+        .execute(conn)
+        .map(|_| ())
+        .map_err(|e| e.into())
+}
+
 pub fn get_subgraph_firehose_cursor(
```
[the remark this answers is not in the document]
maoueh: This is not part of this PR, but I would include that in an upcoming PR that would asyncinfy the various methods found in the `WritableStore` trait.

store/postgres/src/deployment_store.rs
leoyvens: Considering that Near skips blocks this is necessary, so this can be kept as is.

leoyvens (approved):

sduchesneau:
@otaviopace I just rebased, then double-checked this PR and the behavior is as-expected. Would you kindly go ahead and merge this into master ?

PR #3175, opened 2022-01-20
Graphman new command: `index create`

### LINKED ISSUE #3151: graphman command for btree index creation
Inputs something like this:
- Deployment
- Entity / table
- Attributes to include in the index

Result:
- A new btree index for the inputs

Important:
- Create the index **concurrently**
- Prefix index name with `manual_<table-name>_<columns_divided_by_underscores>`
- After completing, check whether the index is _valid_ (because index creation can fail)
  - If it's not, drop it and ask the user to try again, exit

tilacog:
Adds a new subcommand tree `index / create` for `graphman`.

Closes #3151


lutter (approved):
Nice!

PR #3176, opened 2022-01-20
Introduce metrics.md documentation

pwilczynskiclearcode:


lutter (approved):
Thanks for writing this! We've been needing something like this for a while.

PR #3184, opened 2022-01-25
Filter by child entity

kamilkisiela:
#2960

Given the schema:

```graphql
type Purpose @entity {
  id: ID!
  sender: Sender!
  purpose: String!
  createdAt: BigInt!
  transactionHash: String!
}

type Sender @entity {
  id: ID!
  address: Bytes!
  purposes: [Purpose!] @derivedFrom(field: "sender")
  createdAt: BigInt!
  purposeCount: BigInt!
}
```

The `sender_` is added to `Purpose_filter` input object.

```diff
input Purpose_filter {
    ...
+  sender_: Sender_filter
}
```

The `Sender_filter` input object is untouched.

## Questions

**How should I handle `@derivedFrom` directive?**
I see in code that the `where` argument is not generated for fields with `@derivedFrom`.

**Is it find that I hardcoded `c` for root table (the one we select columns from)?**
I noticed that aliases/prefixes for tables and columns are hard-coded so I guess it's fine to assume that `c` represents the root tabel.

**Filters can be deeply nested, is it a problem?**
The logic behind generating the joins is fairly easy and we won't end up with duplicates. It's table <-> table and it starts from "root" table (the one that we select from). 

Every table gets its own prefix/alias, the root table gets `c`. In case of a few level deep filter that adds a condition for `Foo` entity, it ends up being prefixed in the same way as the condition for `Foo` in the first or n-th level.

```sql
WHERE (c1.createdAt > 19283721 AND ( .... ( ... ( c1.purposeCount > 5 OR ... ) ) ) )
```

**Should a left join contain `vid = vid` condition?**
I used `left join` to join tables with `parent.id = child.column AND parent.vid = child.vid` condition. Is it fine? I guess so, because `vid` represents the version and they should match.

Continues #3022

### DIFF
--- a/graphql/src/schema/api.rs
+++ b/graphql/src/schema/api.rs
@@ -376,11 +376,15 @@ fn field_filter_input_values(
                         // `where: { others: ["some-id", "other-id"] }`. In both cases,
                         // we allow ID strings as the values to be passed to these
                         // filters.
-                        field_scalar_filter_input_values(
+                        let mut input_values = field_scalar_filter_input_values(
                             schema,
                             field,
                             &ScalarType::new(String::from("String")),
-                        )
+                        );
+
+                        extend_with_child_filter_input_value(field, name, &mut input_values);
+
+                        input_values
                     }
                 }
                 TypeDefinition::Scalar(ref t) => field_scalar_filter_input_values(schema, field, t),
@@ -438,6 +442,19 @@ fn field_scalar_filter_input_values(
     .collect()
 }
 
+/// Appends a child filter to input values
+fn extend_with_child_filter_input_value(
+    field: &Field,
+    field_type_name: &String,
+    input_values: &mut Vec<InputValue>,
+) {
+    input_values.push(input_value(
+        &format!("{}_", field.name),
+        "",
+        Type::NamedType(format!("{}_filter", field_type_name)),
+    ));
+}
+
 /// Generates `*_filter` input values for the given enum field.
 fn field_enum_filter_input_values(
     _schema: &Document,
@@ -470,33 +487,44 @@ fn field_list_filter_input_values(
         // Decide what type of values can be passed to the filter. In the case
         // one-to-many or many-to-many object or interface fields that are not
         // derived, we allow ID strings to be passed on.
-        let input_field_type = match typedef {
-            TypeDefinition::Interface(_) | TypeDefinition::Object(_) => {
+        let (input_field_type, parent_type_name) = match typedef {
+            TypeDefinition::Object(parent) => {
                 if ast::get_derived_from_directive(field).is_some() {
                     return None;
                 } else {
-                    Type::NamedType("String".into())
+                    (Type::NamedType("String".into()), Some(parent.name.clone()))
                 }
             }
-            TypeDefinition::Scalar(ref t) => Type::NamedType(t.name.to_owned()),
-            TypeDefinition::Enum(ref t) => Type::NamedType(t.name.to_owned()),
+            TypeDefinition::Interface(parent) => {
+                if ast::get_derived_from_directive(field).is_some() {
+                    return None;
+                } else {
+                    (Type::NamedType("String".into()), Some(parent.name.clone()))
+                }
+            }
+            TypeDefinition::Scalar(ref t) => (Type::NamedType(t.name.to_owned()), None),
+            TypeDefinition::Enum(ref t) => (Type::NamedType(t.name.to_owned()), None),
             TypeDefinition::InputObject(_) | TypeDefinition::Union(_) => return None,
         };
 
-        Some(
-            vec!["", "not", "contains", "not_contains"]
-                .into_iter()
-                .map(|filter_type| {
-                    input_value(
-                        &field.name,
-                        filter_type,
-                        Type::ListType(Box::new(Type::NonNullType(Box::new(
-                            input_field_type.clone(),
-                        )))),
-                    )
-                })
-                .collect(),
-        )
+        let mut input_values: Vec<InputValue> = vec!["", "not", "contains", "not_contains"]
+            .into_iter()
+            .map(|filter_type| {
+                input_value(
+                    &field.name,
+                    filter_type,
+                    Type::ListType(Box::new(Type::NonNullType(Box::new(
+                        input_field_type.clone(),
+                    )))),
+                )
+            })
+            .collect();
+
+        if let Some(parent) = parent_type_name {
+            extend_with_child_filter_input_value(field, &parent, &mut input_values);
+        }
+
+        Some(input_values)
     })
 }
 
@@ -1020,6 +1048,7 @@ mod tests {
                 "pets_not",
                 "pets_contains",
                 "pets_not_contains",
+                "pets_",
                 "favoriteFurType",
                 "favoriteFurType_not",
                 "favoriteFurType_in",
@@ -1038,6 +1067,7 @@ mod tests {
                 "favoritePet_not_starts_with",
                 "favoritePet_ends_with",
                 "favoritePet_not_ends_with",
+                "favoritePet_",
             ]
             .iter()
             .map(ToString::to_string)
@@ -1248,5 +1278,57 @@ type Gravatar @entity {
             _ => None,
         }
         .expect("\"metadata\" field is missing on Query type");
+    }
+
+    #[test]
+    fn api_schema_contains_child_entity_filter_on_lists() {
+        let input_schema = parse_schema(
+            r#"
+                type Note @entity {
+                    id: ID!
+                    text: String!
+                    author: User! @derived(field: "id")
+                }
+              
+                type User @entity {
+                    id: ID!
+                    dateOfBirth: String
+                    country: String
+                    notes: [Note!]!
+                }
+            "#,
+        )
+        .expect("Failed to parse input schema");
+        let schema = api_schema(&input_schema).expect("Failed to derived API schema");
+
+        println!("{}", schema);
+
+        let user_filter = schema
+            .get_named_type("User_filter")
+            .expect("User_filter type is missing in derived API schema");
+
+        let filter_type = match user_filter {
+            TypeDefinition::InputObject(t) => Some(t),
+            _ => None,
+        }
+        .expect("User_filter type is not an input object");
+
+        let user_notes_filter_field = filter_type
+            .fields
+            .iter()
+            .find_map(|field| {
+                if field.name == "notes_" {
+                    Some(field)
+                } else {
+                    None
+                }
+            })
+            .expect("notes_ field is missing in the User_filter input object");
+
+        assert_eq!(user_notes_filter_field.name, "notes_");
+        assert_eq!(
+            user_notes_filter_field.value_type,
+            Type::NamedType(String::from("Note_filter"))
+        );
     }
 }
--- a/graphql/src/store/query.rs
+++ b/graphql/src/store/query.rs
@@ -2,6 +2,8 @@ use std::collections::{BTreeMap, BTreeSet, HashSet, VecDeque};
 use std::mem::discriminant;
 
 use graph::data::value::Object;
+use graph::data::value::Value as DataValue;
+use graph::prelude::s::Type;
 use graph::prelude::*;
 use graph::{components::store::EntityType, data::graphql::ObjectOrInterface};
 
@@ -27,6 +29,7 @@ pub(crate) fn build_query<'a>(
     max_first: u32,
     max_skip: u32,
     mut column_names: SelectedAttributes,
+    schema: &ApiSchema,
 ) -> Result<EntityQuery, QueryExecutionError> {
     let entity = entity.into();
     let entity_types = EntityCollection::All(match &entity {
@@ -45,7 +48,7 @@ pub(crate) fn build_query<'a>(
     });
     let mut query = EntityQuery::new(parse_subgraph_id(entity)?, block, entity_types)
         .range(build_range(field, max_first, max_skip)?);
-    if let Some(filter) = build_filter(entity, field)? {
+    if let Some(filter) = build_filter(entity, field, schema)? {
         query = query.filter(filter);
     }
     let order = match (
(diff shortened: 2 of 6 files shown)


lutter:
Could you rebase this to latest master when you get a chance?

dotansimha:
@lutter done! :) 

Note that this PR is only for filtering, for sorting we have this one: https://github.com/graphprotocol/graph-node/pull/3096 (targeted into this PR, not `master`, and @kamilkisiela had some questions there before proceeding).

store/postgres/src/relational_queries.rs
```diff
@@ -803,6 +803,165 @@ impl<'a> QueryFragment<Pg> for PrefixComparison<'a> {
...
+            _ => {}
+        }
+        Ok(())
+    }
+
+    fn find(&self, parent_table: &'a Table, child_table: &'a Table) -> Option<&TableJoin> {
+        self.joins
+            .iter()
+            .filter_map(|join| {
+                if join.parent.name == parent_table.name && join.child.name == child_table.name {
```
lutter, line 862: This doesn't seem right - it's possible that the same parent and child tables appear in one join, e.g., if you have a schema like
```
type User {
  id ID!
}

type Account {
  id ID!
  owner User!
  sponsor User!
}
```

If you now filter accounts on both `owner` and `sponsor`, you'll have to join the account table with the user table twice for different purposes, and probably with different filters.

store/postgres/src/relational_queries.rs
```diff
@@ -803,6 +803,165 @@ impl<'a> QueryFragment<Pg> for PrefixComparison<'a> {
...
+            out.push_sql(format!("{}.", prefix).as_str());
+            out.push_identifier(column)?;
+
+            Ok(())
+        }
+
+        for table_join in &self.joins {
+            out.push_sql("\n");
+
+            out.push_sql("left join ");
```
lutter, line 932: you do want an `inner join` here; `left join` is shorthand for `left outer join`, and we want to suppress all parent rows that have no matching children.

store/postgres/src/relational_queries.rs
```diff
@@ -803,6 +803,165 @@ impl<'a> QueryFragment<Pg> for PrefixComparison<'a> {
...
+                table_join.parent_column.name.as_str(),
+                &table_join.parent_prefix,
+            )?;
+
+            out.push_sql(" AND ");
+
+            // Match versions
+            push_column_with_prefix(out.reborrow(), "vid", &table_join.child_prefix)?;
+            out.push_sql(" = ");
+            push_column_with_prefix(out.reborrow(), "vid", &table_join.parent_prefix)?;
```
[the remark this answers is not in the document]
kamilkisiela: I switched from vid to block

store/postgres/src/relational_queries.rs
kamilkisiela: It has now different, simpler implementation

store/postgres/src/relational_queries.rs
kamilkisiela: Thanks, used `inner join`.

store/postgres/src/relational_queries.rs
```diff
@@ -867,24 +869,126 @@ impl<'a> fmt::Display for QueryFilter<'a> {
...
+        child_filter_ancestor: bool,
+    ) -> Result<(), StoreError> {
+        use EntityFilter::*;
+        match filter {
+            And(filters) | Or(filters) => {
+                for filter in filters {
+                    Self::valid_attributes(filter, table, layout, child_filter_ancestor)?;
+                }
+            }
+            Child(_, entity, child_filter) => {
```
lutter, line 976: This check should also make sure that the attribute name is valid for `table`, i.e., call `table.column_for_field(attr)?`

lutter (commented):
Getting closer! I actually hadn't thought about the issue that forced you to use `group by` but I think using the `exists(select 1 ..)` strategy I outlined in the comment will make this easier.

We'll also need tests for all this; the best place for them is probably `graphql/tests/query.rs` - I recently revamped them to (hopefully) make it easier to add tests there. The tests for this should use child filters that traverse `Musician.bands`, `Musician.written_songs`, `Band.members`, `Song.written_by` and `SongStat.song`. For example, what I have in mind for `Musician.bands` is something like `musicians(where: { bands_: { name: "something" } }) { id }`

store/postgres/src/relational_queries.rs
```diff
@@ -953,9 +985,73 @@ impl<'a> QueryFilter<'a> {
...
+        out.push_sql(parent_column.name.as_str());
+
+        out.push_sql(" and ");
+
+        // TODO: we need a $block here
+        // // Match by block
+        // out.push_sql(inner_prefix);
+        // out.push_identifier(BLOCK_RANGE_COLUMN)?;
+        // out.push_sql(" @> ");
+        // out.push_bind_param::<Integer, _>(&block_number)?;
```
kamilkisiela, line 1036: @lutter I'm not sure what would be the best way to pass the block number. I could drop the implementation for the `QueryFragment` and extend `walk_ast` to accept `block`.

store/postgres/src/relational_queries.rs
```diff
@@ -2706,7 +2859,8 @@ impl<'a> FilterQuery<'a> {
         write_column_names(column_names, table, &mut out)?;
         self.filtered_rows(table, filter, out.reborrow())?;
         out.push_sql("\n ");
-        self.sort_key.order_by(&mut out)?;
+        out.push_sql(" GROUP BY c.id, c.vid "); // TODO: ask if it's fine to group by id and vid, always
```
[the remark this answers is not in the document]
kamilkisiela: I think right now it skips generation of a child filter input for derived fields

graphql/tests/query.rs
```diff
@@ -174,6 +185,34 @@ fn test_schema(id: DeploymentHash, id_type: IdType) -> Schema {
...
+        band: Band
+        author: User!
+    }
+
+    type User @entity {
+        id: ID!
+        name: String!
+        bandReviews: [BandReview] @derivedFrom(field: \"author\")
+        songReviews: [SongReview] @derivedFrom(field: \"author\")
+        reviews: [Review]
```
kamilkisiela, line 214: @lutter is it correct usage of an interface?

graphql/src/schema/api.rs
```diff
@@ -940,14 +990,287 @@ mod tests {
...
+                "body_not_contains_nocase",
+                "body_starts_with",
+                "body_starts_with_nocase",
+                "body_not_starts_with",
+                "body_not_starts_with_nocase",
+                "body_ends_with",
+                "body_ends_with_nocase",
+                "body_not_ends_with",
+                "body_not_ends_with_nocase",
+                // I'm not sure if this is the right behavior, I think there should be no author and author_* fields, just author_
```
kamilkisiela, line 1244: Could somebody confirm?

lutter: Yes, I think that's true; AFAICT, `author_cond: ...` would be the same as `author_ { id_cond: .. }`. The first form would also not work if `author` is a derived field.

graphql/src/store/query.rs
```diff
@@ -141,6 +149,14 @@ fn build_fulltext_filter_from_object(
     )
 }
 
+fn resolve_type_name(ty: &Type) -> &String {
+    match ty {
+        Type::ListType(inner) => resolve_type_name(inner),
+        Type::NonNullType(inner) => resolve_type_name(inner),
+        Type::NamedType(name) => name,
+    }
+}
```
lutter, line 158: This is `get_base_type` from `graph/src/data/graphql` (just do something like `use graph::data::graphql::TypeExt as _` to be able to use it)

graphql/tests/query.rs
```diff
@@ -553,6 +553,63 @@ fn can_query_many_to_many_relationship() {
...
+        let exp = object! {
+            musicians: vec![
+                object! { name: "Tom", mainBand: object! { id: "b2"} }
+            ]
+        };
+
+        let data = extract_data!(result).unwrap();
+        assert_eq!(data, exp);
+    })
+}
```
lutter, line 611: Nice!

store/postgres/src/relational_queries.rs
```diff
@@ -884,24 +887,56 @@ impl<'a> fmt::Display for QueryFilter<'a> {
...
             And(filters) | Or(filters) => {
                 for filter in filters {
-                    Self::valid_attributes(filter, table)?;
+                    Self::valid_attributes(filter, table, layout, child_filter_ancestor)?;
                 }
             }
+            Child(attr, entity, child_filter, _) => {
+                if child_filter_ancestor {
+                    return Err(StoreError::QueryExecutionError(
+                        "Only a single level sub filter is allowed".to_string(),
```
lutter, line 923: How about something like `Child filters can not be nested`?

store/postgres/src/relational_queries.rs
```diff
@@ -932,7 +967,114 @@ impl<'a> QueryFilter<'a> {
...
+            .layout
+            .table_for_entity(entity_type)
+            .expect("Table for child entity not found");
+
+        let child_prefix = "i.";
+        let parent_prefix = "c.";
+
+        out.push_sql("exists (select 1");
+
+        out.push_sql(" from ");
```
lutter, line 994: Very minor but these two `push_sql` could be collapsed into one

store/postgres/src/relational_queries.rs
```diff
@@ -932,7 +967,114 @@ impl<'a> QueryFilter<'a> {
...
+            // If the parent is derived,
+            // the child column is picked based on the provided attribute
+            // and the parent column is the primary key of the parent table
+            let child_column = child_table
+                .column_for_field(attribute)
+                .expect("Column for an attribute not found");
+            let parent_column = self.table.primary_key();
+
+            if child_column.is_list() {
+                // Type A: p.id = any(c.{parent_field})
```
lutter, line 1011: Shouldn't that be `c.id = any(i.{parent_field})` ? And similar in the comments for the other cases

store/postgres/src/relational_queries.rs
```diff
@@ -932,7 +967,114 @@ impl<'a> QueryFilter<'a> {
...
+            }
         }
+
+        out.push_sql(" and ");
+
+        // Match by block
+        out.push_sql(child_prefix);
+        out.push_identifier(BLOCK_RANGE_COLUMN)?;
+        out.push_sql(" @> ");
+        out.push_bind_param::<Integer, _>(&self.block)?;
```
lutter, line 1060: This needs to use `BlockRangeColumn::contains` - the way it is written now only works for mutable entities, but would fail for immutable ones since they don't have a `block_range` column, only a `block$` column

lutter (changes requested):
This is getting really, really close. There's a few smallish things to touch up, but other than that, this is almost ready.

graph/src/components/store/mod.rs
```diff
@@ -82,6 +82,19 @@ impl From<&str> for EntityType {
 
 impl CheapClone for EntityType {}
 
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct EntityFilterDerivative(bool);
```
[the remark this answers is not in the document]
kamilkisiela: 845b35e79336f0c1a793de7bd702da2ea7cbef5d

graphql/src/schema/api.rs
```diff
@@ -338,40 +374,67 @@ fn field_list_filter_input_values(
...
+            let mut input_values: Vec<InputValue> = vec![];
+
+            if let Some(parent) = parent_type_name {
+                extend_with_child_filter_input_value(field, &parent, &mut input_values);
+            }
+
+            return Some(input_values);
+        }
+
+        let input_field_type = input_field_type.unwrap();
```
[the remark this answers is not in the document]
kamilkisiela: 1baae95d0e8270fad283022d639e4b55a683e218

graphql/src/store/query.rs
kamilkisiela: 866883c02f2c310f724cf7555939ea8a31547382

store/postgres/src/relational_queries.rs
kamilkisiela: f00526fefdf012e559f5a06220aad0fa1d968f86

store/postgres/src/relational_queries.rs
kamilkisiela: 7305f53a28989758fcf3ce04123940ac559c03f1

store/postgres/src/relational_queries.rs
kamilkisiela: f6721a3f344acdc80d0d5dfec1146854b95ee24f

store/postgres/src/relational_queries.rs
kamilkisiela: a7c70e42cbb1919be46b9af02c4a129f228217b1

graphql/src/schema/api.rs
```diff
@@ -228,7 +228,30 @@ fn field_filter_input_values(
...
+                            schema,
+                            field,
+                            &ScalarType::new(String::from("String")),
+                        );
+
+                        extend_with_child_filter_input_value(field, name, &mut input_values);
+
+                        input_values
+                    }
+                }
```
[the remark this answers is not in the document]
kamilkisiela: 474af82baa1b8f9e444b6b8e07e7c0a86927870c

lutter (approved):

lutter:
This is good to go - there's one small thing around how the commits are structured/named: see [this](https://github.com/graphprotocol/graph-node/blob/master/CONTRIBUTING.md#commit-messages-and-pull-requests) for how we usually do that. With this PR, it might be best to just squash everything into a mega commit `all: Add child filters`

kamilkisiela:
@lutter I squashed all commits and renamed it

mpaler:
@kamilkisiela @lutter and anyone else involved in this PR. Kinda late to the party on this one, but I'd like to offer a HUGE thank you for getting this in the the graph-node. It's just such an amazing boost to querying. Thank you!!

PR #3155, opened 2022-01-13
GraphQL API Versioning

kamilkisiela:
This is the initial implementation of versioning.

To do:
- [ ] resolve version of WebSocket connections
- [ ] switch from header-based approach to url-based
- [ ] make sure we use default (latest) version correctly
- [ ] remove the example implementation of a feature flag

I left few comments on the interesting parts of the code to make it easier to review.
- [Where do we manage versions](https://github.com/graphprotocol/graph-node/pull/3155#discussion_r784789121)
- [How do we define "latest" version](https://github.com/graphprotocol/graph-node/pull/3155#discussion_r784789581)
- [How does the Gateway fetch a list of supported versions?](https://github.com/graphprotocol/graph-node/pull/3155#discussion_r784792602)
- [How do we define feature flags?](https://github.com/graphprotocol/graph-node/pull/3155#pullrequestreview-852838810)
- [How do we validate the version number provided by the user?](https://github.com/graphprotocol/graph-node/pull/3155#discussion_r784790014)

PR #3185, opened 2022-01-25
all: GraphQL API Versioning

kamilkisiela:
This is the initial implementation of versioning.

- [x] remove the example implementation of a feature flag

I left few comments on the interesting parts of the code to make it easier to review.
- [Where do we manage versions](https://github.com/graphprotocol/graph-node/pull/3185#discussion_r791502259)
- [How do we define "latest" version](https://github.com/graphprotocol/graph-node/pull/3185#discussion_r791502627)
- [How does the Gateway fetch a list of supported versions?](https://github.com/graphprotocol/graph-node/pull/3185#discussion_r791504128)
- [How do we define feature flags?](https://github.com/graphprotocol/graph-node/pull/3185#discussion_r791501883)
- [How do we validate the version number provided by the user?](https://github.com/graphprotocol/graph-node/pull/3185#discussion_r791503067)

Continues #3155

### DIFF
--- a/graph/src/components/versions/features.rs
+++ b/graph/src/components/versions/features.rs
@@ -1,0 +1,2 @@
+#[derive(Clone, PartialEq, Eq, Debug, Ord, PartialOrd, Hash)]
+pub enum FeatureFlag {}
--- a/graph/src/components/versions/registry.rs
+++ b/graph/src/components/versions/registry.rs
@@ -1,0 +1,71 @@
+use crate::prelude::FeatureFlag;
+use itertools::Itertools;
+use lazy_static::lazy_static;
+use semver::{Version, VersionReq};
+use std::collections::HashMap;
+
+lazy_static! {
+    static ref VERSION_COLLECTION: HashMap<Version, Vec<FeatureFlag>> = {
+        vec![
+            // baseline version
+            (Version::new(1, 0, 0), vec![]),
+        ].into_iter().collect()
+    };
+
+    // Sorted vector of versions. From higher to lower.
+    pub static ref VERSIONS: Vec<&'static Version> = {
+        let mut versions = VERSION_COLLECTION.keys().collect_vec().clone();
+        versions.sort_by(|a, b| b.partial_cmp(a).unwrap());
+        versions
+    };
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct ApiVersion {
+    pub version: Version,
+    features: Vec<FeatureFlag>,
+}
+
+impl ApiVersion {
+    pub fn new(version_requirement: &VersionReq) -> Result<Self, String> {
+        let version = Self::resolve(&version_requirement)?;
+
+        Ok(Self {
+            version: version.clone(),
+            features: VERSION_COLLECTION
+                .get(&version)
+                .expect(format!("Version {:?} is not supported", version).as_str())
+                .to_vec(),
+        })
+    }
+
+    pub fn from_version(version: &Version) -> Result<ApiVersion, String> {
+        ApiVersion::new(
+            &VersionReq::parse(version.to_string().as_str())
+                .map_err(|error| format!("Invalid version requirement: {}", error))?,
+        )
+    }
+
+    pub fn supports(&self, feature: FeatureFlag) -> bool {
+        self.features.contains(&feature)
+    }
+
+    fn resolve(version_requirement: &VersionReq) -> Result<&Version, String> {
+        for version in VERSIONS.iter() {
+            if version_requirement.matches(version) {
+                return Ok(version.clone());
+            }
+        }
+
+        Err("Could not resolve the version".to_string())
+    }
+}
+
+impl Default for ApiVersion {
+    fn default() -> Self {
+        // Default to the latest version.
+        // The `VersionReq::default()` returns `*` which means "any version".
+        // The first matching version is the latest version.
+        ApiVersion::new(&VersionReq::default()).unwrap()
+    }
+}
--- a/graph/src/data/query/query.rs
+++ b/graph/src/data/query/query.rs
@@ -7,7 +7,7 @@ use std::sync::Arc;
 
 use crate::{
     data::graphql::shape_hash::shape_hash,
-    prelude::{q, r, DeploymentHash, SubgraphName, ENV_VARS},
+    prelude::{q, r, ApiVersion, DeploymentHash, SubgraphName, ENV_VARS},
 };
 
 fn deserialize_number<'de, D>(deserializer: D) -> Result<q::Number, D::Error>
@@ -112,19 +112,15 @@ impl serde::ser::Serialize for QueryVariables {
 
 #[derive(Clone, Debug)]
 pub enum QueryTarget {
-    Name(SubgraphName),
-    Deployment(DeploymentHash),
+    Name(SubgraphName, ApiVersion),
+    Deployment(DeploymentHash, ApiVersion),
 }
 
-impl From<DeploymentHash> for QueryTarget {
-    fn from(id: DeploymentHash) -> Self {
-        Self::Deployment(id)
-    }
-}
-
-impl From<SubgraphName> for QueryTarget {
-    fn from(name: SubgraphName) -> Self {
-        QueryTarget::Name(name)
+impl QueryTarget {
+    pub fn get_version(&self) -> &ApiVersion {
+        match self {
+            Self::Deployment(_, version) | Self::Name(_, version) => version,
+        }
     }
 }
 
--- a/server/http/src/service.rs
+++ b/server/http/src/service.rs
@@ -5,6 +5,7 @@ use std::task::Poll;
 use std::time::Instant;
 
 use graph::prelude::*;
+use graph::semver::VersionReq;
 use graph::{components::server::query::GraphQLServerError, data::query::QueryTarget};
 use http::header;
 use http::header::{
@@ -89,17 +90,49 @@ where
         self.serve_dynamic_file(self.graphiql_html())
     }
 
+    fn resolve_api_version(
+        &self,
+        request: &Request<Body>,
+    ) -> Result<ApiVersion, GraphQLServerError> {
+        let mut version = ApiVersion::default();
+
+        if let Some(query) = request.uri().query() {
+            let potential_version_requirement = query.split("&").find_map(|pair| {
+                if pair.starts_with("api-version=") {
+                    if let Some(version_requirement) = pair.split("=").nth(1) {
+                        return Some(version_requirement);
+                    }
+                }
+                return None;
+            });
+
+            if let Some(version_requirement) = potential_version_requirement {
+                version = ApiVersion::new(
+                    &VersionReq::parse(version_requirement)
+                        .map_err(|error| GraphQLServerError::ClientError(error.to_string()))?,
+                )
+                .map_err(|error| GraphQLServerError::ClientError(error))?;
+            }
+        }
+
+        Ok(version)
+    }
+
     async fn handle_graphql_query_by_name(
         self,
         subgraph_name: String,
         request: Request<Body>,
     ) -> GraphQLServiceResult {
+        let version = self.resolve_api_version(&request)?;
         let subgraph_name = SubgraphName::new(subgraph_name.as_str()).map_err(|()| {
             GraphQLServerError::ClientError(format!("Invalid subgraph name {:?}", subgraph_name))
         })?;
 
-        self.handle_graphql_query(subgraph_name.into(), request.into_body())
-            .await
+        self.handle_graphql_query(
+            QueryTarget::Name(subgraph_name, version),
+            request.into_body(),
+        )
+        .await
     }
 
     fn handle_graphql_query_by_id(
@@ -108,11 +141,16 @@ where
         request: Request<Body>,
     ) -> GraphQLServiceResponse {
         let res = DeploymentHash::new(id)
-            .map_err(|id| GraphQLServerError::ClientError(format!("Invalid subgraph id `{}`", id)));
+            .map_err(|id| GraphQLServerError::ClientError(format!("Invalid subgraph id `{}`", id)))
+            .and_then(|id| match self.resolve_api_version(&request) {
+                Ok(version) => Ok((id, version)),
+                Err(error) => Err(error),
+            });
+
         match res {
             Err(_) => self.handle_not_found(),
-            Ok(id) => self
-                .handle_graphql_query(id.into(), request.into_body())
+            Ok((id, version)) => self
+                .handle_graphql_query(QueryTarget::Deployment(id, version), request.into_body())
                 .boxed(),
         }
     }
--- a/server/index-node/src/schema.graphql
+++ b/server/index-node/src/schema.graphql
@@ -43,6 +43,7 @@ type Query {
     network: String!
     blockHash: Bytes!
   ): [CachedEthereumCall!]
+  apiVersions(subgraphId: String!): [SubgraphVersion!]!
 }
 
 type SubgraphIndexingStatus {
@@ -177,4 +178,11 @@ type ProofOfIndexingResult {
   block: Block!
   "There may not be a proof of indexing available for the deployment and block"
   proofOfIndexing: Bytes
+}
+
+type ApiVersion {
+  """
+  Version number in SemVer format
+  """
+  version: String!
 }
--- a/server/websocket/src/connection.rs
+++ b/server/websocket/src/connection.rs
@@ -297,7 +297,7 @@ where
                     };
 
                     // Construct a subscription
-                    let target = QueryTarget::Deployment(deployment.clone());
+                    let target = QueryTarget::Deployment(deployment.clone(), Default::default());
                     let subscription = Subscription {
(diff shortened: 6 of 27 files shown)


server/websocket/src/server.rs
```diff
@@ -72,14 +76,18 @@ where
...
-                Ok(state(store, target_from_name(path_segments[2..].join("/"))).await)
-            }
-            &["subgraphs", "network", _, _] => {
-                Ok(state(store, target_from_name(path_segments[1..].join("/"))).await)
+                Ok(state(store, target_from_id(subgraph_id, Default::default())).await)
             }
+            &["subgraphs", "name", _] | &["subgraphs", "name", _, _] => Ok(state(
+                store,
+                target_from_name(path_segments[2..].join("/"), Default::default()), // TODO: version
+            )
```
kamilkisiela, line 84: - [ ] resolve version of WebSocket connections

store/postgres/src/deployment_store.rs
```diff
@@ -276,7 +277,9 @@ impl DeploymentStore {
         // if that's Fred the Dog, Fred the Cat or both.
         //
         // This assumes that there are no concurrent writes to a subgraph.
-        let schema = self.subgraph_info_with_conn(&conn, &layout.site)?.api;
+        let schema = self
+            .subgraph_info_with_conn(&conn, &layout.site, &Default::default())? // TODO: ask if we need a non-default version here
+            .api;
```
kamilkisiela, line 282: - [ ] Do we need a version here? Is default (latest) enough?

store/postgres/src/subgraph_store.rs
```diff
@@ -676,11 +676,11 @@ impl SubgraphStoreInner {
...
     ) -> Result<(Arc<DeploymentStore>, Arc<Site>, ReplicaId), StoreError> {
         let id = match target {
-            QueryTarget::Name(name) => self.mirror.current_deployment_for_subgraph(&name)?,
-            QueryTarget::Deployment(id) => id,
+            QueryTarget::Name(name, _) => self.mirror.current_deployment_for_subgraph(&name)?,
+            QueryTarget::Deployment(id, _) => id,
         };
 
-        let (store, site) = self.store(&id)?;
+        let (store, site) = self.store(&id)?; // TODO: maybe use a tuple (id, version) here?
```
kamilkisiela, line 683: - [ ] not sure if we want to get the store based on `id` or `(id, version)`.

store/postgres/src/subgraph_store.rs
```diff
@@ -855,7 +855,7 @@ impl SubgraphStoreInner {
                 .ok_or_else(|| constraint_violation!("no chain info for {}", deployment_id))?;
             let latest_ethereum_block_number =
                 chain.latest_block.as_ref().map(|ref block| block.number());
-            let subgraph_info = store.subgraph_info(site.as_ref())?;
+            let subgraph_info = store.subgraph_info(site.as_ref(), &Default::default())?; // TODO: ask if we need to use non-default version here
```
kamilkisiela, line 858: - [ ] Do we need a version here? Is default (latest) enough?

store/postgres/src/subgraph_store.rs
```diff
@@ -1054,13 +1054,18 @@ impl SubgraphStoreTrait for SubgraphStore {
 
     fn input_schema(&self, id: &DeploymentHash) -> Result<Arc<Schema>, StoreError> {
         let (store, site) = self.store(&id)?;
-        let info = store.subgraph_info(site.as_ref())?;
+        // TODO: ask if we need to use non-default version here, I think we don't because it's schema provided by user
+        let info = store.subgraph_info(site.as_ref(), &Default::default())?;
```
kamilkisiela, line 1058: - [ ] Do we need a version here? Is default (latest) enough?

I think we don't because it's schema provided by user

graph/src/components/versions/features.rs
```diff
@@ -0,0 +1,5 @@
+#[derive(Clone, PartialEq)]
+pub enum FeatureFlag {
+    // A description of the feature. Give it a little context.
+    BasicOrdering,
```
kamilkisiela, line 4: An example flag, just to demo the versioning.

graph/src/components/versions/registry.rs
```diff
@@ -0,0 +1,112 @@
...
+use std::hash::Hash;
+
+lazy_static! {
+    pub static ref VERSIONS: HashMap<VersionNumber, Vec<FeatureFlag>> = {
+        let supported_versions: Vec<(&str, Vec<FeatureFlag>)> = vec![
+            // baseline version
+            ("1.0.0", vec![]),
+            // Versions with feature flags
+            ("1.1.0", vec![FeatureFlag::BasicOrdering])
+        ];
```
kamilkisiela, line 14: This is something we need to maintain. It's the single source of truth.

graph/src/components/versions/registry.rs
```diff
@@ -0,0 +1,112 @@
...
+        map
+    };
+
+    static ref LATEST_VERSION: String = {
+        let keys: Vec<VersionNumber> = VERSIONS.clone().into_keys().collect();
+
+        let last_version = keys.last().unwrap();
+
+        last_version.0.clone()
+    };
```
kamilkisiela, line 31: Latest version is the last key of `VERSIONS`.

graph/src/components/versions/registry.rs
```diff
@@ -0,0 +1,112 @@
...
+        if major.is_err() || minor.is_err() || patch.is_err() {
+            return Err(format!("Invalid version number: {}", version));
+        }
+
+        if !VERSIONS.contains_key(&VersionNumber::from(version)) {
+            return Err("No versions found".to_string());
+        }
+
+        Ok(())
+    }
```
kamilkisiela, line 71: Validation happens in http and websocket part of the server.

graph/src/data/query/query.rs
```diff
@@ -112,19 +112,20 @@ impl serde::ser::Serialize for QueryVariables {
 
 #[derive(Clone, Debug)]
 pub enum QueryTarget {
-    Name(SubgraphName),
-    Deployment(DeploymentHash),
+    Name(SubgraphName, VersionNumber),
+    Deployment(DeploymentHash, VersionNumber),
 }
```
kamilkisiela, line 117: I tried to look for the best place to put the requested version and `QueryTarget` won the contest.

graphql/src/schema/api.rs
```diff
@@ -84,14 +84,20 @@ pub fn api_schema(input_schema: &Document) -> Result<Document, APISchemaError> {
     let mut schema = input_schema.clone();
     add_directives(&mut schema);
     add_builtin_scalar_types(&mut schema)?;
-    add_order_direction_enum(&mut schema);
+
+    if version.supports(FeatureFlag::BasicOrdering) {
+        add_order_direction_enum(&mut schema);
+    }
```
kamilkisiela, line 90: An example implementation of a feature flag.

server/http/src/service.rs
```diff
@@ -166,17 +166,45 @@ where
...
+                ))
+            })?;
+            VersionNumber::validate(version_value.to_string())
+                .map_err(|error| GraphQLServerError::ClientError(error))?;
+
+            version = version_value.into();
+        }
+
+        Ok(version)
+    }
```
kamilkisiela, line 191: I used `api-version` header because it was the easiest to implement quickly but before merging the Pull Request we should switch to url-based approach (this way we could use GraphiQL - it does not support headers).

server/index-node/src/schema.graphql
```diff
@@ -19,6 +19,7 @@ type Query {
     indexer: Bytes
   ): Bytes
   subgraphFeatures(subgraphId: String!): SubgraphFeatures!
+  subgraphVersions(subgraphId: String!): [SubgraphVersion!]!
```
kamilkisiela, line 22: This is for the Gateway.

server/index-node/src/schema.graphql
```diff
@@ -95,3 +96,10 @@ enum Feature {
   fullTextSearch,
   ipfsOnEthereumContracts,
 }
+
+type SubgraphVersion {
+  """
+  Version number in SemVer format
+  """
+  version: String!
+}
```
kamilkisiela, line 105: Why an object instead of a list? It's extendable and harder to break backwards-compatibility.

server/websocket/src/connection.rs
```diff
@@ -313,7 +313,7 @@ where
                     };
 
                     // Construct a subscription
-                    let target = QueryTarget::Deployment(deployment.clone());
+                    let target = QueryTarget::Deployment(deployment.clone(), Default::default()); // TODO: resolve and validate a version number from WebSocket connection
```
kamilkisiela, line 316: - [ ] resolve version of WebSocket connections

graphql/src/schema/api.rs
```diff
@@ -74,19 +74,20 @@ impl TryFrom<&r::Value> for ErrorPolicy {
...
 
     // Refactor: Don't clone the schema.
     let mut schema = input_schema.clone();
     add_meta_field_type(&mut schema);
-    add_types_for_object_types(&mut schema, &object_types)?;
-    add_types_for_interface_types(&mut schema, &interface_types)?;
-    add_field_arguments(&mut schema, input_schema)?;
-    add_query_type(&mut schema, &object_types, &interface_types)?;
-    add_subscription_type(&mut schema, &object_types, &interface_types)?;
+
```
kamilkisiela, line 85:
```suggestion
```

evaporei:
@kamilkisiela is this ready to review? I see that the CI is failing.

kamilkisiela:
@evaporei yes, the code won't change much even after a rebase. I've got some questions both in code and gh comments, because I'm not sure if it's the right implementation.

server/index-node/src/schema.graphql
lutter: Nice! We should avoid the words 'subgraph version' here since it already has a meaning in the context of `graph-node` as the mapping of subgraph names to deployments (IPFS hash) How about calling this `apiVersions`?  (That naming change should of course happen throughout the code supporting this GraphQL API)

server/websocket/src/connection.rs
lutter: I think we can leave this as-is since there are no subscriptions on the network anyway

store/postgres/src/subgraph_store.rs
lutter: Since the version only affects the API schema, I don't think there's a point in having the version here. I think version should only come into play for the `QueryStore`

store/postgres/src/deployment_store.rs
```diff
@@ -87,7 +87,7 @@ pub struct StoreInner {
     conn_round_robin_counter: AtomicUsize,
 
     /// A cache of commonly needed data about a subgraph.
-    subgraph_cache: Mutex<LruCache<DeploymentHash, SubgraphInfo>>,
+    subgraph_cache: Mutex<LruCache<(DeploymentHash, VersionNumber), SubgraphInfo>>,
```
lutter, line 90: How about keeping `DeploymentHash` as the cache key and changing `SubgraphInfo.api` to be a `HashMap<Version, Arc<ApiSchema>>`? As long as there's only a few API versions, we could even populate that `HashMap` eagerly; once there are more versions, we might want to populate entries there on demand.

That would make it possible to not pass in a version to `subgraph_info` and only use it when we actually need an `ApiSchema` in the `QueryStore`

lutter (changes requested):

kamilkisiela:
@azf20 no updates, I'm trying to implement support for interfaces in #3184 and once it's ready, I'll apply David's comments here.

server/index-node/src/schema.graphql
kamilkisiela: Done

store/postgres/src/deployment_store.rs
kamilkisiela: eaa5c19a8384bd804bc8b25aa11489e3ba2dabc1

kamilkisiela:
@lutter ready for a review again, I implemented everything you asked for. 

Those two failing tests (just the order of fields) are related to the fact I moved the ordering logic behind a feature flag to show what the versioning will look like. Once I remove the example feature flag, it will be fine.

kamilkisiela:
@azf20 I read the GIP and applied some changes to the PR. When we get a version range, the semver package resolves to the most recent applicable version.

graph/src/components/versions/registry.rs
```diff
@@ -0,0 +1,87 @@
...
+            // baseline version
+            (Version::new(1, 0, 0), vec![]),
+            // Versions with feature flags
+            (Version::new(1, 1, 0), vec![FeatureFlag::BasicOrdering])
+        ];
+
+        let mut map = HashMap::new();
+
+        // Sort version by major, minor, patch, from higher to lower.
+        supported_versions.sort_by(|a, b| b.0.partial_cmp(&a.0).unwrap());
```
lutter, line 18: Is this sort necessary? They all go into a `HashMap` afterwards, so doesn't seem like it's needed.

graph/src/components/versions/registry.rs
```diff
@@ -0,0 +1,87 @@
...
+        }
+
+        map
+    };
+
+    static ref LATEST_VERSION: String = {
+        let keys: Vec<Version> = VERSIONS.clone().into_keys().collect();
+
+        // Versions are sorted by major, minor, patch, from higher to lower.
+        keys.first().unwrap().to_string()
```
lutter, line 31: This is wrong `HashMap.into_keys` will return keys in arbitrary order. `BTreeMap` would work, but I think `first` would return the smallest version, so either reverse the iterator or put the keys into a `Vec` and sort that

graph/src/components/versions/registry.rs
```diff
@@ -0,0 +1,87 @@
...
+
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct ApiVersion {
+    pub version: Version,
+    features: Vec<FeatureFlag>,
+}
+
+impl ApiVersion {
+    pub fn new(version_requirement: &VersionReq) -> Result<Self, String> {
+        let version = Self::resolve(&version_requirement);
```
lutter, line 43: It would be cleaner if `resolve` returned `Result<Self, String>` instead of an `Option`

graphql/src/schema/api.rs
```diff
@@ -643,32 +655,36 @@ fn collection_arguments_for_named_type(type_name: &str) -> Vec<InputValue> {
...
         skip,
         first,
         input_value(
+            &"where".to_string(),
+            "",
+            Type::NamedType(format!("{}_filter", type_name)),
+        ),
+    ];
+
+    if version.supports(FeatureFlag::BasicOrdering) {
```
lutter, line 668: I am losing track of the plot a little here - should we really have a version without ordering?

lutter (changes requested):
The main thing we need to clear up is what the initial version(s) should be

graphql/src/schema/api.rs
kamilkisiela: No no, it's just an example of how to use the feature flags
https://github.com/graphprotocol/graph-node/pull/3185#discussion_r791501883

graph/src/components/versions/registry.rs
kamilkisiela: It does now, I did a refactor

graph/src/components/versions/registry.rs
kamilkisiela: I did a refactor and it's now using `VERSION_COLLECTION: HashMap<Version, Vec<FeatureFlag>>` to get the feature flags and `VERSIONS: Vec<&'static Version>` (sorted from latest to oldest version, higher -> lower).

The `VERSIONS` vector is sorted because of the logic in `resolve()` function.

graph/src/components/versions/registry.rs
kamilkisiela: I did a refactor, now only the vector is sorted

kamilkisiela:
@lutter I did a refactor, it includes your suggestions.

lutter (approved):
Looks good now!

@azf20 Are you ok with the base version being `1.0.0`? All fine by me, just wanted to check that that's where we should start.

azf20:
Thanks @kamilkisiela and @lutter! I think starting with 1.0.0 is ok, but also interested in @That3Percent's expectations of a 1.x release

That3Percent:
> I think starting with 1.0.0 is ok, but also interested in @That3Percent's expectations of a 1.x release

I also think a 1.0.0 release is a good choice. Even if there are changes we know we would like to make, that's _always_ the case so it's better to have a 1.0 and 2.0 than 0.x in perpetuity.

That3Percent (commented):
Currently the protocol expects version "0" to exist so that we can be backward-compatible with the existing attestation format. As implemented, graph-node is only supporting 1.0.0 here.

I think this is probably fine, as long as we default to 1.0 upstream from here (indexer agent, gateway) and accept that version "0" should never be queried directly (it would be a historical relic). Thoughts?

azf20:
Agree that this approach seems fine, and I think the GIP already allows for that kind of minimum version specification

azf20:
Hey @kamilkisiela I think this is good to merge, though looks like there are a couple of conflicts first! cc @lutter

Issue #3077, opened 2021-12-16
GraphQL parse/validation caching 


dotansimha:
Since GraphQL `parse` and `validate` phase are static (does not depend on the user input like variables), we can implement caching for both phases (for parser: `String => query::Document`, for validation: `String | query::Document => Vec<ValidationError>`).

This might be more relevant once https://github.com/graphprotocol/graph-node/pull/3057 lands, because if we'll add more validation rules it might add some (probably minimal) overhead. 

Usually it's implemented with LRU.

lutter:
That's a good idea! For our purposes, I think it would be enough to keep a `HashSet` of the shape hashes of the queries that were successfully validated. We already calculate that hash (a `u64`) for all queries anyway.

dotansimha:
https://github.com/graphprotocol/graph-node/pull/3188#issuecomment-1021490277

PR #3188, opened 2022-01-25
graphql: added caching for validation phase

dotansimha:
Closes https://github.com/graphprotocol/graph-node/issues/3077 

### TODO

- [x] Add global cache for validation results
- [ ] Make sure cache does not cause false positives due to hash calculation 

### Overview of this PR

Since `validate` is static (give a query and a schema it will always produce the same result), we can cache it to avoid running it again and again if a given query hash is the same.

### Benchmarks 

Measured on M1 machine, in `--release` mode.

| Phase    | Without Cache | With `HashMap` |
|----------|---------------|----------------|
| validate | ~1.44ms        | <=1µs          |

<details>
<summary>GraphQL query used for testing</summary>

```graphql
{
  purposes(first: 100, subgraphError: allow) {
    id
    sender {
      id
      ...SenderFields
    }
    ...PurposeFields
  }
}

fragment PurposeFields on Purpose {
  purpose
  createdAt
  transactionHash
}

fragment SenderFields on Sender {
  address
  createdAt
  purposeCount
}

```
</details>

> It seems like `parse` takes ~60µs so we don't have to do any significant optimizations at the moment.

### DIFF
--- a/graphql/src/execution/query.rs
+++ b/graphql/src/execution/query.rs
@@ -2,12 +2,13 @@ use graph::data::graphql::DocumentExt as _;
 use graph::data::value::Object;
 use graphql_parser::Pos;
 use graphql_tools::validation::rules::*;
+use graphql_tools::validation::utils::ValidationError;
 use graphql_tools::validation::validate::{validate, ValidationPlan};
 use lazy_static::lazy_static;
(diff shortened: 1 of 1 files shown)


dotansimha:
After a discussion with @lutter , we can drop caching validations at the moment, since it's not adding a real overhead (~1ms).

Issue #3118, opened 2022-01-04
Mandate everything runs in some fixed locale `en_US.UTF8`


saihaj:
> We should mandate that everything runs in some fixed locale (maybe `en_US.UTF8`, though we'd need to look into if there are other locales that are more palatable to the non-English world) `graph-node` could then check on startup that its locale and the locale of the database are the expected one.

_Originally posted by @lutter in https://github.com/graphprotocol/graph-node/issues/1837#issuecomment-984160371_

saihaj:
I search the codebase and I see this comment https://github.com/graphprotocol/graph-node/blob/85ac00cd19cabce1fbcf0979d727fb15105333d8/store/postgres/src/connection_pool.rs#L589-L594 which means non english DBs shouldn't work right cause they get marked as unavailable. That means we don't need to panic or anything? cc @lutter

azf20:
@saihaj I wonder if we can add a specific check for `en_US.UTF8`? I think for determinism it's not enough for it to be in English, I think it needs to be specific

leoyvens:
The POSIX locale is better for performance, so it seems worth discussing if we should choose that.

lutter:
[This discussion](https://dba.stackexchange.com/questions/240930/postgresql-difference-between-collations-c-and-c-utf-8) is a good start. One thing to note is that switching collation would require rebuilding all indexes with `text` columns in them.

azf20:
The majority of indexers responding to the straw poll are running `en_US.UTF8`. Proposing that we adopt that as the standard on the network, but @lutter am wondering what the migration path is for any indexers currently running a different locale? I don't know what the implications are - ideally it would be as simple as a reconfiguration, but presumably indices would need to be regenerated?

lutter:
The 'right' choice would be to use the `C` locale because it performs better in queries etc. than `en_US.UTF8` (I am told that it can speed up comparisons by 10x)

Either way, the safest way to change locale is to recreate the database - the default locale is baked in when the database is initialized. That means the indexers would have to dump their database, recreate it, and restore from backup. There _might_ be ways for us to change that when generating queries, but it'll require some more work within `graph-node`

azf20:
OK - so they wouldn't need to re-index, just restore from backup? Interested in @fordN's opinion from an indexer perspective here?

saihaj:
https://github.com/graphprotocol/graph-node/pull/4163

PR #3191, opened 2022-01-26
store: make comparisons case-insensitive

### LINKED ISSUE #3116: Change `_contains` `_starts_with`, `_ends_with` functionality to be case insensitive

saihaj:
Recreating #3126
Closes #3116﻿

### DIFF
--- a/store/postgres/src/relational_queries.rs
+++ b/store/postgres/src/relational_queries.rs
@@ -873,9 +873,9 @@ impl<'a> QueryFilter<'a> {
             Value::String(s) => {
                 out.push_identifier(column.name.as_str())?;
                 if negated {
-                    out.push_sql(" not like ");
+                    out.push_sql(" not ilike ");
                 } else {
-                    out.push_sql(" like ")
+                    out.push_sql(" ilike ")
                 };
                 if s.starts_with('%') || s.ends_with('%') {
                     out.push_bind_param::<Text, _>(s)?;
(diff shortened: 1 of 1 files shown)


lutter (commented):
I am totally fine with this change, but it is a breaking change in that the same query may now return different data. 

@azf20 Are you ok with merging this?

azf20:
Nice! cc @That3Percent. I think this depends a bit on our apiVersioning story - this is quite a significant change in query behaviour, so I think a question is if we make this part of the `first` graphQL Version or not?

That3Percent:
> Nice! cc @That3Percent. I think this depends a bit on our apiVersioning story - this is quite a significant change in query behaviour, so I think a question is if we make this part of the `first` graphQL Version or not?

This PR seems to have some problems.

For one, this will break existing dApps which were expecting the existing case-sensitive behavior. A better option I think would have been to add case-insensitive variants of these arguments, rather than to modify the existing ones.

Another concern comes quoted from the issue:

> case-sensitivity is locale-dependent, and the response will depend on the locale in which the database is running, but that's not a new thing, we have that with ordering too.

Determinism is table-stakes. We must not introduce non-determinism, and we must eradicate the existing non-determinism. So, that part gets a hard no from me unless all queries using these features are errors unless the non-determinism flag is set in graph-node... which would break all the existing queries.

azf20:
Reasonable on breaking existing dapps - @lutter @saihaj @dotansimha what do you think about introducing a new `i_contains`, `i_starts_with` option in the `where`, to provide specific case insensitive behaviour? My instinct is that the case-sensitivity is not what most users would expect, but I can't think of an easy way to validate if dapps are relying on it.

We are addressing the database locale issue separately: https://github.com/graphprotocol/graph-node/issues/3118

saihaj:
@That3Percent agreed that this is a breaking change and to address it as discussed in the ticket we are adding strict filters #3117  so should we instead keep this as is and then and an insensitive filter?

azf20:
@saihaj it seems that we agree - let's keep this `as is` and introduce a case insensitive filter. We can then monitor usage to try and understand if users are relying on the sensitivity

saihaj:
@azf20 https://github.com/graphprotocol/graph-node/pull/3294

Issue #3114, opened 2022-01-04
Querying for root `__typename` fails due to `null` response


dotansimha:
**Do you want to request a *feature* or report a *bug*?**

bug

**What is the current behavior?**

When querying for root `__typename` (directly under query), `graph-node` replies with an error, due to `null` response.

This is needed because in many GraphQL clients, the `__typename` is being added to every selection-set, in order to get knowledge of the GraphQL.
Also, `graphql-tools` and schema-stitching need that field in order to make sure queries are stitched correctly under the same root `Query` type.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem.**

```graphql
query {
  __typename
}
```

response:

```json
{
  "errors": [
    {
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "message": "Null value resolved for non-null field `__typename`"
    }
  ]
}
```

**What is the expected behavior?**

Should return `"Query"`.

PR #3193, opened 2022-01-26
graphql: resolve __typename if not exists in the resolved object

ardatan:
Fixes https://github.com/graphprotocol/graph-node/issues/3114

Given the following query: 

```graphql
{
  __typename # this one is the problematic one
  purposes {
    __typename
  }
}
```

Should return: 

```json
{
   "__typename": "Query",
   "purposes": [
      {
         "__typename": "Purpose"
      }
   ]
}
```

But it fails to resolve `"__typename": "Query",`, because we don't handle `__typename` not only for root types but all other types.

It works for `Purpose` for example because `__typename` is added by resolvers level(prefetch) for other object types.

### DIFF
--- a/graphql/src/execution/execution.rs
+++ b/graphql/src/execution/execution.rs
@@ -257,7 +257,7 @@ pub(crate) fn execute_root_selection_set_uncached(
         // the data_set SelectionSet
         if is_introspection_field(&field.name) {
             intro_set.push(field)?
-        } else if &field.name == META_FIELD_NAME {
+        } else if &field.name == META_FIELD_NAME || &field.name == "__typename" {
             meta_items.push(field)
         } else {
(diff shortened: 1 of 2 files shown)


dotansimha:
Cool! thanks @ardatan 

At the moment we don't have mutations, and querying for the root Subscriptions `__typename` is not allowed in the spec (enforced via validations). So this fix should cover all relevant use-cases.

dotansimha (approved):

Urigo:
@ardatan should we add a test for this?

ardatan:
@Urigo I added a test to check if `__typename` is correctly resolved for `Query`

saihaj (approved):

lutter (approved):
Nice!

PR #3197, opened 2022-01-27
Fix `graphman run` for the `PollingBlockStream`

otaviopace:
Running `graphman run` with the `PollingBlockStream` could lead to an infinite loop.

This can happen because `graphman run` keeps an [infinite loop](https://github.com/graphprotocol/graph-node/blob/399ee444d1023a95f82404d14bb5e4534d827678/node/src/manager/commands/run.rs#L189) checking if the `deployment_head` passed the `stop_block`, however the `instance_manager` would [stop processing](https://github.com/graphprotocol/graph-node/blob/399ee444d1023a95f82404d14bb5e4534d827678/core/src/subgraph/instance_manager.rs#L622) once the `PollingBlockStream` would get to the `stop_block` and not the `deployment_head`.

I've changed so that we keep the `deployment_head` state in the `instance_manager` and compare the `stop_block` to that.

I've also took the opportunity to remove the last store dependency from the Firehose \o/ 🎉

### DIFF
--- a/chain/ethereum/src/chain.rs
+++ b/chain/ethereum/src/chain.rs
@@ -1,6 +1,5 @@
 use anyhow::{Context, Error};
 use graph::blockchain::BlockchainKind;
-use graph::components::store::WritableStore;
 use graph::data::subgraph::UnifiedMappingApiVersion;
 use graph::env::env_var;
 use graph::firehose::{FirehoseEndpoints, ForkStep};
@@ -186,7 +185,7 @@ impl Blockchain for Chain {
     async fn new_firehose_block_stream(
         &self,
         deployment: DeploymentLocator,
-        writable: Arc<dyn WritableStore>,
+        block_cursor: Option<String>,
         start_blocks: Vec<BlockNumber>,
         filter: Arc<Self::TriggerFilter>,
         metrics: Arc<BlockStreamMetrics>,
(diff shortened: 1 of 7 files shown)


otaviopace:
The only ugly thing is that it ends with a panic ([see the code](https://github.com/graphprotocol/graph-node/blob/399ee444d1023a95f82404d14bb5e4534d827678/core/src/subgraph/instance_manager.rs#L615)) after `graphman run` logs finish:

スクリーンショット 2022-01-27 16 03 02

We can fix this in another PR or if someone has a quick fix for it I can do it here 🙂

leoyvens:
Nice refactors, but on the stop block logic itself, it looks like this isn't guaranteed to stop _exactly_ at the stop block, but at some block past it. Which seems ok to me, but it's good to document that in the graphman run command.

otaviopace:
@leoyvens I just pushed a better description/documentation of the command 🙂 

And added a [new commit](https://github.com/graphprotocol/graph-node/pull/3197/commits/8804d54b1a8f1b20601c08034cb2dbd72b85b71b) that fixes the race condition between the `graphman run` loop and the `instance_manager`.

sduchesneau:
> Nice refactors, but on the stop block logic itself, it looks like this isn't guaranteed to stop _exactly_ at the stop block, but at some block past it. Which seems ok to me, but it's good to document that in the graphman run command.

@leoyvens on a blockchain with skipped blocks (ex: solana, near), you may set a stop_block to 1000, but that block won't exist... 
such a blockchain may have blocks with number 998, then 1002... so we stop right before applying 1002

otaviopace:
Just to post here as well, the logs seem correct now, with no panic 😊 🎉 
スクリーンショット 2022-02-10 18 25 16

sduchesneau (approved):
LGTM!

leoyvens:
> we stop right before applying 1002

@sduchesneau that's not what this is doing, it applies 1002 and then stops.

PR #3198, opened 2022-01-28
Add `graphman` commands: `index list` and `index drop`

tilacog:
This PR adds two new `graphman** subcommands:

## `graphman index list <deployment-hash> <entity-name>`

Lists all indexes for the given Entity table.

Each index will be printed to `stdout` as its own definition command, like:

```
CREATE [UNIQUE] INDEX <index-name> ON <table-name> USING <index-method> (<columns>)
```

## `graphman index drop <deployment-hash> <index-name>`

Drops the given index, concurrently.

If the index does not exist, an error message will be displayed to the user.

### DIFF
--- a/node/src/bin/manager.rs
+++ b/node/src/bin/manager.rs
@@ -385,12 +385,21 @@ pub enum IndexCommand {
     ///
     /// This command may be time-consuming.
     Create {
-        /// The id of the deployment
+        /// The id of the deployment.
+        ///
+        /// Can be expressed either as its Qm-hash form or as its SQL schema identifier.
+        #[structopt(empty_values = false)]
         id: String,
-        /// The Entity name, in camel case.
+        /// The Entity name.
+        ///
+        /// Can be expressed either in upper camel case (as its GraphQL definition) or in snake case
+        /// (as its SQL table name).
         #[structopt(empty_values = false)]
         entity: String,
-        /// The Field names, in camel case.
+        /// The Field names.
+        ///
+        /// Each field can be expressed either in camel case (as its GraphQL definition) or in snake
+        /// case (as its SQL colmun name).
         #[structopt(min_values = 1, required = true)]
         fields: Vec<String>,
         /// The index method. Defaults to `btree`.
@@ -402,17 +411,24 @@ pub enum IndexCommand {
     },
     /// Lists existing indexes for a given Entity
     List {
-        /// The id of the deployment
+        /// The id of the deployment.
+        ///
+        /// Can be expressed either as its Qm-hash form or as its SQL schema identifier.
         #[structopt(empty_values = false)]
         id: String,
-        /// The Entity name, in camel case.
+        /// The Entity name.
+        ///
+        /// Can be expressed either in upper camel case (as its GraphQL definition) or in snake case
+        /// (as its SQL table name).
         #[structopt(empty_values = false)]
         entity: String,
     },
 
     /// Drops an index for a given deployment, concurrently
     Drop {
-        /// The id of the deployment
+        /// The id of the deployment.
+        ///
+        /// Can be expressed either as its Qm-hash form or as its SQL schema identifier.
         #[structopt(empty_values = false)]
         id: String,
         /// The name of the index to be dropped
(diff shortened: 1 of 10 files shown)


node/src/manager/commands/index.rs
```diff
@@ -41,3 +41,34 @@ pub async fn create(
...
     }
 }
+
+pub async fn list(
+    store: Arc<SubgraphStore>,
+    id: String,
+    entity_name: String,
+) -> Result<(), anyhow::Error> {
+    let deployment_hash = DeploymentHash::new(id)
+        .map_err(|e| anyhow::anyhow!("Subgraph hash must be a valid IPFS hash: {}", e))?;
```
otaviopace, line 51: No need to change, but you could simplify this with `.with_context` 😄

otaviopace (approved):
LGTM

node/src/manager/commands/index.rs
```diff
@@ -41,3 +41,34 @@ pub async fn create(
         Err(other) => Err(anyhow::anyhow!(other)),
     }
 }
+
+pub async fn list(
+    store: Arc<SubgraphStore>,
+    id: String,
+    entity_name: String,
+) -> Result<(), anyhow::Error> {
+    let deployment_hash = DeploymentHash::new(id)
```
lutter, line 50: It would be desirable to take a namespace `sgd1234` in all places where a deployment hash is allowed. A lot of times, those are easier to work with than deployment hashes.

node/src/bin/manager.rs
```diff
@@ -388,6 +388,25 @@ pub enum IndexCommand {
...
         method: String,
     },
+    /// Lists existing indexes for a given Entity
+    List {
+        /// The id of the deployment
+        #[structopt(empty_values = false)]
+        id: String,
+        /// The Entity name, in camel case.
+        #[structopt(empty_values = false)]
+        entity: String,
```
lutter, line 398: It would be nice if this allowed both entity type names and table names, so that I can say both (and mix and match)
```
graphman index list QmRfCDkbkACHdidsCKADjUDf6S3mMjS381xyzotJDsbqGq LiquidityPosition
``` 
and
```
graphman index list sgd1234 liquidity_position
```

For the deployment lookup, you can use `node::manager::Deployment::lookup` which can be resolved to a `DeploymentLocator` using `Deployment.locator()`

Translating the second argument could just try to look up the table assuming that `entity` is a type name, and if that fails look it up as the SQL name of a table.

node/src/manager/deployment.rs
```diff
@@ -173,3 +173,17 @@ pub fn locate(
...
+    name: &str,
+) -> anyhow::Result<DeploymentLocator> {
+    let deployment_locator = match &Deployment::lookup(pool, name)?[..] {
+        [] => anyhow::bail!("Found no deployment for the given ID"),
+        [deployment_locator] => deployment_locator,
+        _ => anyhow::bail!("Found multiplle deployments for given identifier"),
+    }
+    .locator();
+    Ok(deployment_locator)
+}
```
tilacog, line 189: This function is reused in the `stats analyze` and all `index` subcommands.

The reasoning behind the single return value is that I assumed that we would prefer to issue commands those commands for a specific deployment.

store/postgres/src/deployment_store.rs
```diff
@@ -1444,3 +1470,43 @@ impl DeploymentStore {
...
+                .or_else(|_error| {
+                    let sql_name = SqlName::from(f.as_ref());
+                    table
+                        .column(&sql_name)
+                        .ok_or_else(|| StoreError::UnknownField(f.as_ref().to_string()))
+                })
+                .map(|column| column.name.as_str())
+        })
+        .collect()
+}
```
tilacog, line 1512: I'm not sure if those two free functions should be methods of `Layout` and `Table,` respectively.

Both types use different signatures for lookups depending if the search term uses GraphQ or SQL notation:
- `Layout`:
  - https://github.com/graphprotocol/graph-node/blob/6eb80ad561a579183f902fccfb3cb1826f8855c5/store/postgres/src/relational.rs#L507
  - https://github.com/graphprotocol/graph-node/blob/6eb80ad561a579183f902fccfb3cb1826f8855c5/store/postgres/src/relational.rs#L514
  
- `Table`
  - https://github.com/graphprotocol/graph-node/blob/6eb80ad561a579183f902fccfb3cb1826f8855c5/store/postgres/src/relational.rs#L1182
  - https://github.com/graphprotocol/graph-node/blob/6eb80ad561a579183f902fccfb3cb1826f8855c5/store/postgres/src/relational.rs#L1194

If necessary, I can turn them into methods and, maybe, homogenize those returns values.

PR #3199, opened 2022-01-28
ethereum: Lower mapping gas cost of ethereum call

leoyvens:
This lowers the mapping gas cost of an eth call from 25 to 5 billion.

By benchmarking a uniswap-like subgraph, I've observed that the eth call gas cost causes the overall gas cost to be way overestimated. And 400 calls per mapping might just not be enough for some real world subgraphs.


That3Percent (approved):
I'm assuming you have a reason for changing this, so approving. But, do consider that we would like to discourage ethereum calls because of the overhead of RPC.

leoyvens:
Indeed, but this would be more like forbidding than discouraging, while we don't yet have good alternatives.

PR #3202, opened 2022-01-29
chain/near: added dynamic data source support

maoueh:
#### WIP

This cannot be merged as-is. While doing the work, I realized that how data source are stored (via `StoredDynamicDataSource) and queried is still chain specific and tied to Ethereum.

Some thinking and discussions will be required to decide what will be the best approach to make them chain agnostic and how we are going to have this properly handle backward compatibility for existing stored dynamic data source.

Right now in this PR, I "hijack" the current format and I'm able to make NEAR dynamic data source works.

### DIFF
--- a/chain/ethereum/src/data_source.rs
+++ b/chain/ethereum/src/data_source.rs
@@ -25,7 +25,7 @@ use graph::{
     },
 };
 
-use graph::data::subgraph::{calls_host_fn, DataSourceContext, Source};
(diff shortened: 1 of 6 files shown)


maoueh:
@mangas Can I close this now in regards to #3669?

azf20:
Filipe is out, but agree that this makes sense to close now, thanks!

PR #3210, opened 2022-02-01
First working implementation of Solana

colindickson:


chain/solana/src/chain.rs
```diff
@@ -0,0 +1,299 @@
...
+pub struct Chain {
+    logger_factory: LoggerFactory,
+    name: String,
+    firehose_endpoints: Arc<FirehoseEndpoints>,
+    chain_store: Arc<dyn ChainStore>,
+}
+
+impl std::fmt::Debug for Chain {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "chain: near")
```
maoueh, line 44: replacement needed

chain/solana/src/chain.rs
```diff
@@ -0,0 +1,299 @@
...
+        &self,
+        _deployment: DeploymentLocator,
+        _writable: Arc<dyn WritableStore>,
+        _start_blocks: Vec<BlockNumber>,
+        _subgraph_start_block: Option<BlockPtr>,
+        _filter: Arc<Self::TriggerFilter>,
+        _metrics: Arc<BlockStreamMetrics>,
+        _unified_api_version: UnifiedMappingApiVersion,
+    ) -> Result<Box<dyn BlockStream<Self>>, Error> {
+        panic!("SOLANA does not support polling block stream")
```
maoueh, line 142: It was `NEAR` in capital before because it's own it's spelled, but for `SOLANA` should be `Solana`.

chain/solana/src/chain.rs
```diff
@@ -0,0 +1,299 @@
...
+
+    fn is_firehose_supported(&self) -> bool {
+        true
+    }
+}
+
+pub struct TriggersAdapter {}
+
+#[async_trait]
+impl TriggersAdapterTrait<Chain> for TriggersAdapter {
```
maoueh, line 173: See new NEAR implementation to correctly set up those methods: https://github.com/graphprotocol/graph-node/blob/master/chain/near/src/chain.rs#L183

chain/solana/src/chain.rs
```diff
@@ -0,0 +1,299 @@
...
+
+    async fn triggers_in_block(
+        &self,
+        _logger: &Logger,
+        block: codec::Block,
+        _filter: &TriggerFilter,
+    ) -> Result<BlockWithTriggers<Chain>, Error> {
+        let shared_block = Arc::new(block.clone());
+        let instructions = block.transactions.iter().flat_map(|transaction| {
+            //let transaction_id = transaction.id.clone();
```
maoueh, line 201: Dead

chain/solana/src/chain.rs
```diff
@@ -0,0 +1,299 @@
...
+            let b = shared_block.clone();
+            let tx = transaction.clone();
```
maoueh, line 203: Why do we clone the full transaction but ultimately just use the `id` part of it?

chain/solana/src/chain.rs
```diff
@@ -0,0 +1,299 @@
...
+            transaction.instructions.iter().flat_map(move |instruction| {
+                Some(trigger::InstructionWithInfo {
+                    instruction: instruction.clone(),
```
maoueh, line 206: I did not check everything, but I found it a bit intense to have the need to fully clone the whole instruction, maybe we could keep only a shared reference to the data (it might not be possible due to how we use it, we will need to investigate that further).

chain/solana/src/chain.rs
```diff
@@ -0,0 +1,299 @@
...
+
+    async fn is_on_main_chain(&self, _ptr: BlockPtr) -> Result<bool, Error> {
+        // FIXME (Solana): Might not be necessary for Solana support for now
+        Ok(true)
+    }
+
+    /// Panics if `block` is genesis.
+    /// But that's ok since this is only called when reverting `block`.
+    async fn parent_ptr(&self, block: &BlockPtr) -> Result<Option<BlockPtr>, Error> {
+        // FIXME (NEAR):  Might not be necessary for NEAR support for now
```
maoueh, line 230: Needs to be implemented correctly now since used on reverts to go back in "time" properly, check NEAR implementation.

chain/solana/src/chain.rs
```diff
@@ -0,0 +1,299 @@
...
+                // let header = block.header();
+                // let parent_ptr = header
+                //     .parent_ptr()
+                //     .expect("Genesis block should never be reverted");
+                //todo: ^^^^^^^^^^^^^^^^^^^^^^^^^
+
+                Ok(BlockStreamEvent::Revert(
+                    block.ptr(),
+                    Some(response.cursor.clone()),
+                    None, //Some(parent_ptr), //todo: <----???
```
maoueh, line 286: Required now indeed.

chain/solana/src/chain.rs
```diff
@@ -0,0 +1,299 @@
...
+
+        use ForkStep::*;
+        match step {
+            StepNew => Ok(BlockStreamEvent::ProcessBlock(
+                adapter.triggers_in_block(logger, block, filter).await?,
+                Some(response.cursor.clone()),
+            )),
+
+            StepUndo => {
+                // let header = block.header();
```
maoueh, line 277: Dead code?

chain/solana/src/codec.rs
```diff
@@ -0,0 +1,52 @@
...
+        return None;
+        // if self.previous_id.len() == 0 {
+        //     return None;
+        // }
+        //
+        // Some(
+        //     BlockPtr::try_from((self.previous_id.as_ref(), self.number))
+        //         .expect("invalid block's hash"),
+        // )
+        //todo: ^^^^^
```
maoueh, line 20: Required now, please do.

chain/solana/src/codec.rs
```diff
@@ -0,0 +1,52 @@
...
+    }
+}
+
+impl From<Block> for BlockPtr {
```
maoueh, line 24: I removed those in NEAR to force people go by reference.

chain/solana/src/codec.rs
```diff
@@ -0,0 +1,52 @@
...
+    fn from(b: Block) -> BlockPtr {
+        (&b).into()
+    }
+}
+
+impl<'a> From<&'a Block> for BlockPtr {
+    fn from(b: &'a Block) -> BlockPtr {
+        // let hash = BlockHash::from(b.id)
```
maoueh, line 32: Dead

chain/solana/src/chain.rs
maoueh: This is even more true because those a later filtered, so we pay high copy price to discard 90% of them later. Minimally if we really need to copy, it would be much better to clone after filtering has been done.

chain/solana/src/data_source.rs
```diff
@@ -0,0 +1,406 @@
...
+
+        let handler = match trigger {
+            // A block trigger matches if a block handler is present.
+            SolanaTrigger::Block(_) => match self.handler_for_block() {
+                Some(handler) => &handler.handler,
+                None => return Ok(None),
+            },
+
+            SolanaTrigger::Instruction(instruction_with_block) => {
+                let pid = &instruction_with_block.instruction.program_id;
```
maoueh, line 90: `pid` when I read it later does not directly helps understanding, don't think it's a good acronym in this context here.

chain/solana/src/data_source.rs
```diff
@@ -0,0 +1,406 @@
...
+                let encoded_instruction_pid = pid.as_slice().to_base58();
```
maoueh, line 91: We should maybe think of doing the opposite, the actual manifest value which is a string should be decoded and stored as bytes so we would not pay the `base58` price on each single check here.

chain/solana/src/data_source.rs
```diff
@@ -0,0 +1,406 @@
...
+
+        DataSource::from_manifest(kind, network, name, source, mapping, context)
+    }
+}
+
+impl TryFrom<DataSourceTemplateInfo<Chain>> for DataSource {
+    type Error = Error;
+
+    fn try_from(_info: DataSourceTemplateInfo<Chain>) -> Result<Self, Error> {
+        Err(anyhow!("Near subgraphs do not support templates"))
```
maoueh, line 248: Message.

chain/solana/src/data_source.rs
```diff
@@ -0,0 +1,406 @@
...
+
+        // How this might be implemented if/when Near gets support for templates:
```
maoueh, line 250: I think we can remove it and later when we will need to implemented it, check NEAR implementation (which is now tracked in #3202

chain/solana/src/runtime/abi.rs
```diff
@@ -0,0 +1,439 @@
...
+impl ToAscObj<AscHashArray> for Vec<Vec<u8>> {
+    fn to_asc_obj<H: AscHeap + ?Sized>(
+        &self,
+        heap: &mut H,
+    ) -> Result<AscHashArray, DeterministicHostError> {
+        let content: Result<Vec<_>, _> = self
+            .iter()
+            .map(|x| {
+                asc_new(heap, x.as_slice())
+                // asc_new(heap, slice.as_ref())?
```
maoueh, line 263: Remove dead code in the file here.

chain/solana/src/runtime/abi.rs
```diff
@@ -0,0 +1,439 @@
...
+        account_changes: vec![account_change_1],
+        failed: false,
+        error: None,
+    };
+
+    let asc_instruction = asc_new(&mut heap, &instruction);
+    assert!(asc_instruction.is_ok());
+}
+
+struct BytesHeap {
```
maoueh, line 377: This is shared with NEAR now, we should move it to a common shared place so it can be re-used by both tests.

chain/solana/src/runtime/generated.rs
```diff
@@ -0,0 +1,306 @@
...
+    pub index: u64,
+    pub id: AscPtr<Uint8Array>,
+    pub additional_signatures: AscPtr<AscHashArray>,
+    pub header: AscPtr<AscMessageHeader>,
+    pub account_keys: AscPtr<AscHashArray>,
+    pub recent_blockhash: AscPtr<Uint8Array>,
+    pub log_messages: AscPtr<AscStringArray>,
+    pub instructions: AscPtr<AscInstructionArray>,
+    pub error: AscPtr<AscTransactionError>,
+    pub failed: bool,
```
maoueh, line 126: Double check ordering of fields in `graph-ts`, in the version of the branch I have on `graph-ts`, those are ordered differently as far as I remember (`error` is after `failed`)

chain/solana/src/trigger.rs
```diff
@@ -0,0 +1,482 @@
...
+}
+
+impl InstructionWithInfo {
+    pub fn number(&self) -> BlockNumber {
+        let num = self.block_num as i32;
+        num.into()
+    }
+}
+
+// #[cfg(test)]
```
maoueh, line 165: I think we can delete those tests.

graph/src/firehose/endpoints.rs
```diff
@@ -92,10 +92,12 @@ impl FirehoseEndpoint {
             },
         );
 
-        debug!(logger, "Connecting to firehose to retrieve genesis block");
+        debug!(logger, "Connecting to firehose to retrieve genesis block, start block 118081400, stop block 119081400");
+
         let response_stream = client
             .blocks(firehose::Request {
-                start_block_num: 0,
+                start_block_num: 118081400,
```
maoueh, line 99: Seems wrong, left over probably? If we ever need to limit Solana range, it should not be performed here anyway.

graph/src/runtime/mod.rs
```diff
@@ -240,6 +240,21 @@ pub enum IndexForAscTypeId {
     NearChunkHeader = 84,
     NearBlock = 85,
     NearReceiptWithOutcome = 86,
+    SolanaBlock = 87,
```
maoueh, line 243: Need to be aligned with `graph-ts` code base. I think we should start to have dedicated "range" per chain. Solana could be 200 - 399, Tendermint 400 - 699 (etc.) to ease future addition. Maybe there is restrictions of doing that.

node/src/main.rs
```diff
@@ -544,6 +571,54 @@ fn near_networks_as_chains(
     HashMap::from_iter(chains)
 }
 
+/// Return the hashmap of SOLANA chains and also add them to `blockchain_map`.
+fn solana_networks_as_chains(
```
maoueh, line 575: I think this have been moved to shared location in `master`.

maoueh (changes requested):

chain/solana/src/chain.rs
colindickson: we can check together if there is a better way.

chain/solana/src/chain.rs
colindickson: this is the same implementation as on NEAR.  unless there's a different implementation on another branch?

maoueh: No I mixed it up and thought it was another one, needs to be a `panic!("Should never be called since FirehoseBlockStream cannot resolve it")` (I'll need to update NEAR version later also)

chain/solana/src/runtime/generated.rs
colindickson: ordering is the same on the latest commit in graph-ts

chain/solana/src/runtime/generated.rs
```diff
@@ -0,0 +1,306 @@
...
+#[derive(AscType)]
+pub(crate) struct AscInstruction {
+    pub ordinal: u32,
+    pub parent_ordinal: u32,
+    pub depth: u32,
+    pub program_id: AscPtr<AscHash>,
+    pub account_keys: AscPtr<AscHashArray>,
+    pub data: AscPtr<AscHash>,
+    pub balance_changes: AscPtr<AscBalanceChangeArray>,
+    pub account_changes: AscPtr<AscAccountChangeArray>,
```
abourget, line 42: Let's not push that out at the moment. Let's figure if we can add the `logs` in here.

graph/src/runtime/mod.rs
leoyvens: Agree we should use ranges

maoueh:
Next steps around this PR and Solana integration in general:

- Rebased this PR on top of master
- Update protobuf definition using latest one (https://github.com/streamingfast/proto-solana/blob/develop/sf/solana/codec/v1/codec.proto)
- Remove `account_changes` up to the `graph-ts` bindings and handler because we will get rid of it for version 1
- Update `block_pointer_from_number` (like https://github.com/graphprotocol/graph-node/blob/d0d0cc1da0ca95398536cacf4bd47bf84031a2f7/chain/near/src/chain.rs#L200)
- Start block ingestor like other chain (https://github.com/graphprotocol/graph-node/blob/60cd12ce7dd8ef796f8c8e93002cc7624fd6346c/node/src/main.rs#L348)
  - Implement header only transform and support in `graph-node` (Firehose work will be required to add a specialized transform that returns a `Block` with `transactions`, `transaction_count`, `has_split_account_changes` and `account_changes_file_ref` fields removed).
- Add `TriggerFilter::InstructionFilter` where user's can match specific instructions based on criterion
  - Filter based on account key at specific position in the instruction' account inputs (`- accountInput { atIndex: 0, onKey: 3AWPyBF4CcoFY3zsxAbc3PRTA8JK227FSsfkYVYVFuxJ})`
  - Filter based on first byte of instruction's data, which enables matching by instruction name (add Anchor support here to accepts a name from the IDL definition and auto-map it to instruction's ID?)
  - Review filters to be sure they are correct for initial Solana support
- Improve `triggers_in_block` implementation, instructions should be filtered upfront to triggers, remove as much `clone()` as possible, structures are big and everywhere cloning we can removed, the better `graph-node` will love us
- Implement Firehose Filtering support in `graph-node` (StreamingFast will be doing the adjustment to Firehose to support that, based on the filters reviewed and approved)
- Instruction tree serialization de-risking (see [Risk: Instructions Tree Serialization](#risk:-instructions-tree-serialization))
- Create initial Solana subgraph which would collect SPL token holders (I think it's achievable)
- Deploy to staging

### Risk: Instructions Tree Serialization

We think having contextual information about other instructions in the tree of instructions rooted at the matched instruction is a requirement. Indeed, we expect that the transaction matched by manifest will actually perform other important child instructions that dapps developers will require access to. For example, performing an NFT mint on Solana generates a bunch of child instructions that is important to have access to, like the SPL token created. 

We think that serializing the whole tree to the WASM handler will give a good DX to dapp developer. However, it's a risk because the instruction tree could be quite big in term of AssemblyScript serialized bytes and the time it could take to do it.

We should explore this question as soon as possible to understand the consequences of fully serializing the full instructions tree always. Other possibilities:

- Serialize the tree only on demand (an intrinsics would perform the work only when the subgraph calls it)
- Serialize specific child instruction of the tree only on demand (an intrinsics would perform the work only when the subgraph calls it), raises questions about how we "index" them and user would "find" them
- De-contextualize and send specific instructions on different handler, so we would keep the one instruction == one handler paradigm, raises questions about how this would look like in the manifest

azf20:
Thanks @maoueh (& @jubeless) for this update.

> add Anchor support here to accepts a name from the IDL definition

How complex is this proposal, and what would the Developer Experience look like (would this require changes to upload the IDL as part of the manifest?)

> Create initial Solana subgraph

Is this initial subgraph definition quite well understood & is any further research required here?

> Risk: Instructions Tree Serialization

Is the primary concern here feasibility (i.e. will it be too big in terms of AssemblyScript serialized bytes) or performance (regardless, the serialisation will be too slow)?

Seems like there are still a few outstanding pieces on both the Firehose & Graph Node side here?

maoueh:
> How complex is this proposal, and what would the Developer Experience look like (would this require changes to upload the IDL as part of the manifest?)

It should have expanded further saying indeed this is a risk factor if we decide to have right now.

So good question, we checked the JSON output format, seems rather "easy" to work it. Now the question is how easy "reading" the data is. For the actual account input, I think it will be quite easy actually to do something.

After that, it's a matter to have low level decoding of each Anchor base types. We are not sure which encoding Anchor uses, it's own, Borsh or default Rust C encoding that was used before, would need to check that. I don't think it's utterly complex to have a working version, need the man power to dedicate at least a week to that specific task.

I envision generate the struct types from the IDL and also the decoding code directly, would be all in AssemblyScript.

> Is this initial subgraph definition quite well understood & is any further research required here?

Would be good to sync on that to be sure we are on the same page, Julien told me you discussed with him maybe SPL Token Holders, if it's the case, we already thought about how to have the right "matching" to get the correct one. I will try to sit with Julien and together we will write the manifest and "pseudo-code" for this subgraph.

> Is the primary concern here feasibility (i.e. will it be too big in terms of AssemblyScript serialized bytes) or performance (regardless, the serialisation will be too slow)?

Both in fact, I think it's really something that needs to be checked first, how long does it take mainly and also how many bytes it consumes, will give indication of how many subgraph we would be able to run on N GB of RAM (worst case/average case something around that).

> Seems like there are still a few outstanding pieces on both the Firehose & Graph Node side here?

Yep there is work on both side, right now we are spinning up the whole Firehose infrastructure and ensuring it works with how big the blocks are (even with account data removed).

For filtering, I don't think it will take much time, we have everything to easily create filter(s) and indexes, we should need to decide about the filters we need to support, so the review of the filters available in the manifest is a prior task to have them in Firehose.

halink0803:
Hi, while looking for a way to use subgraph with Solana, I have pumped to this PR. How is the PR going on? I saw there is no update since Feb, and the last comment is at April. Will this PR still work or stale?

I also saw from `chain` folder, there is a source `substreams` is merged? Will it works as source for `solana` as I thought `substreams` already support for `solana`.

maoueh:
So this branch is stalled I'm not 100% sure how it will behave. This was before `substreams` was a thing.

With `substreams` now and how it's currently integrated within `graph-node`, any chain that was `substreams` will be able to integrate directly. Let's continue in Discord with you have further questions.

halink0803:
@maoueh thank you. I just pump into this and wonder about the status of it. If `substreams` is the alternative, I think this PR could be closed.

leoyvens:
Since the focus is currently on substreams support for Solana, I'm closing this for now.

PR #3212, opened 2022-02-01
Tendermint integration

pienkowb:


chain/tendermint/proto/codec.proto
```diff
@@ -0,0 +1,280 @@
+syntax = "proto3";
```
otaviopace, line 1: If possible can you please put a comment about where this was coppied from?

chain/tendermint/proto/codec.proto
```diff
@@ -0,0 +1,280 @@
...
+
+package fig.tendermint.codec.v1;
+
+option go_package = "github.com/figment-networks/tendermint-protobuf-def/codec";
+
+import "gogo.proto";
+import "google/protobuf/descriptor.proto";
```
otaviopace, line 8: Do you know if these dependencies are needed?

chain/tendermint/proto/gogo.proto
```diff
@@ -0,0 +1,147 @@
...
+	optional bool unsafe_marshaler_all = 63023;
+	optional bool unsafe_unmarshaler_all = 63024;
+
+	optional bool goproto_extensions_map_all = 63025;
+	optional bool goproto_unrecognized_all = 63026;
+	optional bool gogoproto_import = 63027;
+	optional bool protosizer_all = 63028;
+	optional bool compare_all = 63029;
+    optional bool typedecl_all = 63030;
+    optional bool enumdecl_all = 63031;
```
otaviopace, line 85: I don't know if this was autogenerated as well, but the indentation is incorrect.

chain/tendermint/src/adapter.rs
```diff
@@ -0,0 +1,42 @@
...
+
+#[derive(Clone, Debug, Default)]
+pub(crate) struct TendermintBlockFilter {
+    pub trigger_every_block: bool,
+}
+
+impl TendermintBlockFilter {
+    pub fn from_data_sources<'a>(iter: impl IntoIterator<Item = &'a DataSource>) -> Self {
+        iter.into_iter()
+            .filter(|data_source| data_source.source.address.is_some())
```
otaviopace, line 30: We can remove the `address` since Tendermint will not use it 🙂

chain/tendermint/src/capabilities.rs
```diff
@@ -0,0 +1,39 @@
...
+
+use anyhow::Error;
+use graph::impl_slog_value;
+
+use crate::DataSource;
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct NodeCapabilities {}
+
+impl PartialOrd for NodeCapabilities {
```
otaviopace, line 13: Probably can be derived like `PartialEq`, so added to line `10`.

chain/tendermint/proto/codec.proto
```diff
@@ -0,0 +1,280 @@
...
+  option (gogoproto.goproto_enum_prefix)    = false;
+
+  BLOCK_ID_FLAG_UNKNOWN = 0 [(gogoproto.enumvalue_customname) = "BlockIDFlagUnknown"];
+  BLOCK_ID_FLAG_ABSENT  = 1 [(gogoproto.enumvalue_customname) = "BlockIDFlagAbsent"];
+  BLOCK_ID_FLAG_COMMIT  = 2 [(gogoproto.enumvalue_customname) = "BlockIDFlagCommit"];
+  BLOCK_ID_FLAG_NIL     = 3 [(gogoproto.enumvalue_customname) = "BlockIDFlagNil"];
+}
+
+message EventList {
+  EventBlock                new_block             = 1;
```
otaviopace, line 35: Why does it have the `new_` prefix? Maybe it's a cosmos thing 🤔

chain/tendermint/src/chain.rs
```diff
@@ -0,0 +1,295 @@
...
+    ) -> Result<BlockWithTriggers<Chain>, Error> {
+        let shared_block = Arc::new(block.clone());
+
+        let mut triggers: Vec<_> = shared_block
+            .events()
+            .into_iter()
+            .map(|event| {
+                TendermintTrigger::Event(Arc::new(codec::EventData {
+                    event: Some(event),
+                    block: Some(block.block().clone()),
```
otaviopace, line 204: I think it's useful to add a comment here with a `FIXME` for optimizing this part of the code at some point.

Right now it's isn't possible to use an `Arc` here because we're using types that are automatically generated.

chain/tendermint/src/chain.rs
```diff
@@ -0,0 +1,295 @@
...
+        });
+
+        let any_block = response
+            .block
+            .as_ref()
+            .expect("block payload information should always be present");
+
+        // Right now, this is done in all cases but in reality, with how the BlockStreamEvent::Revert
+        // is defined right now, only block hash and block number is necessary. However, this information
+        // is not part of the actual bstream::BlockResponseV2 payload. As such, we need to decode the full
```
otaviopace, line 261: @maoueh could this boilerplate be extracted somewhere else for every Firehose chain.

chain/tendermint/src/codec.rs
```diff
@@ -0,0 +1,149 @@
...
+    pub fn header(&self) -> &Header {
+        self.block().header()
+    }
+
+    pub fn events(&self) -> Vec<Event> {
+        self.begin_block_events()
+            .into_iter()
+            .chain(self.tx_events().into_iter())
+            .chain(self.end_block_events().into_iter())
+            .collect()
```
otaviopace, line 29: Here all of the new `Vec` allocations could be turned into `Iterator` ones.

So you don't have to create three vectors and destroy them to create a fourth one. You can just iterate over three and allocate/create one vector.

server/index-node/src/resolver.rs
```diff
@@ -100,9 +100,7 @@ where
 
         let block_hash = field
             .get_required::<H256>("blockHash")
-            .expect("Valid blockHash required")
-            .try_into()
-            .unwrap();
```
otaviopace, line 105: Dumb question: I didn't understand how those two lines can be removed? 🤔

graph/src/blockchain/block_stream.rs
```diff
@@ -165,7 +165,7 @@ pub trait FirehoseMapper<C: Blockchain>: Send + Sync {
 #[derive(Error, Debug)]
 pub enum FirehoseError {
     /// We were unable to decode the received block payload into the chain specific Block struct (e.g. chain_ethereum::pb::Block)
-    #[error("received gRPC block payload cannot be decoded")]
+    #[error("received gRPC block payload cannot be decoded: {0}")]
     DecodingError(#[from] prost::DecodeError),
```
otaviopace, line 169: Where is this being used? I can't find on the codebase when looking for `DecodingError`. I'm curious what will be passed on `{0}` 👀

graph/Cargo.toml
```diff
@@ -55,6 +55,7 @@ wasmparser = "0.78.2"
 thiserror = "1.0.25"
 parking_lot = "0.11.2"
 itertools = "0.10.3"
+subtle-encoding = {version = "0.5", features = ["bech32-preview"]}
```
otaviopace, line 58: Where is this being used?

chain/tendermint/src/data_source.rs
```diff
@@ -0,0 +1,368 @@
...
+    pub name: String,
+    pub source: Source,
+    pub mapping: Mapping,
+    pub context: Arc<Option<DataSourceContext>>,
+    pub creation_block: Option<BlockNumber>,
+}
+
+impl blockchain::DataSource<Chain> for DataSource {
+    fn address(&self) -> Option<&[u8]> {
+        self.source.address.as_ref().map(|x| x.as_bytes())
```
otaviopace, line 38: As we chatted before, `address` can be removed 🙂

chain/tendermint/src/runtime/abi.rs
```diff
@@ -0,0 +1,722 @@
...
+use graph::runtime::{
+    asc_new, AscHeap, AscIndexId, AscPtr, AscType, DeterministicHostError, ToAscObj,
+};
+use graph_runtime_wasm::asc_abi::class::{Array, Uint8Array};
+
+use crate::codec;
+
+pub(crate) use super::generated::*;
+
+impl ToAscObj<AscEventList> for codec::EventList {
```
otaviopace, line 14: I feel bad that you had to write all of this 😓 

This should be automatically generated. I'll fix this soon 🙏

chain/tendermint/src/runtime/generated.rs
```diff
@@ -0,0 +1,693 @@
...
+    }
+}
+
+impl AscIndexId for AscSignedMsgTypeEnum {
+    const INDEX_ASC_TYPE_ID: IndexForAscTypeId = IndexForAscTypeId::TendermintSignedMsgTypeEnum;
+}
+
+#[repr(u32)]
+#[derive(AscType, Copy, Clone)]
+pub(crate) enum AscSignedMsgType {
```
otaviopace, line 204: I wonder if those should have the same values as the ones described in `codec` and `graph-ts`, like `0`, `1`, `2` and `32` respectively.

chain/tendermint/src/protobuf/gogoproto.rs
```diff
@@ -0,0 +1 @@
+
```
tilacog, line 1: As we discussed, we should add this file and `chain/tendermint/src/protobuf/google.protobuf.rs` into `.gitignore`

tilacog (changes requested):

chain/tendermint/src/runtime/generated.rs
```diff
@@ -0,0 +1,693 @@
...
+}
+
+#[repr(C)]
+#[derive(AscType)]
+pub(crate) struct AscCommit {
+    pub height: i64,
+    pub round: i32,
+    pub block_id: AscPtr<AscBlockID>,
+    pub signatures: AscPtr<AscCommitSigArray>,
+    pub _padding: u32,
```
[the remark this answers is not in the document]
pienkowb: > Did you put the padding because you faced memory alignment issues when testing with Tendermint?

That's right. It would be great if we could get rid of this 👍

chain/tendermint/src/runtime/generated.rs
```diff
@@ -0,0 +1,693 @@
+use graph::runtime::{
```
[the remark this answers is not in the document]
pienkowb: We used a tool created by StreamingFast: https://github.com/streamingfast/graph-as-to-rust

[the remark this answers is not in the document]
maoueh: I would need to take some time to refactor it a bit to better handle Enum and other small edge cases.

chain/tendermint/src/runtime/generated.rs
```diff
@@ -0,0 +1,693 @@
...
+pub(crate) enum AscBlockIDFlag {
+    BlockIdFlagUnknown,
+    BlockIdFlagAbsent,
+    BlockIdFlagCommit,
+    BlockIdFlagNil,
+}
+
+impl AscValue for AscBlockIDFlag {}
+
+impl Default for AscBlockIDFlag {
```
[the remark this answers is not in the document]
maoueh: That's an edge case that I need to update the code generator for, enum and their "variant" is not handled correctly, for NEAR I tweaked by hand and did not update the tool.

Because there is a distinction to be made between pure enum and variant enum and the code generator needs to be adapter to account for those.

chain/tendermint/build.rs
```diff
@@ -0,0 +1,8 @@
+fn main() {
+    println!("cargo:rerun-if-changed=proto");
+    tonic_build::configure()
+        .out_dir("src/protobuf")
+        .format(true)
+        .compile(&["proto/codec.proto"], &["proto"])
+        .expect("Failed to compile Figment's Tendermint proto(s)");
```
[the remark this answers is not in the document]
maoueh: That's because we used `StreamingFast` in same file of other chains. Let's settle on using `Firehose` instead, I'll fix the existing one also.

chain/tendermint/src/adapter.rs
```diff
@@ -0,0 +1,42 @@
...
+            .extend(TendermintBlockFilter::from_data_sources(data_sources));
+    }
+
+    fn node_capabilities(&self) -> NodeCapabilities {
+        NodeCapabilities {}
+    }
+}
+
+#[derive(Clone, Debug, Default)]
+pub(crate) struct TendermintBlockFilter {
```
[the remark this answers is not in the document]
roignpar: `TendermintBlockFilter` is used by `TriggerFilter` (defined in the same module) which is used in the tendermint `Blockchain` impl. would the `TriggerFilter` just be empty then?

chain/tendermint/src/runtime/generated.rs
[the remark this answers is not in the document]
pienkowb: @maoueh Is it possible to support enums with gaps in values? There's a bunch of them in the official Tendermint protos (e.g. [this one](https://github.com/tendermint/tendermint/blob/master/proto/tendermint/types/types.proto#L34-L39)) and right now we're getting this error:
```
FAILURE Enum value SignedMsgType#SIGNED_MSG_TYPE_PROPOSAL does not monotically increase starting at 0
```

server/index-node/src/resolver.rs
roignpar: I'm not sure what `try_into` turned the `H256` into, but `BlockPtr` has a `From<(H256, u64)>` impl.
`block_hash` is used to create `block` on the next line: `BlockPtr::from((block_hash, block_number))`

graph/src/blockchain/block_stream.rs
roignpar: Deriving `thiserror::Error` allows using (unnamed) fields in the generated `Display` impl.
In this case `{0}` would be replaced by the `prost::DecodeError` `fmt` output (from its `Display` impl).
Some examples in `thiserror` docs: https://docs.rs/thiserror/latest/thiserror/

graph/Cargo.toml
roignpar: it's not. will remove.

chain/tendermint/src/runtime/abi.rs
```diff
@@ -0,0 +1,722 @@
...
+    fn to_asc_obj<H: AscHeap + ?Sized>(
+        &self,
+        heap: &mut H,
+    ) -> Result<AscHash, DeterministicHostError> {
+        self.0.to_asc_obj(heap)
+    }
+}
+
+/// Map an optional object to its Asc equivalent if Some, otherwise return null.
+fn asc_new_or_null<H, O, A>(
```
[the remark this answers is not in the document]
roignpar: `prost` wraps non-repeated user defined field types in `Option`.
`asc_new_or_null` simply returns `AscPtr::null()` and is used when the fields are really optional.
by contrast, `asc_new_or_missing` returns an error if the field is `None` and is used for fields that should have a value, even if their type is `Option<...>`.

chain/tendermint/src/data_source.rs
```diff
@@ -0,0 +1,368 @@
...
+    }
+
+    fn api_version(&self) -> semver::Version {
+        self.mapping.api_version.clone()
+    }
+
+    fn runtime(&self) -> &[u8] {
+        self.mapping.runtime.as_ref()
+    }
+}
```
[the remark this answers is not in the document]
roignpar: As a byproduct of using a custom `Source` struct in order to remove the `abi` and `address` fields from tendermint data sources I intend to implement the conversion methods (`as_stored_dynamic_data_source` and `from_stored_dynamic_data_source`) using `None` for `address` and an empty string for `abi`.
Should I then still remove the `DataSourceTemplate` support?

chain/tendermint/src/trigger.rs
```diff
@@ -0,0 +1,125 @@
...
+
+impl PartialEq for TendermintTrigger {
+    fn eq(&self, other: &Self) -> bool {
+        match (self, other) {
+            (Self::Block(a_ptr), Self::Block(b_ptr)) => a_ptr == b_ptr,
+            (Self::Event(a), Self::Event(b)) => a.event().event_type == b.event().event_type,
+            _ => false,
+        }
+    }
+}
```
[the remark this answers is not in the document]
roignpar: i think it's better to leave this in since the second branch (the `Event` comparison) only checks the `event_type` and no other fields.

chain/tendermint/src/trigger.rs
```diff
@@ -0,0 +1,125 @@
...
+            (Self::Event(..), Self::Event(..)) => Ordering::Equal,
+        }
+    }
+}
+
+impl PartialOrd for TendermintTrigger {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
```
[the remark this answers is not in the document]
roignpar: it's not possible because the `TendermintTrigger` variant fields are generated by `prost` and don't implement `PartialOrd`, so they can't be compared. we need to rely on the `Ord` implementation in this case.

graph/src/blockchain/block_stream.rs
otaviopace: I understand `thiserror` library usage, I mean in our codebase where this variant is used with a parameter.

chain/tendermint/src/adapter.rs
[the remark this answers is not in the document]
otaviopace: I think so, can you confirm @leoyvens ?

otaviopace:
@patiee @pienkowb I remember that you folks mentioned that you might be able to use the direct `.proto` files from the official `tendermint` repo, did that workout or we're going with the current approach?

If we're keeping this one, I've answered all comments and resolved the ones that nothing should be done about 🙂

pienkowb:
> @patiee @pienkowb I remember that you folks mentioned that you might be able to use the direct `.proto` files from the official `tendermint` repo, did that workout or we're going with the current approach?

We're still experimenting with importing the official `.proto` files, we should reach a conclusion this week. I'll let you know!

chain/tendermint/proto/codec.proto
lavabyrd: Just on this, `gogo.proto` is required, the `descriptor` isn't so it'll be removed

chain/tendermint/proto/gogo.proto
lavabyrd: not sure how this one broke but I've fixed the indentation on it

chain/tendermint/src/adapter.rs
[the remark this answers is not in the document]
leoyvens: Yes if this isn't actually used `TriggerFilter` can just be empty.

chain/tendermint/src/data_source.rs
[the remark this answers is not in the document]
leoyvens: If it's not supported then when should have as little code as possible around templates.

pienkowb:
@otaviopace I'm checking it. BTW, We might want to clean up the commit history before merging.

otaviopace:
We can just squash them 🙂

pienkowb:
@otaviopace The tests are passing now :)

+1 for squashing the commits into one.

Issue #3203, opened 2022-01-29
refactor: prefer GraphQL SDL for base/meta schema


dotansimha:
@lutter thoughts on that? mainly around `api.rs` and the schema we are building there. I feel like using SDL might be simpler for most of the static definitions? 

Also related:
https://github.com/graphprotocol/graph-node/pull/3201
https://github.com/graphprotocol/graph-node/pull/3184

lutter:
Yes, my preference would be to express as much as plain text files (GraphQL SDL) as possible instead of doing that in code. Of course, some of that has to be dynamic since it depends on user-supplied types.

PR #3213, opened 2022-02-02
graphql: refactor and prefer static GraphQL where possible

dotansimha:
Closes https://github.com/graphprotocol/graph-node/issues/3203 

- [x] Move static GraphQL to be in `meta.graphql`
- [x] Added missing `@skip` and `@include` direcrives to the type definitions. 
- [x] Added descriptions to all static type defs.

### DIFF
--- a/graphql/src/schema/meta.graphql
+++ b/graphql/src/schema/meta.graphql
@@ -1,5 +1,27 @@
+# GraphQL core functionality
+scalar Boolean
+scalar ID
+scalar Int
+scalar Float
+scalar String
+
 directive @skip(if: Boolean!) on FIELD | FRAGMENT_SPREAD | INLINE_FRAGMENT
 directive @include(if: Boolean!) on FIELD | FRAGMENT_SPREAD | INLINE_FRAGMENT
+
+# The Graph extensions
+
+"Marks the GraphQL type as indexable entity.  Each type that should be an entity is required to be annotated with this directive."
+directive @entity on OBJECT
+
+"Defined a Subgraph ID for an object type"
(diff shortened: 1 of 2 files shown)


dotansimha:
@lutter rebased this one as well, should be quick and easy review ;)

graphql/src/schema/meta.graphql
```diff
@@ -1,6 +1,28 @@
+# GraphQL core functionality
+scalar Boolean
+scalar ID
+scalar Int
+scalar Float
```
lutter, line 5: We don't have a `Float` type - not sure what will happen if somebody tries to use that, but I'd suspect that it will blow up in some way.

lutter (approved):
Just remove the `Float` type, and this is good to go.

graphql/src/schema/meta.graphql
dotansimha: Thanks for the review @lutter ! 
We are using it as part of the introspection schema (https://github.com/graphprotocol/graph-node/blob/393d0dca8c25a184f140da7cc62596a26d4580e5/graphql/src/introspection/schema.rs#L17), and it's also a built-in in GraphQL, so are you sure it's a good direction to drop it?

dotansimha: I rebased and pushed a fix for it, but not merging yet, waiting for your decision on that. 
Thanks!

lutter: I just checked the code, and we actually do handle it - you are right, we should leave the `Float` in then.

dotansimha: Fixed, thanks! will merge when CI clears

Issue #613, opened 2018-11-29
Support scalable pagination


Arachnid:
Presently, it's possible to query entities using a `where` clause, but this uses offsets from start or end, which likely won't scale well if paging over a large dataset. It'd be good to use the graphql `connection` pattern, or something similar, where result sets return an opaque cursor that can be passed in on subsequent calls to pick up where the previous query left off.

leodasvacas:
Settings offsets is not the most convenient API and we need good pagination support, connections seem like a good model to follow.

> offsets from start or end, which likely won't scale well if paging over a large dataset

Could you elaborate on what's the issue you're envisioning here?

Arachnid:
> Could you elaborate on what's the issue you're envisioning here?

In most database systems, a  query like `SELECT * FROM table LIMIT x OFFSET y` involves the database internally iterating over and discarding the first y results. This results in the cost of paginating over a large dataset being `O(n^2)` instead of `O(n)`. Using cursors, in contrast, doesn't suffer from this issue.

leodasvacas:
@Arachnid I see. Though that seems to be more a concern of implementation than of graphql interface. We could do a good implementation of graphql offsets that doesn't use `OFFSET`, and it's also possible to do a bad implementation of cursors that does use `OFFSET` on the DB.

Arachnid:
True, but I don't think it's possible (at least without low level DB
support) to do a good implementation that uses offsets - so better to fix
the API early.

-Nick

fubhy:
I wrote a little utility hook that takes care of automatically scraping the endpoint for more results (using skip & limit parameters) until it's exhausted:

```ts
import { useQuery } from '@apollo/react-hooks';
import { useRef, useEffect } from 'react';
import { DocumentNode } from 'graphql';

type QueryPair = [DocumentNode, DocumentNode?];

type ProceedOrNotFn = (result: any, expected: number) => boolean;

export function useScrapingQuery([query, more = query]: QueryPair, proceed: ProceedOrNotFn, props?: any) {
  const limit = (props.variables && props.variables.limit) || 100;
  const skip = useRef((props.variables && props.variables.skip) || 0);
  const result = useQuery(query, {
    ...props,
    variables: {
      ...(props && props.variables),
      limit,
      skip,
    },
  });

  useEffect(() => {
    if (!!result.loading || !proceed(result.data, skip.current + limit)) {
      return;
    }

    result.fetchMore({
      query: more,
      variables: {
[...]

PR #3214, opened 2022-02-02
feat: relay compliant cursor pagination

dotansimha:
Related: https://github.com/graphprotocol/graph-node/issues/613

### Background

This PR introduces a non-breaking change, that adds `xyzConnection` filed on a `Query`, and the matching `XyzConnection` type for easy pagination. 

Related: 
- https://relay.dev/graphql/connections.htm
- https://graphql.org/learn/pagination/

For a given input schema: 

```graphql
type Purpose @entity {
  # ...
}
```

We'll get now: 

```graphql
type Query {
  # ...
  purposesConnection(
    first: Int
    last: Int
    after: String
    before: String
    orderBy: Purpose_orderBy
    orderDirection: OrderDirection
    where: Purpose_filter
    subgraphError: _SubgraphErrorPolicy_ = deny
  ): PurposeConnection!
}

type PurposeConnection {
  edges: [PurposeEdge!]!
  pageInfo: PageInfo!
}

type PurposeEdge {
  cursor: String!
  node: Purpose!
}

type PageInfo {
  hasNextPage: Boolean!
  hasPreviousPage: Boolean!
  startCursor: String
  endCursor: String
}
```

### TODO 
- [x] Implement GraphQL schema generation. 
- [x] Cursor calculation
- [x] Forward Pagination 
- [x] Backward Pagination
- [x] When invalid cursor is provided in where clause then throw an error
- [x] `hasNextPage` computation with SQL
- [ ] `hasPreviousPage` computation with SQL 
- [ ] Fix selection of nested entities, right now pagination just breaks if there is a child entity selected.
- [ ] Improvement: hash the `where` clause during `parse` and then use that to encode it. This way when comparing during `after` cursor creation it will be better than encoding and comparing a large string.
- [ ] OrderBy support: Right now the `after/before` cursor computation disregards the OrderBy from the user's query. This means we need to re-write the `after/before` to use a `Bytes` like comparison but this time it also accounts for `order by` in the SQL generation
- [ ] Figure out where the extra `\"` are coming from when encoding `base64` 
- [ ] Improvement use `&str` instead of `String`
- [ ]  `Node` interface (we will likely do this as a follow, since it is not necessary and requires more refactor)


lutter:
My one concern with the proposed schema is the `totalCount` field on `PurposeConnection` since that can be _very_ expensive to calculate. I think it would be better to leave that out for the first cut of this feature.

dotansimha:
> My one concern with the proposed schema is the `totalCount` field on `PurposeConnection` since that can be _very_ expensive to calculate. I think it would be better to leave that out for the first cut of this feature.

Sure, we can support only parts of this spec, thanks for the heads-up!

graphql/src/store/prefetch.rs
```diff
@@ -635,7 +635,9 @@ fn execute_field(
     field_definition: &s::Field,
     selected_attrs: SelectedAttributes,
 ) -> Result<(Vec<Node>, Trace), Vec<QueryExecutionError>> {
-    let multiplicity = if sast::is_list_or_non_null_list_field(field_definition) {
+    let multiplicity = if sast::is_list_or_non_null_list_field(field_definition)
+        || is_connection_type(&field_definition.field_type.get_base_type().to_string())
+    {
```
lutter, line 687: Why do we need that?

graphql/src/store/query.rs
```diff
@@ -68,6 +93,123 @@ pub(crate) fn build_query<'a>(
...
+                    "last", max_skip, n,
+                ));
+            }
+        }
+        Some(r::Value::Null) | None => 0,
+        _ => unreachable!("last is an Int with a default value"),
+    };
+
+    let after = match field.argument_value("after") {
+        Some(r::Value::String(s)) => Some(s.clone()),
```
lutter, line 137: Is this clone needed?

graphql/src/store/query.rs
```diff
@@ -68,6 +93,123 @@ pub(crate) fn build_query<'a>(
...
+        Some(r::Value::Null) | None => None,
+        _ => unreachable!("after is a String"),
+    };
+
+    let before = match field.argument_value("before") {
+        Some(r::Value::String(s)) => Some(s.clone()),
```
lutter, line 143: Is this clone needed?

graphql/src/store/query.rs
```diff
@@ -68,6 +93,123 @@ pub(crate) fn build_query<'a>(
...
+            match decode_base64 {
+                Ok(decoded) => {
+                    let decoded_value = String::from_utf8(decoded);
+
+                    match decoded_value {
+                        Ok(val) => {
+                            // We encode the cursors using `:` as a separator
+                            // split by `:` and getting the first will give us the ID
+                            // these seems to be `"` around the ID so we need to remove them
+                            let id = val.split(":").collect::<Vec<&str>>()[0].replace("\"", "");
```
lutter: Did you ever figure out where the extra `\"` are coming from? We should really get to the bottom of this before merging.

graphql/src/store/query.rs
```diff
@@ -265,6 +407,13 @@ fn build_filter_from_object(
             use self::sast::FilterOp::*;
             let (field_name, op) = sast::parse_field_as_filter(key);
 
+            let actual_entity = match is_connection_type(entity.name()) {
+                false => entity,
+                true => schema
+                    .object_or_interface(&entity.name().replace("Connection", ""))
```
lutter, line 510: You should define a helper `connection_base_type(&str) -> &str` that does this without allocation. It would also be nicer to have a helper on `ApiSchema` that does this entire `match` statement.

graphql/src/store/query.rs
```diff
@@ -265,6 +407,13 @@ fn build_filter_from_object(
...
+                    .unwrap(),
```
lutter, line 511: `unwrap` causes a panic. It would be better to return a `QueryExecutionError` here; for other things we have `StoreError::ConstraintViolation` to report internal errors. It's probably time to add a variant `QueryExecutionError::ConstraintViolation` to report such internal errors - if we have a `None` here, clearly something is wrong in the `graph-node` code.

lutter (commented):
This isn't a review yet, just some comments from looking through some of the commits.

lutter:

PR #3219, opened 2022-02-04
ethereum: Fix adapter to filter proxy calls

### LINKED ISSUE #3194: Failure on Call input decoding
Example:
```
Subgraph failed with non-deterministic error: failed to process trigger: block #9212748 (0x2ca0…024f), transaction 4e6b6ba98b5249874882099903a988d23e50d2746cbccab9ae25e020106e69b5: Generating function inputs for the call Function { name: "issueBounty", inputs: [Param { name: "_issuer", kind: Address, internal_type: None }, Param { name: "_deadline", kind: Uint(256), internal_type: None }, Param { name: "_data", kind: String, internal_type: None }, Param { name: "_fulfillmentAmount", kind: Uint(256), internal_type: None }, Param { name: "_arbiter", kind: Address, internal_type: None }, Param { name: "_paysTokens", kind: Bool, internal_type: None }, Param { name: "_tokenContract", kind: Address, internal_type: None }], outputs: [Param { name: "", kind: Uint(256), internal_type: None }], constant: false, state_mutability: NonPayable } failed, raw input: 788acd7e00000000000001010101010101010101010101104080040080800000010000000001000000000000000000000000000000000000000000000000000100000001ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffdc0000000004000000000000000000000000000000000000000000000000000001000000000000000000000000000000000000000000000000000000002000028001: Invalid data, retry_delay_s: 1800, attempt: 81
```

Source code: https://github.com/graphprotocol/graph-node/blob/657f059021c2c2b95f17d1afee0649f88ba58efa/chain/ethereum/src/data_source.rs#L606.

https://gist.github.com/otaviopace/05a6fc20af6ca267d932cffaf4e0bd98

otaviopace:
Without this the line that does `function_abi.decode_input(&call.input.0[4..])` was failing with the `Err(InvalidData)`.

For some reason some of the OpenSea Proxy contracts don't get recognized as DELEGATECALL, instead they are seen as CALLs, but you can identify them by the input having the wrong size since it's a fallback function call.

A scenario to exemplify the problem, can be seen in this transaction: [0x4499dc75facb4cd27894196aac00bcb0ffbe3d64de87105878451d5637662f90](https://etherscan.io/tx/0x8e992eeb40e18703dd8169a8031e2113311985f1c20e0723a2dc362df40bafb0).

We would get the traces for the block `14043345` (`0xD648D1`) for the contract `0x7Be8076f4EA4A4AD08075C2508e481d6C946D12b`. Then we would find two CALLs for the `atomicMatch_` function. However the first one would have the wrong size (fail at decoding) and the second one would be correct.

That's because of the proxy and because of [how fallbacks are compiled from Solidity (https://ethereum.stackexchange.com/questions/12106/when-does-the-fallback-function-get-called) (omitting the function signature, aka a feel bytes less).

Fixes #3194


chain/ethereum/src/adapter.rs
```diff
@@ -303,24 +303,31 @@ impl EthereumCallFilter {
...
-                } else {
-                    // Ensure the call is to run a function the filter expressed an interest in
-                    signature.contains(&call.input.0[..4])
-                }
-            }
+        if let None = v {
+            return false;
+        }
+
+        let signature = &v.unwrap().1;
```
leoyvens, line 313:
```suggestion
        let signature = match v {
          None => return false,
          Some(v) => &v.1,
        }
```

chain/ethereum/src/adapter.rs
```diff
@@ -303,24 +303,31 @@ impl EthereumCallFilter {
...
+        // Caveat is this catch all clause limits you from matching with a specific call
+        // on the same address
+        if signature.is_empty() {
+            true
+        } else {
+            // Ensure the call is to run a function the filter expressed an interest in
+            let correct_fn = signature.contains(&call.input.0[..4]);
+            // Not multiple of 32, wrong length to decode parameters.
+            // This is probably a delegatecall disguised as a regular call
+            // via fallback.
```
leoyvens, line 327: Is there a reference you can link here if somebody wants to better understand how this happens? Maybe link an example trace in etherscan?

chain/ethereum/src/adapter.rs
```diff
@@ -303,24 +303,31 @@ impl EthereumCallFilter {
...
+            let wrong_input_size = (call.input.0.len() - 4) % 32 != 0;
+
+            correct_fn && !wrong_input_size
```
leoyvens, line 330:
```suggestion
            let correct_input_size = (call.input.0.len() - 4) % 32 == 0;

            correct_fn && correct_input_size
```

leoyvens (approved):
Approved after a few refactors

chain/ethereum/src/adapter.rs
otaviopace: I've wrapped this suggestion with the `self.contract_addresses_function_signatures.get(&call.to)` function call, so there's no more `v` let definition.

chain/ethereum/src/adapter.rs
otaviopace: Done 😊

chain/ethereum/src/adapter.rs
otaviopace: I've improved the comment, can you check if it's good now please?

PR #3233, opened 2022-02-08
Release 0.25.1

otaviopace:
I've pointed this PR to a temporary branch that's a copy of the `v0.25.0` git tag.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1475,7 +1475,7 @@ dependencies = [
 
 [[package]]
 name = "graph"
-version = "0.25.0"
+version = "0.25.1"
 dependencies = [
  "Inflector",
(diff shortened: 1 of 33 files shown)


otaviopace:
I'm also not sure if I should include commits that fix the cargo warnings we got with the upgrade.

mangas (approved):

leoyvens:
I don't think we want https://github.com/graphprotocol/graph-node/pull/2362 which is an old PR cherry picked on this. What would be necessary is probably to include the commits from this PR https://github.com/graphprotocol/graph-node/pull/2896. But this is getting too complicated, lets keep this update out of 0.25.1 and leave it for the next release from master.

PR #3237, opened 2022-02-10
instance_manager: Add env var for chain head subgraph sync status buffer

otaviopace:
This buffer is being added because the chain head can advance
when we're checking if a subgraph is synced or not.

### DIFF
--- a/core/src/subgraph/instance_manager.rs
+++ b/core/src/subgraph/instance_manager.rs
@@ -685,8 +685,17 @@ where
 
             match res {
                 Ok(needs_restart) => {
+                    debug!(
+                        logger,
+                        "Sync status before if";
+                        "synced" => format!("{}", synced),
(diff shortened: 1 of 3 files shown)


core/src/subgraph/instance_manager.rs
```diff
@@ -47,6 +47,14 @@ lazy_static! {
             .parse::<u64>()
             .map(Duration::from_secs)
             .expect("invalid GRAPH_SUBGRAPH_ERROR_RETRY_CEIL_SECS");
+
+    /// How many blocks behind chain head should be the buffer we will consider
+    /// that a subgraph is synced.
+    pub static ref SYNC_STATUS_BUFFER: u64 =
+        std::env::var("GRAPH_SUBGRAPH_SYNC_BUFFER")
```
otaviopace, line 54: @leoyvens also if you have a better naming suggestion for this.

PR #3238, opened 2022-02-10
store: Deploying a synced version should always replace current

lutter:


otaviopace:
@lutter this unit test failed: `fail_unfail_deterministic_error`.

I'll take a look if I can create the test for the deployment promotion right now on the side.

otaviopace (approved):
👏

PR #3240, opened 2022-02-10
Fix MacOS DNS issues on block ingestion

otaviopace:
This is being done because of DNS issues.

@VIVelev if you want to take a look as well since you were having the same issue 🙂

### DIFF
--- a/chain/ethereum/src/ethereum_adapter.rs
+++ b/chain/ethereum/src/ethereum_adapter.rs
@@ -114,9 +114,20 @@ lazy_static! {
         .unwrap_or("1000".into())
         .parse::<usize>()
         .expect("invalid GRAPH_ETHEREUM_BLOCK_INGESTOR_MAX_CONCURRENT_JSON_RPC_CALLS_FOR_TXN_RECEIPTS env var");
+}
 
+#[cfg(not(target_os = "macos"))]
+lazy_static! {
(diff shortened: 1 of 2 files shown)


maoueh:
I have some doubt about this to be honest. What is the actual error reported, is it `getaddrinfo EADDRINFO`?

I would aim much more at maximum open file descriptors limit being hit, which is by default super low on OS X system. I've been bitten by this just recently when the feature has been introduced, hence why I think it might be this.

It could be argue that it's still a reasonable fix to do, to lower support request but we should better understand the issue.

leoyvens:
On my M1 this works fine. I agree with @maoueh that if it can be avoided by tuning the OS then I'd rather not merge this.

The ideal way to attempt to fix this would be to rebase our web3 fork on the latest upstream web3, which uses reqwest as a client which has connection caching so it should hopefully not run into these issues.

otaviopace:
@leoyvens I remember searching about this and it doesn't seem to be a client related problem.

I will merge this then and we can investigate this further as @maoueh said so we can fix this on the source.

maoueh:
What was the exact error shown, never got the details :)

mangas:
```
error trying to connect: dns error: failed to lookup address information: nodename nor servname provided, or not known', node/src/chain.rs:107:25
```

PR #3245, opened 2022-02-14
Refactor Instance Manager - Part 1

otaviopace:
This PR just moves a few things around but it still doens't include the main refactor that extracts current code from the loop into minor components that are less coupled and with more clear / well defined responsibilities.

This is the first part of https://github.com/graphprotocol/graph-node/issues/3246, the second will separate the main loop into more well defined components with better boundaries of responsibilities.

The `instance_manager` inner workings should be the same after this gets merged.


core/src/subgraph/inputs.rs
```diff
@@ -0,0 +1,22 @@
...
+use graph::{
+    blockchain::Blockchain,
+    components::store::{DeploymentLocator, SubgraphFork},
+    data::subgraph::{SubgraphFeature, UnifiedMappingApiVersion},
+    prelude::BlockNumber,
+};
+use std::collections::BTreeSet;
+use std::sync::Arc;
+
+pub struct IndexingInputs<C: Blockchain> {
```
maoueh, line 11: Are we planning on expanding this module further to include more stuff? Maybe it could be merged with the `context.rs` module introduced in this PR.

maoueh (approved):
Approving since my comment is really minor.

Only other thing to mention is there is my Firehose cursor PR that will be heavily impacted by this, what's your merge timeline on this one? Just so that I can decide on rebase now and trying to push it or wait for your merge and rebase on top of it afterwards.

core/src/subgraph/inputs.rs
otaviopace: Do you have any suggestion for the name of this module (`context` + `state` + `inputs`)?
Is your PR approved, what's the number/link of it?

otaviopace: @maoueh I'll merge this then and change this on the next refactor PR, since I do intend on restructuring all of those anyway

PR #3254, opened 2022-02-16
minor optimization: Return early if there are no transaction hashes

tilacog:
This early return prevents an unecessary iteration over the block transactions, where each would be checked for containment against an empty vector.

### DIFF
--- a/chain/ethereum/src/ethereum_adapter.rs
+++ b/chain/ethereum/src/ethereum_adapter.rs
@@ -1664,6 +1664,11 @@ async fn filter_call_triggers_from_unsuccessful_transactions(
             "failed to obtain transaction hash from call triggers"
         ))?;
 
+    // Return early if there are no transaction hashes
+    if transaction_hashes.is_empty() {
+        return Ok(block);
+    }
+
     // And obtain all Transaction values for the calls in this block.
     let transactions: Vec<&Transaction> = {
         match &block.block {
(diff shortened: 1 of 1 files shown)


chain/ethereum/src/ethereum_adapter.rs
```diff
@@ -1664,6 +1664,11 @@ async fn filter_call_triggers_from_unsuccessful_transactions(
             "failed to obtain transaction hash from call triggers"
         ))?;
 
+    // Return early if there are no transaction hashes
+    if transaction_hashes.is_empty() {
```
otaviopace, line 1668: What about moving the code that's below to here as well?

Like this:

```rust
if transaction_hashes.is_empty() || transactions.is_empty() {
    return Ok(block);
}
```

otaviopace: Actually my idea will change behavior here: https://github.com/graphprotocol/graph-node/pull/3254/files#diff-b31a71a95f48aa7d0f11b5ea3501befc2560d749d73872e9eebe8003de85d921R1689

tilacog: It's not possible to move that check earlier because we still have to create the `transactions` value.

But, since we already checked that `transaction_hashes`:
1. is not empty (as this PR), and
2. has a different length than `transactions`,

it is safe to remove that check entirely, I believe.

maoueh (approved):

PR #3256, opened 2022-02-18
Prefetching optimization fix

tilacog:
Fixes the current errors of prefetching optimization feature.

This PR Includes the `block_range` column in some SQL queries.

### DIFF
--- a/store/postgres/src/relational_queries.rs
+++ b/store/postgres/src/relational_queries.rs
@@ -2994,7 +2994,7 @@ fn jsonb_build_object(
         }
         AttributeNames::Select(column_names) => {
             out.push_sql("jsonb_build_object(");
-            let mut iterator = iter_column_names(column_names, table, true).peekable();
+            let mut iterator = iter_column_names(column_names, table, false).peekable();
             while let Some(column_name) = iterator.next() {
                 // field name as json key
                 out.push_sql("'");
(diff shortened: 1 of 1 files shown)


store/postgres/src/relational_queries.rs
```diff
@@ -2942,7 +2942,15 @@ fn jsonb_build_object(
 fn iter_column_names<'a, 'b>(
     attribute_names: &'a BTreeSet<String>,
     table: &'b Table,
+    include_block_range_column: bool,
 ) -> impl Iterator<Item = &'b str> {
+    let extra = if include_block_range_column {
+        ["block_range"].iter()
```
lutter, line 2948: It would be better to use the constant `crate::block_range::BLOCK_RANGE_COLUMN` here

store/postgres/src/relational_queries.rs
```diff
@@ -2951,6 +2959,7 @@ fn iter_column_names<'a, 'b>(
         })
         .map(|column| column.name.as_str())
         .chain(BASE_SQL_COLUMNS.iter().copied())
+        .chain(extra)
```
lutter, line 2962: You can make `extra` just a `vec!` and leave off the `.copied()` from it, as well as from `BASE_SQL_COLUMNS`

lutter (approved):
LGTM, just a couple minor improvements

store/postgres/src/relational_queries.rs
tilacog: Nice! I didn't notice we already had that.

store/postgres/src/relational_queries.rs
tilacog: In this case, wouldn't copying those values (`&str`) be faster than allocating memory for the new vector?

lutter: Yes, you are right; just leave it as-is

PR #3259, opened 2022-02-20
node: make terminating slash in fork-base optional

VIVelev:
A small PR to make the ending slash for the fork-base URL optional. (It simply adds it if it's not there.)

### DIFF
--- a/node/src/bin/manager.rs
+++ b/node/src/bin/manager.rs
@@ -615,14 +615,25 @@ async fn main() {
         Ok(node) => node,
     };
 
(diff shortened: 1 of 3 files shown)


otaviopace (approved):
@VIVelev I thought a bit more and was about to say we might not want this since we do the same URL parsing for all `graph-node` CLI parameters and environment variables.

**However**, for this case we have to do a concatenation, adding the subgraph's `QmHash`. Since the fork base URL is generic (as it says in the name 😆 , base), it's probably better to add this.

Thanks for fixing the issue ❤️

Issue #3247, opened 2022-02-15
Add useful error messages/types to Result<T, ()>


otaviopace:
There are multiple functions that return `Result<T, ()>` in the code, this is equivalent to `Option<T>`, however instead of changing those to optional types, these should return useful error messages/types so it's easier to debug.

Example: https://github.com/graphprotocol/graph-node/blob/79390c39a13f166999ceea8defc55eedd05265ab/graph/src/data/subgraph/mod.rs#L95

PR #3265, opened 2022-02-21
graph: improve error messages

itsdaiego:
Small adjustments regarding [this issue](https://github.com/graphprotocol/graph-node/issues/3247)  :)

### DIFF
--- a/graph/src/data/store/mod.rs
+++ b/graph/src/data/store/mod.rs
@@ -60,18 +60,24 @@ impl SubscriptionFilter {
 pub struct NodeId(String);
 
 impl NodeId {
-    pub fn new(s: impl Into<String>) -> Result<Self, ()> {
+    pub fn new(s: impl Into<String>) -> Result<Self, anyhow::Error> {
         let s = s.into();
 
         // Enforce length limit
         if s.len() > 63 {
-            return Err(());
+            return Err(anyhow!(
+                "NodeId's name cannot be greater than 63 characters {}",
+                s
+            ));
         }
 
         // Check that the ID contains only allowed characters.
         // Note: these restrictions are relied upon to prevent SQL injection
         if !s.chars().all(|c| c.is_ascii_alphanumeric() || c == '_') {
-            return Err(());
+            return Err(anyhow!(
+                "Only alphanumeric and \"_\" characters are allowed {}",
+                s
+            ));
         }
 
         Ok(NodeId(s))
@@ -106,7 +112,7 @@ impl<'de> de::Deserialize<'de> for NodeId {
     {
         let s: String = de::Deserialize::deserialize(deserializer)?;
         NodeId::new(s.clone())
-            .map_err(|()| de::Error::invalid_value(de::Unexpected::Str(&s), &"valid node ID"))
+            .map_err(|_| de::Error::invalid_value(de::Unexpected::Str(&s), &"valid node ID"))
     }
 }
 
--- a/graph/src/data/subgraph/mod.rs
+++ b/graph/src/data/subgraph/mod.rs
@@ -166,7 +166,7 @@ impl TryFromValue for DeploymentHash {
 pub struct SubgraphName(String);
 
 impl SubgraphName {
-    pub fn new(s: impl Into<String>) -> Result<Self, ()> {
+    pub fn new(s: impl Into<String>) -> Result<Self, anyhow::Error> {
         let s = s.into();
 
         // Note: these validation rules must be kept consistent with the validation rules
@@ -174,7 +174,10 @@ impl SubgraphName {
 
         // Enforce length limits
         if s.is_empty() || s.len() > 255 {
-            return Err(());
+            return Err(anyhow!(
+                "SubgraphName's name cannot be empty or greater than 255 characters {}",
+                s
+            ));
         }
 
         // Check that the name contains only allowed characters.
@@ -182,19 +185,22 @@ impl SubgraphName {
             .chars()
             .all(|c| c.is_ascii_alphanumeric() || c == '-' || c == '_' || c == '/')
         {
-            return Err(());
+            return Err(anyhow!(
+                "Only alphanumeric, \"_\" and \"/\" characters are allowed {}",
+                s
+            ));
         }
 
         // Parse into components and validate each
         for part in s.split('/') {
             // Each part must be non-empty
             if part.is_empty() {
-                return Err(());
+                return Err(anyhow!("Part cannot be empty"));
             }
 
             // To keep URLs unambiguous, reserve the token "graphql"
             if part == "graphql" {
-                return Err(());
+                return Err(anyhow!("Part cannot be equal to \"graphql\""));
             }
 
             // Part should not start or end with a special character.
@@ -204,7 +210,10 @@ impl SubgraphName {
                 || !last_char.is_ascii_alphanumeric()
                 || !part.chars().any(|c| c.is_ascii_alphabetic())
             {
-                return Err(());
+                return Err(anyhow!(
+                    "Part cannot start or end with a special character: {}",
+                    s
+                ));
             }
         }
 
@@ -238,7 +247,7 @@ impl<'de> de::Deserialize<'de> for SubgraphName {
     {
         let s: String = de::Deserialize::deserialize(deserializer)?;
         SubgraphName::new(s.clone())
-            .map_err(|()| de::Error::invalid_value(de::Unexpected::Str(&s), &"valid subgraph name"))
+            .map_err(|_| de::Error::invalid_value(de::Unexpected::Str(&s), &"valid subgraph name"))
     }
 }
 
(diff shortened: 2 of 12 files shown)


otaviopace (changes requested):
First and foremost, thanks (and congrats 🎉 ) on your first contribution here 😊 
Well, upon talking with @leoyvens he suggested that those messages actually get used on the other side, as you can see here for example:

```rust
// It was like this:
// We would match the error to a `()` unit value
.map_err(|()| anyhow!("invalid node id `{}`", node))?;
// Now that's like this:
// You can actually pass the error message forward, and improve whoever is consuming it
.map_err(|err| anyhow!("`{}` node id invalid, error: ", node, err))?;
```

Could you fix this please so your change actually uses the new error messages? 🙂

lutter:
> > **Edit**: thinking a bit more perhaps you can even use `.context`/`.with_context` ([anyhow Context docs](https://docs.rs/anyhow/latest/anyhow/trait.Context.html#example)) instead of `.map_err` so you don't have to do any `String` formatting tada

One thing that's very annoying with `with_context` is that error messages get split into multiple lines, which causes them to show up as multiple log lines, which in turn makes finding things in the logs harder than necessary. Having the entire error message on one log line is actually more usable, even if the formatting makes it harder to read.

leoyvens:
@lutter That depends on how you print it [[source](https://docs.rs/anyhow/latest/anyhow/struct.Error.html)], the `{:#}` format will print causes in a single line, but the debug `{:?}` format will print them in separate lines.

graph/src/data/store/mod.rs
```diff
@@ -105,7 +111,7 @@ impl<'de> de::Deserialize<'de> for NodeId {
     {
         let s: String = de::Deserialize::deserialize(deserializer)?;
         NodeId::new(s.clone())
-            .map_err(|()| de::Error::invalid_value(de::Unexpected::Str(&s), &"valid node ID"))
```
otaviopace, line 108: I think you forgot this one?

graph/src/data/subgraph/mod.rs
```diff
@@ -238,7 +247,7 @@ impl<'de> de::Deserialize<'de> for SubgraphName {
     {
         let s: String = de::Deserialize::deserialize(deserializer)?;
         SubgraphName::new(s.clone())
-            .map_err(|()| de::Error::invalid_value(de::Unexpected::Str(&s), &"valid subgraph name"))
+            .map_err(|_| de::Error::invalid_value(de::Unexpected::Str(&s), &"valid subgraph name"))
```
otaviopace, line 250: And this one

graph/src/data/store/mod.rs
```diff
@@ -105,7 +111,7 @@ impl<'de> de::Deserialize<'de> for NodeId {
...
+            .map_err(|_| de::Error::invalid_value(de::Unexpected::Str(&s), &"valid node ID"))
```
otaviopace, line 114: I think you forgot this one?

graph/src/data/subgraph/mod.rs
itsdaiego: @otaviopace  that's actually one of the places that I wasn't sure about :D 

I was reading [this documentation](https://docs.serde.rs/serde/de/trait.Error.html#method.invalid_value) and from my understanding the first value should be the value the we tried to serialize (which seems fine) but the other one is the "expected value", so showing something like "valid subgraph name, error: <some error>", didn't make much sense to me 🤔   

what do you think?

node/src/config.rs
```diff
@@ -380,8 +380,13 @@ pub struct ChainSection {
 
 impl ChainSection {
     fn validate(&mut self) -> Result<()> {
-        NodeId::new(&self.ingestor)
-            .map_err(|()| anyhow!("invalid node id for ingestor {}", &self.ingestor))?;
+        NodeId::new(&self.ingestor).map_err(|err| {
```
otaviopace, line 383: You can use `.context` here instead of adding `with error {}` to the message.

node/src/config.rs
```diff
@@ -817,8 +822,9 @@ impl DeploymentPlacer for Deployment {
                     .indexers
                     .iter()
                     .map(|idx| {
-                        NodeId::new(idx.clone())
-                            .map_err(|()| format!("{} is not a valid node name", idx))
+                        NodeId::new(idx.clone()).map_err(|err| {
```
otaviopace, line 825: You can use `.context` here instead of adding `with error {}` to the message.

node/src/config.rs
```diff
@@ -865,7 +871,8 @@ impl Rule {
             return Err(anyhow!("useless rule without indexers"));
         }
         for indexer in &self.indexers {
-            NodeId::new(indexer).map_err(|()| anyhow!("invalid node id {}", &indexer))?;
+            NodeId::new(indexer)
+                .map_err(|error| anyhow!("invalid node id {} with error `{}`", &indexer, error))?;
```
otaviopace, line 875: You can use `.context` here instead of adding `with error {}` to the message.

node/src/manager/commands/assign.rs
```diff
@@ -31,7 +31,8 @@ pub fn reassign(
     node: String,
     shard: Option<String>,
 ) -> Result<(), Error> {
-    let node = NodeId::new(node.clone()).map_err(|()| anyhow!("illegal node id `{}`", node))?;
+    let node = NodeId::new(node.clone())
+        .map_err(|err| anyhow!("illegal node id `{}` with error `{}`", node, err))?;
```
otaviopace, line 35: You can use `.context` here instead of adding `with error {}` to the message.

node/src/manager/commands/config.rs
```diff
@@ -48,7 +48,7 @@ pub fn pools(config: &Config, nodes: Vec<String>, shard: bool) -> Result<(), Err
         .into_iter()
         .map(|name| {
             NodeId::new(name.replace("-", "_"))
-                .map_err(|()| anyhow!("illegal node name `{}`", name))
+                .map_err(|err| anyhow!("illegal node name `{}` with error `{}`", name, err))
```
otaviopace, line 51: You can use `.context` here instead of adding `with error {}` to the message.

node/src/manager/commands/create.rs
```diff
@@ -5,7 +5,7 @@ use graph_store_postgres::SubgraphStore;
 
 pub fn run(store: Arc<SubgraphStore>, name: String) -> Result<(), Error> {
     let name = SubgraphName::new(name.clone())
-        .map_err(|()| anyhow!("illegal subgraph name `{}`", name))?;
+        .map_err(|err| anyhow!("illegal subgraph name `{}` with error `{}`", name, err))?;
```
otaviopace, line 8: You can use `.context` here instead of adding `with error {}` to the message.

node/src/manager/commands/copy.rs
```diff
@@ -119,7 +119,8 @@ pub async fn create(
     let base_ptr = BlockPtr::from((hash, src_number));
 
     let shard = Shard::new(shard)?;
-    let node = NodeId::new(node.clone()).map_err(|()| anyhow!("invalid node id `{}`", node))?;
+    let node = NodeId::new(node.clone())
+        .map_err(|err| anyhow!("invalid node id `{}` with error `{}`", node, err))?;
```
otaviopace, line 123: You can use `.context` here instead of adding `with error {}` to the message.

otaviopace (changes requested):
I've "re-reviewed" the code and I saw a few occurrences of code that could benefit from using `.context` instead of `.map_err` + `"with error {}"` to the error message.

Can you please take a look again @itsdaiego ?

Sorry for the big delay in the response! 😅

node/src/config.rs
itsdaiego: This one I'm kinda confused, once I use `.with_context` I'm getting this error ->   "`?` couldn't convert the error to `std::string::String`: the trait `From<graph::prelude::Error>` is not implemented for `std::string::String`", no entirely sure how to get around this, any thoughts? 🙏🏽

evaporei: I didn't understand either, not the prettiest thing, but here's the solution:
```rust
NodeId::new(idx.clone())
    .context(|| format!("{} is not a valid node name", idx))
    .map_err(|e| e.to_string())
```

itsdaiego: Thanks, the only way I managed to fix this is like so:

```
      .map(|idx| {
          NodeId::new(idx.clone())
              .with_context(|| format!("{:#} is not a valid node name", idx))
      })
      .collect::<Result<Vec<_>, _>>()
      .unwrap();

```

My understanding is that since we don't return a Result (to deal w/  possible errors) using unwrap would not be a bad idea (maybe?!) 😬  but again not sure about that also.

Which solution you think is more appropriate for this situation? I'm just suggesting mine because it required a little bit less  coe

PR #3268, opened 2022-02-23
Fix skip block pointer update

otaviopace:
This PR does two things:

1. Stop skipping blocks if the subgraph is synced
2. Ignore a revert of a block that wasn't processed

It's necessary because of a bug introduced in #3223 

**Edit**: I'm still testing this locally with a subgraph that's close to chain head. (probably with the [debug forking tool](https://thegraph.com/docs/en/developer/subgraph-debug-forking/))

### DIFF
--- a/store/postgres/src/deployment_store.rs
+++ b/store/postgres/src/deployment_store.rs
@@ -1090,10 +1090,17 @@ impl DeploymentStore {
     ) -> Result<StoreEvent, StoreError> {
         let conn = self.get_conn()?;
         // Unwrap: If we are reverting then the block ptr is not `None`.
-        let block_ptr_from = Self::block_ptr_with_conn(&site.deployment, &conn)?.unwrap();
+        let deployment_head = Self::block_ptr_with_conn(&site.deployment, &conn)?.unwrap();
+
+        // Early return to cover scenario where the BlockStream advanced more than
+        // the deployment head (database). In this case, revert_block_operations shouldn't
+        // do anything, there's nothing to be reverted.
+        if block_ptr_to.number >= deployment_head.number {
+            return Ok(StoreEvent::default());
+        }
 
         // Sanity check on block numbers
-        if block_ptr_from.number != block_ptr_to.number + 1 {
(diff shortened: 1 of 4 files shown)


store/postgres/src/deployment_store.rs
```diff
@@ -1090,10 +1090,17 @@ impl DeploymentStore {
...
         let conn = self.get_conn()?;
         // Unwrap: If we are reverting then the block ptr is not `None`.
-        let block_ptr_from = Self::block_ptr_with_conn(&site.deployment, &conn)?.unwrap();
+        let deployment_head = Self::block_ptr_with_conn(&site.deployment, &conn)?.unwrap();
+
+        // Early return to cover scenario where the BlockStream advanced more than
+        // the deployment head (database). In this case, revert_block_operations shouldn't
+        // do anything, there's nothing to be reverted.
+        if block_ptr_to.number >= deployment_head.number {
+            return Ok(StoreEvent::default());
```
otaviopace, line 1099: The other option (that might be more appropriate) is to change the return value of `revert_block_operations` to `Option<StoreEvent>`

leoyvens: An empty event is fine imo. Also the comment should be more specific in how this can happen.

otaviopace: I'll improve the comment then!

leoyvens (approved):

tilacog (approved):

otaviopace:
Before merging this, I would like your opinion @leoyvens on this:

I've thought a bit more and perhaps it's good if we make the other `revert_block_operations` return `Option` as well, because I don't think we want to run the rest of this block: https://github.com/graphprotocol/graph-node/blob/347c512e96b39534d67ae90f9186d4b710209dcf/core/src/subgraph/runner.rs#L186.

Which includes:

- Change `revert_block_operations` metrics
- Revert `entity_data_sources`
- Clear the `EntityCache`

I could do an early return by stopping that `match` branch execution by handling the `Ok(None)` result.

PR #3270, opened 2022-02-23
graphman `fix block` command

tilacog:
Adds a new graphman command, that can be used either to:

1. compare, diff and prune cached blocks against freshly obtained ones from the configured JRPC provider, or
2. truncate an entire block cache (for a given chain)


node/Cargo.toml
```diff
@@ -43,6 +43,7 @@ diesel = "1.4.8"
 fail = "0.5"
 http = "0.2.5" # must be compatible with the version rust-web3 uses
 prometheus = { version ="0.13.0", features = ["push"] }
+json-structural-diff = {version = "0.1", features = ["colorize"] }
```
tilacog, line 46: This is a new dependency for diffing JSONs in a structured/canonical way. 
Although named `"colorize"`, this feature is only related to pretty-printing the diff output to the user

node/src/bin/manager.rs
```diff
@@ -1,39 +1,35 @@
...
 git_testament!(TESTAMENT);
 
-macro_rules! die {
-    ($fmt:expr, $($arg:tt)*) => {{
-        use std::io::Write;
-        writeln!(&mut ::std::io::stderr(), $fmt, $($arg)*).unwrap();
-        ::std::process::exit(1)
-    }}
-}
-
```
tilacog, line 36: I've took the liberty to remove this macro and use a result returning `main` function, that will also pretty-print `anyhow::Error` values to the end user.

node/src/bin/manager.rs
```diff
@@ -605,10 +634,16 @@ impl Context {
...
+    async fn ethereum_networks(&self) -> anyhow::Result<EthereumNetworks> {
+        let logger = self.logger.clone();
+        let registry = self.metrics_registry();
+        create_ethereum_networks(logger, registry, &self.config).await
+    }
 }
 
 #[tokio::main]
-async fn main() {
+async fn main() -> anyhow::Result<()> {
```
tilacog, line 646: Having a result returning `main` function will allow us to leverage the `?` operator and return early more conveniently.

node/src/manager/commands/fix_block.rs
```diff
@@ -0,0 +1,304 @@
...
+        steps::fetch_single_provider_block(&block_hash, ethereum_adapter, logger).await?;
+    let diff = steps::diff_block_pair(&cached_block, &provider_block);
+    steps::report_difference(diff.as_deref(), &block_hash);
+    if diff.is_some() {
+        steps::delete_block(&block_hash, &chain_store)?;
+    }
+    Ok(())
+}
+
+mod steps {
```
tilacog, line 86: Since there are lots of common steps, I've chosen to group them in submodules and reuse them in the topmost module.

node/src/manager/commands/fix_block.rs
```diff
@@ -0,0 +1,304 @@
...
+        match (iterator.next(), iterator.next()) {
+            (Some(a), None) => Ok(a),
+            (None, None) => bail!("Expected a single {name} but found none."),
+            _ => bail!("Expected a single {name} but found multiple occurrences."),
+        }
+    }
+}
+
+/// Custom range type that supports being parsed from a string.
+mod ranges {
```
tilacog, line 233: This module is an attempt to have ranges that can be parsed from strings like `"10..=15"` and `"20.."`, since Rust's built-in `Range` type is incapable of conversions.

node/src/bin/manager.rs
lutter: Nice, yeah this was definitely a C-ism left over

node/src/bin/manager.rs
```diff
@@ -445,6 +442,36 @@ pub enum IndexCommand {
     },
 }
 
+#[derive(Clone, Debug, StructOpt)]
+pub struct FixBlockCommand {
+    /// Network name (must fit one of the chain)
```
lutter, line 447: Maybe this should be called 'chain name', since you'd use `graphman chain list` to see which ones are valid. I would change teh help text to `Chain name (must be an existing chain, see 'chain list')`

node/src/bin/manager.rs
```diff
@@ -445,6 +442,36 @@ pub enum IndexCommand {
...
+    #[structopt(empty_values = false)]
+    network_name: String,
+    #[structopt(subcommand)]
+    method: FixBlockSubCommand,
+}
```
lutter, line 452: I find that a little awkward - you'd have to say `graphman fix-block $chain by-hash $hash`; I wonder if it wouldn't be better to make this two subcommands `graphman chain check-blocks [-h hash|-n number|-r range] <chain>` and `graphman chain truncate [-f|--force] <chain>`

node/src/bin/manager.rs
```diff
@@ -445,6 +442,36 @@ pub enum IndexCommand {
...
+enum FixBlockSubCommand {
+    /// The number of the target block
+    ByHash { hash: String },
+
+    /// The hash of the target block
+    ByNumber { number: i32 },
+
+    /// A block number range in the form: `start..end`
+    ///
+    /// All valid [`std::ops::Range`] definitions are accepted.
```
lutter, line 464: This requires that users know the Rust syntax for ranges. Would be better to spell that out. Or just have one allowed syntax. A third option is to pass in separate `--from` and `--to` arguments instead of a literal range.

node/src/bin/manager.rs
```diff
@@ -514,21 +541,21 @@ impl Context {
...
-            self.registry,
+            self.metrics_registry(),
             Arc::new(vec![]),
         );
         pool.skip_setup();
         pool
     }
 
-    fn subgraph_store(self) -> Arc<SubgraphStore> {
+    fn subgraph_store(&self) -> Arc<SubgraphStore> {
```
lutter, line 558: The reason all these methods take `self` and not `&self` is to make sure that `StoreBuilder` doesn't get called more than once since that would create more pools than expected. You could e.g. call `ethereum_networks()` first and `store()` second in your code below, and only `await` on the eth networks in the `if let Some(..)`. Generally, it would be nicer if we only connected to the network the user needs instead of all of them, but that can wait for a future enhancement.

node/src/manager/commands/fix_block.rs
```diff
@@ -0,0 +1,300 @@
...
+                Some(x) if x < 0 => anyhow::bail!("Negative block number"),
+                Some(x) => x,
+            };
+            let inclusive = if self.inclusive { 1 } else { 0 };
+            let max = self.upper_bound.map(|x| x + inclusive);
+            Ok((min, max))
+        }
+    }
+
+    impl FromStr for Range {
```
lutter, line 260: I would avoid all this and just have users pass in `--from` and `--to` (inclusive) options.

node/src/manager/commands/fix_block.rs
```diff
@@ -0,0 +1,300 @@
...
+            Self {
+                lower_bound,
+                upper_bound,
+                inclusive,
+            }
+        }
+
+        pub(super) fn min_max(&self) -> anyhow::Result<(i32, Option<i32>)> {
+            let min = match self.lower_bound {
+                None | Some(0) => 1, // Never include the genesis block
```
lutter, line 250: It would be better if trying to do this for block 0 would result in an error instead of silently adjusting to 1

lutter (changes requested):

node/src/bin/manager.rs
tilacog: I agree. 
Would you think it would make sense if we ask for confirmation for open-ended intervals? 
For instance, if the command is invoked with only `--from` or `--to`, we could bring a prompt to confirm the operation.

In that case, we should also provide a `--no-confirm` flag to auto-confirm those prompts.

node/src/bin/manager.rs
tilacog: Got it!
That makes a lot of sense now.
I'll revert that change and add a comment explaining the rationale for moving `self`.

azf20:
@tilacog nudging on the status of this?

tilacog:
I still have to implement the requested changes. I'll try to find some time today to work on them.

node/src/bin/manager.rs
tilacog: Fixed in 284ee30d4

node/src/bin/manager.rs
tilacog: Fixed in 569e9bbd7

node/src/manager/commands/fix_block.rs
tilacog: Fixed in 6a2f96958

node/src/bin/manager.rs
tilacog: Fixed in 180124537

node/src/manager/commands/fix_block.rs
tilacog: Removed in 180124537

node/src/bin/manager.rs
tilacog: Fixed in e8794ae5e

tilacog: Fixed in e8794ae5e

lutter (approved):
Nice! I think this will be very useful to a lot of people

PR #3271, opened 2022-02-24
Allow using `Bytes` as well as `String`/`ID` for the `id` of entities

lutter:
This PR makes it possible to declare entities with an `id` type of `Bytes`; building on previous support that was committed a long time ago, with this PR, the `Bytes` are not converted to a string before handing them to mappings. Instead, such entities show up in the mappings with a bytes value for their ID. Companion PR's for `graph-ts` and `graph-cli` will make that more usable.

An example of what it takes to convert a subgraph to all `Bytes` ids can be found [here](https://github.com/lutter/uniswap-v2-subgraph/tree/bytes)

### DIFF
--- a/graph/src/data/schema.rs
+++ b/graph/src/data/schema.rs
@@ -13,6 +13,7 @@ use crate::prelude::{
 use anyhow::{Context, Error};
 use graphql_parser::{self, Pos};
 use inflector::Inflector;
+use itertools::Itertools;
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
 
@@ -57,6 +58,8 @@ pub enum SchemaValidationError {
          the following fields: {2}"
     )]
     InterfaceFieldsMissing(String, String, Strings), // (type, interface, missing_fields)
+    #[error("Implementors of interface `{0}` use different id types `{1}`. They must all use the same type")]
+    InterfaceImplementorsMixId(String, String),
     #[error("Field `{1}` in type `{0}` has invalid @derivedFrom: {2}")]
     InvalidDerivedFrom(String, String, String), // (type, field, reason)
     #[error("The following type names are reserved: `{0}`")]
@@ -844,6 +847,7 @@ impl Schema {
             self.validate_schema_type_has_no_fields(),
             self.validate_directives_on_schema_type(),
             self.validate_reserved_types_usage(),
+            self.validate_interface_id_type(),
         ]
         .into_iter()
         .filter(Result::is_err)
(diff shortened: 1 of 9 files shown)


dotansimha (approved):

runtime/wasm/src/host_exports.rs
```diff
@@ -150,22 +150,6 @@ impl<C: Blockchain> HostExports<C> {
         );
         poi_section.end();
 
-        let id_insert_section = stopwatch.start_section("host_export_store_set__insert_id");
```
otaviopace, line 153: Shouldn't we keep this to measure the performance with `Bytes`?

store/postgres/src/relational.rs
```diff
@@ -290,6 +290,7 @@ impl Layout {
                     } else {
                         // For interfaces that are not implemented at all, pretend
                         // they have a String `id` field
+                        // see also: id-type-for-unimplemented-interfaces
```
otaviopace, line 293: Woah I like this, less cryptic than the UUIDs 😅

graph/src/data/schema.rs
```diff
@@ -589,6 +594,33 @@ impl Schema {
...
+            .unwrap()
+            .field_type
+            .get_base_type();
+
+        match base_type {
+            "String" => Ok(store::Value::String(key.entity_id.clone())),
+            "Bytes" => Ok(store::Value::Bytes(scalar::Bytes::from_str(
+                &key.entity_id,
+            )?)),
+            _ => todo!(),
```
otaviopace, line 620: Could this `todo!` give a bad error message to an user (eg subgraph developer)?

graph/src/data/store/mod.rs
```diff
@@ -563,11 +564,15 @@ impl Entity {
         v
     }
 
-    /// Try to get this entity's ID
+    /// Return the ID of this entity. If the ID is a string, return the
+    /// string. If it is `Bytes`, return it as a hex string with a `0x`
+    /// prefix. If the ID is not set or anything but a `String` or `Bytes`,
+    /// return an error
     pub fn id(&self) -> Result<String, Error> {
```
otaviopace, line 571: Where is this `id` function used? Depending on where it's used, I'm curious if changing the return type of this to `Bytes` instead of `String` wouldn't help with performance too.

lutter: We can't change it to `Bytes` because some entities might have actual string ids that can't be converted to `Bytes`. It's used in a few places where entity ids are compared (e.g., for the keys in the `EntityCache`) and IIRC for the explicit id that gets passed in from `store_set` (a string)

It would be a bit cleaner to make this an `enum` but that would require changing _a lot_ of code since it will ripple to `EntityKey`. I don't think it will buy us much since for any given entity type, all ids are either strings or bytes, and there's therefore no danger that a `String` id and a `Bytes` id might inadvertently clash.

graph/src/data/schema.rs
lutter: Ooops .. that should return an error; schema validation at some point should have checked that the `id` is either `String` or `Bytes`, but this shouldn't panic. I'll fix it.

runtime/wasm/src/host_exports.rs
lutter: This was only measuring the time it took to insert into an in-memory map, so not super-interesting; I moved that logic to `EntityCache.set` and this operation now gets measured as part of the `host_export_store_set` section which I have never seen take interesting amounts of time.

otaviopace: Oh okay, makes sense!

otaviopace (approved):
💯

Issue #3274, opened 2022-02-24
GraphQL API for entity changes in a subgraph and block


neysofu:
# Objective

We want to provide new debugging capabilities for indexers, which can then be leveraged by our internal tool IXI to bisect PoI mismatches.

# Proposed changes

We should define a new GraphQL indexer route for dumping all entity changes (removals, additions, updates) that happened at a specific block.

```graphql
type Query {
  # ...

  entitiesChangedInBlock(
    subgraphId: String!
    blockNumber: Int!
  ): [EntityData!]!
}

type EntityData {
  name: String!
  fields: [String!]!
}
```

## Previous work

- https://github.com/graphprotocol/integration-testing/blob/98a5988c4d8c8dbd89194d66e1abf19a0ee52927/src/utils/entities.ts#L52: An example of the SQL query that we'd like to expose via the GraphQL route.

lutter:
I don't fully understand the `EntityData` type. Would `name` be the name of the entity type? Ideally, you'd be able to return JSON like this:
```
{
  "Account": [{ "id": "account1", "balance": 200}, {"id": "account2", "balance: 50"}],
  "Transaction": [ .. ]
  ..
}
```
but I am not sure how you would write a GraphQL type for basically arbitrary JSON (arbitrary for at least the array contents) Maybe @dotansimha  has some idea on how to express that?

neysofu:
Good point @lutter, my design for `EntityData` only contains field names and not values. This is not the desired behavior.

I think we have at least two options for mapping entity field values to GraphQL:

- Something akin to https://github.com/taion/graphql-type-json i.e. a custom `scalar JSON` type.
- Marshalling values as JSON, and treat the minified JSON as a `String!`.

PR #3275, opened 2022-02-24
GraphQL route for dumping entity changes in subgraph and block

neysofu:
Closes https://github.com/graphprotocol/graph-node/issues/3274.

The GraphQL API implemented here looks like the following (a bit different from #3274's first draft; after collecting feedback):

```graphql
type Query {
  entityChangesInBlock(subgraphId: String!, blockNumber: Int!): EntityChanges!
}

type EntityChanges {
  updates: [EntityData!]!
  deletes: [String!]!
}

type EntityData {
  name: String!
  fields: [EntityField!]!
}

type EntityField {
  field: String!
  valueAsJSON: String!
}
```

### DIFF
--- a/graph/src/components/store.rs
+++ b/graph/src/components/store.rs
@@ -988,11 +988,15 @@ pub trait SubgraphStore: Send + Sync + 'static {
     /// subgraph has any deployments attached to it
     fn subgraph_exists(&self, name: &SubgraphName) -> Result<bool, StoreError>;
 
-    fn changed_entities_in_block(
+    /// Returns a collection of all [`EntityModification`] items in relation to
+    /// the given [`BlockNumber`]. No distinction is made between inserts and
+    /// updates, which may be returned as either [`EntityModification::Insert`]
+    /// or [`EntityModification::Overwrite`].
+    fn entity_changes_in_block(
         &self,
         subgraph_id: &DeploymentHash,
         block_number: BlockNumber,
-    ) -> Result<BTreeMap<EntityType, Entity>, StoreError>;
+    ) -> Result<Vec<EntityModification<EntityType>>, StoreError>;
 
     /// Return the GraphQL schema supplied by the user
     fn input_schema(&self, subgraph_id: &DeploymentHash) -> Result<Arc<Schema>, StoreError>;
@@ -1350,13 +1354,13 @@ pub trait StatusStore: Send + Sync + 'static {
 /// `EntityOperation`, we already know whether a `Set` should be an `Insert`
 /// or `Update`
 #[derive(Clone, Debug, PartialEq, Eq)]
-pub enum EntityModification {
+pub enum EntityModification<K = EntityKey> {
     /// Insert the entity
-    Insert { key: EntityKey, data: Entity },
+    Insert { key: K, data: Entity },
     /// Update the entity by overwriting it
-    Overwrite { key: EntityKey, data: Entity },
+    Overwrite { key: K, data: Entity },
     /// Remove the entity
-    Remove { key: EntityKey },
+    Remove { key: K },
 }
 
 impl EntityModification {
--- a/graph/src/components/store.rs
+++ b/graph/src/components/store.rs
@@ -988,6 +988,12 @@ pub trait SubgraphStore: Send + Sync + 'static {
     /// subgraph has any deployments attached to it
     fn subgraph_exists(&self, name: &SubgraphName) -> Result<bool, StoreError>;
 
+    fn changed_entities_in_block(
+        &self,
+        subgraph_id: &DeploymentHash,
+        block_number: BlockNumber,
+    ) -> Result<BTreeMap<EntityType, Entity>, StoreError>;
+
     /// Return the GraphQL schema supplied by the user
     fn input_schema(&self, subgraph_id: &DeploymentHash) -> Result<Arc<Schema>, StoreError>;
 
--- a/graph/src/components/store.rs
+++ b/graph/src/components/store.rs
@@ -996,7 +996,7 @@ pub trait SubgraphStore: Send + Sync + 'static {
         &self,
         subgraph_id: &DeploymentHash,
         block_number: BlockNumber,
-    ) -> Result<Vec<EntityModification<EntityType>>, StoreError>;
+    ) -> Result<Vec<EntityOperation>, StoreError>;
 
     /// Return the GraphQL schema supplied by the user
     fn input_schema(&self, subgraph_id: &DeploymentHash) -> Result<Arc<Schema>, StoreError>;
@@ -1354,13 +1354,13 @@ pub trait StatusStore: Send + Sync + 'static {
 /// `EntityOperation`, we already know whether a `Set` should be an `Insert`
 /// or `Update`
 #[derive(Clone, Debug, PartialEq, Eq)]
-pub enum EntityModification<K = EntityKey> {
+pub enum EntityModification {
     /// Insert the entity
-    Insert { key: K, data: Entity },
+    Insert { key: EntityKey, data: Entity },
     /// Update the entity by overwriting it
-    Overwrite { key: K, data: Entity },
+    Overwrite { key: EntityKey, data: Entity },
     /// Remove the entity
-    Remove { key: K },
+    Remove { key: EntityKey },
 }
 
 impl EntityModification {
--- a/graph/src/data/value.rs
+++ b/graph/src/data/value.rs
@@ -215,6 +215,7 @@ impl Value {
             ("Bytes", Value::String(s)) => Ok(Value::String(s)),
             ("BigInt", Value::String(s)) => Ok(Value::String(s)),
             ("BigInt", Value::Int(n)) => Ok(Value::String(n.to_string())),
+            ("JSONObject", Value::Object(obj)) => Ok(Value::Object(obj)),
             (_, v) => Err(v),
         }
     }
--- a/server/index-node/src/resolver.rs
+++ b/server/index-node/src/resolver.rs
@@ -6,10 +6,11 @@ use graph::data::subgraph::features::detect_features;
 use graph::data::subgraph::{status, MAX_SPEC_VERSION};
 use graph::prelude::*;
 use graph::{
-    components::store::StatusStore,
+    components::store::{EntityType, StatusStore},
     data::graphql::{IntoValue, ObjectOrInterface, ValueMap},
 };
 use graph_graphql::prelude::{a, ExecutionContext, Resolver};
+use std::collections::BTreeMap;
 use std::convert::TryInto;
 use web3::types::{Address, H256};
 
@@ -87,36 +88,23 @@ where
         Ok(infos.into_value())
     }
 
-    fn resolve_entities_changed_in_block(
+    fn resolve_entity_changes_in_block(
         &self,
         field: &a::Field,
     ) -> Result<r::Value, QueryExecutionError> {
-        let block_number = field
-            .get_required::<BlockNumber>("blockNumber")
-            .expect("Valid blockNumber required")
-            .try_into()
-            .unwrap();
-
         let subgraph_id = field
             .get_required::<DeploymentHash>("subgraphId")
-            .expect("Valid subgraphId not provided");
+            .expect("Valid subgraphId required");
 
-        let entities = self
+        let block_number = field
+            .get_required::<BlockNumber>("blockNumber")
+            .expect("Valid blockNumber required");
+
+        let entity_changes = self
             .subgraph_store
-            .changed_entities_in_block(&subgraph_id, block_number)?;
+            .entity_changes_in_block(&subgraph_id, block_number)?;
 
-        let response = r::Value::object(
-            entities
-                .into_iter()
-                .map(|(entity_type, _entity)| {
-                    let entity_type = entity_type.to_string();
-                    let value = r::Value::Null; // FIXME
-                    (entity_type, value)
-                })
-                .collect(),
-        );
-
-        Ok(response)
+        Ok(entity_changes_to_graphql(entity_changes))
     }
 
     fn resolve_proof_of_indexing(&self, field: &a::Field) -> Result<r::Value, QueryExecutionError> {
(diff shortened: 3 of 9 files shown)


store/postgres/src/block_range.rs
```diff
@@ -87,6 +87,31 @@ impl ToSql<Range<Integer>, Pg> for BlockRange {
...
+        out.push_sql("lower(");
+        out.push_sql(self.table_prefix);
+        out.push_identifier(BLOCK_RANGE_COLUMN)?;
+        out.push_sql(") = ");
+        out.push_bind_param::<Integer, _>(&self.block)?;
+        out.push_sql(" or upper(");
+        out.push_sql(self.table_prefix);
+        out.push_identifier(BLOCK_RANGE_COLUMN)?;
+        out.push_sql(") = ");
+        out.push_bind_param::<Integer, _>(&self.block)?;
```
lutter, line 109: By doing `lower(block_range) = $b or upper(block_range) = $b` you'll get both the old and the new entity version in your query result. It's worth a thought if the API should actually return an indication to users what change was made (whether the entity was inserted, updated, or deleted at that block) The old version will have block range `[x, $b)`, the new one will have `[$b,y)` If you ran two queries, one where you get the entities with `lower(block_range) = $b` and another one where you only select the `id` where `upper(block_range) = $b` you could piece that together from just that information and don't need to read the entire old version.

Also, when checking the upper bound of the block range it's better to use the (very unwieldy) `coalesce(upper(block_range), 2147483647)` since there is an index for that rather than `upper(block_range)`, but that's a minor detail for now.

lutter (commented):
Nice! Looks good so far!

store/postgres/src/relational_queries.rs
```diff
@@ -1205,6 +1207,96 @@ impl<'a> LoadQuery<PgConnection, EntityData> for FindQuery<'a> {
...
+    fn walk_ast(&self, mut out: AstPass<Pg>) -> QueryResult<()> {
+        out.unsafe_to_cache_prepared();
+
+        for (i, table) in self.tables.iter().enumerate() {
+            if i > 0 {
+                out.push_sql("\nunion all\n");
+            }
+            out.push_sql("select ");
+            out.push_bind_param::<Text, _>(&table.object.as_str())?;
+            out.push_sql(" as entity, to_jsonb(e.*) as data\n");
```
lutter, line 1275: It would be better for performance to only select the id here instead of the entire object, i.e., `select 'EntityType' as entity, e.id from ..` That would require doing the whole `QueryableByName` dance for a new struct, but is not  hard for reading two strings. Just adding this struct should do the trick:
```rust
#[derive(QueryableByName)]
struct EntityDeletion {
    #[sql_type = "Text"]
    entity: String,
    #[sql_type = "Text"]
    id: String,
}
```

store/postgres/src/relational.rs
```diff
@@ -572,6 +573,49 @@ impl Layout {
...
+
+        let mut changes = Vec::new();
+
+        for data in deletes.into_iter() {
+            // The complexity is bad here (we're iterating over the set of
+            // updated entities for every deletion). This query, however, is
+            // not particularly performance-sensitive
+            if !inserts_or_updates
+                .iter()
+                .any(|update| update.entity_type() == data.entity_type())
```
lutter, line 600: I don't understand the logic here - `entity_type()` is something like `Account` or `Token`, but the comparison tries to find which entities were truly deleted rather than updated (delete + insert) The comparison therefore would also have to take the entity's id into account.

graph/src/components/store.rs
```diff
@@ -1344,13 +1354,13 @@ pub trait StatusStore: Send + Sync + 'static {
 /// `EntityOperation`, we already know whether a `Set` should be an `Insert`
 /// or `Update`
 #[derive(Clone, Debug, PartialEq, Eq)]
-pub enum EntityModification {
+pub enum EntityModification<K = EntityKey> {
```
lutter, line 1357: What's the reason to not use a full `EntityKey` for this PR? I think genericizing this actually makes the code less clear.

server/index-node/src/schema.graphql
```diff
@@ -57,6 +59,21 @@ type EthereumIndexingStatus implements ChainIndexingStatus {
   lastHealthyBlock: Block
 }
 
+type EntityChanges {
+  updates: [EntityData!]!
+  deletes: [String!]!
```
lutter, line 64: For `deletes` you need to report both the entity type and the entity id.

server/index-node/src/schema.graphql
```diff
@@ -57,6 +59,21 @@ type EthereumIndexingStatus implements ChainIndexingStatus {
...
+}
+
+type EntityData {
+  name: String!
```
lutter, line 68: AFAICT, this is the entity type. It might be clearer to call this field `entity_type` then.

server/index-node/src/schema.graphql
```diff
@@ -57,6 +59,21 @@ type EthereumIndexingStatus implements ChainIndexingStatus {
...
+  fields: [EntityField!]!
+}
+
+type EntityField {
+  field: String!
+  valueAsJSON: String!
+}
+
```
lutter, line 76: Thinking about the structure of the response a little more, I think we'd want something like
```
{
  "data": {
     "updates": {
        "EntityType": [
            { "id": "..", "attr1": "..", .. },
            { "id": "..", "attr1": "..", .. },
            ..
         ],
        ..
     },
     "deletes": {
        "EntityType": [ "id1", "id2", ... ],
        ..
     }
  }
}
```

With that, attribute values could be serialized with the normal rules for `r::Value`. That would require creating a custom `scalar JSON` type in the GraphQL schema (I think that would just be a declaration)

A lot of this hinges on how this data will be consumed; the other option is to return each entity as a string that is a JSON blob with the entity's data in it, but that's a lot less friendly to consume for clients since they have to parse each entity twice. @Jannis how do you want to consume the responses here?

lutter (changes requested):

store/postgres/src/relational_queries.rs
neysofu: Makes sense, let's `select` as little data as possible. I wanted to reuse types we already had and compromise on query performance here, but I guess it's a good idea.

store/postgres/src/relational.rs
neysofu: I was under the impression that, for a given block number and entity type, you'd have at most **one** row that is either updated/inserted/deleted. I.e. you can't have two entities with the same entity type deleted at the same block. That would mean that we don't need entity ID comparisons.

graph/src/components/store.rs
neysofu: Lots of unnecessary data redundancy, mainly. We don't need to know `EntityKey::subgraph_id` for every entity because it's an invariant.

By the way, now that I'm thinking about it, it would be best to actually use `EntityOperation` rather than `EntityModification`. Still, the discussion above stands still.

server/index-node/src/schema.graphql
neysofu: Sure!

server/index-node/src/schema.graphql
neysofu: Yup, JSON blobs are not as user-friendly. I'm not familiar with GraphQL custom scalars, but if it doesn't require lots of extra code and we can reuse `r::Value` I can easily do that.

store/postgres/src/relational.rs
neysofu: Indeed, block ranges for any given entity don't overlap, so we'd never have more than a single entity per entity type here. Am I missing something?

Jannis: > I was under the impression that, for a given block number and entity type, you'd have at most one row that is either updated/inserted/deleted.

Any number of entities can be deleted / updated / inserted in a block.

neysofu: Indeed, but you can't have two entities that are deleted / updated / inserted in a block with the same entity type, or can you?

server/index-node/src/schema.graphql
Jannis: GraphQL strongly types the keys in objects like `"updates"`, so we can't use arbitrary entity type names as keys — unless we make the schema something like

```graphql
type EntityChanges {
  updates: JSONObject
  deletes: JSONObject
}
```
which is almost like using an `Any` type of sorts. The best I think we can do is something like this:

```graphql
type EntityChanges! {
  updates: [EntityTypeUpdates!]!
  deletions: [EntityTypeDeletions!]!
}

type EntityTypeUpdates {
  type: String!
  entities: [JSONObject!]!
}

type EntityTypeDeletions {
  type: String!
  entities: [ID!]!
}
```

which would return something like

```json
{
  "data": {
     "updates": [
        {
          "type": "EntityType",
          "entities": [
            { "id": "..", "attr1": "..", .. },
            { "id": "..", "attr1": "..", .. },
            ..
          ],
          }
        ..
     ],
     "deletes": {
        { "type": "EntityType", "entities": [ "id1", "id2", .. ] },
        ..
     }
  }
}
```

---

If we wanted the result to be even more specific (wrt types), we could replace the `JSONObject` type with something like

```graphql
type ChangedEntity {
  type: String!
  fields: [ChangedEntityField!]!
}

type ChangedEntityField {
  name: String!
  value: ChangedValue!
}

interface ChangedValue {
  type: String!
}

type ChangedBigIntValue implements ChangedValue {
  type: String!
  value: BigInt!
}

type ChangedStringValue implements ChangedValue {
  type: String!
  value: String
}

type ChangedListValue implements ChangedValue {
  type: String!
  value: [ChangedValue!]
}

# Repeat for all possible value types.
```

But I think that's overkill. I'd be ok with entities being represented with `scalar JSON`, which we should just be able to add at the top of the GraphQL schema.

Jannis: Having said all this, I like Filippo's proposed schema with fields in a list, having a `name: String!` and `value: JSON` field. I think that's a good compromise because the response structure is typed down to the field values — which could really be anything.

Jannis: To me `Vec<{ key: String, value: Any }>` and `Map<String, Any>` are more or less the same. I can easily build a map from a list of key/value pairs returned by GraphQL on the other side.

server/index-node/src/schema.graphql
```diff
@@ -57,6 +59,21 @@ type EthereumIndexingStatus implements ChainIndexingStatus {
...
+  deletes: [String!]!
+}
+
+type EntityData {
+  entityType: String!
+  fields: [EntityField!]!
+}
+
+type EntityField {
+  field: String!
```
Jannis: I'd prefer `name`

server/index-node/src/schema.graphql
```diff
@@ -57,6 +59,21 @@ type EthereumIndexingStatus implements ChainIndexingStatus {
...
+  valueAsJSON: String!
```
Jannis: I'd prefer `value: JSON!`.

server/index-node/src/schema.graphql
```diff
@@ -57,6 +59,21 @@ type EthereumIndexingStatus implements ChainIndexingStatus {
   lastHealthyBlock: Block
 }
 
+type EntityChanges {
+  updates: [EntityData!]!
+  deletes: [String!]!
+}
+
+type EntityData {
+  entityType: String!
```
Jannis, line 69: `type` would be my preference.

graph/src/components/store.rs
neysofu: Using `EntityOperation` + `EntityKey` now. Marking this as resolved.

store/postgres/src/relational_queries.rs
neysofu: Fixed.

server/index-node/src/schema.graphql
neysofu: Having the map of entities themselves as a JSON blob is quite bad and we lose a lot of GraphQL type information, I agree. That's why I'm listing entities in a `r::List`. Fields themselves can and should use JSON to not complicate things unnecessarily though :+1:

store/postgres/src/relational.rs
Jannis: Ah, no. Entities are either deleted or updated/inserted but not both in the same block. (At least not when looking at the database. They may be deleted and recreated when processing the block but that would not show here because these operations are rolled into one before transacting to the database.)

Jannis: But if it's either or, then you can just assume that the two sets (updates, deletes) are disjoint and you don't need the inner loop. Instead you should be able to take all deletions as they are, no?

server/index-node/src/schema.graphql
neysofu: I still don't understand why we expect multiple entities with the same entity type to be returned from this query.

Jannis (changes requested):

server/index-node/src/schema.graphql
Jannis: We don't. But then we don't need to check the two sets for collisions, right?

neysofu: I settled on this API:

```
type EntityChanges! {
  updates: [EntityTypeUpdates!]!
  deletions: [EntityTypeDeletions!]!
}

type EntityTypeUpdates {
  type: String!
  entities: [JSONObject!]!
}

type EntityTypeDeletions {
  type: String!
  entities: [ID!]!
}
```

Marking this as resolved and continuing the discussion about update/removal cross-checking to another comment.

server/index-node/src/schema.graphql
neysofu: :+1:

server/index-node/src/schema.graphql
neysofu: Moved it all inside a `JSONObject`; marking as resolved.

server/index-node/src/schema.graphql
neysofu: Done, see https://github.com/graphprotocol/graph-node/pull/3275#discussion_r816660558. Marking as resolved.

server/index-node/src/schema.graphql
```diff
@@ -19,6 +20,8 @@ type Query {
     indexer: Bytes
   ): Bytes
   subgraphFeatures(subgraphId: String!): SubgraphFeatures!
+  subgraph(subgraphId: String!): SubgraphFeatures!
```
Jannis, line 23: I think this was added by accident?

Jannis (changes requested):

graph/src/data/value.rs
```diff
@@ -215,6 +215,7 @@ impl Value {
             ("Bytes", Value::String(s)) => Ok(Value::String(s)),
             ("BigInt", Value::String(s)) => Ok(Value::String(s)),
             ("BigInt", Value::Int(n)) => Ok(Value::String(n.to_string())),
+            ("JSONObject", Value::Object(obj)) => Ok(Value::Object(obj)),
```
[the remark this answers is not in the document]
neysofu: @jannis Can you clarify a bit what you mean here? I believe scalars can be any GraphQL values. I don't have any automatic tests here, but querying `entityChangesInBlock` happily returns some GraphQL objects as `JSONObject`. This is kinda what https://github.com/taion/graphql-type-json does.

server/index-node/src/schema.graphql
neysofu: Ops! Indeed :smile:

server/index-node/src/resolver.rs
```diff
@@ -338,6 +359,75 @@ where
...
+        let map = BTreeMap::from([
+            (
+                "type".to_string(),
+                r::Value::String(entity_type.to_string()),
+            ),
+            (
+                "entities".to_string(),
+                r::Value::List(ids.into_iter().map(r::Value::String).collect()),
+            ),
+        ]);
```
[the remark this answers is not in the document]
neysofu: Amazing, didn't know about this.

server/index-node/src/resolver.rs
```diff
@@ -338,6 +359,75 @@ where
...
+                                    .into_iter()
+                                    .map(|(name, value)| (name, value.into()))
+                                    .collect(),
+                            )
+                        })
+                        .collect(),
+                ),
+            ),
+        ]);
+        updates_graphql.push(r::Value::object(map));
```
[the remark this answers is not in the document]
neysofu: :+1: I didn't realize `object!` was so useful.

store/postgres/src/relational.rs
neysofu: Addressed with proper documentation, comments, and performing lookups of duplicates with a `HashSet`.

Jannis (changes requested):
One last request: can we ensure that all returned values are sorted deterministically so we don't have to do it in the client? I see we're doing it for entity fields, but not for entities, entity types, deletions etc.

neysofu:
@jannis I'm also removing `__typename` since it's redundant data.

Jannis (approved):
Fantastic!

PR #3276, opened 2022-02-24
runtime: Improve error message for error initing AS global

leoyvens:
Previously such errors in the WASM start function would generate a `"WASM runtime thread terminated abnormally"` with newlines in it, making it impossible to search for in logging tools. Now it properly bubbles up.

### DIFF
--- a/core/src/subgraph/runner.rs
+++ b/core/src/subgraph/runner.rs
@@ -425,7 +425,8 @@ where
                                     .unwrap()
                                     .remove(&self.inputs.deployment.id);
 
-                                error!(logger, "Subgraph failed with non-deterministic error: {}", e;
+                                let message = format!("{:#}", e).replace("\n", "\t");
+                                error!(logger, "Subgraph failed with non-deterministic error: {}", message;
                                     "attempt" => backoff.attempt,
                                     "retry_delay_s" => backoff.delay().as_secs());
 
--- a/runtime/wasm/src/mapping.rs
+++ b/runtime/wasm/src/mapping.rs
@@ -61,25 +61,15 @@ pub fn spawn_module<C: Blockchain>(
                     trigger,
                     result_sender,
                 } = request;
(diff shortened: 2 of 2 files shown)


core/src/subgraph/runner.rs
```diff
@@ -425,7 +425,8 @@ where
                                     .unwrap()
                                     .remove(&self.inputs.deployment.id);
 
-                                error!(logger, "Subgraph failed with non-deterministic error: {}", e;
+                                let message = format!("{:#}", e).replace("\n", "\t");
+                                error!(logger, "Subgraph failed with non-deterministic error: {}", message;
```
leoyvens, line 429: Here the message is formatted as in other places.

runtime/wasm/src/mapping.rs
```diff
@@ -61,25 +61,15 @@ pub fn spawn_module<C: Blockchain>(
...
-                let section = host_metrics.stopwatch.start_section("module_init");
-                let module = WasmInstance::from_valid_module_with_ctx(
+                let result = instantiate_module_and_handle_trigger(
                     valid_module.cheap_clone(),
                     ctx,
+                    trigger,
                     host_metrics.cheap_clone(),
                     timeout,
                     experimental_features,
-                )?;
```
leoyvens, line 74: This early return was the issue, I fixed it by refactoring this into a separate fn.

core/src/subgraph/runner.rs
lutter: Is that from using anyhow's `with_context`? That doesn't play nice with keeping things on one log line.

leoyvens: The `{:#}` is to print anyhow causes, in a single line. Still newlines must be replaced because wasm traps included traces with newlines.

otaviopace (approved):

Issue #3257, opened 2022-02-18
Improve block stream pipelining with dynamic data sources


Jannis:
Two ideas:
- Reduce block stream pipeline size to zero if we're creating a lot of dynamic data sources
- Avoid restarting the block stream by using broader filters that make it unnecessary

leoyvens:
Because this regresses sparse subgraphs with templates, we plan to turn this on by default only when using firehose.

leoyvens:
This been implemented as intended so this can be closed.

PR #3277, opened 2022-02-24
subgraph: Add template handlers to filters

mangas:
Fixes https://github.com/graphprotocol/graph-node/issues/3257

### DIFF
--- a/chain/ethereum/src/adapter.rs
+++ b/chain/ethereum/src/adapter.rs
@@ -22,6 +22,7 @@ use graph::{
 
 use crate::capabilities::NodeCapabilities;
 use crate::data_source::BlockHandlerFilter;
+use crate::Mapping;
 use crate::{data_source::DataSource, Chain};
 
 pub type EventSignature = H256;
@@ -143,6 +144,22 @@ impl bc::TriggerFilter<Chain> for TriggerFilter {
             traces: self.requires_traces(),
         }
     }
+
+    fn extend_with_template(
+        &mut self,
+        data_sources: impl Iterator<Item = <Chain as bc::Blockchain>::DataSourceTemplate>,
+    ) {
+        for data_source in data_sources.into_iter() {
+            self.log
+                .extend(EthereumLogFilter::from_mapping(&data_source.mapping));
+
+            self.call
+                .extend(EthereumCallFilter::from_mapping(&data_source.mapping));
+
+            self.block
+                .extend(EthereumBlockFilter::from_mapping(&data_source.mapping));
+        }
+    }
 }
 
 #[derive(Clone, Debug, Default)]
@@ -200,8 +217,22 @@ impl EthereumLogFilter {
         this
     }
 
+    pub fn from_mapping(iter: &Mapping) -> Self {
+        let mut this = EthereumLogFilter::default();
+
+        for sig in iter.event_handlers.iter().map(|e| e.topic0()) {
+            this.wildcard_events.insert(sig);
+        }
+
+        this
+    }
+
     /// Extends this log filter with another one.
     pub fn extend(&mut self, other: EthereumLogFilter) {
+        if other.is_empty() {
+            return;
+        };
+
         // Destructure to make sure we're checking all fields.
         let EthereumLogFilter {
             contracts_and_events_graph,
@@ -292,6 +323,8 @@ pub(crate) struct EthereumCallFilter {
     // start_block and the set of function signatures
     pub contract_addresses_function_signatures:
         HashMap<Address, (BlockNumber, HashSet<FunctionSelector>)>,
+
+    pub wildcard_signatures: HashSet<FunctionSelector>,
 }
 
 impl EthereumCallFilter {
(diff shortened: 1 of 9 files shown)


chain/ethereum/src/chain.rs
```diff
@@ -51,7 +51,7 @@ lazy_static! {
     static ref MAX_BLOCK_RANGE_SIZE: BlockNumber = env_var("GRAPH_ETHEREUM_MAX_BLOCK_RANGE_SIZE", 2000);
 
     /// Ideal number of triggers in a range. The range size will adapt to try to meet this.
-    static ref TARGET_TRIGGERS_PER_BLOCK_RANGE: u64 = env_var("GRAPH_ETHEREUM_TARGET_TRIGGERS_PER_BLOCK_RANGE", 100);
+    static ref TARGET_TRIGGERS_PER_BLOCK_RANGE: u64 = env_var("GRAPH_ETHEREUM_TARGET_TRIGGERS_PER_BLOCK_RANGE", 1000);
```
otaviopace, line 54: I'm not sure I understood why this changed and how it will behave with template handlers for filters.

chain/ethereum/src/adapter.rs
```diff
@@ -305,34 +330,62 @@ impl EthereumCallFilter {
...
+            // If the call is to a contract with no specified functions, keep the call
+            //
+            // Allows the ability to genericly match on all calls to a contract.
+            // Caveat is this catch all clause limits you from matching with a specific call
+            // on the same address
+            Some(v) if v.1.is_empty() => return true,
+            // There are some relevant signatures to test
+            Some(v) => Some(&v.1),
+            // There are no signatures and no wildcards
+            None if self.wildcard_matches.is_empty() => return false,
```
otaviopace, line 342: I'm not sure I get why this is `false` 🤔

chain/ethereum/src/chain.rs
mangas: https://github.com/graphprotocol/graph-node/blob/465e7ad1a9e90cd529620318e37688a2eb784579/graph/src/blockchain/polling_block_stream.rs#L379

With the increase in triggers, the ranges went down to 1-10 and this more or less doubles the ranges

graph/src/data/schema.rs
```diff
@@ -847,6 +847,8 @@ impl Schema {
         &self,
         schemas: &HashMap<SchemaReference, Arc<Schema>>,
     ) -> Result<(), Vec<SchemaValidationError>> {
+        // Using this since std::array's .into_iter() doesn't work
+        // as expected (at least as of rustc 1.53)
```
leoyvens, line 851: This is a rebase artifact, this comment should be removed.

leoyvens (changes requested):
If `EthereumCallFilter::extend` won't affect the wildcard filter, then it should assert that the filter is empty.

chain/ethereum/src/adapter.rs
```diff
@@ -143,6 +144,18 @@ impl bc::TriggerFilter<Chain> for TriggerFilter {
...
+        &mut self,
+        data_sources: impl Iterator<Item = <Chain as bc::Blockchain>::DataSourceTemplate>,
+    ) {
+        for data_source in data_sources.into_iter() {
+            self.log
+                .extend(EthereumLogFilter::from_mapping(&data_source.mapping));
+            self.call
+                .extend(EthereumCallFilter::from_mapping(&data_source.mapping));
+        }
+    }
```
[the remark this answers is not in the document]
mangas: this would mean essentially turning it on for every block no? that's why I didn't add it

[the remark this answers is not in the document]
leoyvens: Only if the template actually contains block handlers.

chain/ethereum/src/adapter.rs
```diff
@@ -143,6 +144,23 @@ impl bc::TriggerFilter<Chain> for TriggerFilter {
...
+        for data_source in data_sources.into_iter() {
+            self.log
+                .extend(EthereumLogFilter::from_mapping(&data_source.mapping));
+            self.call
+                .extend(EthereumCallFilter::from_mapping(&data_source.mapping));
+
+            let block = EthereumBlockFilter::from_mapping(&data_source.mapping);
+            if !block.is_empty() {
+                self.block.extend(block);
+            }
```
leoyvens, line 161: Extending with an empty `block` is a noop. If the `if` is meant as an optimization, then it can be implemented within `extend`.

chain/ethereum/src/adapter.rs
```diff
@@ -433,6 +495,13 @@ pub(crate) struct EthereumBlockFilter {
 }
 
 impl EthereumBlockFilter {
+    pub fn from_mapping(mapping: &Mapping) -> Self {
+        Self {
+            contract_addresses: HashSet::new(),
+            trigger_every_block: mapping.block_handlers.len() != 0,
+        }
```
leoyvens, line 502: This isn't taking into account call filters on blocks https://thegraph.com/docs/en/developer/create-subgraph-hosted/#supported-filters

mangas: Is this available for templates as well?

leoyvens: It is afaik, you can test it out to be sure

mangas: I think we need to have this broader filter for now in this version, we can optimise it later. 

Making this change is going to require wider changes, the Block filter properties are used directly instead of using a match statement and the firehose filters don't support this either

node/src/main.rs
```diff
@@ -320,13 +320,18 @@ async fn main() {
             );
             graph::spawn_blocking(job_runner.start());
         }
+        let restart_on_dynamic_data_source =
+            !env::var_os("EXPERIMENTAL_DISABLE_RESTART_ON_DYNAMIC_DS")
+                .map(|_| true)
+                .unwrap_or(false);
```
leoyvens, line 326: What do you think of trying to give this change a more meaningful and shorter name. Perhaps "static filters"? This would mean renaming `restart_on_dynamic_data_source` to `static_filters_enabled` and flipping the conditionals since enabling static filters means restarts are disabled.

mangas: sounds good!

leoyvens (approved):

PR #3283, opened 2022-02-25
to_vec access modifier changed for TypedArray enum

georg-getz:


georg-getz:
@leoyvens should be fine now.

otaviopace (approved):

PR #3290, opened 2022-02-26
store: Reduce number of queries to update subgraph_deployment

lutter:
When updating the metadata in subgraph_deployment, we currently do two to
three roundtrips to the database to

  - update the block pointer
  - update the entity count
  - update the firehose cursor (sometimes)

With this change, all three roundtrips are folded into one update
statement.

### DIFF
--- a/store/postgres/src/deployment.rs
+++ b/store/postgres/src/deployment.rs
@@ -251,6 +251,76 @@ pub fn features(conn: &PgConnection, site: &Site) -> Result<BTreeSet<SubgraphFea
         .collect()
 }
 
+pub fn transact_block(
+    conn: &PgConnection,
+    site: &Site,
+    ptr: &BlockPtr,
+    firehose_cursor: Option<&str>,
+    full_count_query: &str,
+    count: i32,
+) -> Result<(), StoreError> {
+    use crate::diesel::BoolExpressionMethods;
+    use subgraph_deployment as d;
+
+    // Work around a Diesel issue with serializing BigDecimals to numeric
+    let number = format!("{}::numeric", ptr.number);
+
+    let count_sql = if count == 0 {
+        // This amounts to a noop - the entity count does not change
+        "entity_count".to_string()
+    } else {
+        entity_count_sql(full_count_query, count)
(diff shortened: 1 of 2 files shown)


store/postgres/src/deployment.rs
```diff
@@ -251,6 +251,86 @@ pub fn features(conn: &PgConnection, site: &Site) -> Result<BTreeSet<SubgraphFea
...
+        // This amounts to a noop - the entity count does not change
+        "entity_count".to_string()
+    } else {
+        format!(
+            "coalesce((nullif(entity_count, -1)) + ({count}),
+                  ({full_count_query}))",
+            full_count_query = full_count_query,
+            count = count
+        )
+    };
```
leoyvens, line 292: This duplication with `fn update_entity_count` is unfortunate, would it be possible to not duplicate this code?

leoyvens (commented):
This will now write the firehose cursor when its value is `""`, ping @maoueh to confirm that the `!= ""` check was just an optimization and not necessary for correctness.

maoueh:
@leoyvens @lutter It's all good to write the cursor if it's `""`, a Firehose cursor of value `""` is semantically equal to no cursor at all. Ultimately, at the request level, we send `""` to say to the service there is no cursor, so it's all good.

The thing is will change however is that it will effectively delete the firehose cursor it it existed. But that's good since upcoming PR #3174 was doing it already. 

Either this PR or #3174 will need to be updated to remove the https://github.com/graphprotocol/graph-node/pull/3174/files#diff-ec9db820ff81f5904c7de003a414c9756dd8eb81b539122f0a70811752f2f313R125

store/postgres/src/deployment.rs
lutter: I factored that scary comment and string generation into a separate function.

lutter:
I also changed the logic of `deployment::transact_block` to set the firehose cursor to `null` when it is `Some("")` - even if it's all the same to the firehose, it's cleaner that way.

leoyvens (approved):

PR #3291, opened 2022-02-26
runtime: Only call EntityCache.get once in store_get

lutter:


lutter:
> oops, nice catch

probably not a big deal in practice as it just refetches from the catch, but it does look better w/o the extra call ;)

Issue #3236, opened 2022-02-09
Errored subgraphs sometimes resume correctly after restart


aasseman:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
Graph-node v0.25.0.

Some subgraph's indexing would hang up, declaring a deterministic error. But after one (or more) restarts of the graph-node, would resume with seemingly no issues. Observed on `QmTBxvMF6YnbT1eYeRx9XQpH4WvxTV53vdptCCZFiZSprg` (Enzyme) and `Qmf5XXWA8zhHbdvWqtPcR3jFkmb5FLR4MAefEYx8E3pHfr` (Graph network).

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem.**
Index `QmTBxvMF6YnbT1eYeRx9XQpH4WvxTV53vdptCCZFiZSprg` or `Qmf5XXWA8zhHbdvWqtPcR3jFkmb5FLR4MAefEYx8E3pHfr`.

**What is the expected behavior?**
The error should *really* be deterministic. Here it seems that due to unknown factors, a subgraph can be stuck, then unstuck.

Logs for `QmTBxvMF6YnbT1eYeRx9XQpH4WvxTV53vdptCCZFiZSprg`:
```
Feb 08 03:02:22.060 DEBG Trying again after eth_call RPC call failed (attempt #1) with result Err(Web3Error(Transport("Error(IncompleteMessage)"))), block_hash: 0xc3b82bc6897a2f490c757dec90fa5ac07043b64e7e442822f8427b240ba8641e, block_number: 14162904, sgd: 10, subgraph_id: QmTBxvMF6YnbT1eYeRx9XQpH4WvxTV53vdptCCZFiZSprg, component: SubgraphInstanceManager 
[...]

PR #3292, opened 2022-02-26
Improve debug forking

otaviopace:
I do accept any suggestions on the CLI API, I'm not sure those parameters should be passed where I've put them (and with those names).

This PR adds:

- `--debug-fork` CLI option
- `--start-block` CLI option

With those you can basically get **any subgraph** from the Hosted Service and debug it at **any block** without having the source code to change the **start block**, which debug forking currently requires ([docs](https://thegraph.com/docs/en/developer/subgraph-debug-forking/)).

Eg (enzyme fork that I've been able to reproduce #3236) :

```
cargo run -p graph-node -- \
          --postgres-url $THEGRAPH_STORE_POSTGRES_DIESEL_URL \
          --ethereum-rpc $ETHEREUM_RPC_URL \
          --subgraph QmTBxvMF6YnbT1eYeRx9XQpH4WvxTV53vdptCCZFiZSprg \
          --ipfs $IPFS_URL \
          --debug-fork QmTBxvMF6YnbT1eYeRx9XQpH4WvxTV53vdptCCZFiZSprg \
          --start-block 0x8e28f0485a3bbcca851d502c2ca1120607b7178bf997ca8a40f4a457b8acb964:14178262 \
          --fork-base $FORK_BASE_URL
```

I know a lot of parameters, that's why I'm not sure if this is the way to go with the API design.

@VIVelev if you can give any suggestions here, thanks a lot your tool is super useful ❤️

### DIFF
--- a/node/src/main.rs
+++ b/node/src/main.rs
@@ -425,13 +425,31 @@ async fn main() {
                 .debug_fork
                 .map(DeploymentHash::new)
                 .map(|h| h.expect("Debug fork hash must be a valid IPFS hash"));
+            let start_block = opt
+                .start_block
+                .map(|block| {
+                    let mut split = block.split(":");
+                    (
+                        // BlockHash
+                        split.next().unwrap().to_owned(),
+                        // BlockNumber
+                        split.next().unwrap().parse::<i64>().unwrap(),
+                    )
+                })
+                .map(|(hash, number)| BlockPtr::try_from((hash.as_str(), number)))
+                .map(Result::unwrap);
 
             graph::spawn(
                 async move {
                     subgraph_registrar.create_subgraph(name.clone()).await?;
                     subgraph_registrar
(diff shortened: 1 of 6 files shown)


node/src/main.rs
```diff
@@ -389,13 +389,35 @@ async fn main() {
...
+                .map(|block| {
+                    let mut split = block.split(":");
+                    // (BlockHash, BlockNumber)
+                    (
+                        split.next().unwrap().to_owned(),
+                        split.next().unwrap().parse::<i64>().unwrap(),
+                    )
+                })
+                .map(|(hash, number)| BlockPtr::try_from((hash.as_str(), number)))
+                .map(Result::unwrap);
```
otaviopace, line 407: Mostly I need to clean this part up.

mangas (approved):

PR #3294, opened 2022-02-26
store/graph: add insensitive matching filters

### LINKED ISSUE #3293: Add new filters for case insensitive match `_contains_insensitive` , `_starts_with_insensitive`, `_ends_with_insensitive`
Reasonable on breaking existing dapps - @lutter @saihaj @dotansimha what do you think about introducing a new `i_contains`, `i_starts_with` option in the `where`, to provide specific case insensitive behaviour? My instinct is that the case-sensitivity is not what most users would expect, but I can't think of an easy way to validate if dapps are relying on it.

We are addressing the database locale issue separately: https://github.com/graphprotocol/graph-node/issues/3118

_Originally posted by @azf20 in https://github.com/graphprotocol/graph-node/issues/3191#issuecomment-1048875273_

saihaj:
Closes #3293﻿
Precede https://github.com/graphprotocol/graph-node/pull/3191


store/postgres/src/relational_queries.rs
```diff
@@ -818,9 +820,13 @@ impl<'a> QueryFilter<'a> {
...
             | StartsWith(attr, _)
+            | StartsWithInsensitive(attr, _)
             | NotStartsWith(attr, _)
+            | NotStartsWithInsensitive(attr, _)
             | EndsWith(attr, _)
-            | NotEndsWith(attr, _) => {
+            | EndsWithInsensitive(attr, _)
+            | NotEndsWith(attr, _)
+            | NotEndsWithInsensitive(attr, _) => {
                 table.column_for_field(attr)?;
```
tilacog, line 830: LGTM but it might be a sign that we could use a `EntityFilter::attr(&self) -> Option<Attribute>` getter method for conciseness.  

Maybe we could open a new issue for this.

store/postgres/src/relational_queries.rs
```diff
@@ -867,18 +873,20 @@ impl<'a> QueryFilter<'a> {
...
+        };
         match value {
             Value::String(s) => {
                 out.push_identifier(column.name.as_str())?;
-                if negated {
-                    out.push_sql(" not like ");
-                } else {
-                    out.push_sql(" like ")
-                };
+                out.push_sql(operation);
```
tilacog, line 889: Nice!

tilacog (approved):

lutter:
This might be bikeshedding, but could we rename the filters from `_insensitive` to `_nocase`? That sounds a little more natural to me, and I hate being insensitive ;)

saihaj:
updated @lutter :)

lutter (approved):
Thanks for making the change!

saihaj:
> @saihaj can this be merged?

Yes @otaviopace I don't know if I am allowed to merge or if there are blog post or something we would want for this?

PR #3296, opened 2022-02-27
store: Correctly find last vid for copying from immutable tables

lutter:


leoyvens (approved):

lutter:
> This is a mini fix for #3201 right?

Yes, it was a place where we didn't take the difference in db schema into account

Issue #2980, opened 2021-11-18
Filter an Ethereum Firehose for addresses of interest


azf20:
In order to better support syncing sparse subgraphs (where there are not many events), Graph Node should be able to ask the firehose only for blocks that include calls to a contract of interest. This would radically reduce the number of blocks that the graph node needs to process in order to sync the subgraph.

- This should be a simple filter to optimise for performance
- When syncing the chain head, graph node will need to update the subgraph's block pointer

This will likely require a similar implementation to #2942 

- [x] Investigation (Firehose, BlockStream, filters, ...)
- [ ] Implementation
- [ ] Run/test against real Firehose

otaviopace:
In a brief chat with @sduchesneau he mentioned that they still need to implement a filter in the Firehose side (they have a similar old thing there, but it'll probably not apply to what we need today).

The format passed would be something like this:
```
to in ["0xaaaaaaaaaaaaaa","0xbbbbbbbbbbbb...",...]
```

Where `to` is the address of the contract called in any call/subcall. And any transaction that matches this contract (`to`), will be excluded from the Firehose block (or maybe even the full block, but this might be an optimization for later).

@sduchesneau feel free to correct me if I said something wrong 🙂

otaviopace:
## Notes

After further code investigation I've noted a couple of things:

- In our `chain/ethereum` crate we have a [`TriggerFilter` structure](https://github.com/graphprotocol/graph-node/blob/8155ed5299e96a79a1554642486bcf88254b4a59/chain/ethereum/src/adapter.rs#L118) that contains the addresses, events, etc related to a subgraph (originated from it's data sources). The data that will be sent to Firehose will come from here.
- The [`firehose::BlocksRequestV2` structure](https://github.com/graphprotocol/graph-node/blob/8155ed5299e96a79a1554642486bcf88254b4a59/graph/src/firehose/dfuse.bstream.v1.rs#L39) will need to accept these addresses/contracts.
- And the [`firehose::BlockStream`](https://github.com/graphprotocol/graph-node/blob/8155ed5299e96a79a1554642486bcf88254b4a59/graph/src/blockchain/firehose_block_stream.rs#L126) will need to extract the addresses from the `TriggerFilter`, to send them in the `BlocksRequestV2` structure.

## Technical questions

[...]

PR #3298, opened 2022-02-28
add firehose transform proto

mangas:
Part of https://github.com/graphprotocol/graph-node/issues/2980


sduchesneau:
You don't need the indexes.proto file, it is used only internally in the firehose

PR #3299, opened 2022-02-28
add --ipfs flag to 'graphman run' to control the IPFS URL

sduchesneau:
One can now run:

```
graphman --ipfs=localhost:5001 run {...}
```
or leave the default value `https://api.thegraph.com/ipfs/`

### DIFF
--- a/node/src/bin/manager.rs
+++ b/node/src/bin/manager.rs
@@ -63,6 +63,14 @@ pub struct Opt {
     pub node_id: String,
     #[structopt(
         long,
+        value_name = "{HOST:PORT|URL}",
+        default_value = "https://api.thegraph.com/ipfs/",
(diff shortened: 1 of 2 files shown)


leoyvens:
This should imo not have a default value, but I suppose we could mention the url to our gateway in the help string.

sduchesneau:
> This should imo not have a default value, but I suppose we could mention the url to our gateway in the help string

I used the default value set by @mangas , my intention was to simplify the life of the developer trying to "run a subgraph locally for a test" or something similar.
I don't mind removing the default, but it does seem like making it harder for the dev to use the tool.

leoyvens:
Ok I don't feel strongly.

leoyvens (approved):

PR #3304, opened 2022-03-01
Fix some Clippy warnings and wasteful clones

neysofu:
I went through most of Clippy's warning for the workspace and I fixed some classes of warnings that seemed reasonable.

- Unnecessary clones.
- Unnecessary references passed as parameters to function calls (e.g. `PgConnection`, `i32`).
- Wasteful consumption of iterators (e.g. via `.count() > 0`, `.find(...).is_some()`).
- Seemingly unnecessary conversions between `BlockNumber` and `i32`, which are equivalent (@lutter is this to improve clarity and explicitness or a leftover from previous type definitions? I can revert this if there's a reason for that).
- Replacing most `&Vec<T>` arguments with `&[T]`.
- Replacing some pattern matching with more succinct alternatives.

### DIFF
--- a/store/postgres/src/catalog.rs
+++ b/store/postgres/src/catalog.rs
@@ -214,7 +214,7 @@ pub fn account_like(conn: &PgConnection, site: &Site) -> Result<HashSet<String>,
         .select((ts::table_name, ts::is_account_like))
         .get_results::<(String, Option<bool>)>(conn)
         .optional()?
-        .unwrap_or_default()
+        .unwrap_or(vec![])
         .into_iter()
         .filter_map(|(name, account_like)| {
             if account_like == Some(true) {
--- a/graphql/src/schema/api.rs
+++ b/graphql/src/schema/api.rs
@@ -246,7 +246,7 @@ fn field_filter_input_values(
             })
         }
         Type::ListType(ref t) => {
-            Ok(field_list_filter_input_values(schema, field, t).unwrap_or_default())
+            Ok(field_list_filter_input_values(schema, field, t).unwrap_or(vec![]))
         }
         Type::NonNullType(ref t) => field_filter_input_values(schema, field, t),
     }
--- a/graphql/src/store/prefetch.rs
+++ b/graphql/src/store/prefetch.rs
@@ -424,7 +424,7 @@ impl<'a> Join<'a> {
             // query are always joined first, and may then be overwritten by the merged selection
             // set under the object type condition. See also: e0d6da3e-60cf-41a5-b83c-b60a7a766d4a
             let values = parent.id().ok().and_then(|id| grouped.get(&*id).cloned());
-            parent.set_children(response_key.to_owned(), values.unwrap_or_default());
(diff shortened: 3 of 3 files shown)


chain/near/src/trigger.rs
```diff
@@ -175,7 +176,8 @@ mod tests {
             receipt: receipt().unwrap(),
         }));
 
-        let result = blockchain::MappingTrigger::to_asc_ptr(trigger, &mut heap, &GasCounter::new());
+        let result =
+            blockchain::MappingTrigger::to_asc_ptr(trigger, &mut heap, &GasCounter::default());
```
leoyvens, line 180: I disagree with clippy here, nothing wrong with a `new` method that does the same thing as `default`.

neysofu: Yeah... the reasoning behind Clippy's suggestion is that `Default` integrates better with `std::mem::take`, `unwrap_or_default` and such, but I guess it's a preference thing, really.

neysofu: I'll define `GasCounter::new` as `GasCounter::default`, as to have both.

neysofu:
@dotansimha my only concern is that some lints are questionable and we definitely don't want to blindly `cargo clippy --fix`. Granted, we could have fine-grained control over which lints to run and which not to, but that's a bit of a hassle. I think we can add Clippy to the CI as long as it simply reports warnings and doesn't block builds. This way we can keep an eye out on antipatterns but not let it get in the way when we don't want to follow its suggestions.

store/postgres/src/catalog.rs
```diff
@@ -214,7 +214,7 @@ pub fn account_like(conn: &PgConnection, site: &Site) -> Result<HashSet<String>,
         .select((ts::table_name, ts::is_account_like))
         .get_results::<(String, Option<bool>)>(conn)
         .optional()?
-        .unwrap_or(vec![])
+        .unwrap_or_default()
```
leoyvens, line 217: I don't think this improves readability, `unwrap_or(vec![])` is shorter and documents the type.

neysofu: :+1:

dotansimha:
> @dotansimha my only concern is that some lints are questionable and we definitely don't want to blindly `cargo clippy --fix`. Granted, we could have fine-grained control over which lints to run and which not to, but that's a bit of a hassle. I think we can add Clippy to the CI as long as it simply reports warnings and doesn't block builds. This way we can keep an eye out on antipatterns but not let it get in the way when we don't want to follow its suggestions.

I guess maybe we can run it in CI without `--fix`? Just to have visibility on the issues and warnings, (and maybe in the future let it fail PRs in case they are violating some important lint rules?)

neysofu:
That would be nice to have IMHO.

leoyvens:
I'm for running clippy on CI, though we don't need to agree with every lint and can allow those we don't want.

neysofu:
Opened https://github.com/graphprotocol/graph-node/pull/3310 for CI talks.

leoyvens (approved):

PR #3308, opened 2022-03-02
ethereum: increase default reorg threshold to 250

leoyvens:
That's what we run on hosted.


lutter (approved):

lutter:
Just noticed that the comments for these env vars are now out of date - maybe just delete them since they aren't really helpful. Also, shockingly, these two vars are not documented in `docs/environment-variables.md` And finally, maybe we should just have one variable for both? I don't know why anybody would ever use `ancestor_count > reorg_threshold`

PR #3315, opened 2022-03-03
block stream: Cleanup block stream metrics

leoyvens:
Now that the block stream is pipelined, it should not be starting sections concurrently with the runner. This removes the `"filter_call_triggers_from_unsuccessful_transactions"` which was the only one within the block stream. The `deployment_head` metric is moved from the block stream to the runner. This made the metrics dead parameters in many places.

It's likely that we'll want to reintroduce stopwatch metrics in the blockstream, separating the metrics for each stage as done in #3177, but it seems best to remove them now and start from a clean slate when reintroducing them with pipelining and firehose in mind.


lutter (approved):
Excellent!

PR #3321, opened 2022-03-04
Add GraphQL route for fetching block data

### LINKED ISSUE #3311: GraphQL API for block data
When we have detected a difference in the indexing results of two indexers, we need to gather more details and contextual information to debug what could have led to that difference.

Differences are always starting at a specific block and one of the details that we need is the _block_ itself that graph-node used for processing the subgraph in question.

This is (roughly) the schema that we need:

```graphql
type Query {
  blockData(network: String!, blockNumber: Int!): JSONObject
}
```

We cannot guarantee that graph-node always has this information (hence the nullable return value) but typically it should. A generic `JSONObject` seems reasonable as the result because we may be dealing with Ethereum blocks, NEAR blocks and others and probably don't want to type these too strictly in this schema. We'd likely forget to keep these types in sync with the actual types used when processing.

graph-node stores Ethereum block data in an `ethereum_blocks` table. Not sure about NEAR blocks. Either way, there's probably a method in one of the stores to get the right data for a network.

neysofu:
Closes #3311. This is the type signature of the GraphQL route:

```graphql
type Query {
  blockData(network: String!, blockHash: Bytes!): JSONObject
}
```

I'm also introducing a new `Store` trait which maintains separation of concerns between `BlockStore` and `SubgraphStore` while replacing some clutter of generics with associated types within `index-node`.

### DIFF
--- a/server/index-node/src/resolver.rs
+++ b/server/index-node/src/resolver.rs
@@ -1,45 +1,48 @@
 use either::Either;
-use graph::blockchain::{Blockchain, BlockchainKind};
-use graph::components::store::EntityType;
-use graph::data::value::Object;
-
-use graph::data::subgraph::features::detect_features;
-use graph::data::subgraph::{status, MAX_SPEC_VERSION};
-use graph::prelude::*;
-use graph::{
-    components::store::StatusStore,
-    data::graphql::{object, IntoValue, ObjectOrInterface, ValueMap},
-};
-use graph_graphql::prelude::{a, ExecutionContext, Resolver};
 use std::collections::BTreeMap;
 use std::convert::TryInto;
 use web3::types::{Address, H256};
 
+use graph::blockchain::{Blockchain, BlockchainKind};
+use graph::data::subgraph::features::detect_features;
+use graph::data::subgraph::{status, MAX_SPEC_VERSION};
+use graph::data::value::Object;
+use graph::prelude::*;
+use graph::{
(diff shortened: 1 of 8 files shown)


server/index-node/src/resolver.rs
```diff
@@ -102,12 +93,62 @@ where
...
+                error!(
+                    self.logger,
+                    "Failed to fetch block data; block not found";
+                    "network" => network,
+                    "block_hash" => format!("{}", block_hash),
+                );
+                r::Value::Null
+            }
+            Ok(mut blocks) => {
+                assert!(blocks.len() == 1, "Multiple blocks with the same hash");
```
Jannis, line 136: I wouldn't want this to crash a node, so I'd prefer if we just logged this with `error!`.

Jannis (changes requested):

server/index-node/src/resolver.rs
neysofu: I believe `"blocks_pkey" PRIMARY KEY, btree (hash)` guarantees us that the node will never fetch multiple blocks from a single hash

neysofu: So basically the `assert!` is there not to log an error but rather to provide error information in case we have an actual bug.

Jannis: Ok, true. That's fine then.

Jannis (approved):

neysofu:
Rebasing and merging.

PR #3322, opened 2022-03-04
Split huge `graph::components::store` module

neysofu:
The module `graph::components::store`, which has a lot of responsibilities, is currently sitting at over 1800 lines and we can probably do a better job at managing all of that code. I moved all code in `store.rs` to a new `store/` directory, along with some child modules (`err.rs`, `cache.rs`, `traits.rs`) to make `store/mod.rs` itself a tad slimmer.


Jannis (approved):
Good idea!

neysofu:
This is really conflict-prone, so I'll merge immediately.

PR #3325, opened 2022-03-04
Fix revert and unfailure

otaviopace:
This PR fixes:

- Refactors `BlockStream` revert events to be handled by the `SubgraphRunner` instead of the `Store`;
- Deterministic unfailure, where we would end up erroring in `graph-node` restarts with error `DuplicateBlockProcessing`, that outputs `subgraph `QmHash` has already processed block 11310909; there are most likely two (or more) nodes indexing this subgraph`.

### DIFF
--- a/core/src/subgraph/runner.rs
+++ b/core/src/subgraph/runner.rs
@@ -185,8 +185,20 @@ where
 
                 let (block, cursor) = match event {
                     Some(Ok(BlockStreamEvent::ProcessBlock(block, cursor))) => (block, cursor),
-                    Some(Ok(BlockStreamEvent::Revert(subgraph_ptr, revert_to_ptr, cursor))) => {
-                        info!(&logger, "Reverting block to get back to main chain"; "current" => &subgraph_ptr, "revert_to" => &revert_to_ptr);
+                    // In the case of the PollingBlockStream, the stream_ptr can be ahead of the
+                    // subgraph's deployment head (database), because this component is pipelined.
+                    Some(Ok(BlockStreamEvent::Revert(stream_ptr, revert_to_ptr, cursor))) => {
+                        // Current deployment head in the database / WritableAgent Mutex cache.
+                        //
+                        // Safe unwrap because in a Revert event we're sure the subgraph has
+                        // advanced at least once.
+                        let subgraph_ptr = self.inputs.store.block_ptr().unwrap();
+                        if revert_to_ptr.number >= subgraph_ptr.number {
+                            info!(&logger, "Block to revert is higher than subgraph pointer, nothing to do"; "subgraph_ptr" => &subgraph_ptr, "stream_ptr" => &stream_ptr, "revert_to_ptr" => &revert_to_ptr);
+                            continue;
+                        }
+
+                        info!(&logger, "Reverting block to get back to main chain"; "subgraph_ptr" => &subgraph_ptr, "stream_ptr" => &stream_ptr, "revert_to_ptr" => &revert_to_ptr);
 
                         if let Err(e) = self
                             .inputs
--- a/store/postgres/src/deployment_store.rs
+++ b/store/postgres/src/deployment_store.rs
@@ -1135,25 +1135,17 @@ impl DeploymentStore {
         site: Arc<Site>,
         block_ptr_to: BlockPtr,
         firehose_cursor: Option<&str>,
-    ) -> Result<Option<StoreEvent>, StoreError> {
+    ) -> Result<StoreEvent, StoreError> {
         let conn = self.get_conn()?;
         // Unwrap: If we are reverting then the block ptr is not `None`.
         let deployment_head = Self::block_ptr_with_conn(&site.deployment, &conn)?.unwrap();
 
-        // Early return to cover scenario where the BlockStream advanced more than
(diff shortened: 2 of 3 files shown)


core/src/subgraph/runner.rs
```diff
@@ -185,8 +185,20 @@ where
...
-                    Some(Ok(BlockStreamEvent::Revert(subgraph_ptr, revert_to_ptr, cursor))) => {
-                        info!(&logger, "Reverting block to get back to main chain"; "current" => &subgraph_ptr, "revert_to" => &revert_to_ptr);
+                    // In the case of the PollingBlockStream, the stream_ptr can be ahead of the
+                    // subgraph's deployment head (database), because this component is pipelined.
+                    Some(Ok(BlockStreamEvent::Revert(stream_ptr, revert_to_ptr, cursor))) => {
+                        // Current deployment head in the database / WritableAgent Mutex cache.
+                        //
+                        // Safe unwrap because in a Revert event we're sure the subgraph has
+                        // advanced at least once.
+                        let subgraph_ptr = self.inputs.store.block_ptr().unwrap();
```
otaviopace, line 195: Be aware that all of the `subgraph_ptr` occurrences in this `match` revert branch now are using the **database block**, and **not** the **BlockStream one**.

Take a look below at the `metrics` and `revert_data_sources` code.

core/src/subgraph/runner.rs
```diff
@@ -185,8 +185,20 @@ where
 
                 let (block, cursor) = match event {
                     Some(Ok(BlockStreamEvent::ProcessBlock(block, cursor))) => (block, cursor),
-                    Some(Ok(BlockStreamEvent::Revert(subgraph_ptr, revert_to_ptr, cursor))) => {
-                        info!(&logger, "Reverting block to get back to main chain"; "current" => &subgraph_ptr, "revert_to" => &revert_to_ptr);
+                    // In the case of the PollingBlockStream, the stream_ptr can be ahead of the
+                    // subgraph's deployment head (database), because this component is pipelined.
+                    Some(Ok(BlockStreamEvent::Revert(stream_ptr, revert_to_ptr, cursor))) => {
```
maoueh, line 190: Should we remove the `stream_ptr` entirely then, it's not used anymore expect for logging purposes, so maybe it's not even worth it have it returned anymore?

otaviopace: I was wondering the same, but I didn't remove because I wasn't sure about the firehose part.

core/src/subgraph/runner.rs
leoyvens: It should be the same value afaict

leoyvens: Maybe in the skip blocks case the block stream could be ahead

core/src/subgraph/runner.rs
maoueh: Firehose could be pipelined itself (when you say `pipeline` you are referring to wrapping through the`BufferedBlockStream` right), so it shouldn't matter here if we are through Firehose or not.

The only difference is that FirehoseBlockStream can revert more than 1 block, just as an info.

And as a side note, and I think that when we have "filtering/trimming" in place to have Firehose, `BufferedBlockStream` default value for Firehose should be at least 2, so we do the "decoding" work in parallel to always have block ready in the pipeline (right now each block are "full", so it's probably a bad idea to have each subgraph maybe having 2 blocks ready, will see in the integration testing in the upcoming weeks :D)

leoyvens: I agree we can remove the first parameter of `BlockStreamEvent::Revert`.

graph/src/blockchain/block_stream.rs
```diff
@@ -209,7 +209,7 @@ pub enum BlockStreamEvent<C: Blockchain> {
     // The payload is the current subgraph head pointer, which should be reverted, such that the
     // parent of the current subgraph head becomes the new subgraph head.
     // An optional pointer to the parent block will save a round trip operation when reverting.
```
leoyvens, line 211: The comment needs to be updated.

graph/src/blockchain/polling_block_stream.rs
```diff
@@ -61,10 +61,10 @@ enum ReconciliationStep<C>
 where
     C: Blockchain,
 {
-    /// Revert(from, to) the current block pointed at by the subgraph pointer. The pointer is to the current
+    /// Revert(to) the current block pointed at by the subgraph pointer. The pointer is to the current
     /// subgraph head, and a single block will be reverted so the new head will be the parent of the
     /// current one. The second BlockPtr is the parent.
```
leoyvens, line 66: This one as well.

core/src/subgraph/runner.rs
otaviopace: I just removed the field, thanks for pointing this out @maoueh ❤️

graph/src/blockchain/polling_block_stream.rs
otaviopace: all done ✅

graph/src/blockchain/block_stream.rs
otaviopace: all done ✅

leoyvens (approved):

PR #3326, opened 2022-03-04
ethereum: log warning and continue on parse input failures

### LINKED ISSUE #3314: Raw input decode failure when processing trigger
process_triggers function in `graph-node/core/src/subgraph/instance.rs` is failing on one specific raw input on block #14277019. It worked on similar input size on a previous block. etherscan wasn't able to decode the data either. It is currently breaking some subgraphs. 

failing transaction: https://etherscan.io/tx/0x5f0b0ea26e74c9d796209920bbae08f5ebb4ae053ef8b01c7848bd0a2b85691d

working transaction:
https://etherscan.io/tx/0xe3a320d46603d8b1749e1e95bd309791bd50ee7e17531254e801ba0fec4518bb

Before the mapping is even run, the graph-node fails with this error:

`
ERRO Subgraph instance failed to run: failed to process trigger: block #14277019 (0x2bae…9554), transaction 5f0b0ea26e74c9d796209920bbae08f5ebb4ae053ef8b01c7848bd0a2b85691d: Generating function inputs for the call Function { name: "atomicMatch_", inputs: [Param { name: "addrs", kind: FixedArray(Address, 14) }, Param { name: "uints", kind: FixedArray(Uint(256), 18) }, Param { name: "feeMethodsSidesKindsHowToCalls", kind: FixedArray(Uint(8), 8) }, Param { name: "calldataBuy", kind: Bytes }, Param { name: "calldataSell", kind: Bytes }, Param { name: "replacementPatternBuy", kind: Bytes }, Param { name: "replacementPatternSell", kind: Bytes }, Param { name: "staticExtradataBuy", kind: Bytes }, Param { name: "staticExtradataSell", kind: Bytes }, Param { name: "vs", kind: FixedArray(Uint(8), 2) }, Param { name: "rssMetadata", kind: FixedArray(FixedBytes(32), 5) }], outputs: [], constant: false, state_mutability: Payable } failed, raw input: ab834bab0000000000000000000000...: Invalid data
`

There was a similar bug #3194 that was fixed but this issue still exists. 
raw_input:
```ab834bab0000000000000000000000007f268357a8c2552623316e2562d90e642bb538e500000000000000000000000043ab2a731be043163f15a361aac0b3ff3ffede4c000000000000000000000000de8194fc7b6642b70b7e7c8d6e7a18de2a3abbde0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000baf2127b49fc93cbca6269fade0f7f31df4c88a7000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000007f268357a8c2552623316e2562d90e642bb538e5000000000000000000000000de8194fc7b6642b70b7e7c8d6e7a18de2a3abbde00000000000000000000000000000000000000000000000000000000000000000000000000000000000000005b3256965e7c3cf26e11fcaf296dfc8807c01073000000000000000000000000baf2127b49fc93cbca6269fade0f7f31df4c88a7000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000370000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000016345785d8a0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000621926d20000000000000000000000000000000000000000000000000000000000000000038a763ed2d595140fcacca5bb198f0db170c6dc72c0a31c81d7d8249587da320000000000000000000000000000000000000000000000000000000000000370000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000016345785d8a000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000062168cd000000000000000000000000000000000000000000000000000000000621fc7a45478430ae159363193537ab0f9b1abf89e50bc86cd94c1056bf1e648de4c6de10000000000000000000000000000000000000000000000000000000000000001000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000006a000000000000000000000000000000000000000000000000000000000000007c000000000000000000000000000000000000000000000000000000000000008e00000000000000000000000000000000000000000000000000000000000000a000000000000000000000000000000000000000000000000000000000000000b200000000000000000000000000000000000000000000000000000000000000b60000000000000000000000000000000000000000000000000000000000000001b000000000000000000000000000000000000000000000000000000000000001b2e8de92a86092f4185e1997772ecae8adde1ea2e79a6d8d3218725fb9c7bdf1f2e921cbc9cc62f60de7f037372f315d613ac91e261f50311c8223d18735b97bd2e8de92a86092f4185e1997772ecae8adde1ea2e79a6d8d3218725fb9c7bdf1f2e921cbc9cc62f60de7f037372f315d613ac91e261f50311c8223d18735b97bd000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000e4fb16a595000000000000000000000000000000000000000000000000000000000000000000000000000000000000000043ab2a731be043163f15a361aac0b3ff3ffede4c00000000000000000000000057d819152a8bc86d097fed1dcaa6259b5380a4b10000000000000000000000000000000000000000000000000000000000000084000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000c000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000e4fb16a595000000000000000000000000de8194fc7b6642b70b7e7c8d6e7a18de2a3abbde000000000000000000000000000000000000000000000000000000000000000000000000000000000000000057d819152a8bc86d097fed1dcaa6259b5380a4b10000000000000000000000000000000000000000000000000000000000000084000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000c000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000e400000000ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000e4000000000000000000000000000000000000000000000000000000000000000000000000ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000```

### PR DESCRIPTION
**Pull request by mmchougule**
Subgraph gets terminated when the graph-node fails to decode the call data with function_abi here: `function_abi.decode_input(&call.input.0[4..])`

failing transaction: https://etherscan.io/tx/0x5f0b0ea26e74c9d796209920bbae08f5ebb4ae053ef8b01c7848bd0a2b85691d

Failing on the graph studio and etherscan also fails to decode this data. This PR skips the transaction that we aren't able to parse using the abi, logs the warning, and continues instead of completely failing the subgraph.

Without this, to fix and continue, we need to do something similar to the following which is not efficient time + replicating data: 
1. new subgraph needs to be deployed with a shortened `function_abi` but this isn't an issue with subgraphs itself
2. start indexing all over for the new subgraph which would take a while
3. use `grafting`  which copies the entire data from one schema to the next which takes a few hours for 10+million records

Fixes #3314

### DIFF
--- a/chain/ethereum/src/data_source.rs
+++ b/chain/ethereum/src/data_source.rs
@@ -614,13 +614,9 @@ impl DataSource {
                       Ok(val) => val,
                       Err(err) => {
                         warn!(logger, "Failed parsing inputs, skipping"; "error" => &err.to_string());
-                        Vec::new()
+                        return Ok(None)
                       },
                     };
-                
-                if tokens.len() == 0 {
-                  return Ok(None)
-                }
 
                 ensure!(
                     tokens.len() == function_abi.inputs.len(),
--- a/chain/ethereum/src/data_source.rs
+++ b/chain/ethereum/src/data_source.rs
@@ -18,7 +18,7 @@ use graph::{
     prelude::{
         async_trait,
         ethabi::{Address, Contract, Event, Function, LogParam, ParamType, RawLog},
-        info, serde_json,
+        info, warn, serde_json,
         web3::types::{Log, Transaction, H256},
         BlockNumber, CheapClone, DataSourceTemplateInfo, Deserialize, EthereumCall,
         LightEthereumBlock, LightEthereumBlockExt, LinkResolver, Logger, TryStreamExt,
@@ -602,7 +602,7 @@ impl DataSource {
                 // Take the input for the call, chop off the first 4 bytes, then call
                 // `function.decode_input` to get a vector of `Token`s. Match the `Token`s
                 // with the `Param`s in `function.inputs` to create a `Vec<LogParam>`.
-                let tokens = function_abi
+                let tokens = match function_abi
                     .decode_input(&call.input.0[4..])
                     .with_context(|| {
                         format!(
@@ -610,7 +610,17 @@ impl DataSource {
                             &function_abi,
                             hex::encode(&call.input.0)
                         )
-                    })?;
+                    }) {
+                      Ok(val) => val,
+                      Err(err) => {
+                        warn!(logger, "Failed parsing inputs, skipping"; "error" => &err.to_string());
+                        Vec::new()
+                      },
+                    };
+                
+                if tokens.len() == 0 {
+                  return Ok(None)
+                }
 
                 ensure!(
                     tokens.len() == function_abi.inputs.len(),
--- a/chain/ethereum/src/data_source.rs
+++ b/chain/ethereum/src/data_source.rs
@@ -18,7 +18,7 @@ use graph::{
     prelude::{
         async_trait,
         ethabi::{Address, Contract, Event, Function, LogParam, ParamType, RawLog},
-        info, warn, serde_json,
+        info, serde_json, warn,
         web3::types::{Log, Transaction, H256},
         BlockNumber, CheapClone, DataSourceTemplateInfo, Deserialize, EthereumCall,
         LightEthereumBlock, LightEthereumBlockExt, LinkResolver, Logger, TryStreamExt,
@@ -602,21 +602,21 @@ impl DataSource {
                 // Take the input for the call, chop off the first 4 bytes, then call
                 // `function.decode_input` to get a vector of `Token`s. Match the `Token`s
                 // with the `Param`s in `function.inputs` to create a `Vec<LogParam>`.
-                let tokens = match function_abi
-                    .decode_input(&call.input.0[4..])
-                    .with_context(|| {
+                let tokens = match function_abi.decode_input(&call.input.0[4..]).with_context(
+                    || {
                         format!(
                             "Generating function inputs for the call {:?} failed, raw input: {}",
                             &function_abi,
                             hex::encode(&call.input.0)
                         )
-                    }) {
-                      Ok(val) => val,
-                      Err(err) => {
+                    },
+                ) {
+                    Ok(val) => val,
+                    Err(err) => {
                         warn!(logger, "Failed parsing inputs, skipping"; "error" => &err.to_string());
-                        return Ok(None)
-                      },
-                    };
+                        return Ok(None);
+                    }
+                };
 
                 ensure!(
                     tokens.len() == function_abi.inputs.len(),

### CODE REVIEW

File: chain/ethereum/src/data_source.rs [hunk 1]
```diff
@@ -602,15 +602,25 @@ impl DataSource {
...
-                    })?;
+                    }) {
+                      Ok(val) => val,
+                      Err(err) => {
+                        warn!(logger, "Failed parsing inputs, skipping"; "error" => &err.to_string());
+                        Vec::new()
+                      },
+                    };
+                
+                if tokens.len() == 0 {
```
otaviopace, line 621: I'm just not sure if we should do this validation here or in the `adapter`, like I've done [here](https://github.com/graphprotocol/graph-node/pull/3219/files#diff-2bd955f61f661e1160bebb53dd34de5cfc4200f5900c169cbbe7a0f3471c932eR333).

What do you think @leoyvens ?

mmchougule: Thanks, @otaviopace for the review. The only reason I had to add the validation here is that the next line `ensure(tokens.len() == function_abi.inputs.len()` fails due to the length of tokens being 0. So instead of returning a None to continue, it was failing completely.

leoyvens: It looks like this would `None` if the function has no arguments which is incorrect.

chain/ethereum/src/data_source.rs
```diff
@@ -602,15 +602,25 @@ impl DataSource {
...
                             "Generating function inputs for the call {:?} failed, raw input: {}",
                             &function_abi,
                             hex::encode(&call.input.0)
                         )
-                    })?;
+                    }) {
+                      Ok(val) => val,
+                      Err(err) => {
+                        warn!(logger, "Failed parsing inputs, skipping"; "error" => &err.to_string());
+                        Vec::new()
```
leoyvens, line 617: This should be `return Ok(None)` instead of `Vec::new()`.

chain/ethereum/src/data_source.rs
mmchougule: Thanks. Removed it. :)

otaviopace (approved):

otaviopace:
@mmchougule the formatting is incorrect, you can run `cargo fmt` locally to fix it 🙂

mmchougule:
Thank you, @otaviopace. Just ran it and pushed.

otaviopace:
I'm rerunning CI, it should pass now 🙏

mmchougule:
Hey @evaporei, do you know when this will get merged and deployed to the graph studio? We want to resume indexing our subgraph for development and publish it to the decentralized network. :)

leoyvens:
This looks good to merge, thanks for the contribution @mmchougule

PR #3336, opened 2022-03-07
Asyncify various methods in `WritableStore` abd `SubgraphStore`

maoueh:
Also asyncify a bunch of dependencies transitively that were using those.

#### `WritableStore`

- `block_ptr`
- `block_cursor`
- `delete_block_cursor`
- `start_subgraph_deployment`

#### `QueryStore`

- `block_ptr`

#### `SubgraphStore`

- `writable_for_network_indexer`
- `least_block_ptr`

### DIFF
--- a/core/src/subgraph/instance_manager.rs
+++ b/core/src/subgraph/instance_manager.rs
@@ -152,11 +152,15 @@ where
             let store = store.clone();
             let logger = logger.clone();
 
-            store
-                .start_subgraph_deployment(&logger)
-                .map_err(Error::from)
-                .await
-                .map_err(Error::from)?
+            // `start_subgraph_deployment` is blocking.
+            task::spawn_blocking(move || {
+                store
+                    .start_subgraph_deployment(&logger)
+                    .map_err(Error::from)
+            })
+            .await
+            .map_err(Error::from)
+            .and_then(|x| x)?;
         }
 
         let manifest: SubgraphManifest<C> = {
--- a/core/src/subgraph/instance_manager.rs
+++ b/core/src/subgraph/instance_manager.rs
@@ -152,15 +152,11 @@ where
             let store = store.clone();
             let logger = logger.clone();
 
-            // `start_subgraph_deployment` is blocking.
-            task::spawn_blocking(move || {
-                store
-                    .start_subgraph_deployment(&logger)
-                    .map_err(Error::from)
-            })
-            .await
-            .map_err(Error::from)
-            .and_then(|x| x)?;
+            store
+                .start_subgraph_deployment(&logger)
+                .map_err(Error::from)
+                .await
+                .map_err(Error::from)?
         }
 
         let manifest: SubgraphManifest<C> = {
@@ -259,7 +255,7 @@ where
 
         // Initialize deployment_head with current deployment head. Any sort of trouble in
         // getting the deployment head ptr leads to initializing with 0
-        let deployment_head = store.block_ptr().map(|ptr| ptr.number).unwrap_or(0) as f64;
+        let deployment_head = store.block_ptr().await.map(|ptr| ptr.number).unwrap_or(0) as f64;
         block_stream_metrics.deployment_head.set(deployment_head);
 
         let host_builder = graph_runtime_wasm::RuntimeHostBuilder::new(
--- a/core/src/subgraph/instance_manager.rs
+++ b/core/src/subgraph/instance_manager.rs
@@ -148,20 +148,7 @@ where
         // sources; if the subgraph is a graft or a copy, starting it will
         // do the copying and dynamic data sources won't show up until after
         // that is done
-        {
-            let store = store.clone();
-            let logger = logger.clone();
-
-            // `start_subgraph_deployment` is blocking.
-            task::spawn_blocking(move || {
-                store
-                    .start_subgraph_deployment(&logger)
-                    .map_err(Error::from)
-            })
-            .await
-            .map_err(Error::from)
-            .and_then(|x| x)?;
-        }
+        store.start_subgraph_deployment(&logger).await?;
 
         let manifest: SubgraphManifest<C> = {
             info!(logger, "Resolve subgraph files using IPFS");
--- a/graph/src/components/store/traits.rs
+++ b/graph/src/components/store/traits.rs
@@ -156,7 +156,7 @@ pub trait WritableStore: Send + Sync + 'static {
     async fn delete_block_cursor(&self) -> Result<(), StoreError>;
 
     /// Start an existing subgraph deployment.
-    async fn start_subgraph_deployment(&self, logger: &Logger) -> Result<(), StoreError>;
+    fn start_subgraph_deployment(&self, logger: &Logger) -> Result<(), StoreError>;
 
     /// Revert the entity changes from a single block atomically in the store, and update the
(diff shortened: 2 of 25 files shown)


leoyvens (changes requested):

store/postgres/src/writable.rs
```diff
@@ -165,6 +170,7 @@ impl WritableStore {
             store.start_subgraph(logger, self.site.clone(), graft_base)?;
             self.store.primary_conn()?.copy_finished(self.site.as_ref())
         })
+        .await
```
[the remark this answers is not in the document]
maoueh: I did it because it was required at some point to have it `async`. So I would beed to bridge with some `async` and `non-async` at some point. What I mean is that if I revert `start_subgraph_deployment` asyncification, I would need to bridge `async` and `non-async` somewhere. If I keep it, I would need to do this 
`async`/`non-async` in another place.

What's the best way to keep `start_subgraph_deployment` async but fix the problem you mention here, without asyncify all `graft_pending` `start_subgraph` and `primary_conn`? Maybe the answer is to actually asyncify all those.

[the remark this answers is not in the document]
leoyvens: The solution we had so far is to wrap it in `task::spawn_blocking` in the instance manager, we can continue to do that.

core/src/subgraph/instance_manager.rs
```diff
@@ -143,15 +143,11 @@ where
...
 
-            // `start_subgraph_deployment` is blocking.
-            task::spawn_blocking(move || {
-                store
-                    .start_subgraph_deployment(&logger)
-                    .map_err(Error::from)
-            })
-            .await
-            .map_err(Error::from)
-            .and_then(|x| x)?;
```
leoyvens, line 154: Here.

core/src/subgraph/instance_manager.rs
```diff
@@ -143,15 +143,11 @@ where
             let store = store.clone();
             let logger = logger.clone();
 
-            // `start_subgraph_deployment` is blocking.
-            task::spawn_blocking(move || {
```
leoyvens, line 147: Here.

graph/src/components/store/traits.rs
```diff
@@ -156,17 +156,17 @@ pub trait SubgraphStore: Send + Sync + 'static {
...
-    fn block_cursor(&self) -> Option<String>;
+    async fn block_cursor(&self) -> Option<String>;
 
     /// Deletes the current Firehose `cursor` this deployment is currently at.
-    fn delete_block_cursor(&self) -> Result<(), StoreError>;
+    async fn delete_block_cursor(&self) -> Result<(), StoreError>;
 
     /// Start an existing subgraph deployment.
-    fn start_subgraph_deployment(&self, logger: &Logger) -> Result<(), StoreError>;
+    async fn start_subgraph_deployment(&self, logger: &Logger) -> Result<(), StoreError>;
```
mangas, line 169: I generally prefer to handle this type of concurrency on the caller, this is not really about the async

maoueh: Please expand because I do not get what you mean here.

mangas: What I'm suggesting is that we don't make this particular one async, as in, let this call be blocking and let the caller decide whether do send it to a task or block or it. @leoyvens any preference about this one?

store/postgres/src/writable.rs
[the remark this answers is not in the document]
maoueh: I've reverted the deployment change and did use `futures03::executor::block_on` when I needed to bridge the inverse way.

Please review accordingly.

graph/src/components/store/traits.rs
leoyvens: On the general pattern we want to follow, the most important thing is to avoid doing blocking operations within a non-blocking task, which is an easy mistake to make when the store exposes blocking fns. So it's safest to expose async functions when in doubt.

But in this case I agree it's deep rabbit hole for a function that has few call sites, so this is fine.

store/postgres/src/writable.rs
```diff
@@ -384,12 +383,16 @@ impl WritableStoreTrait for WritableAgent {
...
 
-    async fn start_subgraph_deployment(&self, logger: &Logger) -> Result<(), StoreError> {
-        self.store.start_subgraph_deployment(logger).await?;
+    fn start_subgraph_deployment(&self, logger: &Logger) -> Result<(), StoreError> {
+        self.store.start_subgraph_deployment(logger)?;
 
         // Refresh all in memory state in case this instance was used before
-        *self.block_ptr.lock().unwrap() = self.store.block_ptr().await?;
-        *self.block_cursor.lock().unwrap() = self.store.block_cursor().await?;
+        futures03::executor::block_on(async {
```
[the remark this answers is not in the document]
maoueh: I'll change, copied over from other places.

graph/src/components/store/traits.rs
maoueh: I achieved a good chunk and I reached the point where I need to do something so that  `self.store.primary_conn()` would not block the whole thread. I wasn't sure so decided to stop there. Would you have any insights, think some ground work would be need for the `primay_conn` stuff, is it possible?

I'm willing to maybe open a new PR early next week to convert even more stuff up.

graphql/src/store/resolver.rs
```diff
@@ -76,12 +76,9 @@ impl StoreResolver {
...
         let deployment2 = deployment.clone();
-        let block_ptr = graph::spawn_blocking_allow_panic(move || {
-            Self::locate_block(store_clone.as_ref(), bc, deployment2.clone())
-        })
-        .await
-        .map_err(|e| QueryExecutionError::Panic(e.to_string()))
-        .and_then(|x| x)?; // Propagate panics.
+        let block_ptr = Self::locate_block(store_clone.as_ref(), bc, deployment2.clone())
+            .await
+            .map_err(|e| QueryExecutionError::Panic(e.to_string()))?;
```
leoyvens, line 81:
```suggestion
            ?;
```
No need for error gymnastics now that the spawn_blocking is gone.

leoyvens (changes requested):

store/postgres/src/writable.rs
```diff
@@ -383,16 +383,16 @@ impl WritableStoreTrait for WritableAgent {
...
     }
 
-    fn start_subgraph_deployment(&self, logger: &Logger) -> Result<(), StoreError> {
-        self.store.start_subgraph_deployment(logger)?;
+    async fn start_subgraph_deployment(&self, logger: &Logger) -> Result<(), StoreError> {
+        let store = self.store.cheap_clone();
+        let logger = logger.cheap_clone();
+        graph::spawn_blocking_allow_panic(move || store.start_subgraph_deployment(&logger))
+            .await
+            .map_err(Error::from)??;
```
leoyvens, line 391: I had never seen `??` as a way to flatten a result, fun.

leoyvens (approved):
`start_subgraph_deployment` is looking good and asyncifying the tests is exactly what you should do, it's all looking refined, thanks for working through these `async` nuances.

store/postgres/src/writable.rs
maoueh: Seems like it does the `Result<>` unwrap with return syntax desugaring twice :)

PR #3338, opened 2022-03-08
ethereum: Update web3 and ethabi

leoyvens:
This rebases our web3 fork, which reduces the [required patches](https://github.com/graphprotocol/rust-web3/commits/graph-patches-onto-0.18). They have been submitted upstream https://github.com/tomusdrw/rust-web3/pull/616 so we might be able to drop the fork.

This also does a pass at some other deps such as `rand`, `bytes` and `uuid`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -105,17 +105,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "async-native-tls"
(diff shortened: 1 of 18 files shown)


mangas:
This needs a rebase

mangas (approved):

PR #3339, opened 2022-03-08
feat: allow configuring genesis block number from ethereum options

### LINKED ISSUE #2919: [Feature Request] Substrate Frontier runtime upgrade Compatibility

Feature

In a polkadot parachain where it uses Frontier for their WASM runtime, evm feature sometimes does not start from day 1(i.e. block 0). It might add EVM for testing its connection between relay chain or go through necessary tests before they add evm integration. Due to this, current graph node is not able to index evm since current code is set to index from block 0 [here](https://github.com/graphprotocol/graph-node/blob/60262c1eb2900c712ad517ef6c022d157c14c90f/chain/ethereum/src/ethereum_adapter.rs#L875). I would like to suggest to get input from config to set genesis block of evm to index.

The error shows in shiden with `"Ethereum node could not find genesis block"`  with the following log
```
Oct 13 16:30:16.455 DEBG Trying again after eth_getBlockByNumber(0, false) RPC call failed (attempt #1) with result Err(Ethereum node could not find genesis block), provider: shiden-rpc-0
Oct 13 16:30:16.811 DEBG Trying again after eth_getBlockByNumber(0, false) RPC call failed (attempt #2) with result Err(Ethereum node could not find genesis block), provider: shiden-rpc-0
Oct 13 16:30:17.177 DEBG Trying again after eth_getBlockByNumber(0, false) RPC call failed (attempt #3) with result Err(Ethereum node could not find genesis block), provider: shiden-rpc-0
```

I expect adding `gen_block` input in config and add that in `ChainSection` or `stores` so that genesis block argument can be used in `ethereum_adapter.rs`

ChrisLahaye:
resolves #2919

Signed-off-by: Chris Lahaye <mail@chrislahaye.com>


ChrisLahaye:
This commit is suppose to add better support for Substrate Frontier, an Ethereum compatibility layer for Substrate, in particular for chains where compatibility was add after block zero. Using such chains with graph-node does not work as the genesis block cannot be queried because they are hard-coded to be block zero. This commit allows configuring the genesis block number at start-up using the existing ethereum options.

**Old syntax**
NETWORK_NAME:[CAPABILITIES]:URL

**New syntax**
NETWORK_NAME:[CAPABILITIES]:URL[,GENESIS_BLOCK_NUMBER]

I understand that this is probably a long shot as the genesis block being block zero is a pretty fundamental thing, and therefore, probably hard-coded and implied widely across the project.

https://github.com/graphprotocol/graph-node/blob/6f4adda0220cb7a9f24a801db066f92a00a0a81b/core/src/subgraph/registrar.rs#L455 https://github.com/graphprotocol/graph-node/blob/6f4adda0220cb7a9f24a801db066f92a00a0a81b/graph/src/components/ethereum/types.rs#L43 https://github.com/graphprotocol/graph-node/blob/6f4adda0220cb7a9f24a801db066f92a00a0a81b/chain/ethereum/src/data_source.rs#L761

I must admit that my knowledge about this project and rust is pretty limited. I would like to collaborate and work on this feature, although I would require some direction. I will just leave this here and hope to receive some replies on how to approach this.

azf20:
Thanks @ChrisLahaye! @leoyvens interested in your take on this approach - I wonder if this shouldn't be chain agnostic?

leoyvens:
This is not relevant with firehose, since it already doesn't assume that the genesis block is 0.

My only issue is with where to put this config, I don't think we want to expand on the CLI config. This could fit into the [config file format](https://github.com/graphprotocol/graph-node/blob/master/docs/config.md#configuring-ethereum-providers), under the provider.

maoueh:
@azf20 On other chain that uses `Firehose`, `0` always means give me genesis block from Firehose perspective, so tweaking it in `graph-node` is not required. Although something of interest where knowing the genesis block would be interesting would be to validate that the manifest's start block is always >= genesis block. But this could be provided by Firehose endpoint directly, Dan from Figment proposed adding such endpoint to the Firehose service (to know the genesis block, latest block known block) and this could be used instead of specifying it at the command line per network.

@ChrisLahaye There will be some other places where genesis block is implicitly required. For example when starting a subgraph, we do https://github.com/graphprotocol/graph-node/blob/master/core%2Fsrc%2Fsubgraph%2Fregistrar.rs#L455-L465.

This would need to be adapter also, because if the `start_block` is specified on the custom block number, it means we would not find the `block_number - 1`.

There might be other places where it will need to be handled, maybe when validating

ChrisLahaye:
@maoueh Correct, I referred to the places that I could find in my previous comment. Most of them don't have access to the chain struct that I used to store the genesis block number. I first want to gather some feedback before passing down various structs and updating function parameters. First we should decide whether this is worth pursuing, if so, then I will start updating these places one by one.

PR #3340, opened 2022-03-08
Handle arbitrarily large `Bytes` attributes

### LINKED ISSUE #2287: Bytes field of over a certain size won't be indexable

lutter:
Treat `Bytes` attributes similar to how we treat `String` attributes by only indexing a prefix of them and adjusting queries to take advantage of that index as much as possible.

The `id` field is treated special (just as it was for `String`) in that we do not make provisions for arbitrarily large values in it and instead require that those values are small enough to fit into an index; the limit for the size of `id` values is therefore around 2kB, which should not be an issue in practice.

Closes #2287 
Supersedes #2747


tilacog (approved):

lutter:
This was really merged, but is marked closed because of some git hiccup

PR #3343, opened 2022-03-09
GraphQL route to expose call cache contents by block

### LINKED ISSUE #3312: GraphQL API for Ethereum call cache
When we have detected a difference in the indexing results of two indexers, we need to gather more details and contextual information to debug what could have led to that difference.

Differences are always starting at a specific block and one of the details that we need is the Ethereum call cache that graph-node used for processing the subgraph in question.

This is (roughly) the schema that we need:

```graphql
type Query {
  cachedEthereumCalls(network: String!, blockNumber: Int!): [CachedEthereumCallData!]!
}

type CachedEthereumCallData {
  ...
}
```

neysofu:
Closes #3312.

Initially, I tried to retrace all contract calls in the given block using the logic inside `EthereumAdapter`, then query the cache to see which of those calls were present in the cache, if any. This operates under the assumption that the data coming from the `EthereumAdapter` at the time of cache insertion and cache lookup are identical, which is not necessarily true, as we wouldn't need this debugging API in the first place if that was the case.

After some back and forth, I realized there was no way I could guarantee total consistency of the results obtained this way with the actual contents of the cache and I landed on the original, simpler design - an additional index to the `call_cache` tables.

### DIFF
--- a/graph/src/components/store/traits.rs
+++ b/graph/src/components/store/traits.rs
@@ -358,7 +358,8 @@ pub trait ChainStore: Send + Sync + 'static {
 }
 
 pub trait EthereumCallCache: Send + Sync + 'static {
-    /// Cached return value.
+    /// Returns the return value of the provided Ethereum call, if present in
+    /// the cache.
     fn get_call(
         &self,
         contract_address: ethabi::Address,
@@ -366,7 +367,11 @@ pub trait EthereumCallCache: Send + Sync + 'static {
         block: BlockPtr,
     ) -> Result<Option<Vec<u8>>, Error>;
 
-    // Add entry to the cache.
+    /// Returns all cached calls for a given `block`. This method does *not*
+    /// update the last access time of the returned cached calls.
+    fn get_calls_in_block(&self, block: BlockPtr) -> Result<Vec<CachedEthereumCall>, Error>;
+
+    /// Stores the provided Ethereum call in the cache.
     fn set_call(
         &self,
         contract_address: ethabi::Address,
--- a/server/index-node/src/resolver.rs
+++ b/server/index-node/src/resolver.rs
@@ -159,6 +159,91 @@ where
         })
     }
 
+    fn resolve_cached_ethereum_calls(
+        &self,
+        field: &a::Field,
+    ) -> Result<r::Value, QueryExecutionError> {
+        let network = field
+            .get_required::<String>("network")
+            .expect("Valid network required");
+
+        let block_hash = field
+            .get_required::<H256>("blockHash")
+            .expect("Valid blockHash required");
+
+        let chain = if let Ok(c) = self
+            .blockchain_map
+            .get::<graph_chain_ethereum::Chain>(network.clone())
+        {
+            c
+        } else {
+            error!(
+                self.logger,
+                "Failed to fetch cached Ethereum calls; nonexistent network";
+                "network" => network,
+                "block_hash" => format!("{}", block_hash),
+            );
+            return Ok(r::Value::Null);
+        };
+        let chain_store = chain.chain_store();
+        let call_cache = chain.call_cache();
+
+        let block_number = match chain_store.block_number(block_hash) {
+            Ok(Some((_, n))) => n,
+            Ok(None) => {
+                error!(
+                    self.logger,
+                    "Failed to fetch cached Ethereum calls; block not found";
+                    "network" => network,
+                    "block_hash" => format!("{}", block_hash),
+                );
+                return Ok(r::Value::Null);
+            }
+            Err(e) => {
+                error!(
+                    self.logger,
+                    "Failed to fetch cached Ethereum calls; storage error";
+                    "network" => network.as_str(),
+                    "block_hash" => format!("{}", block_hash),
+                    "error" => e.to_string(),
+                );
+                return Ok(r::Value::Null);
+            }
+        };
+        let block_ptr = BlockPtr::new(block_hash.into(), block_number);
+
+        let calls = match call_cache.get_calls_in_block(block_ptr) {
+            Ok(c) => c,
+            Err(e) => {
+                error!(
+                    self.logger,
+                    "Failed to fetch cached Ethereum calls; storage error";
+                    "network" => network.as_str(),
+                    "block_hash" => format!("{}", block_hash),
+                    "error" => e.to_string(),
+                );
+                return Err(QueryExecutionError::StoreError(Error::from(e).into()));
+            }
+        };
+
+        Ok(r::Value::List(
+            calls
+                .into_iter()
+                .map(|cached_call| {
+                    object! {
+                        idHash: &cached_call.blake3_id[..],
+                        block: object! {
+                            hash: cached_call.block_ptr.hash.hash_hex(),
+                            number: cached_call.block_ptr.number,
+                        },
+                        contractAddress: &cached_call.contract_address[..],
+                        returnValue: &cached_call.return_value[..],
+                    }
+                })
+                .collect::<Vec<r::Value>>(),
+        ))
+    }
+
     fn resolve_proof_of_indexing(&self, field: &a::Field) -> Result<r::Value, QueryExecutionError> {
         let deployment_id = field
             .get_required::<DeploymentHash>("subgraph")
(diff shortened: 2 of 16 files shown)


server/index-node/src/schema.graphql
```diff
@@ -4,7 +4,16 @@ scalar Bytes
...
+
+# An obscure object type. Note that this is equivalent to a JSON object, rather
+# than a generic JSON value; as such it cannot be e.g. a string or an integer.
 scalar JSONObject
+# A string in the format of `YYYY-MM-DD`. This is the standard adopted by
+# RFC3339, which is equivalent to the "ISO 8601 extended format".
+# See also:
+# - https://datatracker.ietf.org/doc/html/rfc3339#section-5.6
+# - https://en.wikipedia.org/wiki/ISO_8601#Calendar_dates
+scalar Date
```
otaviopace, line 16: I'm unsure about adding a Scalar. We haven't done this in a while, so I don't know if it's fine and we have already discussed about this as a team.

server/index-node/src/schema.graphql
```diff
@@ -22,6 +31,10 @@ type Query {
   subgraphFeatures(subgraphId: String!): SubgraphFeatures!
   entityChangesInBlock(subgraphId: String!, blockNumber: Int!): EntityChanges!
   blockData(network: String!, blockHash: Bytes!): JSONObject
+  cachedEthereumCalls(
+    network: String!
+    blockHash: Bytes!
```
otaviopace, line 36: Sorry I didn't see you've addressed [my comment](https://github.com/graphprotocol/graph-node/issues/3312#issuecomment-1074149494) before hand 😅

graph/src/components/store/traits.rs
```diff
@@ -358,15 +358,20 @@ pub trait ChainStore: Send + Sync + 'static {
...
         encoded_call: &[u8],
         block: BlockPtr,
     ) -> Result<Option<Vec<u8>>, Error>;
 
-    // Add entry to the cache.
+    /// Returns all cached calls for a given `block`. This method does *not*
+    /// update the last access time of the returned cached calls.
+    fn get_calls_in_block(&self, block: BlockPtr) -> Result<Vec<CachedEthereumCall>, Error>;
+
+    // Stores the provided Ethereum call in the cache.
```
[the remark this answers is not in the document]
neysofu: Fixed, thanks! Marking as resolved.

otaviopace (commented):
Everything looks good to me, but I would be more comfortable if @tilacog or @lutter take a look into this because I'm not as aware about the `call_cache` tables as them.

So if possible I would like any of them to double check the `get_calls_in_block` function in the `store/postgres/src/chain_store.rs` file and the `.sql` ones too. 🙂

server/index-node/src/schema.graphql
neysofu: Yeah, that's fair. We were unsure about the same when introducing `JSONObject`: https://github.com/graphprotocol/graph-node/pull/3275#discussion_r817147165. We ultimately decided that was fine but didn't really discuss `Date` as far as I can remember! Let's wait for some feedback.

server/index-node/src/schema.graphql
```diff
@@ -4,7 +4,16 @@ scalar Bytes
 scalar ID
 scalar Int
 scalar String
+
+# An obscure object type. Note that this is equivalent to a JSON object, rather
```
leoyvens, line 8: 'opaque' might be a better term than 'obscure'.

server/index-node/src/schema.graphql
leoyvens: Since this is just the indexing status api, I don't mind experimenting with a date scalar.

leoyvens: Since this is just the indexing status api, I don't mind experimenting with a date scalar.

store/postgres/migrations/2022-03-09-07000_call_cache_block_number_idx/up.sql
```diff
@@ -0,0 +1,18 @@
+create index if not exists
+	eth_call_cache_block_number_idx
+on
+	eth_call_cache(block_number);
+
+do $$
+declare
+	tables cursor for select namespace from chains;
```
lutter, line 8: One thing I just noticed: `chains` is the global directory of all chains. It is maintained only in the primary, though gets copied to the other shards periodically, but only as a fallback in case the primary database goes down. The list of chains in a specific shard is in `ethereum_networks`, and that table should be consulted for the index creation here.

server/index-node/src/schema.graphql
neysofu: Marking as resolved. It seems we're okay with this.

store/postgres/migrations/2022-03-09-07000_call_cache_block_number_idx/up.sql
neysofu: Thanks for catching this! Fixed in https://github.com/graphprotocol/graph-node/pull/3343/commits/b440fad40c8e605dceeaf76f17f6671601724dc4.

neysofu:
Resolved conflicts with `master`.

server/index-node/src/resolver.rs
```diff
@@ -158,6 +159,92 @@ where
...
+
+        let chain = if let Ok(c) = self
+            .blockchain_map
+            .get::<graph_chain_ethereum::Chain>(network.clone())
+        {
+            c
+        } else {
+            error!(
+                self.logger,
+                "Failed to fetch cached Ethereum calls; nonexistant network";
```
lutter, line 182: typo: 'nonexistent'

server/index-node/src/resolver.rs
```diff
@@ -158,6 +159,92 @@ where
...
+                return Ok(r::Value::Null);
+            }
+            Err(e) => {
+                error!(
+                    self.logger,
+                    "Failed to fetch cached Ethereum calls; storage error";
+                    "network" => network.as_str(),
+                    "block_hash" => format!("{}", block_hash),
+                    "error" => e.to_string(),
+                );
```
lutter, line 209: It would be better to return all these `error!` logs as `QueryExecutionError` in some form; that way, the response to these requests will contain the error, which is much more convenient for troubleshooting (e.g., to figure out why your lookup on the network `mainet` doesn't return anything)

store/postgres/src/chain_store.rs
```diff
@@ -939,6 +936,63 @@ mod data {
...
+                        call_cache.id(),
+                        call_cache.return_value(),
+                        call_cache.contract_address(),
+                        call_meta.accessed_at(),
+                    ))
+                    .inner_join(
+                        call_meta.table().on(call_meta
+                            .contract_address()
+                            .eq(call_cache.contract_address())),
+                    )
```
lutter, line 978: It might be that this join will be very slow in practice, since `call_cache` doesn't have an index on `contract_address`; we'll see how it performs, since there's also a query plan where the lookup by `block_number` happens first and then `call_meta` gets joined along its pk index. If things go wrong, we might have to drop reporting `accessed_at`.

lutter (approved):

store/postgres/src/chain_store.rs
neysofu: Thanks for noticing. Let's keep an eye on this in case performance is unacceptable.

server/index-node/src/resolver.rs
neysofu: Good point, fixed in https://github.com/graphprotocol/graph-node/pull/3343/commits/31a121441011358fe7b202250554df8f060291f2 as well.

store/postgres/src/chain_store.rs
neysofu: On second thought, we don't need the last access timestamp, so we can avoid the `JOIN` at all. https://github.com/graphprotocol/graph-node/pull/3343/commits/f4a7ff9b1f67654d51452d18491ba4355e61f75a fixes this.

neysofu:
Improved query performance (thanks @lutter) and error handling; fixed typos. Rebasing and merging.

store/postgres/migrations/2022-03-09-07000_call_cache_block_number_idx/up.sql
```diff
@@ -0,0 +1,18 @@
+create index if not exists
+	eth_call_cache_block_number_idx
+on
+	eth_call_cache(block_number);
+
+do $$
+declare
+	tables cursor for select namespace from ethereum_networks;
```
lutter, line 8: One more thing I just noticed: for chains that are stored in the shared/public tables, this will also try to create the migration, and fail. The query needs to be `select namespace from ethereum_networks where namespace != 'public'`

PR #3348, opened 2022-03-10
More robust `proofOfIndexing` GraphQL route

neysofu:
Depending on the provided value for a new `accessToken` parameter, we'll either respond with a public POI signature or the actual signature for the given `indexer` address.


server/index-node/src/schema.graphql
```diff
@@ -18,6 +18,7 @@ type Query {
     blockNumber: Int!
     blockHash: Bytes!
     indexer: Bytes
+    accessToken: String
```
Jannis, line 21: I'd prefer the access token to be provided via a `Authorization: Bearer <token>` header because adding it here is a breaking change to the `proofOfIndexing` resolver. `Bearer` may not strictly be needed but IIRC clients often assume bearer tokens are used by default.

neysofu: Makes sense, I should be able to get this sorted by tomorrow.

server/index-node/src/auth.rs
```diff
@@ -0,0 +1,59 @@
...
+            (None, _) => true,
+            // Protection is active, but no access token was provided.
+            (Some(_), None) => false,
+            (Some(a), Some(b)) => {
+                // When comparing secrets to untrusted user data, we have to be
+                // careful about timing attacks. Constant-time comparison is the
+                // standard choice in these situations, but it can be quite
+                // convoluted. Instead, we'll compare the BLAKE3 hashes of the
+                // two values: this way we don't have to worry about timing
+                // attacks nor vetting a constant-time comparison crate.
```
otaviopace, line 39: Noice 👏

otaviopace (approved):
LGTM ✅ 

Not sure if @Jannis has something else to add up or if can just merge this.

Jannis (approved):
LGTM!

PR #3351, opened 2022-03-11
Fix Firehose chains not starting properly

maoueh:
It appears that requesting `start_block_num: 0, stop_block_num: 0` does not stop the stream after receiving the genesis block as I expected. Instead, the stream continue to send blocks forever.

This caused the `block_ptr_for_number` to never return properly, causing the `NET_WAIT` timeout to kick which was ultimately leading to the network not being added to the blockchain map and hence, no subgraph could run for this network.

To prevent looping infinitely, we stop as soon as a new received block's number is higher than the latest received block's number, in which case it means it's an event for a block we are not interested in.

### DIFF
--- a/graph/src/firehose/endpoints.rs
+++ b/graph/src/firehose/endpoints.rs
@@ -140,15 +140,38 @@ impl FirehoseEndpoint {
 
         debug!(logger, "Retrieving block(s) from firehose");
 
-        let mut block_in_longuest_chain: Option<BlockPtr> = None;
+        let mut latest_received_block: Option<BlockPtr> = None;
         while let Some(message) = block_stream.next().await {
             match message {
-                Ok(v) => block_in_longuest_chain = Some(decode_firehose_block::<M>(&v)?.ptr()),
+                Ok(v) => {
+                    let block = decode_firehose_block::<M>(&v)?.ptr();
+
+                    match latest_received_block {
+                        None => {
+                            latest_received_block = Some(block);
+                        }
+                        Some(ref actual_ptr) => {
+                            // We want to receive all events related to a specific block number,
+                            // however, in some circumstances, it seems Firehose would not stop sending
+                            // blocks (`start_block_num: 0 and stop_block_num: 0` on NEAR seems to trigger
+                            // this).
+                            //
+                            // To prevent looping infinitely, we stop as soon as a new received block's
+                            // number is higher than the latest received block's number, in which case it
+                            // means it's an event for a block we are not interested in.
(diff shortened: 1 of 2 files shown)


node/src/chain.rs
```diff
@@ -332,7 +332,7 @@ where
                     // on the URL)
                     Ok(Err(e)) | Err(e) => {
                         error!(logger, "Connection to provider failed. Not using this provider";
-                                       "error" =>  e.to_string());
+                                       "error" =>  format!("{:#}", e));
```
otaviopace, line 335: Was this done for debugging purposes? It's probably wise to keep it so it's easier to understand the problem in the future

graph/src/firehose/endpoints.rs
```diff
@@ -140,15 +140,38 @@ impl FirehoseEndpoint {
 
         debug!(logger, "Retrieving block(s) from firehose");
 
-        let mut block_in_longuest_chain: Option<BlockPtr> = None;
+        let mut latest_received_block: Option<BlockPtr> = None;
         while let Some(message) = block_stream.next().await {
```
otaviopace, line 144: We can improve this in the future, but I think this `while` loop might be able to be turned into a `loop` statement, since it returns a value on `break` 🙂

otaviopace (approved):
Nice catch, thanks for the fix! 🧤

graph/src/firehose/endpoints.rs
maoueh: Not familiar with this construct, do you have an example somewhere?

node/src/chain.rs
leoyvens: imo we should try to remember to always format errors like this, in case they are anyhow errors.

maoueh: Yes, but it led to nothing interesting. 

However, this is the best way to print `anyhow::Error`. If the anyhow error has multiple causes, they will all be printed when using the `:#` formatting directive while `to_string()` or `{}` only prints the outermost error, see https://docs.rs/anyhow/latest/anyhow/struct.Error.html#display-representations

graph/src/firehose/endpoints.rs
otaviopace: I mean this: https://doc.rust-lang.org/reference/expressions/loop-expr.html#break-and-loop-values

node/src/chain.rs
maoueh: Ahah Leo was faster than me :D At least I gave a good reference link!

PR #3354, opened 2022-03-11
Avoid checking chain continuity with subgraph block ptr if subgraph never processed any block

maoueh:
FirehoseBlockStream needs to peform a chain continuity check when starting from a subgraph's existing block ptr because we cannot be sure to respect chain continuity when switching from RPC syncing to Firehose syncing in Ethereum.

However, the previous logic to check this was a little bit flawed. Indeed, the assumption that a `None` subgraph current block would mean the subgraph was never synced never actually match. Indeed, at deployment, even before the subgraph has synced something, the subgraph current block is set to the parent of the manifest's start block number. This means we were always doing the check if the subgraph was freshly deployed.

We now changed the logic. If the subgraph ptr is set but is < than the manifest start block num, it means the subgraph never synced anything in which case, we shall skip the continuity check. That will temporarily solve the Firehose Filters problem when deploying a fresh new Firehose subgraph with filters. It will also speed up starting subgraph on other chains since we will completely skip the continuity check there since it's not required.

### DIFF
--- a/graph/src/blockchain/firehose_block_stream.rs
+++ b/graph/src/blockchain/firehose_block_stream.rs
@@ -109,23 +109,18 @@ fn stream_blocks<C: Blockchain, F: FirehoseMapper<C>>(
     // initial subgraph block pointer points to the parent block of the manifest's start block, which is usually
     // equivalent (but not always) to manifest's start block number - 1.
     //
-    // Hence, we only need to check the chain continuiy if the subgraph block ptr is higher or equal to the
-    // subgraph actual block pointer, indeed, only in this case (and there is no firehose cursor) it means the
-    // subgraph was started and advanced with something else than Firehose and as such, chain continuity check
-    // needs to be performed.
-    let mut check_subgraph_continuity = false;
-    if latest_cursor == "" {
-        if let Some(ref subgraph_current_block) = subgraph_current_block {
-            debug!(&logger, "Checking if subgraph current block is after manifest start block";
-                "subgraph_current_block_number" => subgraph_current_block.number,
-                "manifest_start_block_number" => manifest_start_block_num,
-            );
-
-            if subgraph_current_block.number >= manifest_start_block_num {
-                debug!(&logger, "Going to check continuity of chain on first block");
-                check_subgraph_continuity = true
-            }
-        }
+    // Hence, we only need to check the chain continuity if the subgraph current block ptr is higher or equal
+    // to the subgraph manifest's start block number. Indeed, only in this case (and when there is no firehose
+    // cursor) it means the subgraph was started and advanced with something else than Firehose and as such,
+    // chain continuity check needs to be performed.
+    let mut check_subgraph_continuity = must_check_subgraph_continuity(
+        &logger,
+        &subgraph_current_block,
+        &latest_cursor,
+        manifest_start_block_num,
+    );
+    if check_subgraph_continuity {
+        debug!(&logger, "Going to check continuity of chain on first block");
     }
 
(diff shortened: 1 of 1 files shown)


graph/src/blockchain/firehose_block_stream.rs
```diff
@@ -101,11 +101,32 @@ fn stream_blocks<C: Blockchain, F: FirehoseMapper<C>>(
...
+    // When the graph node deploys a new subgraph, it always create a subgraph ptr for this subgraph, the
+    // initial subgraph block pointer points to the parent block of the manifest's start block, which is usually
+    // equivalent (but not always) to manifest's start block number - 1.
+    //
+    // Hence, we only need to check the chain continuiy if the subgraph block ptr is higher or equal to the
+    // subgraph actual block pointer, indeed, only in this case (and there is no firehose cursor) it means the
+    // subgraph was started and advanced with something else than Firehose and as such, chain continuity check
+    // needs to be performed.
+    let mut check_subgraph_continuity = false;
+    if latest_cursor == "" {
```
mangas, line 117: I think this would be more readable as a match statement. Something along the lines of: 

```
let check_continuity = match subgraph_current_block {
   Some(curr_block) if latest_cursor.is_empty() => {
       debug!(&logger, "Checking if subgraph current block is after manifest start block";
                "subgraph_current_block_number" => subgraph_current_block.number,
                "manifest_start_block_number" => manifest_start_block_num,
            );

                debug!(&logger, "Going to check continuity of chain on first block");
                subgraph_current_block.number >= manifest_start_block_num
  } ,
  _ => false,
}
```

mangas: The first part could go into a function with some tests too, just to prevent someone from making a small mistake on a refactor in the future

maoueh: I totally agree.

graph/src/blockchain/firehose_block_stream.rs
```diff
@@ -101,11 +101,32 @@ fn stream_blocks<C: Blockchain, F: FirehoseMapper<C>>(
...
-    let mut check_subgraph_continuity = latest_cursor == "" && subgraph_current_block.is_some();
+    // If a Firehose cursor is present, it's used to resume the stream and as such, there is no need to
+    // perform the chain continuity check.
+    //
+    // If there was no cursor, now we need to check if the subgraph current block is set to something.
+    // When the graph node deploys a new subgraph, it always create a subgraph ptr for this subgraph, the
+    // initial subgraph block pointer points to the parent block of the manifest's start block, which is usually
+    // equivalent (but not always) to manifest's start block number - 1.
+    //
+    // Hence, we only need to check the chain continuiy if the subgraph block ptr is higher or equal to the
```
leoyvens, line 112:
```suggestion
    // Hence, we only need to check the chain continuity if the subgraph block ptr is higher or equal to the
```

graph/src/blockchain/firehose_block_stream.rs
```diff
@@ -101,11 +101,32 @@ fn stream_blocks<C: Blockchain, F: FirehoseMapper<C>>(
...
+    // subgraph actual block pointer, indeed, only in this case (and there is no firehose cursor) it means the
```
leoyvens, line 113: > subgraph actual block pointer

Is this referring to the start block number?

maoueh: Yeah it's a mistake in the comment you are right, it should be `manifest's start block number` here

PR #3359, opened 2022-03-14
Fix 'graphman copy create'

lutter:
Copying a subgraph that had a start block set would fail after the fix in PR #3287; ultimately, the bug was caused because `SubgraphDeploymentEntity` allows setting things that should not be set when creating a subgraph. Improve the code by using a different struct for creating a deployment.

### DIFF
--- a/store/postgres/src/deployment.rs
+++ b/store/postgres/src/deployment.rs
@@ -13,12 +13,15 @@ use diesel::{
     sql_query,
     sql_types::{Nullable, Text},
 };
-use graph::data::subgraph::{schema::SubgraphManifestEntity, SubgraphFeature};
+use graph::data::subgraph::schema::SubgraphError;
+use graph::data::subgraph::{
+    schema::{DeploymentCreate, SubgraphManifestEntity},
+    SubgraphFeature,
+};
 use graph::prelude::{
     anyhow, bigdecimal::ToPrimitive, hex, web3::types::H256, BigDecimal, BlockNumber, BlockPtr,
     DeploymentHash, DeploymentState, Schema, StoreError,
 };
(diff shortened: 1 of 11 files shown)


store/postgres/src/deployment.rs
```diff
@@ -890,33 +893,24 @@ pub fn create_deployment(
...
+        d::synced.eq(false),
         d::health.eq(SubgraphHealth::Healthy),
         d::fatal_error.eq::<Option<String>>(None),
         d::non_fatal_errors.eq::<Vec<String>>(vec![]),
         d::earliest_ethereum_block_hash.eq(b(&earliest_block)),
         d::earliest_ethereum_block_number.eq(n(&earliest_block)),
-        d::latest_ethereum_block_hash.eq(b(&latest_block)),
-        d::latest_ethereum_block_number.eq(n(&latest_block)),
+        d::latest_ethereum_block_hash.eq(sql("null")),
+        d::latest_ethereum_block_number.eq(sql("null")),
```
tilacog, line 913: Not critical, but I think this could be expressed as `None`, since those fields are `Nullable`. 
https://github.com/graphprotocol/graph-node/blob/f267b3d2aa6908ce024a94244378b4e91c9f21f5/store/postgres/src/deployment.rs#L69-L70
```suggestion
        d::latest_ethereum_block_hash.eq(None),
        d::latest_ethereum_block_number.eq(None),
```

tilacog (approved):
Awesome.
I liked that the new dedicated struct improves readability a lot.

store/postgres/src/deployment.rs
lutter: Yeah, that's what I tried to do first, but you have to have pretty ugly type annotations to explain the `Option<_>` type to the compiler, so I copped out and just used `null`.

PR #3364, opened 2022-03-15
Store environment variables and flags in a global struct

neysofu:
This PR streamlines environment var. -related logic.

Validation, storage and access all happens in a single place i.e. `graph::env::ENV_VARS`. Chain-specific configuration is available in `graph_chain_ethereum::ENV_VARS` (no configuration needed for NEAR and Tendermint as of now, but we'd use the exact same pattern). We're using [`envconfig`](https://github.com/greyblake/envconfig-rs) internally, which is based on `FromStr`. See also the helpers `graph::env::{NoUnderscores, EnvVarBoolean, WithDefaultUsize}`.

Future work: remove `graph::env::{env_var, unsafe_env_var}` and create our own `unsafe-config` mechanism if needed (see https://github.com/graphprotocol/graph-node/pull/2830).


graph/src/env.rs
```diff
@@ -53,3 +64,955 @@ pub fn env_var<E: std::error::Error + Send + Sync, T: FromStr<Err = E> + Eq>(
...
+    #[envconfig(
+        from = "GRAPH_GRAPHQL_WARN_RESULT_SIZE",
+        default = "18446744073709551615"
+    )]
+    graphql_warn_result_size: WithoutUnderscores<usize>,
+    // usize::MAX
+    #[envconfig(
+        from = "GRAPH_GRAPHQL_ERROR_RESULT_SIZE",
+        default = "18446744073709551615"
+    )]
```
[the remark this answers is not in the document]
neysofu: Unfortunately, it's doesn't. `envy` only supports symbols, meaning we can use default functions and constants, while `envconfig` only supports literal. Once `envy` supports both (https://github.com/serde-rs/serde/issues/368), we should 100% switch to `envy` imo.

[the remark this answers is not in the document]
leoyvens: Personally I'd rather write `default = "ten_thousand"` than `default = "18446744073709551615"`. But it's a bikeshed. @mangas do you want to cast the deciding vote?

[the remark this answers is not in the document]
mangas: I would definitely prefer not to hardcode the value, especially because usize is platform dependent. We could work around this implementing a new type with FromStr which could basically parse this and either panic or default to usize::MAX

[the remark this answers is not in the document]
neysofu: We have no control over `FromStr` if the environment variable is not present at all. We must provide a literal default if we stick with `envconfig`.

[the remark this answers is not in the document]
mangas: In that case default could be "" and it would be handled in the FromStr implementation tough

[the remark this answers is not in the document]
neysofu: Uuuh, I see what you mean now. We can have something like a `struct ParseOrElse<T: FromStr>(T)` and it can work. Really nice!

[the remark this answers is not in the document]
neysofu: Fixed in https://github.com/graphprotocol/graph-node/pull/3364/commits/203d4340dcb06771202c0e48cd75b0fc450ae52b, marking as resolved.

chain/ethereum/src/env.rs
```diff
@@ -0,0 +1,221 @@
...
+            inner: RwLock::new(Inner::from_env()?),
+        })
+    }
+
+    /// Refreshes all internally stored environment variables to reflect the
+    /// current environment variable values.
+    ///
+    /// Only available in debug builds.
+    #[cfg(debug_assertions)]
+    pub fn refresh(&self) -> Result<(), envconfig::Error> {
```
leoyvens, line 52: I believe we decided to go with setters rather than a blanket refresh. My opinions on that:

1. Setters should be added on demand, most env vars are never modified by any test.
2. We should avoid the locks, especially returning lock guards, and especially a coarse lock like this one. It means that in theory borrowing an env var value can block any other code trying to borrow any other env var.
3. Locks should be added at the env var field level, only if the env var has a setter and is not a primitive that has a corresponding atomic type.

mangas: I think we could simplify this design a lot by just having a normal structure with public fields. If you need to test it you can pass the vars as an argument. The static reference should be kept strictly for compatibility and never really added in new code IMO. 

This need for sync is just a byproduct of the wrong design pattern. Env vars should be immutable and they should be a structure that can be built as an argument (maybe we need a build for this, maybe we use a trait so you can mock it), IMO a simple struct is much easier to understnad and use. If we implement Default for that it also helps when passing these for tests.

mangas: Bottomline is that these envs should NOT be built to be mutated, at all IMO

neysofu: Forgot about the `refresh` thing, will update shorty :smile:  thanks for noticing.

As for locking, I agree it's not ideal, but it's a `RwLock` so while there is some overhead indeed, I'd expect the overhead to be constant and not grow with the number of readers, which won't be blocked. Sure, I can do finer-grained locking, even using atomics, it seems a bit like a premature optimization to me, that's all.

> It means that in theory borrowing an env var value can block any other code trying to borrow any other env var.

Am I missing something? It's not a mutex, so that won't happen

neysofu: @mangas if we decide *not* to allow any mutability/refreshing/setters for testing, then I'd absolutely use `pub` fields, I just thought that wasn't the team's consensus at the current time. We can reevaluate at standup today if you feel strongly about this.

mangas: I think we didn't agree in making these arguments everywhere, which I kind of understand because it's prolly a lot of work and a big refactor for little value. That being said I would say if we need to test it we should refactor the relevant code instead of baking in an "easy way" to get around doing it properly.

neysofu: Ok, understood! I believe the only code that needs this pattern in order to be tested rn is the IPFS link resolver, so we can do that. So basically you're suggesting to modify the IPFS link resolver and immediately forgo the whole mutability thing?

neysofu: I'm a bit worried about having some areas of the codebase using the `lazy_static` and some others having env vars as extra arguments though, I feel it might be confusing.

mangas: Yes I think we should fix whatever is immediate and remove the mutability. As for the rest I guess we could check afterwards how big of a refactor it would be to remove the `lazy_static` but at the same rate it it's probably still an improvement and from there we can just use the boyscout rule. Any new code will follow the new pattern and any old code we touch can be brought into the current pattern.

neysofu: Okay. @leoyvens how do you feel this?

leoyvens: My comment on readers blocking each other was wrong because I was thinking of a mutex. If there is a simple refactor for the unit tests that would avoid the need for setters then we can just do that. I didn't understand where we would still need lazy_static (except for `EnvVars`).

In terms of the ideal ultimate solution, I'm not sure yet. Much of this code is called from multiple binaries (graph node, tests, examples, matchstick...) and it is desirable that they can all read the same env vars without re-declaring them. But for testability we want configs to be as arguments so they can be overridden.

Building on @mangas suggestion, perhaps each thing should eventually be refactored to have a `FooConfig`, and `FooConfig::default()` reads from the `EnvVars` globals. This way we have global defaults, that can set by globally defined env vars, but can still override them in tests.

mangas: if we're all in agreement I think we should just remove the complexity, leave an empty struct and implement a Default. We can fix forward from that point 🤔

leoyvens: By complexity you mean remove the getters? It does seem like if there are no setters and no locks, that they are not necessary.

leoyvens: I'm not sure `EnvVars` needs to implement `Default`, as my suggestion is to have another layer of config structs, per component that needs configuration to be overridden by tests, which implements a `Default` that reads from `EnvVars`.

mangas: Right, what I mean is EnvVars should not exist at all. As I understand it, it was just there to provide the locking and abstraction. The EthereumEnvs should have defaults as they also currently do. The packages that use these may choose to have some internal config that is generate from these. But these should be the source of truth for how default is configured, specifically at some point in the future we could generate the flag documentation from it as @tilacog suggested

graph/src/env.rs
```diff
@@ -65,13 +65,26 @@ pub fn env_var<E: std::error::Error + Send + Sync, T: FromStr<Err = E> + Eq>(
 
 pub struct EnvVars {
     inner: Inner,
+    log_query_timing: Vec<String>,
```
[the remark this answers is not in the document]
otaviopace: Looking again perhaps it's better to avoid changing things that can lead to different behavior, since this PR is big enough.

graph/src/env.rs
```diff
@@ -455,12 +463,16 @@ impl EnvVars {
...
-    ///
-    /// Set by the environment variable `GRAPH_GETH_ETH_CALL_ERRORS`, separated
-    /// by `;`.
-    pub fn geth_eth_call_errors(&self) -> &[String] {
-        &self.geth_eth_call_errors
+    /// Set by the flag `GRAPH_ETHEREUM_FETCH_TXN_RECEIPTS_IN_BATCHES`. Enabled
+    /// by default on macOS (to avoid DNS issues) and disabled by default on all
+    /// other systems.
+    pub fn ethereum_fetch_receipts_in_batches(&self) -> bool {
+        let default = cfg!(target_os = "macos");
```
otaviopace: noice, much better 🙌 

I hated that if I've done 😅

core/src/link_resolver.rs
```diff
@@ -192,9 +176,7 @@ impl LinkResolverTrait for LinkResolver {
         )
         .await?;
 
-        // FIXME: Having an env variable here is a problem for consensus.
-        // Index Nodes should not disagree on whether the file should be read.
-        let max_file_size: Option<u64> = read_u64_from_env(MAX_IPFS_FILE_SIZE_VAR);
+        let max_file_size = ENV_VARS.max_ipfs_file_bytes().map(|n| n as u64);
```
otaviopace: Will we have the same `panic` as before with `read_u64_from_env`? I imagine the crate you're using already handles this when trying to parse from the env var right?

Because it will do `u64::from_str` the same way.

graph/src/env.rs
```diff
@@ -850,6 +856,12 @@ struct Inner {
     graphql_max_skip: u32,
     #[envconfig(from = "GRAPHQL_ALLOW_DEPLOYMENT_CHANGE", default = "false")]
     graphql_allow_deployment_change: EnvVarBoolean,
+    // usize::MAX
+    #[envconfig(
+        from = "GRAPH_GRAPHQL_WARN_RESULT_SIZE",
+        default = "18446744073709551615"
```
otaviopace: Not sure if we could patch it, but it would be nice if `envconfig` accepted some kind of computation and/or value such as `usize::MAX`.

graph/src/log/mod.rs
```diff
@@ -43,7 +43,7 @@ pub fn logger(show_debug: bool) -> Logger {
                 FilterLevel::Info
             },
         )
-        .parse(ENV_VARS.log_levels())
+        .parse(ENV_VARS.log_levels().unwrap_or_default())
```
otaviopace: I guess now I see @leoyvens 's point about `Default`.

core/src/link_resolver.rs
```diff
@@ -295,6 +295,10 @@ impl LinkResolverTrait for LinkResolver {
 mod tests {
     use super::*;
     use serde_json::json;
+    use std::env;
+
+    const MAX_IPFS_MAP_FILE_SIZE_VAR: &'static str = "GRAPH_MAX_IPFS_MAP_FILE_SIZE";
+    const MAX_IPFS_FILE_SIZE_VAR: &'static str = "GRAPH_MAX_IPFS_FILE_BYTES";
```
otaviopace: How did this work before? 🤔

otaviopace (commented):
First and foremost, thanks a lot for tackling this ❤️ 

Second, I'm not sure it's very safe to merge this, I know you've done a meticulous job to go env var by env var, but we're talking about a **huge** change, around 100 commits that changes **all of `graph-node` env vars**.

I don't want to get in the way of merging this and I've also came late to the review, but __maybe__ we should split this PR up, what do you guys think? I would like to hear more opinions @mangas @leoyvens

graph/src/env.rs
neysofu: This is fixed already in the latests commits. We're using const generics as a workaround, see `graph::env::WithDefaultUsize`.

neysofu:
> First and foremost, thanks a lot for tackling this heart
> 
> Second, I'm not sure it's very safe to merge this, I know you've done a meticulous job to go env var by env var, but we're talking about a **huge** change, around 100 commits that changes **all of `graph-node` env vars**.
> 
> I don't want to get in the way of merging this and I've also came late to the review, but **maybe** we should split this PR up, what do you guys think? I would like to hear more opinions @mangas @leoyvens

Yeah. I hate to say this because I really want to get this done with, but the responsible thing to do is to split this up.

mangas:
> > First and foremost, thanks a lot for tackling this heart
> > Second, I'm not sure it's very safe to merge this, I know you've done a meticulous job to go env var by env var, but we're talking about a **huge** change, around 100 commits that changes **all of `graph-node` env vars**.
> > I don't want to get in the way of merging this and I've also came late to the review, but **maybe** we should split this PR up, what do you guys think? I would like to hear more opinions @mangas @leoyvens
> 
> Yeah. I hate to say this because I really want to get this done with, but the responsible thing to do is to split this up.

On the one hand I feel like PRs of this size should be split but on the other hand this is 90% boilerplate. The only way that would make sense is to add the structures and then gradually move a handful of vars every PR but that might also take a long time and we'll only end up with a bunch of PRs that we need to get through if some value is off 🤔  So it's good to raise but I'm not sure the solution would be better than what's already done 😛  @leoyvens any thoughts?

core/src/link_resolver.rs
neysofu: They were at the top of the module, now they're only needed for tests! But the newest design allows for:

```rust
let mut env_vars = EnvVars::default();
env_vars.mappings.max_ipfs_file_size = 1337;
```

So these var. names are not there anymore.

core/src/link_resolver.rs
neysofu: Yeah, we'd panic.

graph/src/env.rs
```diff
@@ -122,4 +132,27 @@ struct Inner {
...
+impl FromStr for EnvVarBoolean {
+    type Err = String;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "" | "true" | "1" => Ok(Self(true)),
+            "false" | "0" => Ok(Self(false)),
+            _ => Err("Invalid env. var. flag, expected true / false / 1 / 0".to_string()),
+        }
+    }
```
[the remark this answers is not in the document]
neysofu: :partying_face:

graph/src/log/mod.rs
neysofu: Good point! Will fix this.

graph/src/env.rs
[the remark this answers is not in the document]
neysofu: Marking as resolved, then.

neysofu:
Some of my latest changes:

- Removed all update capabilities for both `graph::env::EnvVars` and `graph_chain_etherum::env::EnvVars`. They are now 100% immutable.
- Transformed all getters into public `struct` fields.
- Fixed conflicts with `master` (`GRAPH_LOG_TIME_FORMAT`).
- Brought `GRAPH_POI_ACCESS_TOKEN` under `EnvVars` (introduced in https://github.com/graphprotocol/graph-node/pull/3348).
- Updated the IPFS link resolver to rely on an explicit `EnvVars` instance, as suggested by @mangas, because its tests relied on env. vars. overwriting.
- Implemented `Default` for `EnvVars`, which is a simple `.clone()` of `ENV_VARS`.

The Git history is a mess, apologies for that. I rebased what I could, but obviously rebasing all getters as struct fields for 100+ commits is impossible, short of creating a new PR.

leoyvens:
I share @otaviopace's concern that this is a big change and it's easy for a mistake to be made, but is someone is willing to do a careful line by line review, checking that the env vars are used in the correct places, have the same name and same default value as before, I'm ok with moving forward, we should just squash when merging. Also only a few env vars are critical, for most getting the value wrong wouldn't be a disaster.

@mangas could you do the line by line review so we're confident merging this?

mangas:
> I share @otaviopace's concern that this is a big change and it's easy for a mistake to be made, but if someone is willing to do a careful line by line review, checking that the env vars are used in the correct places, have the same name and same default value as before, I'm ok with moving forward, we should just squash when merging. Also only a few env vars are critical, for most getting the value wrong wouldn't be a disaster.
> 
> @mangas could you do the line by line review so we're confident merging this?

Happy to help, would be great if we have another pair of eyes as well

mangas (approved):

neysofu:
@mangas thanks for the review. Does https://github.com/graphprotocol/graph-node/pull/3391 remove the `FIREHOSE_FILTERS` env var entirely?

mangas:
Yes it does

neysofu:
I'll wait until standup to see if anybody else wants to do a thorough review before merging.

evaporei (approved):
💯 

One thing though, we might want to change [this](https://github.com/graphprotocol/graph-node/blob/master/docs/environment-variables.md) later.

And also, as mentioned in meetings, this should go through the integration cluster.

Congrats on the big refactor 👏 🚀

Issue #3246, opened 2022-02-14
Refactor the SubgraphRunner (old InstanceManager)


evaporei:
The reasoning behind this is because:
- Currently it's a very fragile piece of code to change and/or add features. It's extremely easy to add bugs there (done myself a couple of times 😅)
- And it's extremely important since it's the core of how indexing works and how blocks are processed. As mentioned before, any mistakes there can change indexing behavior drastically.

I've done a [PR](https://github.com/graphprotocol/graph-node/pull/3245) moving a few structures and functions around but it isn't the refactor yet. I've read the current `instance_manager` and to gather possible components and objectives that this refactor should accomplish, some of them could be: 
- StreamManager (.next/.is_canceled)
- Better define/separate
  - Store usage
  - State
- Handle metrics and logging in a decent way (aka remove some of the clones/cheap_clones and perhaps have a standardized way of using this structure).

File to be refactored: https://github.com/graphprotocol/graph-node/blob/9ebd32fd4d16a0ea3d5d3cc9b2c20274499cce0e/core/src/subgraph/instance_manager.rs

PR #3367, opened 2022-03-17
Refactor `SubgraphRunner` (old `InstanceManager`) - Part 2

evaporei:
This is a continuation of #3245.

Related issue #3246

This Pull Request tries to address two problems / solves two things:
- Make the whole `Runner` and it's member fields/structs more flat, easier to access and give more useful meaning in general to them;
- Divide the `run` code into three clear paths, each handling a `BlockStreamEvent` variant.

Here's a pseudocode that exemplifies the second part:

```rust
impl SubgraphRunner {
    fn run(mut self) -> Result<...> {
        loop {
            let stream = BlockStream::new(...);

            loop {
                let event = stream.next();

                match self.handle_stream_event(event) {
                    Action::Continue => continue,
                    Action::Stop => return Ok(()),
                    Action::Restart => break,
                };
            }
        }
    }

    fn handle_stream_event(event: BlockStreamEvent) -> Result<Action> {
        match event {
            Some(Ok(BlockStreamEvent::ProcessBlock(block, cursor))) =>
                self.handle_process_block(block, cursor),

            Some(Ok(BlockStreamEvent::Revert(revert_to_ptr, cursor))) => 
                self.handle_revert(revert_to_ptr, cursor),

            Some(Err(e)) => self.handle_err(e),

            None => unreachable!("The block stream stopped producing blocks"),
        }
    }
}

enum Action {
    Continue,
    Stop,
    Restart,
}

trait StreamEventHandler<C: Blockchain> {
    async fn handle_process_block(
        &mut self,
        block: BlockWithTriggers<C>,
        cursor: Option<String>,
    ) -> Result<Action, Error>;

    async fn handle_revert(
        &mut self,
        revert_to_ptr: BlockPtr,
        cursor: Option<String>,
    ) -> Result<Action, Error>;

    async fn handle_err(
        &mut self,
        err: CancelableError<Error>,
    ) -> Result<Action, Error>;
}

impl<C, T> StreamEventHandler<C> for SubgraphRunner<C, T>
where
    C: Blockchain,
    T: RuntimeHostBuilder<C>,
{
    async fn handle_process_block(
        &mut self,
        block: BlockWithTriggers<C>,
        cursor: Option<String>,
        cancel_handle: CancelHandle,
    ) -> Result<Action, Error> {
        // current branching that handles a block to be processed
    }

    async fn handle_revert(
        &mut self,
        subgraph_ptr: BlockPtr,
        revert_to_ptr: BlockPtr,
        cursor: Option<String>,
    ) -> Result<Action, Error> {
        // current branching that handles the revert
    }

    async fn handle_err(
        &mut self,
        err: CancelableError<Error>,
    ) -> Result<Action, Error> {
        // current branching that handles the error
    }
}
```

There's still work to be done in:

- Making the new `IndexingState` use methods instead of direct property access to its public fields, so that mutability is more controlled;
- Moving the new `trait` to another module;
- Extracting these methods from the `Runner`:
  - `process_block`;
  - `process_triggers`;
  - `create_dynamic_data_sources`;
  - `persist_dynamic_data_sources`;
- etc.


leoyvens:
We should make sure to test this in the integration cluster before merging.

otaviopace:
@leoyvens agreed 👍

mangas (approved):
This LGTM and is a good cleanup but we should still put this through integration just in case.

evaporei:
Same POI results against `v0.25.2` on the integration cluster 🎉 

Thanks for the review guys! 😊 ❤️

PR #3372, opened 2022-03-18
Near firehose filters

mangas:


chain/near/src/chain.rs
```diff
@@ -223,9 +228,14 @@ impl TriggersAdapterTrait<Chain> for TriggersAdapter {
                         return None;
                     }
 
+                    let receipt = outcome.receipt.as_ref()?.clone();
+                    if !receipt_filter.matches(&receipt.receiver_id) {
```
mangas, line 232: @maoueh @leoyvens does this look correct to you?

chain/near/src/chain.rs
```diff
@@ -235,7 +245,10 @@ impl TriggersAdapterTrait<Chain> for TriggersAdapter {
             .map(|r| NearTrigger::Receipt(Arc::new(r)))
             .collect();
 
-        trigger_data.push(NearTrigger::Block(shared_block.cheap_clone()));
+        // TODO: Confirm this is correct
+        if block_filter.trigger_every_block {
```
mangas, line 249: @maoueh @leoyvens does this look correct to you?

Should we always have a trigger even if we have no block handlers?

chain/near/src/chain.rs
maoueh: So you are filtering earlier in data pipeline more or less correct?

mangas: Previously this was not filtered, so all receipts would be added to the stream. This makes it so that only the subscribed ones will be processed. I guess the short answer is yes :)

chain/near/src/adapter.rs
```diff
@@ -1,17 +1,28 @@
...
 }
 
 impl bc::TriggerFilter<Chain> for TriggerFilter {
     fn extend<'a>(&mut self, data_sources: impl Iterator<Item = &'a DataSource> + Clone) {
-        self.block
-            .extend(NearBlockFilter::from_data_sources(data_sources));
+        self.block_filter
+            .extend(NearBlockFilter::from_data_sources(data_sources.clone()));
+        self.receipt_filter
+            .extend(NearReceiptFilter::from_data_sources(data_sources));
```
leoyvens, line 25: Lets destructure `self`.

chain/near/src/adapter.rs
```diff
@@ -25,10 +36,71 @@ impl bc::TriggerFilter<Chain> for TriggerFilter {
...
+/// NearReceiptFilter requires the account to be set, it will match every receipt where `source.account` is the recipient.
+/// see docs: https://thegraph.com/docs/en/supported-networks/near/
+#[derive(Clone, Debug, Default)]
+pub(crate) struct NearReceiptFilter {
+    pub accounts: HashSet<Account>,
+}
+
+impl NearReceiptFilter {
+    pub fn matches(&self, account: &String) -> bool {
+        !self.is_empty() && self.accounts.contains(account)
```
leoyvens, line 74: The `is_empty` check is redundant.

chain/near/src/adapter.rs
```diff
@@ -25,10 +36,71 @@ impl bc::TriggerFilter<Chain> for TriggerFilter {
...
+        accounts.is_empty()
+    }
+
+    pub fn from_data_sources<'a>(iter: impl IntoIterator<Item = &'a DataSource>) -> Self {
+        iter.into_iter()
+            .filter(|data_source| {
+                data_source.source.account.is_some()
+                    && !data_source.mapping.receipt_handlers.is_empty()
+            })
+            .fold(Self::default(), |mut filter_opt, data_source| {
```
leoyvens, line 89: This is more of a `collect` than a `fold`.

chain/near/src/chain.rs
maoueh: To me, it's a good change but from my understanding, it changes a bit the paradigm the `graph-node` is using. 

Transform everything to `triggers` then check if the triggers match or not. Now it's more if the the block does not contain anything of interest, filter it right away.

I imagine it might change the logs a bit, but should improve performance reducing the needs to actually create triggers that will ultimately do not match.

We should probably update the other Firehose implementation (expect Ethereum which is different) to use this pattern.

chain/near/src/adapter.rs
```diff
@@ -37,7 +109,7 @@ pub(crate) struct NearBlockFilter {
 impl NearBlockFilter {
     pub fn from_data_sources<'a>(iter: impl IntoIterator<Item = &'a DataSource>) -> Self {
         iter.into_iter()
-            .filter(|data_source| data_source.source.account.is_some())
+            .filter(|data_source| !data_source.mapping.block_handlers.is_empty())
```
leoyvens, line 112:
```rust
let trigger_every_block = iter.into_iter().any(|data_source| !data_source.mapping.block_handlers.is_empty());
```

chain/near/src/chain.rs
leoyvens: It looks right to me, the trigger should only be added if there are block handlers.

leoyvens (approved):

chain/near/src/chain.rs
sduchesneau: ```data_source.mapping.block_handlers.is_empty``` is the only condition that sets trigger_every_block to true, so I don't think it could trigger without a block handler (this may be the same thing that @leoyvens is saying right before I hit 'Comment')

chain/near/src/chain.rs
maoueh: Checking more, if we go that route, we would probably need to change other part of the code like [match_and_decode](https://github.com/graphprotocol/graph-node/blob/5e7e4b3e9ea656bf3dc23907efed1c676fe8224a/chain/near/src/data_source.rs#L43)

The conditional in there is not required anymore because there would be no `Trigger` at all because it's filtered up-front.

chain/near/src/chain.rs
maoueh: All comments about filtering up in https://github.com/graphprotocol/graph-node/pull/3372/files#r830173700 applies here also.

mangas: https://github.com/graphprotocol/graph-node/pull/3372/files#diff-99488584ebb0c0295142cf3018e8097c0cd6b5bdf74964b23b1d7fa23d42ba8aR112 this is why i was asking. Before we were using source being set as filter rather than block handlers existing. And we would also set this to true in that case. 

Now it's exactly as you describe, it will only be true if there is a block handler

chain/near/src/chain.rs
```diff
@@ -348,3 +361,146 @@ impl FirehoseMapperTrait<Chain> for FirehoseMapper {
...
+            .into_iter()
+            .filter_map(|x| match x {
+                crate::trigger::NearTrigger::Block(b) => b.header.clone().map(|x| x.height),
+                _ => None,
+            })
+            .collect()
+    }
+
+    fn random_block(height: u64, receiver_id: &String) -> codec::Block {
+        codec::Block {
```
maoueh, line 472: Beautiful!

maoueh (changes requested):
If we keep it the pre-filtering in `TriggerAdapter`, we need to change logic in `match_and_decode`.

I'm implicitly approving the pre-filter way of doing things.

chain/near/src/adapter.rs
mangas: Not sure I follow what you mean here.

mangas: do you mean map then collect and then .extend on the hashset?

chain/near/src/adapter.rs
mangas: nice :)

chain/near/src/chain.rs
leoyvens: The architecture allows optional filtering within the firehose, and then within the block stream, as optimizations. But I don't see how that affects `match_and_decode`, which matches at the data source level.

mangas:
> If we keep it the pre-filtering in `TriggerAdapter`, we need to change logic in `match_and_decode`.
> 
> I'm implicitly approving the pre-filter way of doing things.

https://github.com/graphprotocol/graph-node/blob/2f770b039669a4454666f7ffcba6735229dea70d/chain/near/src/data_source.rs#L43

The match_and_decode implementation seems to be compatible unless I'm missing something here

chain/near/src/adapter.rs
leoyvens: Not even extend, just initialize `accounts` to the collect.

chain/near/src/chain.rs
leoyvens: I expected it to have randomness given the name. I much prefer `handcrafted_artisan_rust_sourced_block`.

mangas: 🤣  I'll rename

maoueh:
Discussed with @mangas. We agreed the `match_and_decode` can stay like this as it does no harm. But it opens a bigger discussion that the `match_and_decoder` has lost the semantic of its name now.

We agreed to move on and discuss more broadly the overall flow as it probably affects all future Firehose integration.

For example, Solana transforms each and every instructions to a trigger to later filtering what, 95% of them out. Probably it's the same on Tendermint.

maoueh (approved):

chain/near/src/adapter.rs
mangas: ah yes of course! 🤦

chain/near/src/chain.rs
mangas: I think it's probably worth discussing down the line, if we eventually arrive at the conclusion that all the chains fit within our "new" paradigm then we can refactor accordingly.

PR #3375, opened 2022-03-18
tests: Add integration test for start blocks

Jannis:
This adds an integration test for the start blocks feature.


Jannis:
I don't understand how this can have passed the start blocks integration test. Will need to test again locally.

leoyvens:
If this is revisited it should be with the new end-to-end test fixture, so closing for now.

PR #3382, opened 2022-03-21
store: Temporarily turn prefix comparison off for Bytes attrs

lutter:


store/postgres/src/relational.rs
```diff
@@ -1164,10 +1164,13 @@ impl Column {
     /// lengths. Such columns may contain very large values and need to be
     /// handled specially for indexing
     pub fn has_arbitrary_size(&self) -> bool {
+        // TODO: Turn this back on for `ColumnType::Bytes`. Before we can do
+        // that, we need to make sure that the generated queries work with
+        // both old (no prefix index) and new schemas
```
leoyvens, line 1169: Lets link the comment with the ignored test, so the test is not forgotten.

leoyvens (approved):

PR #3391, opened 2022-03-23
Firehose filter config

mangas:


graph/src/firehose/endpoints.rs
```diff
@@ -25,6 +25,7 @@ pub struct FirehoseEndpoint {
     pub provider: String,
     pub uri: String,
     pub token: Option<String>,
+    pub filters: bool,
```
maoueh, line 28: `filtersEnabled`? At first read seems like a list.

maoueh (approved):
Approving despite the comment about the `filters` name so nothing is blocked.

graph/src/firehose/endpoints.rs
mangas: renamed :)

PR #3396, opened 2022-03-25
Refactor `SubgraphRunner::run`

neysofu:
As per title.

No functional changes expected, this PR simply breaks the huge `SubgraphRunner::run` method in multiple pieces, easier to reason and understand. cc @tilacog


neysofu:
This is something I worked on earlier, but didn't realize it overlapped so much with https://github.com/graphprotocol/graph-node/pull/3367. Closing, maybe reopening something similar in the future if we still want to improve things.

Issue #3072, opened 2021-12-15
File data sources


Jannis:
GIP https://forum.thegraph.com/t/gip-file-data-sources/2721

Status:
- [x] Manifest parsing #3743
- [x] File Monitor PR #3411
- [x] DB representation of file data sources #3742
- [x] Subgraph runner and runtime
- [x] Causality region isolation
- [ ] graph-cli support

leoyvens:
### File data sources engineering plan

This describes the components required for an implementation of file data sources backed directly by an ipfs client and therefore non-deterministic, as described in this GIP https://forum.thegraph.com/t/gip-file-data-sources/2721. There is still uncertainity if data from multiple chains would be indexed with a total order of events or not. The intention of this first pass is to have the simplest possible non-deterministic implementation, so it is not blocked by the design of the deterministic version.

#### File monitor

The file monitor receives requests to monitor the availability of files, and once the file become available it returns its contents through a channel. It should also notify when a file becomes unavailable. Each graph node process will instantiate one file monitor to serve all subgraphs.

For ipfs, the monitor will loop through the queue of files to check and pool them with `object.stat`. limited by max number of requests in flight and a minimum time between rechecks. To scale this component to a load of millions of files, multiple queues with different priority levels can be implemented so that new requests are pooled more frequently than old requests.

#### DB representation of file data sources 

Dynamic file data sources need to be stored in the DB. I can think of a few options:

1. Reuse `dynamic_ethereum_contract_data_source` .

2. Introduce a specialized table for file data sources.

3. Introduce a new table to be used for dynamic data sources for all new blockchains, probably with a json field for network-specific fields.

Option 3 seems attractive as it would be a more general solution, but it requires a more careful design of a schema that serves all blockchains.

#### Data source independency

Data sources and entities are now associated with an access group, which is a numeric id. An entity has the same access group as the data source that created it. Only entites from the same access group are visible to a data source mapping calling `store.get`. If two access groups attempt to set a same entity, that is a non-determinstic error until we figure out the ideal behaviour.

Non-file data sources all have access group 0. Each file data source gets its own access group. Group ids are represented in the cache and in the DB, to enforce access rules.

#### File data source execution

Because availability chain history cannot be reliably replayed, the modification of subgraph state by file data source must not depend on availability history. Therefore a file data source handler should not able to know how many times it ran in the past due to changes in availablility.

[...]

PR #3411, opened 2022-03-30
Polling file monitor

leoyvens:
As part of #3072, this adds the components to poll IPFS for the availability of a file. This does _not_ track unavailability, once a file is found it is not polled again.

The load management approach is to have a configurable capacity for the IPFS service, which is shared among all subgraphs, and then a per-subgraph queue of pending requests. This should effectively distribute the service capacity fairly among subgraphs.

The implementation has two pieces, the monitoring algorithm which is generic over any `tower::Service` and an implementation of `Service` for IPFS.

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -23,8 +23,11 @@ semver = "1.0.12"
 serde = "1.0"
 serde_json = "1.0"
 serde_yaml = "0.8"
-
+# tower = { git = "https://github.com/tower-rs/tower.git", features = ["util"] } # SOON 0.5
+tower = { git = "https://github.com/leoyvens/tower.git", features = ["util", "limit"], branch = "fix-call-all-pending-stream" } # SOON 0.5
 graph-runtime-wasm = { path = "../runtime/wasm" }
+# tower-test = { git = "https://github.com/tower-rs/tower.git" } # DEV DEP
+tower-test = { git = "https://github.com/leoyvens/tower.git", branch = "fix-call-all-pending-stream"} # DEV DEP
 
 [dev-dependencies]
 graph-mock = { path = "../mock" }
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -23,13 +23,14 @@ semver = "1.0.12"
 serde = "1.0"
 serde_json = "1.0"
 serde_yaml = "0.8"
-# tower = { git = "https://github.com/tower-rs/tower.git", features = ["util"] } # SOON 0.5
-tower = { git = "https://github.com/leoyvens/tower.git", features = ["util", "limit"], branch = "fix-call-all-pending-stream" } # SOON 0.5
+# Switch to crates.io once tower 0.5 is released
+tower = { git = "https://github.com/tower-rs/tower.git", features = ["util", "limit"] }
 graph-runtime-wasm = { path = "../runtime/wasm" }
-# tower-test = { git = "https://github.com/tower-rs/tower.git" } # DEV DEP
-tower-test = { git = "https://github.com/leoyvens/tower.git", branch = "fix-call-all-pending-stream"} # DEV DEP
+cid = "0.8.3"
+anyhow = "1.0"
 
 [dev-dependencies]
+tower-test = { git = "https://github.com/tower-rs/tower.git" }
 graph-mock = { path = "../mock" }
 walkdir = "2.3.2"
 test-store = { path = "../store/test-store" }
--- a/core/src/polling_monitor/ipfs_service.rs
+++ b/core/src/polling_monitor/ipfs_service.rs
@@ -1,0 +1,105 @@
+use anyhow::{anyhow, Error};
+use bytes::Bytes;
+use cid::Cid;
+use futures::{Future, FutureExt};
+use graph::{
+    cheap_clone::CheapClone,
+    ipfs_client::{IpfsClient, StatApi},
+    tokio::sync::Semaphore,
+};
+use std::{pin::Pin, sync::Arc, task::Poll, time::Duration};
+use tower::Service;
+
+const CLOUDFLARE_TIMEOUT: u16 = 524;
+const GATEWAY_TIMEOUT: u16 = 504;
+
+#[derive(Clone)]
+pub struct IpfsService {
+    client: IpfsClient,
+    max_file_size: u64,
+    timeout: Duration,
+    concurrency_limiter: Arc<Semaphore>,
+}
+
+impl CheapClone for IpfsService {
+    fn cheap_clone(&self) -> Self {
+        Self {
+            client: self.client.cheap_clone(),
+            max_file_size: self.max_file_size,
+            timeout: self.timeout,
+            concurrency_limiter: self.concurrency_limiter.cheap_clone(),
+        }
+    }
+}
+
+impl IpfsService {
+    #[allow(dead_code)]
+    pub fn new(
+        client: IpfsClient,
+        max_file_size: u64,
+        timeout: Duration,
+        concurrency_limit: u16,
+    ) -> Self {
+        Self {
+            client,
+            max_file_size,
+            timeout,
+            concurrency_limiter: Arc::new(Semaphore::new(concurrency_limit as usize)),
+        }
+    }
+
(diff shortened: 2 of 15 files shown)


core/Cargo.toml
```diff
@@ -23,10 +23,14 @@ serde = "1.0"
 serde_json = "1.0"
 serde_yaml = "0.8"
 fail = "0.5"
-
+# Switch to crates.io once tower 0.5 is released
```
evaporei, line 26: Perhaps add a `TODO: ` prefix here? 🤔

graph/src/log/mod.rs
```diff
@@ -53,6 +53,10 @@ pub fn logger(show_debug: bool) -> Logger {
     Logger::root(drain, o!())
 }
 
+pub fn discard() -> Logger {
+    Logger::root(slog::Discard, o!())
```
evaporei, line 57: Dumb question, why/what is this for? 🤔

core/src/polling_monitor/mod.rs
```diff
@@ -0,0 +1,232 @@
...
+        let monitor = spawn_monitor(MockService(svc), tx, log::discard());
+
+        // Basic test, single file is immediately available.
+        monitor.monitor("req-0");
+        send_response(&mut handle, Some("res-0")).await;
+        assert_eq!(rx.recv().await, Some(("req-0", "res-0")));
+    }
+
+    #[tokio::test]
+    async fn polling_monitor_unordered() {
```
evaporei, line 165: I'm not sure I've followed how this tests the "unorderedness" of the PollingMonitor since the test does indeed send/receive in a specific order, can you explain it a bit further please?

core/src/polling_monitor/mod.rs
```diff
@@ -0,0 +1,232 @@
...
+                        }
+                    }
+
+                    // Object not found, push the id to the back of the queue.
+                    Ok((id, None)) => queue.lock().push_back(id),
+
+                    // Error polling, log it and push the id to the back of the queue.
+                    Err((id, e)) => {
+                        debug!(logger, "error polling";
+                                    "error" => format!("{:#}", e),
```
evaporei, line 79: Noice

core/src/polling_monitor/mod.rs
```diff
@@ -0,0 +1,232 @@
...
+
+pub mod ipfs_service;
+
+/// Spawn a monitor that actively polls a service. Whenever the service has capacity, the monitor
+/// pulls object ids from the queue and polls the service. If the object is not present or in case
+/// of error, the object id is pushed to the back of the queue to be polled again.
+///
+/// The service returns the request ID along with errors or responses. The response is an
+/// `Option`, to represent the object not being found. `S::poll_ready` should never error.
+pub fn spawn_monitor<ID, S, E, Response: Send + 'static>(
```
evaporei, line 22: Would it make sense for the API to be `PollingMonitor::spawn` instead of a detached function `spawn_monitor`?

Perhaps you could even create a private `new` with these two lines:

```rust
    let queue = Arc::new(Mutex::new(VecDeque::new()));
    let (wake_up_queue, queue_woken) = watch::channel(());
```

core/src/polling_monitor/ipfs_service.rs
```diff
@@ -0,0 +1,97 @@
...
+        }
+    }
+
+    async fn call(&self, cid: Cid) -> Result<Option<Bytes>, Error> {
+        let cid_str = cid.to_string();
+        let stat = match self.client.object_stat(&cid_str, self.timeout).await {
+            Ok(stat) => stat,
+            Err(e) => match e.status().map(|e| e.as_u16()) {
+                Some(GATEWAY_TIMEOUT) | Some(CLOUDFLARE_TIMEOUT) => return Ok(None),
+                _ if e.is_timeout() => return Ok(None),
```
evaporei, line 53: Could these two `match` arms be together? 🤔

core/src/polling_monitor/ipfs_service.rs
```diff
@@ -0,0 +1,97 @@
...
+use std::{pin::Pin, sync::Arc, task::Poll, time::Duration};
+use tower::Service;
+
+const CLOUDFLARE_TIMEOUT: u16 = 524;
+const GATEWAY_TIMEOUT: u16 = 504;
+
+#[derive(Clone)]
+pub struct IpfsService {
+    client: IpfsClient,
+    max_file_size: u64,
```
evaporei, line 15: Do we know how much this will be?

evaporei (approved):
The logic seems correct, I've added a bunch of comments for further discussion 🙂

core/Cargo.toml
```diff
@@ -23,10 +23,14 @@ serde = "1.0"
...
+tower = { git = "https://github.com/tower-rs/tower.git", features = ["util", "limit"] }
 graph-runtime-wasm = { path = "../runtime/wasm" }
+cid = "0.8.3"
+anyhow = "1.0"
 
 [dev-dependencies]
+tower-test = { git = "https://github.com/tower-rs/tower.git" }
```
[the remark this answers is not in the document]
leoyvens: The above is `tower`, this is `tower-test`. They live in the same repo.

graph/src/log/mod.rs
leoyvens: Convenience to construct a logger that discards messages.

core/src/polling_monitor/ipfs_service.rs
```diff
@@ -0,0 +1,97 @@
...
+    timeout: Duration,
+    concurrency_limiter: Arc<Semaphore>,
+}
+
+impl CheapClone for IpfsService {
```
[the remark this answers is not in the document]
leoyvens: We could, I like the idea.

core/src/polling_monitor/mod.rs
leoyvens: It sends in the order `req-0`, `req-1`, meaning `req-1` is called first on the service. But the response for `req-0`  arrives first and is received without waiting for `req-1`. This tests that we're using `.call_all(stream).unordered()`.

core/src/polling_monitor/ipfs_service.rs
leoyvens: This was the best I came up with but open to suggestions.

core/src/polling_monitor/ipfs_service.rs
leoyvens: For our other IPFS uses we use an env var `GRAPH_MAX_IPFS_FILE_BYTES` that defaults to 256MB. I'll probably just start with that when IpfsService needs to be initialized somewhere.

core/src/polling_monitor/mod.rs
leoyvens: `PollingMonitor::spawn` is something I considered, but I wanted to try something different because `PollingMonitor` is just a handle to the task, and to me `Handle::spawn()` reads a bit backwards compared to `fn spawn() -> Handle`.

leoyvens:
There's no rush to merge this, so I'll leave it as PR for a while to see if tower 0.5 is released in the meantime.

core/src/polling_monitor/ipfs_service.rs
leoyvens: That wouldn't work, the `if` clause applies to the entire pattern, I just tried it on the [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=7290983d0e09110ef5710411c85958b4).

core/src/polling_monitor/mod.rs
```diff
@@ -0,0 +1,232 @@
...
+            }
+        })
+    };
+
+    {
+        let queue = queue.cheap_clone();
+        graph::spawn(async move {
+            let mut responses = service.call_all(queue_to_stream).unordered().boxed();
+            while let Some(response) = responses.next().await {
+                match response {
```
Theodus, line 64: For each `PollingMonitor<S>`, I think it would make sense to have metrics for:
- queue depth
- timeout rate
- error rate

Theodus (approved):
LGTM

PR #3413, opened 2022-03-31
Add hardware requirements to README.md

### LINKED ISSUE #3412: clang: error: unable to execute command: Killed
Hi,

I've been following the tutorial so far by using the cheapest Hetzner Cloud CX11 node and at the step of building graph-node with `cargo`, I'm getting `clang: error: unable to execute command: Killed`.

I found online that it has to do with the machine's ram. But I've now rescaled to 4GB and it still doesn't work. Is that normal? If so, I think these requirements should be listed in the "Get Started"

TimDaub:
- Fixes #3412

I ended up being able to build with cargo after rescaling my server to 8GB RAM.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,10 @@ To build and run this project you need to have the following installed on your s
 
 For Ethereum network data, you can either run your own Ethereum node or use an Ethereum node provider of your choice.
 
+**Minimum Hardware Requirements:**
+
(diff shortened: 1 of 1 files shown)


leoyvens (approved):
Thanks for measuring this and adding it to the README!

PR #3416, opened 2022-03-31
Fix some graphman bugs and improve usability

lutter:
The `graphman index` commands wouldn't work if the same deployment is used by multiple names.

This PR also overhauls the logic for how deployments are looked up to make the lookup uniform across various commands

### DIFF
--- a/node/src/manager/commands/assign.rs
+++ b/node/src/manager/commands/assign.rs
@@ -1,41 +1,52 @@
-use std::sync::Arc;
+use graph::prelude::{anyhow::anyhow, Error, NodeId};
+use graph_store_postgres::{command_support::catalog, connection_pool::ConnectionPool};
 
-use graph::{
-    prelude::{anyhow::anyhow, Error, NodeId, SubgraphStore as _},
-    slog::Logger,
-};
-use graph_store_postgres::SubgraphStore;
+use crate::manager::deployment::DeploymentSearch;
 
-use crate::manager::deployment::locate;
+pub async fn unassign(primary: ConnectionPool, search: &DeploymentSearch) -> Result<(), Error> {
+    let locator = search.locate_unique(&primary)?;
 
-pub async fn unassign(
-    logger: Logger,
-    store: Arc<SubgraphStore>,
-    hash: String,
-    shard: Option<String>,
-) -> Result<(), Error> {
-    let deployment = locate(store.as_ref(), hash, shard)?;
+    let conn = primary.get()?;
+    let conn = catalog::Connection::new(conn);
 
-    println!("unassigning {}", deployment);
-    store
-        .writable(logger, deployment.id)
-        .await?
-        .unassign_subgraph()?;
+    let site = conn
+        .locate_site(locator.clone())?
+        .ok_or_else(|| anyhow!("failed to locate site for {locator}"))?;
+
+    println!("unassigning {locator}");
+    conn.unassign_subgraph(&site)?;
 
     Ok(())
 }
 
 pub fn reassign(
-    store: Arc<SubgraphStore>,
-    hash: String,
+    primary: ConnectionPool,
+    search: &DeploymentSearch,
     node: String,
-    shard: Option<String>,
 ) -> Result<(), Error> {
     let node = NodeId::new(node.clone()).map_err(|()| anyhow!("illegal node id `{}`", node))?;
-    let deployment = locate(store.as_ref(), hash, shard)?;
+    let locator = search.locate_unique(&primary)?;
 
-    println!("reassigning {} to {}", deployment, node.as_str());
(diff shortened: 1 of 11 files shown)


node/src/manager/deployment.rs
```diff
@@ -179,11 +181,17 @@ pub fn find_single_deployment_locator(
...
+            .into_iter()
+            .map(|deployment| deployment.locator()),
+    )
+    .into_iter()
+    .collect();
+    let deployment_locator = match locators.len() {
+        0 => anyhow::bail!("Found no deployment for `{}`", name),
+        1 => locators.pop().unwrap(),
+        n => anyhow::bail!("Found {} deployments for `{}`", n, name),
+    };
```
tilacog: Interesting!
I didn't realize a deployment could have multiple names.

node/src/manager/deployment.rs
```diff
@@ -18,6 +21,127 @@ use graph_store_postgres::{command_support::catalog as store_catalog, Shard, Sub
...
+                let query = query.filter(ds::subgraph.eq(&hash));
+                match shard {
+                    Some(shard) => query.filter(ds::shard.eq(shard)).load(conn)?,
+                    None => query.load(conn)?,
+                }
+            }
+            DeploymentSearch::Deployment { namespace } => {
+                query.filter(ds::name.eq(&namespace)).load(conn)?
+            }
+        };
```
tilacog, line 122: This is awesome.

node/src/manager/commands/run.rs
```diff
@@ -31,6 +30,15 @@ use graph_core::{
...
 
+fn locate(store: &dyn SubgraphStore, hash: &str) -> Result<DeploymentLocator, anyhow::Error> {
+    let mut locators = store.locators(&hash)?;
+    match locators.len() {
+        0 => bail!("could not find subgraph {hash} we just created"),
+        1 => Ok(locators.pop().unwrap()),
+        n => bail!("there are {n} subgraphs with hash {hash}"),
+    }
+}
+
```
tilacog, line 41: This _"get a single item from a collection"_ pattern happens so often that I created a convenience function for it [here](https://github.com/graphprotocol/graph-node/pull/3270/files#diff-e85bd4cf7aa1c61ad3464c284292e8adedecd56161573109222d0e397714ab6cR218-R230). 
In the future, I'll try to make that even more generic/flexible so we can reuse it.

node/src/manager/commands/assign.rs
```diff
@@ -1,41 +1,52 @@
...
+                println!("deployment {locator} is already assigned to {cur}");
+            } else {
+                println!("reassigning {locator} to {node} (was {cur})");
+                conn.reassign_subgraph(&site, &node)?;
+            }
+        }
+        None => {
+            println!("assigning {locator} to {node}");
+            conn.assign_subgraph(&site, &node)?;
+        }
```
tilacog, line 48: That is very convenient.

tilacog (approved):

tilacog:
While reviewing this, I noticed that the `DeploymentId` inner value is being accessed directly outside the `store` crate, even though it's not supposed to:

https://github.com/graphprotocol/graph-node/blob/9da5493dbdc3fa7421179b9970d237d0556aedad/graph/src/components/store/mod.rs#L762-L767

Do you think we could prevent this by changing its inner field visibility and implementing a getter?

```diff
-pub struct DeploymentId(pub i32);
+pub struct DeploymentId(pub(crate) i32);
...
+pub fn id(&self) -> i32 { self.0 }
```

Please let me know if this would be useful, and I'll come up with a follow-up PR :)

node/src/manager/deployment.rs
lutter: Yeah, that's what keeps things interesting. When people deploy an IPFS hash that was already running under a different name, we just put an entry in `subgraph` and `subgraph_version` to point to that deployment. So the subgraph name -> deployment mapping is many-to-one

node/src/manager/commands/run.rs
lutter: The reason this is special and not used with `DeploymentSearch` is that that wants the primary pool, and we only have a `SubgraphStore` here. The `run` code could probably be changed to also access the primary pool. But it's a smaller issue (and `run` needs a lot more TLC anyway)

lutter:
> Do you think we could prevent this by changing its inner field visibility and implementing a getter?
> 
> ```diff
> -pub struct DeploymentId(pub i32);
> +pub struct DeploymentId(pub(crate) i32);
> ...
> +pub fn id(&self) -> i32 { self.0 }
> ```
> 
> Please let me know if this would be useful, and I'll come up with a follow-up PR :)

I don't have strong opinions either way, but a getter feels cleaner than than just `self.0` - maybe a `to_i32()` function would be better than `id`

PR #3418, opened 2022-04-01
runner: Change transact block error mesage

leoyvens:
And remove the match by leveraging `context` and `?`. Read with whitespace diff disabled.


graph/src/blockchain/mod.rs
```diff
@@ -149,7 +149,7 @@ pub enum IngestorError {
     ReceiptUnavailable(H256, H256),
 
     /// An unexpected error occurred.
-    #[error("Ingestor error: {0}")]
```
leoyvens, line 152: This is an unrelated improvement to print ingestor errors with causes.

core/src/subgraph/runner.rs
```diff
@@ -361,56 +361,49 @@ where
...
+        store
+            .transact_block_operations(
+                block_ptr,
+                firehose_cursor,
+                mods,
+                stopwatch,
+                data_sources,
+                deterministic_errors,
+            )
+            .context("Failed to transact block operations")?;
```
evaporei, line 373: Will this create multiple lines like @lutter mentioned the other day? 🤔

evaporei (approved):

core/src/subgraph/runner.rs
leoyvens: Only if printed as `{:?}`, and the places that print this use `{:#}` and actually explicitly remove any newlines that sneak in due to wasm stack traces https://github.com/graphprotocol/graph-node/blob/102a96ba38393a1bc7ad3403acd7698e9bc07fba/core/src/subgraph/runner.rs#L719

PR #3420, opened 2022-04-02
Remove SubgraphRunner Clones 🔫 

evaporei:
This is a refactor that removes unnecessary:

- Parameters/Arguments;
- Variables;
- Clones.

And chooses usage of `self.` or `&` (references) instead 😉 

Related issue #3246

### DIFF
--- a/core/src/subgraph/instance.rs
+++ b/core/src/subgraph/instance.rs
@@ -188,9 +188,9 @@ where
     pub(crate) fn add_dynamic_data_source(
         &mut self,
         logger: &Logger,
-        data_source: C::DataSource,
-        templates: Arc<Vec<C::DataSourceTemplate>>,
-        metrics: Arc<HostMetrics>,
+        data_source: &C::DataSource,
+        templates: &Arc<Vec<C::DataSourceTemplate>>,
+        metrics: &Arc<HostMetrics>,
     ) -> Result<Option<Arc<T::Host>>, Error> {
         // Protect against creating more than the allowed maximum number of data sources
         if let Some(max_data_sources) = ENV_VARS.subgraph_max_data_sources {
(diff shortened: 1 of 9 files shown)


core/src/subgraph/instance.rs
```diff
@@ -188,9 +188,9 @@ where
     pub(crate) fn add_dynamic_data_source(
         &mut self,
         logger: &Logger,
-        data_source: C::DataSource,
-        templates: Arc<Vec<C::DataSourceTemplate>>,
-        metrics: Arc<HostMetrics>,
+        data_source: &C::DataSource,
```
mangas, line 191: if these are going to be cloned further down maybe it's worth letting the caller clone it and in this case I think you could just pass the ownership down in line 212

core/src/subgraph/runner.rs
```diff
@@ -217,16 +201,15 @@ where
         // very contrived subgraph would be able to observe this.
         while block_state.has_created_data_sources() {
             // Instantiate dynamic data sources, removing them from the block state.
-            let (data_sources, runtime_hosts) = self.create_dynamic_data_sources(
-                logger.clone(),
-                host_metrics.clone(),
-                block_state.drain_created_data_sources(),
-            )?;
+            let (data_sources, runtime_hosts) =
+                self.create_dynamic_data_sources(block_state.drain_created_data_sources())?;
```
mangas, line 205: I wonder if it would be clearer to pass a &mut blockstate here and then let that function deal with the draining

mangas (approved):

core/src/subgraph/instance.rs
evaporei: Great suggestion! Just fixed on 971275892d915cc5361454fa92851dde42163a17

core/src/subgraph/runner.rs
evaporei: Hmmm, I kind of like the [current definition](https://github.com/graphprotocol/graph-node/blob/master/graph/src/components/subgraph/instance.rs#L69) that does an assert first.

Not that it couldn't be improved. Well I'll leave this one for out of scope of this PR, thanks for the review!!

PR #3425, opened 2022-04-04
LinkResolver refactor

mangas:


lutter (approved):
LGTM. It would be good to have a commit message that explains why this change is necessary

tests/Cargo.toml
```diff
@@ -4,10 +4,16 @@ version = "0.25.2"
...
 tokio-stream = "0.1"
 futures = "0.3.13"
 port_check = "0.1.5"
 anyhow = "1.0.54"
 lazy_static = "1.4.0"
+graph-node = { path = "../node" }
+graph-core = { path = "../core" }
+graph = { path = "../graph" }
+graph-mock = { path = "../mock" }
+graph-store-postgres = { path = "../store/postgres" }
```
leoyvens, line 19: Why did all these deps become necessary?

mangas: This is for code that's not yet committed, the tests I've been working on

leoyvens: Then they should be removed if they are dead deps for now.

leoyvens: Then they should be removed if they are dead deps for now.

core/src/subgraph/instance_manager.rs
```diff
@@ -154,7 +144,7 @@ where
                 deployment.hash.cheap_clone(),
                 manifest,
                 // Allow for infinite retries for subgraph definition files.
-                &self.link_resolver.as_ref().clone().with_retries(),
```
leoyvens, line 157: Was the `with_retries()` lost in the refactor?

mangas: https://github.com/graphprotocol/graph-node/pull/3447

PR #3426, opened 2022-04-04
chain/ethereum: Fix broken deployment_eth_rpc_* metrics registration

Jannis:
I noticed the following errors being logged frequently:

    ERRO registering metric [deployment_eth_rpc_request_duration] because it was already registered, component: MetricsRegistry
    ERRO registering metric [deployment_eth_rpc_errors] because it was already registered, component: MetricsRegistry

I also noticed that graph-node wouldn't correctly observe these metrics because of this. The reason for the errors was that the metrics were recreated each time a new block stream for a deployment was created.

This commit changes it so we remember the `SubgraphEthRpcMetric` associated with each deployment instead of trying to recreate them each time.

### DIFF
--- a/chain/ethereum/src/chain.rs
+++ b/chain/ethereum/src/chain.rs
@@ -23,9 +23,9 @@ use graph::{
     },
 };
 use prost::Message;
-use std::collections::HashSet;
+use std::collections::{HashMap, HashSet};
 use std::iter::FromIterator;
-use std::sync::Arc;
+use std::sync::{Arc, Mutex};
 
 use crate::data_source::DataSourceTemplate;
 use crate::data_source::UnresolvedDataSourceTemplate;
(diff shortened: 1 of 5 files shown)


leoyvens:
I'd suggest having `SubgraphEthRpcMetrics` internally use `fn global_counter_vec` (and we'll need to add `fn global_gauge_vec`).

chain/ethereum/src/chain.rs
```diff
@@ -157,7 +159,15 @@ impl Blockchain for Chain {
             self.eth_adapters.cheapest_with(capabilities)?.clone()
         };
 
-        let ethrpc_metrics = Arc::new(SubgraphEthRpcMetrics::new(self.registry.clone(), &loc.hash));
+        let ethrpc_metrics = self
+            .deployment_rpc_metrics
+            .lock()
+            .unwrap()
+            .entry(String::from(loc.hash.as_str()))
```
Jannis, line 166: I'm using an `Arc<Mutex` here to avoid changing `triggers_adapter` to take a `&mut self`. But that could be an alternative.

leoyvens: `&mut` on `Chain` instances is unlikely to work. See my suggestion above to not have to track this map in the `Chain`.

Jannis:
@leoyvens Good point, that _is_ cleaner. Pushed the change to this PR.

leoyvens (approved):

PR #3427, opened 2022-04-04
env vars: Do not derive debug

leoyvens:


leoyvens:
Changed this to implement debug, but without printing any values.

PR #3435, opened 2022-04-05
Add Firehose blockstream metrics

Jannis:
This adds a few metrics for Firehose block streams, all labeled with deployment and provider. Things that these metrics are tracking:

- How often do we reconnect block streams?
- How long do connection attempts take until they succeed/fail?
- How many of these reconnection attempts are successful/fail?
- How much time do we spend receiving and processing stream events?
- What kinds of events (proceed, revert, error) are we receiving?

### DIFF
--- a/core/src/metrics/registry.rs
+++ b/core/src/metrics/registry.rs
@@ -18,6 +18,7 @@ pub struct MetricsRegistry {
     global_counter_vecs: Arc<RwLock<HashMap<u64, CounterVec>>>,
     global_gauges: Arc<RwLock<HashMap<u64, Gauge>>>,
     global_gauge_vecs: Arc<RwLock<HashMap<u64, GaugeVec>>>,
+    global_histogram_vecs: Arc<RwLock<HashMap<u64, HistogramVec>>>,
 }
 
 impl MetricsRegistry {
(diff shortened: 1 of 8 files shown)


core/src/metrics/registry.rs
```diff
@@ -217,6 +219,28 @@ impl MetricsRegistryTrait for MetricsRegistry {
...
+        let histograms = HistogramVec::new(opts, variable_labels)?;
+        let id = histograms.desc().first().unwrap().id;
+        let maybe_histogram = self.global_histogram_vecs.read().unwrap().get(&id).cloned();
+        if let Some(histograms) = maybe_histogram {
+            Ok(histograms)
+        } else {
+            self.register(name, Box::new(histograms.clone()));
+            self.global_histogram_vecs
+                .write()
+                .unwrap()
```
evaporei, line 238: Just to confirm, all of these `.unwrap`s are safe to do right?

evaporei (approved):

core/src/metrics/registry.rs
leoyvens: This by definition only panics if something else has already panicked, and in this use there is no possibility of panic while holding the lock.

PR #3438, opened 2022-04-06
move manifest parsing

mangas:


core/src/subgraph/instance_manager.rs
```diff
@@ -137,42 +201,6 @@ impl<S: SubgraphStore> SubgraphInstanceManager<S> {
         // that is done
         store.start_subgraph_deployment(&logger).await?;
```
leoyvens, line 138: This comment is being explicit about the intent of starting the deployment before loading the data sources, and this PR changes it to the opposite.

leoyvens:
@mangas can this be closed?

PR #3440, opened 2022-04-07
docs: Add how to implement a new chain

evaporei:
This document describes how developers can add a new chain to `graph-node`.

It doesn't go over everything in detail, but it's a generalized guide that we can start with and improve on top of.

### DIFF
--- a/docs/implementation/add-chain.md
+++ b/docs/implementation/add-chain.md
@@ -1,0 +1,274 @@
+# Adding support for a new chain
+
+## Context
+
+`graph-node` started as a project that could only index EVM compatible chains, eg: `ethereum`, `xdai`, etc.
+
+It was known from the start that with growth we would like `graph-node` to be able to index other chains like `NEAR`, `Solana`, `Cosmos`, list goes on...
+
+However to do it, several refactors were necessary, because the code had a great amount of assumptions based of how Ethereum works.
+
+At first there was a [RFC](https://github.com/graphprotocol/rfcs/blob/10aaae30fdf82f0dd2ccdf4bbecf7ec6bbfb703b/rfcs/0005-multi-blockchain-support.md) for a design overview, then actual PRs such as:
+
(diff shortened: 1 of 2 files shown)


leoyvens:
If this were in rustdoc rather than a markdown, then we might be able to require that the examples compile, so they won't become outdated.

evaporei:
@leoyvens but I don't think we want that for those, the examples I've provided are incomplete, the traits require a lot of method implementations.

leoyvens:
Alright, we'll just remember to keep this updated.

evaporei:
You know, actually you just gave me an idea, we could use things like `todo!`s and `unimplemented!` macros in the code. Ok next iteration of this should use this then, great idea!!!

leoyvens:
@evaporei isn't that the perfect use case for a cargo [example](https://doc.rust-lang.org/cargo/reference/cargo-targets.html#examples).

pienkowb (approved):
Looks good to me 👍

docs/implementation/add-chain.md
```diff
@@ -0,0 +1,274 @@
...
+
+- https://github.com/graphprotocol/graph-node/pull/2272
+- https://github.com/graphprotocol/graph-node/pull/2292
+- https://github.com/graphprotocol/graph-node/pull/2399
+- https://github.com/graphprotocol/graph-node/pull/2411
+- https://github.com/graphprotocol/graph-node/pull/2453
+- https://github.com/graphprotocol/graph-node/pull/2463
+- https://github.com/graphprotocol/graph-node/pull/2755
+
+All new chains, besides the EVM compatible ones, are integrated using [StreamingFast](https://www.streamingfast.io/)'s [Firehose](https://github.com/streamingfast/firehose). The integration consists of chain specific `protobuf` files with the type definitions.
```
abourget, line 21: Maybe link to firehose.streamingfast.io

evaporei: Fixed!

PR #3441, opened 2022-04-07
store: Organize relational module

leoyvens:
The `relational.rs` file felt too big so this organizes it a bit. The ddl generation logic is moved to the `ddl.rs` submodule. The test model is extracted to its own file.


lutter (approved):
Nice! Definitely helps readability

PR #3443, opened 2022-04-07
Put receipts into `Arc`s to prevent excessive cloning.

tilacog:
This PR aims to improve performance issues introduced in #3373.

It envelops `TransactionReceipt`s in `Arcs`s to share them without cloning inside the `parse_log_triggers` function.

### DIFF
--- a/chain/ethereum/src/codec.rs
+++ b/chain/ethereum/src/codec.rs
@@ -284,6 +284,9 @@ impl Into<EthereumBlockWithCalls> for &Block {
                             effective_gas_price: None,
                         })
                     })
+                    // Transaction receipts will be shared along the code, so we put them into an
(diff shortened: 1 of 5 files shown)


tilacog:
The second commit in this PR  preventively aims to eliminate a [redundant vector allocation](https://github.com/graphprotocol/graph-node/pull/3443/commits/be507bdcd77e288ae61ccbc2897c66ac0e6c1ac5#diff-b31a71a95f48aa7d0f11b5ea3501befc2560d749d73872e9eebe8003de85d921R1058-R1060) introduced in the first commit, but it required `Arc` to be used in more places than it might be necessary.

We could consider ignoring/reverting the second commit if we want to use less `Arc`s around.

leoyvens (approved):

PR #3444, opened 2022-04-07
Restore lost concurrency when scanning for triggers

tilacog:
This PR aims to improve performance issues introduced in #3373.

It reestablishes concurrency when scanning for triggers in a block range and resolving a block hash.

### DIFF
--- a/chain/ethereum/src/ethereum_adapter.rs
+++ b/chain/ethereum/src/ethereum_adapter.rs
@@ -13,7 +13,7 @@ use graph::prelude::tokio::try_join;
 use graph::{
     blockchain::{block_stream::BlockWithTriggers, BlockPtr, IngestorError},
     prelude::{
-        anyhow::{self, anyhow, bail, ensure},
+        anyhow::{self, anyhow, bail, ensure, Context},
         async_trait, debug, error, ethabi,
         futures03::{self, compat::Future01CompatExt, FutureExt, StreamExt, TryStreamExt},
         hex, info, retry, serde_json as json, stream, tiny_keccak, trace, warn,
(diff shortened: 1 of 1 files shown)


chain/ethereum/src/ethereum_adapter.rs
```diff
@@ -1347,24 +1347,28 @@ pub(crate) async fn blocks_with_triggers(
...
+        })
+        .compat();
+
+    // Join on triggers and block hash resolution
+    let (triggers, to_hash) = futures03::join!(trigger_futs.try_concat(), to_hash_fut);
+
+    // Unpack and handle possible errors in the previously joined futures
+    let triggers =
+        triggers.with_context(|| format!("Failed to obtain triggers for block {}", to))?;
+    let to_hash = to_hash.with_context(|| format!("Failed to infer hash for block {}", to))?;
```
tilacog, line 1371: I decided not to use `try_join!` as it discards one error if the two operations fail.

leoyvens: Well, this will discard the `to_hash` error if both fail, so it's not so different from `try_join`, which would be fine imo.

tilacog: You have a point.

I was reluctant to use `try_join` because it would only return the first error to resolve and discard the slowest, so I was concerned about indeterminism in this case. 

If omitting a second error situation isn't a problem, I can change this to use `try_join` or just leave it as it is.
If not, I can rewrite this to provide a combined error message for when we have two errors.

Which one do you think would be preferable?

leoyvens: interesting point about determinism. I think we can keep this as-is then.

leoyvens (approved):

PR #3446, opened 2022-04-08
Use reserved discriminant segments for `IndexForAscTypeId`

tilacog:
This PR changes the organization of  `IndexForAscTypeId` variants for future blockchain implementations.

**Motivation**:
The incentive for this change is the expected (and desired) growth of supported blockchains in Graph-Node. Each new blockchain implementation also comes with a myriad of new *AssemblyScript* types defined in [Graph-TS](github.com/graphprotocol/graph-ts) that need to be represented in the runtime. Currently, we have been appending new type IDs to the end of `IndexForAscTypeId` enum, but the addition of new types proved to be convoluted and hard to reason about. We aim to improve the developer experience by removing this kind of friction.

**Rationale**:
We propose that each blockchain implementation will have a fixed and contiguous discriminant space of 500 items, starting to be enforced after our latest released implementations: **Ethereum** and **Near**.

Our goal is that every new supported blockchain, for instance, **Tendermint**, will have room to include new AssemblyScript class IDs without interweaving its discriminants with other blockchains.

**TODO**:
- We need to reorganize the variants in graph-ts' [`TypeId`](https://github.com/graphprotocol/graph-ts/blob/main/global/global.ts#L15) enum to exactly match the variants and their discriminants rearranged by this PR.

### DIFF
--- a/graph/src/runtime/mod.rs
+++ b/graph/src/runtime/mod.rs
@@ -139,12 +139,23 @@ macro_rules! impl_asc_type {
 
 impl_asc_type!(u8, u16, u32, u64, i8, i32, i64, f32, f64);
 
-// The numbers on each variant could just be comments hence the
-// `#[repr(u32)]`, however having them in code enforces each value
-// to be the same as the docs.
+/// Contains type IDs and their discriminants for every blockchain supported by Graph-Node.
+///
+/// Each variant corresponds to the unique ID of an AssemblyScript concrete class used in the
+/// [`runtime`].
+///
+/// # Rules for updating this enum
+///
+/// 1 .The discriminants must have the same value as their counterparts in `TypeId` enum from
+///    graph-ts' `global` module. If not, the runtime will fail to determine the correct class
+///    during allocation.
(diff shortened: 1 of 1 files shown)


graph/src/runtime/mod.rs
```diff
@@ -243,70 +245,94 @@ pub enum IndexForAscTypeId {
...
+    // NextNearType = 87,
+    // AnotherNearType = 88,
+    // ...
+    // LastNearType = 999,
+
+    // Reserved discriminant space for more Ethereum type IDs: [1000, 1499]
+    TransactionReceipt = 1000,
+    Log = 1001,
+    ArrayH256 = 1002,
+    ArrayLog = 1003,
```
lutter, line 260: I assume these are all not in real use yet?

lutter (approved):

graph/src/runtime/mod.rs
tilacog: Positive.
Those types were created for #3373, which haven't gone live yet.

PR #3451, opened 2022-04-10
Assume that immutable entities are actually new during indexing 

lutter:
This is a better approach to get to the same result as in PR #3449 - we already know that there shouldn't be any point in trying to look up new immutable entities from the store.

We therefore optimize the code for the common happy path, that users use immutable entities correctly to avoid store interactions that are useless in that case to improve indexing performance.

### DIFF
--- a/graph/src/data/schema.rs
+++ b/graph/src/data/schema.rs
@@ -579,6 +579,8 @@ pub struct Schema {
 
     // Maps an interface name to the list of entities that implement it.
     pub types_for_interface: BTreeMap<EntityType, Vec<ObjectType>>,
+
+    immutable_types: HashSet<EntityType>,
 }
 
 impl Schema {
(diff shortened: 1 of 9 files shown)


graph/src/data/schema.rs
```diff
@@ -579,19 +579,32 @@ pub struct Schema {
...
-    /// schema, and should not be used otherwise
-    pub fn new(id: DeploymentHash, document: s::Document) -> Self {
-        Schema {
-            id,
+    /// Create a new schema. The document must already have been validated
+    //
+    // TODO: The way some validation is expected to be done beforehand, and
+    // some is done here makes it incredibly murky whether a `Schema` is
+    // fully validated. The code should be changed to make sure that a
+    // `Schema` is always fully valid
```
evaporei, line 592: This section is using `//` instead of `///` purposely right?

evaporei (approved):

PR #3454, opened 2022-04-11
add runner end-to-end test - setup

mangas:
Basic setup and simple runner test that does no work.


graph/src/blockchain/block_stream.rs
```diff
@@ -80,8 +82,30 @@ pub trait BlockStream<C: Blockchain>:
 {
 }
 
-pub trait BlockStreamBuilder<C: Blockchain> {
-    fn build() -> Box<dyn BlockStream<C>>;
+/// BlockStreamBuilder is an abstraction that would separate the logic for building streams from the blockchain trait
+/// the goal is that this api is as close as possible to existing implementations in order to make the refactor easier.
```
leoyvens, line 86: This second line of the comment is useful as context for this PR, but I wouldn't keep it as a permanent comment since then it's not clear what refactor this is talking about.

graph/src/blockchain/mod.rs
```diff
@@ -84,7 +84,7 @@ pub trait Blockchain: Debug + Sized + Send + Sync + Unpin + 'static {
     type TriggersAdapter: TriggersAdapter<Self>;
 
     /// Trigger data as parsed from the triggers adapter.
-    type TriggerData: TriggerData + Ord + Send + Sync;
+    type TriggerData: TriggerData + Ord + Send + Sync + Clone;
```
leoyvens, line 87: Needing a `Clone` bound is always unfourtunate. Since this is only for tests, I'd prefer to remove it here and add `where C::TriggerData: Clone` bounds where necessary, to flag the places where cloning happens.

tests/tests/runner.rs
```diff
@@ -0,0 +1,299 @@
...
+
+#[async_trait]
+impl<C: Blockchain> BlockStreamBuilder<C> for StaticStreamBuilder<C> {
+    fn build_firehose(
+        &self,
+        _chain: &C,
+        _deployment: DeploymentLocator,
+        _block_cursor: Option<String>,
+        _start_blocks: Vec<graph::prelude::BlockNumber>,
+        _subgraph_current_block: Option<graph::blockchain::BlockPtr>,
```
leoyvens, line 253: Here and in `build_polling`, we probably want to use the `block_cursor` or  the `subgraph_current_block` as a starting index into `self.events`, for tests that exercise block stream restarts. You don't need to think through this now, but it's worth adding a comment so we don't forget.

leoyvens (changes requested):

leoyvens (approved):

PR #3468, opened 2022-04-15
runner: Ignore skip block ptr if close to chain head

evaporei:
This will make the `sync status` more robust, since the `skip block pointer updates timer` will be skipped if the subgraph is close to the chain head.

### DIFF
--- a/core/src/subgraph/runner.rs
+++ b/core/src/subgraph/runner.rs
@@ -586,6 +586,13 @@ where
         if block.trigger_count() == 0
             && self.state.skip_ptr_updates_timer.elapsed() <= SKIP_PTR_UPDATES_THRESHOLD
             && !self.state.synced
+            && !close_to_chain_head(
+                &block_ptr,
+                self.inputs.chain.chain_store().cached_head_ptr().await?,
+                // The "skip ptr updates timer" is ignored when a subgraph is at most 1000 blocks
+                // behind the chain head.
+                1000,
+            )
         {
             return Ok(Action::Continue);
         } else {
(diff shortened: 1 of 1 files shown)


core/src/subgraph/runner.rs
```diff
@@ -608,9 +615,12 @@ where
             Ok(action) => {
                 // Once synced, no need to try to update the status again.
                 if !self.state.synced
-                    && is_deployment_synced(
+                    && close_to_chain_head(
                         &block_ptr,
                         self.inputs.chain.chain_store().cached_head_ptr().await?,
+                        // We consider a subgraph synced when it's at least 1 block behind the
```
leoyvens, line 621:
```suggestion
                        // We consider a subgraph synced when it's at most 1 block behind the
```

core/src/subgraph/runner.rs
```diff
@@ -586,6 +586,13 @@ where
         if block.trigger_count() == 0
             && self.state.skip_ptr_updates_timer.elapsed() <= SKIP_PTR_UPDATES_THRESHOLD
             && !self.state.synced
+            && !close_to_chain_head(
+                &block_ptr,
+                self.inputs.chain.chain_store().cached_head_ptr().await?,
+                // The "skip ptr updates timer" is ignored when a subgraph is at least 1000 blocks
```
leoyvens, line 592:
```suggestion
                // The "skip ptr updates timer" is ignored when a subgraph is at most 1000 blocks
```

leoyvens (approved):

core/src/subgraph/runner.rs
evaporei: Fixed, thanks!

core/src/subgraph/runner.rs
evaporei: Fixed, thanks!

PR #3485, opened 2022-04-21
new stable-hash implementation

mangas:


graph/src/data/store/scalar.rs
```diff
@@ -193,30 +192,58 @@ impl bigdecimal::ToPrimitive for BigDecimal {
...
+
 impl StableHash for BigDecimal {
-    fn stable_hash<H: StableHasher>(&self, mut sequence_number: H::Seq, state: &mut H) {
+    fn stable_hash<H: stable_hash::StableHasher>(&self, field_address: H::Addr, state: &mut H) {
         let (int, exp) = self.as_bigint_and_exponent();
         // This only allows for backward compatible changes between
         // BigDecimal and unsigned ints
-        exp.stable_hash(sequence_number.next_child(), state);
-        BigInt(int).stable_hash(sequence_number, state);
+        StableHash::stable_hash(&exp, field_address.child(0), state);
```
That3Percent, line 214: The comment says this allows for backward compatible changes, but this implementation does not provide backward compatibility for schema changes as advertised.

One of the defining features of the `stable-hash` crate is that it allows (limited) schema evolution without breaking existing hashes. Adding new fields to a structure is one allowable change, but so is changing between integer types. For example, you can change from a `u8` to an `i16` since the new type contains the old type. Changing from `u32` -> `BigInt` -> `BigDecimal` should also be supported.

This implementation does not allow for those schema evolutions, because it uses `field_adddress.child(1)` for the data in the `BigInt`.

Using the new `stable-hash` API, we can not only support a conversion to `BigDecimal` from an unsigned int (as the comment suggests), but we can support conversion from signed ints as well!

We happen to know that ints are written using `field_address` for a trimmed array of the little endian bytes, and `.child(0)` for the negative sign (if any). So using `.child(1)` for the exponent will be backward compatible without causing collisions.

Typically this kind of assumption would be dangerous, but it's reasonable in this case because `stable-hash` _can't_ change this implementation detail without breaking the contract.

graph/src/data/store/scalar.rs
```diff
@@ -193,30 +192,58 @@ impl bigdecimal::ToPrimitive for BigDecimal {
...
     }
 }
 
+impl StableHash for BigInt {
+    fn stable_hash<H: stable_hash::StableHasher>(&self, field_address: H::Addr, state: &mut H) {
+        AsInt {
+            is_negative: self.0.sign() == BigIntSign::Minus,
+            little_endian: &self.to_bytes_le().1,
+        }
+        .stable_hash(field_address.child(0), state)
```
That3Percent, line 243: For backward compatible schema evolution, this should omit this `.child(0)`

graph/src/data/store/scalar.rs
```diff
@@ -509,9 +536,19 @@ impl fmt::Debug for Bytes {
...
+    ) {
+        stable_hash_legacy::utils::AsBytes(&self.0).stable_hash(sequence_number, state);
+    }
+}
+
 impl StableHash for Bytes {
-    fn stable_hash<H: StableHasher>(&self, sequence_number: H::Seq, state: &mut H) {
-        AsBytes(&self.0).stable_hash(sequence_number, state);
+    fn stable_hash<H: stable_hash::StableHasher>(&self, field_address: H::Addr, state: &mut H) {
+        AsBytes(&self.0).stable_hash(field_address.child(0), state);
```
That3Percent, line 551: This should omit the `.child(0)` as a transparent wrapper

graph/src/data/subgraph/schema.rs
```diff
@@ -201,20 +202,53 @@ impl Display for SubgraphError {
...
+    ) {
+        let SubgraphError {
+            subgraph_id,
+            message,
+            block_ptr,
+            handler,
+            deterministic,
+        } = self;
+        stable_hash_legacy::StableHash::stable_hash(
+            &subgraph_id,
```
That3Percent, line 219: Nit: These `&` aren't be required because it's inferred by the destructuring from `&self`

graphql/src/execution/execution.rs
```diff
@@ -89,16 +88,45 @@ struct HashableQuery<'a> {
 /// different representations. This is considered not an issue. The worst
 /// possible outcome is that the same query will have multiple cache entries.
 /// But, the wrong result should not be served.
+impl stable_hash_legacy::StableHash for HashableQuery<'_> {
```
That3Percent, line 91: I don't think we need the legacy version for this type since it's not used in the PoI. We can of course keep it around, but if you want to find opportunities to remove bloat this is one. There are probably some other types this applies to as well.

That3Percent (changes requested):

graph/src/data/store/scalar.rs
mangas: I made some changes here but not sure this is what you suggested or if you wanted something like AsInt to be used. 

If you could let me know exactly what the code should look like it might be easier

That3Percent:
```rust
// This implementation allows for backward compatible changes from integers (signed or unsigned)
// when the exponent is zero.
let (int, exp) = self.as_bigint_and_exponent();
exp.stable_hash(field_address.child(1));
// Normally it would be a red flag to pass field_address in after having used a child slot.
// But, we know the implementation of StableHash for BigInt will not use child(1) and that
// it will not in the future due to having no forward schema evolutions for ints and the
// stability guarantee.
//
// For reference, ints use child(0) for the sign and write the little endian bytes to the parent slot.
int.stable_hash(field_address);
```

That3Percent: Can we write a unit test to verify backward compatibility to make sure?

graph/src/blockchain/types.rs
```diff
@@ -81,10 +81,27 @@ pub struct BlockPtr {
...
 impl CheapClone for BlockPtr {}
 
+impl stable_hash_legacy::StableHash for BlockPtr {
+    fn stable_hash<H: stable_hash_legacy::StableHasher>(
+        &self,
+        mut sequence_number: H::Seq,
+        state: &mut H,
+    ) {
+        stable_hash_legacy::utils::AsBytes(self.hash.0.as_ref())
+            .stable_hash(sequence_number.next_child(), state);
```
That3Percent, line 91: I see that you added a destructure to the one struct that I put the comment on, but this comment applies to all structs/tuples implementing `StableHash` (modern and legacy)

That3Percent (approved):
Nice! Let a small comment, but it's optional.

graph/src/blockchain/types.rs
mangas: yep sure I'll add those, totally forgot

mangas: fixed

That3Percent:
Next time it would be easier to review if instead of using force-push we did a squash and merge at the end. Seeing deltas makes it easier to not review the entire change from scratch and not miss anything.

PR #3489, opened 2022-04-21
store: Avoid panic if GRAPH_STORE_CONNECTION_MIN_IDLE exceeds pool_size

lutter:


leoyvens (approved):
Huh I didn't know we had this config

PR #3492, opened 2022-04-25
Make traversing derived relationships work when using `Bytes` ids

lutter:
There will also be tests that expose this issue, and make sure it stays fixed. But since that involves an extensive rewrite of `graphql/tests/query.rs`, it will happen in a separate PR so we can at least roll out this fix first.

### DIFF
--- a/graph/src/env/store.rs
+++ b/graph/src/env/store.rs
@@ -98,6 +98,11 @@ pub struct EnvVarsStore {
     /// Setting this to `0` disables pipelined writes, and writes will be
     /// done synchronously.
     pub write_queue_size: usize,
+
+    /// This is just in case new behavior causes issues. This can be removed
+    /// once the new behavior has run in the hosted service for a few days
+    /// without issues.
+    pub disable_error_for_toplevel_parents: bool,
 }
 
 // This does not print any values avoid accidentally leaking any sensitive env vars
@@ -139,6 +144,7 @@ impl From<InnerStore> for EnvVarsStore {
             connection_min_idle: x.connection_min_idle,
             connection_idle_timeout: Duration::from_secs(x.connection_idle_timeout_in_secs),
             write_queue_size: x.write_queue_size,
+            disable_error_for_toplevel_parents: x.disable_error_for_toplevel_parents.0,
         }
     }
 }
@@ -184,4 +190,6 @@ pub struct InnerStore {
     connection_idle_timeout_in_secs: u64,
     #[envconfig(from = "GRAPH_STORE_WRITE_QUEUE", default = "5")]
     write_queue_size: usize,
+    #[envconfig(from = "GRAPH_DISABLE_ERROR_FOR_TOPLEVEL_PARENTS", default = "false")]
+    disable_error_for_toplevel_parents: EnvVarBoolean,
 }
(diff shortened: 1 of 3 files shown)


store/postgres/src/relational_queries.rs
```diff
@@ -2171,6 +2203,26 @@ impl<'a> FilterCollection<'a> {
...
+            FilterCollection::SingleWindow(window) => Ok(Some(window.parent_type())),
+            FilterCollection::MultiWindow(windows, _) => {
+                let parent_types: HashSet<_> =
+                    HashSet::from_iter(windows.into_iter().map(|window| window.parent_type()));
+                if parent_types.len() > 1 {
+                    return Err(graph::constraint_violation!(
+                        "all implementors of an interface must use the same type for their `id`"
+                    ));
+                }
+                Ok(parent_types.iter().next().map(|t| t.to_owned()))
```
neysofu, line 2222: We could also use `all_equal()`, as a possibly clearer alternative to the `HashSet` and `> 1` logic:

```suggestion
                if windows.iter().map(FilterWindow::parent_type).all_equal() {
                    Ok(Some(windows[0].parent_type().to_owned()))
                } else {
                    Err(graph::constraint_violation!(
                        "all implementors of an interface must use the same type for their `id`"
                    ))
                }
```

Just food for thought; feel free to ignore this.

store/postgres/src/relational_queries.rs
```diff
@@ -473,8 +474,32 @@ impl EntityData {
...
+                                    out.insert_entity_data("g$parent_id".to_owned(), value);
+                                } else {
+                                    // A query that does not have parents
+                                    // somehow returned parent ids. We have no
+                                    // idea how to deserialize that
+                                    if parent_type.is_none() {
+                                        return Err(graph::constraint_violation!(
+                                            "query unexpectedly produces parent ids"
+                                        ));
+                                    }
```
neysofu, line 495: Why do we need this check? I think it's redundant, at this point in the code we already know that `parent_type.is_none()`.

neysofu (changes requested):
Looks good, just some minor comments.

graph/src/env/store.rs
```diff
@@ -184,4 +190,6 @@ pub struct InnerStore {
     connection_idle_timeout_in_secs: u64,
     #[envconfig(from = "GRAPH_STORE_WRITE_QUEUE", default = "5")]
     write_queue_size: usize,
+    #[envconfig(from = "GRAPH_DISABLE_ERROR_FOR_TOPLEVEL_PARENTS", default = "false")]
+    disable_error_for_toplevel_parents: bool,
```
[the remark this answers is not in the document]
lutter: Nice! I didn't know that existed. Changed.

store/postgres/src/relational_queries.rs
lutter: Yes, that's absolutely right. Fixed.

store/postgres/src/relational_queries.rs
lutter: I didn't know about `all_equal`. I like your way a lot better. Changed.

lutter:

neysofu (approved):
Looks great! Feel free to merge 🌈

PR #3502, opened 2022-04-27
Rewrite graphql query tests

lutter:
The rewrite was prompted by needing to run the same tests with `Bytes` and `String` as the `id` type in a subgraph. It also streamlines writing tests a lot and reduces the notational overhead of existing tests. That should make it easier to write more tests.

### DIFF
--- a/store/postgres/src/relational.rs
+++ b/store/postgres/src/relational.rs
@@ -665,21 +665,31 @@ impl Layout {
                 }
                 query.load::<EntityData>(conn)
             })
-            .map_err(|e| match e {
-                diesel::result::Error::DatabaseError(
-                    diesel::result::DatabaseErrorKind::__Unknown,
-                    ref info,
-                ) if info.message().starts_with("syntax error in tsquery") => {
-                    QueryExecutionError::FulltextQueryInvalidSyntax(info.message().to_string())
+            .map_err(|e| {
+                use diesel::result::DatabaseErrorKind;
+                use diesel::result::Error::*;
+                // Sometimes `debug_query(..)` can't be turned into a
+                // string, e.g., because `walk_ast` for one of its fragments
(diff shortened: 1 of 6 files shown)


leoyvens:
Failure on `test graft`:
```
thread 'graft' panicked at 'deleting test entities succeeds: Unknown(deadlock detected)', store/postgres/tests/graft.rs:266:10
```
This seems to be a spurious failure, not the first time I see it. I'll try re-running.

lutter:
> Failure on `test graft`:
> 
> ```
> thread 'graft' panicked at 'deleting test entities succeeds: Unknown(deadlock detected)', store/postgres/tests/graft.rs:266:10
> ```
> 
> This seems to be a spurious failure, not the first time I see it. I'll try re-running.

Yes, I've had that before, too but haven't been able to pinpoint why that's happening (I don't think it's related to this PR) It seems to happen when one test is tearing down its fixtures and the other is setting its fixtures up at the same time. Might be related to pipelined store writes.

store/postgres/src/relational.rs
```diff
@@ -665,21 +665,31 @@ impl Layout {
...
+            .map_err(|e| {
+                use diesel::result::DatabaseErrorKind;
+                use diesel::result::Error::*;
+                // Sometimes `debug_query(..)` can't be turned into a
+                // string, e.g., because `walk_ast` for one of its fragments
+                // returns an error. When that happens, avoid a panic from
+                // simply calling `to_string()` on it, and output a string
+                // representation of the `FilterQuery` instead of the SQL
+                let mut query_text = String::new();
+                match write!(query_text, "{}", debug_query(&query_clone)) {
```
leoyvens, line 677: This is such a footgun, nice fix!

leoyvens (approved):

Issue #2097, opened 2020-12-26
Use TlsMode::Prefer instead of TlsMode:None in notification_listener


joe-bowman:
In notification_listener, we create a postgresql connection with the param `TlsMode::None`. There is no reason to not use `TlsMode::Prefer` here and not force operators to run postgresql with `Host` over `SslHost` in pg_hba.conf.

https://github.com/graphprotocol/graph-node/blob/39287ae65a1827d0fbaf79cbc91844549ac00576/store/postgres/src/notification_listener.rs#L121

Elsewhere, `Diesel` is used and this respects the `postgres_url` passed on the command line.

joe-bowman:
Sadly, this is not as easy as it seems.

The postgres v0.15.2 library support for TLS requires an _old_ version of openssl - and upgrading runs into various issues (notably mutable borrowing of the Postgresql Client object for both Client.notifications() and a new JsonNotification object).

It'd be great to get this looked into, because disabling TLS for this feels the wrong thing to do..

leoyvens:
@joe-bowman are you still interested in this? We've since updated the `postgres` crate, if you want to try this again.

qiaozhengyuan:
Would hope this change gets deployed as when I tried to run the docker image with my own postgresql service, it failed because my postgresql enforces tls.

wup-dps:
Any update here? Does the graph-nodel support postgres over SSL?

JasoonS:
> Does the graph-node support postgres over SSL?

AFAIK not unfortunately. It is a big pity since most self maintained graphs would use a postgres instance that requires SSL.

mb-shorai:
> The postgres v0.15.2 library support for TLS requires an _old_ version of openssl - and upgrading runs into various issues (notably mutable borrowing of the Postgresql Client object for both Client.notifications() and a new JsonNotification object).

@joe-bowman how can one reproduce the issue?

mb-shorai:
I guess it can be closed now that https://github.com/graphprotocol/graph-node/pull/3503 is merged, although there is some room for improvements (e.g. don't turn the certificate validation off, but make it configurable).

PR #3503, opened 2022-04-27
Support TLS PostgreSQL connections in notification_listener

mb-shorai:
Hi, this PR implements TLS PSQL connections support for the `notification_listener`, as various managed database providers allow encrypted connections only, and it is generally a bad idea to disable encryption.

Related issue: https://github.com/graphprotocol/graph-node/issues/2097

With these changes it is possible to connect to databases with both TLS support and without. Please note that certificate verification is currently **disabled**, because it fits my use-case, and I don't have any experience with Rust whatsoever to make it configurable to support more use-cases. So I would be grateful if someone takes it over from here.

You can test it e.g. locally when running Postgres with the following command:

```yaml
version: '3'
services:
  postgres:
    image: postgres
    ports:
      - '5432:5432'
    command:
      [
        "postgres",
        "-cshared_preload_libraries=pg_stat_statements",
        "-cssl=on",
        "-cssl_cert_file=/etc/ssl/certs/ssl-cert-snakeoil.pem",
        "-cssl_key_file=/etc/ssl/private/ssl-cert-snakeoil.key"
      ]
    environment:
      POSTGRES_USER: graph-node
      POSTGRES_PASSWORD: let-me-in
      POSTGRES_DB: graph-node
```

And monitor Postgres connections with the following command:

```sql
SELECT datname,usename,ssl,client_addr,application_name
FROM pg_stat_ssl
JOIN pg_stat_activity
ON pg_stat_ssl.pid = pg_stat_activity.pid;
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1708,8 +1708,10 @@ dependencies = [
  "lazy_static",
  "lru_time_cache",
  "maybe-owned",
+ "openssl",
  "pin-utils",
  "postgres",
+ "postgres-openssl",
  "rand",
  "serde",
  "stable-hash",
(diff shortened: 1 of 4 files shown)


azf20:
Thanks @mb-shorai ! @lutter interested in your thoughts here?

lutter (approved):
Nice! I just checked that our current setup still works with these changes, and it does! Could you rebase onto latest master? I'll then merge this PR.

lutter:
NM - just found the rebase button :sunglasses:

lutter:
Merged it - thanks a ton for doing this!

Issue #3405, opened 2022-03-29
New DB representation for data sources


leoyvens:
The DB storage of dynamic data sources needs to change to reflect that graph node is now multichain and to fullfill the requirements of file data sources.

Dynamic data sources are currently stored in `sugraphs.dynamic_ethereum_contract_data_source`. This proposes that each subgraph will gain a `sgd*.data_source` table under its own DB namespace. The proposed schema for this table is:

```sql
create table sgd*.data_sources$ (
      vid integer generated by default as identity primary key,
      block_range int4range not null,
      access_group integer generated by default as identity,
      manifest_idx integer not null,
      parent integer references sgd*.data_source,
      id bytea,
      param bytea,
      context jsonb,
  }
```

Each field is explained and justified below.  This is a good time to bikeshed on all of their names.

`vid`: Just a primary key, has no particular meaning.

`block_range`: The use of a block range is meant to allow an efficent implementation of data source removal, which is not currently a feature but quickly becomes necessary for use cars such as replacing the IPFS file associated with a user profile. Since we antecipate that the query "which data sources are live at this block" will be relevant, a range makes sense compared to separate start and end block fields.

`access_group` : Works as described [here](https://github.com/graphprotocol/graph-node/issues/3072#issuecomment-1058176460) in the 'data source indepedency' section. This will also be added to entity tables. We let the DB handle generating new access groups from a sequence when necessary, since these need not be deterministic.

`manifest_idx` : Serves the purposes that `name` currently serves, to look up the definition in the manifest. This more efficient representation stores the position in the manifest rather than the name.

`id`: An optional user given id, in case the user needs to refer to a specfic data source after creation, such as for removal.

`param`: The stored representation of the `params` argument [passed on creation](https://github.com/graphprotocol/graph-ts/blob/b7a90e10d553d45d242bf87e75c2f798be65a5c3/common/datasource.ts#L10). Examples of stored data: Address for Ethereum, account id for near, file CID for file data sources. A `bytea` representation is optimal for current uses, and we can always add a `jsonb` alternative if required in the future.

`context`: Optional data soure context, as it exists today.

`parent`: The data source which triggered the craetion of this data source. Tracking parent-child wil relationships will likely be required for features such as ipfs data sources created from other data sources. It's assumed in this design that each data source has a single parent.

Unblocks:
#3072
#3202

Ping @lutter @maoueh for review.

leoyvens:
[...]

PR #3505, opened 2022-04-27
store: alter deployment_schemas.version column to int

leoyvens:
The plan is to use this column to control migrations such as the one to a dedicated data source table described in #3405.

This changes that column to be an integer rather than an enum, so we don't need to add a DB enum variant on each new version.

### DIFF
--- a/graph/src/components/store/err.rs
+++ b/graph/src/components/store/err.rs
@@ -1,9 +1,8 @@
+use super::{BlockNumber, DeploymentHash, DeploymentSchemaVersion};
+use crate::prelude::QueryExecutionError;
 use anyhow::{anyhow, Error};
 use thiserror::Error;
 use tokio::task::JoinError;
-
-use super::{BlockNumber, DeploymentHash};
-use crate::prelude::QueryExecutionError;
 
 #[derive(Error, Debug)]
 pub enum StoreError {
@@ -51,7 +50,9 @@ pub enum StoreError {
     #[error("panic in subgraph writer: {0}")]
     WriterPanic(JoinError),
     #[error(
-        "found deployment schema version {0} which is not supported. Did you downgrade Graph Node?"
+        "found schema version {0} but this graph node only supports versions up to {}. \
+         Did you downgrade Graph Node?",
+        DeploymentSchemaVersion::LATEST
     )]
     UnsupportedDeploymentSchemaVersion(i32),
 }
--- a/graph/src/components/store/err.rs
+++ b/graph/src/components/store/err.rs
@@ -50,6 +50,10 @@ pub enum StoreError {
     Poisoned,
     #[error("panic in subgraph writer: {0}")]
     WriterPanic(JoinError),
+    #[error(
+        "found deployment schema version {0} which is not supported. Did you downgrade Graph Node?"
+    )]
+    UnsupportedDeploymentSchemaVersion(i32),
 }
 
 // Convenience to report a constraint violation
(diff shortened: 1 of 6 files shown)


store/postgres/migrations/2022-04-26-125552_alter_deployment_schemas_version/up.sql
```diff
@@ -0,0 +1,62 @@
...
+--  Some views are recreated because they depend on the altered column.
+
+-- This is left for the fdw logic which runs after migrations to recreate.
+drop schema if exists primary_public cascade;
+
+drop view if exists info.subgraph_info;
+drop view if exists info.all_sizes;
+drop materialized view if exists info.subgraph_sizes;
+
+alter table if exists deployment_schemas alter version type integer using 0;
```
lutter, line 11: I wouldn't say `if exists` here - if that table doesn't exist something is seriously wrong, and we should fail the migration.

store/postgres/src/primary.rs
```diff
@@ -114,14 +114,30 @@ table! {
...
-/// will result in an error.
-#[derive(DbEnum, Debug, Clone, Copy)]
+/// This versions changes to the storage schema of deployments.
+#[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
 pub enum DeploymentSchemaVersion {
-    Split,
-    Relational,
+    /// Baseline version, in which:
+    /// - A relational schema is used.
+    /// - Each deployment has its own namespaces for entity tables.
```
lutter, line 122: typo: 'namespace'

store/postgres/src/primary.rs
```diff
@@ -1116,7 +1131,13 @@ impl<'a> Connection<'a> {
             return Ok(site);
         }
 
-        self.create_site(shard, src.deployment.clone(), src.network.clone(), false)
+        self.create_site(
+            shard,
+            src.deployment.clone(),
+            src.network.clone(),
+            src.schema_version,
```
lutter, line 1138: This will be an important detail when we have more than one schema version: will copying create a deployment using the source's schema version or the latest version? My gut feeling is that it should always be the latest version, in which case `create_site` shouldn't take this as an argument and always use `CURRENT` for the new site.

store/postgres/src/primary.rs
```diff
@@ -114,14 +114,30 @@ table! {
     }
 }
 
-/// We used to support different layout schemes. The old 'Split' scheme
-/// which used JSONB layout has been removed, and we will only deal
-/// with relational layout. Trying to do anything with a 'Split' subgraph
-/// will result in an error.
-#[derive(DbEnum, Debug, Clone, Copy)]
+/// This versions changes to the storage schema of deployments.
```
lutter, line 117: I don't understand this sentence

leoyvens:
Thanks for the review, I've addressed the comments and renamed `CURRENT` to `LATEST`, to better reflect that we intend to support multiple schema versions.

graph/src/components/store/err.rs
```diff
@@ -50,6 +50,10 @@ pub enum StoreError {
     Poisoned,
     #[error("panic in subgraph writer: {0}")]
     WriterPanic(JoinError),
+    #[error(
+        "found deployment schema version {0} which is not supported. Did you downgrade Graph Node?"
+    )]
+    UnsupportedDeploymentSchemaVersion(i32),
```
lutter, line 56: It might be friendlier if the error message was 
```
deployment '{Qm..}'({sgd123}) uses schema version {N} but this graph-node only supports versions up to {M}. Did you downgrade graph-node?
```

lutter (approved):

graph/src/components/store/err.rs
leoyvens: I've improved this to show the max version, the deployment id will hopefully show up in the log or higher up in the error context. This required moving `DeploymentSchemaVersion` to the `graph` crate, but that was likely to be required eventually anyways.

PR #3508, opened 2022-04-27
Record timing information for failed GraphQL queries

lutter:


server/http/src/service.rs
```diff
@@ -189,18 +182,15 @@ where
...
-        let query = GraphQLRequest::new(body).compat().await;
+        let query = parse_graphql_request(&body);
 
         let result = match query {
             Ok(query) => service.graphql_runner.run_query(query, target).await,
             Err(GraphQLServerError::QueryError(e)) => QueryResult::from(e).into(),
             Err(e) => return Err(e),
         };
 
-        if let Some(id) = result.first().and_then(|res| res.deployment.clone()) {
```
evaporei, line 200: So it seems before we would only observe the metrics for known deployment ids, now if there's none the string `unknown` will be used, do we want to track metrics for those? Which scenarios would this value be `None`?

server/http/src/request.rs
```diff
@@ -4,64 +4,45 @@ use hyper::body::Bytes;
...
-                    .map(Some)
-            }
-            _ => Err(GraphQLServerError::ClientError(
-                "Invalid query variables provided".to_string(),
-            )),
-        }?;
-
-        Ok(Async::Ready(Query::new(document, variables)))
-    }
+pub fn parse_graphql_request(body: &Bytes) -> Result<Query, GraphQLServerError> {
```
evaporei, line 7: Nice refactor 👏

server/http/src/service.rs
lutter: That, and we also wouldn't record timing and counts for failed queries, something Martynas wanted to see for some investigation he was doing

evaporei (approved):

PR #3516, opened 2022-04-29
graph, graphql: Represent map keys in GraphQL query results as Box<str>

lutter:
I've been looking for ways to reduce the size of the internal representation of query results since that often takes up much more space then the resulting JSON. I suspect that a big driver of that is that `String` is very memory-inefficient for short strings - the string `id` as a `String` takes 24+2 bytes.

One way to reduce the amount of space taken by short strings, and there are a lot of them in query results, would be string interning, where we could get away with as little as 2 or 4 bytes of storage per string, depending on how we intern strings. But interning the map keys that we use in `r::Value::Object` is a bit of an undertaking: we'd want to build a string pool for the API schema, and then put a string pool derived from additional keys (field aliases) introduced by the GraphQL query 'on top' of that.

Rather than doing that, we can switch from `String` for these keys to `Box<str>` which reduces the memory overhead for each string by 8 bytes, so that the string `id` takes 16+2 bytes. The difference between the two is that `Box<str>` doesn't support mutating the string, but we don't need that for the keys in query results.

If this fairly simple change turns out to be worthwhile, it points the way to (a) that interning strings would also be worthwhile and (b) that it would be worth using `Box<str>` in more places where we deal with immutable strings (e.g., the GraphQL AST or `Entity`)

I am using a newtype for `Box<str>` in this PR and, because naming is hard, called it `Word`.

### DIFF
--- a/graph/examples/stress.rs
+++ b/graph/examples/stress.rs
@@ -3,6 +3,7 @@ use std::collections::{BTreeMap, HashMap};
 use std::iter::FromIterator;
 use std::sync::atomic::{AtomicUsize, Ordering::SeqCst};
 
+use graph::data::value::Word;
 use graph::prelude::{lazy_static, q};
 use rand::SeedableRng;
 use rand::{rngs::SmallRng, Rng};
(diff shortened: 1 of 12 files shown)


leoyvens (approved):
Nice, `Box<str>` is underused.

lutter:
> Nice, `Box<str>` is underused.

I blame the fact that there's no nice obvious name for this type. (and that `String` should have a slightly scarier name)

PR #3525, opened 2022-05-03
graph, store: Remove GRAPH_ACCOUNT_TABLES setting

lutter:
Setting tables to account-like should now be done with `graphman stats account-like`. This setting was a holdover from before this command was available.

### DIFF
--- a/store/postgres/src/relational.rs
+++ b/store/postgres/src/relational.rs
@@ -864,15 +864,7 @@ impl Layout {
         site: Arc<Site>,
     ) -> Result<Arc<Self>, StoreError> {
         let account_like = crate::catalog::account_like(conn, &self.site)?;
-        let is_account_like = {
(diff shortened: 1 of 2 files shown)


store/postgres/src/relational.rs
```diff
@@ -1215,18 +1207,13 @@ impl Table {
...
-            .contains(qualified_name.as_str());
-
         let immutable = defn.is_immutable();
 
         let table = Table {
             object: EntityType::from(defn),
             name: table_name,
             qualified_name,
-            is_account_like,
+            is_account_like: false,
```
leoyvens, line 1216: Why not check `catalog::account_like` here?

leoyvens: Got it, that seems worth a comment here.

leoyvens (approved):

PR #3531, opened 2022-05-04
Sergey/prost bump version to 0 10

soleinik-figment:
Notes: 
1. tonic-build v0.7 does not support rustfmt on generated files, hence #[rustfmt::skip]
2. tonic's Endpoint.connect_lazy() used to return Result and not anymore

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -160,6 +160,49 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "cdb031dd78e28731d87d56cc8ffef4a8f36ca26c38fe2de700543e627f8a464a"
 
 [[package]]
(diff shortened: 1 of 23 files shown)


maoueh (commented):
Seems good to me, I enable the CI to see that everything works.

evaporei (approved):

evaporei:
Actually, can't you run `cargo fmt` after doing the `build` so the generated files are formatted?

soleinik-figment:
As an additional step before pushing to git hub? If yes - then yes... it's just anyone who will work with *.proto file will have to remember to do so :-) as generated files will not be formatted. If we really want - earlier tonic-build has this function so we could try to copy it into build.rs...

PR #3532, opened 2022-05-05
Small change to memory accounting in `LfuCache` and various changes to the cache stress testing tool

lutter:
The only commit with an impact on production code is the last commit in this PR which tries to account for the fact that we cannot free the `CacheEntry` itself from the `LfuCache`

The rest of the changes expand what can be tested with the cache stress testing tool, and try to get it closer to how we actually use the `LfuCache`

### DIFF
--- a/graph/src/util/lfu_cache.rs
+++ b/graph/src/util/lfu_cache.rs
@@ -43,9 +43,12 @@ impl<K, V: Default + CacheWeight> CacheEntry<K, V> {
 }
 
 impl<K: CacheWeight, V: Default + CacheWeight> CacheEntry<K, V> {
-    /// Estimate the size of a `CacheEntry` with the given key and value
+    /// Estimate the size of a `CacheEntry` with the given key and value. Do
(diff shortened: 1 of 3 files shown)


leoyvens (approved):

graph/src/util/lfu_cache.rs
```diff
@@ -181,6 +184,13 @@ impl<K: Clone + Ord + Eq + Hash + Debug + CacheWeight, V: CacheWeight + Default>
         max_weight: usize,
         stale_period: u64,
     ) -> Option<(usize, usize, usize)> {
+        use crate::prelude::lazy_static;
+        lazy_static! {
+            // Setting `DEAD_WEIGHT` is dangerous since it can lead to a
+            // situation where an empty cache is bigger than the max_weight
+            // which leads toa panic
```
leoyvens, line 191:
```suggestion
            // which leads to a panic
```

lutter: aah .. damn tyops. Fixed

PR #3538, opened 2022-05-06
Run `ANALYZE` after grafting

### LINKED ISSUE #3153: Fix grafting to analyze tables at the end of copying

tilacog:
Fixes #3153.

### DIFF
--- a/node/src/bin/manager.rs
+++ b/node/src/bin/manager.rs
@@ -886,7 +886,6 @@ async fn main() {
                     let (store, primary_pool) = ctx.store_and_primary();
                     let subgraph_store = store.subgraph_store();
                     commands::stats::analyze(subgraph_store, primary_pool, deployment, &entity)
-                        .await
                 }
             }
         }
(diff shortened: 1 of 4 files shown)


tilacog:
I'm not sure if transforming the `analyze` async function to sync is desirable, but I assumed that would be easier than converting the upstream call hierarchy to async. 

I'll be happy to implement it the other way around if it is preferable.

lutter:
> I'm not sure if transforming the `analyze` async function to sync is desirable, but I assumed that would be easier than converting the upstream call hierarchy to async.

I think removing the `async` is fine - the old use was in the CLI, where it doesn't really matter, and the new use is in a blocking context already, so not having `async` shouldn't be a problem.

It would probably be better to use the existing `conn` though in `start_subgraph` rather than getting a new one for each `analyze` command; that would require splitting the current `analyze` into an `analyze` and `analyze_with_conn` where `analyze` gets a connection and calls `analyze_with_conn`, and `analyze_with_conn` gets a `conn: &PgConnection` as an additional argument.

lutter (approved):
Just split into `analyze` and `analyze_with_conn` and I then this is good to merge

PR #3539, opened 2022-05-06
docker: Install cmake which is required by prost

lutter:
Requirement was added in this PR:
https://github.com/graphprotocol/graph-node/pull/3531

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -14,6 +14,7 @@ ARG TAG_NAME=unknown
 ADD . /graph-node
 
 RUN cd /graph-node \
+    && apt-get update && apt-get install -y cmake \
     && rustup component add rustfmt \
     && RUSTFLAGS="-g" cargo install --locked --path node \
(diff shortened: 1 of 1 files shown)


tilacog (approved):

docker/Dockerfile
```diff
@@ -14,6 +14,7 @@ ARG TAG_NAME=unknown
 ADD . /graph-node
 
 RUN cd /graph-node \
+    && apt-get update && apt-get install -y cmake \
```
evaporei, line 17: Can we remove this [line's](https://github.com/graphprotocol/graph-node/blob/28803bf48a9ee8df29342445360d87699a23f9e3/docker/Dockerfile#L77) `update`?

evaporei (approved):

docker/Dockerfile
lutter: No, they happen in two unrelated containers

PR #3540, opened 2022-05-06
Modify error string to remove Ethereum

lavabyrd:
With the addition of additional protocols, deploying a subgraph with an invalid upstream provider would return an error String that incorrectly states Ethereum network not supported. This just removes the Ethereum part of the string to allow it to make more sense

error_string


lutter (approved):
Nice! Thanks for doing that

PR #3543, opened 2022-05-09
Rename Tendermint to Cosmos

pienkowb:
Apart from the renaming, this PR covers:

- Updated protobuf structure
- Support for Firehose filtering
- Transaction handler
- Fixing the field type mismatch for enums


chain/near/src/runtime/generated.rs
```diff
@@ -207,7 +207,7 @@ impl AscIndexId for AscActionEnum {
 #[repr(C)]
 #[derive(AscType)]
 pub(crate) struct AscPublicKey {
-    pub kind: u32,
+    pub kind: i32,
```
evaporei, line 210: Why these changed for `near`? 🤔

pienkowb: Here's the related discussion: https://figment-work.slack.com/archives/C02RRP8H8LU/p1651842790074239

evaporei: Umnn, can you paste a part of it here? I don't have access to Figment's Slack 😓

pienkowb: Sure, it's a conversation I started on the `#wg-subgraph-api` channel on May 6th:
> I've noticed that in the Asc bindings for some enums an `i32` field is mapped into a `u32` field (e.g. [here](https://github.com/graphprotocol/graph-node/blob/master/chain/near/src/protobuf/sf.near.codec.v1.rs#L276-L277=) and [here](https://github.com/graphprotocol/graph-node/blob/master/chain/near/src/runtime/generated.rs#L221=)). We did the same in the Cosmos integration and I'm wondering if it shouldn't be `i32` in both places. Is this a concern at all?

evaporei (approved):

evaporei:
@pienkowb it seems you'll need to solve conflicts with the `main` branch.

PR #3557, opened 2022-05-11
Reduce memory needs of `graph::data::Object`

lutter:
We use a lot of these when generating  query responses; reducing their size even by just a pointer should therefore reduce the memory needs of query processing


leoyvens (approved):
_fat pointers_

PR #3561, opened 2022-05-16
Fix a problem with mixing mutable and immutable entities in interfaces

lutter:


store/postgres/src/relational_queries.rs
```diff
@@ -2434,7 +2444,15 @@ impl<'a> SortKey<'a> {
...
-        let br_column = if ENV_VARS.store.order_by_block_range {
+        // If there is more than one table, we are querying an interface,
+        // and the order is on an attribute in that interface so that all
+        // tables have a column for that. It is therefore enough to just
+        // look at the first table to get the name
+        let table = collection
+            .first_table()
+            .expect("an entity query always contains at least one entity type/table");
+
+        let br_column = if collection.all_mutable() && ENV_VARS.store.order_by_block_range {
```
evaporei, line 2455: This is a bit of a stretch, but I was wondering, would it make sense to divide the queries in two? One for the immutable tables and the other for mutable ones?

evaporei (approved):

store/postgres/src/relational_queries.rs
lutter: We'd have to look at query plans and see if performance is better if we split into two; for now, I think this is fine, queries that mix mutable and immutable entities should be relatively rare. Once some pop up a problematic, we should revisit.

Issue #3566, opened 2022-05-17
Derived entities with interface and Bytes as ID lead to error in gql query 


0xSkly:
Given a derived field with an interface
```gql
 type Pool {
  interfactions: [PoolInteraction!]! @derivedFrom(field: 'pool)
 ...
}

interface PoolInteraction {
 id: Bytes!
 ....
}

type Swap implements PoolInteraction @entity {
 id: Bytes!
 ...
}
```

When running a gql query against that with
```gql
query {
  pools {
    interactions {
      ... on Swap {
        tokenIn {
          name
        }
        tokenOut {
          name
        }
      }   
    }
  }
}
```

results in this error:
`Failed to get entities from store: operator does not exist: text = bytea`

If i run the top level query on the interface: 
```gql
query {
  poolInteractions {
    ... on Swap {
      tokenIn {
        name
      }
      tokenOut {
        name
      }
    }
  }
}
```
this works.

See https://api.thegraph.com/subgraphs/name/0xskly/beethovenx-v2 (might be broken at times, cause experimenting on it)

PR #3567, opened 2022-05-17
store: Handle related interface attributes with Bytes id correctly

lutter:
Fixes https://github.com/graphprotocol/graph-node/issues/3566


store/postgres/src/relational_queries.rs
```diff
@@ -91,6 +91,31 @@ fn bytes_as_str(id: &str) -> String {
     id.trim_start_matches("\\x").to_owned()
 }
 
+impl IdType {
+    /// Add `ids`  as a bind variable to `out`, using the right SQL type
```
leoyvens, line 95:
```suggestion
    /// Add `ids` as a bind variable to `out`, using the right SQL type
```
```suggestion
    /// Add `ids`  as a bind variable to `out`, using the right SQL type
```

leoyvens (approved):

PR #3577, opened 2022-05-22
Add ipfs.getBlock

Theodus:
This adds an additional host export for fetching IPLD blocks from IPFS, similar to IPFS cat.


runtime/wasm/src/module/mod.rs
```diff
@@ -472,6 +472,12 @@ impl<C: Blockchain> WasmInstance<C> {
         //
         // For reference, search this codebase for: ff652476-e6ad-40e4-85b8-e815d6c6e5e2
         link!("ipfs.cat", ipfs_cat, "host_export_ipfs_cat", hash_ptr);
+        link!(
+            "ipfs.getBlock",
+            ipfs_get_block,
+            "host_export_ipfs_get_block",
+            hash_ptr
+        );
```
leoyvens, line 480: You can gate the function linking by `allow_non_deterministic_ipfs`.

leoyvens (changes requested):

core/src/link_resolver.rs
```diff
@@ -217,6 +217,58 @@ impl LinkResolverTrait for LinkResolver {
...
+        .await?;
+
+        let max_cache_file_size = self.env_vars.mappings.max_ipfs_cache_file_size;
+        let max_file_size = self.env_vars.mappings.max_ipfs_file_bytes.map(|n| n as u64);
+        restrict_file_size(&link.link, &stat, &max_file_size)?;
+
+        let link = link.link.clone();
+        let this = self.clone();
+        let logger = logger.clone();
+        let data = retry_policy(self.retry, "ipfs.cat", &logger)
```
leoyvens, line 243:
```suggestion
        let data = retry_policy(self.retry, "block.get", &logger)
```

core/src/link_resolver.rs
```diff
@@ -217,6 +217,58 @@ impl LinkResolverTrait for LinkResolver {
         Ok(data)
     }
 
+    async fn get_block(&self, logger: &Logger, link: &Link) -> Result<Vec<u8>, Error> {
+        if let Some(data) = self.cache.lock().unwrap().get(&link.link) {
```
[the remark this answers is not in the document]
Theodus: Ah, good catch. The CIDs would refer to the same data, but they are interpreted by the IPFS node differently. So we can't use the same cache. I think it's fine for our use case to avoid caching these results, at least for now.

runtime/wasm/src/module/mod.rs
Theodus: Done

core/src/link_resolver.rs
Theodus: Fixed, using `"ipfs.getBlock"`

Theodus:
It seems that the rebase can be done automatically

leoyvens:
@Theodus But it might fail to compile, since the PR switched from `object stat` to `dag stat` which is now recommended, so there is direct interaction with this code.

Theodus:
Rebased

Theodus:
It looks like the switch from object stat to dag stat has broken the support for fetching IPLD blocks. This is because the IPLD blocks we use aren't IPFS dag nodes.

Theodus:
CI failure seems unrelated: `thread 'graft' panicked at 'deleting test entities succeeds: Unknown(deadlock detected)', store/postgres/tests/graft.rs:266:10`

leoyvens:
Yes that's spurious, rerunning tests. Wild that the blocks are not recognized as dags, do they use a non-standard codec?

leoyvens (approved):

Theodus:
IPFS files & dag objects use an IPLD block format called UnixFS. IPFS also provides storage for IPLD blocks of any supported format via the block API. UnixFS happens to use the dag-pb codec, but more importantly it has expectations for the contents of the data and link sections of those dag-pb blocks. Those expectations won't hold for non-UnixFS blocks.

PR #3591, opened 2022-05-25
elasticsearch: Use connection pool

leoyvens:
This is less overhead than creating a new client for each request. Particularly we were seeing this generate excessive DNS requests.

### DIFF
--- a/graph/src/log/elastic.rs
+++ b/graph/src/log/elastic.rs
@@ -27,6 +27,8 @@ pub struct ElasticLoggingConfig {
     pub username: Option<String>,
     /// The Elasticsearch password (optional).
     pub password: Option<String>,
+    /// A client to serve as a connection pool to the endpoint.
+    pub client: Client,
 }
 
 /// Serializes an slog log level using a serde Serializer.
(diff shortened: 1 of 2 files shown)


graph/src/log/elastic.rs
```diff
@@ -265,14 +267,16 @@ impl ElasticDrain {
                 batch_url.set_path("_bulk");
 
                 // Send batch of logs to Elasticsearch
-                let client = Client::new();
-
                 let header = match config.general.username {
-                    Some(username) => client
+                    Some(username) => config
+                        .general
+                        .client
```
lutter, line 273: Shouldn't we use a `ClientBuilder` and bake the headers into the client instead of setting them with every request? I actually don't think it matters all that much. I had a brief look at the `reqwest` sources, and couldn't see how this change would avoid DNS requests.

lutter (approved):
LGTM

graph/src/log/elastic.rs
leoyvens: I think things are easier to follow if the headers are set where the request is made. On avoiding a DNS request, I'd assume that a new connection would generate a DNS request while reusing an existing one wouldn't. But not an expert here.

PR #3594, opened 2022-05-25
Improve memory efficiency of `EntityCache`

lutter:
This PR makes the `EntityCache` more memory efficient, meaning the same amount of memory will be able to store more entities with two changes:

* Use `Word` (`Box<str>`) for entity types and ids
* Remove the subgraph id from `EntityKey` since that adds an overhead of 70 bytes _for every cache entry_

### DIFF
--- a/chain/substreams/src/trigger.rs
+++ b/chain/substreams/src/trigger.rs
@@ -4,7 +4,7 @@ use anyhow::Error;
 use graph::{
     blockchain::{self, block_stream::BlockWithTriggers, BlockPtr},
     components::{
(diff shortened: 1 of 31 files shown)


leoyvens (approved):
Code lgtm, good thing we didn't have `EntityKey` in the actual `ProofOfIndexingEvent`. The change is very mechanical but it touches many critical pieces of the code, so might be worth an integration cluster run.

lutter:
I ran this in the integration cluster for > 24hrs without any incident. Merging it therefore

PR #3570, opened 2022-05-18
ipfs: Switch from object stat to dag stat

leoyvens:
It turns out the `object` API has been deprecated for a while. Resolves #3565.

PR #3597, opened 2022-05-27
ipfs: Replace dag/stat with files/stat

leoyvens:
This fixes a couple issues in #3570.

`dag/stat` turns out to be unreliable, even with the `progress=false` argument to get a consistent response format, the IPFS node seemed to struggle with it. So we use `files/stat` which is less expensive as it just reads the file metadata, and is probably closer to the deprecated `object/stat`.

The IPFS version in docker and CI is downgraded to 0.10.0, to workaround a startup issue affecting Mac M1s.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,7 @@ jobs:
     runs-on: ubuntu-latest
     services:
       ipfs:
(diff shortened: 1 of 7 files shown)


Theodus (approved):
LGTM

Issue #3609, opened 2022-06-02
Arweave "owner" transaction filtering currently uses the account's public key, rather than the easier to find short form


azf20:
The public key format is the long form which is to well known in Arweave space, which is used more to the short form displayed by the block explorer for example. Here a current long form version https://github.com/ChainSafe/subgraph-arweave-example/blob/master/subgraph.yaml#L11 and short form https://viewblock.io/arweave/address/AOvWwnkMZTItUjTEy4MNB5RJatrMf52f2RNEX8zwq5w.

evaporei:
https://github.com/graphprotocol/graph-node/pull/3613

PR #3613, opened 2022-06-02
Filipe/arweave filter sha256

mangas:
Fixes https://github.com/graphprotocol/graph-node/issues/3609

Changes:
1. Transaction filter now uses both public keys and sha256 of those public keys
2. Filtering will try the long form (as is used right now for blocks from firehose) and then try the sha256
3. When parsing the data sources, we'll calculate the "short version" of the pubkey and store it alongside any sources that defined those keys. 

Arweave uses base64url(sha256(pubkey)) in their explorer and that is much easier for users to retrieve than the pubkey, so the goal of this PR is to enable filtering by either. Currently firehose only supports pubkey format on the protobuf. With these changes we can just switch from checking the owner to checking sha256 as soon as it is available and possibly remove the pubkey matching completing since we can calculate the checksum version when creating the filters.

### DIFF
--- a/chain/arweave/Cargo.toml
+++ b/chain/arweave/Cargo.toml
@@ -12,6 +12,7 @@ graph = { path = "../../graph" }
 prost = "0.10.1"
 prost-types = "0.10.1"
 serde = "1.0"
+sha2 = "0.10.2"
 
 graph-runtime-wasm = { path = "../../runtime/wasm" }
 graph-runtime-derive = { path = "../../runtime/derive" }
--- a/chain/arweave/src/adapter.rs
+++ b/chain/arweave/src/adapter.rs
@@ -2,9 +2,13 @@ use crate::capabilities::NodeCapabilities;
 use crate::{data_source::DataSource, Chain};
 use graph::blockchain as bc;
 use graph::prelude::*;
+use sha2::{Digest, Sha256};
 use std::collections::HashSet;
 
 const MATCH_ALL_WILDCARD: &str = "";
+// Size of sha256(pubkey)
+const SHA256_LEN: usize = 32;
+
 #[derive(Clone, Debug, Default)]
 pub struct TriggerFilter {
     pub(crate) block_filter: ArweaveBlockFilter,
@@ -41,13 +45,22 @@ impl bc::TriggerFilter<Chain> for TriggerFilter {
 /// see docs: https://thegraph.com/docs/en/supported-networks/arweave/
 #[derive(Clone, Debug, Default)]
 pub(crate) struct ArweaveTransactionFilter {
-    owners: HashSet<Vec<u8>>,
+    owners_pubkey: HashSet<Vec<u8>>,
+    owners_sha: HashSet<Vec<u8>>,
     match_all: bool,
 }
 
 impl ArweaveTransactionFilter {
     pub fn matches(&self, owner: &[u8]) -> bool {
-        self.match_all || self.owners.contains(owner)
+        if self.match_all {
+            return true;
+        }
+
+        if owner.len() == SHA256_LEN {
+            return self.owners_sha.contains(owner);
+        }
+
+        self.owners_pubkey.contains(owner) || self.owners_sha.contains(&sha256(owner))
     }
 
     pub fn from_data_sources<'a>(iter: impl IntoIterator<Item = &'a DataSource>) -> Self {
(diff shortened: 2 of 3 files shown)


chain/arweave/src/adapter.rs
```diff
@@ -90,3 +114,94 @@ impl ArweaveBlockFilter {
...
+    const ARWEAVE_PUBKEY_EXAMPLE: &str = "x-62w7g2yKACOgP_d04bhG8IX-AWgPrxHl2JgZBDdNLfAsidiiAaoIZPeM8K5gGvl7-8QVk79YV4OC878Ey0gXi7Atj5BouRyXnFMjJcPVXVyBoYCBuG7rJDDmh4_Ilon6vVOuHVIZ47Vb0tcgsxgxdvVFC2mn9N_SBl23pbeICNJZYOH57kf36gicuV_IwYSdqlQ0HQ_psjmg8EFqO7xzvAMP5HKW3rqTrYZxbCew2FkM734ysWckT39TpDBPx3HrFOl6obUdQWkHNOeKyzcsKFDywNgVWZOb89CYU7JFYlwX20io39ZZv0UJUOEFNjtVHkT_s0_A2O9PltsrZLLlQXZUuYASdbAPD2g_qXfhmPBZ0SXPWCDY-UVwVN1ncwYmk1F_i35IA8kAKsajaltD2wWDQn9g5mgJAWWn2xhLqkbwGbdwQMRD0-0eeuy1uzCooJQCC_bPJksoqkYwB9SGOjkayf4r4oZ2QDY4FicCsswz4Od_gud30ZWyHjWgqGzSFYFzawDBS1Gr_nu_q5otFrv20ZGTxYqGsLHWq4VHs6KjsQvzgBjfyb0etqHQEPJJmbQmY3LSogR4bxdReUHhj2EK9xIB-RKzDvDdL7fT5K0V9MjbnC2uktA0VjLlvwJ64_RhbQhxdp_zR39r-zyCXT-brPEYW1-V7Ey9K3XUE";
+
+    #[test]
+    fn transaction_filter_some_empty_matches_all() {
+        let dss = vec![
+            new_datasource(None, 10),
+            new_datasource(Some("".into()), 10),
+            new_datasource(Some(ARWEAVE_PUBKEY_EXAMPLE.into()), 10),
+            new_datasource(
+                Some("2cf24dba5fb0a30e26e83b2ac5b9e29e1b161e5c1fa7425e73043362938b9824".into()),
```
maoueh, line 138: I expect developers to rather use their the short form version of the address defined like `zrmq3cxYSHSlSQWSPferKnppq_-kTKW6uGf3w5XPUcw` (see https://viewblock.io/arweave/address/zrmq3cxYSHSlSQWSPferKnppq_-kTKW6uGf3w5XPUcw) and not a hex version of it.

Where this hex requirements came from?

chain/arweave/src/adapter.rs
```diff
@@ -56,17 +62,35 @@ impl ArweaveTransactionFilter {
                     && !data_source.mapping.transaction_handlers.is_empty()
             })
             .map(|ds| {
-                base64_url::decode(&ds.source.owner.clone().unwrap_or_default()).unwrap_or_default()
+                let owner = ds.source.owner.clone().unwrap_or_default();
```
maoueh, line 65: `clone` could be made only `owner.len() == HEX_SHA256_LEN`, the other branch of the if does not need an owning copy.

chain/arweave/src/adapter.rs
```diff
@@ -40,12 +44,14 @@ impl bc::TriggerFilter<Chain> for TriggerFilter {
...
-    owners: HashSet<Vec<u8>>,
+    owners_pubkey: HashSet<Vec<u8>>,
+    owners_hex: HashSet<Vec<u8>>,
+    match_all: bool,
 }
 
 impl ArweaveTransactionFilter {
     pub fn matches(&self, owner: &[u8]) -> bool {
-        self.owners.contains(owner)
+        self.match_all || self.owners_pubkey.contains(owner) || self.owners_hex.contains(owner)
```
maoueh, line 54: `owner` is always in long form here, I don't see how it can actually work.

chain/arweave/src/adapter.rs
```diff
@@ -56,17 +62,35 @@ impl ArweaveTransactionFilter {
...
+
+        let owners_hex: Vec<Vec<u8>> = hex
+            .into_iter()
+            .chain(long.iter().map(|bs| digest_bytes(bs).into_bytes()))
+            .collect();
+
+        let owners_pubkey = HashSet::from_iter(long);
         Self {
-            owners: HashSet::from_iter(owners),
+            match_all: owners_pubkey.contains(&vec![]),
```
maoueh, line 86: This only contains long form version no? This means it would match all if short form is provided.

chain/arweave/src/adapter.rs
```diff
@@ -56,17 +62,35 @@ impl ArweaveTransactionFilter {
                     && !data_source.mapping.transaction_handlers.is_empty()
             })
             .map(|ds| {
-                base64_url::decode(&ds.source.owner.clone().unwrap_or_default()).unwrap_or_default()
+                let owner = ds.source.owner.clone().unwrap_or_default();
+
+                if owner.len() == HEX_SHA256_LEN {
+                    owner.into_bytes()
+                } else {
+                    base64_url::decode(&owner).unwrap_or_default()
```
maoueh, line 70: Are we not checking because the input has already be validated?

maoueh (changes requested):

chain/arweave/src/adapter.rs
mangas: The hex here is hex(sha256(pubkey)) , I was under the impression that this is how we go from the long version to the short version

chain/arweave/Cargo.toml
```diff
@@ -22,3 +23,4 @@ graph-core = { path = "../../core" }
 graph-store-postgres = { path = "../../store/postgres" }
 pretty_assertions = "0.7.2"
 test-store = { path = "../../store/test-store" }
+hex = "0.4.3"
```
evaporei, line 26: I think you should probably use `graph::prelude::hex` instead of adding this dependency twice.

chain/arweave/src/adapter.rs
mangas: This has been fixed, the correct format is base64url(sha256(pubkey))

chain/arweave/src/adapter.rs
mangas: this was fixed in a differnet PR

chain/arweave/src/adapter.rs
mangas: for the time being it will just match on `owners_pubkey` but this means we can switch as soon as we have the short format coming from firehose. Performance is prolly not a concern here yet

chain/arweave/src/adapter.rs
mangas: This was refactored in the meantime

chain/arweave/src/adapter.rs
mangas: actually this was already there but it's prolly a good idea to validate it

chain/arweave/Cargo.toml
mangas: well spotted, it's not actually needed anymore :) thanks

chain/arweave/src/adapter.rs
mangas: fixed

mangas: added some more fixes and some tests as well

evaporei (approved):

maoueh (approved):

Issue #3600, opened 2022-05-30
Issue of using Bytes as IDs


haowang1013:
**Do you want to request a *feature* or report a *bug*?**

- bug

**What is the current behavior?**

- Using Bytes as entity IDs seems to cause failure when resolving references.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem.**

- I created a test repo demonstrating the problem [here](https://github.com/haowang1013/graph-test)
- The graphs uses [openzeppelin-subgraphs](https://github.com/OpenZeppelin/openzeppelin-subgraphs) and it's now indexing the BYAC contract on mainnet
- Using the following 2 entities as [example](https://github.com/haowang1013/graph-test/blob/main/generated/byac.schema.graphql):
```
type ERC721Contract @entity(immutable: true) {
	id: Bytes!
	asAccount: Account!
	supportsMetadata: Boolean
	name: String
	symbol: String
	tokens: [ERC721Token!]! @derivedFrom(field: "contract")
	operators: [ERC721Operator!]! @derivedFrom(field: "contract")
	transfers: [ERC721Transfer!]! @derivedFrom(field: "contract")
}

type ERC721Transfer implements Event @entity(immutable: true) {
	id: ID!
	emitter: Account!
	transaction: Transaction!
	timestamp: BigInt!
	contract: ERC721Contract!
	token: ERC721Token!
	from: Account!
	to: Account!
}
```

So the `ERC721Contract` has a list of transfers of type `ERC721Transfer` using the derived field of `contract`

When using the query below [here](https://api.thegraph.com/subgraphs/id/QmZcEF5HAGU1maRMSQTGgAUxxrugiw9taSY7qfwe84HesE)
```
{
  erc721Contracts {
    id
    transfers(first: 5) {
      contract {
        id
      }
    }
  }
}
```

It couldn't resolve the `contract` field of the `ERC721Transfer` with an error of 
```
Null value resolved for non-null field `contract`
```

I know that the `contract` field must be set for the `ERC721Transfer`, otherwise they won't show up as `transfers` of `ERC721Contract`, but for some reason the reference cannot be resolved directly in the query.

I also did an experiment to change `ERC721Contract` to use `ID!` as the id and that seems to fixed it, so this is only a problem of using `Bytes!` as the id.

**What is the expected behavior?**

- Reference of entities using `Bytes` as id should be correctly resolved.

lutter:
This is now live on the hosted service and the query in this issue works

PR #3617, opened 2022-06-04
graph, graphql, store: Use the correct parent type for type d queries

lutter:
When we stitch the result of a query of type d back into the parents, we need to make sure that the parent ids we pass through the query use the type of the parent's id field. Previously, the code was using the type of the child's id field but that has no relation to the correct type.

Fixes https://github.com/graphprotocol/graph-node/issues/3600

### DIFF
--- a/graphql/tests/query.rs
+++ b/graphql/tests/query.rs
@@ -161,6 +161,7 @@ fn test_schema(id: DeploymentHash, id_type: IdType) -> Schema {
         id: @ID@!
         title: String!
         writtenBy: Musician!
+        publisher: Publisher!
         band: Band @derivedFrom(field: \"originalSongs\")
     }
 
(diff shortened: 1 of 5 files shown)


graphql/tests/query.rs
```diff
@@ -586,6 +592,32 @@ fn query_variables_are_used() {
     );
 }
 
+#[test]
+fn bytes_child() {
```
leoyvens, line 596: Do we already have a test for when the parent is `Bytes` and the child is `ID`?

leoyvens (approved):

graphql/tests/query.rs
lutter: Yes, for example `can_query_one_to_many_relationships_in_both_directions` does that with `{writtenSongs .. { writtenBy { id } }` when songs use a `Bytes` id

lutter: That's not quite what we want (`writtenBy` is derived) Just added a test that checks that combination

Issue #3501, opened 2022-04-26
Grafting does not respect deterministic failed status


leoyvens:
To reproduce:
1. Have a subgraph fail with a deterministic error at block N.
2. Graft it at block N.
3. The grafted subgraph will continue syncing from N + 1, when it should be failed at block N.

evaporei:
The `graft` feature always starts from the `base` subgraph at block N + 1. Upon testing the described scenario on `graph-node` version `0.24.2`, I bumped into this error message:

```
✖ Failed to deploy to Graph node http://127.0.0.1:8020/: subgraph validation error: [the graft base is invalid: failed to graft onto `QmWt3quQnYFhwypj93Uco7ex3UmtdeRrLAhYY5cbQL2Vn6` at block 11416275 since it has only processed block 11416241]
```

Which makes sense, because it was only on `0.25.X` that `graph-node` started processing deterministic failures (saving the deployment pointer).

My proposal is to validate at creation time, just like the error above, that subgraphs can't be created at the failed block, because `graft`s start at N + 1.

If a developer wishes to reproduce the failure in a graft, they should define in their manifest block N - 1 (N being where the base failed). We can give a nice error message in the validation suggested above 🙂 

What do you think @leoyvens ?

This is where the failure happens for the curious ones: https://github.com/graphprotocol/graph-node/blob/master/graph/src/data/subgraph/mod.rs#L478

PR #3627, opened 2022-06-06
Add validation for grafts created at failed block

evaporei:
Fixes https://github.com/graphprotocol/graph-node/issues/3501

### DIFF
--- a/graph/src/data/subgraph/mod.rs
+++ b/graph/src/data/subgraph/mod.rs
@@ -458,16 +458,22 @@ impl Graft {
     ) -> Result<(), SubgraphManifestValidationError> {
         use SubgraphManifestValidationError::*;
 
+        let last_processed_block = store
+            .least_block_ptr(&self.base)
+            .await
+            .map_err(|e| GraftBaseInvalid(e.to_string()))?;
+        let is_base_healthy = store
+            .is_healthy(&self.base)
+            .await
+            .map_err(|e| GraftBaseInvalid(e.to_string()))?;
+
         // We are being defensive here: we don't know which specific
         // instance of a subgraph we will use as the base for the graft,
         // since the notion of which of these instances is active can change
         // between this check and when the graft actually happens when the
         // subgraph is started. We therefore check that any instance of the
         // base subgraph is suitable.
-        match (
-            store.least_block_ptr(&self.base).await.map_err(|e| GraftBaseInvalid(e.to_string()))?,
-            store.is_healthy(&self.base).await.map_err(|e| GraftBaseInvalid(e.to_string()))?,
-        ) {
+        match (last_processed_block, is_base_healthy) {
             (None, _) => Err(GraftBaseInvalid(format!(
                 "failed to graft onto `{}` since it has not processed any blocks",
                 self.base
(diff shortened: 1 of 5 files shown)


evaporei:
It seems both the `integration` and `unit` tests are failing for this, I'll investigate why is that 🔍

graph/src/data/subgraph/mod.rs
```diff
@@ -451,37 +451,53 @@ pub struct Graft {
...
+    ( $( $x:expr ),* ) => {
+        {
+            vec![
+                $(
+                    SubgraphManifestValidationError::GraftBaseInvalid($x),
+                )*
+            ]
+        }
+    };
+}
```
leoyvens, line 464: I wouldn't use a macro for this.

graph/src/data/subgraph/mod.rs
```diff
@@ -451,37 +451,53 @@ pub struct Graft {
...
         // between this check and when the graft actually happens when the
         // subgraph is started. We therefore check that any instance of the
         // base subgraph is suitable.
-        match store.least_block_ptr(&self.base).await {
-            Err(e) => gbi(e.to_string()),
-            Ok(None) => gbi(format!(
+        match (
+            store.least_block_ptr(&self.base).await,
+            store.is_healthy(&self.base).await,
+        ) {
```
leoyvens, line 480: If I understand correctly, isn't this disallowing grafts for any failed subgraph?

leoyvens (changes requested):

graph/src/data/subgraph/mod.rs
evaporei: Why not? Do you want to avoid people just using this enum variant by default?

Do you think [this](https://github.com/graphprotocol/graph-node/pull/3627/commits/089302835981ae1db4b8999123f4d9c5a362192c) looks better?

graph/src/data/subgraph/mod.rs
evaporei: No, take a look at the lines below:

```rust
// Base didn't process any block: graft should fail.
(Ok(None), _) => /* ... */,
// Base is healthy, however graft must be prior to processed block.
(Ok(Some(ptr)), Ok(true)) if ptr.number < self.block => /* ... */,
// Base failed, it can be grafted, but only for blocks prior to the failure.
(Ok(Some(ptr)), Ok(false)) if !(self.block < ptr.number) => /* ... */,
```

evaporei: Also, there's a test I made to so we can be sure failed subgraphs can be grafted prior to the failed block 🙂 
https://github.com/graphprotocol/graph-node/pull/3627/files#diff-8c7c963bca4ee64c86ba1963ed4c5f86dc40f89ccd3e890e4c508ba262e413f3R191

graph/src/data/subgraph/mod.rs
leoyvens: You can have just a function that takes a vec of `String`. A macro is kind of a big hammer.

leoyvens: But I can see that you want the single parameter case to not need a wrapping vec, in that aspect the macro is nice.

evaporei: This addresses the issue, but I think it can become cleaner https://github.com/graphprotocol/graph-node/pull/3627/commits/87839c489a1c1ed88bd2ea87138d72eaf8662849

Maybe it's fine for now.

evaporei: I've cleaned it more according to what we've spoke on Slack 🙂

leoyvens (approved):
Good to squash and merge.

Issue #3648, opened 2022-06-09
Graph Node start sync from block 0


vietanhduong:
**Do you want to request a *feature* or report a *bug*?**
* Feature

**What is the current behavior?**
Currently, `graph-node` start sync from block 0.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem.**

**What is the expected behavior?**
We can tell the graph node what the genesis block is. Because, there are some chains, the genesis block start from 1.

```console
Jun 09 11:21:31.137 WARN Trying again after eth_getBlockByNumber(0, false) RPC call failed (attempt #10) with result Err(Ethereum node could not find genesis block), provider: velas-rpc-0
```

khelaia:
add startBlock:
in your subgraph.yaml

leoyvens:
Closed by #3650

PR #3650, opened 2022-06-10
chore: specify genesis block via env

vietanhduong:
Currently, the genesis block fixed with `0`, but there are some chains, the genesis block start from `1`. 

This PR support specify the genesis block by `GRAPH_ETHEREUM_GENESIS_BLOCK_NUMBER`, default value is `0`.

Related Issue: https://github.com/graphprotocol/graph-node/issues/3648

### DIFF
--- a/chain/ethereum/src/env.rs
+++ b/chain/ethereum/src/env.rs
@@ -83,6 +83,12 @@ pub struct EnvVars {
     /// `GRAPH_ETHEREUM_TARGET_TRIGGERS_PER_BLOCK_RANGE`. The default value is
     /// 100.
     pub target_triggers_per_block_range: u64,
+    /// These are some chains, the genesis block is start from 1 not 0. If this
+    /// flag is not set, the default value will be 0.
+    ///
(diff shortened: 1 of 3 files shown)


vietanhduong:
Refer issue: https://github.com/graphprotocol/graph-node/issues/3648

leoyvens:
Hi, @vietanhduong, thank you for the patch, have you verified that this patch allows graph node to successfully sync subgraphs on those chains?

vietanhduong:
I have just tested with `velas` chain, anh it's works. The unit test failed seem lated to these codes on main branch. 

I have run tests on my branch, and everything is fine.

leoyvens (approved):

lutter:
I fear a proper fix for this will require quite a bit more work: the genesis block should really be specified in `graph-node.toml` for each chain, and needs to also be persisted in the database so it's accessible in other places; I assume there's also quite a few places where we implicitly assume that the number of the genesis block is 0 which we'd need to find and update.

vietanhduong:
@lutter I understand what you consider, but for almost case, the genesis block will start with 0. If there is some problem with the genesis block is start with `1`, I think a PR to fix these problems are necessary.

PR #3651, opened 2022-06-10
chain/ethereum: Remove call.input.len() validation

evaporei:
We don't have to check whether the call.input.len() is multiple of 32 because we now skip invalid decoding of calls.

- https://github.com/graphprotocol/graph-node/pull/3326
- ae6951328df4270e3153897220fdef1822055859

This is essentially reverting https://github.com/graphprotocol/graph-node/pull/3219 because of the new change introduced in https://github.com/graphprotocol/graph-node/pull/3326.

Related `ethabi` patch: https://github.com/graphprotocol/ethabi/commit/441c09f1e688f8c63c80dbe7ecc36fd64c745a9e

### DIFF
--- a/chain/ethereum/src/adapter.rs
+++ b/chain/ethereum/src/adapter.rs
@@ -445,22 +445,11 @@ impl EthereumCallFilter {
             return false;
         }
 
-        // Make sure the call input size is multiple of 32, otherwise we can't decode it.
-        // This is due to the Ethereum ABI spec: https://docs.soliditylang.org/en/v0.8.11/abi-spec.html
+        // The `call.input.len()` is validated in the
+        // DataSource::match_and_decode function.
+        // Those calls are logged as warning and skipped.
         //
-        // A scenario where the input could have the wrong size is when a `delegatecall` is
-        // "disguised" as a `call`. For example for this couple of transactions/calls:
-        // 1. https://etherscan.io/tx/0x8e992eeb40e18703dd8169a8031e2113311985f1c20e0723a2dc362df40bafb0
-        // 2. https://etherscan.io/tx/0x7c391bcfa2007f84e123ad47ea72e2d6ffb2fbab81deff0990b0c499e0664a92
-        //
-        // The first one is acting as a proxy to the second one (an atomicMatch_).
-        // If you try to decode the first call as it is/comes from a `traces` RPC request, it
-        // will fail because it has a smaller size/length.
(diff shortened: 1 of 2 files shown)


chain/ethereum/src/adapter.rs
```diff
@@ -445,22 +445,11 @@ impl EthereumCallFilter {
...
-        // The first one is acting as a proxy to the second one (an atomicMatch_).
-        // If you try to decode the first call as it is/comes from a `traces` RPC request, it
-        // will fail because it has a smaller size/length.
-        let correct_input_size = (call.input.0.len() - 4) % 32 == 0;
-        if !correct_input_size {
-            return false;
-        }
-
+        // - https://github.com/graphprotocol/graph-node/pull/3326
+        // - ae6951328df4270e3153897220fdef1822055859
```
leoyvens, line 452: This comment is hard to understand if you don't have all the context. Maybe this should say that decoding failures are filtered elsewhere, and have a 'see also' to that code.

evaporei: Ok, I'll improve on it!

evaporei: Do you think it's better now or do you have some other suggestion? 📝

evaporei:
The CI is failing, I'll fix it.

leoyvens (approved):

chain/ethereum/src/data_source.rs
```diff
@@ -628,6 +628,8 @@ impl DataSource {
                     },
                 ) {
                     Ok(val) => val,
+                    // Link to where the previous validation was
```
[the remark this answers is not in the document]
evaporei: Done.

leoyvens (approved):
Good to squash and merge

alcercu:
Thanks for the quick answer @leoyvens! I was asking because I thought this patch was aimed at solving this:
https://github.com/evaporei/poh-investigation
But in the hosted service version of the subgraph, the error is still present:
https://thegraph.com/hosted-service/subgraph/kleros/proof-of-humanity-mainnet

Let me know if you need more info! Thanks a lot for looking into this!

leoyvens:
@alcercu The subgraph will unfortunately need to be resynced to be able to catch the missing calls.

alcercu:
Perfect, thanks @leoyvens and @evaporei! 🙌🏻

PR #3671, opened 2022-06-15
store/postgres: warn if locale is not en_US.utf8

### LINKED ISSUE #3118: Mandate everything runs in some fixed locale `en_US.UTF8`

saihaj:
Check for user's locale on startup and if it is not `en_US.utf8` panic.

Closes #3118


dotansimha (approved):

leoyvens:
We need a guide for what an indexer should do when encountering this error, and the error message should link to that guide.

dotansimha (approved):

azf20:
@saihaj from discussion with @leoyvens , I think we may want to make the error message a bit more informative (i.e. directing the user to change their DB locale to match, or change the env variable)

lutter:
Here's how I think we should try and manage using the same locale for all indexers. We'll probably be stuck on `en_US.UTF-8` because there are fewer indexers that need to change, though `C` would be more preferable for performance.

1. Communicate a timeline and some writeup on what indexers need to do
1. Print a warning when the locale is not the specified one for a while (probably months, definitely two releases)
2. Switch to enforce the locale, i.e., the current behavior of panicking on startup

saihaj:
https://github.com/graphprotocol/graph-node/pull/4163

PR #3675, opened 2022-06-17
store: Only create exclusion constraints in debug builds

lutter:
In release builds, we create the same index as the exclusion constraint but not the constraint since constraint enforcement is fairly slow. Keeping the constraint for debug builds will have tests that violate the constraint blow up.

We could make that an explicit configuration, but we have so many of them already.

### DIFF
--- a/store/postgres/src/catalog.rs
+++ b/store/postgres/src/catalog.rs
@@ -47,6 +47,13 @@ table! {
     }
 }
 
+// In debug builds (for testing etc.) create exclusion constraints, in
+// release builds for production, skip them
+#[cfg(debug_assertions)]
+const CREATE_EXCLUSION_CONSTRAINT: bool = true;
+#[cfg(not(debug_assertions))]
+const CREATE_EXCLUSION_CONSTRAINT: bool = false;
+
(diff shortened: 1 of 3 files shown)


store/postgres/src/catalog.rs
```diff
@@ -47,6 +47,13 @@ table! {
     }
 }
 
+// In debug builds (for testing etc.) create exclusion constraints, in
+// release builds for production, skip them
+#[cfg(debug_assertions)]
+const CREATE_EXCLUSION_CONSTRAINT: bool = true;
+#[cfg(not(debug_assertions))]
+const CREATE_EXCLUSION_CONSTRAINT: bool = true;
```
leoyvens, line 55:
```suggestion
const CREATE_EXCLUSION_CONSTRAINT: bool = false;
```

leoyvens (changes requested):

store/postgres/src/catalog.rs
lutter: oh wow .. fixed that, thanks!

leoyvens (approved):

PR #3677, opened 2022-06-17
feat(graph,graphql): filter by child entity (interfaces)

kamilkisiela:
Continuation of #3184, adds support for interfaces.

<details>
<summary>GraphQL Schema</summary>

```graphql
type Musician @entity {
  id: ID!
  name: String!
  mainBand: Band
  bands: [Band!]!
  writtenSongs: [Song]! @derivedFrom(field: "writtenBy")
}

type Band @entity {
  id: ID!
  name: String!
  members: [Musician!]! @derivedFrom(field: "bands")
  reviews: [BandReview] @derivedFrom(field: "band")
  originalSongs: [Song!]!
}

type Song @entity {
  id: ID!
  title: String!
  writtenBy: Musician!
  publisher: Publisher!
  band: Band @derivedFrom(field: "originalSongs")
  reviews: [SongReview] @derivedFrom(field: "song")
}

type SongStat @entity {
  id: ID!
  song: Song @derivedFrom(field: "id")
  played: Int!
}

type Publisher {
  id: Bytes!
}

interface Review {
  id: ID!
  body: String!
  author: User!
}

type SongReview implements Review @entity {
  id: ID!
  body: String!
  song: Song
  author: User!
}

type BandReview implements Review @entity {
  id: ID!
  body: String!
  band: Band
  author: User!
}

type User @entity {
  id: ID!
  name: String!
  bandReviews: [BandReview] @derivedFrom(field: "author")
  songReviews: [SongReview] @derivedFrom(field: "author")
  reviews: [Review] @derivedFrom(field: "author")
}

```
</details>

<details>
<summary>GraphQL Query</summary>

```graphql
query {
  users(first: 5, orderBy: id, where: { reviews_: { body_starts_with:  "Good" } }) {
    name
    reviews {
      body
    }
  }
}
```
</details>

<details>
<summary>SQL statement</summary>

```sql
select
  'User' as entity,
  to_jsonb(c.*) as data
from
  (
    select
      *
    from
      "sgd445"."user" c
    where
      c.block_range @> $1
      and (
        (
          exists (
            select
              1
            from
              "sgd445"."song_review" as i
            where
              c."id" = i."author"
              and i.block_range @> $2
              and (i."body" like $3)
          )
          or exists (
            select
              1
            from
              "sgd445"."band_review" as i
            where
              c."id" = i."author"
              and i.block_range @> $4
              and (i."body" like $5)
          )
        )
      )
    order by "id", block_range
    limit 100
  ) c
```
</details>

### DIFF
--- a/graphql/src/schema/api.rs
+++ b/graphql/src/schema/api.rs
@@ -228,7 +228,7 @@ fn field_filter_input_values(
                 .get_named_type(name)
                 .ok_or_else(|| APISchemaError::TypeNotFound(name.clone()))?;
             Ok(match named_type {
-                TypeDefinition::Object(_) => {
+                TypeDefinition::Object(_) | TypeDefinition::Interface(_) => {
                     let mut input_values = match ast::get_derived_from_directive(field) {
                         // Only add `where` filter fields for object and interface fields
                         // if they are not @derivedFrom
@@ -245,23 +245,6 @@ fn field_filter_input_values(
                     };
                     extend_with_child_filter_input_value(field, name, &mut input_values);
                     input_values
-                }
-                TypeDefinition::Interface(_) => {
-                    // Only add `where` filter fields for object and interface fields
-                    // if they are not @derivedFrom
-                    if ast::get_derived_from_directive(field).is_some() {
-                        vec![]
-                    } else {
-                        // We allow filtering with `where: { other: "some-id" }` and
-                        // `where: { others: ["some-id", "other-id"] }`. In both cases,
-                        // we allow ID strings as the values to be passed to these
-                        // filters.
-                        field_scalar_filter_input_values(
-                            schema,
-                            field,
-                            &ScalarType::new(String::from("String")),
-                        )
-                    }
                 }
                 TypeDefinition::Scalar(ref t) => field_scalar_filter_input_values(schema, field, t),
                 TypeDefinition::Enum(ref t) => field_enum_filter_input_values(schema, field, t),
@@ -381,11 +364,14 @@ fn field_list_filter_input_values(
                     )
                 }
             }
-            TypeDefinition::Interface(_) => {
+            TypeDefinition::Interface(parent) => {
                 if ast::get_derived_from_directive(field).is_some() {
-                    (None, None)
+                    (None, Some(parent.name.clone()))
                 } else {
-                    (Some(Type::NamedType("String".into())), None)
+                    (
+                        Some(Type::NamedType("String".into())),
+                        Some(parent.name.clone()),
+                    )
                 }
             }
             TypeDefinition::Scalar(ref t) => (Some(Type::NamedType(t.name.to_owned())), None),
@@ -1174,6 +1160,7 @@ mod tests {
                 "name_ends_with_nocase",
                 "name_not_ends_with",
                 "name_not_ends_with_nocase",
+                "pets_",
                 "favoritePet",
                 "favoritePet_not",
                 "favoritePet_gt",
@@ -1194,6 +1181,7 @@ mod tests {
                 "favoritePet_ends_with_nocase",
                 "favoritePet_not_ends_with",
                 "favoritePet_not_ends_with_nocase",
+                "favoritePet_",
                 "_change_block"
             ]
             .iter()
--- a/graphql/src/schema/api.rs
+++ b/graphql/src/schema/api.rs
@@ -354,24 +354,12 @@ fn field_list_filter_input_values(
         // derived, we allow ID strings to be passed on.
         // Adds child filter only to object types.
         let (input_field_type, parent_type_name) = match typedef {
-            TypeDefinition::Object(parent) => {
+            TypeDefinition::Object(ObjectType { name, .. })
+            | TypeDefinition::Interface(InterfaceType { name, .. }) => {
                 if ast::get_derived_from_directive(field).is_some() {
-                    (None, Some(parent.name.clone()))
+                    (None, Some(name.clone()))
                 } else {
-                    (
-                        Some(Type::NamedType("String".into())),
-                        Some(parent.name.clone()),
-                    )
-                }
-            }
-            TypeDefinition::Interface(parent) => {
-                if ast::get_derived_from_directive(field).is_some() {
-                    (None, Some(parent.name.clone()))
-                } else {
-                    (
-                        Some(Type::NamedType("String".into())),
-                        Some(parent.name.clone()),
-                    )
+                    (Some(Type::NamedType("String".into())), Some(name.clone()))
                 }
             }
             TypeDefinition::Scalar(ref t) => (Some(Type::NamedType(t.name.to_owned())), None),
--- a/graphql/src/store/query.rs
+++ b/graphql/src/store/query.rs
@@ -261,21 +261,40 @@ fn build_child_filter_from_object(
     let child_entity = schema
         .object_or_interface(type_name)
         .ok_or(QueryExecutionError::InvalidFilterError)?;
-    let filter = build_filter_from_object(child_entity, object, schema)?;
+    let filter = Box::new(build_filter_from_object(child_entity, object, schema)?);
     let derived = field.is_derived();
+    let attr = match derived {
+        true => sast::get_derived_from_field(child_entity, field)
+            .ok_or(QueryExecutionError::InvalidFilterError)?
+            .name
+            .to_string(),
+        false => field_name,
+    };
 
-    Ok(EntityFilter::Child(Child {
-        attr: match derived {
-            true => sast::get_derived_from_field(child_entity, field)
-                .ok_or(QueryExecutionError::InvalidFilterError)?
-                .name
-                .to_string(),
-            false => field_name,
-        },
-        entity_type: EntityType::new(type_name.to_string()),
-        filter: Box::new(filter),
-        derived,
-    }))
+    if child_entity.is_interface() {
+        Ok(EntityFilter::Or(
+            child_entity
+                .object_types(schema.schema())
+                .expect("Interface is not implemented by any types")
+                .iter()
+                .map(|object_type| {
+                    EntityFilter::Child(Child {
+                        attr: attr.clone(),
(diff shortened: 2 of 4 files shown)


kamilkisiela:
@azf20 @lutter please let me know what other use cases I should cover in the tests.

graphql/tests/query.rs
```diff
@@ -154,6 +163,7 @@ fn test_schema(id: DeploymentHash, id_type: IdType) -> Schema {
         id: ID!
         name: String!
         members: [Musician!]! @derivedFrom(field: \"bands\")
+        reviews: [BandReview] @derivedFrom(field: \"band\")
```
lutter, line 166: Is there a reason why this can't be `[BandReview!]!` like the other fields? IIRC, we do not allow arrays with nulls in them anyway. (There's a couple more places here that raise the same question)

graphql/tests/query.rs
```diff
@@ -174,6 +185,34 @@ fn test_schema(id: DeploymentHash, id_type: IdType) -> Schema {
...
+        author: User!
+    }
+
+    type User @entity {
+        id: ID!
+        name: String!
+        bandReviews: [BandReview] @derivedFrom(field: \"author\")
+        songReviews: [SongReview] @derivedFrom(field: \"author\")
+        reviews: [Review] @derivedFrom(field: \"author\")
+    }
```
lutter, line 215: It looks like all the interface fields in these tests are `@derivedFrom`. Could some of them be made into direct fields? In general, we want to make sure that all four combinations of derived/direct and single value/array fields work.

lutter (changes requested):
Looks good in general; surprising that it was this easy! I am a little worried about using `or` in queries, but we can look more into this once we have some examples of queries that use this feature and are slow and try out alternative queries.

The tests should be adapted a little, but other than that, looks good!

graphql/tests/query.rs
kamilkisiela: I added a few new test scenarios but I don't know exactly how the interfaces are supposed to work, what is possible, and what is not. I'm guessing here a bit. One test is failing but it yells at me about some non-existing operator (`operator does not exist: bytea = text`). I'm lost at this point :)

kamilkisiela:
I don't know exactly how the interfaces are supposed to work, what is possible, and what is not with and without the `@derivedFrom` directive. I'm guessing here a bit. Waiting for a few actual use cases to match my implementation against some real and confirmed to be working examples/tests.

graphql/tests/query.rs
lutter: That error means that somewhere where a SQL query compares two columns with `=`, one is of type `bytea` and the other of type `text`. That's an error somewhere in query generation, not necessarily connected to child filters. I wanted to try that out myself, but couldn't build this branch since `cargo` gets unhappy when it tries to check out `https://github.com/dotansimha/graphql-tools-rs?branch=clone-validation-error#89085f31` and says
```
object not found - no match for id (89085f316a71b390fecd9bd4c3f55c3884c33169); class=Odb (9); code=NotFound (-3)
```

The best way to get to the bottom of this is to turn on logging of all statements in Postgres with `alter system set log_statement = 'all';` in `psql` in your test database. (You might have to restart Postgres after this) With this set, run the failing test and the Postgres logs will have the query that is causing that error. Actually, I just noticed that the error message from the test has the full query, though I still recommend setting this.

lutter:
I just made a big table to get an overview of what combinations of parent and child types are possible and came up with this:

| parent intf | child intf | derived | list | parent   | child        |
|-------------|------------|---------|------|----------|--------------|
| f           | f          | f       | f    | Musician | mainBand     |
| f           | f          | f       | t    | Musician | bands        |
| f           | f          | t       | f    | Song     | band         |
| f           | f          | t       | t    | Band     | members      |
| f           | t          | f       | f    | User     | latestReview |
| f           | t          | f       | t    | Song     | media        |
| f           | t          | t       | f    | ---      | ---          |
| f           | t          | t       | t    | User     | reviews      |
| t           | f          | f       | f    | Review   | author       |
| t           | f          | f       | t    | ---      | ---          |
| t           | f          | t       | f    | Media    | song         |
| t           | f          | t       | t    | ---      | ---          |
| t           | t          | f       | f    | ---      | ---          |
| t           | t          | f       | t    | ---      | ---          |
| t           | t          | t       | f    | ---      | ---          |
| t           | t          | t       | t    | ---      | ---          |

They all are for a query along the lines of
```
{
  parents(where: { child_: { someField_filter: "value" } }) { .. stuff .. }
}
```

The columns mean:
* _parent intf_: should the parent be an interface
* _child intf_: should the child be an interface
* _derived_: should the child be derived
* _list_: should the child be a list/array
* _parent_: the type of the parent to use
* _child_: the name of the child field to use

I tried to fill this in with examples that I see possible from the test schema and marked the ones where I think that's not possible in the schema with `---` - since there's so many of them, I think we should try and make sure we cover all the cases that I marked as possible (they might all be covered already, haven't checked against the tests yet) and get them working. Once that's done, let's merge this PR and work on another one to cover more of the possible combinations here.

kamilkisiela:
@lutter the table says that `Song` type (not an interface) can have `media: [Media!]!` where `Media` is an interface.

For this query:

```graphql
query {
  songs(
    first: 100, 
    orderBy: id, 
    where: { 
      media_: {
        title_starts_with: "Cheesy Tune"
      }
    }
  ) {
    title
    media {
      title
    }
  }
}
```

This is the generated SQL query:

```sql
select 'Song' as entity, to_jsonb(c.*) as data from (select  *
  from "sgd52"."song" c
where c.block_range @> 1 and ((exists (select 1 from "sgd52"."photo" as i where i."id" = any(c."media") and i.block_range @> 1 and (i."title" like '%Cheesy Tune%')) or exists (select 1 from "sgd52"."video" as i where i."id" = any(c."media") and i.block_range @> 1 and (i."title" like '%Cheesy Tune%'))))
order by "id", block_range
limit 100) c
```

That query is valid and works.

The `operator does not exist: bytea = text` error comes from another SQL query that is not related to my code.

```sql
with matches as (select c.* from unnest($1::bytea[]) as q(id)
 cross join lateral (select 'Photo' as entity, c.id, c.vid, p.id::text as g$parent_id, c.block_range
/* children_type_c */  from rows from (unnest($2), reduce_dim(array[array['0xf1', '0xf2']]::text[][])) as p(id, child_ids) cross join lateral (select  *  from "sgd54"."photo" c where c.block_range @> $3 and q.id = p.id and c.id = any(p.child_ids)) c
union all
select 'Video' as entity, c.id, c.vid, p.id::text as g$parent_id, c.block_range
/* children_type_c */  from rows from (unnest($4), reduce_dim(array[array['0xf1', '0xf2']]::text[][])) as p(id, child_ids) cross join lateral (select  *  from "sgd54"."video" c where c.block_range @> $5 and q.id = p.id and c.id = any(p.child_ids)) c
order by "id", block_range
 limit 100) c)
select m.*, to_jsonb("c".*)|| jsonb_build_object('g$parent_id', m.g$parent_id) as data
  from "sgd54"."photo" c, matches m
 where c.vid = m.vid and m.entity = 'Photo'
union all
select m.*, to_jsonb("c".*)|| jsonb_build_object('g$parent_id', m.g$parent_id) as data
  from "sgd54"."video" c, matches m
 where c.vid = m.vid and m.entity = 'Video'
 order by g$parent_id, "id"
```

graphql/src/schema/api.rs
```diff
@@ -381,11 +364,14 @@ fn field_list_filter_input_values(
                     )
                 }
             }
-            TypeDefinition::Interface(_) => {
+            TypeDefinition::Interface(parent) => {
```
wminshew, line 367: apologies if I'm missing something (relatively new to rust) but this expression can be combined w line 357 into `TypeDefinition::Object(parent) | TypeDefinition::Interface(parent) => ` as in line 231 i think? both of these match expressions (L358-366, L368-375) look the same to me

kamilkisiela: `parent` is different in both and I think Rust does not support unions. If I would merge them and try to access `parent.name` I would get an error.

wminshew: I see -- so it works in L231 bc you're discarding the object anyway and here the compiler ~can't understand they both have `.name` independently? and so in this case it would need to use .. some kind of trait?

appreciate the response, thanks!

lutter: Nice catch - you can make this work if you destructure further:
```rust
            TypeDefinition::Object(ObjectType { name, .. })
            | TypeDefinition::Interface(InterfaceType { name, .. }) => {
                if ast::get_derived_from_directive(field).is_some() {
                    (None, Some(name.clone()))
                } else {
                    (Some(Type::NamedType("String".into())), Some(name.clone()))
                }
            }
```

lutter:
> @lutter the table says that `Song` type (not an interface) can have `media: [Media!]!` where `Media` is an interface.
> ... 
> Can you confirm it's not supported by graph-node or it's a bug?

Yes, that is indeed a pre-existing bug related to using `Bytes` for ids, caused by the `media { title }` part of the GraphQL query. I just added a commit to this PR that fixes that issue. You might want to move that commit before the commit that adds the test, I just didn't want to change too much in your PR.

I am glad that adding these tests actually uncovered that issue!

graphql/src/store/query.rs
```diff
@@ -287,6 +289,39 @@ fn build_child_filter_from_object(
...
+
+                    Ok(EntityFilter::Child(Child {
+                        attr: attr.clone(),
+                        entity_type: EntityType::new(child_entity.name().to_string()),
+                        filter: filter.clone(),
+                        derived,
+                    }))
+                })
+                .collect::<Result<Vec<EntityFilter>, QueryExecutionError>>()?,
+        ))
```
lutter, line 324: Wow .. nice. That's a pretty nasty case. We'll see how it works out in practice, but it could be that all those `or`'s end up being slow and we need to change how the query is generated; but it's easier to determine that once we have an example of that.

lutter (approved):
Great! Let's merge it!

PR #3678, opened 2022-06-17
Fast PoI

That3Percent:
Changes:
* (Performance) Introduces Fast PoI implementation, but currently switched off (needs to be enabled through subgraph manifest)
* (Performance) Optimize WASM Heap access by avoiding unnecessary heap allocations and writes
* (Hygiene) Merge AscGet and TryAscGet
* (Hygiene) Use const generics for FromAscObj [u8; LEN] instead of duplicated code

### DIFF
--- a/chain/near/src/trigger.rs
+++ b/chain/near/src/trigger.rs
@@ -153,7 +153,6 @@ mod tests {
         runtime::gas::GasCounter,
         util::mem::init_slice,
     };
-    use second_stack::Stack;
 
     #[test]
     fn block_trigger_to_asc_ptr() {
@@ -424,7 +423,6 @@ mod tests {
 
     struct BytesHeap {
         api_version: graph::semver::Version,
-        stack: Stack,
         memory: Vec<u8>,
     }
 
@@ -433,7 +431,6 @@ mod tests {
             Self {
                 api_version,
                 memory: vec![],
-                stack: Stack::new(),
             }
         }
     }
@@ -452,10 +449,6 @@ mod tests {
             let mut data = [std::mem::MaybeUninit::<u8>::uninit(); 4];
             let init = self.read(offset, &mut data, gas)?;
             Ok(u32::from_le_bytes(init.try_into().unwrap()))
-        }
-
-        fn stack(&self) -> &Stack {
-            &self.stack
         }
 
         fn read<'a>(
--- a/chain/near/src/trigger.rs
+++ b/chain/near/src/trigger.rs
@@ -450,7 +450,7 @@ mod tests {
 
         fn read_u32(&self, offset: u32, gas: &GasCounter) -> Result<u32, DeterministicHostError> {
             let mut data = [std::mem::MaybeUninit::<u8>::uninit(); 4];
-            let init = self.init(offset, &mut data, gas)?;
+            let init = self.read(offset, &mut data, gas)?;
             Ok(u32::from_le_bytes(init.try_into().unwrap()))
         }
 
@@ -458,8 +458,8 @@ mod tests {
             &self.stack
         }
 
-        fn init<'s, 'a>(
-            &'s self,
+        fn read<'a>(
+            &self,
             offset: u32,
             buffer: &'a mut [std::mem::MaybeUninit<u8>],
             _gas: &GasCounter,
--- a/chain/near/src/trigger.rs
+++ b/chain/near/src/trigger.rs
@@ -5,11 +5,7 @@ use graph::cheap_clone::CheapClone;
 use graph::prelude::hex;
 use graph::prelude::web3::types::H256;
 use graph::prelude::BlockNumber;
-use graph::runtime::asc_new;
-use graph::runtime::gas::GasCounter;
-use graph::runtime::AscHeap;
-use graph::runtime::AscPtr;
-use graph::runtime::DeterministicHostError;
+use graph::runtime::{asc_new, gas::GasCounter, AscHeap, AscPtr, DeterministicHostError};
 use std::{cmp::Ordering, sync::Arc};
 
 use crate::codec;
@@ -146,7 +142,7 @@ pub struct ReceiptWithOutcome {
 
 #[cfg(test)]
 mod tests {
-    use std::{convert::TryFrom, mem::MaybeUninit};
+    use std::convert::TryFrom;
 
     use super::*;
 
@@ -155,6 +151,7 @@ mod tests {
         data::subgraph::API_VERSION_0_0_5,
         prelude::{hex, BigInt},
         runtime::gas::GasCounter,
+        util::mem::init_slice,
     };
     use second_stack::Stack;
 
@@ -495,15 +492,7 @@ mod tests {
 
             let src = &self.memory[start_offset..end_offset_exclusive];
 
-            // See also 6d2f040c-9361-4cf9-8b3c-426f87cc0b21
-            unsafe {
-                let uninit_src: &[MaybeUninit<u8>] = std::mem::transmute(src);
-                buffer.copy_from_slice(uninit_src);
-                // This is copied from slice_assume_init_mut, which is
-                // currently an unstable API
-                let buffer = &mut *(buffer as *mut [MaybeUninit<u8>] as *mut [u8]);
-                Ok(buffer)
-            }
+            Ok(init_slice(src, buffer))
         }
 
         fn api_version(&self) -> graph::semver::Version {
--- a/chain/near/src/trigger.rs
+++ b/chain/near/src/trigger.rs
@@ -146,7 +146,7 @@ pub struct ReceiptWithOutcome {
 
 #[cfg(test)]
 mod tests {
-    use std::convert::TryFrom;
+    use std::{convert::TryFrom, mem::MaybeUninit};
 
     use super::*;
 
@@ -438,7 +438,7 @@ mod tests {
         }
     }
 
-    impl AscHeap for BytesHeap {
+    unsafe impl AscHeap for BytesHeap {
         fn raw_new(
             &mut self,
             bytes: &[u8],
@@ -448,12 +448,12 @@ mod tests {
             Ok((self.memory.len() - bytes.len()) as u32)
         }
 
-        fn get(
-            &self,
+        fn init<'s, 'a>(
+            &'s self,
             offset: u32,
-            size: u32,
+            buffer: &'a mut [std::mem::MaybeUninit<u8>],
             _gas: &GasCounter,
-        ) -> Result<Vec<u8>, DeterministicHostError> {
+        ) -> Result<&'a mut [u8], DeterministicHostError> {
             let memory_byte_count = self.memory.len();
             if memory_byte_count == 0 {
                 return Err(DeterministicHostError::from(anyhow!(
@@ -462,7 +462,7 @@ mod tests {
             }
 
             let start_offset = offset as usize;
-            let end_offset_exclusive = start_offset + size as usize;
+            let end_offset_exclusive = start_offset + buffer.len();
 
             if start_offset >= memory_byte_count {
                 return Err(DeterministicHostError::from(anyhow!(
@@ -480,7 +480,17 @@ mod tests {
                 )));
             }
 
-            return Ok(Vec::from(&self.memory[start_offset..end_offset_exclusive]));
+            let src = &self.memory[start_offset..end_offset_exclusive];
+
+            // See also 6d2f040c-9361-4cf9-8b3c-426f87cc0b21
+            unsafe {
+                let uninit_src: &[MaybeUninit<u8>] = std::mem::transmute(src);
+                buffer.copy_from_slice(uninit_src);
+                // This is copied from slice_assume_init_mut, which is
+                // currently an unstable API
+                let buffer = &mut *(buffer as *mut [MaybeUninit<u8>] as *mut [u8]);
+                Ok(buffer)
+            }
         }
 
         fn api_version(&self) -> graph::semver::Version {
--- a/chain/near/src/trigger.rs
+++ b/chain/near/src/trigger.rs
@@ -438,7 +438,7 @@ mod tests {
         }
     }
 
-    unsafe impl AscHeap for BytesHeap {
+    impl AscHeap for BytesHeap {
         fn raw_new(
             &mut self,
             bytes: &[u8],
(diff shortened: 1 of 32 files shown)


chain/near/src/trigger.rs
```diff
@@ -480,7 +480,17 @@ mod tests {
                 )));
             }
 
-            return Ok(Vec::from(&self.memory[start_offset..end_offset_exclusive]));
+            let src = &self.memory[start_offset..end_offset_exclusive];
+
+            // See also 6d2f040c-9361-4cf9-8b3c-426f87cc0b21
+            unsafe {
+                let uninit_src: &[MaybeUninit<u8>] = std::mem::transmute(src);
```
mangas, line 487: Could this block be achieved without unsafe? 

From the docs: 
```
transmute is incredibly unsafe. There are a vast number of ways to cause [undefined behavior](https://doc.rust-lang.org/stable/reference/behavior-considered-undefined.html) with this function. transmute should be the absolute last resort.
```

graph/src/components/subgraph/proof_of_indexing/online.rs
```diff
@@ -54,82 +92,75 @@ pub enum BlockEventStream {
...
-                ref mut vec_length,
-                handler_start: _,
-                seq_no,
-                ref mut digest,
-            } => {
-                *vec_length += 1;
-                event.stable_hash(seq_no.next_child(), digest);
-            }
-        }
+        let children = &[
```
mangas, line 146: Could use some more information about what is going on here. I might just be missing context but overall this section is not very clear to me even with the existing comments.

graph/src/util/stable_hash_glue.rs
```diff
@@ -0,0 +1,82 @@
...
+            fn stable_hash<H: stable_hash_legacy::StableHasher>(&self, sequence_number: H::Seq, state: &mut H,
+            ) {
+                let Self(transparent) = self;
+                $(let transparent = $e(transparent);)*
+                stable_hash_legacy::StableHash::stable_hash(&transparent, sequence_number, state);
+            }
+        }
+    };
+}
+pub use crate::_impl_stable_hash as impl_stable_hash;
```
mangas, line 62: why do we need to re-export with a different name?

graph/src/util/stable_hash_glue.rs
```diff
@@ -0,0 +1,82 @@
+use stable_hash::{StableHash, StableHasher};
+use stable_hash_legacy::prelude::{
+    StableHash as StableHashLegacy, StableHasher as StableHasherLegacy,
+};
+
+/// Implements StableHash and StableHashLegacy. This macro supports two forms:
+/// Struct { field1, field2, ... } and Tuple(transparent). Each field supports
+/// an optional modifier. For example: Tuple(transparent: AsBytes)
+#[macro_export]
+macro_rules! _impl_stable_hash {
```
mangas, line 10: would it be worth adding some tests to ensure changes to this code don't break anything?

leoyvens:
I have a couple worries on the heap optimization aspect of this.

On `second-stack`, the scoped bump allocator is a cool thing, but being thread local means that the buffer would never be freed for as long as the subgraph is running, this is likely to increase steady state memory usage of index nodes running multiple subgraphs. If the memory were freed at the end of the handler execution, I believe we could still get most of the performance benefit while additionally making it suitable for async handler execution should that be a requirement in the future.

Making the `AscHeap` trait unsafe seems excessive. I only see `asc_get_array` relying on that unsafe invariant, is it possible to remove that assumption on the initialization of the buffer, even if at perhaps a small cost to the performance of `asc_get_array`?

graph/src/util/stable_hash_glue.rs
That3Percent: Yes, I can add some backwards-compatibility checks for the PoI in particular.

mangas, line 10: I also think this macro would probably be more suited to be added to the stable-hash crate since it mostly deals with the details from that crate and how it implements this trait for structs. Is there a reason why we shouldn't move it there and add some test coverage for it? Possibly even backport it to the old version of stable-hash is that makes sense? In that case I guess this would just rely those two? Thoughts?

graph/src/util/stable_hash_glue.rs
mangas: understood, thanks for clarifying 🙏

chain/near/src/trigger.rs
That3Percent: Not really. Working with `MaybeUninit` infers that there will be unsafety somewhere. And though `transmute` comes with strong warnings, it is currently the only way to work with `MaybeUninit` effectively. There are some nightly APIs that will simplify our use of unsafe, like [write_slice](https://doc.rust-lang.org/stable/std/mem/union.MaybeUninit.html#method.write_slice) but for now we have to copy-paste the internals of those APIs.

That3Percent:
@leoyvens 
> being thread local means that the buffer would never be freed for as long as the subgraph is running, this is likely to increase steady state memory usage of index nodes running multiple subgraphs. If the memory were freed at the end of the handler execution, I believe we could still get most of the performance benefit while additionally making it suitable for async handler execution should that be a requirement in the future.
> 
> Making the `AscHeap` trait unsafe seems excessive. I only see `asc_get_array` relying on that unsafe invariant, is it possible to remove that assumption on the initialization of the buffer, even if at perhaps a small cost to the performance of `asc_get_array`?

That3Percent:
@leoyvens 

> being thread local means that the buffer would never be freed for as long as the subgraph is running, this is likely to increase steady state memory usage of index nodes running multiple subgraphs. If the memory were freed at the end of the handler execution, I believe we could still get most of the performance benefit while additionally making it suitable for async handler execution should that be a requirement in the future.

Typically, if async use constrains the number of threads to a small amount (eg: 2x cores plus some odd one-offs). I just did a quick check to see what the threading model is in graph-node though and it looks like we are using a thread per subgraph deployment, so the concern is warranted.

We can add an API `second_stack::try_free() -> Result<(), usize>` which will return `Ok(())` if the memory can be freed, and `Err(len)` if the heap is currently in-use further up in the stack. We can then call this function at the end of handler execution. This will lose some efficiency by requiring re-allocating memory, but will also lower the steady-state memory pressure.

I want to better understand this part though: "additionally making it suitable for async handler execution should that be a requirement in the future" since I think more async improves the situation by re-cycling memory across tasks/subgraphs as threads are recycled.

That3Percent:
> Making the `AscHeap` trait unsafe seems excessive. I only see `asc_get_array` relying on that unsafe invariant, is it possible to remove that assumption on the initialization of the buffer, even if at perhaps a small cost to the performance of `asc_get_array`?

You are correct that this is the only case we rely on that unsafe invariant, and that the code is motivated by performance.

I tend to take an almost militaristic stance on performance because I've observed that doing so consistently makes a huge difference in the long run. A 2% improvement daily is > 1500x gain over 1 year. And though it may seem like hyperbole, I've seen this story play out when people turn out to the typical "this performance is too small to be worth it" culture. Surprisingly, in that story the code ended up more maintainable/understandable in the end as well because unnecessary abstraction was often a culprit. Long story short, when code is on the hot path the "gain too small" argument rarely resonates with me. In this case, I think we have a better option anyway...

If the `AscHeap` usage was object-safe we could add `AscHeap::get_array<const LEN: usize>(offset: u32, gas: &GasCounter)` which wouldn't require unsafe at all. Since we can't do that easily...

For now, how about simply `AscHeap::get_u32`? That's hyper specific but covers the existing calls for this method.

graph/src/util/stable_hash_glue.rs
That3Percent: This is done.

That3Percent: We now have macros in both the fast and legacy stable-hash crates, and a graph-node macro that delegates to the other two.

That3Percent:
@mangas Addressed all concerns.

@leoyvens Optimistically went with `AscHeap::get_u32` to reduce unsafe. Awaiting your final verdict on `second-stack` threadlocal usage.

runtime/wasm/src/module/mod.rs
```diff
@@ -652,30 +657,56 @@ impl<C: Blockchain> AscHeap for WasmInstanceContext<C> {
...
+                )))?;
+
+            // See also 6d2f040c-9361-4cf9-8b3c-426f87cc0b21
+            // When MaybeUninit::write_slice is stabilized, this won't
+            // require any unsafe code and will be a one-liner
+            let uninit_src: &[MaybeUninit<u8>] = std::mem::transmute(src);
+            buffer.copy_from_slice(uninit_src);
+            // This is copied from slice_assume_init_mut, which is
+            // currently an unstable API
+            let buffer = &mut *(buffer as *mut [MaybeUninit<u8>] as *mut [u8]);
```
leoyvens, line 707: This is copy pasted in two places, could it be extracted to a safe utility function somewhere in `graph`?

leoyvens:
I am happy with `get_u32`.

Thanks for the breakdown on the thread local vs owned API. The `GasCounter` situation is unfortunate, but it does not need to repeat itself with this stack, I believe it can be nicely added as a field in `WasmInstanceContext`. Based on your other arguments I do still prefer the owned API as it would be easier to learn and reason about.

graph/src/runtime/asc_heap.rs
```diff
@@ -11,12 +13,14 @@ pub trait AscHeap {
     /// Allocate new space and write `bytes`, return the allocated address.
     fn raw_new(&mut self, bytes: &[u8], gas: &GasCounter) -> Result<u32, DeterministicHostError>;
 
-    fn get(
-        &self,
+    fn init<'s, 'a>(
```
leoyvens, line 16: Now that this doesn't actually guarantee that the buffer is init'ed, maybe this should be renamed (back to `get` or something more creative). Also I think the `'s` lifetime can be elided.

That3Percent:
> I believe it can be nicely added as a field in `WasmInstanceContext`.

I refactored `second-stack` to introduce an owned `Stack`: [https://docs.rs/second-stack/0.3.3/second_stack/struct.Stack.html](https://docs.rs/second-stack/0.3.3/second_stack/struct.Stack.html)

But, I fail to see how to add this cleanly to `WasmInstanceContext`. It may just be my unfamiliarity with `graph-node`. My best guess is to add `AscHeap::second_stack(&self) -> &Stack`. To use then you would need to call both `second_stack` and then `init`, which seems a bit redundant. Is that what you had in mind or is there a better API I'm missing?

leoyvens:
@That3Percent `WasmInstanceContext` is what implements `AscHeap`, so the stack would be accessible through `self.stack`.

That3Percent:
@leoyvens 
> @That3Percent `WasmInstanceContext` is what implements `AscHeap`, so the stack would be accessible through `self.stack`. But probably I'm the one missing something we can chat about this.

I made a commit for this which works. It demonstrates the "problems" which aren't problems so much now that it's implemented.

The caller of `heap.init` needs the slice beforehand. (heap.init cannot return a local borrow). So, that infers we call `heap.stack().uninit_slice(..)` using the `AscHeap` trait, not `WasmInstanceCtx`. But, if we add `fn stack(&self) -> &Stack` to `AscHeap`, then `WasmInstance` can no longer implement `AscHeap` because if it tries to implement this method by proxy it must return a borrow of a temporary.

Thankfully only tests relied on the implementation of `AscHeap` for `WasmInstance`. So, there was clean way to go by introducing a method which also made the tests more terse.

The commit may be more clear: [https://github.com/graphprotocol/graph-node/pull/3678/commits/1b0f8a03a2391c55c50548e80854841a9d7e931a](https://github.com/graphprotocol/graph-node/pull/3678/commits/1b0f8a03a2391c55c50548e80854841a9d7e931a)

runtime/wasm/src/module/mod.rs
That3Percent: Done

graph/src/runtime/asc_heap.rs
That3Percent: Done

mangas (approved):
LGTM

That3Percent:
@leoyvens So I guess the next steps are you running this on the integration cluster?

mangas:
@azf20 i believe it's just a sanity check because of the runtime changes. The POI now have some test coverage with hardcore values generated from the previous release (I think)

That3Percent:
@azf20 
> Do we need to upgrade subgraphs with a new ApiVersion to test them?

Before we can do that there are two places in the code that need to be modified to switch this on. They can be found by searching for `a0a79c0f-919f-4d97-a82 a-439a1ff78230`. All that needs to be done is to change `ProofOfIndexingVersion::Legacy` to `ProofOfIndexingVersion::Fast` if the new ApiVersion is enabled.

> And aren't POI divergences expected (as it's a new POI algorithm)?

The ApiVersion should prevent divergences. Unfortunately this means that existing subgraphs won't benefit.

azf20:
OK, thanks - yes I think we do need a new apiVersion here, so I think that would still be a necessary addition to this PR @mangas ?

mangas:
> OK, thanks - yes I think we do need a new apiVersion here, so I think that would still be a necessary addition to this PR @mangas ?

I think this PR is quite large in size and scope, we should prolly have a separate PR for the version changes

That3Percent:
Don't merge this yet. I need to make a fix, but can't until next week.

That3Percent:
@leoyvens

I just pushed a fix to remove the usage of second-stack. The problem was that I introduced undefined behavior when I refactored to move from the threadlocal allocation to the `Stack` owning an allocation. It is non-trivial to get `Stack` to pass the borrow checker without UB. And, it would also be non-trivial to use `Stack` correctly in graph-node with the API that would be required to avoid UB.

The latest change is still going to be a performance boost in all cases as compared to before the PR, but it will not be as significant as could have been afforded by a threadlocal second-stack because we still allocate a `Vec` in some cases (though, at least we don't zero it anymore in those cases).

The PR should be ready to merge again.

leoyvens:
Good you caught that! The switch to a sort of smallvec approach looks good, might go a long way for performance. Could you just rebase this?

That3Percent:
@leoyvens 

Rebased

That3Percent:
@leoyvens

Unfortunately, `cargo check --release --all-targets` fails on `master`, and therefore also fails now in this rebased PR. Since this works without either `--release` or `--all-targets` I'm guessing there is a `#[cfg(not(debug_assertions))]` or something along those lines throwing it off.

In any case, not affected by this PR but something to fix for the graph-node team. Some of the errors are bizarre, like this one...
```
error[E0599]: no method named `clone` found for struct `std::sync::Arc<_>` in the current scope
  --> store/postgres/src/relational/query_tests.rs:36:43
   |
36 |     let catalog = Catalog::for_tests(site.clone()).expect("Can not create catalog");
   |                                           ^^^^^ method not found in `std::sync::Arc<_>`
```

mangas:
This PR ran successfully on the integration cluster for 24h, I think we can merge it

That3Percent:
Well, turns out I was wrong about the API causing potential UB. `second-stack` has no UB that I'm aware of and the issue I was worried about was a false alarm. I've gone over the API and done code review with the Rust "Dark Arts" folks (including the implementer of Stacked Borrows), and run miri on the fuzz tests. That's about as good as it gets.

If anyone wants to re-add that optimization you are free to do so in a separate PR.

PR #2220, opened 2021-02-25
Make CI run faster

tilacog:

PR #3689, opened 2022-06-24
Improve end-to-end test runner

leoyvens:
This continues the work by @mangas in #3454 and #3466, towards a lightweight end-to-end unit testing framework. The goal is for tests to have as much end-to-end coverage as possible while being fast and reliable in local and CI testing, and also being simple to write and debug.

The existing integration test suite spins up a full graph node process, ganache, PG and IPFS, all in docker, deploys contracts and a subgraph through truffle and graph-cli and then runs the tests through a JS scripts. It integrates many pieces but has been slow and flaky on CI, despite efforts to improve it in #2220. And lately nobody really wants to maintain it or write new tests for it. So an alternative is needed.

The proposed approach does not depend on solidity and ganache, and the tests are all in rust. The sacrifice in coverage is that the block stream is entirely mocked. Not testing the block stream is less concerning now that we're moving towards the firehose block stream is much leaner than the polling block stream.

This repurposes the existing sample test to be a reimplementation the `data-source-revert` integration test. For this to be possible the improvements were:
- Deploy a local subgraph through graph-cli, rather than a subgraph existing on a remote IPFS endpoint.
- Use config files for the store.
- Ensure the name and deployment are removed before and after the test.
- Support reorgs in the static block stream.
- Minor change to the subgraph runner, to stop a subgraph if the block stream ends.
- Check the health of the subgraph.

There is still only one test so the code isn't yet very organized, but as we add more tests it will be possible to extract the reusable parts so this will actually become a framework.

### DIFF
--- a/tests/tests/runner.rs
+++ b/tests/tests/runner.rs
@@ -1,44 +1,18 @@
 use graph_chain_ethereum::chain::BlockFinality;
-use graph_store_postgres::SubgraphStore;
-use graph_tests::fixture::build_subgraph;
-use std::env::VarError;
-use std::pin::Pin;
+use graph_tests::fixture::{self, test_ptr};
 use std::sync::Arc;
-use std::task::{Context, Poll};
 use std::time::Duration;
-use tokio::fs::read_to_string;
 
-use anyhow::{anyhow, Error};
-use async_stream::stream;
+use anyhow::anyhow;
 use ethereum::trigger::{EthereumBlockTriggerType, EthereumTrigger};
-use futures::{Stream, StreamExt};
-use graph::blockchain::block_stream::{
-    BlockStream, BlockStreamBuilder, BlockStreamEvent, BlockWithTriggers,
-};
-use graph::blockchain::{
-    Block, BlockHash, BlockPtr, Blockchain, BlockchainMap, ChainIdentifier, RuntimeAdapter,
-    TriggersAdapter, TriggersAdapterSelector,
-};
-use graph::cheap_clone::CheapClone;
-use graph::components::store::{BlockStore, DeploymentLocator};
-use graph::env::ENV_VARS;
-use graph::firehose::{FirehoseEndpoint, FirehoseEndpoints};
-use graph::ipfs_client::IpfsClient;
+use graph::blockchain::block_stream::BlockWithTriggers;
+use graph::blockchain::{Block, BlockPtr};
 use graph::prelude::ethabi::ethereum_types::{H256, U64};
 use graph::prelude::{
-    async_trait, BlockNumber, DeploymentHash, LightEthereumBlock, LoggerFactory, MetricsRegistry,
-    NodeId, SubgraphAssignmentProvider, SubgraphName, SubgraphRegistrar, SubgraphStore as _,
-    SubgraphVersionSwitchingMode,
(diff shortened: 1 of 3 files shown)


tests/tests/runner.rs
```diff
@@ -33,81 +35,119 @@ use graph_core::{
...
+    let hash = {
+        let mut line = deploy_output
+            .lines()
+            .find(|line| line.contains(ID_PREFIX))
+            .expect("found no matching line");
+        if !line.starts_with(ID_PREFIX) {
+            line = &line[5..line.len() - 5]; // workaround for colored output
+        }
+        DeploymentHash::new(line.trim_start_matches(ID_PREFIX)).unwrap()
+    };
```
mangas, line 78: Prolly worth moving most stuff up to this line into a common::setup to be re-usable but also help read the test content

tests/tests/runner.rs
```diff
@@ -33,81 +35,119 @@ use graph_core::{
...
+                },
+            )),
+            trigger_data: vec![EthereumTrigger::Block(
+                test_ptr(1),
+                EthereumBlockTriggerType::Every,
+            )],
+        };
+        let block_2_reorged_hash = H256::from_low_u64_be(12);
+        let block_2_reorged = BlockWithTriggers::<Chain> {
+            block: graph_chain_ethereum::chain::BlockFinality::Final(Arc::new(
```
mangas, line 108: Could also be useful adding a new_final_block fn to use as a shorthand for this stuff, especially if we mostly care about 3 fields most of the time

tests/tests/runner.rs
```diff
@@ -288,11 +331,22 @@ async fn setup(
         logger_factory,
         provider: subgraph_provider,
         store: subgraph_store,
+        deployment_locator,
     }
 }
 
+fn cleanup(subgraph_store: &SubgraphStore, name: &SubgraphName, hash: &DeploymentHash) {
```
mangas, line 338: 🔥

tests/tests/runner.rs
```diff
@@ -345,15 +405,35 @@ impl<C: Blockchain> Stream for StaticStream<C> {
...
 fn stream_events<C: Blockchain>(
-    blocks: Vec<BlockStreamEvent<C>>,
-) -> impl Stream<Item = Result<BlockStreamEvent<C>, Error>> {
+    blocks: Vec<BlockWithTriggers<C>>,
+    current_idx: Option<usize>,
+) -> impl Stream<Item = Result<BlockStreamEvent<C>, Error>>
+where
+    C::TriggerData: Clone,
+{
     stream! {
```
mangas, line 414: Could be worth adding a comment somewhere around here describing how this works

mangas (approved):

tests/tests/runner.rs
leoyvens: Added a `fixture` module and moved that there.

tests/tests/runner.rs
leoyvens: There are nice helpers we could have for building chains, but I want to see some more tests to optimize for the patterns we see.

leoyvens:
Thanks for the review, I've also moved `fn setup` to a new `fixture` module so we can start separating the reusable code from test-specific code.

PR #3697, opened 2022-06-29
ethereum: Detect xdai eth_call revert

leoyvens:
We've seen [this subgraph](https://thegraph.com/hosted-service/subgraph/superfluid-finance/protocol-dev-xdai?selected=logs) get stuck on this. The RPC response is:
```
Error { code: ServerError(-32015), message: "VM execution error.", data: Some(String("revert")) }
```


hellwolf:
👀 thanks!

leoyvens:
@hellwolf to be clear, this will only benefit the subgraph if it uses `try_*` calls to handle reverts. If you don't understand why this is reverting, we'll try to reproduce the call manually so you can investigate.

PR #3710, opened 2022-07-05
Expose expensive queries file as an option

kaiwetlesen:
This pull request replaces the defined constant with EXPENSIVE_QUERIES with a parameter of a similar name. It also adds documentation in the testament describing the functionality exposed with the expensive queries file. Logs into the information log which file is used.

### DIFF
--- a/node/src/main.rs
+++ b/node/src/main.rs
@@ -47,14 +47,20 @@ use tokio::sync::mpsc;
 
 git_testament!(TESTAMENT);
 
-fn read_expensive_queries() -> Result<Vec<Arc<q::Document>>, std::io::Error> {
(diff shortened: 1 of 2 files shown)


kaiwetlesen:
Background
==========

This PR came from my reviewing the `graph-node` code in order to determine the reason why `graph-node` does not respond to `SIGTERM` or other typical graceful exit signalling. I figure it'd be good to expose this potentially useful feature as a configurable parameter.

azf20:
Thanks @kaiwetlesen! @lutter interested in your thoughts on this one

lutter (approved):
Nice! At some point, we should rethink this whole mechanism, but that's independent of this PR

PR #3713, opened 2022-07-08
Enable fork base for graph-node Docker image

yoshiwarab:
This PR updates the start script in the docker directory to support the fork-base graph-cli argument.  The argument is always passed to graph-node command and defaults to `https://api.thegraph.com/subgraphs/id/` if no fork_base environment variable is present.  This should make it easy to use debug forking while running graph-node locally with Docker.

### DIFF
--- a/docker/start
+++ b/docker/start
@@ -50,7 +50,6 @@ wait_for_ipfs() {
 }
 
 run_graph_node() {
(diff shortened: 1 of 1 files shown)


leoyvens:
Thanks for the contribution, it would be best for the default to be no fork base.

yoshiwarab:
Updated, thanks for the feedback!

Issue #3715, opened 2022-07-11
Add setting to log GraphQL Validation failures


azf20:

schmidsi:
It would be great if those logs are also sent to the client that sent the query. A response could look something like:

```json
{
  "data": { ... },
  "extensions": { "validations": ["Line 13: Null provided for non-nullable value"] }
}
```

PR #3716, opened 2022-07-11
feat(graph, graphql): allow to run GraphQL validations silently

dotansimha:
Closes https://github.com/graphprotocol/graph-node/issues/3715 

## Changes in this PR 

- Change `ENABLE_GRAPHQL_VALIDATIONS` to be enabled by default.
- Added `SILENT_GRAPHQL_VALIDATIONS` env var that is (enabled by default)
- Document environment variables related to GraphQL validations 

## Behavior changes 

- No breaking changes are expected by this PR. 
- For every given incoming query, the GraphQL validations will run now, but will report the results silently to the log in case of an error.

### DIFF
--- a/graphql/src/execution/query.rs
+++ b/graphql/src/execution/query.rs
@@ -15,7 +15,9 @@ use graph::data::graphql::{ext::TypeExt, ObjectOrInterface};
 use graph::data::query::QueryExecutionError;
 use graph::data::query::{Query as GraphDataQuery, QueryVariables};
 use graph::data::schema::ApiSchema;
-use graph::prelude::{info, o, q, r, s, BlockNumber, CheapClone, Logger, TryFromValue, ENV_VARS};
+use graph::prelude::{
+    info, o, q, r, s, warn, BlockNumber, CheapClone, Logger, TryFromValue, ENV_VARS,
+};
 
 use crate::execution::ast as a;
 use crate::query::{ast as qast, ext::BlockConstraint};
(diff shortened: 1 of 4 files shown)


graphql/src/execution/query.rs
```diff
@@ -152,12 +154,25 @@ impl Query {
...
+                        )
+                    })
+                    .collect());
+            } else {
+                warn!(
+                  &logger,
+                  "GraphQL Validation failure";
+                  "query" => &query.query_text,
+                  "variables" => &query.variables_text,
+                  "errors" => validation_errors.iter().map(|e| e.message.clone()).collect::<Vec<_>>().join("\n")
```
tilacog, line 173: I believe the `.join("\n")` part would result in multi-line logs, which we are currently trying to avoid. Could we change the `"errors"` section to something different?

Maybe use an array formatting, like: `["first error", "second error"]`?

tilacog (commented):
LGTM.

Just left a comment about a possible multiline logging case that we would prefer to avoid. I think it should be good to merge after this.

I also noticed that we don't have any docs about GraphqQL validation environment variables, should we add them to the `docs/environment-variables.md` file?

schmidsi (commented):
I assume this is only about logging in the graph-node and not sending the validation results through graphql extensions as [described here](https://github.com/graphprotocol/graph-node/issues/3715#issuecomment-1180490879), right?

graphql/src/execution/query.rs
dotansimha: Sure, that makes sense. I changed it, thanks!

dotansimha:
> I assume this is only about logging in the graph-node and not sending the validation results through graphql extensions as [described here](https://github.com/graphprotocol/graph-node/issues/3715#issuecomment-1180490879), right?

Yeah, I didn't want to affect the response. Also to implement that we might need a more significant change in how we handle the response from `Query::new` (right @tilacog ? Because we'll need to "hint" the `Runner` that it needs to add this kind of error to the final response).

dotansimha:
> Just left a comment about a possible multiline logging case that we would prefer to avoid. I think it should be good to merge after this.

Thanks, this is fair! Fixed :) 

> I also noticed that we don't have any docs about GraphqQL validation environment variables, should we add them to the `docs/environment-variables.md` file?

Sure, done!

tilacog (approved):

Issue #3011, opened 2021-11-26
Fix/remove _padding fields in NEAR Rust/ASc conversion types


Jannis:

maoueh:
`AscEnum` also has an hard-coded `_padding` field [pub _padding: u32, // Make padding explicit.](https://github.com/graphprotocol/graph-node/blob/2e164c4a70be3361ad29e378d941fb578b442035/runtime/wasm/src/asc_abi/class.rs#L456-L457) that could be removed if proper alignment would be written to AssemblyScript memory correctly.

PR #3722, opened 2022-07-12
Automated generation of the `Asc` types

### LINKED ISSUE #2985: Automated type generation for new chains (Rust <> AS types)
Currently for each chain we have to create all of the `Asc` (AssemblyScript) types in the graph-node `runtime`.

This not only requires defining the types and fields themselves, but also to implement a few traits/interfaces, such as `ToAscObj` and `FromAscObj`.

Those should be automated via macros so that we have:

- Less code overall (remove all conversion code related to `ToAscObj`/`FromAscObj`)
- Less code required for adding new types + fields on the runtime (reduce developer time necessary on those)

pienkowb:
This PR resolves #2985 and #3011.


leoyvens:
Looks like this is accidentally updating all the dependencies in Cargo.lock

pienkowb:
> Looks like this is accidentally updating all the dependencies in Cargo.lock

@leoyvens Right. It should be good now.

runtime/derive/src/generate_asc_type.rs
```diff
@@ -0,0 +1,180 @@
...
+            }
+        } else {
+            "N/A".into()
+        }
+    } else {
+        "N/A".into()
+    }
+}
+
+//recurcive
```
evaporei, line 163: Typo in 'recursive'

runtime/derive/src/generate_asc_type.rs
```diff
@@ -0,0 +1,180 @@
...
+                    }
+                }
+                "Option" => match &ps.arguments {
+                    syn::PathArguments::AngleBracketed(v) => {
+                        if let syn::GenericArgument::Type(syn::Type::Path(p)) = &v.args[0] {
+                            let tp_nm = path_to_string(&p.path);
+                            if is_scalar(&tp_nm) {
+                                format!("Option<{}>", tp_nm)
+                            } else {
+                                format!("graph::runtime::AscPtr<crate::protobuf::Asc{}>", tp_nm)
```
maoueh, line 129: I imagine that's the reason for the introduction of the intermediary `protobuf` module. Should be documented in the top of the macro (each of them if it affects them all).

runtime/derive/src/lib.rs
```diff
@@ -4,7 +4,103 @@ extern crate proc_macro;
...
+                let _comma: syn::Token![,] = content.parse()?;
+            }
+        }
+
+        Ok(TypeParamList(params))
+    }
+}
+
+//intended use is in build.rs with tonic_build's type_attribute(<...>, <...>) to generate type implementation of graph::runtime::ToAscObj
+//attributes: optional list of required fields '__required__{name:TypeName}' and enumerations field decraration with types, i.e. sum{single: ModeInfoSingle,multi: ModeInfoMulti}
```
maoueh, line 76: Do we still need support to manually pass required and enumeration explicitly?

runtime/derive/src/lib.rs
```diff
@@ -65,18 +179,68 @@ fn asc_type_derive_struct(item_struct: ItemStruct) -> TokenStream {
...
+    } else {
+        (quote! {}, quote! {})
+    };
 
     TokenStream::from(quote! {
         impl #impl_generics graph::runtime::AscType for #struct_name #ty_generics #where_clause {
             fn to_asc_bytes(&self) -> Result<Vec<u8>, graph::runtime::DeterministicHostError> {
+                #no_fields_return_bytes
+
                 let in_memory_byte_count = std::mem::size_of::<Self>();
```
maoueh, line 205: This gives the Rust `repr(C)` memory size in bytes. Now that we implement a proper padding, the later assertion below (need to expand context in PR to see the line in question):

```
assert_eq!(bytes.len(), in_memory_byte_count, "Alignment mismatch for {}, re-order fields or explicitly add a _padding field", stringify!(#struct_name));
```

Now that we deal with padding ourself, I'm pretty sure the assertion will triggers now in some occasions where padding is different between `repr(C)` and AssemblyScript rules.

We will need to see if Edge&Node here how we are going to ensure this is thoroughly tested because such change if we get it wrong can lead to invisible bugs where the data will be just wrong without any signs of crashes.

runtime/derive/src/lib.rs
```diff
@@ -65,18 +179,68 @@ fn asc_type_derive_struct(item_struct: ItemStruct) -> TokenStream {
...
-                // Right now we do not properly implement the alignment rules dicted by
-                // AssemblyScript. As such, we here enforce that the structure is tighly
-                // packed and that no implicit padding has been added.
-                //
+                let mut offset = 0;
+                // max field alignment will also be struct alignment which we need to pad the end
+                let mut max_align = 0;
+
+                #(
+                    let field_align = std::mem::align_of::<#field_types>();
```
maoueh, line 213: I'm a bit worried by this because this uses Rust rules to determine the alignment of the field and not AssemblyScript rules.

You would think they are fully equivalent but usage of `std::mem::size_of` above tends to show that it's untrue that those methods are fully equivalent to AssemblyScript rules in all occasions.

I did a quick test however based on https://github.com/AssemblyScript/assemblyscript/blob/67457f3824efe80cfcb34465bd901c10c2e0c15c/tests/compiler/builtins.ts#L424-L428 (seems to be log2 values) and all usage for base types seems to work properly: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=0ad7be7dfdafad100557b4fe41bcd80d.

Anyway, hopefully it works for every cases that we require, but that is not 100% sure to me.

maoueh (changes requested):
Can you also remove dead code that is commented but unsued.

My main concerns is the AssemblyScript padding that was introduced, just need to ensure we properly test it on existing subgraphs.

chain/cosmos/build.rs
```diff
@@ -1,7 +1,54 @@
...
+    let array_types = types
+        .iter()
+        .flat_map(|(_, t)| t.fields.iter())
+        .filter(|t| t.is_array)
+        .map(|t| t.type_name.clone())
+        .collect::<std::collections::HashSet<_>>();
+
+    let mut builder = tonic_build::configure().out_dir("src/protobuf");
+
+    for (name, ptype) in types {
```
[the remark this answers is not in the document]
soleinik-figment: yes, agree... for common code might be issues with proto file location as it's local to the crate. something to pay attention to

chain/cosmos/proto/firehose/annotations.proto
```diff
@@ -0,0 +1,9 @@
+package firehose;
```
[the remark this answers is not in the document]
soleinik-figment: Let's discuss

runtime/derive/src/generate_asc_type.rs
soleinik-figment: there is https://crates.io/crates/protobuf (parser) and crate::protobuf
https://crates.io/crates/protobuf - is to parse proto files and annotate prost structs with meta info... unless we are missing your point... let's discuss

runtime/derive/src/lib.rs
soleinik-figment: No, this is meta info derived from proto files

runtime/test/src/test_padding.rs
```diff
@@ -0,0 +1,420 @@
+use crate::protobuf;
+use graph::prelude::tokio;
+
+use self::data::BadFixed;
+
+const WASM_FILE_NAME: &str = "test_padding.wasm";
+const SUB_GRAPH_NAME: &str = "bogus_sub_graph_id";
```
evaporei, line 7: I think `subgraph` here can be together, without the `_` 🙂

runtime/test/src/test_padding.rs
```diff
@@ -0,0 +1,420 @@
...
+
+            bytes.extend_from_slice(&self.nonce.to_asc_bytes()?);
+            bytes.extend_from_slice(&self.str_suff.to_asc_bytes()?);
+            bytes.extend_from_slice(&self.tail.to_asc_bytes()?);
+
+            // assert_eq!(
+            //     bytes.len(),
+            //     in_memory_byte_count,
+            //     "Alignment mismatch for AscBad, re-order fields or explicitely add a _padding field",
+            // );
```
evaporei, line 86: Is this comment because that's what the test is testing?

evaporei (changes requested):
Congrats on making this one through 👏 
I just added a few comments on very minor things to adjust.
After those I will definitely merge this ❤️

evaporei (approved):

PR #3734, opened 2022-07-13
Filipe/select poi algo

mangas:


graph/src/data/subgraph/api_version.rs
```diff
@@ -50,7 +53,7 @@ impl UnifiedMappingApiVersion {
     ) -> Result<Self, DifferentMappingApiVersions> {
         let unique_versions: BTreeSet<Version> = versions.collect();
 
-        let all_below_referential_version = unique_versions.iter().all(|v| *v < API_VERSION_0_0_5);
+        let all_below_referential_version = unique_versions.iter().all(|v| *v < API_VERSION_0_0_6);
```
leoyvens, line 56: These checks should not have been changed from 0.0.5, that's why there are test failures.

leoyvens (changes requested):

store/postgres/src/deployment_store.rs
```diff
@@ -553,13 +557,21 @@ impl DeploymentStore {
             api_schema(&schema.document).map_err(|e| StoreError::Unknown(e.into()))?;
         schema.add_subgraph_id_directives(site.deployment.clone());
 
+        let poi_version = Version::from_str(&spec_version).map_err(anyhow::Error::from)?;
```
leoyvens, line 560:
```suggestion
        let spec_version = Version::from_str(&spec_version).map_err(anyhow::Error::from)?;
```

leoyvens (changes requested):

Issue #3060, opened 2021-12-13
Add `timestamp` to `_meta.block`


schmidsi:
**Do you want to request a *feature* or report a *bug*?**

A feature

**What is the current behavior?**

If a subgraph consumer wants to know the freshness of their subgraph, they can send a query like:

```gql
{
  _meta {
    block {
      number
    }
  }
}
```

Now they know until which block that subgraph is indexed, but without another source (JSON RPC endpoint like Infura or Alchemy) of what is the number of the latest block they do not know how fresh the subgraph is.

**What is the expected behavior?**

It would be much simpler to be also able to query for the timestamp of the block that a subgraph has indexed. Like:

```gql
{
  _meta {
    block {
      number
      timestamp
    }
  }
}
```

With a timestamp, they can directly calculate the freshness by comparing it to the system time.

azf20:
Discussed yesterday with @lutter and @leoyvens, this is not a quick win as it would require a call to get the timestamp from the block cache (currently the `_meta` field just uses the hash and number which are stored on the deployment)

leoyvens:
Actually in a call by block hash going to the block cache is always necessary anyways, so this wouldn't add overhead in that case which is the relevant one in the network.

schmidsi:
@azf20 Circling back on this. This was an issue yesterday as we did not receive new blocks from one of the chains. The front-ends do not know much about it unless they have a second source of block info to compare. Do you still think this is not a quick win?

mangas:
For Hash constraint: 
Currently it looks like for hash we only call `store.block_number` which returns `Option<BlockNumber>`, to introduce this we would need to either change the `BlockNumber` type or introduce a new trait function that either gets the entire Block or number+timestamp? 

The `ChainStore` trait already something similar, which I think prolly need a new type here in order to add more fields return (instead of adding to the tuple) 
```
 /// Find the block with `block_hash` and return the network name and number
   357     fn block_number(&self, hash: &BlockHash) -> Result<Option<(String, BlockNumber)>, StoreError>;
```
----------
For number constraint: 

It looks like we don't actually query anything, in this case the timestamp either cannot be provided or we'd need to add the call here as well.
```
 138             BlockConstraint::Number(number) => {
     1                 check_ptr(state, number)?;
     2                 // We don't have a way here to look the block hash up from
     3                 // the database, and even if we did, there is no guarantee
     4                 // that we have the block in our cache. We therefore
     5                 // always return an all zeroes hash when users specify
[...]

PR #3738, opened 2022-07-14
Add timestamp to meta block

mangas:
Fixes: https://github.com/graphprotocol/graph-node/issues/3060

This change should add a timestamp for EVM chains, should work well for both RPC and firehose ingestors. 
The fact the query depends heavily on the format of the block means it will not work for any block format that doesn't add the expected timestamp field inside a top level block field. 

To solve this further we could ensure that all block formats have the necessary timestamp or alternative we could extract/duplicate that field outside the block and run some migrations. The downside of the migration approach is that it needs to be communicated more widely. This migration would also be potentially long due to the size of the block cache tables.

### DIFF
--- a/graph/src/components/store/traits.rs
+++ b/graph/src/components/store/traits.rs
@@ -361,8 +361,14 @@ pub trait ChainStore: Send + Sync + 'static {
     /// may purge any other blocks with that number
     fn confirm_block_hash(&self, number: BlockNumber, hash: &BlockHash) -> Result<usize, Error>;
 
-    /// Find the block with `block_hash` and return the network name and number
-    fn block_number(&self, hash: &BlockHash) -> Result<Option<(String, BlockNumber)>, StoreError>;
+    /// Find the block with `block_hash` and return the network name, number and timestamp if present.
+    /// Currently, the timestamp is only returned if it's present in the top level block. This format is
+    /// depends on the chain and the implementation of Blockchain::Block for the specific chain.
+    /// eg: {"block": { "timestamp": 123123123 } }
+    fn block_number(
+        &self,
+        hash: &BlockHash,
+    ) -> Result<Option<(String, BlockNumber, Option<String>)>, StoreError>;
 
     /// Tries to retrieve all transactions receipts for a given block.
     async fn transaction_receipts_in_block(
--- a/graph/src/components/store/traits.rs
+++ b/graph/src/components/store/traits.rs
@@ -415,6 +415,14 @@ pub trait QueryStore: Send + Sync {
 
     fn block_number(&self, block_hash: &BlockHash) -> Result<Option<BlockNumber>, StoreError>;
 
+    /// Returns the blocknumber as well as the timestamp. Timestamp depends on the chain block type
+    /// and can have multiple formats, it can also not be prevent. For now this is only available
+    /// for EVM chains both firehose and rpc.
+    fn block_number_with_timestamp(
+        &self,
+        block_hash: &BlockHash,
+    ) -> Result<Option<(BlockNumber, Option<String>)>, StoreError>;
+
     fn wait_stats(&self) -> Result<PoolWaitStats, StoreError>;
 
     async fn has_deterministic_errors(&self, block: BlockNumber) -> Result<bool, StoreError>;
--- a/graphql/src/schema/meta.graphql
+++ b/graphql/src/schema/meta.graphql
@@ -55,6 +55,8 @@ type _Block_ {
   hash: Bytes
   "The block number"
   number: Int!
+  "Timestamp of the block if available, format depends on the chain" 
+  timestamp: String
 }
 
 enum _SubgraphErrorPolicy_ {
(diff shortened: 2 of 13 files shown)


store/postgres/src/chain_store.rs
```diff
@@ -593,30 +594,57 @@ mod data {
...
+                block: Block,
+            }
+
             let number = match self {
                 Storage::Shared => {
                     use public::ethereum_blocks as b;
 
                     b::table
-                        .select(b::number)
+                        .select((b::number, b::data))
```
lutter, line 615: Instead of getting the whole block from the db, we should just get the timestamp. In SQL, that would look like `select data->'block'->>'timestamp' from chain41.blocks limit 1;` I believe you can get that through Diesel by doing `select(b::number, sql("data->'block'->>'timestamp'"))` though it might require some `QueryableByName` struct.

There's a difference between the JSON structure of how a full and a light block is stored, so you'll have to use `coalesce(data->'block'->>'timestamp', data->>'timestamp')` (see the comment in `Storage::blocks`) 

A while ago we tried to make the chain store dumb wrt the JSON structure of `data`, but I don't see how we can maintain that for this without dragging tens of kB around for each lookup like this.

tilacog (approved):

leoyvens:
Has this been tested on a firehose block stream?

mangas:
@leoyvens on a firehose stream is the thing I was testing locally but the blocks look the same on the database. The firehose ingestor I didn't test because as far as I can tell it is not wired but didn't spend too much on that either

graphql/src/schema/meta.graphql
```diff
@@ -55,6 +55,8 @@ type _Block_ {
   hash: Bytes
   "The block number"
   number: Int!
+  "Timestamp of the block if available, format depends on the chain"
```
leoyvens, line 58: This query is deterministic, and we don't have good control of whether the timestamp is available.

An idea to better control this is: Make this field non-nullable, and return a new `QueryExecutionError` variant if it is requested but we can't produce it, and make sure that error returns `false` from `is_attestable`.

leoyvens (changes requested):

graphql/src/schema/meta.graphql
mangas: This query is already non deterministic because it will return different results for the same block depending on whether the request was done using hash or number. This behaviour we're introducing is consistent with the behaviour of hash. I'm not sure introducing an error for a nullable field is going to produce the UX since it's not trivial to understand the different without being familiar with the code at some level.

store/postgres/src/chain_store.rs
```diff
@@ -1690,12 +1698,15 @@ impl ChainStoreTrait for ChainStore {
             .confirm_block_hash(&conn, &self.chain, number, hash)
     }
 
-    fn block_number(&self, hash: &BlockHash) -> Result<Option<(String, BlockNumber)>, StoreError> {
+    fn block_number(
+        &self,
+        hash: &BlockHash,
+    ) -> Result<Option<(String, BlockNumber, Option<String>)>, StoreError> {
         let conn = self.get_conn()?;
```
leoyvens, line 1705: This would ideally be asyncified to use `with_conn`.

graph/src/components/store/traits.rs
```diff
@@ -355,8 +355,14 @@ pub trait ChainStore: Send + Sync + 'static {
...
-    /// Find the block with `block_hash` and return the network name and number
-    fn block_number(&self, hash: &BlockHash) -> Result<Option<(String, BlockNumber)>, StoreError>;
+    /// Find the block with `block_hash` and return the network name, number and timestamp if present.
+    /// Currently, the timestamp is only returned if it's present in the top level block. This format is
+    /// depends on the chain and the implementation of Blockchain::Block for the specific chain.
+    /// eg: {"block": { "timestamp": 123123123 } }
+    fn block_number(
+        &self,
+        hash: &BlockHash,
+    ) -> Result<Option<(String, BlockNumber, Option<String>)>, StoreError>;
```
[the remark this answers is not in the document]
mangas: I'm not sure this refactoring should go on the same PR though. I'm happy to make the changes and raise a new PR for it

graph/src/components/store/traits.rs
```diff
@@ -403,6 +409,14 @@ pub trait QueryStore: Send + Sync {
 
     fn block_number(&self, block_hash: &BlockHash) -> Result<Option<BlockNumber>, StoreError>;
 
+    /// Returns the blocknumber as well as the timestamp. Timestamp depends on the chain block type
+    /// and can have multiple formats, it can also not be prevent. For now this is only available
+    /// for EVM chains both firehose and rpc.
+    fn block_number_with_timestamp(
+        &self,
+        block_hash: &BlockHash,
+    ) -> Result<Option<(BlockNumber, Option<String>)>, StoreError>;
```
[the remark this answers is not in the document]
mangas: it's an package internal name, I went with I thought was an accurate description 😛 happy to rename if you have suggestions

store/postgres/src/chain_store.rs
mangas: this one is also used in a few places so I'd rather take that as a separate PR

leoyvens (approved):

PR #3739, opened 2022-07-14
Remove typename from POI

evaporei:
After an entity merge, if the Entity Cache was cleared, the `__typename`
field would come up again and polute mappings and by consequence, the
POI. This PR fixes the issue/bug alongside with it's respective (new) test.

The merging behavior is described [here](https://thegraph.com/docs/en/developer/assemblyscript-api/#updating-existing-entities) in the docs (second item).

```
### Updating existing entities
There are two ways to update an existing entity:

1. Load the entity with e.g.`Transfer.load(id)`, set properties on the entity, then `.save()` it back to the store.
2. Simply create the entity with e.g. `new Transfer(id)`, set properties on the entity, then `.save()` it to the store. If the entity already exists, the changes are merged into it.
```


leoyvens (approved):

leoyvens:
Nice work with the test!

leoyvens (approved):

PR #3742, opened 2022-07-17
Private data sources table

leoyvens:
This fills in the sql implementation in `dynds/private.rs`, and sets V1 as the deployment schema version for new deployments.

This maintains compatibility with the deployment schema V0, which uses the shared table. This does not implement migrations from V0 to V1, so copies are enabled for all versions.

The new stores template references by manifest index, but the shared table stores them by name and the runtime creates the by name, so a map between template name and index had to be plumbed into various places.

This was tested manually with uniswap-v3, and the `data_source_revert` end-to-end test was modified to additionally test restarting a subgraph with dynamic data sources.

Part of #3405, the only remaining work being migrations which we'll want to do eventually.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,7 +2,13 @@
 
 ## Unreleased
 
-- ...
+### New DB table for dynamic data sources
+
+For new subgraph deployments, dynamic data sources will be recorded under the `sgd*.data_sources$`
+table, rather than `subgraphs.ethereum_contract_data_source`. As a consequence
+new deployments will not work correctly on earlier graph node versions, so
+_downgrading to an earlier graph node version is not supported_.
+See issue #3405 for other details.
 
 ## 0.27.0
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,7 +5,7 @@
 ### New DB table for dynamic data sources
 
 For new subgraph deployments, dynamic data sources will be recorded under the `sgd*.data_sources$`
-table, rather than `subgraphs.ethereum_contract_data_source`. As a consequence
+table, rather than `subgraphs.dynamic_ethereum_contract_data_source`. As a consequence
 new deployments will not work correctly on earlier graph node versions, so
 _downgrading to an earlier graph node version is not supported_.
 See issue #3405 for other details.
--- a/graph/src/data/subgraph/mod.rs
+++ b/graph/src/data/subgraph/mod.rs
@@ -764,16 +764,27 @@ impl<C: Blockchain> UnresolvedSubgraphManifest<C> {
             ));
         }
 
+        let ds_count = data_sources.len();
+        if ds_count as u64 + templates.len() as u64 > u32::MAX as u64 {
+            return Err(anyhow!(
+                "Subgraph has too many declared data sources and templates",
+            ));
+        }
+
         let (schema, data_sources, templates, offchain_data_sources) = try_join4(
             schema.resolve(id.clone(), &resolver, logger),
             data_sources
                 .into_iter()
-                .map(|ds| ds.resolve(&resolver, logger))
+                .enumerate()
+                .map(|(idx, ds)| ds.resolve(&resolver, logger, idx as u32))
                 .collect::<FuturesOrdered<_>>()
                 .try_collect::<Vec<_>>(),
             templates
                 .into_iter()
-                .map(|template| template.resolve(&resolver, logger))
+                .enumerate()
+                .map(|(idx, template)| {
+                    template.resolve(&resolver, logger, ds_count as u32 + idx as u32)
+                })
                 .collect::<FuturesOrdered<_>>()
                 .try_collect::<Vec<_>>(),
             offchain_data_sources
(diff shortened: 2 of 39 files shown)


leoyvens:
Grafting from a V0 subgraph isn't working, I'll fix that.

graph/src/data/subgraph/mod.rs
```diff
@@ -727,16 +727,27 @@ impl<C: Blockchain> UnresolvedSubgraphManifest<C> {
             ));
         }
 
+        let ds_count = data_sources.len();
+        if ds_count as u64 + templates.len() as u64 > u32::MAX as u64 {
```
Theodus, line 731:
```suggestion
        let ds_count = data_sources.len() as u32;
        if ds_count.checked_add(templates.len() as u32).is_none() {
```

Theodus: nit: I find this a bit more clear.

Theodus (approved):
LGTM

graph/src/data/subgraph/mod.rs
Theodus: Yup, there is. You're initial solution is probably better.

NEWS.md
```diff
@@ -2,10 +2,21 @@
 
 ## Unreleased
 
+### Performance
+- Pipeline store writes #3084 #3177
+
+### New DB table for dynamic data sources
+
+For new subgraph deployments, dynamic data sources will be recorded under the `sgd*.data_sources$`
+table, rather than `subgraphs.ethereum_contract_data_source`. As a consequence
```
lutter, line 11: That should be `subgraphs.dynamic_ethereum_contract_data_source`

store/postgres/src/dynds/private.rs
```diff
@@ -96,10 +121,93 @@ impl DataSourcesTable {
...
+                .execute(conn)?;
+        }
+
+        Ok(inserted_total)
     }
 
     pub(crate) fn revert(&self, conn: &PgConnection, block: BlockNumber) -> Result<(), StoreError> {
-        todo!()
+        // Use `@>` to leverage the gist index.
+        // This assumes all ranges are of the form [x, +inf).
```
lutter, line 165: If block ranges are always `[x, +inf)`, it would be better to not use a range at all, and just have a `block$` that stores `x` since that could be supported by a normal BTree index. But if we do need the flexibility to disable data sources afte ra certain block, this is ok.

store/postgres/src/dynds/private.rs
```diff
@@ -96,10 +121,93 @@ impl DataSourcesTable {
...
+
+        let count = sql_query(&query)
+            .bind::<Integer, _>(target_block)
+            .execute(conn)?;
+
+        // Test that both tables have the same contents.
+        debug_assert!(
+            self.load(conn, target_block).map_err(|e| e.to_string())
+                == dst.load(conn, target_block).map_err(|e| e.to_string())
+        );
```
lutter, line 209: Nice!

lutter (approved):
LGTM! I have a few detail questions, but assuming they aren't a real issue this is good to merge.

store/postgres/src/dynds/private.rs
```diff
@@ -78,16 +78,41 @@ impl DataSourcesTable {
...
+            .load::<Tuple>(conn)?;
+
+        Ok(tuples
+            .into_iter()
+            .map(|(block_range, manifest_idx, param, context)| {
+                let creation_block = match block_range.0 {
+                    Bound::Included(block) => Some(block),
+
+                    // Should never happen.
+                    Bound::Excluded(_) | Bound::Unbounded => None,
```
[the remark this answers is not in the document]
leoyvens: Indeed

NEWS.md
leoyvens: Thanks!

store/postgres/src/dynds/private.rs
leoyvens: Yes this is trying to be future compatible with removing data sources by closing their range.

store/postgres/src/dynds/private.rs
```diff
@@ -96,10 +121,93 @@ impl DataSourcesTable {
...
+        let count = dst.table.clone().count().get_result::<i64>(conn)?;
+        if count > 0 {
+            return Ok(count as usize);
+        }
+
+        // Assumes all ranges are of the form `[n, +inf)`.
+        let query = format!(
+            "\
+            insert into {dst}(block_range, causality_region, manifest_idx, parent, id, param, context)
+            select int4range(lower(e.block_range), null), e.causality_region, e.manifest_idx,
```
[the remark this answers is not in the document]
leoyvens: The desired behaviour for this to be future compatible with closed ranges would be to literally copy the range `[x, y]` if `y <= target_block`, but if `y > target_block` then we'd open the range with `[x, null)`. Do you know what would be the right SQL to achieve this?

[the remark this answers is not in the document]
lutter: You will have fun with [`case` expressions](https://www.postgresql.org/docs/current/functions-conditional.html), something like 
```sql
case
  when upper(e.block_range) <= $target_block then e.block_range
  else int4range(lower(e.block_range), null)
end
```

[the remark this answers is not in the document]
leoyvens: That seems to work, thanks!

PR #3749, opened 2022-07-19
Stepd/fix poi cursor on deterministic error

sduchesneau:
(depends on https://github.com/graphprotocol/graph-node/pull/3748) 

Fixes the following scenario:
* A deterministic error happens at block 15
* The store updates its current head to be block=15, cursor={15}
* The deployment stops with "Subgraph Error"

When we restart the graph-node instance, it will try again:
* it detects that current head block 15 has errors
* It reverts the block 15 (down to 14...)
* It starts the stream again with start_block=15 (but also, with cursor={15})
* firehose skips the block 15 because that's how firehose cursors work (exclusive...)
* firehose sends block 16... the subgraph_deployment now has invalid data !

This patch prevents updating the cursor in the store when a deterministic error happened, so that it keeps the previous cursor. It relies on the fact that the next time that this subgraph is started, the problematic block will be reverted, so the cursor will be valid at that moment.

@leoyvens do you agree with this approach ? 
(ping @maoueh)

### DIFF
--- a/core/src/subgraph/runner.rs
+++ b/core/src/subgraph/runner.rs
@@ -81,7 +81,8 @@ where
                 let _outcome = self
                     .inputs
                     .store
-                    .unfail_deterministic_error(&current_ptr, &parent_ptr).await?;
+                    .unfail_deterministic_error(&current_ptr, &parent_ptr)
+                    .await?;
             }
         }
 
--- a/core/src/subgraph/runner.rs
+++ b/core/src/subgraph/runner.rs
@@ -81,7 +81,7 @@ where
                 let _outcome = self
                     .inputs
                     .store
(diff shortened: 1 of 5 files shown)


leoyvens:
This bit of code seems to be already trying to address this scenario
https://github.com/graphprotocol/graph-node/blob/15493e9960c5400d0d155f3b8226b5b5ef80042e/store/postgres/src/deployment_store.rs#L1379-L1384

So could the issue be that the block stream is not reset after an unfail, here
https://github.com/graphprotocol/graph-node/blob/15493e9960c5400d0d155f3b8226b5b5ef80042e/core/src/subgraph/runner.rs#L660

leoyvens (approved):

PR #3759, opened 2022-07-21
graphql: cache GraphQL validation

kamilkisiela:
I ran the same query ~55 times and got these results:

```
No caching (58 samples)

p95:      47.75µs
p99:      73.83µs
average:  24.85µs
median:   20.13µs

Caching (56 samples)
    
p95:       4.63µs
p99:       4.92µs
average:   2.08µs
median:    1.67µs
```

Part of https://github.com/graphprotocol/graph-node/issues/3077

### DIFF
--- a/graphql/src/execution/query.rs
+++ b/graphql/src/execution/query.rs
@@ -5,10 +5,11 @@ use graphql_tools::validation::rules::*;
 use graphql_tools::validation::utils::ValidationError;
 use graphql_tools::validation::validate::{validate, ValidationPlan};
 use lazy_static::lazy_static;
+use parking_lot::Mutex;
 use std::collections::{BTreeMap, HashMap, HashSet};
 use std::hash::{Hash, Hasher};
 use std::iter::FromIterator;
-use std::sync::{Arc, Mutex, PoisonError};
+use std::sync::Arc;
 use std::time::Instant;
 use std::{collections::hash_map::DefaultHasher, convert::TryFrom};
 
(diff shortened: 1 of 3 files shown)


graphql/src/execution/query.rs
```diff
@@ -150,8 +156,16 @@ impl Query {
...
+        let mut validation_cache = GRAPHQL_VALIDATION_CACHE
+            .lock()
+            .unwrap_or_else(PoisonError::into_inner);
+        let validation_errors = validation_cache.entry(query.shape_hash).or_insert_with(|| {
+            validate(
+                &schema.document(),
+                &query.document,
+                &GRAPHQL_VALIDATION_PLAN,
+            )
+        });
```
leoyvens, line 168: It seems the lock is held for a larger scope than necessary, maybe you can do it all in one line:
```suggestion
        GRAPHQL_VALIDATION_CACHE
            .lock()
            .unwrap_or_else(PoisonError::into_inner)
            .entry(query.shape_hash).or_insert_with(|| {
            validate(
                &schema.document(),
                &query.document,
                &GRAPHQL_VALIDATION_PLAN,
            )
        });
```

leoyvens (changes requested):

kamilkisiela:
Applied suggested changes, ran `cargo fmt` and squashed commits

leoyvens:
Still doesn't build.

graphql/src/execution/query.rs
kamilkisiela: @leoyvens not sure how to do it in one line and assign the result to `let validation_errors`

leoyvens:
Lock lifetimes are tricky. I've built on your changes so that the lock is only held when reading or inserting to the cache, but importantly not held when actually validating the query. As you had noticed this requires a bit of cloning.

leoyvens:
@kamilkisiela could you review and re-run the benchmarks?

kamilkisiela:
@leoyvens thank you. I re-run the benchmarks and it looks faster now.

leoyvens (approved):

PR #3766, opened 2022-07-22
Filipe/substreams option runtime

mangas:


leoyvens:
Why isn't it sufficient to use a fake `RuntimeHostBuilder`?

leoyvens:
Hmm I see that replacing the `RuntimeHostBuilder` isn't so simple.

mangas:
> Hmm I see that replacing the `RuntimeHostBuilder` isn't so simple.

This way is also more explicit and requires less "hacks" IMO

core/src/subgraph/instance.rs
```diff
@@ -90,7 +94,12 @@ where
...
         host_metrics: Arc<HostMetrics>,
-    ) -> Result<T::Host, Error> {
+    ) -> Result<Option<T::Host>, Error> {
+        let host_builder = match self.host_builder {
+            None => return Ok(None),
+            Some(ref host_builder) => host_builder,
+        };
+
         let mapping_request_sender = {
             let module_bytes = data_source.runtime();
```
leoyvens, line 104: What about making this controlled by the data source, so we change `DataSource::runtime` to return an option, and if it returns none then no host is created.

leoyvens (changes requested):

runtime/wasm/src/host.rs
```diff
@@ -53,7 +53,7 @@ impl<C: Blockchain> RuntimeHostBuilderTrait<C> for RuntimeHostBuilder<C> {
     type Req = MappingRequest<C>;
 
     fn spawn_mapping(
-        raw_module: Vec<u8>,
+        raw_module: &Vec<u8>,
```
leoyvens, line 56:
```suggestion
        raw_module: &[u8],
```

runtime/wasm/src/mapping.rs
```diff
@@ -13,15 +13,15 @@ use std::thread;
 
 /// Spawn a wasm module in its own thread.
 pub fn spawn_module<C: Blockchain>(
-    raw_module: Vec<u8>,
+    raw_module: &Vec<u8>,
```
leoyvens, line 16:
```suggestion
    raw_module: &[u8],
```

leoyvens (changes requested):

leoyvens (approved):

Issue #3732, opened 2022-07-13
ethereum_chain_head_number is not tracking as expected for firehose-based chains


azf20:
From the Hosted Service monitoring, it appears that this metric is not showing up as expected. Despite the metric name, it _should_ work for Firehose-based, non EVM chains.

azf20:
Note from @lutter:
```
The metric is updated by the chain head listener. That listens to updates the block ingestor sends out and is only used for non-firehose networks. For firehose network, that metric is never updated.
```

azf20:
This was fixed

PR #3771, opened 2022-07-25
ethereum_chain_head_number is not tracking as expected for firehose-based chains #3732

soleinik-figment:


maoueh:
I'm not sure why the other files were changed in this PR, I assume it might be leftover from switching branches. I think it should be remove from this PR.

As for the PR itself, seems all good to me, approving.

maoueh (approved):

store/postgres/src/chain_store.rs
```diff
@@ -1556,7 +1556,10 @@ impl ChainStoreTrait for ChainStore {
 
         let ptr = block.ptr();
         let hash = ptr.hash_hex();
-        let number = ptr.number as i64;
+        let number = ptr.number as i64; //block height
+
+        //this will send an update via prostgres, channel: chain_head_updates
```
leoyvens, line 1561:
```suggestion
        //this will send an update via postgres, channel: chain_head_updates
```

leoyvens (approved):

PR #3781, opened 2022-07-27
Allow a `number_gte` constraint before the subgraph start/earliest block

lutter:


graphql/src/store/resolver.rs
```diff
@@ -145,9 +145,19 @@ impl StoreResolver {
                 // See 7a7b9708-adb7-4fc2-acec-88680cb07ec1
                 Ok(BlockPtr::from((web3::types::H256::zero(), number as u64)))
             }
-            BlockConstraint::Min(number) => {
-                check_ptr(state, number)?;
-                Ok(state.latest_block.cheap_clone())
+            BlockConstraint::Min(min) => {
+                let ptr = state.latest_block.cheap_clone();
+                if ptr.number < min {
+                    return Err(QueryExecutionError::ValueParseError(
```
leoyvens, line 151: `ValueParseError` doesn't seem like the most suitable variant.

leoyvens (approved):

graphql/src/store/resolver.rs
lutter: But it's always been a `ValueParseError` ...

leoyvens: Just a passing thought, no in-depth opinion here

PR #3782, opened 2022-07-27
graphql: Reenable validations by default

leoyvens:


leoyvens:
Still questions around the caching by query shape.

PR #3787, opened 2022-07-29
substreams trigger processor

mangas:


chain/substreams/src/trigger.rs
```diff
@@ -85,3 +100,133 @@ impl blockchain::TriggersAdapter<Chain> for TriggersAdapter {
...
+                        causality_region,
+                        logger,
+                    );
+
+                    state.entity_cache.set(key, Entity::from(data))?;
+                }
+                Operation::Delete => {
+                    let key = EntityKey {
+                        subgraph_id: self.locator.hash.clone(),
+                        entity_type: EntityType::new("some-type".into()),
```
maoueh, line 221: Are missing a real value here? If unused, I would add a comment explaining why we do this.

chain/substreams/src/trigger.rs
```diff
@@ -85,3 +100,133 @@ impl blockchain::TriggersAdapter<Chain> for TriggersAdapter {
...
+        for entity_change in block.entities_changes.entity_changes.iter() {
+            match entity_change.operation() {
+                Operation::Unset => {
+                    // Potentially an issue with the server side or
+                    // we are running an outdated version. In either case we should abort.
+                    return Err(MappingError::Unknown(anyhow!("Detected UNSET entity operation, either a server error or there's a new type of operation and we're running an outdated protobuf")));
+                }
+                Operation::Create | Operation::Update => {
+                    let entity_type: &str = &entity_change.entity;
+                    let entity_id: String = String::from_utf8(entity_change.id.clone())
```
maoueh, line 152: I don't see why `clone` would be required here.

chain/substreams/src/trigger.rs
```diff
@@ -85,3 +100,133 @@ impl blockchain::TriggersAdapter<Chain> for TriggersAdapter {
...
+                                Value::Int(i32::from_be_bytes(bytes))
+                            }
+                            (Type::Bytes, false) => {
+                                Value::Bytes(Bytes::from(field.new_value.as_ref()))
+                            }
+                            (Type::String, false) => Value::String(
+                                String::from_utf8(field.new_value.clone())
+                                    .map_err(|e| MappingError::Unknown(anyhow::Error::from(e)))?,
+                            ),
+                            _ => {
```
maoueh, line 194: Aren't we missing the handling of `(_, true)` case which means the field is not set? Or maybe we are already using field presence via `optional` keyword and this is waiting to be removed from the proto?

maoueh: I see we talk about `null` value in the default case below, are we not dealing with `null` value to "unset" a field?

maoueh (changes requested):
Requesting changes but it's really more question/interrogations.

chain/substreams/src/trigger.rs
mangas: entity_change is a reference and so the vec can't be moved.

chain/substreams/src/trigger.rs
mangas: Yep

chain/substreams/src/trigger.rs
maoueh: But `String::from_utf8` is accepting a reference of bytes `&[u8]` creating a new `String` instance, I don't see `String::from_utf8` owning the buffer so I don't see why the `clone` is necessary.

maoueh: Scratch all that! I was looking at `str::from_utf8` sorry.

maoueh (approved):

PR #3788, opened 2022-07-29
graphql: Add deterministic error test

leoyvens:


graphql/tests/query.rs
```diff
@@ -1755,3 +1755,43 @@ fn can_query_root_typename() {
...
+
+    run_test_sequentially(|store| async move {
+        let deployment = setup(
+            store.as_ref(),
+            "testDeterministicError",
+            BTreeSet::new(),
+            IdType::String,
+        )
+        .await;
+
```
evaporei, line 1772: Not necessary for your PR, but I would like for our tests to populate `latest_ethereum_block_{number,hash}` before failures, like this:

```rust
        test_store::transact_and_wait(
            &STORE.subgraph_store(),
            &deployment,
            BLOCKS[0].clone(),
            vec[]!,
        )
        .await
        .unwrap();
```

evaporei (approved):

graphql/tests/query.rs
leoyvens: Yhea these tests are a bit strange in their setup but I'll leave that for another time.

PR #3790, opened 2022-07-29
test(store): Fix race condition in graft test

leoyvens:
This just might fix the flakiness we've seen in these tests.


lutter:
Nice! Let's hope that does the trick

leoyvens:
@lutter from seeing this reproduce again in more recent CI runs, I believe it didn't 😞

PR #3791, opened 2022-07-29
Add runtime support for offchain data sources & templates

Theodus:
This implements most of the runtime support for handling offchain data sources. Unfortunately there's a lot of boilerplate here, but I think it's best for a first pass since my initial approach at something more cohesive became a massive time sink.


graph/src/blockchain/mod.rs
```diff
@@ -81,22 +81,23 @@ pub trait Block: Send + Sync {
...
-    type UnresolvedDataSourceTemplate: UnresolvedDataSourceTemplate<Self>;
+    type UnresolvedDataSourceTemplate: UnresolvedDataSourceTemplate<Self> + Clone;
 
     /// Trigger data as parsed from the triggers adapter.
     type TriggerData: TriggerData + Ord + Send + Sync + Debug;
 
     /// Decoded trigger ready to be processed by the mapping.
     /// New implementations should have this be the same as `TriggerData`.
-    type MappingTrigger: MappingTrigger + Debug;
+    type MappingTrigger: Send + Sync + Debug;
```
leoyvens, line 100: Should we just move the `ToAscPtr` bound here, for convenience.

core/src/subgraph/runner.rs
```diff
@@ -556,6 +574,81 @@ where
...
+            source: source.clone(),
+            data: Arc::new(data),
+        });
+
+        let block_state = self
+            .ctx
+            .instance
+            .process_trigger(
+                &self.logger,
+                &Arc::default(),
```
leoyvens, line 627: This is not great, though it does seem to meet our initial requirements. Worth a comment that this will need further refactoring at some point.

core/src/subgraph/instance.rs
```diff
@@ -78,8 +78,8 @@ where
 
             // Create services for static offchain data sources
             if let DataSource::Offchain(ds) = &ds {
-                if let Some(source) = &ds.source {
-                    offchain_monitor.monitor(source.clone());
+                if ds.source.is_some() {
```
leoyvens, line 81: Why this `is_some`, shouldn't we let it error if the ds is somehow missing a source?

core/src/subgraph/runner.rs
```diff
@@ -561,6 +574,81 @@ where
...
+            BlockState::<C>::new(self.inputs.store.cheap_clone(), LfuCache::new());
+        let entity_cache = std::mem::take(&mut self.state.entity_lfu_cache);
+
+        self.persist_dynamic_data_sources(
+            &mut block_state.entity_cache,
+            matching_data_sources
+                .cloned()
+                .map(data_source::DataSource::Offchain)
+                .collect(),
+        );
```
leoyvens, line 612: The handling of `persist_dynamic_data_sources` confuses me, why in the main loop we only filter for onchain data sources? And why is this being called here at all?

leoyvens:
Great stuff!

One thing I hadn't previously considered is handling reverts. In the DB we get that for free, but we need to also revert the in-memory state. It might be that this https://github.com/graphprotocol/graph-node/blob/c75e0d7570d33b6669cd52c989b476d190a24013/core/src/subgraph/instance.rs#L200 just happens to be sufficent, but do we additionally need any revert handling in the offchain monitor?

leoyvens:
This PR is now sufficient to to merge for internal testing.

There is a severe though infrequent known bug, which is that if a file data source is created on block B1, but the files are only found on a later block B2 and B2 is reverted, then the entities will be gone forever. The intended fix is to write the entities for file data sources on the creation block rather than the current block, this way they will only be reverted if the data source also is. This will be the first follow up to this PR in case it isn't done in time for merging.

Theodus (commented):
LGTM, thanks! Forbidding static file data sources for now seems like a sensible decision.

store/postgres/src/dynds/private.rs
```diff
@@ -220,4 +240,55 @@ impl DataSourcesTable {
...
+                .bind::<Nullable<Binary>, _>(param.as_ref().map(|p| &**p))
+                .bind::<Nullable<Jsonb>, _>(context)
+                .bind::<Nullable<Integer>, _>(creation_block)
+                .execute(conn)?;
+
+            if count > 1 {
+                // Data source deduplication enforces this invariant.
+                // See also: data-source-is-duplicate-of
+                return Err(constraint_violation!(
+                    "expected to remove at most one offchain data source but removed {}, ds: {:?}",
```
lutter, line 285: Since this will abort the txn, I would phrase the error as `expected to remove at most one offchain data source but would have removed ...` to make it a little less scary (nothing was lost in this case)

store/postgres/src/dynds/private.rs
```diff
@@ -150,19 +157,32 @@ impl DataSourcesTable {
                 ));
             }
 
-            let query = format!(
-                "insert into {}(block_range, manifest_idx, causality_region, param, context) \
-                 values (int4range($1, null), $2, $3, $4, $5)",
-                self.qname
-            );
+            // Onchain data sources have the causality region explicitly set to 0, while offchain
+            // data sources have an unique causality region assigned from the sequence.
```
lutter, line 161: This comment would match the code better if was something like
```rust
// Offchain data sources have a unique causality region assigned from a sequence in the database,
// while onchain data sources always have causality region 0
```

store/postgres/src/dynds/private.rs
```diff
@@ -95,26 +96,34 @@ impl DataSourcesTable {
...
-                    creation_block,
-                }
-            })
+            .map(
+                |(block_range, manifest_idx, param, context, causality_region)| {
+                    let creation_block = match block_range.0 {
+                        Bound::Included(block) => Some(block),
+
+                        // Should never happen.
+                        Bound::Excluded(_) | Bound::Unbounded => {
```
lutter, line 112: Since we're dealing with integers, wouldn't `Bound::Excluded(block)` be the same as `Bound::Included(block+1)`? It probably won't happen but isn't really fatal.

lutter (commented):
I only looked at `dynds/private.rs`, but that all makes sense to me.

store/postgres/src/dynds/private.rs
leoyvens: Well noted!

store/postgres/src/dynds/private.rs
leoyvens: Well we don't use or plan to use excluded initial bounds, so we can't know what they would mean. And our usage of `lower(block_range)` assumes an included bound, so it's best to blow up if an excluded bound is ever found.

leoyvens (approved):

leoyvens:
Integration cluster shows no divergences, merging!

PR #3810, opened 2022-08-03
fix(firehose): Add a firehose `connect_timeout`

leoyvens:


sduchesneau (approved):
LGTM, it correctly times out after 10s in my local tests and tries reconnecting.

PR #3813, opened 2022-08-04
wire substreams

mangas:
Enable substreams trigger processor and manifest handling

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1619,6 +1619,7 @@ dependencies = [
  "graph-chain-cosmos",
  "graph-chain-ethereum",
  "graph-chain-near",
+ "graph-chain-substreams",
  "graph-mock",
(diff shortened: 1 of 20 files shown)


maoueh (approved):
Only one question more as a curiosity, why now using `Box<dyn TrigggerProcessor>` instead of the generic parameter?

mangas:
> Only one question more as a curiosity, why now using `Box<dyn TrigggerProcessor>` instead of the generic parameter?

I discussed this with @leoyvens, my first approach was also to use the generic but it ends up creeping up to the Blockchain which would then also depend on RuntimeHostBuilder or we end up having to introduce a new type to build the trigger processor. Which is also not great. We think the performance implication of the dynamic dispatch will make up for the simpler code.

leoyvens (approved):

PR #3814, opened 2022-08-04
move debug stage above production

computeronix:
suggestion: move graph-node-debug stage above the production stage graph-node

For those that don't use --target then they will always get the production image by default

This also helps for those building from Docker Hub Cloud where a target can't be specified


lutter (approved):
Sorry, this fell through the cracks, but looks good to me. Thanks for the patch!

PR #3816, opened 2022-08-04
Fix provider selection when limit is set

lutter:
Provider selection with a limit did not work as intended because the node name was not passed around correctly, so that rules to set the limit were always evaluated for the node `default` (commit 209786c9 fixes that)

To make configuring limits easier and provide more insight into provider selection, this PR also adds a command `graphman config provider` that makes it possible to see which providers are eligible. Use it as `graphman --node-id index_node_0 --config /some/where/graph-node.toml config provider mainnet`

### DIFF
--- a/chain/ethereum/src/network.rs
+++ b/chain/ethereum/src/network.rs
@@ -28,23 +28,30 @@ pub struct EthereumNetworkAdapters {
 }
 
 impl EthereumNetworkAdapters {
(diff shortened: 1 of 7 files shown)


evaporei (approved):
LGTM :shipit:

dizzyd:
We'll merge this in next week after file data sources table has had a chance to bake.

PR #3818, opened 2022-08-05
fix(firehose): http2_adaptive_window

leoyvens:
We currently multiplex all firehose connections on a single http2 connection, and under load we're seeing possible flow control issues, possibly a [hyper bug](https://github.com/hyperium/hyper/issues/2899). The default per-connection http2 window is small and equal to the per-stream window. <strike>Enabling adaptive windows should give us a much bigger connection window, and possibly solve the issue at the current loads</strike>. This sets the connection window to the maximum value, effectively disabling it.

### DIFF
--- a/graph/src/firehose/endpoints.rs
+++ b/graph/src/firehose/endpoints.rs
@@ -56,7 +56,9 @@ impl FirehoseEndpoint {
                 .expect("TLS config on this host is invalid"),
             _ => panic!("invalid uri scheme for firehose endpoint"),
         }
-        .connect_timeout(Duration::from_secs(10));
+        .connect_timeout(Duration::from_secs(10))
+        .http2_keep_alive_interval(Duration::from_secs(30))
+        .http2_adaptive_window(true);
 
         let uri = endpoint.uri().to_string();
         //connect_lazy() used to return Result, but not anymore, that makes sence since Channel is not used immediatelly
--- a/graph/src/firehose/endpoints.rs
+++ b/graph/src/firehose/endpoints.rs
@@ -49,16 +49,28 @@ impl FirehoseEndpoint {
             .parse::<Uri>()
             .expect("the url should have been validated by now, so it is a valid Uri");
 
-        let endpoint = match uri.scheme().unwrap_or(&Scheme::HTTP).as_str() {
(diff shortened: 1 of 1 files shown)


maoueh (approved):

graph/src/firehose/endpoints.rs
```diff
@@ -56,7 +56,9 @@ impl FirehoseEndpoint {
                 .expect("TLS config on this host is invalid"),
             _ => panic!("invalid uri scheme for firehose endpoint"),
         }
-        .connect_timeout(Duration::from_secs(10));
+        .connect_timeout(Duration::from_secs(10))
+        .http2_keep_alive_interval(Duration::from_secs(30))
+        .http2_adaptive_window(true);
```
sduchesneau, line 61: This causes disconnections when connecting to firehose through a GCP Load Balancer. (`error reading a body from connection: unexpected end of file`)

graph/src/firehose/endpoints.rs
```diff
@@ -56,7 +56,9 @@ impl FirehoseEndpoint {
                 .expect("TLS config on this host is invalid"),
             _ => panic!("invalid uri scheme for firehose endpoint"),
         }
-        .connect_timeout(Duration::from_secs(10));
+        .connect_timeout(Duration::from_secs(10))
+        .http2_keep_alive_interval(Duration::from_secs(30))
```
sduchesneau, line 60: This is too low IMHO. While I didn't experience disconnections with this setting, some backends are very restrictive on this (at 10secs, it DOES generate similar disconnections as with http2_adaptive_window)

graph/src/firehose/endpoints.rs
leoyvens: Thanks for testing that! I've removed this, and took a different approach of setting a very large window size.

graph/src/firehose/endpoints.rs
leoyvens: I've removed the keep alive for now, to focus this PR on the stalling bug.

leoyvens:
This now sets the connection window to the maximum value, effectively disabling it. I was able to locally reproduce that this fixes the block stream stalling issue. I'm also convinced this is not a hyper bug but is http2 flow control working as designed, for our use case we should opt out of the connection level flow control.

PR #3833, opened 2022-08-09
Firehose connection pool

leoyvens:
This adds a `conn_pool_size` field to the toml configuration of firehose providers, which enables load balancing across multiple connections to the provider's endpoint. It defaults to 10.

Also does some refactoring around `FirehoseEndpoint::new`.


sduchesneau (approved):
great

sduchesneau:
Just need to check the tests that are hanging though!

leoyvens:
oops, fixed test

PR #3876, opened 2022-08-25
Feature/substreams v2 blocks

Eduard-Voiculescu:


chain/ethereum/proto/codec.proto
```diff
@@ -130,19 +192,19 @@ message TransactionTrace {
     // All transactions that ever existed prior Berlin fork before EIP-2718 was implemented.
     TRX_TYPE_LEGACY = 0;
 
-    // Transaction that specify an access list of contract/storage_keys that is going to be used
+    // Transaction that specicy an access list of contract/storage_keys that is going to be used
```
mangas, line 195:
```suggestion
    // Field that specifies an access list of contract/storage_keys that is going to be used
```

chain/ethereum/proto/codec.proto
```diff
@@ -130,19 +192,19 @@ message TransactionTrace {
...
     // in this transaction.
     //
     // Added in Berlin fork (EIP-2930).
     TRX_TYPE_ACCESS_LIST = 1;
 
-    // Transaction that specifies an access list just like TRX_TYPE_ACCESS_LIST but in addition defines the
+    // Transaction that specifis an access list just like TRX_TYPE_ACCESS_LIST but in addition defines the
```
mangas, line 201:
```suggestion
    // Transaction that specifies an access list just like TRX_TYPE_ACCESS_LIST but in addition defines the
```

mangas (changes requested):
fix typos, otherwise LGTM

PR #3877, opened 2022-08-25
firehose: Add timeout to `stream_blocks` request

leoyvens:
We have a timeout configured on the tonic endpoint which is catching most but somehow not all hangs.


maoueh:
Seems good, and we should see the connection timing out due to logging of `Err()`.

maoueh (approved):

PR #3896, opened 2022-08-31
Adding tests for decoding of gRPC server response from substreams

Eduard-Voiculescu:


chain/substreams/src/trigger.rs
```diff
@@ -271,3 +237,208 @@ where
         Ok(state)
     }
 }
+
+fn decode_entity_change(field: &Field, entity: &String) -> Result<Value, MappingError> {
+    return match field.value_type() {
```
mangas, line 242: you don't need this return

chain/substreams/src/trigger.rs
```diff
@@ -271,3 +237,208 @@ where
...
+            let mut bytes: [u8; 8] = [0; 8];
+            bytes.copy_from_slice(field.new_value.as_ref());
+            Ok(Value::Int(i64::from_be_bytes(bytes) as i32))
+        }
+        Type::Bytes => Ok(Value::Bytes(Bytes::from(field.new_value.as_ref()))),
+        Type::String => Ok(Value::String(
+            String::from_utf8(field.new_value.clone())
+                .map_err(|e| MappingError::Unknown(anyhow::Error::from(e)))?,
+        )),
+    };
```
mangas, line 270: remove this ;

chain/substreams/src/trigger.rs
```diff
@@ -271,3 +237,208 @@ where
...
+        for case in cases.into_iter() {
+            let value: Value = decode_entity_change(&case.field, &case.entity).unwrap();
+            let mut actual_value: String = String::new();
+            match case.field.value_type {
+                1 => actual_value = value.as_big_decimal().unwrap().to_string(),
+                2 => actual_value = value.as_bigint().unwrap().to_string(),
+                3 => actual_value = value.as_int().unwrap().to_string(),
+                4 => actual_value = value.as_bytes().unwrap().to_string(),
+                5 => actual_value = value.as_string().unwrap(),
+                _ => {}
```
mangas, line 439: replace {} with assert!(false)

chain/substreams/src/trigger.rs
```diff
@@ -271,3 +237,208 @@ where
...
+                },
+                entity: "Block".to_string(),
+                expected_new_value: "97892.6".to_string(),
+            },
+        ];
+
+        for case in cases.into_iter() {
+            let value: Value = decode_entity_change(&case.field, &case.entity).unwrap();
+            let mut actual_value: String = String::new();
+            match case.field.value_type {
```
mangas, line 433: you can use let actual_value = match ... and this way you don't need to define it as mut etc

mangas (approved):

PR #3897, opened 2022-08-31
add substreams config provider

mangas:
Adds substreams as a provider type rather than a protocol.

### DIFF
--- a/node/src/main.rs
+++ b/node/src/main.rs
@@ -4,9 +4,7 @@ use ethereum::{
 };
 use git_testament::{git_testament, render_testament};
 use graph::blockchain::firehose_block_ingestor::FirehoseBlockIngestor;
-use graph::blockchain::{
-    Block as BlockchainBlock, Blockchain, BlockchainKind, BlockchainMap, ChainIdentifier,
-};
+use graph::blockchain::{Block as BlockchainBlock, Blockchain, BlockchainKind, BlockchainMap};
 use graph::components::store::BlockStore;
 use graph::data::graphql::effort::LoadManager;
 use graph::env::EnvVars;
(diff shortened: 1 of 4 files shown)


evaporei:
LGTM, is anything missing here?

matthewdarwin:
See discussion here: https://forum.thegraph.com/t/substreams-into-subgraphs-a-simple-integration/3542

node/src/main.rs
```diff
@@ -236,6 +234,14 @@ async fn main() {
         create_firehose_networks(logger.clone(), &config)
     };
 
+    let substreams_networks_by_kind = if query_only {
+        BTreeMap::new()
+    } else {
+        create_substreams_networks(logger.clone(), &config)
+    };
+
+    println!("#### {:?}", substreams_networks_by_kind);
```
evaporei, line 243: Is this intentional?

mangas: nope 😛

evaporei (approved):

node/src/main.rs
```diff
@@ -334,10 +319,12 @@ async fn main() {
             node_id.clone(),
             metrics_registry.clone(),
             firehose_networks_by_kind.get(&BlockchainKind::Ethereum),
+            substreams_networks_by_kind.get(&BlockchainKind::Ethereum),
```
maoueh, line 322: This will start to be interesting because this force a `substreams` right not be for a "known" network. `substreams` brings the possibility to support an infinite number of chain as long as an `substreams` endpoint exist. 

Going full free form raises interesting question though. I think it would be a good subject for Wednesday weekly call.

mangas: For querying there are some aspects like block resolution that require the block cache to be populated. This is why we've gone this way. By being an endpoint type, we can require the chain to be ingested and thus we can rely on the ChainStore for block resolution and "close to head" status

mangas: Otherwise we'd need a "substreams" ingestor if that makes sense

Issue #3665, opened 2022-06-15
Add a `graphman prune` command


lutter:
The command will have the form `graphman prune <subgraph> <offset>` and remove all entity versions that were deleted or updated at least `offset` many blocks before the current subgraph head; for a pruned subgraph, the index-node API will report that block number as the earliest bock, and queries at a block height before that block will fail with an error. Pruning will not affect query results for queries at a block after the pruned block, and therefore simply limits how far back time-travel queries can reach.

Since pruning removes a huge amount of data that is usually not accessed by queries, it speeds up queries significantly.

As part of this issue, pruning will be a one-time action, i.e., it only removes history at the point in time when it is run.

azf20:
@lutter on the "earliest block", I wonder if that might break some application's assumptions (which might rely on that for example to identify how synced a subgraph is)

lutter:
> @lutter on the "earliest block", I wonder if that might break some application's assumptions (which might rely on that for example to identify how synced a subgraph is)

The earliest block number would still be accurate, it's just that the hash for the earliest block gets filled with a dummy value. Right now, that value is always the start block for the subgraph; with pruning the block number could/would move up over time. But I have no idea if anybody is relying on the block hash for anything.

PR #3898, opened 2022-08-31
Pruning of subgraphs

lutter:
This PR resolves issue #3665 and adds a command `graphman prune <deployment>` that removes all history from a deployment before a given block. By default, that block is 10,000 blocks before the current subgraph head. After pruning, the deployment can only be queried at block numbers at least as high as that block.

Here's what it looks like when pruning something:
```
prune QmRuorV4Ck1sVdpfpAAwfYXnf3cfSkbDwZvvzWud9SH8Dg[130]
    latest: 15461533
     final: 15461283
  earliest: 15451533

Analyzed 14 tables in 80s
            table              |  entities  |  versions  |  ratio
-------------------------------+------------+------------+--------
add_liquidity_event            |      41750 |      41750 | 100.0%
eth_purchase_event             |    1156610 |    1156610 | 100.0%
exchange                       |        626 |    2186833 |   0.0%
exchange_day_data              |      19353 |    2185127 |   0.9%
exchange_historical_data       |    2508807 |    2508807 | 100.0%
poi2$                          |          1 |    1412019 |   0.0%
remove_liquidity_event         |      25698 |      25698 | 100.0%
token_purchase_event           |    1285494 |    1285494 | 100.0%
transaction                    |    2473179 |    2473179 | 100.0%
uniswap                        |          1 |    1410157 |   0.0%
uniswap_day_data               |       1288 |    1407878 |   0.1%
uniswap_historical_data        |      19804 |    2184587 |   0.9%
user                           |      98807 |      98807 | 100.0%
user_exchange_data             |      19008 |    2441509 |   0.8%

Copy final entities (versions live between 15451533 and 15461283)
            table              |  versions  |    time
-------------------------------+------------+------------
exchange                       |        182 |        70s
exchange_day_data              |        130 |         5s
poi2$                          |        172 |         5s
uniswap                        |        172 |         3s
uniswap_day_data               |        171 |         3s
uniswap_historical_data        |        130 |         4s
user_exchange_data             |        184 |        11s
Finished copying final entity versions in 103s

Blocking writes and switching tables
            table              |  versions  |    time
-------------------------------+------------+------------
exchange                       |       3095 |         3s
exchange_day_data              |     103159 |        70s
poi2$                          |         10 |         0s
uniswap                        |         10 |         0s
uniswap_day_data               |       1410 |         1s
uniswap_historical_data        |     103159 |        67s
user_exchange_data             |     223731 |       154s
Enabling writes. Switching took 362s

Analyzed 7 tables in 5s
            table              |  entities  |  versions  |  ratio
-------------------------------+------------+------------+--------
exchange                       |       3086 |       3277 |  94.2%
exchange_day_data              |     102822 |     103289 |  99.5%
poi2$                          |          1 |        182 |   0.5%
uniswap                        |          1 |        182 |   0.5%
uniswap_day_data               |       1401 |       1581 |  88.6%
uniswap_historical_data        |     102933 |     103289 |  99.7%
user_exchange_data             |     223127 |     223915 |  99.6%

Finished pruning in 553s
```

Space savings from pruning can be dramatic, depending on the subgraph. A test database with ~ 60 subgraphs shrank by about 1/3, with the size of some subgraphs being reduced by 90%.

### DIFF
--- a/store/postgres/src/copy.rs
+++ b/store/postgres/src/copy.rs
@@ -296,8 +296,10 @@ impl AdaptiveBatchSize {
     // get close to TARGET_DURATION for the time it takes to copy one
     // batch, but don't step up batch_size by more than 2x at once
     pub fn adapt(&mut self, duration: Duration) {
+        // Avoid division by zero
+        let duration = duration.as_millis().max(1);
         let new_batch_size =
-            self.size as f64 * TARGET_DURATION.as_millis() as f64 / duration.as_millis() as f64;
+            self.size as f64 * TARGET_DURATION.as_millis() as f64 / duration as f64;
         self.size = (2 * self.size).min(new_batch_size.round() as i64);
     }
 }
--- a/store/postgres/src/copy.rs
+++ b/store/postgres/src/copy.rs
@@ -202,17 +202,17 @@ impl CopyState {
                     })
             })
             .collect::<Result<_, _>>()?;
-        tables.sort_by_key(|table| table.dst.object.to_string());
+        tables.sort_by_key(|table| table.batch.dst.object.to_string());
 
         let values = tables
             .iter()
             .map(|table| {
                 (
-                    cts::entity_type.eq(table.dst.object.as_str()),
+                    cts::entity_type.eq(table.batch.dst.object.as_str()),
                     cts::dst.eq(dst.site.id),
-                    cts::next_vid.eq(table.next_vid),
-                    cts::target_vid.eq(table.target_vid),
-                    cts::batch_size.eq(table.batch_size),
+                    cts::next_vid.eq(table.batch.next_vid),
+                    cts::target_vid.eq(table.batch.target_vid),
+                    cts::batch_size.eq(table.batch.batch_size),
                 )
             })
             .collect::<Vec<_>>();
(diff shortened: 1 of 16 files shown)


evaporei:
I'm not sure this is ready to review, but the CI seems to be failing on compilation issues to `prost` 🤔

store/postgres/src/copy.rs
```diff
@@ -268,14 +273,106 @@ impl CopyState {
...
+
+        Self { size }
+    }
+
+    // adjust batch size by trying to extrapolate in such a way that we
+    // get close to TARGET_DURATION for the time it takes to copy one
+    // batch, but don't step up batch_size by more than 2x at once
+    pub fn adapt(&mut self, duration: Duration) {
+        let new_batch_size =
+            self.size as f64 * TARGET_DURATION.as_millis() as f64 / duration.as_millis() as f64;
```
leoyvens, line 300: I see this code was just moved, but to make sure we avoid division by zero:
```suggestion
            self.size as f64 * TARGET_DURATION.as_millis() as f64 / (duration.as_millis() as f64 + 1);
```

store/postgres/src/deployment.rs
```diff
@@ -1010,3 +1010,31 @@ pub fn set_entity_count(
...
+    update(d::table.filter(d::id.eq(site.id)))
+        .set(d::earliest_block_number.eq(earliest_block))
+        .execute(conn)?;
+    Ok(())
+}
+
+/// Lock the row for `site` in `subgraph_deployment` for update. This lock
+/// is used to coordinate the changes that the subgraph writer makes with
+/// changes that other parts of the system, in particular, pruning make
+//  see also: deployment-lock-for-update
```
leoyvens, line 1030: Since this doesn't return a handle, I assume it locks for the duration of the transaction? Worth adding a note or linking to relevant PG documentation.

store/postgres/src/relational/prune.rs
```diff
@@ -0,0 +1,331 @@
...
+    copy::AdaptiveBatchSize,
+    deployment,
+    relational::{Table, VID_COLUMN},
+};
+
+use super::{Layout, SqlName};
+
+/// Utility to copy relevant data out of a source table and into a new
+/// destination table and replace the source table with the destination
+/// table
```
leoyvens, line 27: Worth commenting somewhere on this file (maybe as a module-level comment) on the approach for the pruning implementation. In particular I'd naively expect it to be implemented by deleting rows, but I see we're copying to a new table first.

leoyvens (approved):

leoyvens:
That progress report looks nice!

store/postgres/src/copy.rs
lutter: Nice idea, added

store/postgres/src/deployment.rs
lutter: Locks in PG are always to the end of the txn - there's no way to unlock those internal lock (pg_advisory_lock is a different story, but they can span txns) In any event, I updated the comment

store/postgres/src/relational/prune.rs
lutter: Good point. I expanded the comment for `prune_by_copying` to explain that better since I envision that we might have a `prune_by_deleting` at some point, too which is better if we only prune a small amount of data.

PR #3902, opened 2022-09-05
make timestamp numeric

mangas:


evaporei (commented):
LGTM, but since this is a Scalar value change, perhaps there's something missing, I would be more confident if @lutter takes a look too

store/postgres/src/chain_store.rs
```diff
@@ -617,12 +617,30 @@ mod data {
...
 
+            fn try_to_int(ts: Option<String>) -> Result<Option<u64>, StoreError> {
+                let ts = match ts {
+                    Some(str) => str,
+                    None => return Ok(None),
+                };
+
+                let radix = if ts.starts_with("0x") { 16 } else { 10 };
+
+                u64::from_str_radix(&ts[2..], radix)
```
lutter, line 628: This isn't right when the string is base 10 - there we don't want to skip the first two chars

graphql/src/schema/meta.graphql
```diff
@@ -55,8 +55,8 @@ type _Block_ {
   hash: Bytes
   "The block number"
   number: Int!
-  "Timestamp of the block if available, format depends on the chain" 
-  timestamp: String
+  "Timestamp of the block if available" 
+  timestamp: Int
```
lutter, line 59: The comment should also say something about how to interpret that number - maybe add `Integer representation of the timestamp stored in blocks for the chain` to tell people that we don't do any other processing

lutter (approved):
Looks good - two comments that should be addressed before merging, but other than that ok.

graphql/src/schema/meta.graphql
```diff
@@ -55,8 +55,8 @@ type _Block_ {
   hash: Bytes
   "The block number"
   number: Int!
-  "Timestamp of the block if available, format depends on the chain" 
-  timestamp: String
+  "Integer representation of the timestamp store in blocks for the chain"
```
mangas, line 58:
```suggestion
  "Integer representation of the timestamp stored in blocks for the chain" 
```

PR #3903, opened 2022-09-05
chain/ethereum: Monitor transaction receipt fetching duration

tilacog:
Adds a Prometheus subgraph metric for the `eth_getTransactionReceipt` call.


chain/ethereum/src/ethereum_adapter.rs
```diff
@@ -1931,6 +1937,7 @@ async fn get_logs_and_transactions(
     let transaction_receipts_by_hash = get_transaction_receipts_for_transaction_hashes(
         &adapter,
         &transaction_hashes_by_block,
+        subgraph_metrics.cheap_clone(),
```
evaporei, line 1940: I think this can be removed

tilacog: Resolved in #abb8382

evaporei (approved):

Issue #3923, opened 2022-09-13
`indexer-service` crashes after sending introspection query to Graph Node


tilacog:
I found this error while the [`indexer-service`](https://github.com/graphprotocol/indexer) was performing an introspection query on the Graph Node:
```
Error: Decorated type deeper than introspection query.
    at getType (/home/tilacog/tiago/code/edge-and-node/local-network-environments/build/graphprotocol/indexer/node_modules/graphql/utilities/buildClientSchema.js:110:15)
    at getType (/home/tilacog/tiago/code/edge-and-node/local-network-environments/build/graphprotocol/indexer/node_modules/graphql/utilities/buildClientSchema.js:103:42)
    at getType (/home/tilacog/tiago/code/edge-and-node/local-network-environments/build/graphprotocol/indexer/node_modules/graphql/utilities/buildClientSchema.js:113:28)
    at buildField (/home/tilacog/tiago/code/edge-and-node/local-network-environments/build/graphprotocol/indexer/node_modules/graphql/utilities/buildClientSchema.js:302:18)
    at keyValMap (/home/tilacog/tiago/code/edge-and-node/local-network-environments/build/graphprotocol/indexer/node_modules/graphql/jsutils/keyValMap.js:29:27)
    at buildFieldDefMap (/home/tilacog/tiago/code/edge-and-node/local-network-environments/build/graphprotocol/indexer/node_modules/graphql/utilities/buildClientSchema.js:294:37)
    at fields (/home/tilacog/tiago/code/edge-and-node/local-network-environments/build/graphprotocol/indexer/node_modules/graphql/utilities/buildClientSchema.js:218:21)
    at resolveObjMapThunk (/home/tilacog/tiago/code/edge-and-node/local-network-environments/build/graphprotocol/indexer/node_modules/graphql/type/definition.js:504:40)
    at defineFieldMap (/home/tilacog/tiago/code/edge-and-node/local-network-environments/build/graphprotocol/indexer/node_modules/graphql/type/definition.js:760:20)
    at GraphQLObjectType._fields (/home/tilacog/tiago/code/edge-and-node/local-network-environments/build/graphprotocol/indexer/node_modules/graphql/type/definition.js:685:26)
```

`git bisect` revealed that commit 2d27e9be932793d65e37985539482f8ccfccaca5 from #3185 probably introduced the bug in the Graph Node code.

The Query:
```graphql


query IntrospectionQuery {
  __schema {
    queryType {
      name
    }
[...]

PR #3924, opened 2022-09-13
graph-server-index-node: Use ApiVersion instead of undefined SubgraphVersion

kamilkisiela:
Related #3923

The `SubgraphVersion` type was never defined.


saihaj (approved):

tilacog:
After running this branch, the issues reported in #3923 no longer occur.

PR #3934, opened 2022-09-14
feat: AND/OR query filters

### LINKED ISSUE #579: Expose AND and OR logical operators in GraphQL query filters
Where filters should allow AND and OR logical operators to combine and nest conditions in the GraphQL queries.

saihaj:
Enables users to execute queries with logical operators.

Query Example:
```graphql
{
  musicians(where: {OR: {name: "John", id: "m2"}}) {
    name
    id
  }
}
```

Question:  should we limit this to only one level deep? 

Closes #579


lutter (changes requested):
Nice! Looks good in general, but there's some test failures.

One question: are these filters case sensitive? Seeing `AND` when everything else is lower case feels a little strange.

saihaj:
> Seeing `AND` when everything else is lower case feels a little strange.

In the ticket and to make it more differentiable made these logical filters uppercase. Don’t have a preference so happy to do whatever you think is the best.

saihaj:
> so you could have X OR (Y AND Z), and you could have X AND (Y OR Z) but not X OR (Y OR Z)?

@azf20 updated it so it no longer is limiting. User can make as much recursively deep `AND`/`OR` filter they if they chose to do so.

 >  there's some test failures.

@lutter fixed the schema tests

graphql/src/schema/api.rs
```diff
@@ -187,13 +187,28 @@ fn add_filter_type(
...
+                default_value: None,
+                directives: vec![],
+            });
             let typedef = TypeDefinition::InputObject(InputObjectType {
                 position: Pos::default(),
                 description: None,
                 name: filter_type_name,
                 directives: vec![],
-                fields: generated_filter_fields,
+                fields: generated_filter_fields.clone(),
```
saihaj, line 211:
```suggestion
                fields: generated_filter_fields,
```

graphql/src/schema/ast.rs
```diff
@@ -12,6 +12,7 @@ use graph::prelude::{s, Error, ValueType};
 
 use crate::query::ast as qast;
 
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
```
saihaj, line 15:
```suggestion
```

dotansimha:
@saihaj i think we need a rebase there, and then we are good to go in terms of code review?

lutter:
Quick note on rebasing: please always rebase with `git rebase master <mybranch>` not by merging master into the branch since that leads to funky looking merge commits.

graphql/tests/query.rs
```diff
@@ -2123,3 +2123,46 @@ fn deterministic_error() {
...
+    ";
+
+    run_query(QUERY, |result, _| {
+        let exp = object! {
+            musicians: r::Value::List(vec![]),
+        };
+        let data = extract_data!(result).unwrap();
+        assert_eq!(data, exp);
+    })
+}
```
lutter, line 2168: It would be good to also have some tests that mix `and` and `or` filters just to make sure that doesn't get broken inadvertently in the future. Maybe something like `where: { or: { and: { name: "John", id: "m1" }, and: { name: "Lisa", id: "m1" }, { mainBand: "b2" } } }` which, AFAICT, should return `John` and `Tom`

lutter (approved):
Looks good! Sorry it took so long to review.

Just add another test like I suggested in the comment (or more if you can think of more ;) ) and then feel free to rebase and commit.

graphql/tests/query.rs
saihaj: added 2 more nested tests

dotansimha (approved):

graphql/tests/query.rs
```diff
@@ -2123,3 +2123,101 @@ fn deterministic_error() {
...
+
+#[test]
+fn can_query_with_and_and_filter() {
+    const QUERY: &str = "
+    query {
+        musicians(
+          where: {
+                or: {and: {name: \"John\", id: \"m1\" },
+                or: {and: {name: \"Lisa\", id: \"m2\" }
```
lutter, line 2202: I thought repeating the same key here was prohibited by the GraphQL spec, or am I misinterpreting this?

Also cc @azf20 for the syntax of writing `(name = 'John' and id = 'm1') or (name = 'Lisa' and id = 'm2')` As an alternative, we could maybe have `and` and `or` take either an object (`and: {name: "John", id: "m1"}`) or an array of objects (`or: [{and: {name: "John", id: "m1"}}, {and: {name: "Lisa", id: "m2"}}]`)

saihaj: It’s the way I formatted the query for test it may look like prohibited syntax but it isn’t :)
```graphql
query {
  musicians(
    where: {
      or: {
        and: { name: "John", id: "m1" }
        or: { and: { name: "Lisa", id: "m2" } }
      }
    }
  ) {
    id
    name
  }
}
```

azf20: @saihaj I am a bit confused on what that query is doing, is it effectively:
```
(name == "John" AND id == "m1") OR (name == "Lisa" AND id == "m2")
```
Or is it something different? If so I think that syntax is a bit confusing

saihaj: yes @azf20 that is what the query is doing

azf20:
If we proceed with https://github.com/graphprotocol/graph-node/pull/4080 should we close this @saihaj ?

PR #3935, opened 2022-09-14
hack: early filter

mangas:


leoyvens:
Given that we fixed the PoI issue in a different way here https://github.com/graphprotocol/graph-node/pull/3951, so this is purely an optimization, and that:
1. This optimization was targeting subgraphs using static filters.
2. Static filters will be dynamically enabled only when the number of data sources is large, even with firehose.

That reduces the number of subgraphs that this would potentially help. So I think we should revisit when we have more benchmark data to decide which optimizations to pursue.

PR #3938, opened 2022-09-14
graphql,server: make `Resolver::resolve_object` async

tilacog:
Currently, the `subgraphFeatures` route is panicking because `Store`s implementation of `Resolver::resolve_object` calls `block_on` from within an async context.

This fix makes the `Resolver::resolve_object` an async method, allowing us to `await` instead of blocking the execution.


lutter (approved):
LGTM!

Issue #3911, opened 2022-09-07
Add index-node endpoint to fetch block hash for a given network:blockNumber


azf20:
- This is required to allow the indexer agent to fetch the block hash so it can deterministically get the POI

PR #3942, opened 2022-09-15
`index-node` endpoint to get a hash from a block number

neysofu:
This new endpoint fills a need that several services (namely EBO and indexer agent) that rely on `graph-node` have, i.e. to fetch a block hash starting from the number. The underlying implementation bridges over to the configured data ingestor for the requested network, allowing these services to get this data directly from `graph-node`.

I've used some declarative macro idiosyncracy to quickly abstract over `BlockchainMap` items with different concrete types.

```graphql
# Query example
{
  blockHashFromNumber(network: "mainnet", blockNumber: 15239780)
}
```

### DIFF
--- a/graphql/src/execution/execution.rs
+++ b/graphql/src/execution/execution.rs
@@ -587,14 +587,17 @@ async fn resolve_field_value(
             .await
         }
 
-        s::Type::NamedType(ref name) => resolve_field_value_for_named_type(
-            ctx,
-            object_type,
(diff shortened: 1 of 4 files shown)


neysofu:
Closes https://github.com/graphprotocol/graph-node/issues/3911.

evaporei (approved):
LGTM

Smart idea to put a match on the Blockchain enum 👏

PR #3947, opened 2022-09-16
graphql: Remove validation cache

lutter:
Initial testing suggests that, after the recent improvement in graphql-tools, validations are now fast enough to not need such a cache.

### DIFF
--- a/graphql/src/execution/query.rs
+++ b/graphql/src/execution/query.rs
@@ -2,10 +2,8 @@ use graph::data::graphql::DocumentExt as _;
 use graph::data::value::Object;
 use graphql_parser::Pos;
 use graphql_tools::validation::rules::*;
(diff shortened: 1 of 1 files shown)


dotansimha:
Thanks @lutter ! I think it might be a good take to cache it regardless? Since for a given input query, the validation result will always be consistent (as long as the schema didn't change)

lutter:
I would first like to see if we really need to cache the validation result; caching that is actually pretty complex since computing a good cache key quickly is not all that easy as #3840 shows. 

I am also not sure how effective that cache can ever be, since the validity of a query also depends on variable values (e.g. rule `VariablesAreInputTypes`) which means that the cached result is only valid for the exact same query with the exact same variable values.

leoyvens (approved):

PR #3950, opened 2022-09-19
Automatically group imports with `cargo +nightly fmt`

neysofu:
While `rustfmt` doesn't have any stable option to lint and reorder imports, nightly `rustfmt` does:

- `group_imports`
- `imports_granularity`

The default value for both of these is `Preserve`, meaning `rustfmt` won't normally touch the imports. However, we may benefit from a `rustfmt`-enforced canonical ordering of imports, given the huge number of imported items in many of our source files. This also makes imports more easily greppable, which can be helpful.

This PR doesn't enforce the new standard in the CI, as that would require devs to install and run nightly `rustfmt` alongside stable `rustc`, which results in bad editor integration. Instead, we're running it as a one-time thing and we may run it again in the future when the imports get messy once again.

To review this PR, follow these steps:
- `git checkout c332a90e3`
- Review `rustfmt.toml`
- `cargo +nightly fmt`
- `git diff cad5b6109`
- Verify that the diff is empty and decide if you like the new formatting

PS: as an extra, I'm also turning on `wrap_comments`, which is also supported on stable `rustfmt`.


leoyvens:
I'm neutral on whether grouped imports are nicer or not. But running this manually every now and then would cause a bunch of merge conflicts with any open PRs, so I'd rather do nothing at least until this option is stable.

neysofu:
Yeah, that's fair. I wanted this to immediately see where a module is imported from across the whole project. I'll see if there's a way to use these options in stable `rustfmt`, otherwise we'll close.

neysofu:
`unstable_features = true` only works on nightly `rustfmt`. Can't do much about it, closing.

PR #3951, opened 2022-09-19
fix false positive POI

mangas:
https://github.com/graphprotocol/graph-node/issues/3963

### DIFF
--- a/store/postgres/src/subgraph_store.rs
+++ b/store/postgres/src/subgraph_store.rs
@@ -228,7 +228,7 @@ impl SubgraphStore {
         }
     }
 
-    pub(crate) async fn get_proof_of_indexing(
+    pub async fn get_proof_of_indexing(
         &self,
         id: &DeploymentHash,
         indexer: &Option<Address>,
--- a/tests/tests/runner.rs
+++ b/tests/tests/runner.rs
@@ -2,6 +2,7 @@ use std::sync::Arc;
 
 use cid::Cid;
 use graph::blockchain::{Block, BlockPtr};
+use graph::env::EnvVars;
 use graph::object;
 use graph::prelude::ethabi::ethereum_types::H256;
 use graph::prelude::{SubgraphAssignmentProvider, SubgraphName};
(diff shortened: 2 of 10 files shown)


tests/tests/runner.rs
```diff
@@ -151,3 +168,57 @@ async fn file_data_sources() {
...
+        let test_dir = format!("./integration-tests/{}", subgraph_name);
+        fixture::build_subgraph(&test_dir).await
+    };
+
+    let blocks = {
+        let block_0 = genesis();
+        let block_1 = empty_block(block_0.ptr(), test_ptr(1));
+        let block_2 = empty_block(block_1.ptr(), test_ptr(2));
+        vec![block_0, block_1, block_2]
+    };
```
leoyvens, line 188: Have you checked that this test fails on master? I'm surprised that it would reproduce with just empty blocks.

leoyvens (changes requested):
The new test should be added to the workspace in `tests/integration-tests/package.json` to avoid adding the huge `package-lock.json`.

tests/tests/runner.rs
mangas:
```
Sep 21 10:10:07.912 DEBG 1 candidate trigger in this block, block_hash: 0x0000000000000000000000000000000000000000000000000000000000000001, block_number: 1, sgd: 9, subgraph_id: Qmepz1JnzZNFHLhYRUBLgi53dnhuukEuKNjCKzsg2g1YrF, component: SubgraphInstanceManager
```

This is printed during the test and is being filtered out by match_and_decode, it's the trigger we need.

leoyvens: Nice.

tests/tests/runner.rs
```diff
@@ -151,3 +168,56 @@ async fn file_data_sources() {
...
+        Some(env_vars),
+    )
+    .await;
+    ctx.start_and_sync_to(stop_block).await;
+
+    let poi = ctx
+        .store
+        .get_proof_of_indexing(&ctx.deployment.hash, &None, test_ptr(1))
+        .await
+        .unwrap();
```
leoyvens, line 209: This is testing the PoI value at block 1, but it's worth also comparing to the value at block 0 or block 2 and ensure it doesn't change.

leoyvens (approved):

store/postgres/src/subgraph_store.rs
```diff
@@ -228,7 +228,7 @@ impl SubgraphStore {
         }
     }
 
-    pub(crate) async fn get_proof_of_indexing(
+    pub async fn get_proof_of_indexing(
```
lutter, line 231: It doesn't seem like there is any need for this change; `master` still compiles with this as `pub(crate)`

lutter: Ah, I missed that; it's fine to leave it as is then

PR #3966, opened 2022-09-20
store: Be considerate of replicas when copying

lutter:
Copying of subgraphs can create a lot of data to be replicated, which can make replicas fall hopelessly behind. With this change, the copy code checks whether there are replicas that are too far behind and backs off of copying for a bit to give them a chance to catch up.

### DIFF
--- a/store/postgres/src/copy.rs
+++ b/store/postgres/src/copy.rs
@@ -38,7 +38,7 @@ use graph::{
 };
 
 use crate::{
-    advisory_lock,
+    advisory_lock, catalog,
     dynds::DataSourcesTable,
     primary::{DeploymentId, Site},
 };
@@ -53,6 +53,16 @@ const INITIAL_BATCH_SIZE: i64 = 10_000;
 const INITIAL_BATCH_SIZE_LIST: i64 = 100;
 const TARGET_DURATION: Duration = Duration::from_secs(5 * 60);
 const LOG_INTERVAL: Duration = Duration::from_secs(3 * 60);
+
+/// If replicas are lagging by more than this, the copying code will pause
+/// for a while to allow replicas to catch up
+const MAX_REPLICATION_LAG: Duration = Duration::from_secs(60);
+/// If replicas need to catch up, do not resume copying until the lag is
+/// less than this
+const ACCEPTABLE_REPLICATION_LAG: Duration = Duration::from_secs(30);
+/// When replicas are lagging too much, sleep for this long before checking
+/// the lag again
+const REPLICATION_SLEEP: Duration = Duration::from_secs(10);
 
 table! {
     subgraphs.copy_state(dst) {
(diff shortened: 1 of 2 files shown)


store/postgres/src/copy.rs
```diff
@@ -744,6 +754,24 @@ impl Connection {
...
+                // Pause copying if replication is lagging behind to avoid
+                // overloading replicas
+                let mut lag = catalog::replication_lag(&self.conn)?;
+                if lag > MAX_REPLICATION_LAG {
+                    loop {
+                        info!(&self.logger,
+                             "Replicas are lagging too much; pausing copying for {}s to allow them to catch up",
+                             REPLICATION_SLEEP.as_secs();
+                             "lag_s" => lag.as_secs());
+                        std::thread::sleep(REPLICATION_SLEEP);
```
neysofu, line 767: Isn't sleeping for 30s excessive here before polling again? I'd imagine there's many cases where the replica catches up within that time. Anyway, this is something that we can measure and tweak later.

store/postgres/src/copy.rs
```diff
@@ -744,6 +754,24 @@ impl Connection {
                 if table.is_cancelled(&self.conn)? {
                     return Ok(Status::Cancelled);
                 }
+
+                // Pause copying if replication is lagging behind to avoid
+                // overloading replicas
+                let mut lag = catalog::replication_lag(&self.conn)?;
+                if lag > MAX_REPLICATION_LAG {
```
neysofu, line 761: This is equivalent to a `while lag > MAX_REPLICATION_LAG` loop.

lutter: No, it's not, since we only get out of the loop when lag falls below `ACCEPTABLE_REPLICATION_LAG`; getting in and out of the loop is asymmetric, we basically hope that if lag is between `ACCEPTABLE_REPLICATION_LAG` and `MAX_REPLICATION_LAG` that it will resolve itself.

store/postgres/src/copy.rs
lutter: Yeah, true; I am lowering this to 10s. Querying for lag every 10s shouldn't cause issues

store/postgres/src/copy.rs
lutter: > This is equivalent to a `while lag > MAX_REPLICATION_LAG` loop.

It's not, since we only get out of the loop when lag falls below ACCEPTABLE_REPLICATION_LAG; getting in and out of the loop is asymmetric, we basically hope that if lag is between ACCEPTABLE_REPLICATION_LAG and MAX_REPLICATION_LAG that it will resolve itself.

neysofu: Oh, right! I didn't catch that. Marking as resolved :smile:

store/postgres/src/copy.rs
neysofu: :+1:

neysofu (approved):
Nice!

PR #3974, opened 2022-09-22
Ship `envsubst` with the Docker image to facilitate TOML templating

### LINKED ISSUE #3946: Make the config.toml file more modular by expanding env vars inside it
Feature request

Only some of the parameters can be inserted via env vars

Similar to  https://github.com/graphprotocol/graph-node/commit/b82ac37784db19261f200b3b3720acd010a3ed19 but for pretty much anything there is to set inside the toml file.

Example:
```
[chains.${CHAIN_0}]
shard = "${SHARD_0}"
provider = [ { label = "${CHAIN_0}", url = "${CHAIN_RPC_0}", features = ["${CHAIN_0_FEATURES}"] } ]```

### PR DESCRIPTION
**Pull request by neysofu**
`envsubst` takes care of all foreseeable templating needs that we might face for TOML configuration files.

From all `envsubst` implementations that I could find, I selected https://github.com/a8m/envsubst because it seemed to be the most complete and well adopted. More fully-fledged templating engines like http://www.confd.io/ are interesting but probably overkill.

Here's an example of a configuration pattern that `envsubst` can deal with, while our current variable expansion logic can't:

```toml
[chains.${CHAIN_0}]
shard = "${SHARD_0}"
provider = [ { label = "${CHAIN_0}", url = "${CHAIN_RPC_0}", features = ["${CHAIN_0_FEATURES}"] } ]
```

Closes #3946.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -4,6 +4,14 @@
 # by running something like the following
 #   docker build --target STAGE -f docker/Dockerfile .
 
+FROM golang:buster as envsubst
+
+# v1.2.0
+ARG ENVSUBST_COMMIT_SHA=16035fe3571ad42c7796bf554f978bb2df64231b
+
+RUN go install github.com/a8m/envsubst/cmd/envsubst@$ENVSUBST_COMMIT_SHA \
+    && strip -g /go/bin/envsubst
+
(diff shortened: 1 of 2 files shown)


evaporei (approved):
💯

PR #3975, opened 2022-09-22
Cosmos: Add the message handler

pienkowb:
This PR extends the Cosmos integration with a new type of handler, which enables subgraph developers to access transaction data in a more granular way.

Each Cosmos transaction contains multiple messages. Until now, in order to access a single message a subgraph developer had to define a transaction handler and loop over the message array inside the mapping function. With the introduction of the message handler, the same goal can be achieved just by adding a message handler and specifying a message type.

The data passed to the message handler includes the transaction context, which is a light-weight structure that contains the most important information about a transaction (hash, index, status code, gas information). The same structure is also passed to the event handler.


evaporei (approved):
Looks correct to me 👍 

I'm just curious, what's `gas_wanted`?

pienkowb:
> Looks correct to me 👍
> 
> I'm just curious, what's `gas_wanted`?

@evaporei  It's the amount of gas requested for a transaction.

PR #3982, opened 2022-09-22
graph,node: Migrate from structopt to clap v3

tilacog:
The `structopt` crate is in maintenance mode, and authors suggest that users migrate to `clap`, which now has derive macros in a similar fashion.

This PR updates all `structopt` occurrences in favor of `clap` without altering the CLI usage.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -55,15 +55,6 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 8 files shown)


lutter (approved):
Nice! I've been meaning to do that for a while, thanks for doing this!

Issue #3260, opened 2022-02-21
Refactor to ease adding new chains


otaviopace:
## Context

When we only had `EVM` compatible chains, adding new ones didn't require much change. To support NEAR (our first non EVM-compatible chain) required multiple refactors, `trait` additions, etc. When that got merged we wondered what could be improved to **ease adding new chains**, in the sense of code structure/reuse and boilerplate.

With the recent addition of Tendermint ([PR](https://github.com/graphprotocol/graph-node/pull/3212)) it's now more clear what could be improved.

## Objective

This will be an umbrella issue connecting all possible improvements/refactors that we can do today to ease new chains being added on the code level, requiring less boilerplate code.

## Possible improvements

The improvements/questions below were noted after a second revision of the Tendermint integration Pull Request.

- Could we somehow concentrate all `build.rs` files logic into a single place? Since adding new Firehose chains will always require this to build `prost` (gRPC library)
- About empty/default implementations:
  - `TriggerFilter`: could we have a default implementation for this that doesn't do anything?
  - `NodeCapabilities`: could we have a default when the chain doesn't require anything specific for this?
  - `RuntimeAdapter`: could we have an empty/default adapter when no logic is needed here?
- `DataSource`:
  - Should we make the getter code here created by a `macro`? I mean in the sense that there are tons of similar getter functions that all chains have to do, the only difference are the fields/types themselves.
  - There are some checks that could be done for all chains (eg: `handlers > 0`). Should we have traits for each type of validation required so the chain chooses to implement what suits best?
- ABI/AS code generation: this is being addressed by #2985
- On `main.rs` (and related binaries) the part that instantiates the chain code is being addressed by a current refactor for `graphman run` that I'm currently working on (which will remove other code duplication there as well)

PR #3985, opened 2022-09-23
`EmptyNodeCapabilities` for chains that don't discriminate nodes

neysofu:
Provides a default `NodeCapabilities` implementor called `EmptyNodeCapabilities`, so that new chains can use that without writing their own, assuming they don't need any custom logic.

Also "fixes" the `PartialOrd` implementation of Ethereum's `NodeCapabilities`, which is not buggy per se but breaks the required trait invariants.

Related to https://github.com/graphprotocol/graph-node/issues/3260.


neysofu:
Alright, I think I see how indeed we can remove all traces of `NodeCapabilities` from the `Blockchain` trait. Some problems arise with `BlockStreamBuilder::build_firehose` though, where the current flow of logic is `TriggerFilter -> NodeCapabilities -> TriggersAdapter` via `Blockchain::triggers_adapter`. This can be transformed into `TriggerFilter -> &[Chain::DataSource] -> TriggersAdapter`, but getting the data sources out of a `TriggerFilter` is not immediate and it would require refactoring that trait as well. I'd leave it for a future PR, unless I'm missing something and it's easier than that.

leoyvens:
Afaict `fn build_firehose` does not have a good reason to be constructing a triggers adapter. I only see it used here 
https://github.com/graphprotocol/graph-node/blob/e5198128bd3d10c95325580ade063f658709fbe4/chain/ethereum/src/chain.rs#L627-L631

That TODO isn't hard to fix, this block needs to be extracted to a new `fn parse_triggers` and that should be used directly instead of going through the triggers adapter
https://github.com/graphprotocol/graph-node/blob/e5198128bd3d10c95325580ade063f658709fbe4/chain/ethereum/src/chain.rs#L529-L536

neysofu:
I had totally forgot about this. I never pursued the follow-up refactoring, should we merge it as it is (after fixing conflicts)?

chain/ethereum/src/capabilities.rs
```diff
@@ -15,28 +13,28 @@ pub struct NodeCapabilities {
...
             self.traces.cmp(&other.traces),
-        ) {
-            (Ordering::Greater, Ordering::Greater) => Ordering::Greater,
-            (Ordering::Greater, Ordering::Equal) => Ordering::Greater,
-            (Ordering::Equal, Ordering::Greater) => Ordering::Greater,
-            (Ordering::Equal, Ordering::Equal) => Ordering::Equal,
-            (Ordering::Less, _) => Ordering::Less,
-            (_, Ordering::Less) => Ordering::Less,
-        }
+        ])
```
leoyvens, line 23: Isn't this just `(&self.archive, &self.traces).cmp((&other.archive, &other.traces))`.

leoyvens (approved):

chain/ethereum/src/capabilities.rs
neysofu: No, Rust tuples define lexicographic ordering, which is not what we use for node capabilities comparisons (e.g. all fields have to be `>=` in order for a `NodeCapabilities` to be `>`, but if we just compare them as tuples, the comparison might stop at `archive` and discard `traces`).

leoyvens: I see, thanks

PR #3989, opened 2022-09-26
fix: Allow grafts to add data sources

### LINKED ISSUE #3960: Adding data sources in graft not working
It's a reasonable use case for grafting to add a data source with a start block larger than the graft block. However the recent change to a private data sources table broke this, because it assumes that the manifest index of the templates is the same as in the source subgraph, and adding a data source will increase the manifest index. This can be fixed by translating the manifest indexes when grafting.

leoyvens:
Resolves #3960. This is done by adding a `raw_yaml` column to the `subgraph_manifest` table, and using it to map the manifest idx from the source to the destination deployment. For existing deployments this column is migrated in the instance manager on subgraph startup. This also changes a bit the order in which things are done in the instance manager, now the subgraph is first fetched from IPFS before being started in the store.

The `data-source-revert` integration test was modified to also test this.

### DIFF
--- a/graph/src/data/subgraph/schema.rs
+++ b/graph/src/data/subgraph/schema.rs
@@ -1,6 +1,6 @@
 //! Entity types that contain the graph-node state.
 
-use anyhow::{anyhow, Error};
+use anyhow::{anyhow, bail, Error};
 use hex;
 use lazy_static::lazy_static;
 use rand::rngs::OsRng;
@@ -110,11 +110,12 @@ pub struct DeploymentCreate {
 
 impl DeploymentCreate {
     pub fn new(
+        raw_manifest: String,
         source_manifest: &SubgraphManifest<impl Blockchain>,
         earliest_block: Option<BlockPtr>,
     ) -> Self {
         Self {
-            manifest: SubgraphManifestEntity::from(source_manifest),
+            manifest: SubgraphManifestEntity::new(raw_manifest, source_manifest),
             earliest_block: earliest_block.cheap_clone(),
             graft_base: None,
             graft_block: None,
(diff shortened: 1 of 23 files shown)


graph/src/data/subgraph/schema.rs
```diff
@@ -163,18 +164,52 @@ pub struct SubgraphManifestEntity {
...
             spec_version: manifest.spec_version.to_string(),
             description: manifest.description.clone(),
             repository: manifest.repository.clone(),
             features: manifest.features.iter().map(|f| f.to_string()).collect(),
             schema: manifest.schema.document.clone().to_string(),
+            raw_yaml: Some(raw_yaml),
         }
     }
+
+    pub fn template_idx_and_name(&self) -> Result<Vec<(i32, String)>, Error> {
```
neysofu, line 182: I understand we're using `i32` to simplify comparing these values with query results, but it would be nice to postpone the conversion and create a type alias for this tuple.

store/postgres/src/dynds/private.rs
```diff
@@ -206,37 +207,80 @@ impl DataSourcesTable {
...
+                &self.manifest_idx,
+                &self.param,
+                &self.context,
+                &self.causality_region,
+            ))
+            .order_by(&self.vid)
+            .load::<Tuple>(conn)?;
+
+        let mut count = 0;
+        for (block_range, src_manifest_idx, param, context, causality_region) in src_tuples {
```
neysofu, line 242: I suspect these N+1 queries may take longer than we'd like, but I don't have a better idea and N is most likely small for most subgraphs.

neysofu (approved):
LGTM, and nice small cleanups along the way! Just left two minor comments.

graph/src/util/error.rs
```diff
@@ -17,3 +17,10 @@ macro_rules! ensure {
         }
     };
 }
+
+#[macro_export]
+macro_rules! bail {
+    ($($err:tt)*) => {
+        return Err(anyhow::anyhow!($($err)*).into());
+    };
+}
```
lutter, line 26: Why don't we just use the existing `anyhow::bail!` macro? If this is truly needed, maybe add a comment why.

graphql/tests/query.rs
```diff
@@ -267,7 +267,7 @@ async fn insert_test_entities(
     manifest: SubgraphManifest<graph_chain_ethereum::Chain>,
     id_type: IdType,
 ) -> DeploymentLocator {
-    let deployment = DeploymentCreate::new(&manifest, None);
+    let deployment = DeploymentCreate::new(String::new(), &manifest, None);
```
lutter, line 270: Could we make the raw yaml an `Option` here? Alternatively, `deployment::set_manifest_raw_yaml` could translate an empty string back to `None`.

store/postgres/src/deployment.rs
```diff
@@ -275,6 +276,21 @@ pub fn features(conn: &PgConnection, site: &Site) -> Result<BTreeSet<SubgraphFea
...
 
+/// This migrates subgraphs that existed before the raw_yaml column was added.
+pub fn set_manifest_raw_yaml(
+    conn: &PgConnection,
+    site: &Site,
+    raw_yaml: &str,
+) -> Result<(), StoreError> {
+    use subgraph_manifest as sm;
+
+    update(sm::table.filter(sm::id.eq(site.id)))
```
lutter, line 287: It would be better to also include `.filter(sm::raw_yaml.ne(raw_yaml))` to avoid unnecessary updates in that table. Otherwise, we'd rewrite the entire table on each node start.

lutter (changes requested):
The general approach looks good! Just some smaller detail comments on how we go about storing the raw yaml.

graph/src/util/error.rs
leoyvens: It doesn't call `.into()`, yhea needs a comment.

graphql/tests/query.rs
leoyvens: If the tests work, I would prefer not to make the interface more option-y just for the tests.

lutter: Oh, I totally missed that that was in the tests. Just ignore what I said.

lutter (approved):

store/postgres/src/deployment.rs
lutter: Sounds good! The sideeffect is that we'll never update the `raw_yaml` once it's set, but that shouldn't be an issue

PR #3990, opened 2022-09-26
firehose: Dont use debug format on firehose errors

leoyvens:
This would print the error with newlines, the best is to use `{:#}` to print with causes.


maoueh (approved):

maoueh:
Merging right away.

Issue #3948, opened 2022-09-17
File data sources: support files in IPFS directories


azf20:
IPFS file data sources based on files in directories (e.g. https://ipfs.io/ipfs/QmaXzZhcYnsisuue5WRdQDH6FDvqkLQX1NckLqBYeYYEfm/1.json) fail with the following error:
```
        "fatalError": {
          "message": "Failed to parse multihash",
        }
```

Example mapping:
```
token.tokenURI = "/" + event.params.tokenId.toString() + ".json";
token.ipfsURI = ipfshash + token.tokenURI;
TokenMetadataTemplate.create(token.ipfsURI);
```
File data sources should be able to reference these kind of files as well as a CID

leoyvens:
Done!

PR #3992, opened 2022-09-26
ipfs_service: Add support for folders

mangas:
https://github.com/graphprotocol/graph-node/issues/3948

### DIFF
--- a/core/tests/fixtures/ipfs_folder/hello.txt
+++ b/core/tests/fixtures/ipfs_folder/hello.txt
@@ -1,0 +1,1 @@
+hello world
--- a/graph/src/data_source/offchain.rs
+++ b/graph/src/data_source/offchain.rs
@@ -7,10 +7,10 @@ use crate::{
     },
     data::store::scalar::Bytes,
     data_source,
+    ipfs_client::CidFile,
     prelude::{DataSourceContext, Link},
 };
 use anyhow::{self, Context, Error};
-use cid::Cid;
 use serde::Deserialize;
 use slog::{info, Logger};
 use std::{fmt, sync::Arc};
@@ -74,7 +74,7 @@ impl DataSource {
 
     pub fn as_stored_dynamic_data_source(&self) -> StoredDynamicDataSource {
         let param = match self.source {
-            Source::Ipfs(link) => Bytes::from(link.to_bytes()),
+            Source::Ipfs(ref link) => Bytes::from(link.to_bytes()),
         };
         let context = self
             .context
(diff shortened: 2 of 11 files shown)


core/tests/fixtures/ipfs_folder/hello.txt
```diff
@@ -0,0 +1 @@
+hello world
```
leoyvens, line 1: Is `hello.txt` being used?

graph/src/data_source/offchain.rs
```diff
@@ -112,14 +114,14 @@ impl DataSource {
     /// used as the value to be returned to mappings from the `dataSource.address()` host function.
     pub fn address(&self) -> Option<Vec<u8>> {
         match self.source {
-            Source::Ipfs(cid) => Some(cid.to_bytes()),
+            Source::Ipfs(CidFile { cid, path: _ }) => Some(cid.to_bytes()),
```
leoyvens, line 117: What happens is that we were using `dataSource.address()` to return the CID in byte form. But now the input can be a CID + path to traverse directories. And paths are strings. A couple options I can see:
1. The one taken here, return just the CID and lose the path.
2. Change this to return an utf8 string, which is more symmetrical with the data source creation which takes a string. The host fn is called `address` and returns an `Address` on the AS side which is strange, but we can improve that on graph-ts.

I lean towards 2, wdyt @azf20.

graph/src/ipfs_client.rs
```diff
@@ -1,14 +1,74 @@
...
 use reqwest::multipart;
 use serde::Deserialize;
+use std::fmt::Display;
 use std::time::Duration;
 use std::{str::FromStr, sync::Arc};
 
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct CidFile {
+    pub cid: Cid,
+    pub path: Option<String>,
```
leoyvens, line 18: Lets add a comment above this field clarifying that it never contains the leading slash.

leoyvens (commented):
Generally looks good! We just have that open question on `dataSource.address()`.

core/tests/fixtures/ipfs_folder/hello.txt
mangas: I left it just so the fold is not empty, can also be used for other tests but I'm not using it otherwise we could get false results on the tests from previous runs.

graph/src/data_source/offchain.rs
mangas: we could also just return the entire path {cid}/{path} in bytes?

leoyvens: Yes I think I wasn't clear, 2 is returning the entire path as a string encoded as a utf8 byte array.

core/tests/fixtures/ipfs_folder/hello.txt
leoyvens: Makes sense.

graph/src/data_source/offchain.rs
azf20: sorry I missed this - yes I think 2 makes sense here.
I had some trouble turning `dataSource.address()` into an ipfsHash I recognised during my testing, but I expect that was user error

leoyvens: That's the consensus then, @mangas lets do 2 and this will be ready to merge.

leoyvens (approved):

PR #4001, opened 2022-09-27
Feature/ethereum provider feature call only

maoueh:
Open a PR so we can discuss it at the weekly tomorrow, right now I feel the feature part is weird for `call_only` has it brings interesting problems like the fact that in most cases, "call_only" provider should never be actually selected. 

The current PR reflects that with a new `cheapest_with_for_call` which tries to use `call_only` provider first then fallback to normal behavior if none are found. So it's prioritizing `call_only` in this case but only in this case.

### DIFF
--- a/chain/ethereum/src/network.rs
+++ b/chain/ethereum/src/network.rs
@@ -23,27 +23,54 @@ pub struct EthereumNetworkAdapter {
     limit: usize,
 }
 
+impl EthereumNetworkAdapter {
+    pub fn call_only(&self) -> bool {
+        self.adapter.call_only
+    }
+}
+
 #[derive(Clone)]
 pub struct EthereumNetworkAdapters {
-    pub adapters: Vec<EthereumNetworkAdapter>,
+    pub full_adapters: Vec<EthereumNetworkAdapter>,
+    pub call_only_adapters: Vec<EthereumNetworkAdapter>,
 }
 
 impl EthereumNetworkAdapters {
+    pub fn is_capabilities_fulfilled(&self, required_capabilities: &NodeCapabilities) -> bool {
+        // Now that we have a split on adapters, we must perform some logic to determine if
+        // capabilities can be fulfilled or not.
+        match (required_capabilities.archive, required_capabilities.traces) {
+            // Those 2 cases cannot be fulfilled by 'call_only', so it means we only inspect 'full_adapters'
+            (false, _) => has_adapter_for(&self.full_adapters, required_capabilities),
+            // This case could be provided by both set so we check both
+            (true, false) => {
+                has_adapter_for(&self.full_adapters, required_capabilities)
+                    || has_adapter_for(&self.call_only_adapters, required_capabilities)
+            }
+            // The case '{ archive: true, traces: true }' is problematic, 'archive' could be provided
+            // by 'call_only_adapters' while 'traces' by 'full_adapters'.
+            (true, true) => {
+                let only_archive = NodeCapabilities {
+                    archive: true,
+                    traces: false,
+                };
+                let only_traces = NodeCapabilities {
+                    archive: false,
+                    traces: true,
+                };
+
+                has_adapter_for(&self.full_adapters, &only_traces)
+                    && (has_adapter_for(&self.full_adapters, &only_archive)
+                        || has_adapter_for(&self.call_only_adapters, &only_archive))
+            }
+        }
+    }
+
     pub fn all_cheapest_with(
         &self,
         required_capabilities: &NodeCapabilities,
     ) -> impl Iterator<Item = Arc<EthereumAdapter>> + '_ {
-        let cheapest_sufficient_capability = self
-            .adapters
-            .iter()
-            .find(|adapter| &adapter.capabilities >= required_capabilities)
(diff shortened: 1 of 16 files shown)


chain/ethereum/src/runtime/runtime_adapter.rs
```diff
@@ -45,9 +45,10 @@ impl blockchain::RuntimeAdapter<Chain> for RuntimeAdapter {
         let call_cache = self.call_cache.cheap_clone();
         let eth_adapter = self
             .eth_adapters
-            .cheapest_with(&NodeCapabilities {
+            .cheapest_with_for_call(&NodeCapabilities {
```
leoyvens, line 48: This is the only place where we'd use a call only adapter, so that's why I thought it could use a dedicated method that doesn't need a `NodeCapabilities` argument.

maoueh:
I have no problem exploring that route and removing modification of the node capabilities. I've pushed an updated version, tell me how it feels and what you would like to see different.

chain/ethereum/src/network.rs
```diff
@@ -60,11 +68,54 @@ impl EthereumNetworkAdapters {
...
+                )
+            })
+    }
+
     pub fn cheapest(&self) -> Option<Arc<EthereumAdapter>> {
         // EthereumAdapters are sorted by their NodeCapabilities when the EthereumNetworks
         // struct is instantiated so they do not need to be sorted here
         self.adapters
             .iter()
+            .filter(|adapter| !adapter.call_only())
```
leoyvens, line 118: Given that everywhere we iterate over `adapters` we filter on this, it seems it would be cleaner to separate the lists of ordinary and call-only adapters.

chain/ethereum/src/network.rs
```diff
@@ -135,9 +186,14 @@ impl EthereumNetworks {
...
-                .adapters
-                .sort_by_key(|adapter| adapter.capabilities)
+            adapters.adapters.sort_by(|a, b| {
+                a.capabilities
+                    .partial_cmp(&b.capabilities)
+                    // We can't define a total ordering over node capabilities,
+                    // so incomparable items are considered equal and end up
+                    // near each other.
+                    .unwrap_or(Ordering::Equal)
+            })
```
leoyvens, line 196: Is this change still necessary, or would `sort_by_key` work again?

maoueh:
I had to made some modification to how exactly `triggers_adapter` is actually called and implemented.

Indeed, in `instance_manager`, we retrieved a `triggers_adapter` based on the subgraph requested capabilities. This was causing errors when defining the provider with 'archive' as 'call_only'.

The problem is that at that point in `instance_manager`, nobody is actually going to use `archive` directly, only `traces` is of interset. But `instance_manager` is chain agnostic, so I couldn't implement the logic there. To keep the validation of the features of a subgraph, I introduced `is_capabilities_fulfilled` on `Chain` trait. That way, the `instance_manager` can validate the features.

And the `triggers_adapter` on Ethereum implementation was changed to remove the need for 'archive' capabilities because it's not needed there. Only when actually getting a provide to perform the `eth_call` of the subgraph that we need `archive`.

Happy to discuss those change over a call if needed.

graph/src/blockchain/mod.rs
```diff
@@ -149,6 +149,12 @@ pub trait Blockchain: Debug + Sized + Send + Sync + Unpin + 'static {
 
     type NodeCapabilities: NodeCapabilities<Self> + std::fmt::Display;
 
+    fn is_capabilities_fulfilled(
```
mangas, line 152: I think the right place for this is on the NodeCapabilities trait since the Blockchain already contains a NodeCapabilities associated type

chain/ethereum/src/network.rs
```diff
@@ -64,17 +91,76 @@ impl EthereumNetworkAdapters {
...
+    }
+
     pub fn remove(&mut self, provider: &str) {
-        self.adapters
+        self.full_adapters
             .retain(|adapter| adapter.adapter.provider() != provider);
     }
 }
 
+fn has_adapter_for<'a>(
```
mangas, line 127: This pattern (firehose + calls adapter) seems to only be necessary for ethereum, could we move this logic to the EthereumAdapterSelector? 

Would it be enough to look for an "eth_call" adapter there? In that case, if there is one we use that, if not we can fallback for an archive, otherwise current firehose configurations would be broken.

graph/src/blockchain/mod.rs
maoueh: Seems like a good idea. 

While checking to do it, I hit a small road block though. My current implementation on Ethereum relies on the `self.firehose_endpoints` which is hold by `chain`.

If I move it to `NodeCapabilities` I'm not sure how I would access this. In the chain agnostic `instance_manager.rs`, the call site looks like:

```
        if !chain.is_capabilities_fulfilled(&required_capabilities, &logger) {
            return Err(anyhow!(
                "deployment {} with required capabilities {} cannot be fulfilled by this instance",
                &deployment,
                &required_capabilities
            ));
        }
```

Do you think we have ways around to fix this?

mangas:
This PR can now be closed, https://github.com/graphprotocol/graph-node/pull/4351 implements this

PR #4008, opened 2022-09-29
enable static filters for SGs with many dynamic datasource 

mangas:
Add a subgraph start check for number of dynamic data sources. If above the defined threshold, force static filters.

### DIFF
--- a/core/src/subgraph/instance_manager.rs
+++ b/core/src/subgraph/instance_manager.rs
@@ -182,7 +182,7 @@ impl<S: SubgraphStore> SubgraphInstanceManager<S> {
         // that is done
         store.start_subgraph_deployment(&logger).await?;
 
-        let (manifest, manifest_idx_and_name) = {
+        let (manifest, manifest_idx_and_name, static_data_sources) = {
             info!(logger, "Resolve subgraph files using IPFS");
 
             let mut manifest = SubgraphManifest::resolve_from_raw(
@@ -218,6 +218,8 @@ impl<S: SubgraphStore> SubgraphInstanceManager<S> {
 
             info!(logger, "Successfully resolved subgraph files using IPFS");
 
+            let static_data_sources = manifest.data_sources.clone();
+
             // Add dynamic data sources to the subgraph
             manifest.data_sources.extend(data_sources);
 
@@ -227,8 +229,11 @@ impl<S: SubgraphStore> SubgraphInstanceManager<S> {
                 manifest.data_sources.len()
             );
 
-            (manifest, manifest_idx_and_name)
+            (manifest, manifest_idx_and_name, static_data_sources)
         };
+
+        let static_filters =
+            self.static_filters || manifest.data_sources.len() >= ENV_VARS.static_filters_threshold;
 
         let onchain_data_sources = manifest
             .data_sources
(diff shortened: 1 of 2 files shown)


core/src/subgraph/instance_manager.rs
```diff
@@ -230,6 +230,9 @@ impl<S: SubgraphStore> SubgraphInstanceManager<S> {
             (manifest, manifest_idx_and_name)
         };
 
+        let static_filters = self.static_filters
+            || (manifest.network_name() == "bsc" && manifest.templates.len() > 10000);
```
lutter, line 234: I think it would make sense to do this for any network.

lutter, line 234: Isn't `manifest.templates.len()` just the number of templates in the manifest? We want to count actual dynamic data sources, so want to look at `data_sources.len()` ?

mangas: Yeah I was trying to avoid including the data_sources but you're right, it's prolly a few in 10000 so that should be fine, I'll change this

mangas: actually this also uses the previously stored data sources, so I need to trim those as well

leoyvens:
Looking at how static filters are currently implemented, this might not have the intended effect. Because it currently will add the filters from the templates but it doesn't prevent the filters from the dynamic data sources from being added.

When static filters are enabled, we need the final filter to be static data sources + templates, ignoring the stored dynamic data sources.

core/src/subgraph/instance_manager.rs
```diff
@@ -218,18 +218,28 @@ impl<S: SubgraphStore> SubgraphInstanceManager<S> {
...
             info!(logger, "Successfully resolved subgraph files using IPFS");
 
-            // Add dynamic data sources to the subgraph
-            manifest.data_sources.extend(data_sources);
-
             info!(
                 logger,
                 "Data source count at start: {}",
                 manifest.data_sources.len()
             );
```
leoyvens, line 225: We do want this log to include the count of dynamic data sources.

core/src/subgraph/instance_manager.rs
```diff
@@ -218,18 +218,28 @@ impl<S: SubgraphStore> SubgraphInstanceManager<S> {
...
+            // Add dynamic data sources to the subgraph
+            manifest.data_sources.extend(data_sources);
+
+            (manifest, manifest_idx_and_name, static_data_sources)
         };
 
+        let dds_count = manifest
+            .data_sources
+            .len()
+            .abs_diff(static_data_sources.len());
```
leoyvens, line 238: I wouldn't bother doing the diff and would just use the total len but this is also fine.

core/src/subgraph/instance_manager.rs
mangas: @leoyvens thoughts about this?

core/src/subgraph/instance_manager.rs
```diff
@@ -244,18 +254,30 @@ impl<S: SubgraphStore> SubgraphInstanceManager<S> {
...
+        // if static_filters is enabled, build a minimal filter with the static data sources and
+        // add the necessary filters based on templates.
+        // if not enabled we just stick to the filter based on all the data sources.
+        // This changes specifically removes dynamic data sources based filters because these can be derived
+        // from templates AND this reduces the cost of egress traffic by making the payloads smaller.
+        let filter = if static_filters {
+            let onchain_data_sources = static_data_sources
+                .iter()
+                .filter_map(|d| d.as_onchain().cloned())
+                .collect::<Vec<_>>();
```
leoyvens, line 266: No need to `collect` if `iter()` will be called right after.

core/src/subgraph/instance_manager.rs
```diff
@@ -244,18 +254,30 @@ impl<S: SubgraphStore> SubgraphInstanceManager<S> {
             .with_context(|| format!("no chain configured for network {}", network))?
             .clone();
 
-        // Obtain filters from the manifest
-        let mut filter = C::TriggerFilter::from_data_sources(onchain_data_sources.iter());
+        // if static_filters is enabled, build a minimal filter with the static data sources and
+        // add the necessary filters based on templates.
+        // if not enabled we just stick to the filter based on all the data sources.
+        // This changes specifically removes dynamic data sources based filters because these can be derived
```
leoyvens, line 260:
```suggestion
        // This specifically removes dynamic data sources based filters because these can be derived
```

leoyvens (approved):

PR #4010, opened 2022-09-29
store: Use a different locking strategy for deployments

lutter:
We used 'select for update' to lock a deployment against concurrent writes. It seems that that has bad consequences for performance in general; advisory locks are more lightweight than locking table rows, hopefully using them here will help write performance.

### DIFF
--- a/store/postgres/src/advisory_lock.rs
+++ b/store/postgres/src/advisory_lock.rs
@@ -11,6 +11,8 @@
 //! We use the following 2x 32-bit locks
 //!   * 1, n: to lock copying of the deployment with id n in the destination
 //!           shard
+//!   * 2, n: to lock the deployment with id n to make sure only one write
+//!           happens to it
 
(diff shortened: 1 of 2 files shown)


store/postgres/src/advisory_lock.rs
```diff
@@ -44,3 +46,10 @@ pub(crate) fn unlock_copying(conn: &PgConnection, dst: &Site) -> Result<(), Stor
         .map(|_| ())
         .map_err(StoreError::from)
 }
+
+pub(crate) fn lock_deployment_xact(conn: &PgConnection, site: &Site) -> Result<(), StoreError> {
+    sql_query(&format!("select pg_advisory_xact_lock(2, {})", site.id))
```
leoyvens, line 51: Does the `2` mean anything here?

leoyvens (approved):

PR #4012, opened 2022-09-30
Code reuse between `graphman` and `graph-node`

neysofu:
https://github.com/graphprotocol/graph-node/pull/3079 introduced some code duplication between `manager/run.rs` and `main.rs`, the intention was to fix things in an upcoming PR but we never came around to it.


tilacog (approved):
I really liked the functions being split now.
Thanks for this!

PR #4016, opened 2022-10-01
docs: fix dead links

wp-lai:
fix dead links of `Schema API` & `Query API` in `getting-started.md` file


leoyvens (approved):
These docs should probably live only on the website, but this is better than a broken link, thanks!

Issue #4014, opened 2022-09-30
File data source duplicates may not be detected after the file is found


leoyvens:
Currently data source duplication is checked against the in-memory list of data sources. If the file is found, we currently remove the file data source in the DB, so if the subgraph is restarted it won't be present in the list and duplicates will not be detected.

The bug can be reproduced by:
1. Create a file data source.
2. The file is found.
3. Restart the subgraph.
4. Create a duplicate of the file data source.
5. The duplicate is not detected.

We can fix this by not removing the data source when the file is found, this way it will always be in memory for the duplicate check. And track the done status through an additional column on `data_sources$`, such as `done bool not null`.

To add the column we'll need to introduce a migration runner when starting the subgraph, migrations would be tied to the `deployment_schemas.version`, and the migration runner will run the necessary migrations to bring the schema up to the latest and bump the version.

PR #4018, opened 2022-10-03
fix(offchain): detect offchain datasource duplication

poonai:
Closes: #4014 

I've tried to tackle the issue 😅 

Notes:
- I've implemented schema migrator for subgraph, which upgrades to the latest schema based on schema version
- instead of removing the offchain, now status is updated to the `done` column

 Will test it properly and update here. Meanwhile, early feedback would be really appreciated.


leoyvens:
There is a test case which is important to add:
1. Create a file data source A.
2. Trigger A, making it done.
3. Create a file data source B, with the same source CID as A, but a different context. So they share the same source but are not duplicates.
4. Trigger B.
5. Ensure A was not triggered again.

I believe this test will currently fail. That is because if A is done we are currently not monitoring it, which is a good optimization but doesn't entirely prevent it from triggering. This can be fixed by returning `None` in `fn match_and_decode` when the data source is done.
https://github.com/graphprotocol/graph-node/blob/d445e92f8b69bc38d168783d9f61ce1f7b705f44/graph/src/data_source/offchain.rs#L60

store/postgres/tests/subgraph.rs
```diff
@@ -1060,3 +1067,85 @@ fn fail_unfail_non_deterministic_error_noop() {
...
+    // run the migration for v2.
+    STORE_RUNTIME.handle().block_on(async {
+        SUBGRAPH_STORE.clear_caches();
+        SUBGRAPH_STORE.run_migration(&deployment.hash).unwrap();
+    });
+
+    // assert whether done column added to the datasource table.
+    run_test_with_conn(|conn| {
+        let query = format!(
+            "SELECT column_name
```
leoyvens, line 1114: Nice work on this test! Lets also select and test `is_nullable`, `column_default ` and if possible `data_type` here.

leoyvens (changes requested):

tests/tests/runner.rs
```diff
@@ -163,13 +168,39 @@ async fn file_data_sources() {
...
+        .unwrap();
+
+    assert_eq!(
+        query_res,
+        Some(object! { ipfsFile1: object!{ id: id , content: "[]" } })
+    );
+
+    ctx.provider.stop(ctx.deployment.clone()).await.unwrap();
+    let stop_block = test_ptr(4);
+    ctx.start_and_sync_to(stop_block).await;
```
leoyvens, line 203: You check the data source count again at the end of the test, and that the data sources are done.

leoyvens (changes requested):

store/postgres/src/subgraph_migrator.rs
```diff
@@ -0,0 +1,67 @@
...
+        return SubgraphMigrator {
+            current_version: site.schema_version,
+            subgraph_store: &subgraph_store,
+            store,
+            site,
+        };
+    }
+
+    pub fn run(mut self) -> Result<(), StoreError> {
+        // upgrade the schema untill it reaches the lastest version.
```
leoyvens, line 31:
```suggestion
        // upgrade the schema until it reaches the latest version.
```

tests/tests/runner.rs
```diff
@@ -163,13 +166,49 @@ async fn file_data_sources() {
...
+    let writebable = ctx
```
leoyvens, line 202:
```suggestion
    let writable = ctx
```

graph/src/components/subgraph/instance.rs
```diff
@@ -51,7 +51,7 @@ impl<C: Blockchain> BlockState<C> {
             deterministic_errors,
             created_data_sources,
             handler_created_data_sources,
-            offchain_to_remove,
+            processed_data_sources: processed_data_source,
```
leoyvens, line 54:
```suggestion
            processed_data_sources,
```

graph/src/components/subgraph/instance.rs
```diff
@@ -61,7 +61,7 @@ impl<C: Blockchain> BlockState<C> {
         }
         deterministic_errors.extend(other.deterministic_errors);
         entity_cache.extend(other.entity_cache);
-        offchain_to_remove.extend(other.offchain_to_remove);
+        processed_data_source.extend(other.processed_data_sources);
```
leoyvens, line 64:
```suggestion
        processed_data_sources.extend(other.processed_data_sources);
```

tests/tests/runner.rs
```diff
@@ -163,13 +166,49 @@ async fn file_data_sources() {
...
+        .store
+        .clone()
+        .writable(ctx.logger.clone(), ctx.deployment.id.clone())
+        .await
+        .unwrap();
+    let data_sources = writebable.load_dynamic_data_sources(vec![]).await.unwrap();
```
leoyvens, line 208: Lets check `data_sources.len()` here.

leoyvens (changes requested):
Just a few minor details

tests/tests/runner.rs
poonai: lol I was about to change, you caught it

store/postgres/src/subgraph_migrator.rs
```diff
@@ -0,0 +1,56 @@
...
+            table = DataSourcesTable::table_name(&self.site.namespace)
+        );
+        self.store.execute_raw(migration)?;
+        self.update_site_version(DeploymentSchemaVersion::V2)?;
+        Ok(())
+    }
+
+    fn update_site_version(&mut self, version: DeploymentSchemaVersion) -> Result<(), StoreError> {
+        let primary = self.subgraph_store.primary_conn()?;
+        primary.update_site_version(&self.site.deployment, &version)?;
```
lutter, line 52: I would prefer if `SubgraphStore` had a `pub(crate)` method `update_version` so that it continues to be the one place where the primary is manipulated (I know this is aspirational since `WritableStore` goes against that, but I would really like to make `primary_conn` a private method at some point)

store/postgres/src/writable.rs
```diff
@@ -63,6 +65,11 @@ impl WritableSubgraphStore {
     fn load_deployment(&self, site: &Site) -> Result<SubgraphDeploymentEntity, StoreError> {
         self.0.load_deployment(site)
     }
+
+    // returns ref of SubgraphStore
+    fn inner(&self) -> &SubgraphStore {
+        &self.0
+    }
```
lutter, line 72: The idea behind `WritableSubgraphStore` was to make it clear what functionality of a `SubgraphStore` gets exposed to the `WritableStore`. Instead of exposing `self.0` wholesale, add method(s) that expose the specific `SubgraphStore` functionality.

store/postgres/src/subgraph_migrator.rs
```diff
@@ -0,0 +1,56 @@
...
+    fn migrate_to_v2(&mut self) -> Result<(), StoreError> {
+        let migration = format!(
+            "
+        alter table {table}
+        add done bool not null default false;
+        ",
+            table = DataSourcesTable::table_name(&self.site.namespace)
+        );
+        self.store.execute_raw(migration)?;
+        self.update_site_version(DeploymentSchemaVersion::V2)?;
```
lutter, line 46: If we get unlucky and `graph-node` gets stopped in between committing `execute_raw` but not yet committing `update_site_version`, we have a subgraph that can't be started again since the migration will fail when it runs a second time. `migrate_to_v2` therefore needs to be idempotent, something like `alter table {table} add column if not exists ...` should do the trick (but please tests this manually by running that statement twice for an example)

lutter (commented):
The general approach looks fine to me. Alternatively, this could be done as a normal diesel migration that looks at
```
select table_schema from information_schema.tables where table_name = 'data_sources$';
```

That would be a lot simpler in terms of managing this migration; [this migration](https://github.com/graphprotocol/graph-node/blob/master/store/postgres/migrations/2022-03-09-07000_call_cache_block_number_idx/up.sql#L6-L20) shows how to run SQL commands for each entry found in some other table

lutter (changes requested):
Meant to set this to 'Request changes' instead of 'Comment'

store/postgres/src/subgraph_migrator.rs
```diff
@@ -0,0 +1,56 @@
...
+            self.migrate_to_v2()?;
+        }
+        Ok(())
+    }
+
+    fn migrate_to_v2(&mut self) -> Result<(), StoreError> {
+        let migration = format!(
+            "
+        alter table {table}
+        add done bool not null default false;
```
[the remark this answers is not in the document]
leoyvens: On the name, I find "processed" a bit ambiguous between partially processed or fully processed. Also "an onchain data source is never processed" doesn't sound right. Open to other ideas: `finished`, `completed`, `finalized`...

I agree that we can make this this `done_at`, but store the block number on which it became done rather than the timestamp.

poonai:
@lutter yess, I agree it will be simpler to use diesel migration script without using additional code. 

Here is the migration script to run migration without using migration runner. 

```
do $$
declare 
   deployments cursor for select * from deployment_schemas where version = 1;
begin
   for deployment in deployments loop
      execute 'alter table ' || deployment.name || '.data_sources$ add done_at int';
      execute 'update deployment_schemas set version = 2 where id = ' || deployment.id;
   end loop;
end;
$$;

```

do let me know what approach we can take, I'll make changes accordingly 

cc: @leoyvens , @lutter

leoyvens:
If that script works, I'm 👍 on using regular migrations for this!

lutter:
@poonai The migration you sketched out above won't work because of sharding: `deployment_schemas` exists in all shards, but is only really maintained in the primary shard ([docs](https://github.com/graphprotocol/graph-node/blob/master/docs/implementation/metadata.md)) while the `data_sources$` tables live in each shard.

Instead of relying on `deployment_schemas` to find `data_sources$` tables that need migrating, we'll have to use the `information_schema` catalog to find `data_sources$` tables that don't have a `done_at` column yet, and we'll also update `deployment_schemas.version`. This migration does that:
```sql
-- Change the tables in each shard for each deployment that needs it
do $$
declare
  deployments cursor for
     select t.table_schema as sgd
       from information_schema.tables t
      where t.table_schema like 'sgd%'
        and t.table_name = 'data_sources$'
        and not exists (select 1 from information_schema.columns c
                         where c.table_name = t.table_name
                           and c.table_schema = t.table_schema
                           and c.column_name = 'done_at');
begin
  for d in deployments loop
    execute 'alter table ' || d.sgd || '.data_sources$ add done_at int';
  end loop;
end;
$$;

-- This will only do something for real in the primary, but is fine to do
-- in all shards
update deployment_schemas
   set version = 2
 where version = 1;
```

To test this, run a `graph-node` (without the migration) and deploy a few subgraphs to it; they don't need to be synced, just run them for a few minutes, then kill `graph-node`. Save the above SQL script somewhere, say `/tmp/up.sql`, then connect to the `graph-node` database with `psql` and do
```
mydb=# begin; -- start a transaction
mydb=# \i /tmp/up.sql
mydb=# -- poke around the database to make sure everything has been set up correctly
mydb=# rollback;
```

With the `rollback`, it is possible to make changes to the SQL script if there are problems and repeat the migration a few times.

For the down migration, I am not terribly concerned about that; you could just make it
```sql
raise 'This migration is irreversible';
```
or you could change the SQL script above to drop the `done_at` column and set `version = 1`. Either is fine by me.

store/postgres/src/subgraph_migrator.rs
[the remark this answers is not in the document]
lutter: I am totally fine with calling this `finished` or the like. Even better might be a comment on `DataSource.done_at` that explains how that flag is used, and what it means. If we have a clear explanation there, I am also fine with `done`.

The reason I wanted a timestamp there rather than the block number is so that indexers can get a sense for how long ago (in wall time) that data source was processed. The motivation for this is that in some situations it would be quite useful with a subgraph's `synced` status to see how long ago that was reached. If storing a block number there implies that that was the block number that was the chain head at the time the data source was finished.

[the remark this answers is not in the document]
leoyvens: I don't feel strongly on the name. @poonai do you have opinions? Agreed on improving the comments.

While I don't expect the block number to ever be deterministic it might be more reproducible and comparable across indexers than timestamps. Anyways this value might never be used at all, it isn't yet clear how the indexing status of offchain data will be reported, so lets not overthink.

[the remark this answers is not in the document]
poonai: I really don't have any strong opinion on this.

leoyvens (approved):

PR #4026, opened 2022-10-04
Revert "fix: Allow grafts to add data sources (#3989)"

leoyvens:
This reverts commit 3b3b458ecabffcc1006a82bacf2df7f17f2859e5. It's causing existing grafted subgraphs to stop indexing.

Just the migration to add the raw_yaml colum is kept.


leoyvens:
After #4027 hopefully this won't be necessary.

leoyvens:
This was fixed without the need of a revert.

PR #4030, opened 2022-10-04
Fix cross shard graft

leoyvens:
First commit reverts most of https://github.com/graphprotocol/graph-node/commit/0c4417e50da3ea40d492ee2373bbfdf82ec09f0f, which was misguided since the graft code is already only called if the graft is pending.

Second commit fixes cross shard grafting, by not assuming that `load_deployment` will succeed on the same `DeploymentStore` as the one in which the source subgraph is deployed. 


lutter (approved):
Nice!

PR #4035, opened 2022-10-06
graphman: new command to hastly delete a subgraph

tilacog:
Introduces a new graphman command to promptly delete all data from a deployment, even if it is running.

TODO: 
- [x] Decide the name and placement for this command
- [x] Documentation

### DIFF
--- a/node/src/bin/manager.rs
+++ b/node/src/bin/manager.rs
@@ -247,13 +247,13 @@ pub enum Command {
     Drop {
         /// The deployment identifier
         deployment: DeploymentSearch,
-        /// List only current version
+        /// Search only for current version
         #[clap(long, short)]
         current: bool,
-        /// List only pending versions
+        /// Search only for pending versions
         #[clap(long, short)]
         pending: bool,
-        /// List only used (current and pending) versions
+        /// Search only for used (current and pending) versions
         #[clap(long, short)]
         used: bool,
         /// Skip confirmation prompt
--- a/node/src/bin/manager.rs
+++ b/node/src/bin/manager.rs
@@ -869,7 +869,7 @@ async fn main() -> anyhow::Result<()> {
                 }
             }
         }
-        Remove { name } => commands::remove::run(ctx.subgraph_store(), name),
+        Remove { name } => commands::remove::run(ctx.subgraph_store(), &name),
         Create { name } => commands::create::run(ctx.subgraph_store(), name),
         Unassign { deployment } => {
             let sender = ctx.notification_sender();
--- a/node/src/bin/manager.rs
+++ b/node/src/bin/manager.rs
@@ -256,7 +256,7 @@ pub enum Command {
         /// List only used (current and pending) versions
         #[clap(long, short)]
         used: bool,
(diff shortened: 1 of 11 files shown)


node/src/bin/manager.rs
```diff
@@ -237,6 +237,29 @@ pub enum Command {
...
+        /// List only current version
+        #[clap(long, short)]
+        current: bool,
+        /// List only pending versions
+        #[clap(long, short)]
+        pending: bool,
+        /// List only used (current and pending) versions
+        #[clap(long, short)]
+        used: bool,
+        /// Skips confirmation prompt
```
lutter, line 259: Should be `Skip` to stay in the imperative style of the rest of the help text.

node/src/bin/manager.rs
```diff
@@ -1097,6 +1120,31 @@ async fn main() -> anyhow::Result<()> {
...
+            pending,
+            used,
+            force,
+        } => {
+            let sender = ctx.notification_sender();
+            let (store, mut pools) = ctx.store_and_pools();
+            let subgraph_store = store.subgraph_store();
+            let primary_pool = pools
+                .remove(&*PRIMARY_SHARD)
+                .expect("there is a primary pool");
```
lutter, line 1135: There's a `ctx.store_and_primary()` which will do most of this for you

node/src/manager/commands/drop.rs
```diff
@@ -0,0 +1,74 @@
+use crate::manager::{
+    deployment::{Deployment, DeploymentSearch},
+    display::List,
+    prompt::prompt_for_confirmation,
+};
+use graph::anyhow::{self, bail};
+use graph_store_postgres::{connection_pool::ConnectionPool, NotificationSender, SubgraphStore};
+use std::sync::Arc;
+
+/// Finds, unassings, record and remove matching deployments.
```
lutter, line 10: type: 'unassigns'

node/src/manager/commands/info.rs
```diff
@@ -5,7 +5,7 @@ use graph_store_postgres::{connection_pool::ConnectionPool, Store};
 
 use crate::manager::deployment::{Deployment, DeploymentSearch};
 
-fn find(
+pub fn find(
```
lutter, line 8: Might be better to make this a method on `DeploymentSearch` if it's used in more than one place

lutter (approved):
Nice! I really like the user interaction. Just address the few comments I left, and then this is good to go.

As for the command name: after looking at the commands, I agree it's much better to make this a new command than to add it to `graphman remove`.

node/src/bin/manager.rs
tilacog: Fixed in 7ff6aa8ba

node/src/bin/manager.rs
tilacog: Fixed in 9ccee3a82

node/src/manager/commands/info.rs
tilacog: Fixed in b2adf2922

Issue #4005, opened 2022-09-28
Bug: wrong start_block when subgraph is removed and added again


sduchesneau:
# This is a *BUG*
Reproduced on current master #750730ce1

# Behavior:
When a subgraph is removed and cleaned up, adding the same subgraph again will result in it starting at block 0 instead of the proper start block.

# How to reproduce

1. Launch a graph-node on an empty postgres database, pointing to thegraph ipfs for simplicity `--ipfs https://api.thegraph.com/ipfs/`

2. Use `httpie` to simplify adding a deployment from ipfs hash:

```
dep=QmXU1g4ju2zNBmDanMBHHuwJyDRG7pCjTK9DdAPvXqfQc4
http -I post http://localhost:8020/ jsonrpc="2.0" id="1" method="subgraph_create" params:="{\"name\": \""$dep"\"}" && http -I post http://localhost:8020/ jsonrpc="2.0" id="1" method="subgraph_deploy" params:="{\"name\": \""$dep"\", \"ipfs_hash\": \""$dep"\", \"version_label\": \"1\"}"; done
```

3. Observe the correct start block in the logs:
`INFO Scanning blocks [10647740, 10647740], range_size: 1, sgd: 1, subgraph_id: QmXU1g4ju2zNBmDanMBHHuwJyDRG7pCjTK9DdAPvXqfQc4, component: BlockStream`

4. Use `graphman` to remove this subgraph completely:

```
graphman remove QmXU1g4ju2zNBmDanMBHHuwJyDRG7pCjTK9DdAPvXqfQc4 && sleep 2
graphman unused record && sleep 2
graphman unused remove
```

5. Deploy the same subgraph once more

```
dep=QmXU1g4ju2zNBmDanMBHHuwJyDRG7pCjTK9DdAPvXqfQc4
http -I post http://localhost:8020/ jsonrpc="2.0" id="1" method="subgraph_create" params:="{\"name\": \""$dep"\"}" && http -I post http://localhost:8020/ jsonrpc="2.0" id="1" method="subgraph_deploy" params:="{\"name\": \""$dep"\", \"ipfs_hash\": \""$dep"\", \"version_label\": \"1\"}"; done
```

6. Observe the graph-node logs, baffled, as it starts at block 0 with this line `Scanning blocks [0, 0]`

```
[...]

PR #4044, opened 2022-10-07
Handle rapid deletion and recreation of subgraphs more gracefully

lutter:
In development and similar settings, often a subgraph gets deleted, completely removed and then immediately redeployed. This usually fails because of some internal caches that `graph-node` keeps. This PR removes these failures on indexing nodes. If there are dedicated query nodes, they will still not pick up the deployment until their internal caches get updated (by default in at most 5 minutes) Installations that use a single node, or when the main concern is the index node, will no longer be affected by this cache issue.

### DIFF
--- a/store/postgres/src/subgraph_store.rs
+++ b/store/postgres/src/subgraph_store.rs
@@ -1264,6 +1264,16 @@ impl SubgraphStoreTrait for SubgraphStore {
         Ok(writable)
     }
 
+    async fn stop_subgraph(&self, loc: &DeploymentLocator) -> Result<(), StoreError> {
+        // Remove the writable from the cache and stop it
+        let deployment = loc.id.into();
+        let writable = self.writables.lock().unwrap().remove(&deployment);
+        match writable {
+            Some(writable) => writable.stop().await,
+            None => Ok(()),
+        }
+    }
+
     fn is_deployed(&self, id: &DeploymentHash) -> Result<bool, StoreError> {
         match self.site(id) {
             Ok(_) => Ok(true),
--- a/store/postgres/src/subgraph_store.rs
+++ b/store/postgres/src/subgraph_store.rs
@@ -373,6 +373,20 @@ impl SubgraphStoreInner {
         Ok(site)
     }
 
+    fn evict(&self, id: &DeploymentHash) -> Result<(), StoreError> {
+        if let Some((site, _)) = self.sites.remove(id) {
+            let store = self.stores.get(&site.shard).ok_or_else(|| {
+                constraint_violation!(
+                    "shard {} for deployment sgd{} not found when evicting",
+                    site.shard,
+                    site.id
+                )
+            })?;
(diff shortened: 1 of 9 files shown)


lutter:
This also fixes #4005

store/postgres/src/subgraph_store.rs
```diff
@@ -1264,6 +1264,16 @@ impl SubgraphStoreTrait for SubgraphStore {
         Ok(writable)
     }
 
+    async fn stop_subgraph(&self, loc: &DeploymentLocator) -> Result<(), StoreError> {
+        // Remove the writable from the cache and stop it
+        let deployment = loc.id.into();
+        let writable = self.writables.lock().unwrap().remove(&deployment);
```
tilacog, line 1288: If I got it right [docs](https://doc.rust-lang.org/std/sync/struct.Mutex.html#method.lock) right, this would only panic due to a programming error, so it is relatively safe to unwrap here.
Is that correct?

store/postgres/src/writable.rs
```diff
@@ -438,10 +439,16 @@ enum Request {
         block_ptr: BlockPtr,
         firehose_cursor: FirehoseCursor,
     },
+    Stop,
+}
+
+enum ExecResult {
+    Continue,
+    Stop,
```
tilacog, line 447: Just a suggestion, so feel free to ignore this comment.

This enum somewhat similar to the [ControlFlow](https://doc.rust-lang.org/std/ops/enum.ControlFlow.html) from `std`. 

While we don't need to return any data out of this, maybe we could redefine it as
```rust
type ExecResult = ControlFlow<()>
```

I know this won't change anything, but it would make room if we want to return data from those contexts in the future.

store/postgres/src/subgraph_store.rs
```diff
@@ -499,6 +513,8 @@ impl SubgraphStoreInner {
         #[cfg(not(debug_assertions))]
         assert!(!replace);
 
+        self.evict(&schema.id)?;
```
tilacog, line 516: Is the reason we call `evict` on deployment creation so we have more confidence that the cache will be cleared?

tilacog (approved):
LGTM.
I left a few questions for me to understand a little more about how deployment handling goes.

store/postgres/src/subgraph_store.rs
lutter: It would panic if the lock is poisoned, i.e., if some other thread paniced while holding the lock. In that case, it's anybody's guess what's happening and panicing here is really the only thing we can do.

store/postgres/src/subgraph_store.rs
lutter: Yes, to make sure that we don't have a stale entry for the same hash in the cache (which could happen if a previous deployment of the same hash was stopped and then redeployed) This is really redundant with the eviction in `stop_subgraph` but since it's a cheap operation seemed safer to do it in both places.

PR #4055, opened 2022-10-11
Deduplicate triggers

evaporei:


graph/src/blockchain/block_stream.rs
```diff
@@ -173,9 +173,22 @@ where
 }
 
 impl<C: Blockchain> BlockWithTriggers<C> {
-    pub fn new(block: C::Block, mut trigger_data: Vec<C::TriggerData>) -> Self {
+    pub fn new(block: C::Block, mut trigger_data: Vec<C::TriggerData>, logger: &Logger) -> Self {
```
mangas, line 176: This to me sounds like a lot of logic for a `new` function. I'd rather change this to something from_trigger_data or something with a comment saying it will dedup etc.

graph/src/blockchain/block_stream.rs
```diff
@@ -173,9 +173,22 @@ where
...
         // This is where triggers get sorted.
         trigger_data.sort();
+
+        let old_len = trigger_data.len();
```
mangas, line 180: https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.dedup_with_count

mangas: I think this could be useful here

graph/src/blockchain/block_stream.rs
evaporei: I just added a doc comment, what do you think?

graph/src/blockchain/block_stream.rs
evaporei: Wow this is pretty cool! However I'll not use it because I only need to log the count previously and after, I think having to parse the result of `dedup_with_count` will be overkill.

graph/src/blockchain/block_stream.rs
mangas: I don't feel strongly about it but I feel like reading the code you would not expect new to do the sort + dedup etc. The doc helps though I'll leave the rest up to you.

graph/src/blockchain/block_stream.rs
mangas: nvm you're right, it's not useful here

mangas (approved):

NEWS.md
```diff
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+#### Upgrade notes
+
+- This release includes a **determinism fix** that should affect very few subgraphs on the network (currently only two). There was an issue that if a subgraph manifest had one data source with no contract address, listening to the same events or calls of another data source that has a specified address, the handlers for those would be called twice. With the fix, this will happen no more, the handler will be called just once like it should.
+  - The two affected deployments are: `Qmccst5mbV5a6vT6VvJMLPKMAA1VRgT6NGbxkLL8eDRsE7` and `Qmd9nZKCH8UZU1pBzk7G8ECJr3jX3a2vAf3vowuTwFvrQg`;
+  - Here's an example [manifest](https://ipfs.io/ipfs/Qmd9nZKCH8UZU1pBzk7G8ECJr3jX3a2vAf3vowuTwFvrQg), taking a look at the data sources of name `ERC721` and `CryptoKitties`, both listen to the `Transfer(...)` event. Considering a block where there's only one occurence of this event, `graph-node` would duplicate it and call `handleTransfer` twice. Now this is fixed and it will be called only once per event/call that happened on chain.
```
leoyvens, line 9: Since this is mentioning the network impact it's important to note any impact to indexers and any actions they should take.

evaporei: Awesome, thanks for the feedback I'll address it :slightly_smiling_face:

evaporei: I just pushed what they should do, what do you think? @leoyvens

leoyvens: lgtm

PR #4056, opened 2022-10-11
datasource event parser

mangas:


leoyvens:
@mangas Do you intend to merge or close this? I think something like this would be useful, but we'd want to think through something that's sufficiently general.

PR #4058, opened 2022-10-12
Nested Sorting: interface by non-derived child entity

kamilkisiela:
Sorting an interface by child-level entity (derived) is not supported yet


lutter:
Is this ready for review or should I wait until you've done your refactor?

kamilkisiela:
@lutter wait until refactor, I will @mention you

graphql/src/store/query.rs
```diff
@@ -561,22 +561,47 @@ fn build_order_by(
...
             OrderByValue::Child(parent_field_name, child_field_name) => {
-                if entity.is_interface() {
-                    return Err(QueryExecutionError::OrderByNotSupportedError(
-                        entity.name().to_owned(),
-                        parent_field_name,
-                    ));
-                }
+                // Finds the field that connects the parent entity with the child entity.
+                // In the case of an interface, we need to find the field on one of the types that implement the interface,
+                // as the `@derivedFrom` directive is only allowed on object types.
```
lutter, line 566: The crazy thing here is that AFAIK it's possible for different object types that implement the same interface to store that attribute directly or derived. I have no idea how well that actually works in practice, but I don't think we enforce anything there.

I don't think we should hold this PR up any longer because of that, but it's likely that this will lead to issues if anybody uses that somewhat crazy mix of derived and direct storage. I have no idea if anybody does that, and we should have a look at the specific subgraph if that ever leads to issues. It might be a good idea though to require that all implementers of the interface use the same storage and return an error otherwise just to make sure we don't inadvertently return incorrect data.

lutter (approved):
Looks good to me. I like how you avoided passing that somewhat mysterious boolean flag around.

To me, the most important thing is that this won't change the behavior of queries that work today - I _think_ that is true, and I don't have a great idea how we could test that more rigorously. The existing query tests should ensure that.

This is good to be merged after a rebase, and addressing the comment about making sorting by an interface where the field has mixed storage in implementers an error. I think it would lead to a SQL error anyway, but an explicit error would be clearer.

dotansimha (approved):

gatleas17 (commented):
Hi

PR #4061, opened 2022-10-12
Single-source crate versions

neysofu:
Uses the Cargo 1.64 feature that allows one to inherit crate metadata from the workspace root's `Cargo.toml` file.

The new script `release.sh` is a refactor of the old `update-cargo-version.sh`, with similar capabilities. Tested on macOS and Linux.


neysofu:
Closes #4061.

scripts/release.sh
```diff
@@ -0,0 +1,111 @@
...
+
+abort () {
+  local FAIL_MSG=$@
+  echo "$FAIL_MSG"
+  exit 1
+}
+
+abort_failed_to_update () {
+  local FILE_NAME=$@
+  abort "💀 Failed to update $FILE_NAME. Aborting."
```
evaporei, line 14: I like the 💀 use here 😆

evaporei (approved):
💯 ❤️

PR #4065, opened 2022-10-14
refactor(offchain): Add causality region to DataSource

leoyvens:
The goal of this PR is for the in-memory representation of an offchain data source to know its causality region. This is in preparation of adding the causality region to entities.

Currently the column exists in the DB and the causality region is assigned at insert time. But we need the causality region to be assigned at creation time. So instead of using a DB sequence, this plumbs the `causality_region_next_value` through various places, so it can be known and bumped when a data source is created.

With the causality region also being added to `StoredDynamicDataSource`, it turned out the `is_offchain` field became redundant and could be removed. Also some queries in `dynds/private.rs` could be simplified.

The data source trait was modified to have an associated `fn from_template_info` instead of using `TryFrom`. This change turned out not to be necessary but I guess it's nice to keep the interface symmetrical with offchain data sources. A lot of the diff in `chain/*` is just moving code due to this.


core/src/subgraph/context/instance.rs
```diff
@@ -166,4 +168,10 @@ where
     pub(super) fn hosts(&self) -> &[Arc<T::Host>] {
         &self.hosts
     }
+
+    pub(super) fn causality_region_next_value(&mut self) -> i32 {
```
mangas, line 172: can you elaborate a bit on this? Why do we need sequencial numbers as causality regions? 

I also question if this is the right place for this logic. If causality regions have logic to them perhaps adding it to the type would be a better place?

leoyvens:
By our discussions, I've added a `CausalityRegion` newtype. I tried encapsulating further the management of the causality region sequence, but that didn't quite work and would need a larger refactoring than I wanted to do in this PR.

core/src/subgraph/context.rs
```diff
@@ -140,6 +140,10 @@ impl<C: Blockchain, T: RuntimeHostBuilder<C>> IndexingContext<C, T> {
 
         Ok(host)
     }
+
+    pub fn causality_region_next_value(&mut self) -> CausalityRegion {
+        self.instance.causality_region_next_value()
```
mangas, line 145: Would it make sense for this to be part of the impl of CausalityRegion? If Onchain next() is Onchain, if Offchain(n) it's Offchain(n+1) or Offchain::default() if nothing is set (ie, the first one)?

core/src/subgraph/context/instance.rs
```diff
@@ -37,6 +39,7 @@ where
         host_builder: T,
         host_metrics: Arc<HostMetrics>,
         offchain_monitor: &mut OffchainMonitor,
+        causality_region_next_value: CausalityRegion,
```
mangas, line 42: from this I think I'd move the logic of next value to CausalityRegion, I think this next value is leaking the specifics of CR

core/src/subgraph/runner.rs
```diff
@@ -475,13 +477,26 @@ where
...
+                    DataSourceTemplate::Offchain(_) => offchain::DataSource::from_template_info(
+                        info,
+                        self.ctx.causality_region_next_value(),
+                    )
+                    .map(DataSource::Offchain),
+                };
+                match res {
+                    Ok(ds) => ds,
+                    Err(e @ DataSourceCreationError::Ignore(..)) => {
+                        warn!(self.logger, "{}", e.to_string());
```
mangas, line 495: you are logging this and returning the error with a message. Usually this message would be printed by the handler of the error. As we discussed, I think the Ignore error semantic leads to it being mishandled very easily if someone doesn't know the error is here. I'd rather keep the log message here and return None. If you think you absolutely need the Ignore enum with a message then I'd suggest dropping the warning the error will contain the same message.

graph/src/components/store/mod.rs
```diff
@@ -800,8 +801,8 @@ pub struct StoredDynamicDataSource {
     pub param: Option<Bytes>,
     pub context: Option<serde_json::Value>,
     pub creation_block: Option<BlockNumber>,
-    pub is_offchain: bool,
     pub done_at: Option<i32>,
+    pub causality_region: CausalityRegion,
```
mangas, line 805: A lot better :raised_hands:

graph/src/data_source/mod.rs
```diff
@@ -138,8 +125,32 @@ impl<C: Blockchain> DataSource<C> {
         match (self, other) {
             (Self::Onchain(a), Self::Onchain(b)) => a.is_duplicate_of(b),
             (Self::Offchain(a), Self::Offchain(b)) => {
+                let offchain::DataSource {
```
mangas, line 128: This could use a unit test

graph/src/data_source/offchain.rs
```diff
@@ -124,29 +128,40 @@ impl DataSource {
...
+        let StoredDynamicDataSource {
+            manifest_idx,
+            param,
+            context,
+            creation_block,
+            done_at,
+            causality_region,
+        } = stored;
+
+        let param = param.context("no param on stored data source")?;
```
mangas, line 149: Does this apply here? https://graphprotocol.slack.com/archives/C0385VCCSBC/p1663958185550939

store/postgres/src/dynds/mod.rs
```diff
@@ -76,3 +77,17 @@ pub(crate) fn update_offchain_status(
...
+pub(crate) fn causality_region_next_value(
+    conn: &PgConnection,
+    site: &Site,
+) -> Result<CausalityRegion, StoreError> {
+    match site.schema_version.private_data_sources() {
+        true => DataSourcesTable::new(site.namespace.clone()).causality_region_next_value(conn),
+
+        // Subgraphs on the legacy shared table do not use offchain data sources, so this value will
+        // never be used.
+        false => Ok(CausalityRegion::ONCHAIN),
```
mangas, line 91: This should definitely be handled by CausalityRegion :stuck_out_tongue:

core/src/subgraph/runner.rs
leoyvens: > you are logging this and returning the error with a message. 

Afaict this is logging and dropping the error after the `continue`.

graph/src/data_source/offchain.rs
leoyvens: I don't think the context methods are to blame for newlines in log messages, we just need to be sure to not print errors, or anything, with the debug format `{:?}`. The annoying thing about anyhow is remembering to use the alternate format specifier `{:#}` to print with causes.

store/postgres/src/dynds/mod.rs
leoyvens: Open to suggestions.

leoyvens:
@mangas Thanks for the detailed review!

A few of your comments are about using the `CausalityRegion` impl to better manage the sequence of causality regions. I'm not sure how to improve on that, I'm open to more detailed suggestions.

mangas (approved):
Approving to unblock, check comments

core/src/subgraph/runner.rs
mangas: Ah yeah don't mind me, you are correct :)

graph/src/data_source/mod.rs
leoyvens: Done

leoyvens:
I've added a `CausalityRegionSeq` to help clarify the sequence management, it helped move some logic out of the store.

store/postgres/src/dynds/mod.rs
```diff
@@ -78,16 +78,15 @@ pub(crate) fn update_offchain_status(
     }
 }
 
-/// The next causality region free to be assigned. Any higher number is also free to be assigned.
-pub(crate) fn causality_region_next_value(
+/// The maximum assigned causality region. Any higher number is therefore free to be assigned.
+pub(crate) fn causality_region_curr_val(
```
mangas, line 82: could probably call this something like latest_causality_region

leoyvens: This is just a nod to PG which uses this name for their equivalent sequence function https://www.postgresql.org/docs/current/functions-sequence.html

PR #4066, opened 2022-10-17
graphman command for clearing the call cache for a given chain

### LINKED ISSUE #3727: New `graphman` command for clearing the call cache for a given chain
We could expand the `graphman chain` command with new tasks that deal with the call cache.

For instance, the following command:

```
$ graphman chain call-cache remove <chain-name> --from <block-pointer> --to <block-pointer>
```

Would delete rows from the call cache between the blocks specified by the `--from/--to` block pointers.

If omitted, the absent pointers would be interpreted as open ends of an interval, potentially allowing the invoker to truncate the entire call cache.

poonai:
close #3727

### DIFF
--- a/docs/graphman.md
+++ b/docs/graphman.md
@@ -369,32 +369,50 @@ Inspect all blocks after block `13000000`:
 
 ### SYNOPSIS
 
-Execute operations on call cache
+Remove the call cache of the specified chain.
+
+If block numbers are not mentioned in `--from` and `--to`, then all the call cache will be removed.
 
 USAGE:
-    graphman chain call-cache <CHAIN_NAME> <SUBCOMMAND>
-
-ARGS:
-    <CHAIN_NAME>    Chain name (must be an existing chain, see 'chain list')
+    graphman chain call-cache <CHAIN_NAME> remove [OPTIONS]
 
 OPTIONS:
-    -h, --help    Print help information
+    -f, --from <FROM>
+            Starting block number
 
-SUBCOMMANDS:
-    help      Print this message or the help of the given subcommand(s)
-    remove    Remove call cache of a specified chain
+    -h, --help
+            Print help information
+
+    -t, --to <TO>
+            Ending block number
 
 ### DESCRIPTION
 
 Remove call cache of a specified chain
 
+### OPTIONS
+
+The `from` and `to` options can be used to decide the block range of the call cache that needs to be removed.
+
+#### `from`
+   
+    graphman --config config.toml chain call-cache ethereum remove --from 10
+
+#### `to`
+
+    graphman --config config.toml chain call-cache ethereum remove --to 10
+
+#### `from` and `to`
+
+    graphman --config config.toml chain call-cache ethereum remove --from 10 --to 50
+
 ### EXAMPLES
 
-Remove call cache by specifying its `from` and `to` block number
+Remove call cache by specifying its `from` and `to` block number:
 
     graphman --config config.toml chain call-cache ethereum remove --from 10 --to 20
 
-Remove all the call cache of the specified chain
+Remove all the call cache of the specified chain:
 
     graphman --config config.toml chain call-cache ethereum remove
 
--- a/docs/graphman.md
+++ b/docs/graphman.md
@@ -392,19 +392,17 @@ Remove the call cache of a specified chain.
 
 ### OPTIONS
 
-The `from` and `to` options can be used to decide the block range of the call cache that needs to be removed.
+The `from` and `to` options are used to decide the block range of the call cache that needs to be removed.
 
 #### `from`
-   
-    graphman --config config.toml chain call-cache ethereum remove --from 10
+
+The `from` option is used to specify the starting block number of the block range. In the absence of `from` option, the first block 
+number will be used as the starting block number.
 
 #### `to`
 
-    graphman --config config.toml chain call-cache ethereum remove --to 10
-
-#### `from` and `to`
-
-    graphman --config config.toml chain call-cache ethereum remove --from 10 --to 50
+The `to` option is used to specify the ending block number of the block range. In the absence of `to` option, the last block 
+number will be used as the ending block number.
 
 ### EXAMPLES
 
--- a/docs/graphman.md
+++ b/docs/graphman.md
@@ -6,7 +6,11 @@
 - [Unused Record](#unused-record)
 - [Unused Remove](#unused-remove)
 - [Drop](#drop)
+<<<<<<< HEAD
 - [Chain Check Blocks](#check-blocks)
+=======
+- [Chain Call Cache](#call-cache)
+>>>>>>> d0434b2b5 (add docs)
 
 <a id="info"></a>
 # ⌘ Info
@@ -289,6 +293,7 @@ Stop, unassign and delete all indexed data from a specific deployment by its sub
 
     graphman --config config.toml drop autor/subgraph-name
 
+<<<<<<< HEAD
 <a id="check-blocks"></a>
 # ⌘ Check Blocks
 
@@ -362,3 +367,34 @@ Inspect a block range, deleting any duplicated blocks:
 Inspect all blocks after block `13000000`:
 
     graphman --config config.toml chain check-blocks mainnet by-range --from 13000000
+=======
+<a id="call-cache"></a>
+# ⌘ Call Cache
+
+### SYNOPSIS
+
+Execute operations on call cache
+
+USAGE:
+    graphman chain call-cache <CHAIN_NAME> <SUBCOMMAND>
+
+ARGS:
+    <CHAIN_NAME>    Chain name (must be an existing chain, see 'chain list')
+
+OPTIONS:
+    -h, --help    Print help information
+
+SUBCOMMANDS:
+    help      Print this message or the help of the given subcommand(s)
+    remove    Remove call cache of a specified chain
+
+### DESCRIPTION
+
+Remove call cache of a specified chain
+
+### EXAMPLES
+
+Remove call cache by specifying its `from` and `to` block pointer.
+
+    graphman --config config.toml chain call-cache ethereum remove --from 10 --to 20
+>>>>>>> d0434b2b5 (add docs)
--- a/docs/graphman.md
+++ b/docs/graphman.md
@@ -388,7 +388,7 @@ OPTIONS:
 
 ### DESCRIPTION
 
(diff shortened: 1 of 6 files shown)


store/postgres/src/chain_store.rs
```diff
@@ -1028,6 +1028,69 @@ mod data {
...
+                        sql_query(query)
+                            .bind::<Integer, _>(from)
+                            .bind::<Integer, _>(to)
+                            .execute(conn)
+                            .map_err(Error::from)
+                    }
+                    (None, None) => {
+                        let query = format!("delete from {}", call_cache.qname);
+                        sql_query(query).execute(conn).map_err(Error::from)
+                    }
```
tilacog, line 1089: In this case, truncating the table would be faster.
Diesel doesn't have methods for `TRUNCATE` commands, but you can check [how we truncate the block cache](https://github.com/graphprotocol/graph-node/blob/2f9a82d21679614f768a7ac0d3c34587ff168eec/store/postgres/src/chain_store.rs#L1396-L1400) for reference.

store/postgres/src/chain_store.rs
```diff
@@ -1028,6 +1028,69 @@ mod data {
...
+                            .execute(conn)
+                            .map_err(Error::from),
+                        (Some(from), Some(to)) => delete_stmt
+                            .filter(cache::block_number.ge(from))
+                            .filter(cache::block_number.le(to))
+                            .execute(conn)
+                            .map_err(Error::from),
+                        (None, None) => delete_stmt.execute(conn).map_err(Error::from),
+                    }
+                }
```
tilacog, line 1057: I'd suggest mutating the `delete_stmt` value only once for each variant, and only calling `.execute()` at the end. 
This reduces code duplication, which helps with maintenance.

Also, since this function's error type is `anyhow::Error`, we can imbue extra details about possible errors. This would also occur only once in this scope.

```suggestion
            use anyhow::Context;
            match self {
                Storage::Shared => {
                    use public::eth_call_cache as cache;
                    let mut delete_stmt = diesel::delete(cache::table);
                    if let Some(from) = from {
                        delete_stmt.filter(cache::block_number.ge(from));
                    }
                    if let Some(to) = from {
                        delete_stmt.filter(cache::block_number.le(to));
                    }
                    delete_stmt.execute(conn).context(
                        "Failed to execute SQL to delete the block cache in shared storage.",
                    )
                }
```

store/postgres/src/chain_store.rs
```diff
@@ -1028,6 +1028,69 @@ mod data {
...
+                        let query = format!(
+                            "delete from {} where block_number >= $1 and block_number <= $2",
+                            call_cache.qname
+                        );
+                        sql_query(query)
+                            .bind::<Integer, _>(from)
+                            .bind::<Integer, _>(to)
+                            .execute(conn)
+                            .map_err(Error::from)
+                    }
```
tilacog, line 1085: We would instead use Diesel methods than plain SQL here, just like you did for the `Store::Shared` variant.

Please refer to my comment over there, as this scope could also benefit from removing the code duplication and using `anyhow::Context`.

store/postgres/src/chain_store.rs
```diff
@@ -1715,6 +1778,12 @@ impl ChainStoreTrait for ChainStore {
             .await
     }
 
+    async fn clear_call_cache(&self, from: Option<i32>, to: Option<i32>) -> Result<usize, Error> {
+        let storage = self.storage.clone();
+        let conn = self.get_conn()?;
+        storage.clear_call_cache(&conn, from, to)
```
tilacog, line 1784: We don't need to clone the store here:

```suggestion
        let conn = self.get_conn()?;
        self.storage.clear_call_cache(&conn, from, to)
```

store/postgres/src/chain_store.rs
tilacog: You can find the value for the call cache table at `call_cache.table`.
Here's an example of that table being used as a Diesel table type:
https://github.com/graphprotocol/graph-node/blob/fed567c028702888683ee801633e81f4c489e7dd/store/postgres/src/chain_store.rs#L1008-L1017

store/postgres/src/chain_store.rs
poonai: `filter` returns `Filter<Self, Predicate>`  instead of mutating `delete_stmt`. I'm worried that, filters won't be applied on `delete_stmt`

tilacog: Good catch! 
In this case, we can use the [`.into_boxed` method](https://docs.rs/diesel/2.0.2/diesel/query_builder/struct.DeleteStatement.html#method.into_boxed) and reassign the new value to the `delete_stmt` variable:
```rust
let mut delete_stmt = diesel::delete(cache::table).into_boxed();
if let Some(from) = from {
    delete_stmt = delete_stmt.filter(cache::block_number.ge(from));
}
if let Some(to) = from {
    delete_stmt = delete_stmt.filter(cache::block_number.le(to));
}
```

store/postgres/src/chain_store.rs
tilacog: It seems that we can't use our call cache Diesel tables for deletions 😓.
They are of type [`diesel_dynamic_schema::Table`](https://docs.rs/diesel-dynamic-schema/0.2.0/diesel_dynamic_schema/struct.Table.html) and don't implement the [`HasTable`](https://docs.rs/diesel/latest/diesel/associations/trait.HasTable.html) trait, which is required for deletion.

poonai: Thanks for verifying it for me. I’ll resolve the comments

docs/graphman.md
```diff
@@ -362,3 +363,56 @@ Inspect a block range, deleting any duplicated blocks:
...
+    graphman --config config.toml chain call-cache ethereum remove --from 10
+
+#### `to`
+
+    graphman --config config.toml chain call-cache ethereum remove --to 10
+
+#### `from` and `to`
+
+    graphman --config config.toml chain call-cache ethereum remove --from 10 --to 50
+
```
tilacog, line 408: The `OPTIONS` section should be dedicated only to explaining what each option does and how it affects the program execution. 

Instead of the examples above, we should describe how `--to` and `--from` affect block range selection. 
For instance, a reader might be interested in whether the `--from` block is included in the range or not (_"Are the range bounds inclusive?"_) and so on.

I believe those literal examples would best fit the `EXAMPLES` section, each preceded by a brief paragraph describing the expected outcome of that invocation.

store/postgres/src/chain_store.rs
```diff
@@ -1028,6 +1039,68 @@ mod data {
...
+                            .execute(conn)
+                            .map_err(Error::from)?;
+                        Ok(())
+                    }
+                    _ => {
+                        unreachable!("truncate should have been executed");
+                    }
+                },
+            }
+        }
```
tilacog, line 1102: Nice!
I really liked how you've factored out truncation. 

I'll only suggest adding some comments explaining a little bit about why we chose each path.

store/postgres/src/chain_store.rs
```diff
@@ -1028,6 +1039,68 @@ mod data {
...
+            &self,
+            conn: &PgConnection,
+            from: Option<i32>,
+            to: Option<i32>,
+        ) -> Result<(), Error> {
+            if from.is_none() && to.is_none() {
+                // fast path to clear the call cache table.
+                self.truncate_call_cache(conn)?;
+                return Ok(());
+            }
```
tilacog, line 1052:
```suggestion
            // If both `from` and `to` arguments are equal to `None`, then truncation should be
            // preferred over deletion as it is a faster operation.
            if from.is_none() && to.is_none() {
                self.truncate_call_cache(conn)?;
                return Ok(());
            }
```

store/postgres/src/chain_store.rs
```diff
@@ -1028,6 +1039,68 @@ mod data {
...
+                    if let Some(from) = from {
+                        delete_stmt = delete_stmt.filter(cache::block_number.ge(from));
+                    }
+                    if let Some(to) = to {
+                        delete_stmt = delete_stmt.filter(cache::block_number.le(to))
+                    }
+                    delete_stmt.execute(conn).map_err(Error::from)?;
+                    Ok(())
+                }
+                Storage::Private(Schema { call_cache, .. }) => match (from, to) {
```
tilacog, line 1066:
```suggestion
                // Because they are dynamically defined, our private call cache tables can't
                // implement all the required traits for deletion. This means we can't use Diesel
                // DSL with them and must rely on the `sql_query` function instead.
                Storage::Private(Schema { call_cache, .. }) => match (from, to) {
```

store/postgres/src/chain_store.rs
```diff
@@ -1028,6 +1039,68 @@ mod data {
...
+                        sql_query(query)
+                            .bind::<Integer, _>(from)
+                            .bind::<Integer, _>(to)
+                            .execute(conn)
+                            .map_err(Error::from)?;
+                        Ok(())
+                    }
+                    _ => {
+                        unreachable!("truncate should have been executed");
+                    }
```
tilacog, line 1099:
```suggestion
                    (None, None) => {
                        unreachable!("truncation was handled at the beginning of this function");
                    }
```

docs/graphman.md
```diff
@@ -7,6 +7,7 @@
 - [Unused Remove](#unused-remove)
 - [Drop](#drop)
 - [Chain Check Blocks](#check-blocks)
+- [Chain Call Cache](#chain-call-cache-remove)
```
tilacog, line 10: Just a detail: We should use the full command name here.
```suggestion
- [Chain Call Cache Remove](#chain-call-cache-remove)
```

docs/graphman.md
```diff
@@ -362,3 +363,56 @@ Inspect a block range, deleting any duplicated blocks:
 Inspect all blocks after block `13000000`:
 
     graphman --config config.toml chain check-blocks mainnet by-range --from 13000000
+
+<a id="chain-call-cache-remove"></a>
+# ⌘ Call Cache
```
tilacog, line 368: As per the comment before, we would also use the full command name here.
```suggestion
# ⌘ Chain Call Cache Remove
```

tilacog (changes requested):
Looking good!
I've left some suggestions, most of which can be applied directly.

tilacog (approved):
Thanks a lot for this!
I'll run it locally just to try it out before merging.

PR #4070, opened 2022-10-18
fix(docker): incorrect config PGDATA

0xdapper:
We are mounting our volume at `/var/lib/postgresql/data` while telling the container to use `/data/postgres` instead, fix the inconsistency so db persistency works as expected.


neysofu:
Thanks for the fix! I added an explanatory comment as to why we need to set `PGDATA`, I hope you don't mind :smiley:

neysofu (approved):

PR #4072, opened 2022-10-18
Make Docker builds faster by caching dependencies

neysofu:
Cargo is not very Docker-friendly by default and even a small change to any source file results in rebuilding the whole dependency tree. There's several known workarounds for this behavior, most of which boil down to "build the dependencies separately from the workspace, so that they get placed in a different caching layer".

[`cargo-chef`](https://github.com/LukeMathWalker/cargo-chef) is a simple tool that fits our use case: it creates a recipe of all dependencies from both `Cargo.toml` and `Cargo.lock`, builds those, and then lets you build the rest of your workspace with a normal `cargo build` that leverages Docker's cache. All builds that don't alter our dependency list will be faster.

This PR also removes `rustfmt` from the `graph-node-build` stage, which doesn't seem to be necessary.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,2 +1,6 @@
-.git
 target
+docker/data
+node_modules
+.dockerignore
+.git
+.gitignore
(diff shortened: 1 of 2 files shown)


mangas (approved):

lutter (approved):
LGTM!

PR #4073, opened 2022-10-18
tests: Improve `file_data_sources` raciness

leoyvens:
Worth the shot.


leoyvens:
I ran this at least 5 times, might actually be better now.

PR #4075, opened 2022-10-18
Feature/fix firehose dynamic data source

sduchesneau:
This PR leverages the feature of firehose-ethereum to fetch a single block (https://github.com/streamingfast/proto/blob/develop/sf/firehose/v2/firehose.proto#L13-L40) at current cursor in its full version.
It includes a bump in the firehose protocol from `sf.firehose.v1.Stream/Blocks` to `sf.firehose.v2.Stream/Blocks`

Firehose endpoints now have a minimal version requirement to because of the upgrade to sf.firehose.v2.Stream/Blocks:
- Requires firehose-ethereum > v1.2.0 (https://github.com/streamingfast/firehose-ethereum/releases/tag/v1.2.0)
- Requires firehose-cosmos > v0.6.0 (https://github.com/streamingfast/firehose-cosmos/releases/tag/v0.6.0)
- Requires firehose-arweave > v1.1.0 (https://github.com/streamingfast/firehose-arweave/releases/tag/v1.1.0)
- Requires firehose-near > `43ffb0f`  (https://github.com/streamingfast/firehose-near/commit/43ffb0f4781905f90e8a4d4bb139a9d8aed12bd7)

When a dynamic data source is added on a block, that block must be fetched back from firehose before we apply the new source to it, because the block that we have in hands is incomplete (transformed by the firehose filter to only contain interesting transactions).


leoyvens:
Does this reproduce with static filters on? In principle the static filter should retrieve all the triggers that could potentially apply to a template, regardless of the dynamic data sources that are actually instantiated.

sduchesneau:
> Does this reproduce with static filters on? 
> Edit: turns out we don't always want to have static filters on with firehose, and will do it adaptively as introduced in https://github.com/graphprotocol/graph-node/pull/4008

Exactly, this problem happens only _without_ static filters, but since the performance of static filters is very bad on some use cases, the adaptive approach is much more appropriate, so both ways must work without determinism issues

sduchesneau:
(note: current unit test failing happen on master)

leoyvens (approved):

leoyvens:
When we mix rpcs and firehoses in a config, it would be good to avoid calling the firehose endpoint for a refetch on a subgraph that's using an rpc blockstream. But we don't have to deal with that right now.

PR #4080, opened 2022-10-20
graphql: AND/OR filter

saihaj:
introduces alternative syntax for doing `AND/OR` filters.

Generated filter will have these new types
```graphql
input Proposal_fitler {
  id: ID
  # ... other variants of `id` like we do today
  collection: String
  # ... other variants of `collection` like we do today
  and: [Proposal_filter]
  or: [Proposal_filter]
}
```

### Example queries
1. using multiple AND filters
```graphql
query {
  proposals(
    where: { and: [{ id_in: "111" }, { collection_contains_nocase: "a" }] }
  ) {
    id
  }
}
```

2. Using multiple OR filters
```graphql
query {
  proposals(
    where: { or: [{ id_in: "111" }, { collection_contains_nocase: "a" }] }
  ) {
    id
  }
}
```
3. Using AND/OR combined
```graphql
query {
  proposals(
    where: {
      or: [{ id_in: "111" }, { collection_contains_nocase: "a" }]
      and: [{ owner_contains: "0x2" }]
    }
  ) {
    id
  }
}
```

Downside with this approach is users can still have nested `AND`/`OR` which can be quite expensive example
```graphql
query {
  proposals(
    where: {
      or: [
        { id_in: "111" }
        {
          collection_contains_nocase: "a"
          and: [{ owner_contains: "0x2" }, { id_in: "22" }]
        }
      ]
      and: [{ owner_contains: "0x2" }]
    }
  ) {
    id
  }
}
```

### DIFF
--- a/graphql/src/schema/api.rs
+++ b/graphql/src/schema/api.rs
@@ -188,23 +188,29 @@ fn add_filter_type(
             let mut generated_filter_fields = field_input_values(schema, fields)?;
             generated_filter_fields.push(block_changed_filter_argument());
 
-            generated_filter_fields.push(InputValue {
-                position: Pos::default(),
-                description: None,
-                name: "and".to_string(),
-                value_type: Type::ListType(Box::new(Type::NamedType(filter_type_name.to_owned()))),
-                default_value: None,
-                directives: vec![],
-            });
+            if !ENV_VARS.graphql.disable_bool_filters {
+                generated_filter_fields.push(InputValue {
+                    position: Pos::default(),
+                    description: None,
+                    name: "and".to_string(),
+                    value_type: Type::ListType(Box::new(Type::NamedType(
+                        filter_type_name.to_owned(),
+                    ))),
+                    default_value: None,
+                    directives: vec![],
+                });
 
-            generated_filter_fields.push(InputValue {
-                position: Pos::default(),
-                description: None,
-                name: "or".to_string(),
-                value_type: Type::ListType(Box::new(Type::NamedType(filter_type_name.to_owned()))),
-                default_value: None,
-                directives: vec![],
-            });
+                generated_filter_fields.push(InputValue {
+                    position: Pos::default(),
+                    description: None,
+                    name: "or".to_string(),
+                    value_type: Type::ListType(Box::new(Type::NamedType(
+                        filter_type_name.to_owned(),
+                    ))),
+                    default_value: None,
+                    directives: vec![],
+                });
+            }
 
             let typedef = TypeDefinition::InputObject(InputObjectType {
                 position: Pos::default(),
--- a/graphql/src/schema/api.rs
+++ b/graphql/src/schema/api.rs
@@ -188,6 +188,24 @@ fn add_filter_type(
             let mut generated_filter_fields = field_input_values(schema, fields)?;
             generated_filter_fields.push(block_changed_filter_argument());
 
+            generated_filter_fields.push(InputValue {
+                position: Pos::default(),
+                description: None,
+                name: "and".to_string(),
+                value_type: Type::ListType(Box::new(Type::NamedType(filter_type_name.to_owned()))),
+                default_value: None,
+                directives: vec![],
+            });
+
+            generated_filter_fields.push(InputValue {
+                position: Pos::default(),
+                description: None,
+                name: "or".to_string(),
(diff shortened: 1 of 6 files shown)


lutter:
I am still trying to wrap my head around what this all means - the notation that GraphQL imposes on us makes this all pretty confusing.

If I have predicates `a`, `b`, `c`, etc. where each predicate is something like `name: "Truss"`, `where: { a, b, c }` and `where: { and: [a, b, c] }` mean the same thing. It gets more murky with
```
where: {
  and: [a, b],
  or: [c, d]
}
```
because we need to define precedence rules. I would expect the above to mean `(a and b) or c or d`. That, I think, could also be written as `where: { or: [ {and: [a, b]}, c, d] }`.

I wonder if we should enforce the latter notation by making it illegal to have `and` and `or` at the same level, since the more explicit notation of  `where: { or: [ {and: [a, b]}, c, d] }` is much clearer.

There's also the mixed case of `where: { a, b, or: [c, d] }` which would mean `a and b and (c or d)` and also useless cases like `where: {a, b, and: [c, d] }` which is the same as `where: {a, b, c, d}`

As for this being expensive: if people start using `or` a lot, especially with no additional constraints, that can become very expensive; I haven't tried it but something like `where: { or: [{token_0: "ABC"}, {token_1: "ABC"}] }` OTOH, we already have plenty of ways to write expensive queries.

azf20:
Thanks so much @saihaj! I think this is close.. I wonder if we can address some of @lutter's concerns by making the approaches exclusive, i.e. the filter can only be one of the following:

Retained for backwards compatibility & simplicity:
```
input Proposal_original_filter {
  id: ID
  # ... other variants of `id` like we do today
  collection: String
}
```

Allows for "or conditions":
```
input Proposal_and_filter {
  or: [Proposal_filter]
}
```

Allows for more nuanced AND filtering, sub-filters with OR etc:
```
input Proposal_or_filter {
  and: [Proposal_filter]
}
```

So in summary:

```
input Proposal_filter = Proposal_original_filter | Proposal_and_filter | Proposal_or_filter
```

azf20:
@saihaj assuming yes, @lutter and I just caught up and this syntax looks good to go, but I think we will just want to add a check to prevent `and and or` filters at the same level, if that makes sense? Otherwise I think this is ready for review

azf20:
@saihaj as this is still in draft, just let us know when this is ready to take a look!

lutter:
One thing: it would be good if there was an (undocumented) feature flag `GRAPH_GRAPHQL_DISABLE_BOOL_FILTERS` to turn off boolean filters. I am not sure how hard it will be to get that in, but if we could control at least whether `or` can be used would be very good as that has the highest potential to create slow queries. Turning off `or` should happen at the point where we parse filters and just result in a `QueryExecutionError::NotSupported` or some such. The flag should default to `false` (i.e., we accept filters by default)

graphql/src/store/query.rs
```diff
@@ -161,91 +161,144 @@ fn parse_change_block_filter(value: &r::Value) -> Result<BlockNumber, QueryExecu
...
+                    Ok(block_number) => Ok(EntityFilter::ChangeBlockGte(block_number)),
+                    Err(e) => Err(e),
+                };
+            }
+            use self::sast::FilterOp::*;
+            let (field_name, op) = sast::parse_field_as_filter(key);
+
+            Ok(match op {
+                And => EntityFilter::And(build_list_filter_from_object(entity, object, schema)?),
+                Or => EntityFilter::Or(build_list_filter_from_object(entity, object, schema)?),
```
lutter, line 269: These two branches should check whether boolean filters are disabled and return a `QueryExecutionError` if they are.

graphql/src/schema/api.rs
```diff
@@ -188,6 +188,24 @@ fn add_filter_type(
...
+
+            generated_filter_fields.push(InputValue {
+                position: Pos::default(),
+                description: None,
+                name: "or".to_string(),
+                value_type: Type::ListType(Box::new(Type::NamedType(filter_type_name.to_owned()))),
+                default_value: None,
+                directives: vec![],
+            });
+
```
lutter, line 208: These two statements should only be executed if boolean filters are not disabled

lutter (changes requested):
This all looks good, you just need to add the environment variable that allows disabling boolean filters

lutter (approved):
Nice! Looks great now!

dotansimha (approved):

Issue #3863, opened 2022-08-22
Debian 10 has reached to EOL and node is using it.


shogochiai:
**Do you want to request a *feature* or report a *bug*?**
Bug report.

**What is the current behavior?**
`./docker/Dockerfile` has a reference to `rust:buster`.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem.**
PGroonga (Japanese fulltext search engine for Postgres) has been finished supporting Buster (Debian 10) due to its EOL (this August). Unsupported OS is the main OS of the node.

**What is the expected behavior?**
Maybe changing it to Trixie (Debian 13) and passing tests.

PR #4081, opened 2022-10-20
Use `bullseye-slim` as base image

neysofu:
Closes https://github.com/graphprotocol/graph-node/issues/3863.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -4,7 +4,7 @@
 # by running something like the following
 #   docker build --target STAGE -f docker/Dockerfile .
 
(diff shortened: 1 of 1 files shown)


matthewdarwin:
Any reason we can't use alpine?

neysofu:
I don't think the size savings are worth it. Dealing with musl libc is a pain sometimes, and its `malloc` performance is notoriously bad (https://github.com/graphprotocol/graph-node/issues/489). AFAIK `mallocng` in musl 1.2.1 improved things but `jemalloc` is still king.

mangas (approved):

Issue #3879, opened 2022-08-26
Firehose connections hang when more than 100 subgraphs are deployed


leoyvens:
See [this commit](https://github.com/graphprotocol/graph-node/commit/2c8173c5d7f7ea139bc5288a947f2782608a2e08) for a reproduction example. It exposes a few causes that seem to be conspiring:

1. The log below can be obtained by running the example with h2 tracing level at `DEBUG`. It shows that the firehose endpoint sets the http2 setting `max_concurrent_streams` to 100, thereby responding with `REFUSED_STREAM` to any additional streams. This limits the number of block streams to 100 on a single connection.
```
DEBUG Connection{peer=Client}: h2::codec::framed_read: received frame=Settings { flags: (0x0), max_concurrent_streams: 100, initial_window_size: 1048576, max_header_list_size: 65536 }
```

2. This setting is per connection, and we were [supposed to be using connection polling](https://github.com/graphprotocol/graph-node/pull/3833), but the example seems to use a single connection no matter what `conn_pool_size` is set to, so that connection polling implementation must not be working or not be balancing based on number of streams (least loaded, round-robin or randomly would work).

3. Tonic seems to not retry establishing the stream after receiving `REFUSED_STREAM`. On retry it logs the below and then goes silent. This might be related to this issue https://github.com/hyperium/tonic/issues/515.
```
2022-08-26T12:18:13.083821Z TRACE tonic::transport::service::reconnect: poll_ready; connected
2022-08-26T12:18:13.083864Z TRACE tonic::transport::service::reconnect: poll_ready; not ready
2022-08-26T12:18:13.083905Z TRACE hyper::client::dispatch: send_when canceled
```

On the potential fix:
If Firehose could reliably set `max_concurrent_streams` to a high value, that would be great. But afaik Firehose itself does not set a limit, so proxies with draconian defaults must be to blame ([nginx for example defaults to 128](https://nginx.org/en/docs/http/ngx_http_v2_module.html#http2_max_concurrent_streams)). Since we don't want to add configuration pitfalls to operators or require specific proxies, we cannot rely on this being set higher than 100 which is the [RFC recommended minimum](https://www.rfc-editor.org/rfc/rfc7540#section-6.5.2).

The tonic bug seems related but even if tonic retried, the stream would probably be refused again.

So the most reliable fix would be to get connection pooling working with an algorithm that seeks to balance the number of streams per connection.

leoyvens:
[...]

PR #4083, opened 2022-10-20
firehose connection loadbalance

mangas:
Fixes https://github.com/graphprotocol/graph-node/issues/3879

### DIFF
--- a/graph/src/firehose/endpoints.rs
+++ b/graph/src/firehose/endpoints.rs
@@ -4,14 +4,13 @@ use crate::{
     cheap_clone::CheapClone,
     components::store::BlockNumber,
     firehose::{decode_firehose_block, ForkStep},
-    prelude::{debug, info},
+    prelude::{anyhow, debug, info},
     substreams,
 };
 use futures03::StreamExt;
 use http::uri::{Scheme, Uri};
-use rand::prelude::IteratorRandom;
 use slog::Logger;
-use std::{collections::BTreeMap, fmt::Display, iter, sync::Arc, time::Duration};
+use std::{collections::BTreeMap, fmt::Display, sync::Arc, time::Duration};
 use tonic::{
     metadata::MetadataValue,
     transport::{Channel, ClientTlsConfig},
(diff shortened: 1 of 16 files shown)


graph/src/firehose/endpoints.rs
```diff
@@ -267,10 +264,12 @@ impl FirehoseEndpoints {
         self.0.len()
     }
 
+    // selects the FirehoseEndpoint with the lest amount of references, which will help with spliting
+    // the load naively across the entire list.
     pub fn random(&self) -> Option<&Arc<FirehoseEndpoint>> {
```
leoyvens, line 269: One aspect that concerns me is a silent hang if we ever hit the 100 stream limit per connection limit. Perhaps a simple way to make the error explicit would be to add a
```
const SUBGRAPHS_PER_CONN: usize = 100;
```
And return an error if the ref count reaches this number. Or even better auto-scale based on this.

leoyvens (changes requested):

graph/src/firehose/endpoints.rs
mangas: will address this on a separate PR as this could benefit from some metrics in order to detect a stall

graph/src/firehose/endpoints.rs
```diff
@@ -264,8 +266,19 @@ impl FirehoseEndpoints {
...
-    pub fn random(&self) -> Option<&Arc<FirehoseEndpoint>> {
-        self.0.iter().min_by_key(|x| Arc::strong_count(x))
+    pub fn random(&self) -> anyhow::Result<Arc<FirehoseEndpoint>> {
+        let endpoint = self
+            .0
+            .iter()
+            .min_by_key(|x| Arc::strong_count(x))
+            .ok_or(anyhow!("no available firehose endpoints"))?;
+        if Arc::strong_count(endpoint) > SUBGRAPHS_PER_CONN {
+            return Err(anyhow!("all connections saturated with {} connections, increase the firehose conn_pool_size", SUBGRAPHS_PER_CONN));
```
leoyvens, line 276: Nice work with the actionable error message!

leoyvens (approved):

leoyvens:
Oh and lets bump the default to 20.

leoyvens (approved):

PR #4089, opened 2022-10-21
feat(offchain): Ignore ds created with invalid cid

### LINKED ISSUE #4086: Creating File Data Sources with strings which aren't a valid multihash fails the subgraph
The subgraph fails with:
```
Failed to parse multihash
```

Expected behaviour: the file data source should not be created and indexing should continue

leoyvens:
Resolves #4086

### DIFF
--- a/core/src/subgraph/runner.rs
+++ b/core/src/subgraph/runner.rs
@@ -16,10 +16,9 @@ use graph::data::subgraph::{
     schema::{SubgraphError, SubgraphHealth, POI_OBJECT},
     SubgraphFeature,
 };
-use graph::data_source::{offchain, DataSource, TriggerData};
+use graph::data_source::{offchain, DataSource, DataSourceCreationError, TriggerData};
 use graph::prelude::*;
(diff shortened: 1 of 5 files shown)


core/src/subgraph/runner.rs
```diff
@@ -475,7 +474,14 @@ where
 
         for info in created_data_sources {
             // Try to instantiate a data source from the template
-            let data_source = DataSource::try_from(info)?;
+            let data_source = match DataSource::from_template_info(info) {
+                Ok(ds) => ds,
+                Err(e @ DataSourceCreationError::Ignore(..)) => {
```
mangas, line 479: This prolly more of a nitpick but I think I would prefer a Ok(None) over an error called Ignore as you'd not know there's an ignore from looking at the signature

leoyvens: I started with that but `None` does not carry the reason for ignoring, which is helpful for debugging.

mangas (approved):

PR #4092, opened 2022-10-22
Manage the statistics target for a deployment

lutter:
Show, set or reset the target for a deployment. The statistics target determines how much of a table Postgres will sample when it analyzes a table. This can be particularly beneficial when Postgres chooses suboptimal query plans for some queries. Increasing the target will make analyzing tables take longer and will require more space in Postgres' internal statistics storage.

See also https://www.postgresql.org/docs/15/planner-stats.html#id-1.5.13.5.3

### DIFF
--- a/node/src/manager/commands/stats.rs
+++ b/node/src/manager/commands/stats.rs
@@ -6,6 +6,7 @@ use crate::manager::deployment::DeploymentSearch;
 use diesel::r2d2::ConnectionManager;
 use diesel::r2d2::PooledConnection;
 use diesel::PgConnection;
+use graph::components::store::DeploymentLocator;
 use graph::components::store::VersionStats;
 use graph::prelude::anyhow;
 use graph_store_postgres::command_support::catalog as store_catalog;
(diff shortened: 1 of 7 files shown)


store/postgres/src/relational.rs
```diff
@@ -162,6 +162,18 @@ impl fmt::Display for SqlName {
     }
 }
 
+impl Borrow<str> for SqlName {
+    fn borrow(&self) -> &str {
+        self.as_str()
+    }
+}
+
```
tilacog, line 170: We don't need to implement `Borrow<str>` for `SqlName`, only for `&SqlName`.

```suggestion
```

node/src/manager/commands/stats.rs
```diff
@@ -165,3 +174,28 @@ pub fn target(
...
+    entity: Option<&str>,
+    columns: Vec<String>,
+    target: i32,
+    no_analyze: bool,
+) -> Result<(), anyhow::Error> {
+    let columns = if columns.is_empty() {
+        vec!["id".to_string(), "block_range".to_string()]
+    } else {
+        columns
+    };
```
tilacog, line 191: Two questions:
1. Should `id` and `block_range` always be included?
2. What about `vid`?

tilacog (approved):
LGTM. Just had a few questions, but everything looks fine.

store/postgres/src/relational.rs
lutter: Nice catch. Taking that out.

node/src/manager/commands/stats.rs
lutter: I haven't seen `vid` be useful here, that's why I left it out. And I felt that not adding to a list of user-provided columns is more understandable than always adding them.

PR #4096, opened 2022-10-24
Use Kaniko on CloudBuild

neysofu:
Not sure if the cloud builder image has BuildKit enabled but I *think* so. Let's try.

https://docs.docker.com/develop/develop-images/build_enhancements/#to-enable-buildkit-builds


neysofu:
Closing because it's not worth it to pursue this; see https://github.com/graphprotocol/graph-node/pull/4187.

Issue #4097, opened 2022-10-24
cargo build bug


mzaprudin:
**bug**

**What is the current behavior?**
I clone the gparh-node repository and run `cargo build`
but I get the next error
>$ cargo build
>error: failed to load manifest for workspace member `/home/m/nodes/graph-node/core`
>
>Caused by:
>  failed to parse manifest at `/home/m/nodes/graph-node/core/Cargo.toml`
>
>Caused by:
>  invalid type: map, expected a string for key `package.edition`

**What is the expected behavior?**
Building without any issue

neysofu:
We require `rustc` 1.64.0. Try running `rustup update`.

mzaprudin:
> We require `rustc` 1.64.0. Try running `rustup update`.

It was helpful
Thank you!

Issue #3230, opened 2022-02-08
Error when cargo build the graph node


bijianing97:
When i builded the graph node, it came error like this:
The code version is master a393098d3fc2036f9ade348f4dc08238feb41d6b 
The cargo version is cargo 1.57.0 (b2e52d7ca 2021-10-21)
The system is Ubuntu 20.04.3 LTS (GNU/Linux 5.4.0-86-generic x86_64)
```
 cargo build
   Compiling zstd-safe v3.0.1+zstd.1.4.9
   Compiling zstd-sys v1.4.20+zstd.1.4.9
   Compiling psm v0.1.14
   Compiling graph-store-postgres v0.25.0 (/mnt2/graph-node/store/postgres)
   Compiling zstd v0.6.1+zstd.1.4.9
error: there is no argument named `table_name`
   --> store/postgres/src/deployment_store.rs:690:40
    |
690 |             let sql = format!("analyze {table_name}");
    |                                        ^^^^^^^^^^^^

error: there is no argument named `table_name`
   --> store/postgres/src/deployment_store.rs:723:46
    |
723 |             let index_name = format!("manual_{table_name}_{column_names_sep_by_underscores}");
    |                                              ^^^^^^^^^^^^

error: there is no argument named `column_names_sep_by_underscores`
   --> store/postgres/src/deployment_store.rs:723:59
    |
723 |             let index_name = format!("manual_{table_name}_{column_names_sep_by_underscores}");
    |                                                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: there is no argument named `index_name`
   --> store/postgres/src/deployment_store.rs:726:58
    |
726 |                 "create index concurrently if not exists {index_name} \
    |                                                          ^^^^^^^^^^^^

error: there is no argument named `schema_name`
   --> store/postgres/src/deployment_store.rs:727:21
    |
727 |                  on {schema_name}.{table_name} using {index_method} \
    |                     ^^^^^^^^^^^^^

error: there is no argument named `table_name`
   --> store/postgres/src/deployment_store.rs:727:35
    |
727 |                  on {schema_name}.{table_name} using {index_method} \
    |                                   ^^^^^^^^^^^^

error: there is no argument named `index_method`
   --> store/postgres/src/deployment_store.rs:727:54
    |
727 |                  on {schema_name}.{table_name} using {index_method} \
    |                                                      ^^^^^^^^^^^^^^

[...]

PR #4098, opened 2022-10-24
Use a `rust-toolchain.toml` to simplify CI and minimize confusion

neysofu:
Would avoid bug reports like https://github.com/graphprotocol/graph-node/issues/4097 and https://github.com/graphprotocol/graph-node/issues/3230 in the future.


leoyvens (approved):

maoueh:
Oh yeah thanks :)

PR #4099, opened 2022-10-24
Make our CI faster

neysofu:
- Switch to `swatinem/rust-cache@v2` for caching Cargo dependencies, which automatically takes care of many subtleties for us and caches as much as possible without reusing artifacts which it shouldn't.
- Set `RUSTFLAGS` as a job-wide environment variable, thus ensuring it becomes part of the cache key.
- Cache `yarn` dependencies - not sure how effective this will be, maybe we'll remove it if it doesn't help.
- Run integration tests concurrently. This is a big one and should help quite a bit, @tilacog the integration testing framework still supports concurrent tests without issues, correct? Setting this to `N_CONCURRENT_TESTS` to 4 because it seems to get the best times out of the GitHub runners. Setting it too high (e.g. 15, the default) results in starvation and the job takes too long.
- Remove our release builds in the CI. They take a long time and rarely, if ever, pick up problems that `cargo test` doesn't.
- Be more selective when `cargo build`ing, ensuring that we're not building unnecessary stuff.

Steps now run noticeably (but not substantially) faster. Integration tests are still by far our bottleneck and more work needs to be done to further optimize them (or just pay up for larger runners).

Example: https://github.com/graphprotocol/graph-node/actions/runs/3328853325/jobs/5505338075 (failed because of a flaky unit test, but otherwise all previous runs looks good).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -114,11 +114,12 @@ jobs:
           args: --bin graph-node
 
       # Integration tests are a bit flaky, running them twice increases the
-      # chances of one run succeeding
+      # chances of one run succeeding.
       - name: Run integration tests (round 1)
         id: integration-tests-1
         uses: actions-rs/cargo@v1
         env:
+          N_CONCURRENT_TESTS: "4"
           TESTS_GANACHE_HARD_WAIT_SECONDS: "30"
         with:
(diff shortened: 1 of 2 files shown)


tilacog (approved):
:rocket:

PR #3410, opened 2022-03-30
WIP: Query performance testing

dotansimha:
- [x] Setup a new entry point for running `graph-node` and all requirements
- [x] Reuse parts from e2e/integration testing
- [ ] Setup a test project with a contract + Subgraph
- [ ] Run K6 as part of the CI workflow 
- [ ] Add GH Actions workflow for running + commenting results on each PR.

PR #4110, opened 2022-10-25
performance-tests: added workflow for running tests in CI (3️⃣)

dotansimha:
❗  Note: this PR is based on https://github.com/graphprotocol/graph-node/pull/4108 and https://github.com/graphprotocol/graph-node/pull/4109 and requires both to be merged first. 

### Background

As part of the effort done in https://github.com/graphprotocol/graph-node/pull/3410 , and based on a request from @dizzyd to break this PR into smaller pieces, here's part 3.

### Changes in this PR 

- [x] Added GitHub Actions workflow file to run and report performance issues on every PR.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -1,0 +1,67 @@
+name: Benchmark
+
+on:
+  - pull_request
+
+jobs:
+  query:
+    name: Query
+    strategy:
+      fail-fast: false
+      matrix:
+        rust: ["stable"]
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v2
+
+      - name: Cache cargo registry
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+          key: test-cargo-${{ hashFiles('**/Cargo.toml') }}
+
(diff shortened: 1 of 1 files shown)


michaelanegus:
Compiled feedback from Graph Node Team:

**4108**
-	Great addition. Graph Node team proposes discussing how to move to a docker-compose where the containers can run entrypoint.sh scripts that take care of the dependency between containers.
**4109** 
-	If we use this on the main repo, we would want to ensure that this will provide a good experience locally. As a consideration, there may be existing tools designed for this that could be used instead of adding infrastructure setup as code. Is this something you’ve considered?
**4110**
-	Can we run this on releases or release candidates instead of on every PR?  

**Other feedback/considerations:** 
-	Performance testing in isolation may not help much without generating serious load and without a well-prescribed setup in terms of hardware, etc. There is significant performance regression possible that will go unnoticed until the system is under serious load (e.g., performance issues from GraphQL validations were not visible in isolated testing but were nearly catastrophic under real load). 
-	Loading enough data into the database by syncing one or more subgraphs where we would see performance issues from the database would take a long time—this sort of test requires having a large dataset that can be quickly imported by restoring or cloning a database.
-	Having a setup where we can programmatically spin up a graph-node and run queries to test correctness end-to-end would be very useful. No strong opinion whether the automation in these PRs gets us there.

Issue #4088, opened 2022-10-21
File data source retries should back off exponentially


azf20:
When a file isn't found, Graph Node should retry, but the retry frequency should decrease exponentially.

Consideration: on subgraph restart, we don't want to overload the IPFS node with retries - retries should be load balanced somehow

leoyvens:
Notes from discussions had on backoff, first some situations which would be addressed:
1. A file is transiently bad.
2. A file is bad forever.
3. Whole endpoint down.

I don't think we should prioritize special treatment of 2, because it's hard to determine, or 3, because we currently rely on a single endpoint anyways. Hopefully a solution for 1 will be good enough for 2 and 3.

Some possible solutions:
a. Rate limit the whole endpoint. Currently we only have a limit for the number of concurrent requests, we should also have a rate limit.
b. Backoff when a file errors. We could apply a base timeout with fuzzing when a file is not found or errors. This helps if there is a small number of bad files, but if the time it takes to poll all the bad files is larger than the timeout then it's not effective.
c. Exponential backoff. This is more effective at scaling to a large number of bad files. The exponent then needs to be tracked per file.

PR #4117, opened 2022-10-28
Ipfs polling rate limiting and backoff from errors

leoyvens:
Resolves https://github.com/graphprotocol/graph-node/issues/4088.

The first commit is a fix for metrics.

The second refactors the ipfs service and adds a rate limit. The refactor seeks to better use tower composability and built-ins.

The third adds a backoff to errored requests in the polling monitor.

### DIFF
--- a/core/src/polling_monitor/mod.rs
+++ b/core/src/polling_monitor/mod.rs
@@ -1,13 +1,16 @@
 mod ipfs_service;
 mod metrics;
 
+use std::collections::HashMap;
 use std::fmt::Display;
+use std::hash::Hash;
 use std::sync::Arc;
 use std::task::Poll;
+use std::time::Duration;
 
 use futures::future::BoxFuture;
 use futures::stream::StreamExt;
-use futures::{stream, FutureExt, TryFutureExt};
+use futures::{stream, Future, FutureExt, TryFutureExt};
 use graph::cheap_clone::CheapClone;
 use graph::parking_lot::Mutex;
 use graph::prelude::tokio;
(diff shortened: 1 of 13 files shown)


core/src/polling_monitor/mod.rs
```diff
@@ -124,7 +171,16 @@ where
...
                                     "object_id" => id.to_string());
                         metrics.errors.inc();
-                        queue.push_back(id);
+
+                        // Requests that return errors could mean there is a permanent issue with
+                        // fetching the given item, or could signal the endpoint is overloaded.
+                        // Either way a backoff makes sense.
+                        let queue = queue.cheap_clone();
+                        let backoff = backoffs.next_backoff(id.clone());
+                        graph::spawn(async move {
```
mangas, line 180: this creates essentially an unbounded number of tasks. How does this affect the unconstrained task the runner is running on? I kinda would prefer a pending queue we could fetch stuff on a background task. This way it's easier to track and it's memory bound. I may be wrong here but the task seems too expensive for what is happening here.

leoyvens: If one of these waits forever then the `queue` would not be dropped even if the polling task ends, but that's not the case since there is a limit to the sleep time.

The tokio scheduler is [very efficient](https://tokio.rs/blog/2019-10-scheduler). And there is no fundamental reason why using tasks would be less efficient than rolling our own way of waiting on concurrent futures. There is `FuturesUnordered`, which in essence a single-task executor, but that's more suitable when there is a list of futures to be executed.

If we wanted to use a single task it might look like:
```rust
let (tx, rx) = mpsc::channel(1);
// Pull from the channel and simultaneously wait on up to a bounded number of futures.
// Uses `FuturesUnordered` under the hood.
task::spawn(rx.buffered_unordered(1_000_000).for_each(|_| ()));

// do `tx.push` this instead of `task::spawn`.
// Blocks if we've hit the `buffered_unordered` limit.
tx.push(async {
 backoff.await;
 queue.push_back(id);
}).await
```

This gives us the possibility of bounding the max number of timeouts being waited on, but I'm not sure that's particularly helpful in our case.

leoyvens: What ultimately protects this from growing without bound is the max data sources setting.

mangas: it's up to you, personally I don't think creating an unbounded number of tasks is great, especially because we can have 100s of SGs on the same runtime with each of them having an unpredictable amount of files. Putting that type of loads on the runtime, even if it's efficient, seems unnecessary but I don't know tokio internals very well so I'll leave it with you to decide.

That example above I guess would work but IIUIC if you have a high backoff time, all the following will wait that time so the last to be processed would be the sum of all the backoffs? I was thinking of place an item that looks like 
```
{
  id
  not_before: i64
}
```

Which would allow quickly scanning through the buffer and just queueing all the ready ones.

mangas:
I also think it would be good to add some test coverage for the rate limit etc even if minimal, with the mock it should be possible to just wait instead of responding which would simulate the rate limit. Same thing for errors I guess? I don't remember seeing that but if it was already there then please ignore this :pray:

mangas (approved):
Approving to unblock, left small comment about testing

core/src/polling_monitor/mod.rs
leoyvens: The `unordered` aspect would avoid one backoff blocking the other. I don't really want to complicate the queue by managing the backoff directly on it, that would probably result in removing items from the middle of the queue which isn't great for performance.

leoyvens:
Thanks for the review!

This didn't include specific testing for the rate limit and the backoff. The existing test for errors will test that the backed-off file is polled again, but it doesn't test the exponential aspect. Time based things, such as rate limits and backoffs, are tricky to test. But hopefully using the tower utilities leaves little room for error. Merging for expediency.

PR #4119, opened 2022-10-31
draft implementation of derived field mapping

poonai:
Cleaning this for review. 

Signed-off-by: poonai <rbalajis25@gmail.com>

### DIFF
--- a/store/postgres/src/writable.rs
+++ b/store/postgres/src/writable.rs
@@ -10,7 +10,7 @@ use graph::data::subgraph::schema;
 use graph::data_source::CausalityRegion;
 use graph::env::env_var;
 use graph::prelude::{
-    BlockNumber, DerviedEntityIds, Entity, MetricsRegistry, Schema, SubgraphDeploymentEntity,
+    BlockNumber, DerivedEntityIds, Entity, MetricsRegistry, Schema, SubgraphDeploymentEntity,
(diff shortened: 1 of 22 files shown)


tests/integration-tests/file-data-sources/schema.graphql
```diff
@@ -6,4 +6,15 @@ type IpfsFile @entity {
 type IpfsFile1 @entity {
   id: ID!
   content: String!
+}
+
+type Holder @entity {
```
poonai, line 11: I'll clean up the test case

store/postgres/src/writable.rs
```diff
@@ -716,6 +726,10 @@ impl Queue {
         }
     }
 
+    fn get_derived_ids(&self, key: &DerivedKey) -> Result<Option<DerivedEntityIds>, StoreError> {
+        self.store.get_dervied_ids(&key, BLOCK_NUMBER_MAX)
+    }
```
leoyvens, line 731: The queue is a pipeline of entity operations going into the store. For example it may be that when processing block `N`, block `N - 1` has not yet been transacted into the store, but is queued. If you access `self.store`, the queue is being bypassed so writes in block `N - 1` would be ignored.

So this needs to take the modifications in the `queue` into account. It's a similar problem to the entity cache, it contains pending writes. This is tricky to handle but looking at the approach in `fn get_many` should be a good start.

PR #4126, opened 2022-11-03
Leo/retry ds create

mangas:


core/src/subgraph/context/instance.rs
```diff
@@ -155,7 +155,35 @@ where
         })
     }
 
-    pub(super) fn revert_data_sources(&mut self, reverted_block: BlockNumber) {
+    pub(super) fn revert_data_sources(
+        &mut self,
+        reverted_block: BlockNumber,
+    ) -> Vec<DataSource<C>> {
```
leoyvens, line 161: Considering we don't need to return removed on-chain data sources, this could return `Vec<offchain::DataSource>`. We could even return just `Vec<offchain::Source>` to minimize cloning.

core/src/subgraph/context/instance.rs
```diff
@@ -155,7 +155,35 @@ where
...
+        // should be fine. *IT'S FINE*
+        if self.causality_region_seq.0 == CausalityRegion::ONCHAIN {
+            return removed;
+        }
+
+        // TODO: replace this with drain_filter once it's stabilised.
+        let (drained, hosts): (Vec<_>, Vec<_>) =
+            self.hosts.iter().cloned().partition(
+                |host| matches!(host.done_at(), Some(done_at) if done_at >= reverted_block),
+            );
```
leoyvens, line 176: I don't think we want to remove any hosts, but just set `done_at` to `None` on the necessary data sources.

tests/src/fixture.rs
```diff
@@ -336,6 +383,7 @@ pub async fn wait_for_sync(
     store: &SubgraphStore,
     hash: &DeploymentHash,
     stop_block: BlockPtr,
+    expect_healthy: bool,
```
leoyvens, line 386: This `expect_healthy` param I added turned out not to be useful in this PR, since we don't actually fail the subgraph. And in a separate PR I've exposed errors here in a [different way](https://github.com/graphprotocol/graph-node/pull/4147/files#diff-d470d1203f4dd0118c3c5ea05685e7be336d1174aad8b4f8b49567cae94552eaR356-R361). So lets remove it.

core/src/subgraph/context/instance.rs
```diff
@@ -155,7 +156,36 @@ where
...
+    pub(super) fn revert_data_sources(
+        &mut self,
+        reverted_block: BlockNumber,
+    ) -> Vec<offchain::Source> {
+        self.revert_quick(reverted_block);
+
+        // The following code handles resetting offchain datasources so in most
+        // cases this is enough processing.
+        // At some point we prolly need to improve the linear search but for now this
+        // should be fine. *IT'S FINE*
```
leoyvens, line 172: The `*IT'S FINE*` is very reassuring but perhaps a bit too reassuring 😄

graph/src/data_source/offchain.rs
```diff
@@ -32,35 +33,34 @@ pub struct DataSource {
...
-        }
-    }
-}
-
 impl DataSource {
     // mark this data source as processed.
     pub fn mark_processed_at(&self, block_no: i32) {
-        *self.done_at.lock().unwrap() = Some(block_no);
+        self.done_at
+            .store(block_no, std::sync::atomic::Ordering::SeqCst);
```
leoyvens, line 44: To be extra paranoid, lets `assert!(block_no != NOT_DONE_VALUE)`.

core/src/subgraph/context/instance.rs
```diff
@@ -164,11 +194,11 @@ where
             .filter(|h| h.creation_block_number() >= Some(reverted_block))
             .is_some()
         {
-            self.hosts.pop();
+            self.hosts.pop().unwrap().data_source();
```
leoyvens, line 197: No need to call `data_source()` here, it's a dead value.

graph/src/data_source/tests.rs
```diff
@@ -31,7 +31,7 @@ fn offchain_duplicate() {
         },
         context: Arc::new(None),
         creation_block: Some(0),
-        done_at: Mutex::new(None),
+        done_at: Arc::new(AtomicI32::new(0)),
```
leoyvens, line 34:
```suggestion
        done_at: Arc::new(AtomicI32::new(-1)),
```

graph/src/data_source/offchain.rs
```diff
@@ -32,35 +33,34 @@ pub struct DataSource {
...
+    }
+
+    pub fn done_at(&self) -> Option<i32> {
+        match self.done_at.load(std::sync::atomic::Ordering::SeqCst) {
+            NOT_DONE_VALUE => None,
+            n => Some(n),
+        }
+    }
+
+    pub fn set_at(&self, block: Option<i32>) {
```
leoyvens, line 59:
```suggestion
    pub fn set_done_at(&self, block: Option<i32>) {
```

core/src/subgraph/context/instance.rs
mangas: image

graph/src/data_source/offchain.rs
mangas: I'm very confused as to how the done got lost :thinking:

core/src/subgraph/context/instance.rs
```diff
@@ -155,7 +156,36 @@ where
...
+        &mut self,
+        reverted_block: BlockNumber,
+    ) -> Vec<offchain::Source> {
+        self.revert_quick(reverted_block);
+
+        // The following code handles resetting offchain datasources so in most
+        // cases this is enough processing.
+        // At some point we prolly need to improve the linear search but for now this
+        // should be fine. *IT'S FINE*
+        if self.causality_region_seq.0 == CausalityRegion::ONCHAIN {
```
leoyvens, line 173: The comment could better explain why this check guarantees that there are no offchain data sources. Perhaps it would help readability to add a method to the sequence such as `.is_minimum()`. Then we can keep the inner value in `CausalityRegionSeq` private.

core/src/subgraph/context/instance.rs
```diff
@@ -155,7 +156,36 @@ where
...
+            .filter(|host| matches!(host.done_at(), Some(done_at) if done_at >= reverted_block))
+            .map(|host| {
+                host.set_done_at(None);
+                // Safe to call unwrap() because only offchain DataSources have done_at = Some
+                host.data_source().as_offchain().unwrap().source.clone()
+            })
+            .collect()
+    }
+
+    fn revert_quick(&mut self, reverted_block: BlockNumber) {
```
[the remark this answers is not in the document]
mangas: the idea was to speak to the complexity of it. It's an internal name so we don't needanything fancy the point was the distinguish the quick part for the slow part. maybe revert_hosts_cheap or something?

core/src/subgraph/context/instance.rs
mangas: I agree with the comment, disagree with making it private. having minimum or whatever is opaque and makes the code impossible to understand without a lot of context. Will try to improve the comment but I'd rather leave ONCHAIN exposed so we have this clarity. Otherwise we could do is_onchain or has_offchain type thing.

core/src/subgraph/context/instance.rs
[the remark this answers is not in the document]
leoyvens: That's good, thanks

leoyvens (approved):
Great work, this is fixing at least three different bugs at once!

PR #4134, opened 2022-11-03
Remove `subgraph_deployment.earliest_ethereum_block_*`

lutter:
These columns weren't really used since PR #3673 but left in place in case we needed to go back to them. This PR completely removes them.


leoyvens (approved):
Nice cleanup!

PR #4144, opened 2022-11-07
Better error message if missing PostgreSQL password

### LINKED ISSUE #4002: could not find user in database while running `cargo run`
bug

Getting error: `thread 'main' panicked at 'connection url's contain enough detail: could not find user in 'postgresql://postgres@localhost:5432/graph-node''`
```bash
Sep 28 06:40:04.162 INFO Creating transport, capabilities: archive, traces, url: https://mainnet.infura.io/v3/..., provider: mainnet-rpc-0
Sep 28 06:40:04.163 INFO Successfully connected to IPFS node at: http://127.0.0.1:5001/
Sep 28 06:40:05.199 DEBG Creating firehose networks [1 chains, ingestor default]
Sep 28 06:40:05.199 DEBG Creating firehose networks [1 chains, ingestor default]
Sep 28 06:40:05.199 WARN Expensive queries file not set to a valid file: /etc/graph-node/expensive-queries.txt
Sep 28 06:40:05.199 DEBG Cleaning up large notifications after about 300s, channel: store_events, component: NotificationListener
thread 'main' panicked at 'connection url's contain enough detail: could not find user in `postgresql://postgres@localhost:5432/graph-node`', node/src/store_builder.rs:103:14
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
while running:
```bash
  --postgres-url postgresql://postgres@localhost:5432/graph-node \
  --ethereum-rpc mainnet:https://mainnet.infura.io/v3/...... \
  --ipfs 127.0.0.1:5001 \
  --debug
```

It says it couldn't find user in the database but I do see the `postgres` user when I do:
- `psql -U postgres`
- then running `\du` gives:
```
                                     List of roles
 Role name |                         Attributes                         | Member of 
-----------+------------------------------------------------------------+-----------
 postgres  | Superuser, Create role, Create DB, Replication, Bypass RLS | {}
```

The error seems to be coming from [`node/src/store_builder.rs:103:14`](https://github.com/graphprotocol/graph-node/blob/master/node/src/store_builder.rs#L103) according to the error message. Since I don't know rust, I couldn't get any info from that file.

I have followed all the steps as mentioned in the [getting-started.md](https://github.com/graphprotocol/graph-node/blob/master/docs/getting-started.md) file, which were:
- `ipfs init`
- `ipfs daemon`
- `initdb -D .postgres -U postgres`
- `pg_ctl -D .postgres -l logfile start`
- `createdb graph-node -U postgres`

`logfile`:
```bash
2022-09-28 07:03:52.858 UTC [5367] LOG:  starting PostgreSQL 14.5 (Ubuntu 14.5-1.pgdg20.04+1) on x86_64-pc-linux-gnu, compiled by gcc (Ubuntu 9.4.0-1ubuntu1~20.04.1) 9.4.0, 64-bit
2022-09-28 07:03:52.858 UTC [5367] LOG:  listening on IPv6 address "::1", port 5432
2022-09-28 07:03:52.858 UTC [5367] LOG:  listening on IPv4 address "127.0.0.1", port 5432
2022-09-28 07:03:52.885 UTC [5367] LOG:  listening on Unix socket "/var/run/postgresql/.s.PGSQL.5432"
2022-09-28 07:03:52.916 UTC [5371] LOG:  database system was shut down at 2022-09-28 07:03:49 UTC
2022-09-28 07:03:52.933 UTC [5367] LOG:  database system is ready to accept connections
```

The graph node should run without error.

Any help on this would be very appreciated.

neysofu:
Fixes #4002.

### DIFF
--- a/store/postgres/src/connection_pool.rs
+++ b/store/postgres/src/connection_pool.rs
@@ -86,7 +86,12 @@ impl ForeignServer {
         let password = String::from_utf8(
             config
                 .get_password()
-                .ok_or_else(|| anyhow!("could not find user in `{}`", SafeDisplay(postgres_url)))?
+                .ok_or_else(|| {
+                    anyhow!(
+                        "could not find password in `{}`; you must provide one.",
+                        SafeDisplay(postgres_url)
+                    )
+                })?
                 .into(),
         )?;
         let port = config.get_ports().get(0).cloned().unwrap_or(5432u16);


store/postgres/src/connection_pool.rs
```diff
@@ -86,7 +86,12 @@ impl ForeignServer {
         let password = String::from_utf8(
             config
                 .get_password()
-                .ok_or_else(|| anyhow!("could not find user in `{}`", SafeDisplay(postgres_url)))?
+                .ok_or_else(|| {
+                    anyhow!(
+                        "could not find password in `{}`; you must provide one.",
```
lutter, line 91: We could make this even more helpful if the error became `Could not find password in `{}`. Make sure to use a valid connection URI following https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING`

lutter (approved):
Nice improvement!

store/postgres/src/connection_pool.rs
neysofu: I *think* the doc you linked says password is optional in URI connection strings; it's `graph-node` that is not fully spec-compliant. Am I reading this right?

```
postgresql://[userspec@][hostspec][/dbname][?paramspec]

where userspec is:

user[:password]

...
```

lutter: Ooh .. yes, you are absolutely right!

Issue #3770, opened 2022-07-25
File data sources: Causality region isolation


leoyvens:
See #3072

leoyvens:
__Problem statement__

For indexing results to be reproducible given a set of available files, regardless of the order in which the files are found, it is necessary to isolate the effects of processing offchain data sources. An onchain data source cannot read or update entities created by an offchain data source, and an offchain data source cannot read or update an entity created by another offchain data source.

Causality regions help us, because by assigning all onchain data sources the same causality region and each offchain data source their own causality region, we can restate the goal as: A data source cannot read or update entities from other causality regions.

The isolation rule can be further refined as read restriction and a write restriction.
- Read restriction: An attempted read from a different causality region returns null.
- Write restriction:  An an attempted update or overwrite to an entity of a different causality region is a _conflict_ which must raise an error.

This issue is about implementing both restrictions.

__Implementation plan__

1. Enforce the `entities` manifest field is enforced on file data sources (PR #4147)
   This way we know which entity types actually need causality region tracking, which reduces the performance costs of checking causality regions, on the principle of "you don't pay for what you don't use".
2. Add a causality region column to entity tables which are used by file data sources (WIP)
3. Write the correct causality region to entities written to those tables.
4. Implement the read restriction.
5. Implement the write restriction, the simplest way to do this is to always create [the exclusion constraint](https://github.com/graphprotocol/graph-node/blob/a3cc5b99a8d98a4d9d8d61a2331f407162be8115/store/postgres/src/catalog.rs#L140-L142) for tables which have a causality region column.
5. End-to-end tests for all the permutations of reading and updating from onchain/offchain and offchain/offchain.

The details on 3 and 4 are not clear yet, what would the best way be to get the causality region value to the store? Maybe add it to the `EntityKey`?

leoyvens:
Done.

PR #4147, opened 2022-11-07
fix(offchain): Enforce entities list on offchain data sources

leoyvens:
First part of #3770.

This enforces that file data sources can only call `entity.get`, `entity.set` and `entity.remove` on entity types they've declared in the `entities:` array in the manifest.

### DIFF
--- a/graph/src/data_source/mod.rs
+++ b/graph/src/data_source/mod.rs
@@ -43,6 +43,15 @@ pub enum DataSourceCreationError {
 }
 
 /// Which entity types a data source can read and write to.
+///
+/// Currently this is only enforced on offchain data sources and templates, based on the `entities`
+/// key in the manifest. This informs which entity tables need an explicit `causality_region` column
+/// and which will always have `causality_region == 0`.
+///
+/// Note that this is just an optimization and not sufficient for causality region isolation, since
+/// generally the causality region is a property of the entity, not of the entity type.
+///
+/// See also: entity-type-access
 pub enum EntityTypeAccess {
     Any,
     Restriced(Vec<EntityType>),
--- a/graph/src/data_source/mod.rs
+++ b/graph/src/data_source/mod.rs
@@ -13,7 +13,7 @@ use crate::{
     },
     components::{
         link_resolver::LinkResolver,
-        store::{BlockNumber, StoredDynamicDataSource},
+        store::{BlockNumber, EntityType, StoredDynamicDataSource},
     },
     data_source::offchain::OFFCHAIN_KINDS,
     prelude::{CheapClone as _, DataSourceContext},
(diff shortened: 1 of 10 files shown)


graph/src/data_source/mod.rs
```diff
@@ -42,6 +42,33 @@ pub enum DataSourceCreationError {
     Unknown(#[from] Error),
 }
 
+/// Which entity types a data source can read and write to.
+pub enum EntityTypeAccess {
```
mangas, line 46: Might be worth adding a bit more here about how this interacts with causality region based isolation

runtime/wasm/src/host_exports.rs
```diff
@@ -97,6 +99,20 @@ impl<C: Blockchain> HostExports<C> {
         }
     }
 
+    fn check_entity_type_access(&self, entity_type: &EntityType) -> Result<(), HostExportError> {
+        match self.entity_type_access.allows(entity_type) {
+            true => Ok(()),
+            false => Err(HostExportError::Deterministic(anyhow!(
```
mangas, line 105: Could this error be added as a typed error? Seems quite specific to be matched on string so I think it could be useful to have an IsolationError subtype

runtime/wasm/src/host_exports.rs
```diff
@@ -97,6 +99,20 @@ impl<C: Blockchain> HostExports<C> {
         }
     }
 
+    fn check_entity_type_access(&self, entity_type: &EntityType) -> Result<(), HostExportError> {
```
mangas, line 102: even though this is not exported I think we should a bit of context here as a comment

tests/tests/runner.rs
```diff
@@ -219,6 +222,19 @@ async fn file_data_sources() {
         assert!(data_source.causality_region == causality_region.next());
         causality_region = causality_region.next();
     }
+
+    let stop_block = test_ptr(5);
+    let err = ctx.start_and_sync_to_error(stop_block).await;
+    let message = "entity type `IpfsFile1` is not on the 'entities' list for data source `File2`. \
```
mangas, line 228: would it be enough to check if this is an isolation error somehow (left a suggestion about subtype but that might be lost). Maybe we could match an important part of the match like IpfsFile1 or somnething rather than the full message which also includes a backtrace. As it is it will be broken as soon as any change is made in this subgraph.

runtime/wasm/src/host_exports.rs
leoyvens: It could, but given that for now we don't need to match on it I think it's fine to just use the `Deterministic` variant.

graph/src/data_source/mod.rs
leoyvens: I've written a more detailed comment.

runtime/wasm/src/host_exports.rs
leoyvens: I've written a short comment and linked it to the one on `enum EntityTypeAccess`.

tests/tests/runner.rs
leoyvens: If the test breaks when functionality changes, that's good! Gives us a chance to check that the change was intended. Even if its for the backtrace which is not directly unrelated to what's being tested.

leoyvens:
@mangas thanks for the review! I've addressed the comments.

mangas (approved):

PR #4152, opened 2022-11-09
store: Improve predictability of the time for each batch during pruning

lutter:


mangas (approved):

lutter:
> This makes the `vid >= $3` and `limit $4` on the outer query redundant right? So we could maybe remove them or explain in the comment that they are redundant.

Yes, they are redundant; I was in a bit of a belt and suspenders mood with that. I'll expand the comment

PR #4162, opened 2022-11-11
Add causality region column

leoyvens:
For now this just tracks the tables that need the column and adds the column to the DDL, but still unconditionally inserts 0. Inserting the correct causality region is follow up work.

Depends on #4147.


lutter:
One comment: since the set of entity types that need to store causality regions is static, this new column should go into `subgraph_manifest`. I'd like to change things so that only data that changes frequently (like the subgraph head pointer) are stored in `subgraph_deployment` and that slow moving data goes into `subgraph_manifest`.

And to bikeshed: `has_causality_region` makes me think this is a boolean column. Maybe `entities_with_causality_region` would be better name (though I'll live with either)

leoyvens:
@mangas thanks taking a stab at this, I've moved your WIP commit to a [separate branch](https://github.com/graphprotocol/graph-node/commit/857dea5c282c73f5015b9b4c2b3071f82851366c). I'll take some things from it but I suspect some of the changes, such as adding a CR to the mapping trigger, are not necessary.

@lutter thanks for the review, I'll make those changes.

core/src/subgraph/registrar.rs
```diff
@@ -620,11 +620,30 @@ async fn create_subgraph_version<C: Blockchain, S: SubgraphStore>(
         "block" => format!("{:?}", base_block.as_ref().map(|(_,ptr)| ptr.number))
     );
 
+    // Entity types that may be touched by offchain data sources need a causality region column.
+    let needs_causality_region = manifest
```
mangas, line 624: I think this could benefit from some test coverage, just making sure that given some manifests, the DeploymentCreate ends up with the right entities

graph/src/components/store/mod.rs
```diff
@@ -104,13 +128,18 @@ pub struct EntityKey {
 
     /// ID of the individual entity.
     pub entity_id: Word,
+
+    pub causality_region: CausalityRegion,
```
mangas, line 132: worth adding a small note on why this needs to be here? I think it's well explained on the main CR comment but might help anyone looking at this code. Possibly along the lines of "the CR this entity belong too, only DS with the same CR should be able to interact with it"

graph/src/data/subgraph/schema.rs
```diff
@@ -136,6 +137,15 @@ impl DeploymentCreate {
         self.debug_fork = fork;
         self
     }
+
+    pub fn entities_with_causality_region(
```
mangas, line 141: offchain_entities?

graph/src/data/subgraph/schema.rs
```diff
@@ -169,17 +179,23 @@ pub struct SubgraphManifestEntity {
...
     pub raw_yaml: Option<String>,
+    pub entities_with_causality_region: Vec<EntityType>,
 }
 
 impl SubgraphManifestEntity {
-    pub fn new(raw_yaml: String, manifest: &super::SubgraphManifest<impl Blockchain>) -> Self {
+    pub fn new(
+        raw_yaml: String,
+        manifest: &super::SubgraphManifest<impl Blockchain>,
+        entities_with_causality_region: Vec<EntityType>,
```
mangas, line 189:
```suggestion
        offchan_entities: Vec<EntityType>,
```

mangas: with or without CR I think should be a detail of the store

graph/src/data_source/causality_region.rs
```diff
@@ -47,6 +49,15 @@ impl CausalityRegion {
     pub const fn next(self) -> Self {
         CausalityRegion(self.0 + 1)
     }
+
+    pub fn from_entity(entity: &Entity) -> Self {
+        CausalityRegion(
+            entity
+                .get("causality_region")
+                .and_then(|v| v.as_int())
+                .unwrap_or(0),
```
mangas, line 58:
```suggestion
                .unwrap_or(CausalityRegion::ONCHAIN),
```

mangas: covering this with a basic unit test would also be good. If we don't replace the constant it could happen that someone changes the value and then this would be broken.

graph/src/data_source/mod.rs
```diff
@@ -206,6 +206,13 @@ impl<C: Blockchain> DataSource<C> {
             Self::Offchain(_) => vec![],
         }
     }
+
+    pub fn causality_region(&self) -> CausalityRegion {
```
mangas, line 210: If not covered already then adding a simple test would be good

graph/src/data_source/mod.rs
```diff
@@ -250,6 +257,13 @@ impl<C: Blockchain> DataSourceTemplate<C> {
         }
     }
 
+    pub fn as_offchain(&self) -> Option<&offchain::DataSourceTemplate> {
```
mangas, line 260: same

store/postgres/src/deployment_store.rs
```diff
@@ -1043,17 +1050,17 @@ impl DeploymentStore {
     ) -> Result<Option<Entity>, StoreError> {
         let conn = self.get_conn()?;
         let layout = self.layout(&conn, site)?;
-        layout.find(&conn, &key.entity_type, &key.entity_id, block)
+        layout.find(&conn, &key, block)
     }
 
     /// Retrieve all the entities matching `ids_for_type` from the
```
mangas, line 1056: Seems to have some more details we could add here now

store/postgres/src/relational.rs
```diff
@@ -500,34 +505,39 @@ impl Layout {
     pub fn find_many(
```
mangas, line 505: maybe comment on what is going on in this function after the changes could be helpful

store/postgres/src/relational.rs
```diff
@@ -1213,6 +1221,10 @@ pub struct Table {
     /// Entities in this table are immutable, i.e., will never be updated or
     /// deleted
     pub(crate) immutable: bool,
+
+    /// Whether this table has an explicit `causality_region` column. If `false`, then the column is
+    /// not present and the causality region for all rows is implicitly `0`.
```
mangas, line 1226:
```suggestion
    /// not present and the causality region for all rows is implicitly `CausalityRegion::ONCHAIN` .
```

store/postgres/src/relational_queries.rs
```diff
@@ -3511,6 +3556,24 @@ impl<'a> QueryFragment<Pg> for CopyEntityBatchQuery<'a> {
             }
             (false, false) => out.push_sql(BLOCK_RANGE_COLUMN),
         }
+
+        match (self.src.has_causality_region, self.dst.has_causality_region) {
```
mangas, line 3560: Not sure if I missed it but I don't think these are now covered in the database tests other than the integration test that uses the ipfs SG right? Can you please ensure this is covered on the store tests? Apologies if it is and I missed it

store/postgres/tests/relational_bytes.rs
```diff
@@ -253,24 +275,29 @@ fn find_many() {
...
+
+        let id_key = EntityKey {
+            entity_id: ID.into(),
+            entity_type: THING.clone(),
+            causality_region: CausalityRegion::ONCHAIN,
+        };
+        let id2_key = EntityKey {
+            entity_id: ID2.into(),
+            entity_type: THING.clone(),
+            causality_region: CausalityRegion::ONCHAIN,
```
mangas, line 297: I guess these are the ones I missed earlier. maybe add one that's not ONCHAIN?

mangas (approved):

mangas:
approving as in general looks good to me. Would be good to fix the comments you think are worth fixing.

graph/src/data/subgraph/schema.rs
leoyvens: An entity table might have a causality region column and stores some entities from onchain data sources. So calling this `offchain_entities` might be more confusing.

graph/src/components/store/mod.rs
leoyvens: Added a comment.

graph/src/data_source/causality_region.rs
leoyvens: I added this to the `relational_bytes::find` test.

graph/src/data_source/mod.rs
leoyvens: I've added a `data_source_helpers` test.

store/postgres/src/deployment_store.rs
leoyvens: Improved the comment a bit.

store/postgres/src/relational.rs
leoyvens: Clarified.

store/postgres/tests/relational_bytes.rs
leoyvens: It would be a bit annoying to test this here since it would be necessary to add to the test layout a new entity type that somehow has a causality region column. I added coverage for looking up offchain entities to the integration tests, and added a `debug_assert!` to the entity cache that ensures cache consistency (so the test is not looking up only on the cache).

store/postgres/src/relational_queries.rs
leoyvens: Yes, all these graft combinations should be tested. But at this point this has been open for too long and I'm thinking of merging it without these test, with the risk that there may be a bug in grafting a subgraph with file data sources.

leoyvens:
This now fully implements the isolation rules. The review comments have been addressed and this is ready for a final review.

core/src/subgraph/provider.rs
```diff
@@ -82,9 +82,7 @@ impl<I: SubgraphInstanceManager> SubgraphAssignmentProviderTrait for SubgraphAss
         {
             // Shut down subgraph processing
             self.instance_manager.stop_subgraph(deployment).await;
-            Ok(())
-        } else {
-            Err(SubgraphAssignmentProviderError::NotRunning(deployment))
```
mangas, line 87: Is this change related to this PR?

store/postgres/src/relational.rs
```diff
@@ -554,18 +565,15 @@ impl Layout {
 
         for entity_data in inserts_or_updates.into_iter() {
             let entity_type = entity_data.entity_type();
-            let mut data: Entity = entity_data.deserialize_with_layout(self, None, false)?;
+            let data: Entity = entity_data.deserialize_with_layout(self, None, true)?;
             let entity_id = Word::from(data.id().expect("Invalid ID for entity."));
             processed_entities.insert((entity_type.clone(), entity_id.clone()));
 
-            // `__typename` is not a real field.
-            data.remove("__typename")
```
mangas, line 562: is this ok to remove now? I seem to remember this being added (by Eva?) because this was causing some issues?

store/postgres/tests/relational_bytes.rs
mangas: I think this is the price we have to pay for the design decision and current test strategy. IMO it's worth adding this coverage even if it takes longer to write. I would, however, add it in a separate PR.

store/postgres/src/relational.rs
leoyvens: This was just a refactor, a couple lines above the last parameter to `deserialize_with_layout` was changed to `true` so that the typename is already removed.

PR #4163, opened 2022-11-11
store/postgres: insist on a database with C locale

lutter:
This PR is a variation of PR #3671 It does two things:

* check that collation and ctype are set to `C` and that encoding is `UTF8`
* panic if this is an empty `graph-node` installation, otherwise log a warning


tilacog (approved):
LGTM!

PR #4170, opened 2022-11-14
`cargo check --release` in the CI

neysofu:
They were removed in https://github.com/graphprotocol/graph-node/pull/4099 but we want to reintroduce them. See https://github.com/graphprotocol/graph-node/pull/4166.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -167,3 +167,26 @@ jobs:
         with:
           command: clippy
           args: --no-deps
+
+  release-check:
+    name: Build in release mode
+    runs-on: ubuntu-latest
+    env:
+      RUSTFLAGS: "-D warnings"
+    steps:
+      - uses: actions/checkout@v2
+      - uses: Swatinem/rust-cache@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -167,3 +167,26 @@ jobs:
...
+      - uses: Swatinem/rust-cache@v2
+      - name: Install dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get -y install libpq-dev
+      - name: Cargo check (debug)
+        uses: actions-rs/cargo@v1
+        with:
+          command: check
+          args: --tests
```
leoyvens, line 187: Could we remove this step of running debug check with `--tests`, maybe making it a bit faster?

neysofu: I think we can, running the tests covers it already. 👍

neysofu: Fixed in 344e95da865554c17a8c66363f9eb8bf71458429.

leoyvens (approved):

PR #4171, opened 2022-11-14
graphql: fields of object should be ANDed always

saihaj:
Let's take this query as an example
```graphql
{
  musicians(where: { or: [{name: "John", id: "m2"},{bandId: "7j"}]}){
    id
  }
}
```

**Before** when the filter for `OR/AND` was parsed and SQL was generated it would flatten all the filters and then all the fields were ORed so it would look like

```rust
EntityFilter::Or([{name: "John"}, {id: "m2"}, {bandId: "7j"}])
```

**After** this PR it will alway AND the fields in an object and then OR the adjacent fields so it would look like

```rust
EntityFilter::Or([EntityFilter::And([{name: "John"}, {id: "m2"}]), EntityFilter::And([{bandId: "7j"}])])
```

### DIFF
--- a/graphql/src/store/query.rs
+++ b/graphql/src/store/query.rs
@@ -206,22 +206,23 @@ fn build_list_filter_from_value(
     schema: &ApiSchema,
     value: &r::Value,
 ) -> Result<Vec<EntityFilter>, QueryExecutionError> {
+    // We have object like this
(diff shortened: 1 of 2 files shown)


dotansimha (approved):

lutter (approved):
thanks for the quick fix!

Issue #3978, opened 2022-09-22
graphman rewind not clearing failed status of deployment


paymog:
**Do you want to request a *feature* or report a *bug*?**

Bug, I think

**What is the current behavior?**

`graphman rewind` does not clear the `synced` or `failed` columns for a subgraph deployment

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem.**

I had a subgraph deployment which had a postgres entry like so

<img width="777" alt="image" src="https://user-images.githubusercontent.com/1986950/191805145-c1d7ee39-c945-4a5e-9f28-7db75d35ef89.png">

Note that it's is failed with a fatal error. It was stuck on block 15402220. I recently rewound it with graphman using a command like 

```
graphman rewind 0xa3643c8c86108af9f4f041f8880f333dd0f35a0b30eecdefad909cd6e7489fae 15401220 --sleep 300 <deployment id>
```

That graphman command succeeded in rewinding the subgraph to block 15401220 (as seen in the above screenshot).

However, when I check the subgraph deployment row for the deployment, I see that it still has a status of `synced=true` and `failed=true` even though it's been rewound.

**What is the expected behavior?**

I'd expect that rewinding a subgraph would clear (or set to false) the `synced` and `failed` columns for the deployment specified by the subgraph.

tilacog:
Thanks for the feedback, @paymog.
Can you confirm if the `synced` and `failed` fields are set to the expected values once the subgraph resumes syncing?

paymog:
I didn't notice those fields update to the expected value after the subgraph resumed indexing. I ended up setting them manually myself using psql.

paymog:
@tilacog have you been able to verify whether this is reproducible?

azf20:
Hi @paymog just to check, you are saying this is the case just immediately after rewind, or also after the subgraph starts syncing again? (Expected current behaviour is that the "failed" status is cleared on restart, while the "synced" status is not cleared after a subgraph has been "synced" for the first time)

paymog:
Immediately after rewind

fordN:
I'm seeing similar behavior.  After rewinding a subgraph it was able to sync back to chain head successfully and continue syncing healthily. However, the failed status has not cleared.  

Here is the current record in subgraphs.subgraph_deployment (you'll see it _does_ have a fatal_error): 
[...]

Issue #4123, opened 2022-11-02
Failed flag should be cleared even if there is no fatal error


leoyvens:
We've seen a few times subgraphs that are syncing fine, but have the `failed` flag set and no fatal error. I'm not sure how this happens, it probably shouldn't, but one thing we can easily address is that the unfail logic is not clearing the failed flag in this situation.

Here it is assuming that if there is no fatal error then the subgraph is not failed:
https://github.com/graphprotocol/graph-node/blob/ee76a6356f79368a61fbb7fbb47ea32568bdd3c7/store/postgres/src/deployment_store.rs#L1437-L1442

leoyvens:
What gets us into this situation could be a rewind or revert, which will delete errors but not set a failed subgraph back to healthy. See #3978.

PR #4174, opened 2022-11-15
Clear `failed` status on graphman rewind

evaporei:
Fixes https://github.com/graphprotocol/graph-node/issues/3978 and https://github.com/graphprotocol/graph-node/issues/4123

This is being done by changing `revert_subgraph_errors` to change both
the `failed` and `status` columns in the case of both being `failed`.
    
Without this change, rewinding failed subgraphs past the failed
block would result in the `failed` status staying the same.


store/postgres/src/deployment_store.rs
```diff
@@ -1174,6 +1174,18 @@ impl DeploymentStore {
                     layout.count_query.as_str(),
                     count,
                 )?;
+
+                // Delete subgraph errors
+                deployment::revert_subgraph_errors(conn, &site.deployment, block_ptr_to.number)?;
+
+                // Clear `failed` column back to `false`
+                deployment::update_deployment_status(
```
evaporei, line 1182: @leoyvens I tried to think a few scenarios where we shouldn't call this, but I think the failed block will always be the last one and we'll always want the deployment to be healthy after a rewind.

Do you have any worries/considerations here?

leoyvens: There is the case of non-fatal errors, where the subgraph health would be `Unhealthy`. If this checked that the health is actually failed I think it would be functionally correct, but we need a test.

Also there are some code organization issues with this approach. `revert_subgraph_errors` is already called from `revert_metadata` so that's redundant. And the health update should be moved into `revert_subgraph_errors` imo.

evaporei: Thanks for the feedback! I've addressed it by stopping calling `revert_subgraph_errors` twice. And changing it to also update the `failed` and `status` columns in the case of both of them being `failed` :wink: 

Do you think we should change an integration test for this?

leoyvens: Nice! A test would be great

neysofu (approved):

Issue #4169, opened 2022-11-14
Resolve Clippy warnings and maybe enforce in CI?


xJonathanLEI:
_(Sorry for not using the template, but this is not a feature or bug report)_

Currently, Clippy generates _89_ warnings on the entire workspace:

```console
$ cargo +nightly clippy -Z unstable-options --keep-going --all --all-targets
--- SNIP ---
warning: `graph` (lib) generated 89 warnings
error: could not compile `graph` due to previous error; 89 warnings emitted
```

CI only _reports_ the number of warnings without blocking merges. I think I undertand the rationale behind this design, but the number of warnings is making it hard for making contributions - it's hard to tell if new Clippy warnings are introduced in the new code added.

Should we start fixing the Clippy warnings, and (maybe) enforce the Clippy lints on CI?

PR #4178, opened 2022-11-16
Fix a bunch of Clippy warnings

neysofu:
As per title. I didn't apply the fixes that were controversial the last time I did a batch of these (e.g. `.unwrap_or_default()`), so I think we should be good.

See #4169.

### DIFF
--- a/chain/arweave/src/adapter.rs
+++ b/chain/arweave/src/adapter.rs
@@ -72,9 +72,7 @@ impl ArweaveTransactionFilter {
             })
             .map(|ds| match &ds.source.owner {
                 Some(str) if MATCH_ALL_WILDCARD.eq(str) => MATCH_ALL_WILDCARD.as_bytes().to_owned(),
-                owner @ _ => {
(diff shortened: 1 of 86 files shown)


neysofu:
Mmm. On one hand I find a lot of value in deterministic linting above style preference, which is why I really like Clippy; on the other hand I understand the frustration with specific lints. In the future I would really like to explore configuration options for Clippy that we can agree on, so that we reap the benefits without having to go through these reviews.

xJonathanLEI:
Sorry for commenting on a merged PR. I think it makes a lot of sense to ignore certain lints that the maintainers don't agree on (especially ones that come with perf penalty), but as @neysofu mentioned having linting that acutally works has a lot of value. For one, it makes contributing to the codebase much easier.

I understand this is not something the team is actively looking to do, but can I maybe submit a PR fixing all the clippy warnings/errors, and for the ones the team doesn't agree, I just update the PR with cliippy ignore rules instead?

_P.S. I maintain the [StarkNet support fork](https://github.com/starknet-graph/graph-node) and it's pretty tough for me to check whether my patch or incoming PRs are abiding by the lints._

PR #4184, opened 2022-11-18
Enable thin LTO for release builds

neysofu:
I would hardly expect this to cause a performance regression; at most we won't see any improvements. So I think we're good rolling this out before further investigation.

This results in a x2.12 `--release` compilation time increase on my machine.

@lutter @leoyvens

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,5 +25,10 @@ license = "MIT OR Apache-2.0"
 [profile.test]
 incremental = false
 
+[profile.release]
+lto = "thin"
+codegen-units = 1
+incremental = false
+
 [profile.dev]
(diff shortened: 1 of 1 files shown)


leoyvens:
A 2x slowdown in release builds is a considerable downside. How will we measure any performance benefits? Is it possible to measure without merging to master?

neysofu:
I was thinking of comparing CPU usage before and after a production rollout, and reverting the commit if we see no improvement. Do you think we should do an isolated rollout to a single node, just like we discussed for a different allocator?

leoyvens:
Yes, I think we can rollout to a single node, and from a tag on this PR rather than master.

PR #4188, opened 2022-11-19
Code coverage experimentations

neysofu:
WIP.


neysofu:
How about we merge this? We can always improve things down the line, but at least we'll start getting nightly reports. @mangas

mangas (approved):

Issue #4195, opened 2022-11-22
Substream failed 


mangas:
```
20/11/2022, 23:15:27

ERROR

Subgraph failed with non-deterministic error: Failed to transact block operations: subgraph writer poisoned by previous error, retry_delay_s: 1800, attempt: 116

20/11/2022, 23:15:27

DEBUG

1 candidate trigger in this block, block_hash: 0xc5e1b7effa3d0a8c4d2bd2631d3b37c46b7ec20c59194e4bd5636525838dd082, block_number: 12287508

20/11/2022, 23:15:26

INFO

Blockstreams connected, component: SubstreamsBlockStream

20/11/2022, 23:15:26

INFO

Blockstreams disconnected, connecting, cursor: 8mpdkyxbzQ0fquuHrDolBaWwLpc_DFttVwPiLxFGj46k9nLAi8vzCWYgbRWDlqz3iUHoQ1Kkjd_MEC9_9ZYG6tHixu5n6CdpEy99kY66rrDpffr2aw0acO8zXePbY9DRWj7Vag3-erYK5YfhOvaPMRc7ZZIjJDW1ijgE9oZceKtCuXIxkjj6JZmE0POTp4FE-bNwELWjkn_2AzN4JUsIOs2DM6OdtmkhbQ==, start_block: 12287507, endpoint_uri: mainnet-substreams-0-0, component: SubstreamsBlockStream

20/11/2022, 23:15:26

DEBUG

Starting block stream

```

Replicate: 

```
cargo run -p graph-node -- --config ~/git/graph-node-dev/config/eth-mainnet-substreams.toml --ipfs https://api.thegraph.com/ipfs/ --start-block 0x02f994fd308229a46c09127837764dfdfb4e025ee4cecd1d92a02f54f07a0330:12777423 --subgraph QmVDT9fsWF3nBez71RjRBTRGVH7mDxSoZQem7Fbc6QDdDA
```

Output
```

Nov 21 22:02:15.691 INFO Blockstreams connected, sgd: 1, subgraph_id: QmVDT9fsWF3nBez71RjRBTRGVH7mDxSoZQem7Fbc6QDdDA, component: SubstreamsBlockStream
[chain/substreams/src/mapper.rs:63] &logger = Logger(component, sgd, subgraph_id)
[chain/substreams/src/mapper.rs:63] "received block {}" = "received block {}"
[chain/substreams/src/mapper.rs:63] number = 12287507
[...]

PR #4197, opened 2022-11-22
ensure substreams-based-subgraphs can be started at specific block or rewound

sduchesneau:
Fixes https://github.com/graphprotocol/graph-node/issues/4195

This feature was omitted in substreams initially, but it is needed for `--start-block` feature, rewind, and probably in other "recovery" scenarios where we cannot start from a cursor.


maoueh (approved):

mangas:
@sduchesneau change looks good, you need to fix the warning

PR #4201, opened 2022-11-23
Fix: run locale check after migrations

tilacog:
PR #4163 introduced a locale check, which runs before migrations are applied.

This PR alters this check, so it runs immediately after migrations.

### DIFF
--- a/store/postgres/src/connection_pool.rs
+++ b/store/postgres/src/connection_pool.rs
@@ -980,18 +980,6 @@ impl PoolInner {
         let conn = self.get().map_err(|_| StoreError::DatabaseUnavailable)?;
 
         let start = Instant::now();
-        if let Err(msg) = catalog::Locale::load(&conn)?.suitable() {
-            if &self.shard == &*PRIMARY_SHARD && primary::is_empty(&conn)? {
-                die(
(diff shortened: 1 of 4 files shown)


store/postgres/src/connection_pool.rs
```diff
@@ -1024,6 +1012,20 @@ impl PoolInner {
...
+                    &pool.logger,
+                    "Database does not use C locale. \
+                    Please check the graph-node documentation for how to set up the database locale",
+                    &msg,
+                );
+            } else {
+                warn!(pool.logger, "{}.\nPlease check the graph-node documentation for how to set up the database locale", msg);
+            }
+        }
+
```
lutter, line 1028: To help with the clarity of logging, it would be better to have this check just before "Setup finished" is logged.

lutter (approved):

store/postgres/src/connection_pool.rs
tilacog: Makes sense.
Fixed in 3ca714f83f221519931f208eab124bc1cf56adde

leoyvens:
CI failures look legit. Seems we're not using the C locale on CI?

Issue #4059, opened 2022-10-12
Add a Firehose Block Ingestor for Ethereum


azf20:
### Situation
Currently, Graph Node configured to use a Firehose for Ethereum indexing still needs to connect to an RPC endpoint in order to ingest blocks to keep the chain head pointer up to date.

### Resolution
Add a Firehose block ingestor for Ethereum, similar to the implementations for other protocols without an existing RPC setup

This should be configurable (i.e. to prefer an RPC or Firehose Block Ingestor)

paymog:
Do we know when this fix will land in a usable docker image?

mangas:
@paymog it looks like it might only be release with 0.30 in Jan.

PR #4204, opened 2022-11-25
enable firehose eth block ingestor

mangas:
Fixes https://github.com/graphprotocol/graph-node/issues/4059

### DIFF
--- a/node/src/main.rs
+++ b/node/src/main.rs
@@ -1,5 +1,6 @@
 use clap::Parser as _;
 use ethereum::chain::{EthereumAdapterSelector, EthereumBlockRefetcher, EthereumStreamBuilder};
+use ethereum::codec::HeaderOnlyBlock;
 use ethereum::{
     BlockIngestor as EthereumBlockIngestor, EthereumAdapterTrait, EthereumNetworks, RuntimeAdapter,
 };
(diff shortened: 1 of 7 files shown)


node/src/main.rs
```diff
@@ -373,13 +374,37 @@ async fn main() {
...
+                    .into_iter()
+                    .partition(|(_, chain)| chain.is_firehose_supported());
 
-                start_block_ingestor(
-                    &logger,
-                    &logger_factory,
-                    block_polling_interval,
-                    ethereum_chains,
-                );
+                start_block_ingestor(&logger, &logger_factory, block_polling_interval, polling);
```
maoueh, line 381: Should we avoid completely the RPC block ingestor if Firehose is enabled? The Firehose one will have better latency for new blocks in general, so seems a fair assumption to use it exclusively if an endpoint for it exist.

mangas: This behaviour is implicit because the IS_FIREHOSE_PREFERRED env var is not set per chain. This means that if it's set and there are firehose endpoints, then firehose is used for ingestor. On the other hand if a specific chain does not have firehose endpoints it needs to run RPC

mangas: We also can't remove this completely because some ethereum compatible chains may exist and use the RPC ingestor

maoueh: The way I'm reading the current code for when both an RPC and a Firehose provider are set is the following (please correct any wrong assumptions about the behavior of your PR):

  ```
    ...
    provider = [
      { label = "firehose", details = { type = "firehose", url = "<url>", token = "<auth_token>", features = [ "filters" ] }},
      { label = "peering", url = "<ethereum_full_node_rpc>", features = [ "archive" ] },
    ]
  ```
  
In this case, `start_block_ingestor` is called which will launch the current `EthereumBlockIngestor`. And since there is a Firehose provider configured for Ethereum, `start_firehose_block_ingestor` will also be called.

I'm commenting on this specific thing where only one of EthereumBlockIngestor or `FirehoseBlockIngestor` must started for a given chain.

maoueh:
Also, I don't see in the PR where we handle the current `EthereumBlockIngestor` [upsert full block block logic](https://github.com/graphprotocol/graph-node/blob/6792c9c633682b8a01172275f3be3c8ef3d48f70/chain/ethereum/src/ingestor.rs#L196-L199). Is it scheduled for another follow-up PR?

maoueh:
For now, no receipts will be written when using a Firehose Block Ingestor on Ethereum, it's not really required anyway since Firehose is able to stream that information directly, so there is no point in saving that information right now.

Only draw back is that switching back to RPC provider is not possible without manual intervetion on the database by removing the accumulated blocks so far before switching back to RPC.

node/src/main.rs
```diff
@@ -373,13 +374,37 @@ async fn main() {
...
+
+                firehose_networks_by_kind
+                    .get(&BlockchainKind::Ethereum)
+                    .map(|eth_firehose_endpoints| {
+                        start_firehose_block_ingestor::<_, Block>(
```
maoueh, line 386: So there is a final thing that I would like to see how we can implement. Right now, Firehose streams a full block back for ingestor but this is definitely not required, all the traces for example, they could be skipped. 

First quick optimization would be to use `HeaderOnlyBlock` struct here, so we decode only a subset of the returned payload reducing greatly the required memory for it.

Second, on Ethereum (and some other networks), we have a specialized transform that stream only a really light block, which make sense here for the block ingestor. Ideally, we would support custom transforms per chain and activate it here, that would greatly reduced the payload between `graph-node` and `Firehose`.

Could be left for a second PR though.

leoyvens:
Code looks safe to merge. But I'll let @maoueh, who did the in-depth review, give the green check.

maoueh (approved):

paymog:
> For now, no receipts will be written when using a Firehose Block Ingestor on Ethereum, it's not really required anyway since Firehose is able to stream that information directly, so there is no point in saving that information right now.
> 
> Only draw back is that switching back to RPC provider is not possible without manual intervetion on the database by removing the accumulated blocks so far before switching back to RPC.

would it be good to add a graphman command to make it easy to roll back to rpc only providers? is it the `chainX.blocks` table that needs to be wiped during a rollback?

Issue #4154, opened 2022-11-09
Cache recent blocks in memory


lutter:
It appears that the fact that each synced subgraph loads the chain head block from the database causes considerable load on databases backing a `ChainStore`. To alleviate that, each `ChainStore` should have a small in-memory cache (say 5 or 10 blocks) so that recent blocks only need to be loaded once per index node instead of once per subgraph.

Issue #17, opened 2018-05-01
Speed up Rust builds by caching ~/.cargo and target/


Jannis:
Travis does the same:

* Travis issue: https://github.com/travis-ci/travis-ci/issues/4470
* Travis implementation: https://github.com/travis-ci/travis-build/commit/624981b17372eeed331c64fdf49468b15fcf1f7c

Issue #14, opened 2018-04-30
Set up cluster monitoring with Google Stackdriver


Jannis:
We'll want to monitor our standalone node (uptime etc.). Google Stackdriver can do this for us and will hopefully integrate well with Kubernetes.

PR #4215, opened 2022-11-30
Cache recent blocks in memory to reduce load on the db

neysofu:
Closes https://github.com/graphprotocol/graph-node/issues/4154.

### DIFF
--- a/store/postgres/src/chain_store.rs
+++ b/store/postgres/src/chain_store.rs
@@ -1686,36 +1686,37 @@ impl ChainStoreTrait for ChainStore {
 
     async fn ancestor_block(
         self: Arc<Self>,
-        ancestor_ptr: BlockPtr,
+        block_ptr: BlockPtr,
         offset: BlockNumber,
     ) -> Result<Option<json::Value>, Error> {
         ensure!(
-            ancestor_ptr.number >= offset,
+            block_ptr.number >= offset,
             "block offset {} for block `{}` points to before genesis block",
             offset,
-            ancestor_ptr.hash_hex()
+            block_ptr.hash_hex()
         );
 
         // Check the local cache first.
-        if let Some(data) = self.recent_blocks_cache.get_block(&ancestor_ptr, offset) {
+        if let Some(data) = self.recent_blocks_cache.get_ancestor(&block_ptr, offset) {
             return Ok(data.1.clone());
         }
 
-        let ancestor_ptr_clone = ancestor_ptr.clone();
+        let block_ptr_clone = block_ptr.clone();
         let chain_store = self.cheap_clone();
         let block_data = self
             .pool
             .with_conn(move |conn, _| {
                 chain_store
                     .storage
-                    .ancestor_block(conn, ancestor_ptr_clone, offset)
-                    .map_err(|e| CancelableError::from(StoreError::from(e)))
+                    .ancestor_block(conn, block_ptr_clone, offset)
+                    .map_err(StoreError::from)
+                    .map_err(CancelableError::from)
             })
             .await?;
 
         if let Some((data, ptr)) = block_data {
             // Update the local cache.
-            self.recent_blocks_cache.set_block(ancestor_ptr, ptr, &data);
+            self.recent_blocks_cache.set_ancestor(block_ptr, ptr, &data);
 
             Ok(Some(data))
         } else {
@@ -1844,73 +1845,73 @@ impl ChainStoreTrait for ChainStore {
 mod recent_blocks_cache {
     use super::*;
 
-    type OffsetFromChainHead = BlockNumber;
-
     struct Inner {
-        blocks: HashMap<OffsetFromChainHead, (BlockPtr, Option<json::Value>)>,
+        chain_head: Option<BlockNumber>,
+        blocks: HashMap<BlockNumber, (BlockPtr, Option<json::Value>)>,
         capacity: usize,
     }
 
     impl Inner {
         fn set_chain_head(&mut self, chain_head: BlockPtr) {
+            self.chain_head = Some(chain_head.number);
             // We invalidate the entire cache when the chain head changes.
             self.blocks.clear();
-            self.blocks.insert(0, (chain_head, None));
+            self.blocks.insert(chain_head.number, (chain_head, None));
         }
 
-        fn get_block(
+        fn get_ancestor(
             &self,
-            ancestor: &BlockPtr,
+            child: &BlockPtr,
             offset: BlockNumber,
         ) -> Option<(BlockPtr, Option<json::Value>)> {
-            // Check if the ancestor itself is in the cache.
-            let ancestor_offset_from_head = self.chain_head_number()? - ancestor.number;
-            match self
-                .blocks
-                .get(&ancestor_offset_from_head)
-                .map(|p| p.0.clone())
-            {
-                None => return None,
-                Some(ptr) if ptr.hash != ancestor.hash => return None,
-                _ => {}
-            }
+            let child_is_cached = match self.blocks.get(&child.number) {
+                Some((ptr, _)) => ptr.hash == child.hash,
+                _ => false,
+            };
 
-            let offset_from_head = ancestor_offset_from_head + offset;
-            self.blocks.get(&offset_from_head).cloned()
+            if child_is_cached {
+                self.blocks.get(&(child.number - offset)).cloned()
+            } else {
+                None
+            }
         }
 
-        fn set_block(&mut self, ancestor: BlockPtr, ptr: BlockPtr, data: &json::Value) {
+        fn contains_block(&self, ptr: &BlockPtr) -> bool {
+            let identity_offset = 0;
+            self.get_ancestor(ptr, identity_offset).is_some()
+        }
+
+        /// Tentatively caches the `ancestor` of a [`BlockPtr`], together with
+        /// its associated `data`.
+        fn set_ancestor(&mut self, child: BlockPtr, ancestor: BlockPtr, data: &json::Value) {
             // We must have a chain head to cache blocks.
-            let chain_head_number = if let Some(n) = self.chain_head_number() {
+            let chain_head = if let Some(n) = self.chain_head {
                 n
             } else {
                 return;
             };
 
-            let offset_from_head = chain_head_number - ptr.number;
             // We only cache blocks that are close to the chain head.
+            let offset_from_head = chain_head - ancestor.number;
             if offset_from_head >= self.capacity as _ {
                 return;
             }
-            // We only cache blocks if their ancestor is in the cache (chain head excluded).
-            if self.get_block(&ancestor, 0).is_none() {
+
+            // We only cache ancestors of blocks that are cached already (chain
+            // head excluded). Otherwise, we might inadvertently cache blocks
+            // that don't belong to the main chain, e.g. uncle blocks.
+            if !self.contains_block(&child) {
                 return;
-            };
+            }
 
             self.blocks
-                .insert(offset_from_head, (ptr, Some(data.clone())));
-        }
-
-        fn chain_head_number(&self) -> Option<BlockNumber> {
-            self.blocks.get(&0).map(|p| p.0.number)
+                .insert(ancestor.number, (ancestor, Some(data.clone())));
         }
     }
 
     /// We cache the most recent blocks in memory to avoid overloading the
     /// database with unnecessary queries close to the chain head. We invalidate
-    /// blocks:
-    /// - After 5 seconds.
-    /// - Whenever the chain head advances.
+    /// blocks whenever the chain head advances.
     pub struct RecentBlocksCache {
         // We protect everything with a global `RwLock` to avoid data races. Ugly...
         inner: RwLock<Inner>,
@@ -1920,6 +1921,7 @@ mod recent_blocks_cache {
         pub fn new(capacity: usize) -> Self {
             RecentBlocksCache {
                 inner: RwLock::new(Inner {
+                    chain_head: None,
                     blocks: HashMap::new(),
                     capacity,
                 }),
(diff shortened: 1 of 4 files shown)


neysofu:
Marked as draft; I need to fix a few things before review.

neysofu:
1. Change the `TimedCache` into a `HashMap` or `Vec`.
2. Make the capacity configurable.

neysofu:
@lutter The PR is ready... but `remove-then-update` fails, and I can't figure out why. Any idea?

store/postgres/src/chain_store.rs
```diff
@@ -1813,6 +1841,120 @@ impl ChainStoreTrait for ChainStore {
...
+    struct Inner {
+        blocks: HashMap<OffsetFromChainHead, (BlockPtr, Option<json::Value>)>,
+        capacity: usize,
+    }
+
+    /// We cache the most recent blocks in memory to avoid overloading the
+    /// database with unnecessary queries close to the chain head. We invalidate
+    /// blocks:
+    /// - After 5 seconds.
+    /// - Whenever the chain head advances.
```
lutter, line 1858: That comment needs to be updated since there is no TTL anymore

store/postgres/src/chain_store.rs
```diff
@@ -1813,6 +1841,120 @@ impl ChainStoreTrait for ChainStore {
...
+                    capacity,
+                }),
+            }
+        }
+
+        pub fn set_chain_head(&self, chain_head: BlockPtr) {
+            // We invalidate the entire cache when the chain head changes.
+            let mut inner = self.inner.write();
+            inner.blocks.clear();
+            inner.blocks.insert(0, (chain_head, None));
```
lutter, line 1878: This is overly aggressive, and will reduce the effectiveness of the cache; it should be enough to remove any entry that is not within `[chain_head - capacity + 1, chain_head]` which would help reduce the churn from this cache. In normal operation, we'd only need to remove one block from the cache when the chain head advances.

I also think it would be simpler if the cache didn't require a call to `set_chain_head` and instead infer that from `set_block`; that makes it less likely that future code changes forget to call this and make the cache ineffective.

store/postgres/src/chain_store.rs
```diff
@@ -1813,6 +1841,120 @@ impl ChainStoreTrait for ChainStore {
...
+
+        pub fn chain_head_ptr(&self) -> Option<BlockPtr> {
+            self.inner.read().blocks.get(&0).map(|p| p.0.clone())
+        }
+
+        pub fn clear(&self) {
+            self.inner.write().blocks.clear();
+        }
+
+        fn chain_head_number_inner(inner: &Inner) -> Option<BlockNumber> {
```
lutter, line 1889: More a stylistic comment: why not make these part of an `impl Inner`? Not a big deal though

store/postgres/src/chain_store.rs
lutter: Oh .. I just realized that the hashmap is keyed off the distance from the chain head. If using a `HashMap`, why not just use the block number directly, even if that makes getting the current max block number a bit more expensive? If we want to go with `0 == chain_head`, then a `VecDeque` might be a better data structure.

store/postgres/src/chain_store.rs
```diff
@@ -1813,6 +1841,120 @@ impl ChainStoreTrait for ChainStore {
...
+            offset: BlockNumber,
+        ) -> Option<(BlockPtr, Option<json::Value>)> {
+            Self::get_block_inner(&*self.inner.read(), ancestor, offset)
+        }
+
+        fn get_block_inner(
+            inner: &Inner,
+            ancestor: &BlockPtr,
+            offset: BlockNumber,
+        ) -> Option<(BlockPtr, Option<json::Value>)> {
```
lutter, line 1905: This is used in a few places to check if that block is in the cache; it has to clone the cached value because of this return type. It would be better to either return references here or to have a separate `has_block_inner` method for places where we just need to check if a block is cached.

store/postgres/src/chain_store.rs
```diff
@@ -1813,6 +1841,120 @@ impl ChainStoreTrait for ChainStore {
...
+            // Check if the ancestor itself is in the cache.
+            let ancestor_offset_from_head = Self::chain_head_number_inner(inner)? - ancestor.number;
+            match inner
+                .blocks
+                .get(&ancestor_offset_from_head)
+                .map(|p| p.0.clone())
```
lutter, line 1911: It's minor, but is this `clone()` necessary?

store/postgres/src/chain_store.rs
```diff
@@ -1813,6 +1841,120 @@ impl ChainStoreTrait for ChainStore {
...
+            } else {
+                return;
+            };
+
+            let offset_from_head = chain_head_number - ptr.number;
+            // We only cache blocks that are close to the chain head.
+            if offset_from_head >= inner.capacity as _ {
+                return;
+            }
+            // We only cache blocks if their ancestor is in the cache (chain head excluded).
```
lutter, line 1946: The impact of this isn't entirely clear to me; could you add a comment why we do that?

store/postgres/src/chain_store.rs
```diff
@@ -1813,6 +1841,120 @@ impl ChainStoreTrait for ChainStore {
...
+
+        pub fn set_chain_head(&self, chain_head: BlockPtr) {
+            // We invalidate the entire cache when the chain head changes.
+            let mut inner = self.inner.write();
+            inner.blocks.clear();
+            inner.blocks.insert(0, (chain_head, None));
+        }
+
+        pub fn chain_head_ptr(&self) -> Option<BlockPtr> {
+            self.inner.read().blocks.get(&0).map(|p| p.0.clone())
```
lutter, line 1882: It would be more readable to use a constant instead of `0` for these lookups

lutter (changes requested):
This looks generally good. It just occurred to me since this touches a very sensitive part of the code, that we should run this change in the integration cluster before merging it.

store/postgres/src/chain_store.rs
neysofu: You're absolutely right!

store/postgres/src/chain_store.rs
neysofu: It is, unfortunately. We can't get a reference out of `self` because it's all guarded by the `RwLock`.

store/postgres/src/chain_store.rs
neysofu: Done, good catch!

store/postgres/src/chain_store.rs
neysofu: I added a `contains_blocks` method on `Inner` (so it's not available in the public API, but it can be made public if we need it in other places). Unfortunately we can't easily return references (we could do `Arc`'s though) because of locking.

store/postgres/src/chain_store.rs
neysofu: I'm now indexing by block number, and not chain head offset. So we're good here!

store/postgres/src/chain_store.rs
neysofu: The comment wasn't really truthful, in retrospect. I tried to make a better job at addressing the issue in the comment and I'll probably add some others. We only want to add blocks to the cache if a child of that block is already present in the cache, because that's the easiest way to ensure the block is part of the main chain with the correct head. Otherwise, we might end up storing uncle blocks.

store/postgres/src/chain_store.rs
```diff
@@ -1813,6 +1841,164 @@ impl ChainStoreTrait for ChainStore {
...
+            } else {
+                None
+            }
+        }
+
+        fn chain_head(&self) -> Option<&BlockPtr> {
+            self.blocks.iter().next_back().map(|b| &b.1.ptr)
+        }
+
+        fn last_block(&self) -> Option<&CachedBlock> {
```
lutter, line 1887: I got a little confused here - isn't that the first block, i.e., the one with the lowest block number? Calling this `first_block` or `earliest_block` would be clearer

store/postgres/src/chain_store.rs
```diff
@@ -1813,6 +1841,164 @@ impl ChainStoreTrait for ChainStore {
...
+            let chain_head = if let Some(ch) = self.chain_head() {
+                ch
+            } else {
+                // We don't have anything in the cache, so we're free to store
+                // everything we want.
+                self.blocks.insert(block.ptr.number, block);
+                return;
+            };
+
+            if chain_head.number == block.ptr.number - 1 && chain_head.hash == block.parent_hash {
```
lutter, line 1919: It might be good to have a utility function to check if one block is the parent of another to make this more readable

store/postgres/src/chain_store.rs
```diff
@@ -1813,6 +1841,164 @@ impl ChainStoreTrait for ChainStore {
...
+                // We have a new chain head that is a direct child of our
+                // previous chain head, so we get to keep all items in the
+                // cache.
+                self.blocks.insert(block.ptr.number, block);
+            } else if block.ptr.number > chain_head.number && block.parent_hash != chain_head.hash {
```
lutter, line 1924: I don't understand this condition; why is the `block.parent_hash != chain_head.hash` needed? We already checked that we didn't advance the chain head by one in the previous branch, so if `block.ptr.number > chain_head.number` we somehow jumped around and the old cache contained a different branch

store/postgres/src/chain_store.rs
lutter: I think using an `Arc<Option<json::Value>>` would be a good thing, but it will require lots of changes across the code base. It would be a very good as a separate follow-on to this PR

store/postgres/src/chain_store.rs
```diff
@@ -1813,6 +1841,164 @@ impl ChainStoreTrait for ChainStore {
...
+        }
+
+        fn last_block(&self) -> Option<&CachedBlock> {
+            self.blocks.iter().next().map(|b| b.1)
+        }
+
+        fn evict_if_necessary(&mut self) {
+            while self.blocks.len() > self.capacity && !self.blocks.is_empty() {
+                let block_num = *self.blocks.iter().next().unwrap().0;
+                self.blocks.remove(&block_num);
```
lutter, line 1894: Isn't this just `self.blocks.pop_first()` ?

lutter (changes requested):
This looks good in general; I would mostly like to understand the logic in `insert_block` a little better before approving

neysofu:
Integration testing went well - it's on `fraction` anybody wants to take a look.

store/postgres/src/chain_store.rs
```diff
@@ -1813,6 +1841,164 @@ impl ChainStoreTrait for ChainStore {
...
+                let ancestor_block_number = child.number - offset;
+                let block = self.blocks.get(&ancestor_block_number)?;
+                Some((&block.ptr, block.data.as_ref()))
+            } else {
+                None
+            }
+        }
+
+        fn chain_head(&self) -> Option<&BlockPtr> {
+            self.blocks.iter().next_back().map(|b| &b.1.ptr)
```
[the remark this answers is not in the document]
neysofu: Yup, same as a `pop_first` - this was stabilized so now it's fair game :+1:

store/postgres/src/chain_store.rs
```diff
@@ -1813,6 +1841,164 @@ impl ChainStoreTrait for ChainStore {
...
+                // cache.
+                self.blocks.insert(block.ptr.number, block);
+            } else if block.ptr.number > chain_head.number && block.parent_hash != chain_head.hash {
+                // We have a new chain head, but it's not a direct child of
+                // our previous chain head. This means that we must
+                // invalidate all the items in the cache before inserting
+                // this block.
+                self.blocks.clear();
+                // Try again after clearing the cache.
+                self.insert_block(block.ptr, block.data, block.parent_hash);
```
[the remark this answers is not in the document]
neysofu: You're right, much more readable now!

store/postgres/src/chain_store.rs
neysofu: >  so if block.ptr.number > chain_head.number we somehow jumped around and the old cache contained a different branch

Indeed. I think in that case we want to clear the cache and remove all blocks that belong to the old branch, right?

lutter: Yes, I think what happens in the branch is right, I just don't understand why the condition `block.ptr.number > chain_head.number` isn't enough here - why the additional condidition `block.parent_hash != chain_head.hash`?

store/postgres/src/chain_store.rs
```diff
@@ -1813,6 +1841,164 @@ impl ChainStoreTrait for ChainStore {
...
+        }
+
+        fn insert_block(
+            &mut self,
+            ptr: BlockPtr,
+            data: Option<json::Value>,
+            parent_hash: BlockHash,
+        ) {
+            fn is_parent_of(parent: &BlockPtr, child: &CachedBlock) -> bool {
+                child.ptr.number == parent.number + 1 && child.parent_hash == parent.hash
```
maoueh, line 1904: `+ 1` and `- 1` on block number is not correct, NEAR for example don't have strictly continuous block number, a valid chain in NEAR can be `#14 (Hash ABC1, Parent XX) -> #17 (Hash EBD2, Parent ABC1)` is valid.

I imagine this code will be hit by NEAR, so it's important to have in mind. I did not check the rest of this PR, but it review of `+ 1` nor `- 1` logic on block number should be carefully reviewed.

lutter: Good point. I think the checks by number are redundant anyway since the hash/parent hash chain should be enough to establish these relationships.

store/postgres/src/chain_store.rs
neysofu: Oh, I got it now. You're right, it's superfluous. The NEAR fix in https://github.com/graphprotocol/graph-node/pull/4215/commits/6d46bacb80dc312f9cfb32c49b581b3ce33c571c removes block number checks entirely (keeping only hash comparisons), so this code doesn't exist anymore.

store/postgres/src/chain_store.rs
neysofu: Good catch! Fixed with https://github.com/graphprotocol/graph-node/pull/4215/commits/6d46bacb80dc312f9cfb32c49b581b3ce33c571c. By relying entirely on parent hashes without any block number comparison, I think we should be good.

neysofu:
Addressed all comments.

lutter (approved):
Nice work. Curious to see it in action.

PR #4186, opened 2022-11-18
docker: use exec form for CMD directive

islishude:
the `CMD start` is equal with `CMD sh -c start`, so pid of graph node is not 1

image

PR #4217, opened 2022-12-01
docker: use exec form for CMD directive

islishude:
the `CMD start` is equal with `CMD sh -c start`, so pid of graph node is not 1

image

#4186


neysofu (approved):
Works great! Let's merge.

PR #4219, opened 2022-12-02
fail when ens rainbow not present

mangas:
Closes https://github.com/graphprotocol/graph-node/issues/4206

### DIFF
--- a/store/postgres/src/subgraph_store.rs
+++ b/store/postgres/src/subgraph_store.rs
@@ -6,7 +6,7 @@ use diesel::{
 };
 use std::{
     collections::{BTreeMap, HashMap},
-    sync::{Arc, Mutex},
+    sync::{atomic::AtomicU8, Arc, Mutex},
 };
 use std::{fmt, io::Write};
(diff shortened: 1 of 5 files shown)


store/postgres/src/subgraph_store.rs
```diff
@@ -1147,23 +1147,67 @@ impl SubgraphStoreInner {
...
+/// loaded. To avoid unnecessary database roundtrips, the empty table check is lazy
+/// and will not be retried. Once the table is checked, any subsequent calls will
+/// just used the stored result.
 struct EnsLookup {
     primary: ConnectionPool,
+    // In order to keep the struct lock free, we'll use u8 for the status:
+    // 0 - Not Checked
+    // 1 - Checked - empty
+    // 2 - Checked - non empty
+    state: AtomicU8,
```
leoyvens, line 1164: This caching has the downside of requiring a graph node reset after the names are loaded.

I don't think this is worth it as an optimization, the `fn is_ens_data_empty` query is cheap and it is run only if the name lookup returned empty.

leoyvens (approved):

PR #4232, opened 2022-12-07
Improve `graphman index list`

lutter:
This was part of the work I did on partial indexes (still unfinished) The code parses index definitions, but most importantly makes the output of `graphman index list` more readable, including fancy colors, and adds options to filter out default indexes and the like.

Example:
image


node/src/manager/commands/index.rs
```diff
@@ -44,13 +55,130 @@ pub async fn list(
...
+
+    if to_sql {
+        for index in indexes {
+            writeln!(term, "{};", index.to_sql(concurrent, if_not_exists)?)?;
+        }
+    } else {
+        let mut first = true;
+        header(&mut term, &indexes, &deployment_locator, entity_name)?;
+        for index in &indexes {
+            if !first {
```
tilacog, line 176: It seems this condition will never be satisfied, as `first` is never set to `false` outside that `if` expression.

```suggestion
            if first {
```

store/postgres/src/relational/index.rs
```diff
@@ -0,0 +1,848 @@
...
+
+        match method {
+            "brin" => Ok(Brin),
+            "btree" => Ok(BTree),
+            "gin" => Ok(Gin),
+            "gist" => Ok(Gist),
+            _ => Err(method),
+        }
+    }
+}
```
tilacog, line 62: Considering our input is borrowed, there is no need to return it as an `Error`. 
I couldn't find where we would use the `Error` value on the rest of this module, so I believe it would be safe just to use `()` here.

Also, there's a debate about whether to use `FromStr/str::parse()` instead of `TryFrom<&str>::try_from`, since the former would be more readable, but that's subjective.

Personally, I find `FromStr` more practical because it doesn't require an import line on every module that uses it.

tilacog (approved):
LGTM
There's just one code change that I think should be reviewed.

This is impressive.
It fundamentally changes how we deal with indices, which I think will be the precursor of promising new features.

node/src/manager/commands/index.rs
lutter: Oh, yes - that was totally busted; fixing it makes the output a lot more readable

store/postgres/src/relational/index.rs
lutter: I agree - I changed this to use `FromStr`

PR #4249, opened 2022-12-16
Refactor integration tests

neysofu:
This is a refactoring of our integration tests and "runner" tests.

1. I'm moving the subgraphs designed for runner tests into a different directory named `tests/runner-tests/`, marking a clear separation between integration and runner tests. The new `yarn.lock` is a duplicate of the one we already had.
2. `runner_tests.rs` is mostly untouched. There's a new helper called `RunnerTestRecipe`, but that's about it.
3. I'm changing the way we generate Docker container names. Instead of global, atomic variables that provide numeric ids, I'm using v4 UUIDs prefixed with descriptive names.
4. `tokio::join!` a lot of `.await`'s, to run things as fast as possible. I also removed the code that waits 100ms after every output line, which I think was inserted out of an abundance of caution but doesn't seem to help with test flakiness.
5. More code polishing in `parallel_tests.rs` (renamed to `integration_tests.rs` for consistency), which mostly doesn't impact the way tests are run.


leoyvens (approved):
Thanks for showing the tests some love! Lgtm after a rebase

tests/integration-tests/package.json
```diff
@@ -2,18 +2,12 @@
...
     "host-exports",
     "non-fatal-errors",
     "overloaded-contract-functions",
     "poi-for-failed-subgraph",
     "remove-then-update",
-    "typename",
-    "value-roundtrip",
-    "dynamic-data-source",
-    "file-data-sources"
+    "value-roundtrip"
```
mangas, line 11: Why did these get removed?

mangas (approved):

PR #4278, opened 2023-01-03
Fix attestability of ValueParseError, EnumCoercionError, ScalarCoercionError

Theodus:
It seems that `ValueParseError` should be attestable. Feel free to quickly scan through to check for other miscategorizations.


leoyvens (approved):
`EnumCoercionError` and `ScalarCoercionError` look clearly deterministic.

leoyvens (approved):

PR #4303, opened 2023-01-19
docs, tests: update links to graph tooling repo and use latest graph-ts/cli versions

saihaj:


tests/integration-tests/data-source-revert/package.json
```diff
@@ -7,7 +7,7 @@
     "deploy:test-grafted": "graph deploy test/data-source-revert-grafted grafted.yaml --version-label v0.0.1 --ipfs $IPFS_URI --node $GRAPH_NODE_ADMIN_URI"
   },
   "devDependencies": {
-    "@graphprotocol/graph-cli": "https://github.com/graphprotocol/graph-cli#main",
-    "@graphprotocol/graph-ts": "https://github.com/graphprotocol/graph-ts#main"
+    "@graphprotocol/graph-cli": "latest",
+    "@graphprotocol/graph-ts": "latest"
```
saihaj, line 11: I think this is better and safer cause essentially we want to try new features

enisdenjo (approved):

dotansimha (approved):

tests/integration-tests/data-source-revert/package.json
dotansimha: Agree!

neysofu: I like the idea, but we must find another way to periodically upgrade our dependencies without `latest`. `yarn.lock` exists specifically to make builds as deterministic as possible, and an upstream change should never break our builds (case in point, CI fails now). I'm working on a PoC with a Dependabot-like workflow, so I'll post more information here. Until then, all of these version changes have to be reverted so the CI succeeds.

neysofu (changes requested):

tests/integration-tests/data-source-revert/package.json
neysofu: Alright, I think I got it. `latest` can stay. We just need to 1. modify `dependabot.yml` and 2. rebuild `yarn.lock`.
@saihaj you can cherrypick from my branch `filippo/yarn-lock-dependabot` if necessary (`https://github.com/graphprotocol/graph-node/commit/592aeae2f6b204b991fc2334161045ad761b583c` and `https://github.com/graphprotocol/graph-node/commit/b1d3633c9c87aedf100f060a99b1e8da7daf17b9`).

neysofu:
@saihaj I saw your message on Slack - I've tried to get this thing working, but no luck. It might require us to upgrade Yarn to v2 or v3 to leverage their new workspace commands...

If you're okay with leaving the `latest` thing out of this PR to revisit it down the road, I think that would be the best course of action.

saihaj:
> If you're okay with leaving the latest thing out of this PR to revisit it down the road, I think that would be the best course of action.

This issue is we have moved combined `graph-cli` and `graph-ts` to monorepo so this is just easier to do

neysofu:
Well, agreed, if we could get it to compile that would be best. Feel free to experiment with it and see if you can get it working.

saihaj:
> hey @neysofu and @saihaj, any update on this, is this blocked or are there changes required here?

need to spend sometime debugging this one but haven't gotten to it yet. Putting this in draft for now

PR #4318, opened 2023-01-24
store: Add a materialized view 'info.chain_sizes'

lutter:
This is just a convenience for admins to make it easier to understand how storage is used in the DB (similar to the already existing `info.subgraph_sizes` and `info.table_sizes`)


leoyvens (approved):

leoyvens:
Lets just merge https://github.com/graphprotocol/graph-node/pull/4162 first, so that the migrations are merged in the order that they are dated.

PR #4321, opened 2023-01-25
Add JSON logging to graph node

### LINKED ISSUE #3853: How to run graph-node with json logging

Feature (?)

Graph node currently does not do JSON logging and the [environment variables](https://github.com/graphprotocol/graph-node/blob/master/docs/environment-variables.md) don't seem to offer a configuration for that

No steps to reproduce, just run the graph node

I expect (?) that there's a configuration option for JSON logging

paymog:
Add JSON logging to graph and make it controllable by an env var, `JSON_LOGGING`.

Fixes #3853

### DIFF
--- a/graph/src/log/mod.rs
+++ b/graph/src/log/mod.rs
@@ -1,3 +1,14 @@
+use std::{fmt, io, result};
+
+use isatty;
+use slog::*;
+use slog_async;
+use slog_envlogger;
+use slog_json;
+use slog_term::*;
+
+use crate::prelude::ENV_VARS;
+
 #[macro_export]
 macro_rules! impl_slog_value {
     ($T:ty) => {
@@ -17,25 +28,24 @@ macro_rules! impl_slog_value {
     };
 }
 
-use isatty;
-use slog::*;
-use slog_async;
-use slog_envlogger;
-use slog_term::*;
-use std::{fmt, io, result};
-
-use crate::prelude::ENV_VARS;
-
 pub mod codes;
 pub mod elastic;
 pub mod factory;
 pub mod split;
 
-pub fn logger(show_debug: bool) -> Logger {
+pub fn logger(show_debug: bool, json: bool) -> Logger {
     let use_color = isatty::stdout_isatty();
     let decorator = slog_term::TermDecorator::new().build();
-    let drain = CustomFormat::new(decorator, use_color).fuse();
-    let drain = slog_envlogger::LogBuilder::new(drain)
+    let json_drain = slog_json::Json::new(std::io::stdout())
+        .add_default_keys()
+        .build();
+    let drain = CustomFormat::new(decorator, use_color);
+    let drain_picker = if json {
+        DrainPicker::JsonFormat(json_drain)
+    } else {
+        DrainPicker::CustomFormat(drain)
+    };
+    let drain = slog_envlogger::LogBuilder::new(drain_picker.fuse())
         .filter(
             None,
             if show_debug {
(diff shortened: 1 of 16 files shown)


neysofu:
Hi @paymog! Thanks for the PR.

I'm conflicted on this. On the one hand, I appreciate the fact that JSON logging facilitates integration with some log analytics platforms e.g. Datadog; on the other, this is a niche configuration setting that may result in fragmentation of log formats used in bug reports. Would you mind expanding a bit on your use case, please?

paymog:
> Hi @paymog! Thanks for the PR.
> 
> I'm conflicted on this. On the one hand, I appreciate the fact that JSON logging facilitates integration with some log analytics platforms e.g. Datadog; on the other, this is a niche configuration setting that may result in fragmentation of log formats used in bug reports. Would you mind expanding a bit on your use case, please?

Hey! Yes, this is precisely to make log ingestion/analysis through platforms like datadog easier. 

Do you mind expanding on why this would be a niche configuration? Do most folks who run graph-node not use log ingestion/analytics platforms? Is there specific data about how many (or few) graph node hosts don't use platforms like datadog for monitoring?

neysofu:
> Do you mind expanding on why this would be a niche configuration? Do most folks who run graph-node not use log ingestion/analytics platforms?

 I definitely expect most indexers to use log explorers, but most (all?) of these platforms allow you to parse log messages with custom rules, thus they accept arbitrary log formats. The set of users that *want* to use log analytics and explorer platforms but *can't* because of a lack of JSON logging is, I would imagine, very small.

paymog:
@neysofu ah yes, I totally agree! The set of users that want to use log analytics and explorer platforms but can't because of a lack of JSON logging is probably very small.

I've also set up a custom parser on datadog. However I find that this custom parser takes quite a bit of time to set up and even then isn't perfect (though that may just be because I haven't spent enough time on it) and is also something I may have to maintain over time if the logs change in some way.

As a follow up question, is it efficient as an ecosystem to have each indexer come up with the relevant parsing expression for their log platform of choice? That seems like quite a bit of duplicated effort for each indexer with not much benefit. 

I guess one alternative to this PR would be to keep track of the correct parsing expression for various platforms. That way when a new indexer decides to join, they can go to a well maintained list and more-or-less copy-paste a parsing expression into their preferred logging platform.

However, I still believe json logging will have far superior results because it means those parsing expressions don't need to be maintained and it means that platforms which don't have highly configurable parsers can still be used.

graph/src/log/mod.rs
```diff
@@ -17,25 +28,24 @@ macro_rules! impl_slog_value {
...
-
-use crate::prelude::ENV_VARS;
-
 pub mod codes;
 pub mod elastic;
 pub mod factory;
 pub mod split;
 
-pub fn logger(show_debug: bool) -> Logger {
+pub fn logger(show_debug: bool, json: bool) -> Logger {
```
mangas, line 36: I think this should take an Option<DrainPicker> if None, default to CustomFormat, if one passed then use that instead

mangas: you can add a `json_logger` function which does that setup and then calls `logger(debug, drain)`

node/src/opt.rs
```diff
@@ -163,6 +163,13 @@ pub struct Opt {
     pub expensive_queries_filename: String,
     #[clap(long, help = "Enable debug logging")]
     pub debug: bool,
+    #[clap(
+        long,
+        value_name = "JSON_LOGGING",
+        env = "JSON_LOGGING",
+        help = "output logs in json format"
+    )]
+    pub json_logging: bool,
```
mangas, line 172: This should also be a configurable option. Could have the values: `default` and `json` for now, others can be added later.

graph/src/log/mod.rs
mangas: could also setup an helper `default_logger` this way the other instances can just be replaced from `logger` to `default_logger` on tests and such

PR #4333, opened 2023-02-01
improve(metrics): Turn "Flushing logs to Elasticsearch" log into metric

leoyvens:
This log is very noisy, and can perfectly be substituted by a metric.


lutter (approved):
Nice!

PR #4340, opened 2023-02-06
store(test-store): use C locale for DB

saihaj:
while working on tests for #3214 I had to make this change. But useful to just land it upstream since it is unrelated to that PR


lutter (approved):
Nice find. Yes, that's definitely needed

Issue #3533, opened 2022-05-05
Remove use of web3 types outside of chain/ethereum


evaporei:
As seen in the [Arweave integration PR](https://github.com/graphprotocol/graph-node/pull/3465/files#diff-d1859cd422acc06d37e3622428bb2178c7a74150788f826e906352e8f13b8ad8R110) the `BlockPtr` hash is limited to 32 bytes / 256 bits (`H256` type).

This should be refactored so `graph-node` can deal with chains that have larger block hashes.

maoueh:
To add a bit more details, this is an implicit limit because underlying chain agnostic part of `graph-node` related to storage with Postgres uses H256 on some of their code path leading to panic/errors on chain with block hash with longer values.

For example [chain_head_ptr](https://github.com/streamingfast/graph-node/blob/b1dfbafcf7aa86b11725be6f703c43326b68f685/store/postgres/src/chain_store.rs#L1397) which does an implicit cast to `H256` which fails on Arweave.

Also [fn graft(](https://github.com/streamingfast/graph-node/blob/5ddc3e1994ec9599522ece67c478c8c3d8769d79/store/postgres/src/deployment.rs#L127)

There is a few other places like that. `hash_to_h256` should be removed also in my opinion.

lutter:
I think we should go even one step further: nothing outside of `chain/ethereum` should be using `H256`

azf20:
Is this fixed by https://github.com/graphprotocol/graph-node/pull/3623 cc @leoyvens ?

leoyvens:
That fixed the issues we've observed so far but it didn't systematically replace web3 types such as H256 and H160 with chain-agnostic ones.

PR #4348, opened 2023-02-07
chain store: replace more H256 by BlockHash

shyba:
This is my attempt at #3533 
I decided to open the PR at this point before going further to confirm if I understood the issue correctly. I can continue it after that or maybe in other PRs since this change is well isolated.

### DIFF
--- a/graphql/src/store/resolver.rs
+++ b/graphql/src/store/resolver.rs
@@ -233,7 +233,7 @@ impl StoreResolver {
                     if ptr.ptr.hash.is_zero() {
                         None
                     } else {
-                        Some(r::Value::String(format!("0x{:}", ptr.ptr.hash.hash_hex())))
+                        Some(r::Value::String(format!("0x{:x}", ptr.ptr.hash)))
                     }
                 })
                 .unwrap_or(r::Value::Null);


mangas (approved):

graphql/src/store/resolver.rs
```diff
@@ -230,11 +230,10 @@ impl StoreResolver {
                     // locate_block indicates that we do not have a block hash
                     // by setting the hash to `zero`
                     // See 7a7b9708-adb7-4fc2-acec-88680cb07ec1
-                    let hash_h256 = ptr.ptr.hash_as_h256();
-                    if hash_h256 == web3::types::H256::zero() {
+                    if ptr.ptr.hash.is_zero() {
                         None
                     } else {
-                        Some(r::Value::String(format!("0x{:x}", hash_h256)))
+                        Some(r::Value::String(format!("0x{:}", ptr.ptr.hash.hash_hex())))
```
maoueh, line 236: Probably that `BlockHash` should implement [`fmt::LowerHex`](https://doc.rust-lang.org/std/fmt/trait.LowerHex.html#)

shyba: Good catch. It already implements `LoweHex` but `hash_hex` was being used everywhere. I went ahead and replaced all usages with formatting in a separate commit. Let me know if that was too much, I can revert to just handling this case.

store/postgres/src/chain_store.rs
```diff
@@ -1700,7 +1700,7 @@ impl ChainStoreTrait for ChainStore {
             block_ptr.number >= offset,
             "block offset {} for block `{}` points to before genesis block",
             offset,
-            block_ptr.hash_hex()
+            format!("{:x}", block_ptr)
```
mangas, line 1703: I'd rather not repeat format everywhere, prolly better just adding a fn for this.

shyba: Reverted to just changing the line pointed out above and using the `fn` as it was. Thanks

shyba:
(Sorry, I didn't mean to remove lutter request for review and can't re-add. That happened automatically when I clicked re-request.)

graph/src/blockchain/types.rs
```diff
@@ -28,8 +28,8 @@ impl BlockHash {
         hex::encode(&self.0)
     }
 
-    pub fn zero() -> Self {
-        Self::from(H256::zero())
+    pub fn is_zero(&self) -> bool {
+        self.0.iter().all(|item| *item == 0)
```
mangas, line 32: https://docs.rs/primitive-types/0.12.1/primitive_types/struct.H256.html#method.is_zero I think this should work?

shyba: Sorry, I don't know if I get it. `self.0` is a `Box[u8]` and not a H256.
They implemented something [similar](https://github.com/paritytech/parity-common/blob/7a9cc3b03c5311c24623c66fe8a750903cf34484/fixed-hash/src/hash.rs#L206) for H256, but maybe there is an easier way.

mangas: ah no, you're right then, I assumed it was an H256. The implementation is similar for H256, I just thought it was an H256 so it would have been good to re-use

neysofu:
I'll close this as it hasn't seen activity in a while @shyba. We can always reopen if necessary :smiley:

PR #4349, opened 2023-02-07
Solve some clippy warnings

shyba:
To ease review I separated commits by `clippy` lint rules so it is easier to revert each, but we can squash before merge if necessary.

Special attention to `3729a474719d76caa7977a31ed7b34b5777d034d` as that required manual intervention. The others were automatic.

About #4169 - I suggest adding to CI the lints that everyone agrees that should be an error (which can be an empty set right now). Then we can fix a category and enable on CI incrementally.

### DIFF
--- a/graph/src/components/versions/registry.rs
+++ b/graph/src/components/versions/registry.rs
@@ -53,7 +53,7 @@ impl ApiVersion {
     fn resolve(version_requirement: &VersionReq) -> Result<&Version, String> {
         for version in VERSIONS.iter() {
             if version_requirement.matches(version) {
-                return Ok(version.clone());
+                return Ok(version);
             }
         }
 
--- a/graph/src/components/versions/registry.rs
+++ b/graph/src/components/versions/registry.rs
@@ -34,7 +34,7 @@ impl ApiVersion {
             version: version.clone(),
             features: VERSION_COLLECTION
                 .get(version)
-                .expect(format!("Version {:?} is not supported", version).as_str())
+                .unwrap_or_else(|| panic!("Version {:?} is not supported", version))
                 .to_vec(),
         })
     }
--- a/graph/src/components/versions/registry.rs
+++ b/graph/src/components/versions/registry.rs
@@ -28,12 +28,12 @@ pub struct ApiVersion {
 
 impl ApiVersion {
     pub fn new(version_requirement: &VersionReq) -> Result<Self, String> {
-        let version = Self::resolve(&version_requirement)?;
+        let version = Self::resolve(version_requirement)?;
 
         Ok(Self {
             version: version.clone(),
             features: VERSION_COLLECTION
-                .get(&version)
+                .get(version)
                 .expect(format!("Version {:?} is not supported", version).as_str())
                 .to_vec(),
         })
--- a/graph/src/components/versions/registry.rs
+++ b/graph/src/components/versions/registry.rs
@@ -35,7 +35,7 @@ impl ApiVersion {
             features: VERSION_COLLECTION
                 .get(version)
                 .unwrap_or_else(|| panic!("Version {:?} is not supported", version))
-                .to_vec(),
+                .clone(),
         })
     }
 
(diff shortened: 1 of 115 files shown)


shyba:
Something that caught my attention on [this commit](https://github.com/graphprotocol/graph-node/pull/4349/commits/20817dc2a4c3280afb832f7b3227001b0580fe5a):
Dropping the `&GasCounter` does nothing, so `clippy` seems right and I changed it, but then I was curious about the original intention. It is used as a function parameter down some lines, so it looks like gas is counted here. Should I remove the comment as well?

mangas:
Thanks a lot for this, it's very appreciated

graph/src/components/versions/registry.rs
```diff
@@ -28,14 +28,14 @@ pub struct ApiVersion {
...
+        let version = Self::resolve(version_requirement)?;
 
         Ok(Self {
             version: version.clone(),
             features: VERSION_COLLECTION
-                .get(&version)
-                .expect(format!("Version {:?} is not supported", version).as_str())
-                .to_vec(),
+                .get(version)
+                .unwrap_or_else(|| panic!("Version {:?} is not supported", version))
```
mangas, line 37: Is this a manual change or clippy? Not sure why this is better?

shyba: Automatic. It comes from [this commit](https://github.com/graphprotocol/graph-node/commit/271871081406a3f0ac245e4d9299c4433d7a5752) from the linter `clippy::expect_fun_call`.
As each rule has its own commit, we can drop the ones that are questionable.
[According to clippy](https://rust-lang.github.io/rust-clippy/stable/index.html#expect_fun_call), it was bad because `The function will always be called.`

neysofu:
Right after we merge this PR, I suggest we start enforcing Clippy in the CI (minus some annoying lints, possibly). Something like `cargo clippy -- -A clippy::len_zero -A clippy::match_like_matches_macro`.

neysofu:
By the way @shyba, how were you able to isolate fixes for specific lints so cleanly? Does Clippy have some feature like `--fix --but-only-for-this-lint`? Or did you apply fixes manually?

runtime/wasm/src/module/mod.rs
```diff
@@ -1677,8 +1677,7 @@ impl<C: Blockchain> WasmInstanceContext<C> {
         gas: &GasCounter,
         hash_ptr: AscPtr<AscString>,
     ) -> Result<AscPtr<AscString>, HostExportError> {
-        // Not enabled on the network, no gas consumed.
-        drop(gas);
+        // GasCounter is not enabled on the network, no gas consumed.
```
neysofu, line 1680: Thank you for drawing my attention to this change. `drop(gas)` should go away indeed, but I think the previous comment was better!

shyba:
> By the way @shyba, how were you able to isolate fixes for specific lints so cleanly? Does Clippy have some feature like --fix --but-only-for-this-lint? Or did you apply fixes manually?

Most were automatic. I used some bash commands.
`cargo clippy -- -W clippy::all 2>&1 | grep github | sed -e ' s/ //g' | sort | uniq -c | sort -h` shows the lint rules sorted by frequency.
`cargo clippy --fix --allow-dirty -- -A clippy::all -W  clippy::manual_str_repeat` will try to fix only the named lint.

I can imagine a bash script for that, but I was reviewing the changes while messing with `mold` and `sccache` so I went ahead to run those manually and compare times. Using `codegen-units=1024`,  `mold` and `sccache` building is way faster and takes between 600-2000mb of RAM instead of what it was taking before. I found that might be useful for development.

runtime/wasm/src/module/mod.rs
shyba: Indeed, will restore it!

shyba:
Rebased, fixed lint as well.

mangas:
LGTM, will leave approval to @neysofu 

@shyba if you do decide to add anything else please do so in a separate PR. Once again, thanks for submitting this PR 🙏

shyba:
@mangas Thanks!
Rebased, but had to change two commits during it. `clippy::map_clone`  and `clippy::redundant_closure`

mangas:
@neysofu @shyba we should probably enable the clippy checks on CI after this

neysofu:
Agreed. I'll review today, and after merging let's figure out a strategy for CI linting.

neysofu (approved):

PR #4350, opened 2023-02-08
Refactor Dockerfile to avoid unnecessary rebuilds and improve code readability.

arnav13081994:
Thank you so much for your amazing work on this repo.

I would also like to contribute in any way as I could and found some potential improvements with the Dockerfile.

This PR does the following:

1. Added `WORKDIR` directive to avoid needing to do `cd` and other manual directory changes.
2. Separated `OS dep layer` from building Rust package. This is done so that we don't need to rebuild rust package
in case of any OS updates.
3. Explicitly adding all Rust member files. This was done so that updates in files that have
nothing to do with the Rust build don't re-trigger unnecessary rebuilds. For example, even changing anything in the docker folder will cause a re-build.
4. Improve general readablity of the Dockerfile.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -40,14 +40,23 @@ ADD server ./server
 ADD store ./store
 ADD tests ./tests
 
+
+RUN RUSTFLAGS="-g" cargo build --release --package graph-node
+
+
+RUN cp target/release/graph-node /usr/local/bin/graph-node
+RUN cp target/release/graphman /usr/local/bin/graphman
+
     # Reduce the size of the layer by removing unnecessary files.
-    && cargo clean \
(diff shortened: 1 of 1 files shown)


neysofu:
Hi @arnav13081994, thank you for the PR! The `WORKDIR` addition is pretty nice, and I like it. I see a few problems with this PR though:

- `ADD`ing files necessary for the build automatically is not worth the effort. That list will, sooner or later, get out of sync and we'll have to debug failed builds. The savings are dubious at best.
- All these extra layers (which admittedly improve readability) consume space on disk. The size of the final image doesn't change, but this does have an impact on caching and storage used on development machines.

I don't think these changes improve the `Dockerfile` meaningfully. If you have some other ideas on how to improve the build process, let's absolutely discuss those - maybe in a separate issue. For now, I'll close this PR.

PR #4351, opened 2023-02-08
Add support for Web3Call adapter

mangas:
- Always prefer web3call adapter for runtime adapter on ethereum if available
- Prevent non call functions on call_only adapters
- Add web3call type configuration
- Require Web3Call endpoints to have the "archive" feature

### DIFF
--- a/chain/ethereum/src/network.rs
+++ b/chain/ethereum/src/network.rs
@@ -1,4 +1,4 @@
-use anyhow::{anyhow, Context};
+use anyhow::{anyhow, bail, Context};
 use graph::cheap_clone::CheapClone;
 use graph::prelude::rand::{self, seq::IteratorRandom};
 use std::cmp::Ordering;
@@ -23,12 +23,26 @@ pub struct EthereumNetworkAdapter {
     limit: usize,
 }
 
-#[derive(Clone)]
+impl EthereumNetworkAdapter {
+    fn is_call_only(&self) -> bool {
+        self.adapter.is_call_only()
+    }
+}
+
+#[derive(Clone, Default)]
 pub struct EthereumNetworkAdapters {
     pub adapters: Vec<EthereumNetworkAdapter>,
+    pub call_only_adapters: Vec<EthereumNetworkAdapter>,
 }
 
 impl EthereumNetworkAdapters {
+    pub fn push_adapter(&mut self, adapter: EthereumNetworkAdapter) {
+        if adapter.is_call_only() {
+            self.call_only_adapters.push(adapter);
+        } else {
+            self.adapters.push(adapter);
+        }
+    }
     pub fn all_cheapest_with(
         &self,
         required_capabilities: &NodeCapabilities,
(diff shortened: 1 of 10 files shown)


chain/ethereum/src/runtime/runtime_adapter.rs
```diff
@@ -43,13 +43,17 @@ impl blockchain::RuntimeAdapter<Chain> for RuntimeAdapter {
...
-                archive: ds.mapping.requires_archive()?,
-                traces: false,
-            })?
-            .cheap_clone();
+        let eth_adapter = match self.eth_adapters.call_only_adapter() {
+            Err(err) => bail!(err),
+            Ok(None) => self
+                .eth_adapters
+                .cheapest_with(&NodeCapabilities {
+                    archive: ds.mapping.requires_archive()?,
```
lutter, line 51: If we are making eth calls, don't we always need an archive node? I am not sure, but my assumption is that this is always `true`

chain/ethereum/src/runtime/runtime_adapter.rs
```diff
@@ -43,13 +43,17 @@ impl blockchain::RuntimeAdapter<Chain> for RuntimeAdapter {
...
+                    traces: false,
+                })?
+                .cheap_clone(),
```
lutter, line 54: It would be better to have this logic (first look for call only adapters, then look in the general adapters) encapsulated in a method on `eth_adapters`, maybe `eth_adapters.cheapest_for_call(..)` since this whole `match` statement has a lot of knowledge about how `EthereumNetworkAdapters` sets itself up internally which would be better to hide away

chain/ethereum/src/network.rs
```diff
@@ -23,12 +23,26 @@ pub struct EthereumNetworkAdapter {
...
+    fn is_call_only(&self) -> bool {
+        self.adapter.is_call_only()
+    }
+}
+
+#[derive(Clone, Default)]
 pub struct EthereumNetworkAdapters {
     pub adapters: Vec<EthereumNetworkAdapter>,
+    pub call_only_adapters: Vec<EthereumNetworkAdapter>,
 }
```
lutter, line 36: Would it make sense to have a separate `EthereumCallAdapter` that can only be used to make calls? The way things are right now, you could get a call-only adapter and then use all kinds of different methods on it even though they would blow up. It would be safer to split that out from `EthereumNetworkAdapter`, but I am not sure how much work that will be.

chain/ethereum/src/network.rs
```diff
@@ -73,6 +87,42 @@ impl EthereumNetworkAdapters {
...
+
+        let adapters = self
+            .call_only_adapters
+            .iter()
+            .min_by_key(|x| Arc::strong_count(&x.adapter))
+            .ok_or(anyhow!("no available call only endpoints"))?;
+
+        // TODO: This will probably blow up a lot sooner than [limit] amount of
+        // subgraphs, since we probably use a few instances.
+        if Arc::strong_count(&adapters.adapter) >= adapters.limit {
```
lutter, line 118: Yeah, I used the `Arc::strong_count` mostly as a hack; great if this works, but we might run into the limits of this approach.

lutter (approved):
I like the idea of making this a different type of provider in the config; call-only really doesn't feel like a feature but a different type of provider.

chain/ethereum/src/network.rs
mangas: I have added some assertions, I thought about this but this would require a significant refactor because we would need to make these functions 2 traits and implement them for different types (normal and call-only) and then change everything that uses these. I do think we should do it, but probably afterwards so that we can start testing ASAP. I think I will add this as a separate issue.

chain/ethereum/src/network.rs
mangas: I think this is another part we probably should change properly, this method was already being used, I don't think it's very good for this part of the code because we have a lot of different copies of the underlying adapters for a single subgraph.

chain/ethereum/src/runtime/runtime_adapter.rs
mangas: This part was already here so I'm not sure if it;s different for other clients that are not ethereum. In the new code I made that assumption

chain/ethereum/src/runtime/runtime_adapter.rs
mangas: good catch, I had added the function later, forgot to replace this part :)

PR #4353, opened 2023-02-09
Fix limit handling on rpc configuration

mangas:
- Set 0 for Disabled
- Omitted hosts can are disabled if list of rules is not empty
- Empty rules = Unlimited
- Limit != 0 does exactly what you'd think it does

### DIFF
--- a/chain/ethereum/src/network.rs
+++ b/chain/ethereum/src/network.rs
@@ -1,5 +1,6 @@
 use anyhow::{anyhow, bail, Context};
 use graph::cheap_clone::CheapClone;
+use graph::firehose::SubgraphLimit;
 use graph::prelude::rand::{self, seq::IteratorRandom};
 use std::cmp::Ordering;
 use std::collections::HashMap;
@@ -20,7 +21,7 @@ pub struct EthereumNetworkAdapter {
     /// strong_count on `adapter` to determine whether the adapter is above
     /// that limit. That's a somewhat imprecise but convenient way to
     /// determine the number of connections
-    limit: usize,
+    limit: SubgraphLimit,
 }
 
 impl EthereumNetworkAdapter {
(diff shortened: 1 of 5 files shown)


graph/src/firehose/endpoints.rs
```diff
@@ -34,15 +34,26 @@ pub struct FirehoseEndpoint {
...
-#[derive(Clone, Debug)]
+// TODO: Find a new home for this type.
+#[derive(Clone, Debug, PartialEq, Ord, Eq, PartialOrd)]
 pub enum SubgraphLimit {
-    Unlimited,
+    Disabled,
     Limit(usize),
-    NoTraffic,
+    Unlimited,
+}
```
lutter, line 47: Do we really need this enum? I don't really think it buys us much above using a plain `usize` (where `Disabled -> 0` and `Unlimited -> usize::MAX`)

lutter (approved):
LGTM. I am not sure that the `SubgraphLimit` enum is really needed, but the behavior is now defintiely what we want (based on the unit tests, nice!)

graph/src/firehose/endpoints.rs
mangas: It's mostly to improve readability since otherwise we'd need to set a value for 0 and some for "not set" in order to differentiate so there would be a few places with Option<usize> where we were not sure of the meaning. Also Unlimited will also get transformed in the different adapters, eg firehose uses 100 max so it gets set to the adapter-specific "maximum"

chain/ethereum/src/network.rs
```diff
@@ -97,9 +97,12 @@ impl EthereumNetworkAdapters {
         &self,
         capabilities: Option<&NodeCapabilities>,
     ) -> anyhow::Result<Arc<EthereumAdapter>> {
-        match self.call_only_adapter()? {
-            Some(adapter) => Ok(adapter),
-            None => self.cheapest_with(capabilities.unwrap_or(&NodeCapabilities {
+        // call_only_adapter can fail if we're out of capcity or,this is fine since
+        // we would want to fallback onto a full adapter
+        // so we will ignore this error and return whatever comes out of `cheapest_with`
```
lutter, line 102: I don't fully understand this comment, especially the 'or, this is fine' throws me off

lutter (approved):
LGTM.

The file [`docs/config.md`](https://github.com/graphprotocol/graph-node/blob/master/docs/config.md#controlling-the-number-of-subgraphs-using-a-provider) also needs an update: the sentence
```
Any node whose name does not match one of these patterns will use mainnet-0 and mainnet-1 for an unlimited number of subgraphs.
```
needs to be changed to say that nodes whose names do not match will not use that provider

chain/ethereum/src/network.rs
mangas: I initially had something else there :P I will fix this, good catch :)

PR #4374, opened 2023-02-16
Implement moving of subgraphs

### LINKED ISSUE #4371: Add a way to move all subgraphs out of a shard
There are situations where it is desirable to move all subgraphs from one shard into another shard (e.g., to consolidate shards) It's currently possible to do that, but pretty laborious: for each subgraph in the shard that should be emptied, it is necessary to run the following commands manually:

```
graphman copy create sgd<src> <dst shard> <index node>
# wait until copying has finished and the copy is synced
graphman copy activate <IPFS hash> <dst shard>
graphman unassign sgd<src>
```

It would be much better if `graph-node` could perform these steps without much user interaction. We should change `graphman copy create` to accept two new flags:
* `--activate`: activate the copy when it has caught up with the chain head. For that, `SyncStore::deployment_synced` will call `primary::Connection::activate`
* `--replace`: when the copy has synced, do what `--activate` does but also unassign the source of the copy so that it will eventually be deleted

With that, a shard can be emptied by running `graphman copy create --replace sgd<src> <dst shard> <index node>` for each subgraph in that shard. It should be possible to issue those commands all at once, `graph-node` should work through that list on its own and appropriately limit load in case there is a huge number of deployments to move, though that needs to be tried out and tested carefully

lutter:
Fixes #4371 (which also describes how this is intended to be used)

### DIFF
--- a/docs/implementation/metadata.md
+++ b/docs/implementation/metadata.md
@@ -105,6 +105,7 @@ shard alongside the deployment's data in `sgdNNN`.
 | `use_bytea_prefix`      | `boolean!` |                                                      |
 | `start_block_hash`      | `bytea`    | Parent of the smallest start block from the manifest |
 | `start_block_number`    | `int4`     |                                                      |
+| `on_sync`               | `text`     | Additional behavior when deployment becomes synced   |
 
 ### `subgraph_deployment_assignment`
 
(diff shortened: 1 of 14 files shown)


neysofu (approved):
LGTM!

PR #4376, opened 2023-02-17
tests: Make running createdb more tolerant to slow Postgres startup

lutter:


fordN (approved):
LGTM 👍

tests/src/docker_utils.rs
```diff
@@ -218,34 +218,46 @@ impl ServiceContainer {
...
+            let mut stream = docker.client.start_exec(&message.id, None);
+            while let Some(_) = stream.next().await { /* consume stream */ }
+
+            // 3. Inspect exec
+            let inspect = docker.client.inspect_exec(&message.id).await?;
+            match inspect.exit_code {
+                Some(0) => return Ok(()),
+                code if attempt < EXEC_TRIES => {
+                    println!(
+                        "failed to run 'createdb' (exit code: {:?}) Will try again",
```
fordN, line 249: Include `attempt` and `EXEC_TRIES` values in the log for context?

tests/src/docker_utils.rs
```diff
@@ -218,34 +218,46 @@ impl ServiceContainer {
...
+                        code
+                    );
+                    std::thread::sleep(Duration::from_secs(5));
```
mangas, line 252: Isn't this going to crash since it's an async func?

lutter: It won't crash, but it will block the executor; switching to `tokio::time::sleep`

tests/src/docker_utils.rs
lutter: Good idea, changed

PR #4389, opened 2023-02-22
Enforce Clippy in the CI

neysofu:
Now that we've merged https://github.com/graphprotocol/graph-node/pull/4349, we can start enforcing Clippy on the CI. The defaults look like a good choice for now.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,2 +1,3 @@
 [alias]
+# Warnings create a lot of noise, we only print errors.
 check-clippy = "clippy --no-deps -- --allow warnings"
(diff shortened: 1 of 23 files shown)


neysofu:
On second thought, many of the warnings are valuable lints (and we still get many of those), so we shouldn't ignore those.

mangas:
as discussed I think we should do away with check-clippy and instead just rely on defaults for now, meaning we could also just run cargo clippy on ci

neysofu:
We can't enable defaults for now; CI will fail. I'll merge this and then we can increase the amount of lints until we can use defaults.

Issue #4394, opened 2023-02-23
JSON-RPC call `reassign_subgraph` in graph-node fails if a deployment has been copied


lutter:
When a deployment has been copied, the IPFS hash does not identify the deployment uniquely any more, and there will be two (or more) deployments with the same hash. `graph-node`'s JSON-RPC `reassign_subgraph` request fails because a lookup of that IPFS hash returns multiple deployments.

The call handler should be changed so that `reassign_subgraph` always changes the active copy if there are multiple copies of a deployment.

On the network, this issue causes the indexer-agent to get stuck since it tries to reassign subgraphs

PR #4395, opened 2023-02-23
all: Indicate whether to only locate active deployment

lutter:
The reassign_subgraph JSON-RPC call will fail if a deployment exists in multiple shards. To disambiguate that, change the call so that it only affects the active deployment.

Fixes https://github.com/graphprotocol/graph-node/issues/4394

### DIFF
--- a/core/src/subgraph/registrar.rs
+++ b/core/src/subgraph/registrar.rs
@@ -411,21 +411,10 @@ where
         hash: &DeploymentHash,
         node_id: &NodeId,
     ) -> Result<(), SubgraphRegistrarError> {
-        let locations = self.store.locators(hash)?;
-        let deployment = match locations.len() {
-            0 => return Err(SubgraphRegistrarError::DeploymentNotFound(hash.to_string())),
-            1 => locations[0].clone(),
-            _ => {
-                return Err(SubgraphRegistrarError::StoreError(
-                    anyhow!(
-                        "there are {} different deployments with id {}",
-                        locations.len(),
-                        hash.as_str()
-                    )
-                    .into(),
-                ))
-            }
-        };
+        let locator = self.store.active_locator(hash)?;
+        let deployment =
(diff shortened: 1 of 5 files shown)


core/src/subgraph/registrar.rs
```diff
@@ -411,7 +411,7 @@ where
         hash: &DeploymentHash,
         node_id: &NodeId,
     ) -> Result<(), SubgraphRegistrarError> {
-        let locations = self.store.locators(hash)?;
+        let locations = self.store.locators(hash, true)?;
```
tilacog, line 414: We expect this to return a vector with a single `DeploymentLocator` (the active copy), right?

tilacog (approved):

graph/src/components/store/traits.rs
```diff
@@ -156,8 +156,12 @@ pub trait SubgraphStore: Send + Sync + 'static {
 
     async fn is_healthy(&self, id: &DeploymentHash) -> Result<bool, StoreError>;
 
-    /// Find the deployment locators for the subgraph with the given hash
-    fn locators(&self, hash: &str) -> Result<Vec<DeploymentLocator>, StoreError>;
+    /// Find the deployment locators for the subgraph with the given hash.
+    /// If `only_active` is `true`, only the locator for the active
+    /// deployment will be returned. If it is `false`, locators for all
+    /// deployments with this hash will be returned.
+    fn locators(&self, hash: &str, only_active: bool)
```
mangas, line 163: I think we should split this into something like `all_locators` and `active_locators` instead of using a flag, when someone is using or reading this function with a bool they need to go read the docs to figure what the bool is used for, when using all/active naming then it's obvious both when writing and when reading the code.

lutter: I just removed the boolean flag and added a `fn active_locator` to the trait

mangas (approved):

PR #4396, opened 2023-02-23
store: Tolerate a missing source for a copy

lutter:
When we check whether we should perform on_sync actions, tolerate the source having been deleted.


mangas (approved):

mangas:
Would be great to ensure this is covered by a test. If possible

lutter:
> Would be great to ensure this is covered by a test. If possible

I just added a test, and even made sure it fails on `master` :)

PR #4397, opened 2023-02-24
all: Make `graphman chain call-cache remove` safer

lutter:
The command defaulted to the dangerous operation of clearing the entie cache. Now, the entire cache can only be cleared by passing the annoyingly named `--remove-netire-cache`.

In a similar vein, it used to be ok to omit --from or --to which would default to the first or last block. They now need to both be specified to make the user be clearer about how much of the call cache they are going to remove.


tilacog (approved):
:+1:

PR #4398, opened 2023-02-24
Improve running subgraph metric accuracy

mangas:
- Always modify alive instances alongside metrics

Ran locally with a forced error, the count is now correctly decreased while waiting for the retry sleep

### DIFF
--- a/core/src/subgraph/instance_manager.rs
+++ b/core/src/subgraph/instance_manager.rs
@@ -1,5 +1,5 @@
 use crate::polling_monitor::IpfsService;
-use crate::subgraph::context::{IndexingContext, SharedInstanceKeepAliveMap};
+use crate::subgraph::context::{IndexingContext, SubgraphKeepAlive};
 use crate::subgraph::inputs::IndexingInputs;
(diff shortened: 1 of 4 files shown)


core/src/subgraph/instance_manager.rs
```diff
@@ -179,11 +173,10 @@ impl<S: SubgraphStore> SubgraphInstanceManager<S> {
             logger_factory,
             subgraph_store,
             chains,
-            manager_metrics: Arc::new(SubgraphInstanceManagerMetrics::new(
-                metrics_registry.cheap_clone(),
+            metrics_registry: metrics_registry.cheap_clone(),
+            instances: SharedInstanceKeepAliveMap::new(Arc::new(
+                SubgraphInstanceManagerMetrics::new(metrics_registry),
```
lutter, line 178: This would be easier to read if `SharedInstanceKeepAliveMap::new` took just the `metrics_registry` as an argument and set up the `SubgraphInstanceManagerMetrics` itself. (And feel free to shorten some of those names)

lutter (approved):

Issue #4237, opened 2022-12-08
Feature idea: `graphman rewind --start-block`


neysofu:
To rewind a subgraph to its start block, without having to manually read the start block from its manifest.

PR #4400, opened 2023-02-26
node: Add start block flag to graphman rewind

incrypto32:
Fixes https://github.com/graphprotocol/graph-node/issues/4237

### DIFF
--- a/node/src/manager/commands/rewind.rs
+++ b/node/src/manager/commands/rewind.rs
@@ -128,20 +128,27 @@ pub async fn run(
     println!("\nRewinding deployments");
     for deployment in &deployments {
         let loc = deployment.locator();
+        let block_store = store.block_store();
         let deployment_details = subgraph_store.load_deployment_by_id(loc.clone().into())?;
-
         let block_ptr_to = block_ptr_to.clone();
 
-        if let Some(block_ptr) = block_ptr_to {
-            subgraph_store.rewind(loc.hash.clone(), block_ptr.clone())?;
-            println!("  ... rewound {}", loc);
-        } else {
-            if let Some(block_ptr) = deployment_details.start_block {
-                subgraph_store.truncate(loc.hash.clone(), block_ptr.clone())?;
+        let start_block = deployment_details.start_block.or_else(|| {
+            block_store
+                .chain_store(&deployment.chain)
+                .and_then(|chain_store| chain_store.genesis_block_ptr().ok())
+        });
+
+        match (block_ptr_to, start_block) {
+            (Some(block_ptr), _) => {
+                subgraph_store.rewind(loc.hash.clone(), block_ptr)?;
+                println!("  ... rewound {}", loc);
+            }
+            (None, Some(start_block_ptr)) => {
+                subgraph_store.truncate(loc.hash.clone(), start_block_ptr)?;
                 println!("  ... truncated {}", loc);
-            } else {
-                // TODO: Remove the if block and truncate even if there is no startBlock set
-                println!("  ... no block pointer to rewind to for {}", loc);
+            }
+            (None, None) => {
+                println!("  ... Failed to find start block for {}", loc);
             }
         }
     }
--- a/node/src/manager/commands/rewind.rs
+++ b/node/src/manager/commands/rewind.rs
@@ -59,12 +59,18 @@ pub async fn run(
     primary: ConnectionPool,
     store: Arc<Store>,
     searches: Vec<DeploymentSearch>,
-    block_hash: String,
-    block_number: BlockNumber,
+    block_hash: Option<String>,
+    block_number: Option<BlockNumber>,
     force: bool,
     sleep: Duration,
+    start_block: bool,
 ) -> Result<(), anyhow::Error> {
     const PAUSED: &str = "paused_";
+
+    // Sanity check
+    if !start_block && (block_hash.is_none() || block_number.is_none()) {
+        bail!("--block-hash and --block-number must be specified when --start-block is not set");
+    }
 
     let subgraph_store = store.subgraph_store();
     let block_store = store.block_store();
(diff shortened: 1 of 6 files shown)


node/src/manager/commands/rewind.rs
```diff
@@ -116,8 +128,19 @@ pub async fn run(
...
+        let deployment_details = subgraph_store.load_deployment_by_id(loc.clone().into())?;
+
+        let block_ptr_to = block_ptr_to.clone().or(deployment_details.start_block);
+
+        if let Some(block_ptr) = block_ptr_to {
+            subgraph_store.rewind(loc.hash.clone(), block_ptr.clone())?;
+            println!("  ... rewound {}", loc);
+        } else {
+            println!(
+                "  ... skipping {} as it doesn't have a startBlock specified in manifest",
```
neysofu, line 140: Instead of skipping deployments for which there's no start block, I think I'd rather first cycle through all deployments and make sure all manifests have a start block. If not, we abort the entire operation.

neysofu (changes requested):
The implementation looks good, I think we should abort the operation if some manifest doesn't have a start block though.

Also, can you please make a note in `NEWS.md` under `Unreleased` that `block-hash` and `block-number` are now flags instead of arguments? That's a breaking change and we must put it in the notes.

lutter:
When rewinding to the start block which basically forces a complete reindexing of the subgraph (and maybe the flag should be called `reindex`) we should not use `delete from` to get rid of the data and instead truncate the data tables with `truncate`. Truncating tables is almost instant whereas deleting everything from a table can take a very long time.

That means the store should somehow become aware of the fact that we are going to throw away all the data. That could either be passed into `SubgraphStore.rewind` or `SubgraphStore.rewind` could check itself whether we will throw away everything and determine the best strategy itself.

incrypto32:
@lutter , thanks for this, what do you think would be the better route? is it making the `rewind_with_conn` in the [deployment_store](https://github.com/graphprotocol/graph-node/blob/master/store/postgres/src/deployment_store.rs#L1221) somehow check if the block is the earliest block available and truncate instead of `revert_block` if thats the case 

Or do you think passing something like `truncate` / `reindex` boolean down from `SubgraphStore.rewind` is better?

store/postgres/src/relational.rs
```diff
@@ -863,6 +863,13 @@ impl Layout {
         Ok(count)
     }
 
+    pub fn truncate_tables(&self, conn: &PgConnection) -> Result<(StoreEvent, i32), StoreError> {
+        for table in self.tables.values() {
+            conn.execute(&format!("TRUNCATE TABLE {}", table.qualified_name))?;
+        }
+        Ok((StoreEvent::new(vec![]), -1))
```
lutter, line 870: Since we truncate, we know that there will be 0 entities left, so there's no point in returning a count.

node/src/manager/commands/rewind.rs
```diff
@@ -116,8 +128,22 @@ pub async fn run(
...
-        println!("  ... rewound {}", loc);
+        let deployment_details = subgraph_store.load_deployment_by_id(loc.clone().into())?;
+
+        let block_ptr_to = block_ptr_to.clone();
+
+        if let Some(block_ptr) = block_ptr_to {
+            subgraph_store.rewind(loc.hash.clone(), block_ptr.clone())?;
+            println!("  ... rewound {}", loc);
+        } else {
+            if let Some(block_ptr) = deployment_details.start_block {
```
lutter, line 139: If the deployment doesn't have a start block, shouldn't we use the genesis block as the point to which we truncate?

store/postgres/src/deployment_store.rs
```diff
@@ -1362,14 +1370,63 @@ impl DeploymentStore {
...
+    ) -> Result<StoreEvent, StoreError> {
+        self.rewind_or_truncate_with_conn(conn, site, block_ptr_to, firehose_cursor, false)
+    }
+
+    fn truncate_with_conn(
+        &self,
+        conn: &PgConnection,
+        site: Arc<Site>,
+        block_ptr_to: BlockPtr,
+        firehose_cursor: &FirehoseCursor,
```
lutter, line 1401: Since the firehose cursor will always be set to `None`, it shouldn't be passed in here; this function can just always use `None`

store/postgres/src/deployment_store.rs
```diff
@@ -1362,14 +1370,63 @@ impl DeploymentStore {
...
+                }
+
                 Ok(event)
             })
         })?;
 
         Ok(event)
     }
 
+    fn rewind_with_conn(
```
lutter, line 1386: Since `rewind_with_conn` and `truncate_with_conn` are so simple, they could be inlined into their callers.

store/postgres/src/deployment_store.rs
```diff
@@ -1362,14 +1370,63 @@ impl DeploymentStore {
...
+    ) -> Result<StoreEvent, StoreError> {
+        let conn = self.get_conn()?;
+
+        // Unwrap: If we are reverting then the block ptr is not `None`.
+        let block_ptr_from = Self::block_ptr_with_conn(&conn, site.cheap_clone())?.unwrap();
+
+        // Sanity check on block numbers
+        if block_ptr_from.number <= block_ptr_to.number {
+            constraint_violation!(
+                "rewind must go backwards, but would go from block {} to block {}",
```
lutter, line 1419: The error should say `truncate`

store/postgres/src/deployment_store.rs
```diff
@@ -1353,7 +1354,14 @@ impl DeploymentStore {
...
 
-                let (event, count) = layout.revert_block(conn, block)?;
+                let event: StoreEvent;
+                let count: i32;
+
+                if truncate {
+                    (event, count) = layout.truncate_tables(conn)?;
+                } else {
+                    (event, count) = layout.revert_block(conn, block)?;
+                }
```
lutter, line 1364: You can write this more idiomatically as
```rust
let (event, count) = if truncate { .. }
```

Actually, I would pull the other block from line 1373 into this, too, and write this as
```rust
let event = if truncate {
                      let event = layout.truncate_tables(conn)?;
                      deployment::set_entity_count(conn, site.as_ref(), layout.count_query.as_str())?;
                      event
                   } else {
                      let (event, count) = layout.revert_block(conn, block)?;
                      deployment::update_entity_count(conn, site.as_ref(), count)?;
                      event
                  };
```

That avoids passing around a count that means very different things for rewinding vs truncate

lutter (changes requested):
Looks good in general - I left some minor comments

store/postgres/src/deployment_store.rs
incrypto32: I did so since these both give two very different results and thought it'd make sense to not use `rewind_or_truncate_with_conn` that has a boolean flag `truncate` everywhere but instead separate it to two functions. 

Will inline it and use `rewind_or_truncate_with_conn` instead.

lutter (approved):
Nice!

neysofu (approved):

PR #4401, opened 2023-02-27
Improve subgraph running count metrics

mangas:
- deployment_count now shows the absolute value of deployed subgraphs for the node
- deployment_running_count now shows the count of running singraphs which decreases when a subgraph is retrying


core/src/subgraph/context.rs
```diff
@@ -64,7 +58,7 @@ where
     C: Blockchain,
 {
     instance: SubgraphInstance<C, T>,
-    pub instances: SubgraphKeepAlive,
+    pub instances: Arc<SubgraphKeepAlive>,
```
neysofu, line 61: What's the rationale for removing `CheapClone` from `SubgraphKeepAlive` and wrapping it in `Arc`? Is it because now it has two `Arc` fields?

neysofu (approved):

core/src/subgraph/context.rs
mangas: oh it's because cheap_clone is already implemented for Arc and I've now wrapped with Arc. I don't think it's actually copied so I could probably leave it without the Arc anyway

neysofu: I think wrapping everything with `Arc` hinders readability, so I try to avoid it unless some fields are quite expensive to clone. We're good though, feel free to merge.

mangas: I think it's a good point about readability, I've reverted that change.

neysofu: 💯

PR #4408, opened 2023-03-01
`BlockchainBuilder` abstraction to simplify instantiating chains

neysofu:
NEAR, Cosmos and Arweave all share the same `Chain::new` type signature. By putting that method into a trait, we can reduce code duplication in the `_networks_as_chains` family of functions in `main.rs`.

### DIFF
--- a/chain/arweave/src/chain.rs
+++ b/chain/arweave/src/chain.rs
@@ -1,8 +1,10 @@
 use graph::blockchain::client::ChainClient;
-use graph::blockchain::{Block, BlockchainKind, BuildableBlockchain, EmptyNodeCapabilities};
+use graph::blockchain::{
+    BasicBlockchainBuilder, Block, BlockchainBuilder, BlockchainKind, EmptyNodeCapabilities,
+};
 use graph::cheap_clone::CheapClone;
 use graph::data::subgraph::UnifiedMappingApiVersion;
-use graph::firehose::{FirehoseEndpoint, FirehoseEndpoints};
+use graph::firehose::FirehoseEndpoint;
 use graph::prelude::MetricsRegistry;
 use graph::{
     blockchain::{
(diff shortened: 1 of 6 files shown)


mangas (approved):
Cool, I'm going to try and do something similar to block ingestors and block streams since all they need now is ChainClient :)

neysofu:
@mangas I went a step further after your review; I completely removed `Chain::new` and rather instantiate directly in `Chain::build`. WDYT? See https://github.com/graphprotocol/graph-node/pull/4408/commits/8d29da588225419af549d0131c5a5ee075456cad.

chain/arweave/src/chain.rs
```diff
@@ -44,20 +44,20 @@ impl std::fmt::Debug for Chain {
     }
 }
 
-impl Chain {
-    pub fn new(
+impl BuildableBlockchain for Chain {
+    fn build(
         logger_factory: LoggerFactory,
-        name: String,
+        chain_id: String,
```
mangas, line 50: This is the network_name, there is also a chain_id whihc is something internal to each chain, I think we should keep the previous naming

chain/arweave/src/chain.rs
```diff
@@ -44,20 +44,20 @@ impl std::fmt::Debug for Chain {
...
         chain_store: Arc<dyn ChainStore>,
-        firehose_endpoints: FirehoseEndpoints,
-        metrics_registry: Arc<dyn MetricsRegistry>,
+        fh_endpoints: FirehoseEndpoints,
```
mangas, line 52: please leave this one as was, I have a PR that will replace FIrehoseEndpoints with ChainClient in most places

graph/src/blockchain/mod.rs
```diff
@@ -135,6 +139,19 @@ impl ChainStoreBlock {
 //     Rpc(C::Client),
 // }
 
+/// A chain-agnostic interface for instantiating chains of a specific kind. Not
+/// all chains may support this interface, but that's okay, as some may require
+/// special logic.
+pub trait BuildableBlockchain: Blockchain {
```
mangas, line 145: This comment is totally subjective so I will leave it just for your consideration.

I think I would rather have a the Blockchain as a type parameter. 

Also have a suggestion for the name 

```
pub trait BlockchainBuilder<C: Blockchain>
```

This means we could have multiple implementations of Builders if needed or a single one that's generic over Blockchain as opposed to having only 1 per Blockchain implementation. 

Other than that I don't have a huge preference and may also be missing something here.

mangas (commented):
Left a few more comments but in general LGTM

neysofu:
I like the `BlockchainBuilder` pattern much better, thank you for the suggestion!

PR #4411, opened 2023-03-02
Further refactor adapter handling, BlockStream, BlockIngestor

mangas:
1. BlockIngestor is now a trait and each Blockchain will built it’s own
2. BlockStream is now also built via a single function on the blockchain trait instead of leaking the firehose/non firehose decision
3. The adapter is now not sticky on the FirehoseBlockStream ie it will retry with different adapters
4. The FirehoseBlockStream now has the provider and subgraph id on every message
5. Make FirehoseBlockIngestor adapter not sticky

Closes:
https://github.com/graphprotocol/graph-node/issues/4415
https://github.com/graphprotocol/graph-node/issues/4416


graph/src/blockchain/mod.rs
```diff
@@ -208,6 +205,8 @@ pub trait Blockchain: Debug + Sized + Send + Sync + Unpin + 'static {
     fn runtime_adapter(&self) -> Arc<dyn RuntimeAdapter<Self>>;
 
     fn chain_client(&self) -> Arc<ChainClient<Self>>;
+
+    fn block_ingestor(&self) -> anyhow::Result<Box<dyn BlockIngestor>>;
```
neysofu, line 209: Arweave, NEAR, and Cosmos pretty much have identical implementations. Unfortunately `Self::Block: prost::Message` is not guaranteed so we can't have a default implementation, among other reasons, but a helper function might be nice.

node/src/main.rs
```diff
@@ -841,141 +813,21 @@ fn near_networks_as_chains(
...
+                )),
             )
         })
         .collect();
 
-    for (chain_id, firehose_chain) in chains.iter() {
-        blockchain_map
-            .insert::<graph_chain_near::Chain>(chain_id.clone(), firehose_chain.chain.clone())
+    for (chain_id, chain) in chains.iter() {
+        blockchain_map.insert::<graph_chain_near::Chain>(chain_id.clone(), chain.clone())
```
neysofu, line 829: I hope these changes don't conflict too much with #4408, but we'll see. Let's merge this first as it's bigger.

node/src/opt.rs
```diff
@@ -186,14 +186,6 @@ pub struct Opt {
...
     )]
     pub elasticsearch_password: Option<String>,
-    #[clap(
-        long,
-        value_name = "MILLISECONDS",
-        default_value = "1000",
-        env = "ETHEREUM_POLLING_INTERVAL",
-        help = "How often to poll the Ethereum node for new blocks"
-    )]
-    pub ethereum_polling_interval: u64,
```
neysofu, line 196: Why did you remove this from `Opt` and put it in `EnvVars`?

neysofu: It's not bad, just curious.

neysofu (changes requested):

graph/src/blockchain/mod.rs
```diff
@@ -51,6 +51,12 @@ use self::{
     client::ChainClient,
 };
 
+#[async_trait]
+pub trait BlockIngestor: 'static + Send + Sync {
+    async fn run(self: Box<Self>);
+    fn network_name(&self) -> String;
```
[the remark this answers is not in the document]
mangas: This change will create issues with the reference since it's not 'static and it's an async context, will end up forcing to_string() being called elsewhere, prolly not worth the optimisation.

graph/src/blockchain/mod.rs
mangas: True, I think this is prolly a good further improvement for a different PR though

mangas: https://github.com/graphprotocol/graph-node/issues/4418

node/src/main.rs
mangas: I'm happy to refactor the other PR I think it's not too bad but they touch some of the same things

node/src/opt.rs
mangas: because I didn't want to add a dependency on the node package from the ethereum package and I don't think this would be used as a flag since it's eth specific.

mangas: we could leave it in both places but if used as an argument AND var we would have to make this more complicated than it should be 😛

chain/ethereum/src/chain.rs
```diff
@@ -101,14 +104,66 @@ impl BlockStreamBuilder<Chain> for EthereumStreamBuilder {
...
+                    "no adapter for network {} with capabilities {}",
+                    chain.name, requirements
+                )
+            });
+
+        let logger = chain
+            .logger_factory
+            .subgraph_logger(&deployment)
+            .new(o!("component" => "BlockStream"));
+        let chain_store = chain.chain_store().clone();
```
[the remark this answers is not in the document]
mangas: What happened with clippy? Shouldn't it detect this kind of thing on CI ?

graph/src/blockchain/mod.rs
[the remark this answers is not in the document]
neysofu: Got it! Resolving.

graph/src/blockchain/mod.rs
neysofu: :+1:

node/src/opt.rs
neysofu: Not adding the crate as a dependency is a good reason I guess, yeah I'm okay with it. Resolving.

neysofu (approved):

chain/ethereum/src/chain.rs
```diff
@@ -291,92 +349,41 @@ impl Blockchain for Chain {
         self.adapter_selector
             .triggers_adapter(loc, capabilities, unified_api_version)
     }
-
-    async fn new_firehose_block_stream(
+    async fn new_block_stream(
         &self,
         deployment: DeploymentLocator,
-        block_cursor: FirehoseCursor,
+        store: Arc<dyn WritableStore>,
```
lutter, line 355: Same comment about passing in `WritableStore`

chain/ethereum/src/chain.rs
```diff
@@ -291,92 +349,41 @@ impl Blockchain for Chain {
...
         start_blocks: Vec<BlockNumber>,
-        subgraph_current_block: Option<BlockPtr>,
         filter: Arc<Self::TriggerFilter>,
         unified_api_version: UnifiedMappingApiVersion,
     ) -> Result<Box<dyn BlockStream<Self>>, Error> {
+        let current_ptr = store.block_ptr();
+        if !self.chain_client().is_firehose() {
```
lutter, line 361: `match self.chain_client()` is much better here since it protects against changes to that enum leading to the wrong behavior here

chain/ethereum/src/env.rs
```diff
@@ -83,6 +83,9 @@ pub struct EnvVars {
     /// Set by the flag `GRAPH_ETHEREUM_GENESIS_BLOCK_NUMBER`. The default value
     /// is 0.
     pub genesis_block_number: u64,
+    /// The time to wait between polls when using polling block ingestor.
+    /// The value is set in millis and the default is 1000.
+    pub ingestor_polling_interval: Duration,
```
lutter, line 88: Not something for this PR, but we should think about making this a per-chain setting in the config file. Same for reorg threshold

graph/src/blockchain/block_stream.rs
```diff
@@ -113,7 +113,7 @@ pub trait BlockStreamBuilder<C: Blockchain>: Send + Sync {
 
     async fn build_polling(
         &self,
-        chain: Arc<C>,
+        chain: &C,
```
lutter, line 116: Nice, since this makes it clear that we are not holding on to the entire `Blockchain`

graph/src/blockchain/firehose_block_stream.rs
```diff
@@ -125,20 +120,23 @@ where
     where
         F: FirehoseMapper<C> + 'static,
     {
+        if !client.is_firehose() {
+            panic!("Firehose block stream called with rpc endpoint");
```
lutter, line 124: Is this user error, i.e., a misconfiguration or a code error? If it's user error would be good to tell them to change their config. Basically: if a user sees this error what should they do? Fix their config or report a bug?

graph/src/blockchain/mock.rs
```diff
@@ -268,6 +272,19 @@ impl<C: Blockchain> RuntimeAdapter<C> for MockRuntimeAdapter {
     }
 }
 
+pub struct MockBlockIngestor {}
```
lutter, line 275: Is this actually used anywhere? Couldn't find it with a grep through the sources

lutter (changes requested):
My main concern with this PR is that it changes some connection errors during startup into panics. We've had situations where it was not possible to start the block ingestor for all chains successfully, and we should make sure that that continues to work even if some chain clients have lost their mind.

graph/src/blockchain/firehose_block_ingestor.rs
```diff
@@ -192,3 +157,69 @@ where
...
+                    ..Default::default()
+                })
+                .await;
+
+            match result {
+                Ok(stream) => {
+                    info!(logger, "Blockstream connected, consuming blocks");
+
+                    // Consume the stream of blocks until an error is hit
+                    latest_cursor = self.process_blocks(latest_cursor, stream).await
```
lutter, line 210: I know it's not from this PR, but we should have a new type `struct FirehoseCursor(Box<str>)` for passing around firehose cursors instead of just using `String` for clarity. Something for the todo list

node/src/main.rs
```diff
@@ -88,6 +85,21 @@ fn read_expensive_queries(
...
 }
 
+macro_rules! collect_ingestors {
+    ($acc:ident, $($chain:ident),+) => {
+        $(
+        $chain.iter().for_each(|(network_name, chain)| {
+            let ingestor = chain.block_ingestor().expect(&format!(
+                "Failed to create ingestor for network {}",
+                network_name
+            ));
```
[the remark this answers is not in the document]
mangas: This is somewhat intentional but I guess this is pretty bad if the provider is down, I will revert to logging this instead

chain/substreams/src/chain.rs
```diff
@@ -172,6 +172,10 @@ impl Blockchain for Chain {
     fn chain_client(&self) -> Arc<ChainClient<Self>> {
         self.client.clone()
     }
+
+    fn block_ingestor(&self) -> anyhow::Result<Box<dyn BlockIngestor>> {
+        panic!("Substreams rely on the block ingestor from the network they are processing")
```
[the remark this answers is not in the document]
mangas: This would only be called if the code is wrong, this should never be called that's why it's a panic

chain/arweave/src/chain.rs
```diff
@@ -107,12 +110,11 @@ impl Blockchain for Chain {
         unimplemented!("This chain does not support Dynamic Data Sources. is_refetch_block_required always returns false, this shouldn't be called.")
     }
 
-    async fn new_firehose_block_stream(
+    async fn new_block_stream(
         &self,
         deployment: DeploymentLocator,
-        block_cursor: FirehoseCursor,
+        store: Arc<dyn WritableStore>,
```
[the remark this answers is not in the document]
mangas: I get that I thought of adding an enum where you could pass one or the other but leads to some iffy code. I don't like this pattern either but it sort of hides the difference between firehose and rpc. Alternatively I could move the 2 functions needed to a separate trait and require that here. This way it's a lot clearer what we need from this argument. What do you think?

graph/src/blockchain/firehose_block_stream.rs
mangas: This is also a case of reporting a bug as there would have been some code path that called this when it was not using firehose

graph/src/blockchain/mock.rs
mangas: you're right, I had added this as a type on the trait but then removed it. Will remove this as well

graph/src/blockchain/firehose_block_ingestor.rs
mangas: https://github.com/graphprotocol/graph-node/issues/4419

chain/arweave/src/chain.rs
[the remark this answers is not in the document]
lutter: Can't you just have the caller call `store.block_cursor()` and `store.block_ptr()` and pass those in? All I am after is to make it clear that the block stream has no real interest in the `WritableStore`

graph/src/blockchain/firehose_block_stream.rs
lutter: Actually, if these are fatal code problems, then it's better to use `unreachable!` instead of `panic!`

chain/arweave/src/chain.rs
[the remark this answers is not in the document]
mangas: I completely agree I just wanted to avoid the caller having to deal with the firehose/non firehose duality since they don't really need to know anything about it. By providing the store (which I agree is a lot of surface for what we actually need) we can let the callee decide which one to use. ie ignore block_cursor if not needed. 

I absolutely agree that this doesn't seem good as is though so I'll try to find a better solution.

node/src/main.rs
```diff
@@ -90,6 +86,24 @@ fn read_expensive_queries(
...
+        $(
+        $chain.iter().for_each(|(network_name, chain)| {
+            let logger = $logger.new(o!("network_name" => network_name.clone()));
+            match chain.block_ingestor() {
+                Ok(ingestor) =>{
+                    info!(logger, "Started block ingestor");
+                    $acc.push(ingestor);
+                }
+                Err(err) => error!(&logger,
+                    "Failed to create block ingestor for network {}",err),
```
lutter, line 100: Better wording: `"Failed to create block ingestor: {}", err`

lutter (approved):

Issue #4420, opened 2023-03-05
`ExponentialBackoff` left shift overflow


YaroShkvorets:
**Do you want to request a *feature* or report a *bug*?**
bug

**What is the current behavior?**
When using `ExponentialBackoff` struct, it takes 32 attempts to overflow the left shift operator on u32 [here](https://github.com/graphprotocol/graph-node/blob/e7d1555427cacc69e40751e0ea308e257b7bae9b/graph/src/util/backoff.rs#L36).
With [current parameters](https://github.com/graphprotocol/graph-node/blob/e7d1555427cacc69e40751e0ea308e257b7bae9b/graph/src/blockchain/substreams_block_stream.rs#L182) (`base` = 0.5s, `ceiling` = 45s) it takes about 20 minutes of missing internet connection for that to happen

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem.**
Obvious looking at the code

**What is the expected behavior?**
Should continue with the `delay == ceiling`

PR #4421, opened 2023-03-06
graph: fix backoff overflow

YaroShkvorets:
This PR fixes an issue with `backoff` left shift overflow #4420 
Also adds tests for that struct.

### DIFF
--- a/graph/src/util/backoff.rs
+++ b/graph/src/util/backoff.rs
@@ -33,7 +33,7 @@ impl ExponentialBackoff {
     }
 
     pub fn delay(&self) -> Duration {
-        let mut delay = self.base.saturating_mul(1 << self.attempt);
+        let mut delay = self.base.saturating_mul(1u32 << self.attempt.min(31));
         if delay > self.ceiling {
             delay = self.ceiling;
         }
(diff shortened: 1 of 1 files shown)


graph/src/util/backoff.rs
```diff
@@ -33,7 +33,7 @@ impl ExponentialBackoff {
     }
 
     pub fn delay(&self) -> Duration {
-        let mut delay = self.base.saturating_mul(1 << self.attempt);
+        let mut delay = self.base.saturating_mul(1u32 << self.attempt.min(31));
```
lutter, line 36: I think it would be better to make sure that `next_attempt` never increments `self.attempt` to more than 31. We can also do the `min` here in addition, but I think we should also clamp `self.attempt` to 31 explicitly

lutter (changes requested):
Nice find!

graph/src/util/backoff.rs
YaroShkvorets: Thought it would be better to let it increment to have a sense how many total attempts there were, i.e. for logging.

lutter: True. That makes sense

lutter (approved):

PR #4424, opened 2023-03-06
`EmptyRuntimeAdapter` to reduce code duplication and simplify adding support for new chains.

neysofu:
All chains except Ethereum currently don't expose any host functions in their runtime adapters. Might as well have an `EmptyRuntimeAdapter` that can be reused by whatever chains.


graph/src/blockchain/empty_runtime_adapter.rs
```diff
@@ -0,0 +1,22 @@
+use std::marker::PhantomData;
+
+use super::{Blockchain, HostFn, RuntimeAdapter};
+
+/// A [`RuntimeAdapter`] that does not expose any host functions.
+#[derive(Debug, Clone)]
+pub struct EmptyRuntimeAdapter<C>(PhantomData<C>);
```
mangas, line 7: What do you think about `NoopRuntimeAdapter`?

mangas (approved):

graph/src/blockchain/empty_runtime_adapter.rs
neysofu: Done!

PR #4426, opened 2023-03-06
Refactor some code in Cosmos' `trigger.rs`

neysofu:
1. `CheapClone::cheap_clone` mirrors `Clone` functionality by default, so it rarely needs to use custom code.
2. ~~Although not immediately obvious, I'm pretty sure the custom implementation of `PartialEq` is identical to what `derive` what do, so I'm removing it.~~ nvm, I don't want to deal with determinism issues at this time so I'll leave it as it is.

### DIFF
--- a/chain/cosmos/src/trigger.rs
+++ b/chain/cosmos/src/trigger.rs
@@ -69,23 +69,7 @@ pub enum CosmosTrigger {
     Message(Arc<codec::MessageData>),
 }
 
-impl CheapClone for CosmosTrigger {
-    fn cheap_clone(&self) -> CosmosTrigger {
-        match self {
-            CosmosTrigger::Block(block) => CosmosTrigger::Block(block.cheap_clone()),
-            CosmosTrigger::Event { event_data, origin } => CosmosTrigger::Event {
-                event_data: event_data.cheap_clone(),
-                origin: *origin,
-            },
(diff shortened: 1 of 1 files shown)


chain/cosmos/src/trigger.rs
```diff
@@ -69,50 +69,7 @@ pub enum CosmosTrigger {
...
-                CosmosTrigger::Transaction(transaction_data.cheap_clone())
-            }
-            CosmosTrigger::Message(message_data) => {
-                CosmosTrigger::Message(message_data.cheap_clone())
-            }
-        }
-    }
-}
-
-impl PartialEq for CosmosTrigger {
```
mangas, line 90: do we have any coverage for this?

neysofu: I don't think we do. I'll think of a way to test this, even though it looks correct at a glance.

mangas: you can prolly compare the expanded code

neysofu: I'll mark this as resolved as I have decided not to tackle it at this time, I don't feel like sweating over determinism issues. I've limited this PR to the `CheapClone` refactor.

mangas (approved):

PR #4427, opened 2023-03-06
Remove `MockMetricsRegistry` in favor of the real one

neysofu:
`MetricsRegistry` doesn't really need to be mocked; in fact, it's very easy to instantiate. I provided a new `mock_new` method that discards all logs related to metrics to simulate the old behavior, even if not really needed from what I can see.


core/src/metrics/registry.rs
```diff
@@ -42,6 +42,10 @@ impl MetricsRegistry {
         }
     }
 
+    pub fn mock_new() -> Self {
```
mangas, line 45: I think mock() is enough. Also I take it you prolly cant add #[cfg(debug_assertions)], but just in case

neysofu: I renamed it to `mock`, I agree it's better. Unfortunately we can't add #[cfg(test)] without adding feature flags to all crates in the workspace (`cfg(test)` otherwise doesn't propagate to depedendencies).

chain/ethereum/src/network.rs
```diff
@@ -391,7 +392,7 @@ mod tests {
     async fn adapter_selector_unlimited() {
         let chain = "mainnet".to_string();
         let logger = graph::log::logger(true);
-        let mock_registry: Arc<dyn MetricsRegistry> = Arc::new(MockMetricsRegistry::new());
+        let mock_registry: Arc<dyn MetricsRegistryTrait> = Arc::new(MetricsRegistry::mock_new());
```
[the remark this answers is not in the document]
neysofu: I created https://github.com/graphprotocol/graph-node/issues/4438 to track this! Planning on addressing it in a future PR.

mangas (approved):

PR #4434, opened 2023-03-08
Derived fields getter

### LINKED ISSUE #4004: Access and save derived fields in mappings
During indexing it is not currently possible to access derived fields. Users who want to access and update entities in those fields therefore need to store them as an array. This leads to unwieldy code, and can lead to massive arrays

Given the following schema:
```
Holder @entity {
  id: ID!
  tokens: [Token!]! @derivedFrom(field: "holder")
}

Token @entity {
  id: ID!
  holder: Holder!
  color: String
}
```

Users can load and save derived entities:
```
const holder = Holder.load("0xAbDe42")
const holderTokens = holder.tokens.load() // loads the derived entities as an array
const updatedTokens = holderTokens.map((token) => {
    const newToken = token;
    newToken.color = 'Red';
    return newToken;
  })

updatedTokens.save() // saves the array of entities in one line
```

Considerations:
- Should there be a limit on the number of derived entities which are loaded into the mapping?
- Do we want to allow for additional filtering / limiting by the user?
- Is there a risk to the "multi-save"?

flametuner:
This PR aims to add a getter for `derivedFields`. Fixes #4004 

### Roadmap
- [x] add `store.loadRelated` host function
- [x] create query for `find_derived`
- [x] retrive derived entities from the database
- [x] create ASC conversion
- [x] get derived from queue
- [x] optimize the get derived query to filter entities ids
- [x] tests


graph/src/components/store/mod.rs
```diff
@@ -137,6 +137,33 @@ pub struct EntityKey {
...
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum EntityMultiKey {
+    /// A filter that matches all entities of a given type.
+    Derived(EntityDerived),
+
+    /// A filter that matches a specific entity.
+    Single(EntityKey),
+}
```
lutter, line 165: I haven't read through the changes in detail, but two questions come to mind from this:
1. How are you going to deal with aliasing? The same entity can now appear with as a `Single` and as (several) `Derived` in the cache. That will lead to correctness issues
2. The size of an `EntityMultiKey` is quite a bit bigger than an `EntityKey` which will increase the amount of memory the `EntityCache` takes or reduce the number of entities that can go into the cache, either of which are not desirable.

flametuner: 1. I'm planning to keep in the `Derived` cache only an Array<String> containing the `ids` so the entity cache won't be duplicated. With that, I can cache the whole Entity as usual with it's own `EntityKey`. Currently, the only missing feature is to add/remove the id of a newly created/removed entities to keep the `Derived` cache updated.
2. The `EntityKey` is used in multiple places all over the code and changing it would be almost a complete refactor of the code. The safest way I found was to wrap it inside the `EntityMultiKey` enum and use it only in the `EntityCache`. I agree that `EntityMultiKey` is bigger than `EntityKey` and I'm open to suggestions.

lutter: What I had in mind was what I outlined in the comment on #4004 : make the lookup understand how to chase down how the original entity and the entities you want to look up are related, and then treat the result as normal entities so that there's no changes needed to the `EntityCache`. Basically, the flow I have in mind is:
1. mapping calls `loadRelated`
2. internally transform that into the right query (which will depend on whether the field being looked up is derived or not, see [here](https://github.com/graphprotocol/graph-node/blob/master/docs/implementation/sql-query-generation.md) for how that's done in general)
3. run that query and get back a list of entities
4. insert them into the cache as if they had been looked up one-by-one with `load`
5. return to the mappings

flametuner: Ok, so essentially, the response of `loadRelated()` would not be cached, just call to the database loading the entities to the cache.

I mean, if I call `loadRelated()` twice it should query the database twice. That's way simpler than what I'm currently doing.

lutter: Aah .. now I see where you're coming from. Yes, this shouldn't be cached since it will be too hard to track what changes will cause the response to change. When we save an entity, we have no idea if that will affect the list of related entities somewhere else. For example, in the data model from #4004,  you could just save a new `Token` entity (or modify an existing one) and set its `holder` to one for which you already used `loadRelated`. To keep the cache accurate, we would have to find all entities that could be affected by that and update the cached list of things. 

If it turns out that repeatedly calling `loadRelated` is an important use case, we can try and add that later.
For now, just having it so that each `loadRelated` call goes to the database and adds the individual entities to the cache without tracking that they were put there by `loadRelated` is good enough.

flametuner: Ok, I removed all the cache logic, now it's way simpler. You can continue to review the PR 😊

flametuner:
I splitted and renamed the structs for a more semantic naming and usage.

flametuner:
@lutter do you think the implementation for the `Queue` with a `Writer::Async` is needed? And if so, how do you suggest to implement?

flametuner:
> From a quick glance at your last commit, it looks like you figured out what needs to happen: the query needs to be run against the store and the result needs to be updated according to the changes still in the queue similar to what `get` and `get_many` do.

Yes, I figured out! The only point of optimization is to filter in the database query the entities ids that we already have from queue.

flametuner:
@lutter I did some optimizations with the query but I would be more secure if I could test it. Do you have any suggestions where I can place the tests for it?

lutter (changes requested):
This all looks very good. There is one issue in `writable.rs` that needs to be addressed and a few small stylistic comments.

lutter:
> @lutter I did some optimizations with the query but I would be more secure if I could test it. Do you have any suggestions where I can place the tests for it?

I think the best place for tests would be in `graph/tests/entity_cache.rs` so you can test the entire path of loading derived entities. The tests in there right now use a `MockStore`, but this test should use the proper `test_store::STORE` etc. It would have to set up its own subgraph with a schema that uses a derived field, populate it with data and then test loading, probably with some edge cases, e.g., for an entity where the set of related entities is empty. Have a look at the tests in `store/postgres/tests` for some of the helpers we have to set up test subgraphs.

runtime/wasm/src/module/mod.rs
```diff
@@ -507,6 +507,14 @@ impl<C: Blockchain> WasmInstance<C> {
         link!("abort", abort, message_ptr, file_name_ptr, line, column);
 
         link!("store.get", store_get, "host_export_store_get", entity, id);
+        link!(
+            "store.loadRelated",
+            store_load_related,
+            "host_export_store_get_derived",
```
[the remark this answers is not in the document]
flametuner: You are right! This is a missing part of the code when I changed from get_derived to load_related

graph/src/data/schema.rs
```diff
@@ -643,6 +643,43 @@ impl Schema {
...
+            Ok((base_type, field.as_str().unwrap()))
+        } else {
+            Err(anyhow!(
+                "Entity {}[{}]: field `{}` is not derived",
+                key.entity_type,
+                key.entity_id,
+                key.entity_field,
+            ))
+        }
+    }
```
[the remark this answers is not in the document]
flametuner: Ohh, now I get it, I didn't know you were planning to use `loadRelated` for every `[Child!]!` inside an entity, derived or not. I thought it was already working. I'll create an issue for it.

store/postgres/src/writable.rs
```diff
@@ -792,6 +803,55 @@ impl Queue {
...
+                tracker.update(req.as_ref());
+                match req.as_ref() {
+                    Request::Write {
+                        block_ptr, mods, ..
+                    } => {
+                        if tracker.visible(block_ptr) {
+                            for emod in mods {
+                                let key = emod.entity_ref();
+                                // we select just the entities that match the query
+                                if derived_query.entity_type == key.entity_type {
```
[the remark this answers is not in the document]
flametuner: You are absolutely right! I missed that one. Thank you.

store/postgres/src/relational.rs
```diff
@@ -539,6 +539,32 @@ impl Layout {
         Ok(entities)
     }
 
+    pub fn find_derived(
+        &self,
+        conn: &PgConnection,
+        derived_query: &DerivedEntityQuery,
+        block: BlockNumber,
+        excluded_keys: &Option<Vec<EntityKey>>,
```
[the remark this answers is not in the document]
flametuner: Hahaha, that's funny! I was doing exactly this and changed to `Option` because I saw in some internet post that we should use Option on  "null values" like `-1`, empty string `""` and I thought a `vec![]` would match this case.

But indeed, in this case there is no difference between `None` and `Some(vec![])`. I was just doing two checks instead of one.

runtime/wasm/src/module/mod.rs
```diff
@@ -1015,6 +1023,33 @@ impl<C: Blockchain> WasmInstanceContext<C> {
...
+            &mut self.ctx.state,
+            entity_type.clone(),
+            id.clone(),
+            field.clone(),
+            gas,
+        )?;
+
+        let entities: Vec<Vec<(String, Value)>> = entities
+            .iter()
+            .map(|entity| entity.clone().sorted())
```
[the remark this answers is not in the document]
flametuner: That's great! TIL.

graph/src/data/schema.rs
```diff
@@ -643,6 +643,43 @@ impl Schema {
         }
     }
 
+    pub fn get_type_for_field(&self, key: &LoadRelatedRequest) -> Result<(&str, &str), Error> {
```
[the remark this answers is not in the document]
flametuner: How is this extensible when we start to use non-derived fields? Is there any `&Field` that represent the relationship between the parent and the child?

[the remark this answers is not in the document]
lutter: Yes, the difference is that with a derived field the schema looks like

```graphql
type Parent {
  children [Child!]! @derivedFrom(field: "parent")
}
```

and for a non-derived field it is
```graphql
type Parent {
  children [Child!]!
}
```

In that case, the parent has a column `children` of either type `text[]` or `bytea[]` depending on the type of `Child.id`

graph/src/data/schema.rs
```diff
@@ -665,11 +680,32 @@ impl Schema {
                 )
             })?;
         if field.is_derived() {
-            let derived_from = field.find_directive("derivedFrom").unwrap();
+            let derived_from = field.find_directive("derivedfrom").unwrap();
```
lutter, line 683: It should really be spelled `derivedFrom`

lutter (changes requested):
Apart from the small type for `derivedFrom`, this looks good. I just realized that this PR does not take one other situation into account: the field that is mentioned in the `@derivedFrom` can also be an array. This [document](https://github.com/graphprotocol/graph-node/blob/master/docs/implementation/sql-query-generation.md) goes through all the possible permutations. But I would also rather address this in a follow-on PR; I believe that taking that into account only requires some small changes to the query generation logic.

graph/src/data/schema.rs
flametuner: Thanks, it was a typo while adding the `&Field` return.

flametuner:
> Apart from the small type for `derivedFrom`, this looks good. I just realized that this PR does not take one other situation into account: the field that is mentioned in the `@derivedFrom` can also be an array. This [document](https://github.com/graphprotocol/graph-node/blob/master/docs/implementation/sql-query-generation.md) goes through all the possible permutations. But I would also rather address this in a follow-on PR; I believe that taking that into account only requires some small changes to the query generation logic.

wow, I didn't know that. Everything that I know about subgraphs was reading the [Official documentation](https://thegraph.com/docs/en/developing/creating-a-subgraph/) and it doesn't have anything related to `@derivedFrom` with single entity (I found within the code) and with array of entities.

I'll look through the document you sent and, if it is the case, try to fail gracefully until the follow-on PR.

lutter (approved):

lutter:
This all looks good now.

@azf20 I forget, what is the process for adding a new host function? I think we need to add a API version 0.0.8 and make sure the new host function is only available with that version?

lutter:
@flametuner can you rebase to latest master?

flametuner:
@lutter is there anything blocking this PR? Please let me know so I can fix it.

lutter:
> @lutter is there anything blocking this PR? Please let me know so I can fix it.

This all looks good - can you rebase once more? I can then merge it.

flametuner:
> This all looks good - can you rebase once more? I can then merge it.

Done

lutter:
@flametuner I think it needs a little bit of additional work - most of the CI checks are failing

flametuner:
> @flametuner I think it needs a little bit of additional work - most of the CI checks are failing

Rebase didn't update some errors types in my code. Just fixed that.

lutter:
@flametuner Thanks a ton for this PR. Great work!

PR #4440, opened 2023-03-09
Instrument `RecentBlocksCache` with metrics

### LINKED ISSUE #4391: Metrics to track chain head caching
We want some Prometheus metrics to accurately track hit rates and other stats from #4154.

neysofu:
Fixes #4391. These new metrics could help us with identifying and root-causing performance and load issues on the database.

### DIFF
--- a/store/postgres/src/chain_store.rs
+++ b/store/postgres/src/chain_store.rs
@@ -1392,6 +1392,20 @@ impl ChainStoreMetrics {
             .with_label_values(&[network])
             .dec();
     }
+
+    pub fn record_cache_hit(&self, network: &str) {
+        self.chain_head_cache_hits
+            .get_metric_with_label_values(&[network])
+            .unwrap()
+            .inc();
+    }
+
+    pub fn record_cache_miss(&self, network: &str) {
+        self.chain_head_cache_misses
+            .get_metric_with_label_values(&[network])
+            .unwrap()
+            .inc();
+    }
 }
 
 pub struct ChainStore {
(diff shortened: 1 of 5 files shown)


store/postgres/src/chain_store.rs
```diff
@@ -1877,6 +1954,42 @@ mod recent_blocks_cache {
...
+                .chain_head_cache_oldest_block_num
+                .get_metric_with_label_values(&[&self.network])
+                .unwrap()
+                .set(self.earliest_block().map(|b| b.ptr.number).unwrap_or(0) as f64);
+
+            self.metrics
+                .chain_head_cache_latest_block_num
+                .get_metric_with_label_values(&[&self.network])
+                .unwrap()
+                .set(self.earliest_block().map(|b| b.ptr.number).unwrap_or(0) as f64);
```
leoyvens, line 1990: This is using `earliest_block()`, shouldn't it be using `chain_head()`?

store/postgres/src/chain_store.rs
```diff
@@ -1877,6 +1954,42 @@ mod recent_blocks_cache {
...
+                .inc();
+        }
+
+        fn record_cache_miss(&self) {
+            self.metrics
+                .chain_head_cache_misses
+                .get_metric_with_label_values(&[&self.network])
+                .unwrap()
+                .inc();
+        }
```
leoyvens, line 1971: You could move `fn record_cache_hit` and `fn record_cache_miss` into `ChainStoreMetrics`.

leoyvens (approved):

store/postgres/src/chain_store.rs
neysofu: Good catch!

store/postgres/src/chain_store.rs
neysofu: :+1: fixed in https://github.com/graphprotocol/graph-node/pull/4440/commits/efbf522284bd3cc705939d4419ed7343d77e9564.

Issue #1823, opened 2020-07-30
Fulltext search allows `where` conditions but ignores them


lutter:
The following query against [everest](https://thegraph.com/explorer/subgraph/graphprotocol/everest) should only return one result, but returns four:
```graphql
{
  projectSearch(text: "Project:*", where: {name:"Everest"}, orderBy: totalVotes) {
    name
  }
}
```

The problem is that the SQL query we end up running completely ignores `where` and `orderBy`:
```sql
select 'Project' as entity,
           to_jsonb(c.*) as data
  from (select * 
          from sgd28.project c
         where c.block_range @> $1
           and project_search @@ to_tsquery($2)
         order by ts_rank("project_search", to_tsquery($3)) asc nulls last, "id" limit 100) c
```

Jannis:
We currently only validate the parts of the GraphQL API that we _expect_, not the ones that we don't expect. Fulltext search fields like `projectSearch` don't support `where` and `orderBy`, they only support `text`. So the behavior here is correct.

It would be nice to have richer validation but we've been doing ok without it and it would be a lot of effort now (and probably break a lot of queries out there). Closing this as won't fix for now.

PR #4442, opened 2023-03-10
Fulltext filter

### LINKED ISSUE #4412: Filter and sort on Fulltext search
Feature

I have a specific usecase for text search that is currently not possible with The Graph.

I'm building a marketplace running on Smart Contracts where only the sellers who have at minimum X stake can create listings. Furthermore, the incentive for a seller to stake more tokens is that the order of the products depends on the seller current stake.

Last but not least, a listing has a time frame where it can be found from creation to expiration.

The problem with this, is that full text search doesn't have filters, so it's not possible to filter only for valid listings. The "solution" that I found was to create an entity called `SearchableListing` , save it and have a derived field inside an entity `Block`. So I created a blockHandler that every block checks for the list of searchableListing, removing them.

But then, there was another problem where I can't really access derivedFields inside the subgraph and for this to work I had to actively create an array inside block, maintain it, and you know the mess this is going.

As I said earlier, I want the listings to be sorted by seller stake and this should be true for searching too, at least between X listings with the same text proximity, they should be sorted desc.

Request:  `where` and `orderBy` in a full text search.

flametuner:
After some conversations in #4413, I splitted the PR and this is the filter for fulltext search queries.
This PR partially fixes #4412, at least the filter part.

There's also another issue #1823, where this PR can be useful.

As requested by @lutter in #4413, I added some tests for fulltext search and full text search filter.

### DIFF
--- a/graph/src/components/store/mod.rs
+++ b/graph/src/components/store/mod.rs
@@ -184,6 +184,7 @@ pub enum EntityFilter {
     NotEndsWithNoCase(Attribute, Value),
     ChangeBlockGte(BlockNumber),
     Child(Child),
+    Fulltext(Attribute, Value),
 }
 
 // A somewhat concise string representation of a filter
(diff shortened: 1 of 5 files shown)


lutter (approved):
Very nice! Great stuff!

PR #4445, opened 2023-03-11
Add facilities to gte more details during indexing

lutter:
These changes make it so that an individual deployment can be set to log all store interactions to help diagnose problems with the deployment.

Right now, this is turned on by manipulating the database directly; if it turns out to be useful, we should add a `graphman` command to turn this on.

### DIFF
--- a/store/postgres/src/deployment.rs
+++ b/store/postgres/src/deployment.rs
@@ -307,32 +307,41 @@ pub struct ManifestInfo {
     pub description: Option<String>,
     pub repository: Option<String>,
     pub spec_version: String,
+    pub instrument: bool,
 }
 
 impl ManifestInfo {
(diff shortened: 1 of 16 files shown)


store/postgres/src/deployment.rs
```diff
@@ -307,32 +307,41 @@ pub struct ManifestInfo {
...
+                sm::features,
             ))
             .filter(sm::id.eq(site.id))
             .first(conn)?;
         let input_schema = Schema::parse(s.as_str(), site.deployment.clone())?;
 
+        // Using the features field to store the instrument flag is a bit
+        // backhanded, but since this will be used very rarely, should not
+        // cause any headaches
+        let instrument = features.iter().any(|s| s == "instrument");
```
leoyvens, line 337: If we're hacking this here, can we at least ensure that the real manifest doesn't set this?

Issue #4443, opened 2023-03-10
Cyclic deps with test_store


flametuner:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**

Currently, `rust-analyzer` can't index the whole project because of some cyclic deps.

```
[ERROR project_model::workspace] cyclic deps: graph_core(CrateId(175)) -> graph_chain_substreams(CrateId(172)), alternative path: graph_chain_substreams(CrateId(172)) -> graph_core(CrateId(175))
[ERROR project_model::workspace] cyclic deps: test_store(CrateId(431)) -> graph(CrateId(156)), alternative path: graph(CrateId(156)) -> test_store(CrateId(431))
[ERROR project_model::workspace] cyclic deps: test_store(CrateId(431)) -> graph_chain_ethereum(CrateId(166)), alternative path: graph_chain_ethereum(CrateId(166)) -> graph(CrateId(156)) -> test_store(CrateId(431))
[ERROR project_model::workspace] cyclic deps: test_store(CrateId(431)) -> graph_graphql(CrateId(177)), alternative path: graph_graphql(CrateId(177)) -> graph(CrateId(156)) -> test_store(CrateId(431))
[ERROR project_model::workspace] cyclic deps: test_store(CrateId(431)) -> graph_mock(CrateId(181)), alternative path: graph_mock(CrateId(181)) -> graph(CrateId(156)) -> test_store(CrateId(431))
[ERROR project_model::workspace] cyclic deps: test_store(CrateId(431)) -> graph_node(CrateId(182)), alternative path: graph_node(CrateId(182)) -> graph(CrateId(156)) -> test_store(CrateId(431))
[ERROR project_model::workspace] cyclic deps: test_store(CrateId(431)) -> graph_store_postgres(CrateId(195)), alternative path: graph_store_postgres(CrateId(195)) -> graph(CrateId(156)) -> test_store(CrateId(431))
```

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem.**

Using `rust-analyzer analysis-stats .` or using the VSCode extension in the project gives this problem.

The development mainly of tests is more difficult because there are a lot of the following error in functions from `test-store`:

```
expected &{unknown}, found &<Type>
```

It just affects the rust-analyzer, cargo continues to build well.

**What is the expected behavior?**

It should index all the files from the different project of the workspace using rust-analyzer.

[...]

PR #4448, opened 2023-03-13
Remove some cyclic crate dependencies

lutter:
Partial fix for #4443 

Unfortunately, the remaining cyclic dependencies are much harder to remove. We might have to have separate test crates for tests in other crates that require the `test_store`, since that ends up being pretty much top-level for testing, similar to the `node` crate for normal code.

Also, it would be nice to get rid of the `MetricsRegistryTrait` and just have the `MetricsRegistry` struct, but that's not directly related to cyclic dependencies.


lutter:
One thing that just occurred to me: we could break the remaining cyclic dependencies fairly easily if we moved tests that require the `test-store` crate into that crate, i.e. tests from `core` or `graphql` that use the `test-store` would get put there.

leoyvens (approved):

lutter:
@leoyvens what are your thoughts on pulling all tests that use `test-store` into that crate?

PR #4463, opened 2023-03-16
graph: fix substreams `startBlock` selection

### LINKED ISSUE #4449: Picking `startBlock` for substreams data source
[This logic](https://github.com/graphprotocol/graph-node/blob/860becc40e9d4350adc1372e430f354c3ae18727/chain/substreams/src/data_source.rs#L175-L178) of picking the startBlock for the module to stream, where we pick the MIN `initial_block` among **all** modules in the package, seems superfluous and wrong.

For some complex substreams it could pick wrong block values resulting in indexing errors like this:
```
An error occurred while streaming blocks: status: InvalidArgument, message: "start block 100000 smaller than request outputs for module \"graph_out\" with start block 1000000"
```

IMO We should just take the `initial_block` of the module being requested.

`substreams pack` already assigns initial_block for modules that don't have `initial_block` set, based on its ancestors, so trying to silently override it here seems unnecessary.

YaroShkvorets:
This fixes #4449 by choosing `initial_block` of the module provided in the manifest.

It also verifies that module name provided in the subgraph manifest exists in the substreams manifest and if not, `graph deploy` errors out rather than letting it deploy and then unsuccessfully retry RPC requests to the substreams endpoint.

### DIFF
--- a/chain/substreams/src/data_source.rs
+++ b/chain/substreams/src/data_source.rs
@@ -277,7 +277,6 @@ mod test {
 
     use crate::{DataSource, Mapping, UnresolvedDataSource, UnresolvedMapping, SUBSTREAMS_KIND};
 
-
     #[test]
     fn parse_data_source() {
         let ds: UnresolvedDataSource = serde_yaml::from_str(TEMPLATE_DATA_SOURCE).unwrap();
--- a/chain/substreams/src/data_source.rs
+++ b/chain/substreams/src/data_source.rs
@@ -280,6 +280,8 @@ mod test {
         components::link_resolver::LinkResolver,
         prelude::{async_trait, serde_yaml, JsonValueStream, Link},
         slog::{o, Discard, Logger},
+        substreams::module::{Kind, KindMap, KindStore},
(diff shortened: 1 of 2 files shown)


YaroShkvorets:
Oh. Making it pass those tests would be non-trivial without the substreams package...

YaroShkvorets:
Ready for review

chain/substreams/src/data_source.rs
```diff
@@ -340,14 +346,35 @@ mod test {
...
+            modules: Some(graph::substreams::Modules {
+                modules: vec![graph::substreams::Module {
+                    name: "output".into(),
+                    initial_block: 123,
+                    binary_entrypoint: "entry".into(),
+                    binary_index: 0,
+                    kind: None,
+                    inputs: vec![],
+                    output: None,
+                }],
```
maoueh, line 362: Actually, could you add a second module with a different block so we see the code is working properly.

maoueh (changes requested):

maoueh (approved):
Once the tests pass, I'll merge this.

PR #4468, opened 2023-03-17
graph-chain-ethereum: Avoid adapters with errors

mangas:
- Wire EndpointMetrics for rpc endpoints
- Use a percentage of traffic to retest errored adapters

### DIFF
--- a/chain/ethereum/src/ethereum_adapter.rs
+++ b/chain/ethereum/src/ethereum_adapter.rs
@@ -10,7 +10,6 @@ use graph::data::subgraph::API_VERSION_0_0_7;
 use graph::prelude::ethabi::ParamType;
 use graph::prelude::ethabi::Token;
 use graph::prelude::tokio::try_join;
-use graph::url::Url;
 use graph::{
     blockchain::{block_stream::BlockWithTriggers, BlockPtr, IngestorError},
     prelude::{
@@ -61,8 +60,6 @@ use crate::{
 #[derive(Debug, Clone)]
 pub struct EthereumAdapter {
     logger: Logger,
-    pub(crate) url: Url,
-    /// The label for the provider from the configuration
     provider: String,
     web3: Arc<Web3<Transport>>,
     metrics: Arc<ProviderEthRpcMetrics>,
@@ -85,7 +82,6 @@ impl CheapClone for EthereumAdapter {
         Self {
             logger: self.logger.clone(),
             provider: self.provider.clone(),
-            url: self.url.clone(),
             web3: self.web3.cheap_clone(),
             metrics: self.metrics.cheap_clone(),
             supports_eip_1898: self.supports_eip_1898,
@@ -102,15 +98,11 @@ impl EthereumAdapter {
     pub async fn new(
         logger: Logger,
         provider: String,
-        url: &str,
         transport: Transport,
         provider_metrics: Arc<ProviderEthRpcMetrics>,
         supports_eip_1898: bool,
         call_only: bool,
     ) -> Self {
-        // Unwrap: The transport was constructed with this url, so it is valid and has a host.
-        let url = graph::url::Url::parse(url).unwrap();
-
         let web3 = Arc::new(Web3::new(transport));
 
         // Use the client version to check if it is ganache. For compatibility with unit tests, be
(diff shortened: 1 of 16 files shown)


chain/ethereum/src/network.rs
```diff
@@ -28,15 +33,43 @@ impl EthereumNetworkAdapter {
...
         self.adapter.is_call_only()
     }
+
+    pub fn get_capacity(&self) -> AvailableCapacity {
+        self.limit.get_capacity(Arc::strong_count(&self.adapter))
+    }
+
+    pub fn current_error_count(&self) -> u64 {
+        self.endpoint_metrics
+            .get_count(&self.adapter.url.as_str().into())
```
leoyvens, line 43: Can we use `provider` instead to identify the adapter, just to keep the url private?

chain/ethereum/src/network.rs
```diff
@@ -13,10 +15,13 @@ use crate::adapter::EthereumAdapter as _;
...
 use crate::EthereumAdapter;
 
+pub const DEFAULT_ADAPTER_ERROR_RETEST_PERCENT: f64 = 0.2;
+
 #[derive(Debug, Clone)]
 pub struct EthereumNetworkAdapter {
+    endpoint_metrics: Arc<EndpointMetrics>,
     pub capabilities: NodeCapabilities,
-    adapter: Arc<EthereumAdapter>,
+    pub adapter: Arc<EthereumAdapter>,
```
leoyvens, line 24: This `pub` seems unecessary.

leoyvens (changes requested):

chain/ethereum/src/ethereum_adapter.rs
```diff
@@ -841,7 +838,7 @@ impl EthereumAdapter {
 #[async_trait]
 impl EthereumAdapterTrait for EthereumAdapter {
     fn url_hostname(&self) -> &str {
```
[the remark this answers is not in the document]
mangas: removed

chain/ethereum/src/network.rs
mangas: for adapters we can, it leaks a bit of the impl details onto config, added comments where relevant

chain/ethereum/src/network.rs
mangas: removed

leoyvens (changes requested):
My suggestion would be that we switch to always identifying providers by name rather than url, for metrics purposes. Urls can contain sensitive information such as auth data, so it's best to restrict the places where they are exposed. So instead of `ChainSection::provider_urls` we'd have `ChainSection::providers` exposing just the provider names for the metrics.

leoyvens:
Having the 'paren provider' seems like a solution. You are concerned that the user changes the provider name in the config and loses the continuity of the metrics? The same could be said if the URL changes, I don't think one is more likely to change than the other.

mangas:
> Having the 'parent provider' seems like a solution. You are concerned that the user changes the provider name in the config and loses the continuity of the metrics? The same could be said if the URL changes, I don't think one is more likely to change than the other.

My concern is more about the code allowing for a dev to do the wrong thing inadvertently since the details of the conn pool leak into the different components like the metrics and deduplication

node/src/chain.rs
```diff
@@ -130,11 +130,13 @@ pub fn create_substreams_networks(
...
 
-                for i in 0..firehose.conn_pool_size {
+                for _ in 0..firehose.conn_pool_size {
                     parsed_networks.insert(
                         name.to_string(),
                         Arc::new(FirehoseEndpoint::new(
-                            &format!("{}-{}", provider.label, i),
+                            // This label needs to be the original label so that the metrics
+                            // can be deduped.
+                            &provider.label,
```
leoyvens, line 139: This is a functional change right, instead of segregating the metrics by individual conns, they are grouped for the whole pool? Makes sense to me, but just confirming.

leoyvens (approved):

node/src/chain.rs
mangas: The metrics will be tagged with provider/host, result, request type and so on. I changed from hosts to providers which means that the dedup needs to happen on the provider level (eg, if one conn of providerA fails, all of the providerA conns should be de-prioritised until checked again).

PR #4469, opened 2023-03-17
Use GitHub issue forms

neysofu:
This PR removes the current GitHub issue template and replaces it with two [issue forms](https://github.blog/changelog/2021-06-23-issues-forms-beta-for-public-repositories/). One issue form is for bug reports, while the other is for feature requests. The goal would be to get more structured information out of our GitHub issues. Users still have the freedom to select a blank template if they don't wish to use any of the two premade templates.

I have deployed this feature in my personal fork of `graph-node`, here: https://github.com/neysofu/graph-node/issues/new/choose. Feel free to try it out.


azf20 (approved):
LGTM!

PR #4470, opened 2023-03-17
Remove `MetricsRegistryTrait` and use just `struct MetricsRegistry`

### LINKED ISSUE #4438: Remove the `MetricsRegistry` trait and directly reference the `struct` everywhere
PR https://github.com/graphprotocol/graph-node/pull/4427 has removed `MockMetricsRegistry`, leaving us with just a single `MetricsRegistry` trait implementation (the `struct` with the same name). Considering how there will hardly be a need for reintroducing mocking for it again, we should probably remove the trait entirely and rely solely on the `struct`.

neysofu:
Fixes #4438. See also #4443, which recently touched on some of this code.

It's all in one big commit, apologies, as I didn't know how to best break it up into smaller pieces. The only changes are pretty much:

- Delete `MetricsRegistryTrait` and add its default method implementations to `MetricsRegistry`.
- Change a bunch of imports.


lutter (approved):
Nice! One less trait we don't need, and I had totally forgotten how awkward the split between the trait and the impl was.

PR #4471, opened 2023-03-17
Upgrade to Rust 1.68.0

neysofu:
- Rust 1.67 replaces `std::sync::mpsc` with the `crossbeam-channel` implementation, improving performance.
- Rust 1.68 introduces Cargo's new "sparse indexing protocol", which TL;DR reduces download times to https://crates.io and is especially useful for CI. It's not yet the default option, so I turned it on in our `.cargo/config.toml`.

I'm also removing `crossbeam-channel` and replacing it with `std::sync::mpsc`, since now they are effectively the same thing.

### DIFF
--- a/node/Cargo.toml
+++ b/node/Cargo.toml
@@ -20,7 +20,6 @@ graphql-parser = "0.4.0"
 futures = { version = "0.3.1", features = ["compat"] }
 lazy_static = "1.2.0"
 url = "2.3.1"
-crossbeam-channel = "0.5.7"
 graph = { path = "../graph" }
 graph-core = { path = "../core" }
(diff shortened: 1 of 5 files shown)


node/Cargo.toml
```diff
@@ -42,6 +41,6 @@ toml = "0.7.1"
 shellexpand = "2.1.0"
 termcolor = "1.2.0"
 diesel = "1.4.8"
-http = "0.2.5" # must be compatible with the version rust-web3 uses
+http = "0.2.5"                                                      # must be compatible with the version rust-web3 uses
```
neysofu, line 44: Ops, forgot to turn off my TOML autoformatter.

PR #4483, opened 2023-03-20
store: Kill the background writer when Writable is dropped

lutter:


leoyvens:
Abort is kind of a big hammer, it can drop the task at any await point, and it's hard to know that this won't cause issues, such as preventing some error cleanup from running. In this case it does seem safe, but still I'd find it nicer if the `Queue` had a cancelation mechanism, such as holding a `oneshot::Receiver` which is the writer selects on together with `queue.peek()`, so that it can be canceled when waiting on an empty queue or getting the next request.

lutter:
Yeah, maybe I was being overly cautious with this PR - there already is a way to gracefully shut down which is queuing a `Request::Stop`. And looking through the [main loop](https://github.com/graphprotocol/graph-node/blob/master/store/postgres/src/writable.rs#L592-L636) for the writer, it only keeps running as long as we successfully process 'normal' requests.

The only way we could leak a task here is if we don't shut the task down gracefully. Maybe we should call `queue.stop()` from `drop` instead to guard against that, though it would be another `graph::block_on(queue.stop())` and I am a bit queasy about doing something like this from `drop`

All this is to say: I agree that we should use the `oneshot` mechanism instead.

leoyvens:
@lutter Using `stop` from `Drop` sounds like it would achieve the same thing, I agree we wouldn't want to use `block_on` in `Drop`, I think we could make it synchronous if the bounded queue had a "force push" that ignores the semaphore. I'm fine with whichever solution you prefer here.

lutter:
Closing this for now since I don't think I'll finish this any time soon, and it doesn't seem overly pressing

PR #4485, opened 2023-03-21
graph: Use a map with interned keys for `Entity`

lutter:
This is an implementation of a string pool that I've had kicking around locally for a long time. The idea behind this is that we have a lot of places where we deal with maps whose keys are strings, and those keys come from a known, fixed set of strings. On the indexing side, those are the names of attributes from the subgraph schema, and for queries it's those names plus field aliases in the query.

Moving from maps with string keys to the `Object` struct, which is a map keyed on interned strings, should reduce memory consumption quite a bit.

The main reason to open this PR is to start a discussion around this approach before I go and plumb this into the places where we deal with such maps, mostly the `Entity` type during indexing and the `r::Value` type for queries.


leoyvens (commented):
The interner implementation looks good! Perhaps the next step, before a full refactor, is to switch the `Entity` implementation to use the interner `Object`, and see if that brings up any new concerns? One thing I'm curious about is the stable hash implementation, and if we can keep that consistent when switching to this.

lutter:
> before a full refactor, is to switch the `Entity` implementation to use the interner `Object`, and see if that brings up any new concerns?

Yes, that's what I have been working on - it involves quite a bit of change since we need to get rid of `Entity::new` and similar methods that create entities and make the schema a factory for entities (that's where the string pool lives) I'll add to this PR once I have something that's halfway understandable.

> One thing I'm curious about is the stable hash implementation, and if we can keep that consistent when switching to this.

It should follow the implementation for `HashMap` and not change the stable hash - I need a helper from the stable hash crate to become public to do that, but it shouldn't be very hard since the helper operates on an iterator over `(&str, Value)`.

graph/src/util/intern.rs
```diff
@@ -0,0 +1,373 @@
...
+}
+
+/// A pool of interned strings. Pools can be organized hierarchically with
+/// lookups in child pools also considering the parent pool. The chain of
+/// pools from a pool through all its ancestors act as one big pool to the
+/// outside.
+pub struct AtomPool {
+    base: Option<Arc<AtomPool>>,
+    base_sym: AtomInt,
+    atoms: Vec<Box<str>>,
```
That3Percent, line 54: If you are attempting to reduce memory consumption, would it be better to use `Arc` instead of `Box` here? The `str` in `atoms` and `words` are identical.

lutter:
This PR now integrates the `AtomPool` into `graph-node` and uses it to back `Entity`, i.e., the indexing side now uses objects with interned keys to represent entities. Most of this PR is concerned with moving schema-handling code between crates/modules with the goal of making the new `InputSchema` a factory for entities since it is now not allowed to create free-standing entities - they all need to be based on a schema (except for entities in tests)

graph/src/util/intern.rs
lutter: Yes, that might be a win depending on the average size of those strings. `Arc` introduces an overhead of 24 bytes, whereas `str` is 16 bytes, so the savings come down to how many strings fit into 8 bytes. But for now, the main win of interning will be to reduce hundreds of copies of the same string to two.

graph/src/schema/input_schema.rs
```diff
@@ -282,3 +291,61 @@ impl Inner {
         Entity::try_make(self.pool.clone(), iter)
     }
 }
+
+/// Create a new pool that contains the names of all the types defined
+/// in the document and the names of all their fields
+fn atom_pool(document: &s::Document) -> AtomPool {
+    let mut pool = AtomPool::new();
+    // These two entries are always required
```
leoyvens, line 305: three entries.

leoyvens (approved):
Amazing work! It's hard to review in depth since the PR is quite extensive, but the commits were meticulously organized as always, I read through them and they look good. We should give this a short run on the test cluster because refactoring `Entity` is sensitive and we should make sure PoIs are unaffected.

graph/src/schema/input_schema.rs
lutter: Heh .. two of them shouldn't be there. I'll have a follow-on PR that removes `__typename` and `g$parent_id`; they are only for the query side.

lutter:
I added one more commit to reduce the size of `u16` - not much of a win for now, but it'll make it easier when we improve the memory layout of `Object` further.

mangas:
Just for future reference, I think it would have been nicer to have the self-contained implementation and tests in one PR with a refactor in a separate PR

graph/src/components/store/entity_cache.rs
```diff
@@ -170,7 +177,8 @@ impl EntityCache {
             }
             None => {
                 let value = self.schema.id_value(&key)?;
-                entity.set("id", value);
+                // unwrap: our AtomPool always has an id in it
+                entity.set("id", value).unwrap();
```
mangas, line 181: could be useful to have a set_id fn that doesn't return an error? I think it would make it easier to use correctly

graph/src/schema/input_schema.rs
```diff
@@ -0,0 +1,357 @@
...
+use crate::prelude::q::Value;
+use crate::prelude::{s, DeploymentHash};
+use crate::schema::api_schema;
+use crate::util::intern::AtomPool;
+
+use super::fulltext::FulltextDefinition;
+use super::{ApiSchema, Schema, SchemaValidationError};
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct InputSchema {
```
mangas, line 22: Could you please add some comments to help understand what this is supposed to be used for?

graph/src/schema/input_schema.rs
```diff
@@ -0,0 +1,357 @@
...
+    inner: Arc<Inner>,
+}
+
+#[derive(Debug, PartialEq)]
+pub struct Inner {
```
mangas, line 27: Does it make sense for Inner to be pub?

graph/src/schema/input_schema.rs
```diff
@@ -0,0 +1,357 @@
...
+
+    pub fn parse(raw: &str, id: DeploymentHash) -> Result<Self, Error> {
+        let schema = Schema::parse(raw, id)?;
+
+        Ok(Self::create(schema))
+    }
+}
+
+impl Inner {
+    pub fn api_schema(&self) -> Result<ApiSchema, anyhow::Error> {
```
mangas, line 84: could you also add some comments to these other exported fns ?

graph/src/util/intern.rs
```diff
@@ -0,0 +1,658 @@
+//! Interning of strings.
+//!
+//! This module provides an interned string pool `AtomPool` and a map-like
```
mangas, line 3: nitpick: I don't think this really is a util, it's more of a data structure as state, also intern doesn't really describe what is contained in the file so I think either a re-usable crate if possible or at least a different package name/path would benefit readability

graph/src/schema/api.rs
```diff
@@ -71,6 +73,220 @@ impl TryFrom<&r::Value> for ErrorPolicy {
     }
 }
 
+#[derive(Debug)]
+pub struct ApiSchema {
```
mangas, line 77: Comment would be great

mangas (approved):
See comments

graph/src/components/store/entity_cache.rs
lutter: I'll have another PR that gets rid of single-key changes to entities - they are only needed in tests. In general, an `Entity` is used to shuttle data between WASM and the store, and we don't really need to update an `Entity` once it's been constructed (this `id` setting logic then moves to `store_set` before the `Entity` is constructed) That has the nice side-effect that an `Entity` always has an `id` and `Entity.id()` can just return `String`

graph/src/schema/input_schema.rs
lutter: Added a comment

graph/src/schema/input_schema.rs
```diff
@@ -0,0 +1,357 @@
...
+            .field(&key.entity_field)
+            .ok_or_else(|| {
+                anyhow!(
+                    "Entity {}[{}]: unknown field `{}`",
+                    key.entity_type,
+                    key.entity_id,
+                    key.entity_field,
+                )
+            })?;
+        if field.is_derived() {
```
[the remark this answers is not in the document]
lutter: This is code that I moved verbatim from `schema.rs`; I purposely did not touch the code to make the review a bit easier.

graph/src/schema/api.rs
lutter: This is code that I moved here from somewhere else as-is; I'll add a comment

graph/src/schema/input_schema.rs
lutter: I removed the `impl Inner` and the `Deref`

graph/src/schema/input_schema.rs
lutter: A lot of this is just existing code moved here from other places - for better or worse, a lot of that doesn't have comments, and commenting it all would be a pretty big undertaking. I'll add some comments, but can't do that for everything here.

lutter:
Just for reference: I ran this PR in the integration cluster for ~ 24 hours without any PoI differences.

lutter:
Rebased to latest `master`

PR #4489, opened 2023-03-22
store: Disable collecting write metrics for synced deployments

lutter:
It's been a long-standing annoyance that the metrics for writing keep getting updated for synced subgraphs.

### DIFF
--- a/store/postgres/src/writable.rs
+++ b/store/postgres/src/writable.rs
@@ -842,6 +842,10 @@ impl Queue {
     fn poisoned(&self) -> bool {
         self.poisoned.load(Ordering::SeqCst)
     }
+
+    fn deployment_synced(&self) {
+        self.stopwatch.disable()
+    }
 }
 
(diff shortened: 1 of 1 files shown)


leoyvens:
I think `deployment_synced` is only called once though, so if the graph node is restarted it won't be called again and the stopwatch will re-enable.

lutter:
> I think `deployment_synced` is only called once though, so if the graph node is restarted it won't be called again and the stopwatch will re-enable.

It looks to me like `deployment_synced` is called once every time `graph-node` starts: that call is [here](https://github.com/graphprotocol/graph-node/blob/master/core/src/subgraph/runner.rs#L782-L803) and `self.state.synced` is initialized to `false` [here](https://github.com/graphprotocol/graph-node/blob/master/core/src/subgraph/runner.rs#L61) There are no other assignments to it.

leoyvens:
Ah right, I assumed that was initialized from the DB.

leoyvens (approved):
Nice!

Issue #4496, opened 2023-03-27
[Bug] Pruning can set `earliest_block` incorrectly


lutter:
### Bug report

When copying or grafting onto a subgraph that is being pruned at the same time (either through an explicit command or through the feature in PR #4429 the `earliest_block` of the destination deployment can be set incorrectly because it the value is read from the source when copying starts, which can take hours or days, and written to the destination as one of the last things during copying. While the copy is ongoing, it is possible that the source gets pruned and the destination contains no or incomplete data for the blocks between the old and the new value of `earliest_block`.

The `earliest_block` of the destination will be correct once the destination is pruned, too. 

The fix for this issue is to read the `earliest_block` again from the source, as late in the copy process as possible, and use that as the `earliest_block` of the destination.

### Some information to help us out

- [ ] Tick this box if this bug is caused by a regression found in the latest release.
- [ ] Tick this box if this bug is specific to the hosted service.
- [X] I have searched the issue tracker to make sure this issue is not a duplicate.

### OS information

None

PR #4502, opened 2023-03-28
Ensure `earliest_block` on copies is set correctly

lutter:
Fixes https://github.com/graphprotocol/graph-node/issues/4496

### DIFF
--- a/store/postgres/src/connection_pool.rs
+++ b/store/postgres/src/connection_pool.rs
@@ -54,10 +54,22 @@ impl ForeignServer {
         format!("shard_{}", shard.as_str())
     }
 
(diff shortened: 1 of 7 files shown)


store/postgres/src/connection_pool.rs
```diff
@@ -202,6 +214,8 @@ impl ForeignServer {
             "subgraph_deployment_assignment",
             "subgraph",
             "subgraph_version",
+            "subgraph_deployment",
+            "subgraph_manifest",
```
leoyvens, line 218: Mapping these could be useful for some analytics queries as well, nice.

leoyvens (approved):
Looks good!

PR #4505, opened 2023-03-28
Docs for pruning and some internal renaming

lutter:
This PR provides some user-level explanation of how pruning works and renames the copy strategy for pruning to 'rebuild'. Since we already have a 'copy' operation in `graph-node`, calling the strategy 'rebuild' reduces the risk for confusion between the two very different operations.

### DIFF
--- a/docs/implementation/pruning.md
+++ b/docs/implementation/pruning.md
@@ -1,12 +1,15 @@
 ## Pruning deployments
 
-Pruning is an operation that deletes data from a deployment that is only
-needed to respond to queries at block heights before a certain block. In
-GraphQL, those are only queries with a constraint `block { number: <n> } }`
-or a similar constraint by block hash where `n` is before the block to
-which the deployment is pruned. Queries that are run at a block height
-greater than that are not affected by pruning, and there is no difference
-between running these queries against an unpruned and a pruned deployment.
+Subgraphs, by default, store a full version history for entities, allowing
+consumers to query the subgraph as of any historical block. Pruning is an
+operation that deletes entity versions from a deployment older than a
+certain block, so it is no longer possible to query the deployment as of
+prior blocks. In GraphQL, those are only queries with a constraint `block {
+number: <n> } }` or a similar constraint by block hash where `n` is before
+the block to which the deployment is pruned. Queries that are run at a
+block height greater than that are not affected by pruning, and there is no
+difference between running these queries against an unpruned and a pruned
+deployment.
 
 Because pruning reduces the amount of data in a deployment, it reduces the
 amount of storage needed for that deployment, and is beneficial for both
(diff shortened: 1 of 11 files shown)


docs/implementation/pruning.md
```diff
@@ -0,0 +1,82 @@
+## Pruning deployments
+
+Pruning is an operation that deletes data from a deployment that is only
```
azf20, line 3: I might start with the higher level context for, maybe something like...
By default, subgraphs store a full version history for entities, allowing consumers to query the subgraph as of any historical block.  Pruning is an operation that deletes entity versions from a deployment older than a certain block, so it is no longer possible to query that deployment as of prior blocks. In GraphQL...

docs/implementation/pruning.md
```diff
@@ -0,0 +1,82 @@
...
+`ETHEREUM_REORG_THRESHOLD` to make sure that reverts can never conflict
+with pruning.
+
+Pruning is started by running `graphman prune`. That command will perform
+an initial prune of the deployment and set the subgraph's `history_blocks`
+setting which is used to periodically check whether the deployment has
+accumulated more history than that. Whenever the deployment does contain
+more history than that, the deployment is automatically repruned. If
+ongoing pruning is not desired, pass the `--once` flag to `graphman
+prune`. Ongoing pruning can be turned off by setting `history_blocks` to a
```
azf20, line 35: To check my understanding, the turning off pointer here is saying that if you pruned once with (say) 10,000 blocks (setting `history_block` to 10,000), if you want to turn off pruning you might call `graphman prune --history 1000000000` so 1B blocks, which is _effectively_ no pruning)?

docs/implementation/pruning.md
```diff
@@ -0,0 +1,82 @@
...
+much database work.
+
+Pruning uses two different strategies for how to remove unneeded data:
+rebuilding tables and deleting old entity versions. Deleting old entity
+versions is straightforward: this strategy deletes rows from the underlying
+tables. Rebuilding tables will copy the data that should be kept from the
+existing tables into new tables and then replaces the existing tables with
+these much smaller tables. Which strategy to use is determined for each
+table individually, and governed by the settings for
+`GRAPH_STORE_HISTORY_REBUILD_THRESHOLD` and
```
azf20, line 56: Are these thresholds 0-1 (i.e. 0.5 is 50%)? Or 0-100?

docs/implementation/pruning.md
```diff
@@ -0,0 +1,82 @@
...
+`GRAPH_STORE_HISTORY_DELETE_THRESHOLD`: if we estimate that we will remove
+more than `REBUILD_THRESHOLD` of the table, the table will be rebuilt. If
+we estimate that we will remove a fraction between `REBUILD_THRESHOLD` and
+`DELETE_THRESHOLD` of the table, unneeded entity versions will be
```
azf20, line 60: Are there checks that REBUILD_THRESHOLD is greater than DELETE_THRESHOLD (does it matter?)

docs/implementation/pruning.md
```diff
@@ -0,0 +1,82 @@
...
+is not changed at all. With both strategies, operations are broken into
+batches that should each take `GRAPH_STORE_BATCH_TARGET_DURATION` seconds
+to avoid causing very long-running transactions.
+
+### Caveats
+
+Pruning is a user-visible operation and does affect some of the things that
+can be done with a deployment:
+
+* because it removes history, it restricts how far back time-travel queries
```
azf20, line 71: maybe worth linking to the time travel docs page?

docs/implementation/pruning.md
```diff
@@ -0,0 +1,82 @@
...
+prune operation is performed. When pruning finishes, it updates the
+`earliest_block` for the deployment. The `earliest_block` can be retrieved
+through the `index-node` status API, and `graph-node` will return an error
+for any query that tries to time-travel to a point before
+`earliest_block`. The value of `history_blocks` must be greater than
+`ETHEREUM_REORG_THRESHOLD` to make sure that reverts can never conflict
+with pruning.
+
+Pruning is started by running `graphman prune`. That command will perform
+an initial prune of the deployment and set the subgraph's `history_blocks`
```
azf20, line 30: Is that initial prune now async (i.e. it doesn't block indexing?)

azf20 (approved):
Looks good to me, just some ideas in the comments!

docs/implementation/pruning.md
lutter: Nice, incorporated.

docs/implementation/pruning.md
lutter: yes, that's exactly what I meant here

docs/implementation/pruning.md
lutter: Just checked the code - there are actually no checks; you could use that to make REBUILD_THRESHOLD lower than DELETE_THRESHOLD which would disable rebuilding

docs/implementation/pruning.md
lutter: Yes, it's between 0 and 1, added that to the text

docs/implementation/pruning.md
lutter: Just looked at the time-travel doc, and it's super low-level about how rows in the db are manipulated. Seems we miss more of a user-level explanation of it.

docs/implementation/pruning.md
lutter: Good point, added a paragraph fro that. It blocks indexing with the rebuild strategy while it copies nonfinal entities. I also added another paragraph explaining what log output to look for.

Issue #4304, opened 2023-01-20
Firehose only provider should be possible for Ethereum


maoueh:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**

If only a Firehose provider is specified in the graph node config for an Ethereum chain, the chain is not wired correctly and will not be available for indexing.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem.**

With this config:

```
[general]

[store]
[store.primary]
connection = "postgresql://graph-node:let-me-in@localhost:5432/graph-node"
weight = 1
pool_size = 10

[chains]
ingestor = "block_ingestor_node"
[chains.mainnet]
shard = "primary"
protocol = "ethereum"
provider = [
  { label = "firehose", details = { type = "firehose", url = "$FIREHOSE_ENDPOINT", token = "$STREAMING_FAST_API_TOKEN", features = [ "filters" ] }},
]

[deployment]
[[deployment.rule]]
shard = "primary"
indexers = [ "default" ]
```

No chain is available at all. 

**What is the expected behavior?**

1 Ethereum `mainnet` chain should be available.

**Details**

Some hunting led me to [create_ethereum_networks_for_chain](https://github.com/graphprotocol/graph-node/blob/02bff21448bc40714a0ce2a4b8cd01abb2919056/node/src/chain.rs#L399). This function is what actually populate the initial `eth_networks` that is used later to connect to adapter and firehose endpoints.

This function is however only looking for `web3` provider, so if no RPC node is configured at all, the chain is not created as no provider is set up.

paymog:
possibly related to https://github.com/graphprotocol/graph-node/issues/4059?

azf20:
Closing this as I believe it is now resolved, please re-open @maoueh if this is still live

matthewdarwin:
@azf20 is it really resolved that i need no RPC provider now?

azf20:
@matthewdarwin you still need a `web3call` provider if the subgraph requires `eth_call` (this could be an EVMX rather than an archive node)

paymog:
Seems that RPC providers will be needed until https://github.com/graphprotocol/graph-node/issues/4431 is resolved

matthewdarwin:
Yes you need an RPC provider for eth_call, but if you have no eth_call, then should not be needed.

YaroShkvorets:
Doesn't seem to be resolved.

Getting this error on latest version:
```
[...]

PR #4508, opened 2023-03-30
chain/ethereum: RuntimeAdapter lazy EthAdapter

mangas:
Fixes https://github.com/graphprotocol/graph-node/issues/4304

### DIFF
--- a/chain/ethereum/src/network.rs
+++ b/chain/ethereum/src/network.rs
@@ -193,6 +193,10 @@ impl EthereumNetworks {
         }
     }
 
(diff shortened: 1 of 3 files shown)


neysofu (changes requested):
Instead on making the truple `Option`al, which is a bit of a long type, I think it would be better to have an `fn insert_empty(&mut self, name: String)` or `fn insert_noop(&mut self, name: String)`. I'll leave the exact naming choice to you as you see fit.

neysofu (approved):

PR #4510, opened 2023-04-01
Restructure integration tests to make rust-analyzer happy

lutter:
This PR removes the cyclic dev-dependencies that make [rust-analyzer almost useless](https://github.com/rust-lang/rust-analyzer/issues/14167) by moving integration tests from their individual crates into the `test-store` crate. It's a little weird to have the tests outside their crates, but almost all integration tests really do need the whole system directly or indirectly.

Given a choice between a little weirdness with test placement and a rust-analyzer that by now is basically unable to find anything, I'd rather live with the weirdness.


leoyvens (approved):
Great quality of life improvement for all of us!

I guess `store: Several improvements to Writable` is mostly unrelated, but that looks good as well.

lutter:
> I guess `store: Several improvements to Writable` is mostly unrelated, but that looks good as well.

That one was actually necessary (and the hardest part of this PR) since all the `store/postgres` tests now run as one test, and the single-step logic caused tests to deadlock.

PR #4533, opened 2023-04-11
Avoid `writer poisoned` errors

lutter:
If a subgraph encountered an error, it would get restarted with some backoff. If that error was caused by some store interaction, the `WritableStore` would remain in poisoned state and refuse to process any more writes.

This PR changes that so that the `WritableStore` is restarted when a subgraph is restarted, clearing the error. That requires that we clear any internal state that is based on assumptions of what has been written so far. I am not entirely sure that the changes in `core/src/subgraph/runner.rs` is enough to do that, and would appreciate a thorough review of what other state might have to be cleared.

### DIFF
--- a/store/postgres/src/writable.rs
+++ b/store/postgres/src/writable.rs
@@ -1318,9 +1318,22 @@ impl WritableStoreTrait for WritableStore {
 
     async fn restart(self: Arc<Self>) -> Result<Option<Arc<dyn WritableStoreTrait>>, StoreError> {
         if self.poisoned() {
+            // When the writer is poisoned, the background thread has
+            // finished since `start_writer` returns whenever it encounters
+            // an error. Just to make extra-sure, we log a warning if the
+            // join handle indicates that the writer hasn't stopped yet.
             let logger = self.store.logger.clone();
-            if let Err(e) = self.stop().await {
-                warn!(logger, "Writable had error when stopping, it is safe to ignore this error"; "error" => e.to_string());
+            match &self.writer {
+                Writer::Sync(_) => { /* can't happen, a sync writer never gets poisoned */ }
+                Writer::Async { join_handle, queue } => {
+                    let err = match queue.check_err() {
+                        Ok(()) => "error missing".to_string(),
+                        Err(e) => e.to_string(),
+                    };
+                    if !join_handle.is_finished() {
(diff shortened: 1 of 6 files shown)


leoyvens:
For correctness, you should use `SubgraphRunner::revert_state`, which is suppose to clear any dirty state written at or beyond a given block number.

The current approach also has a performance issue, which is that it clears the entity cache on every `Action::Restart`. But that is also used [when a dynamic data source is created](https://github.com/graphprotocol/graph-node/blob/e556be6fd490d5a03f8e72c78fc9795f8251903f/core/src/subgraph/runner.rs#L480), where clearing the cache would be unnecessarily detrimental to performance. So we want to revert the state in the error case, but not in the created data source case. One way to do this would be to add an `Action::RestartFromError(BlockPtr)`, which is used when restarting mid-block, with potentially dirty state.

lutter:
I added a commit that hopefully addresses that

store/postgres/src/writable.rs
```diff
@@ -1312,4 +1314,20 @@ impl WritableStoreTrait for WritableStore {
     async fn flush(&self) -> Result<(), StoreError> {
         self.writer.flush().await
     }
+
+    async fn restart(self: Arc<Self>) -> Result<Option<Arc<dyn WritableStoreTrait>>, StoreError> {
+        if self.poisoned() {
+            let logger = self.store.logger.clone();
+            if let Err(e) = self.stop().await {
+                warn!(logger, "Writable had error when stopping, it is safe to ignore this error"; "error" => e.to_string());
+            }
```
leoyvens, line 1323: Is it sufficient to send a stop request here, shouldn't this also join the task handle of the writer, to ensure is finished?

lutter: I am actually wondering if this is too defensive - if the writer is poisoned, it's because we had an error in the `Queue`, and `start_writer` will return, i.e. shut down, on any error. Maybe I should just remove that code and just leave a comment here?

lutter: Actually, this would always fail - when a writer has been poisoned, it doesn't allow adding anything more to the queue. I changed this code to just log a warning if the join handle indicates that the background writer is still running. Joining here is a bit tricky since it risks blocking indexing of that subgraph if we really do have a problem with the writer stopping.

lutter:
Rebased to latest master.

lutter:
Woops .. I screwed up, I thought this had been approved already, and merged it.

PR #4536, opened 2023-04-12
Update rust-toolchain.toml

maoueh:
I think we are safe to use any version within the `1.xx` release train.

Maybe due to a local setup problem, I had this error when trying `cargo check` after pulling latest `master` branch:

```bash
$ cargo check
error: the 'cargo' binary, normally provided by the 'cargo' component, is not applicable to the '1.68.0-aarch64-apple-darwin' toolchain
```

I tried `rustup update` but didn't change anything. Changing `rust-toolchain.toml` to use `1.68` fixed it.

Maybe a side effect of "changing" which Rust version to use (?). Anyway, I think it's quite safe to not force the usage of a particular minor version.

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "1.68.0"
+channel = "1.68"
 profile = "default"


maoueh:
The integration test failures seems unrelated to this PR.

leoyvens:
I've been able to fix that error in the past by uninstalling and reinstalling the rustup toolchain. Seems to be a rustup bug.

neysofu (changes requested):
I would suggest not removing the patch number, for the reasons described in my comment above. Upgrading to 1.69.0 would be nice, though.

PR #4540, opened 2023-04-14
core, graph, runtime: Add store.get_in_block

lutter:
This adds a new host fn `get_in_block` that will only look for entities that have been created or updated in the current block. This can improve indexing performance quite a bit if it is known that a certain entity has to have been created in the same block.

A typical situation for this is that one handler creates a `Transaction` from some on-chain event, and a later handler wants to access this transaction if it exists. In the case where the transaction does not exist, a normal `store.get` will have to go to the database just to find out that the entity does not exist; if the subgraph author already knows that the entity must have been created in the same block, using `store.get_in_block` avoids this database roundtrip. For some subgraphs, these missed lookups can contribute significantly to the indexing time.

### DIFF
--- a/runtime/wasm/src/module/mod.rs
+++ b/runtime/wasm/src/module/mod.rs
@@ -9,6 +9,7 @@ use std::time::Instant;
 
 use anyhow::anyhow;
 use anyhow::Error;
+use graph::components::store::GetScope;
 use graph::slog::SendSyncRefUnwindSafeKV;
 use never::Never;
 use semver::Version;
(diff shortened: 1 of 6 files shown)


runtime/wasm/src/module/mod.rs
```diff
@@ -535,6 +536,13 @@ impl<C: Blockchain> WasmInstance<C> {
             id,
             field
         );
+        link!(
+            "store.get_in_block",
+            store_get,
```
mangas, line 541: This should be store_get_in_block, right?

mangas (approved):
Check comment, otherwise looks good

runtime/wasm/src/module/mod.rs
lutter: Ooh .. nice catch! Will fix that

PR #4541, opened 2023-04-14
core: Log how many entries the cache had at each block

lutter:


leoyvens (approved):

leoyvens:
This looks like it could be a metric.

PR #4556, opened 2023-04-19
bump substreams client to sf.substreams.rpc.v2.Stream/Blocks

sduchesneau:
Update substreams protocol to sf.substreams.rpc.v2.Stream/Blocks
This breaks compatibility with previous substreams servers that only support 'sf.substreams.v1.Stream/Blocks'


chain/substreams/src/mapper.rs
```diff
@@ -17,89 +14,80 @@ impl SubstreamsMapper<Chain> for Mapper {
...
-            )
-        });
-
-        if outputs.is_empty() {
-            return Ok(None);
-        }
+        match message {
+            Some(SubstreamsMessage::BlockUndoSignal(undo)) => {
+                let valid_block = match undo.last_valid_block {
+                    None => return Err(SubstreamsError::MissingClockError),
```
mangas, line 22: This should prolly have a different error message?

chain/substreams/src/mapper.rs
```diff
@@ -17,89 +14,80 @@ impl SubstreamsMapper<Chain> for Mapper {
...
-                        Ok(Some(BlockStreamEvent::Revert(
-                            parent_ptr,
-                            FirehoseCursor::from(cursor.clone()),
-                        )))
-                    }
-                    StepUnknown => {
-                        panic!("unknown step should not happen in the Firehose response")
-                    }
-                }
+            None => {
```
mangas, line 86: Could you add a comment describing when this should happen?

graph/build.rs
```diff
@@ -18,4 +18,11 @@ fn main() {
         .out_dir("src/substreams")
         .compile(&["proto/substreams.proto"], &["proto"])
         .expect("Failed to compile Substreams proto(s)");
+
+    tonic_build::configure()
+        .protoc_arg("--experimental_allow_proto3_optional")
+        .extern_path(".sf.substreams.v1", "crate::substreams")
```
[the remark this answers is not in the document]
sduchesneau: no, the sf.substreams.v1 still exists for the packages, manifests and such. Only, the service layer in it has been deprecated and moved to sf.substreams.rpc.v2 (the v2 to avoid confusion, even if we change namespace)
this extern_path directive informs the rpcv2 tonic handler about where to find the sf.substreams.v1 stuff.

mangas (approved):

Issue #4544, opened 2023-04-17
Shed light into why some subgraphs spend all their time in `unknown [process]`


azf20:
For example, `lutter/marketplace` (`QmVC7pdoc58HgcfXWFzdDVLtdq8VMxntArenExZGWEF5d6`) based on [decentraland/marketplace](https://github.com/decentraland/marketplace/tree/master/indexer) is basically stalled because it's spending all its time there

azf20:
This appears to be particularly prevalent for subgraphs using substream datasources

azf20:
Unknown is significantly reduced in recent deployments of substream-based subgraphs, closing.

PR #4558, opened 2023-04-20
core: Add more stopwatch sections to runner process

mangas:
May or may not help with https://github.com/graphprotocol/graph-node/issues/4544


leoyvens (approved):
Looks great

PR #4570, opened 2023-04-24
ipfs: Remove concurrency limit, rely on rate limit

leoyvens:
The concurrency limit, because it uses a semaphore, had unfortunate interactions with the use of `call_all` in the polling monitor. `CallAll` can internally have requests which are holding the semaphore, so it must be polled regularly to check on those requests. If the task holding the `CallAll` hangs on some other future, that can lead to a deadlock.

In our case, it can hang in the `response_sender.send((id, response)).await`, if the channel is full and the subgraph runner takes a long time to check it:
https://github.com/graphprotocol/graph-node/blob/9f4a1821146b18f6f49165305e9a8c0795120fad/core/src/polling_monitor/mod.rs#L155

This could be fixed, but the footgun would remain so it seemed best to rely only on the rate limit which can't deadlock.

This also adds a log when a file is not found.

### DIFF
--- a/docs/environment-variables.md
+++ b/docs/environment-variables.md
@@ -78,8 +78,8 @@ those.
   may use (in bytes, defaults to 256MB).
 - `GRAPH_MAX_IPFS_CACHE_SIZE`: maximum number of files cached (defaults to 50).
 - `GRAPH_MAX_IPFS_CACHE_FILE_SIZE`: maximum size of each cached file (in bytes, defaults to 1MiB).
-- `GRAPH_IPFS_REQUEST_LIMIT`: Limits both concurrent and per second requests to IPFS for file data
-   sources. Defaults to 100.
+- `GRAPH_IPFS_REQUEST_LIMIT`: Limits both per second requests to IPFS for file data sources.
+   Defaults to 100.
 
 ## GraphQL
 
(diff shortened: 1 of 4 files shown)


docs/environment-variables.md
```diff
@@ -78,8 +78,8 @@ those.
   may use (in bytes, defaults to 256MB).
 - `GRAPH_MAX_IPFS_CACHE_SIZE`: maximum number of files cached (defaults to 50).
 - `GRAPH_MAX_IPFS_CACHE_FILE_SIZE`: maximum size of each cached file (in bytes, defaults to 1MiB).
-- `GRAPH_IPFS_REQUEST_LIMIT`: Limits both concurrent and per second requests to IPFS for file data
-   sources. Defaults to 100.
+- `GRAPH_IPFS_REQUEST_LIMIT`: Limits both per second requests to IPFS for file data sources.
+   Defaults to 100.
```
lutter, line 82: Maybe better: 'Limits the number of requests per second to IPFS for ...'

lutter (approved):

docs/environment-variables.md
leoyvens: Thanks!

PR #4572, opened 2023-04-25
Remove single-key changes to entities

lutter:
This PR does two big things:

* `Entity` is now guaranteed to have an `id`, as that is checked on construction
* Production code never makes changes to individual keys

Besides calrifying that entities are really just a way to shuttle data between the mappings and the database, this will make it possible to implement more efficient bulk-only changes to entities in the future.


core/src/subgraph/runner.rs
```diff
@@ -1061,11 +1062,14 @@ async fn update_proof_of_indexing(
 
         // Put this onto an entity with the same digest attribute
         // that was expected before when reading.
-        let new_poi_entity = entity! {
-            entity_cache.schema =>
-            id: entity_key.entity_id.to_string(),
-            digest: updated_proof_of_indexing,
-        }?;
+        let data = vec![
```
mangas, line 1065: This form seems a bit harder to read and seems a bit longer too. Is this to avoid changing the macro?

core/src/subgraph/runner.rs
```diff
@@ -1061,11 +1062,14 @@ async fn update_proof_of_indexing(
...
+            (
+                graph::data::store::ID.clone(),
+                Value::from(entity_key.entity_id.to_string()),
+            ),
+            (Word::from("digest"), Value::from(updated_proof_of_indexing)),
```
mangas, line 1070: digest could prolly be a const if it's used in other places as well

graph/src/data/store/mod.rs
```diff
@@ -595,6 +594,11 @@ where
     }
 }
 
+lazy_static! {
+    /// The name of the id attribute, `"id"`
+    pub static ref ID: Word = Word::from("id");
```
mangas, line 599: this could prolly be done for digest as well

runtime/wasm/src/host_exports.rs
```diff
@@ -182,6 +182,34 @@ impl<C: Blockchain> HostExports<C> {
...
+        fn check_id(key: &EntityKey, prev_id: &str) -> Result<(), anyhow::Error> {
+            if prev_id != key.entity_id.as_str() {
+                Err(anyhow!(
+                    "Value of {} attribute 'id' conflicts with ID passed to `store.set()`: \
+                {} != {}",
+                    key.entity_type,
+                    prev_id,
+                    key.entity_id,
+                ))
+            } else {
```
mangas, line 194: This else can be removed and instead do return Err...

store/postgres/src/relational_queries.rs
```diff
@@ -239,6 +239,9 @@ impl ForeignKeyClauses for Column {
 }
 
 pub trait FromEntityData: Sized {
+    /// Whether to include the internal keys `__typename` and `g$parent_id`.
+    const WITH_INTERNAL_KEYS: bool;
```
mangas, line 243: 🔥

store/postgres/src/relational_queries.rs
```diff
@@ -494,30 +500,34 @@ impl EntityData {
...
                         Some(Ok((Word::from("__typename"), T::Value::from_string(e))))
+                    } else {
+                        None
                     }
                 });
                 let entries = map.into_iter().filter_map(move |(key, json)| {
                     // Simply ignore keys that do not have an underlying table
                     // column; those will be things like the block_range that
                     // is used internally for versioning
                     if key == "g$parent_id" {
```
mangas, line 513: I know this is not from this PR but these names should defo be placed in constants, including the lazy Word const IMO

mangas (approved):
LGTM

Issue #4565, opened 2023-04-24
Add failover support for substreams providers


azf20:
Provider failover is now supported in Graph Node, but this has not been enabled for substreams providers.

PR #4578, opened 2023-04-27
graph: Add adapter balancing to substreams

mangas:
Fixes https://github.com/graphprotocol/graph-node/issues/4565

### DIFF
--- a/graph/src/blockchain/substreams_block_stream.rs
+++ b/graph/src/blockchain/substreams_block_stream.rs
@@ -1,7 +1,7 @@
 use super::block_stream::SubstreamsMapper;
+use super::client::ChainClient;
 use crate::blockchain::block_stream::{BlockStream, BlockStreamEvent};
 use crate::blockchain::Blockchain;
-use crate::firehose::FirehoseEndpoint;
 use crate::prelude::*;
 use crate::substreams::response::Message;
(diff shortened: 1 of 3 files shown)


graph/src/blockchain/substreams_block_stream.rs
```diff
@@ -186,13 +180,18 @@ fn stream_blocks<C: Blockchain, F: SubstreamsMapper<C>>(
...
 
     try_stream! {
+            let endpoint = client.firehose_endpoint()?;
+            let logger = logger.new(o!("deployment" => deployment.clone(), "provider" => endpoint.provider.to_string()));
+
         loop {
             info!(
                 &logger,
                 "Blockstreams disconnected, connecting";
                 "endpoint_uri" => format_args!("{}", endpoint),
```
lutter, line 190: Not related to this PR, but for RPC providers we stopped logging URL's because they can contain sensitive information (passwords) and mostly just log provider names. Might be a good idea for firehose endpoints, too.

lutter (approved):

graph/src/blockchain/substreams_block_stream.rs
mangas: Display has been explicitly implemented for the endpoint, only prints the provider

PR #4583, opened 2023-04-28
Write changes in batches during syncing

lutter:
With this PR, when a subgraph syncs, changes are accumulated into batches for a while (5 minutes by default) or until they reach a certain size (10MB) This should considerably speed up syncing of subgraphs whose speed is limited by the database. Batching only kicks in when the write queue has more than one entry - it turns out that holding writes for a while is pretty tricky, but for subgraphs where the bottleneck is the database, this should not matter.

Most of this PR is concerned with restructuring the code around writing entities so that those changes can be captured in a new data structure `graph::components::store::write::Batch` to which more changes can then be added.

Note that this PR sits on top of #4572 and we'll need to merge that first before merging this.

### DIFF
--- a/core/src/subgraph/inputs.rs
+++ b/core/src/subgraph/inputs.rs
@@ -26,9 +26,6 @@ pub struct IndexingInputs<C: Blockchain> {
     pub poi_version: ProofOfIndexingVersion,
     pub network: String,
 
-    // Correspondence between data source or template position in the manifest and name.
-    pub manifest_idx_and_name: Vec<(u32, String)>,
-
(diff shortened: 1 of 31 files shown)


lutter:
I just added a change to hold off on writing a batch during syncing until it is big or old enough according to the configuration. It's not clear whether that will lead to the best performance - there's some experimentation needed for that.

lutter:
This shows to have some performance problems in integration testing. I'll close this for now until I figured out what's happening there. Will create a new PR once I have solved the problems.

PR #4603, opened 2023-05-08
De-`Arc` `IpfsClient`

neysofu:
- `IpfsClient` is already `CheapClone`, so we don't need to wrap it in an `Arc` to pass is around.
- `reqwest::Client` already uses an `Arc` internally (this is not an implementation detail, but something documented and intended), so we don't need to wrap it in an `Arc`,
- Many `CheapClone` implementations can and should use the default implementation of `CheapClone::cheap_clone`, because they don't need custom behavior.

### DIFF
--- a/graph/src/ipfs_client.rs
+++ b/graph/src/ipfs_client.rs
@@ -110,6 +110,8 @@ pub struct AddResponse {
 #[derive(Clone)]
 pub struct IpfsClient {
     base: Arc<Uri>,
+    // reqwest::Client doesn't need to be `Arc` because it has one internally
+    // already.
     client: reqwest::Client,
 }
 
--- a/graph/src/ipfs_client.rs
+++ b/graph/src/ipfs_client.rs
@@ -110,7 +110,7 @@ pub struct AddResponse {
 #[derive(Clone)]
 pub struct IpfsClient {
     base: Arc<Uri>,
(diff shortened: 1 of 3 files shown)


leoyvens:
The redundant `CheapClone` implementations have a purpose, it's so that when a field is added to the struct, you are forced to think if it still is `CheapClone`, because it needs to be added to the implementation. If anything, I'd want to remove the default implementation of `fn cheap_clone` so that it always needs to be implemented manually. How do you feel about it given this consideration?

neysofu:
Oooh, I never realized. It makes so much sense and I feel dumb, eheh.

Agreed that removing the default implementation would be a smart move. Ideally we'd have a procedural macro that can `derive` it for us, but it's likely overkill.

I have removed all changes to `CheapClone` implementations in this PR :+1: should be good for another round review.

graph/src/ipfs_client.rs
```diff
@@ -110,7 +110,7 @@ pub struct AddResponse {
 #[derive(Clone)]
 pub struct IpfsClient {
     base: Arc<Uri>,
-    client: Arc<reqwest::Client>,
+    client: reqwest::Client,
```
leoyvens, line 113: I'd add a comment that this is an `Arc` internally.

leoyvens (approved):

PR #4604, opened 2023-05-08
set id field downstream when making entity change

Eduard-Voiculescu:
With recent changes done to the make entity, we have to send a create with the "id" field AND all the update entity changes also need to contain a "id" field change.

### DIFF
--- a/chain/substreams/src/trigger.rs
+++ b/chain/substreams/src/trigger.rs
@@ -20,7 +20,6 @@ use graph_runtime_wasm::module::ToAscPtr;
 use lazy_static::__Deref;
 
 use crate::codec;
-use crate::codec::value::Typed;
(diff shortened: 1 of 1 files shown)


Eduard-Voiculescu:
@mangas I have an enormous amount of abi test stuff that don't pass locally? Am I missing a file or a generate script to run?

chain/substreams/src/trigger.rs
```diff
@@ -217,6 +217,11 @@ where
                         logger,
                     );
 
+                    data.insert(
+                        Word::from("id"),
+                        decode_value(&codec::value::Typed::String(entity_id)).unwrap(),
```
lutter, line 222: you could just use `Value::from(&entity_id)` here and save a trip through `decode_value`

Eduard-Voiculescu: Good point, I made the changes.

lutter (approved):

Issue #4608, opened 2023-05-09
Don't set gas limit on Arbitrum Goerli eth_call


leoyvens:
#### Background

This workaround is necessary due to the situation of subgraphs such as [jsgphd/bracketxtest](https://thegraph.com/hosted-service/subgraph/jsgphd/bracketxtest?selected=logs) (QmU5wTEai4cyAXxAb2PNFJDHVfanzRJNW4jMLpyGjF3tNW), which is stuck with an error that looks like:
```
Trying again after eth_call RPC call for block #19141221 (b912555a9514fe9baf8f7c473fa537d8f854495fc9311f00292497911751a489) failed (attempt #9) with result Err(Web3Error(Rpc(Error { code: ServerError(-32000), message: "err: intrinsic gas too low (supplied gas 50000000)", data: None }))), provider: arbitrum-goerli-0
```
This is actually caused by high base fees on the Goerli L1, which are accounted for in the L2 gas cost, though they maybe shouldn't be. In the short term, we can disable the gas limit on this particular network so the subgraphs can make progress while we figure out a long term solution that respects determinism.

#### Solution

Have a list of EIP155 chain ids, configurable through an env var, for which no gas limit should be set. This gives us an escape hatch for this issue and will be an option for future issues related to eth_call gas costs, if they arise.

#### Implementation

The env var can be a comma-separated list decimal of numbers, corresponding to the chain ids which will not have a gas limit set. Suggestions welcome for the env var name. The default value for this env var will be the Arbitrum Goerli chain id, which is 421613.

We already store the chain `net_version`, which is equal to the `chainId`, at least for the chains we care about. We want to access that in the `RuntimeAdapter` which does the `ethereum_call`.

First a refactor is necessary to store the `ChainIdentifier` directly in the `ChainStore`, this can be done in `ChainStore::new` https://github.com/graphprotocol/graph-node/blob/0e9f517fd312a90acf6509f6a5e919646ca52bed/store/postgres/src/chain_store.rs#L1428, this `ChainIdentifier` should then be exposed through a getter method.

Then when the `RuntimeAdapter` is constructed here, the chain store is in scope so the `ChainIdentifier` can accessed from it and provided to `RuntimeAdapter::new`
https://github.com/graphprotocol/graph-node/blob/0e9f517fd312a90acf6509f6a5e919646ca52bed/node/src/main.rs#L727-L730

[...]

PR #4609, opened 2023-05-10
chain,node: Add support to skip setting gas field for certain chains

incrypto32:
Solves #4608

### DIFF
--- a/chain/ethereum/src/env.rs
+++ b/chain/ethereum/src/env.rs
@@ -83,6 +83,10 @@ pub struct EnvVars {
     /// The time to wait between polls when using polling block ingestor.
     /// The value is set in millis and the default is 1000.
     pub ingestor_polling_interval: Duration,
+    /// Set by the flag `ETHEREUM_CALL_GAS_SKIP`. The default value is empty.
+    /// This is a comma separated list of chains for which the gas field will not be set
+    /// when calling `eth_call`.
+    pub eth_call_skip_gas: Vec<String>,
 }
 
 // This does not print any values avoid accidentally leaking any sensitive env vars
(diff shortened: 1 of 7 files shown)


chain/ethereum/src/env.rs
```diff
@@ -83,6 +83,10 @@ pub struct EnvVars {
     /// The time to wait between polls when using polling block ingestor.
     /// The value is set in millis and the default is 1000.
     pub ingestor_polling_interval: Duration,
+    /// Set by the flag `ETHEREUM_CALL_GAS_SKIP`. The default value is empty.
+    /// This is a comma separated list of chains for which the gas field will not be set
```
leoyvens, line 87:
```suggestion
    /// This is a comma separated list of chain ids for which the gas field will not be set
```

chain/ethereum/src/runtime/runtime_adapter.rs
```diff
@@ -45,6 +58,11 @@ impl blockchain::RuntimeAdapter<Chain> for RuntimeAdapter {
         let call_cache = self.call_cache.cheap_clone();
         let eth_adapters = self.eth_adapters.cheap_clone();
         let archive = ds.mapping.requires_archive()?;
+        let eth_call_gas = ENV_VARS
+            .eth_call_skip_gas
+            .contains(&self.chain_identifier.net_version)
+            .then(|| None)
+            .unwrap_or(Some(ETH_CALL_GAS));
```
leoyvens, line 65: This doesn't read very well, it would be better with an `if` or `match`.

chain/ethereum/src/env.rs
```diff
@@ -171,4 +181,6 @@ struct Inner {
     genesis_block_number: u64,
     #[envconfig(from = "ETHEREUM_POLLING_INTERVAL", default = "1000")]
     ingestor_polling_interval: u64,
+    #[envconfig(from = "ETHEREUM_CALL_SKIP_GAS", default = "")]
```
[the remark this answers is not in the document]
leoyvens: Also lets include the Arbitrum Goerli chain id in the default values.

leoyvens (commented):
This looks good! I just have some minor comments.

PR #3350, opened 2022-03-11
Return non-fatal errors inside `SubgraphIndexingStatus`

neysofu:
Closes https://github.com/graphprotocol/graph-node/issues/2442.

PR #4615, opened 2023-05-12
Update nonFatalErrors in subgraphs.subgraph_deployment table

incrypto32:
Solves #3350

### DIFF
--- a/core/src/subgraph/runner.rs
+++ b/core/src/subgraph/runner.rs
@@ -478,6 +478,7 @@ where
                 persisted_data_sources,
                 deterministic_errors,
                 processed_data_sources,
+                is_non_fatal_errors_active,
             )
             .await
             .context("Failed to transact block operations")?;
--- a/store/postgres/src/deployment.rs
+++ b/store/postgres/src/deployment.rs
@@ -775,7 +775,29 @@ pub fn fail(
 ) -> Result<(), StoreError> {
     let error_id = insert_subgraph_error(conn, error)?;
 
-    update_deployment_status(conn, id, SubgraphHealth::Failed, Some(error_id))?;
+    update_deployment_status(conn, id, SubgraphHealth::Failed, Some(error_id), None)?;
+
+    Ok(())
+}
+
+pub fn update_non_fatal_errors(
+    conn: &PgConnection,
+    deployment_id: &DeploymentHash,
+    health: SubgraphHealth,
+    non_fatal_errors: Option<&[SubgraphError]>,
+) -> Result<(), StoreError> {
+    let error_ids = non_fatal_errors.map(|errors| {
+        errors
+            .iter()
+            .map(|error| {
+                hex::encode(stable_hash_legacy::utils::stable_hash::<SetHasher, _>(
+                    error,
+                ))
+            })
+            .collect::<Vec<_>>()
+    });
+
+    update_deployment_status(conn, deployment_id, health, None, error_ids)?;
 
     Ok(())
 }
@@ -802,6 +824,7 @@ pub fn update_deployment_status(
     deployment_id: &DeploymentHash,
     health: SubgraphHealth,
     fatal_error: Option<String>,
+    non_fatal_errors: Option<Vec<String>>,
 ) -> Result<(), StoreError> {
     use subgraph_deployment as d;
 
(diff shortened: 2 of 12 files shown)


neysofu (approved):

core/src/subgraph/runner.rs
```diff
@@ -495,6 +496,17 @@ where
...
+        // For subgraphs with nonFatalErrors feature enabled,
+        // We need to update the nonFatalErrors field in the subgraph_deployment table.
+        if has_errors && is_non_fatal_errors_active {
+            let errors = deterministic_errors.clone();
+            self.inputs
+                .store
+                .update_non_fatal_errors(Some(errors.to_vec()))
+                .await
+                .context("Failed to update nonFatalErrors")?;
+        }
```
lutter, line 508: This is going to cause trouble if `graph-node` gets stopped in between calling `transact_block_operations` and this logic here. All changes for a block need to be made through the call to `transact_block_operations`, and they need to happen in one database txn to make sure the db state is always consistent (either all changes for a block get written or not written)

lutter (changes requested):
It might be best to wait with this until I have merged #4606 since that introduces a lot of changes to how things get written.

core/src/subgraph/runner.rs
incrypto32: what would you suggest here? I could add a bool parameter `non_fatal_errors_enabled`  to `transact_block_operations` and then run it inside there if that sounds good

core/src/subgraph/runner.rs
```diff
@@ -468,6 +468,7 @@ where
         } = block_state;
 
         let first_error = deterministic_errors.first().cloned();
+        let deterministic_errors = Arc::new(deterministic_errors);
```
lutter, line 471: You shouldn't put this into an `Arc`; that way, you don't have to call `clone().to_vec()` below and can just pass `deterministic_errors` into `transact_block_operations`

store/postgres/src/deployment_store.rs
```diff
@@ -1118,6 +1118,16 @@ impl DeploymentStore {
                     )?;
                 }
 
+                if !batch.deterministic_errors.is_empty() && batch.is_non_fatal_errors_active {
```
lutter, line 1121: It would be better to pull this `if` into the previous `if` statement so that you only have to say `if batch.is_non_fatal_errors_active { .. }`

store/postgres/src/deployment.rs
```diff
@@ -775,7 +775,29 @@ pub fn fail(
...
+    update_deployment_status(conn, id, SubgraphHealth::Failed, Some(error_id), None)?;
+
+    Ok(())
+}
+
+pub fn update_non_fatal_errors(
+    conn: &PgConnection,
+    deployment_id: &DeploymentHash,
+    health: SubgraphHealth,
+    non_fatal_errors: Option<Vec<SubgraphError>>,
```
lutter, line 787: If you make this an `Option<&[SubgraphError]>` the caller doesn't need to clone - it's not a big deal in this case since errors happen rarely, but on hotter code paths it would make a difference

lutter (approved):
I left a few stylistic comments that would be good to address before merging, but other than that this looks good.

It would also be good to have a test that checks that the non-fatal errors on a deployment get set now when that should happen (it's enough if this is just a test within the store) so we don't accidentally regress.

One thing I am not clear on is whether `fail_subgraph` still needs to also store the error - it seems like that is just additional work now?

core/src/subgraph/runner.rs
incrypto32: got it, thanks for pointing it out, did so since previously non_fatal_errors were being written in the same function and deterministic_errors get moved, now we do that in `transact_block_operations` so its no longer required to be Arc

incrypto32:
> One thing I am not clear on is whether `fail_subgraph` still needs to also store the error - it seems like that is just additional work now?

@lutter in this case `fail_subgraph` wont get called because the subgraph progresses, but i believe `fail_subgraph` always did extra work since all errors are being written inside `transact_block_operations` when `insert_subgraph_errors` is being called. 
What `fail_subgraph` should ideally be doing is to just calculate the error_id (instead of inserting the error and getting the id back) and then updating this `subgraph_deployment` table. That is what i have done in `update_non_fatal_errors`. 

As you said since it works and ignores the duplicate insert silenty, we should not touch `fail_subgraph` anyways atleast for nows?

incrypto32:
@lutter i have added tests, transact some errors and then check that status to check if nonFatalErrors array has something in there.

Also squashed the commits into two.

lutter (approved):
Looks great! Thanks for adding the tests.

Issue #4325, opened 2023-01-27
Rules-based system to control index creation and similar attributes


lutter:
There are a few things that can be set manually through `graphman` or directly by manipulating the database that affect how subgraph data is stored, and can help query performance. Those items include:

* custom indexes (multicolumn indexes or partial indexes)
* creation of default attribute indexes (a lot of them are never used)
* account-like optimization
* statistics target

An example of what that could like like in YAML form is
```yaml
# Sketch of what an index definition file could look like. The file will
# contain many entries like this one
- rule: Messari analytics subgraphs
  match:
    # All subgraphs whose name matches this regex will be treated according
    # to these actions
    name: messari/.*
  actions:
    - entity:
        name: LiquidityPoolSnapshot
        index:
          - on: [pool, timestamp]
    - entity:
        name: LiquidityPool
        # Turn off attribute indexes
        default_indexes: false
        # Set the table to account-like
        account-like: true
        # Increase the statistics target
        stats-target: 300
        # Create some custom indexes
        index:
          # Use a GiST index since we use block_range
          - on: [block_range, timestamp, id]
            method: gist
          # A partial index that only covers blocks after a certain point
          # TODO: since the block number means different things on
          # different networks, we should either use a timespan or let users
          # match on network
          - on: [timestamp, id]
            after: 15664000
          - on: [pool_address, block_range, timestamp]
            method: gist
```

We need to think more about the way in which operators interact with these rules. Changing rules should not require a restart of `graph-node`. It should also be possible to list rules, delete them etc. Rules should be applied when a subgraph is deployed, but there should also be tooling to apply a new rule to an existing deployment.

leoyvens:
Indexes as configuration are a fascinating idea!

Matching on subgraph name regex is still whack-a-mole because similar subgraphs get deployed under different names or orgs for many reasons. I'd want to go further and, by default, match the indexes on entities rather than on subgraphs. This catches subgraph forks under any name. And given the push for subgraph schema standardization by Messari, matching on entities would benefit all subgraphs using standard schemas.

Since entities are defined in GraphQL, that might be a good language for specifying the indexes. [Prisma indexes](https://www.prisma.io/docs/concepts/components/prisma-schema/indexes) offer prior art on using a graphql-y syntax for this. Something like:
```graphql
type LiquidityPool
  @entity
  @flags(defaultIndexes: false, accountLike: true)
  @index(fields: ["block_range", "pool", "timestamp"], method: gist)
[...]

PR #4622, opened 2023-05-15
node: Add perf config file support

mangas:
First iteration of https://github.com/graphprotocol/graph-node/issues/4325's implementation. 

Some unsolved issues in MVP:
- Rules are read in random order 
- If multiple rules apply to the same subgraph one of them is selected
- No tool to check how the rules apply to existing subgraphs. 
- Currently history_blocks cannot be replaced via config, only applied to new deployments

### DIFF
--- a/core/src/subgraph/config.rs
+++ b/core/src/subgraph/config.rs
@@ -1,0 +1,26 @@
+use graph::prelude::{regex::Regex, SubgraphName};
+
+#[derive(Default, Debug)]
+pub struct SubgraphPerfRules {
+    pub rules: Vec<(Regex, SubgraphPerfConfig)>,
+}
+
+impl SubgraphPerfRules {
+    pub fn config_for_name(&self, name: &SubgraphName) -> Option<SubgraphPerfConfig> {
+        self.rules.iter().find_map(|(regex, config)| {
+            if regex.is_match(name.as_str()) {
+                Some(config.clone())
+            } else {
+                None
+            }
+        })
+    }
+}
+
+#[derive(Default, Debug, Clone)]
+// TODO: Remove clone and use Arc or something on the RulesStructure
+// Make this struct combinable by setting relevant priority per field,
+// eg prefer shorter history_blocks etc
+pub struct SubgraphPerfConfig {
+    pub history_blocks: i32,
+}
--- a/core/src/subgraph/registrar.rs
+++ b/core/src/subgraph/registrar.rs
@@ -13,6 +13,8 @@ use graph::prelude::{
     SubgraphRegistrar as SubgraphRegistrarTrait, *,
 };
 
+use crate::SubgraphPerfRules;
+
 pub struct SubgraphRegistrar<P, S, SM> {
     logger: Logger,
     logger_factory: LoggerFactory,
@@ -24,6 +26,7 @@ pub struct SubgraphRegistrar<P, S, SM> {
     node_id: NodeId,
     version_switching_mode: SubgraphVersionSwitchingMode,
     assignment_event_stream_cancel_guard: CancelGuard, // cancels on drop
+    perf_config: Arc<SubgraphPerfRules>,
 }
 
 impl<P, S, SM> SubgraphRegistrar<P, S, SM>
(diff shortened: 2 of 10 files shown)


core/src/subgraph/config.rs
```diff
@@ -0,0 +1,26 @@
...
+#[derive(Default, Debug)]
+pub struct SubgraphPerfRules {
+    pub rules: Vec<(Regex, SubgraphPerfConfig)>,
+}
+
+impl SubgraphPerfRules {
+    pub fn config_for_name(&self, name: &SubgraphName) -> Option<SubgraphPerfConfig> {
+        self.rules.iter().find_map(|(regex, config)| {
+            if regex.is_match(name.as_str()) {
+                Some(config.clone())
```
lutter, line 12: If this method returns a `Option<&SubgraphPerfConfig>`, we don't need to clone; the caller can always decide whether they need a clone or not.

node/src/main.rs
```diff
@@ -146,6 +148,17 @@ async fn main() {
         std::process::exit(0);
     }
 
+    let subgraph_config: SubgraphPerfRules = match env_vars.performance_config_file {
+        Some(ref path) => {
+            info!(logger, "Reading subgraph configuration file `{}`", path);
+            toml::from_str::<SubgraphConfigSection>(&read_to_string(path).unwrap())
+                .unwrap()
+                .try_into()
+                .unwrap()
```
lutter, line 157: This would be better as a method on `SubgraphConfigSection` - `main` doesn't need to know that this is TOML.

Also, there's a lot of unwrapping here. What do error messages look like when there's a syntax error in the file or when the file doesn't even exist? It's important that the error messages give users enough information to fix their mistakes.

core/src/subgraph/registrar.rs
```diff
@@ -24,6 +26,7 @@ pub struct SubgraphRegistrar<P, S, SM> {
     node_id: NodeId,
     version_switching_mode: SubgraphVersionSwitchingMode,
     assignment_event_stream_cancel_guard: CancelGuard, // cancels on drop
+    perf_config: Arc<SubgraphPerfRules>,
```
lutter, line 29: Does this need to be an `Arc`? It seems to me the registrar could just take ownership of the rules

node/src/perf_config.rs
```diff
@@ -0,0 +1,96 @@
...
+mod test {
+    use crate::perf_config::MatchSelection;
+
+    #[test]
+    fn parses_correctly() {
+        let content = r#"
+        [rules]
+
+        [rules.name] 
+        match = { type = "subgraph_name", name = ".*" }
```
lutter, line 67: I don't understand why we need both `type` and `name` here. That seems a bit cumbersome and just having `match { name = "foo.*" }` should be enough.

lutter (changes requested):

lutter:
I think the most important thing to fix before we can use this are the first two points you mention - evaluating rules in random order will be very confusing, and can lead to cases where restarts change what will effectively be applied.

mangas:
> I think the most important thing to fix before we can use this are the first two points you mention - evaluating rules in random order will be very confusing, and can lead to cases where restarts change what will effectively be applied.

IMO, this is not something to fix right now, we can still release it and test it. Only applies to new deployments anyway. I was thinking the fix for that could be that rules have priorities, in this case, prefer higher values of history_blocks so you could have a catch-all X and a specific rule Y, when evaluating the rules, the higher value gets priority. On other properties later it could be different. 

Evaluating in a specific order would probably require a change in syntax from dictionary to array which I'm not really sure would be for the best. Regardless, after fixing the comments IMO we could still use this as a trial and add more features as we go.

node/src/perf_config.rs
mangas: I was thinking later we could add an owner match instead of just name

node/src/perf_config.rs
```diff
@@ -0,0 +1,96 @@
+use std::collections::HashMap;
+
+use graph::{anyhow, prelude::regex::Regex};
+use graph_core::{SubgraphPerfConfig, SubgraphPerfRules};
+use serde::{Deserialize, Serialize};
+
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub struct SubgraphConfigSection {
+    rules: HashMap<String, SubgraphConfigRule>,
```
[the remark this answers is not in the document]
mangas: I don't think that will help much since the main issue is that usually people would prolly be expecting a first match and a catch-all rule. Internally, the consistent ordering is not super important (I think).

neysofu:
Is this PR made obsolete by #4622? Should we close it?

PR #4624, opened 2023-05-17
perf: Avoid cloning cached entities

leoyvens:
This should be a measurable performance improvement. And the lifetime adjustments surprisingly didn't cause much trouble. The only case where a clone is still required is if the entity needs to be mutated by applying pending updates, so we use a `Cow` to handle that. The entity sorting step is also faster now, because it no longer requires cloning the entity keys, rather we just sort the interned values.


lutter (approved):
Nice! Getting rid of that clone should help a lot!

leoyvens:
Integration cluster likes it, merging

PR #4625, opened 2023-05-17
Fix oneshot cancelled errors

### LINKED ISSUE #2797: Improve error message for runtime panics (`oneshot canceled`)
Nowadays when the `runtime` code panics before the mappings code, we save something like this to the `SubgraphErrors` table:

```
Subgraph instance failed to run: failed to process trigger: block #15730765 (0x3fc4…457f), transaction f7fd51db1b13c53b571d767503af8d6ad946708fe8ff67de8bceab3dd30f69d6: Mapping terminated before handling trigger: oneshot canceled, code: SubgraphSyncingFailure
```

The message `Mapping terminated before handling trigger: oneshot canceled` is actually an `anyhow` context ([that happens here](https://github.com/graphprotocol/graph-node/blob/master/runtime/wasm/src/host.rs#L195)) of `Mapping terminated before handling trigger` with `oneshot canceled`, which is the panic from the channel.

Maybe look [here](https://github.com/graphprotocol/graph-node/blob/master/runtime/wasm/src/mapping.rs#L54).

- [ ] Investigation on where exactly we're losing the error message
- [ ] Implement the solution

incrypto32:
Fixes #2797

### DIFF
--- a/runtime/wasm/src/mapping.rs
+++ b/runtime/wasm/src/mapping.rs
@@ -9,8 +9,9 @@ use graph::data_source::{MappingTrigger, TriggerWithHandler};
 use graph::prelude::*;
 use graph::runtime::gas::Gas;
 use std::collections::BTreeMap;
+use std::panic::AssertUnwindSafe;
 use std::sync::Arc;
-use std::thread;
+use std::{panic, thread};
 
 /// Spawn a wasm module in its own thread.
 pub fn spawn_module<C: Blockchain>(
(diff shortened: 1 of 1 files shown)


runtime/wasm/src/mapping.rs
```diff
@@ -52,14 +53,32 @@ where
...
+                    );
+
+                    panic!("TEST PANIC");
+                    result
+                }));
+
+                let result = match result {
+                    Ok(result) => result,
+                    Err(panic_info) => {
+                        let err_msg = if let Some(payload) = panic_info.downcast_ref::<&str>() {
```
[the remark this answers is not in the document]
incrypto32: gotcha

runtime/wasm/src/mapping.rs
```diff
@@ -52,14 +53,33 @@ where
...
+                        let err_msg = if let Some(payload) = panic_info
+                            .downcast_ref::<String>()
+                            .map(String::as_str)
+                            .or(panic_info.downcast_ref::<&str>().copied())
+                        {
+                            anyhow!("Subgraph panicked with message: {}", payload)
+                        } else {
+                            anyhow!("Subgraph panicked with an unknown payload.")
+                        };
+                        println!("Subgraph panicked with message: {}", err_msg);
```
leoyvens, line 79: Usually we don't want `println!` in the code, we could turn this into a log but it seems just redundant with the error, lets remove this.

leoyvens (changes requested):

runtime/wasm/src/mapping.rs
incrypto32: Got it, will keep that in mind going forward, addressed the changes

leoyvens (approved):

PR #4628, opened 2023-05-17
store: Do not create GIN indexes for numeric arrays

lutter:
GIN indexes are pretty expensive to build and maintain. For fields of types like `[BigInt!]!` or `[BigDecimal!]!`, they also don't help much since [they only support](https://www.postgresql.org/docs/current/gin-builtin-opclasses.html) looking for exact values, which is very unlikely to happen in a query, and even less unlikely to be the filter that the query planner would use.


leoyvens (approved):
I'm also suspicious of whether GIN indexes carry their weight.

PR #4631, opened 2023-05-19
Optimize match and decode, turn on static filters

leoyvens:
We saw great effect from static filters on uniswap, stabilizing the firehose connection and the overall throughput of the subgraph. But they do cause a large number of false positive triggers to be received, resulting in a lot of time being spent by `fn match_and_decode` to discard those false positives.

Right now the total number of calls to `match_and_decode` would be: `n_triggers * n_data_sources`. This PR optimizes trigger matching by introducing a hash set of addresses to filter out triggers which are certain not to match any data sources, and this requires only a hash operation for each trigger.

The second commit turns on static filters for any subgraph with more than 10k data sources, under the assumption that with this optimization static filters will be an improvement for most subgraphs.

### DIFF
--- a/core/src/subgraph/context/instance.rs
+++ b/core/src/subgraph/context/instance.rs
@@ -7,7 +7,7 @@ use graph::{
     },
     prelude::*,
 };
-use std::collections::{HashMap, HashSet};
+use std::collections::HashMap;
 
 use super::OffchainMonitor;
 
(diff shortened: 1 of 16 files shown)


core/src/subgraph/context/instance.rs
```diff
@@ -111,6 +116,11 @@ where
                 sender
             }
         };
+
+        if let Some(address) = data_source.address() {
+            self.host_addresses.insert(address.into());
```
lutter, line 121: I am not sure if I am missing something, but I don't see `self.host_addresses` used anywhere to check if an address is there or not.

lutter (approved):
Looks good! I am not sure if `SubgraphInstance.host_addresses` is actually used.

core/src/subgraph/context/instance.rs
leoyvens: Oh wow seems I completely lost track of where I was going with this 😅. It's not only the AIs that can hallucinate. I'll rework this to actually use the hashset.

core/src/subgraph/context/instance.rs
```diff
@@ -204,11 +202,121 @@ where
...
+
+    fn pop(&mut self) {
+        let Some(host) = self.hosts.pop() else { return };
+        let address = host.data_source().address();
+        match address {
+            Some(address) => {
+                // Unwrap and assert: The same host we just popped must be the last one in `hosts_by_address`.
+                let hosts = self.hosts_by_address.get_mut(address.as_slice()).unwrap();
+                let idx = hosts.pop().unwrap();
+                assert_eq!(idx, self.hosts.len());
```
lutter, line 286: Is this running inside some panic handler that will stop the subgraph if this assert gets tripped? Otherwise, it would be better to return an error so that only this subgraph stops, and others can continue

lutter (approved):
LGTM! Just one comment on how asserts will impact indexing other subgraphs

core/src/subgraph/context/instance.rs
leoyvens: Thanks for the review, I've checked and this panic will only stop the runner thread for the subgraph that trips it.

leoyvens:
Integration cluster failures look unrelated, merging.

PR #4633, opened 2023-05-20
Add subgraph-specific settings

lutter:
This PR achieves a very similar result as #4622 but with a few important differences:

* rule evaluation is predictable as rules are always evaluated in the order in which they are listed in the settings file
* the command `graphman config check` will check the settings file for validity if one is set in the environment
* the command `graphman config setting <name>` will print what setting will be used when deploying a new subgraph `name`

The syntax of the file is also slightly different to streamline it a bit:
```toml
[[setting]]
match = { name = "foo/.*" }
history_blocks = 100

[[setting]]
match = { name = ".*/bar" }
history_blocks = 200
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1580,6 +1580,7 @@ dependencies = [
  "serde_derive",
  "serde_json",
  "serde_plain",
+ "serde_regex",
  "serde_yaml",
  "slog",
  "slog-async",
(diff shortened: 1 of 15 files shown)


leoyvens (approved):

leoyvens:
This lgtm, for subgraph-wide configs. For configuring entity/table level configs such as indexes, I'd still want us to think about orienting around matching the tables directly rather than subgraph names, but that's just a vague thought for now.

PR #4635, opened 2023-05-22
Add MappingTrigger.error_context()  to error logs

### LINKED ISSUE #3694: Add transaction hash to SubgraphError and error message when available

feature request

For a runtime null-pointer error, it looks like the below:

```
Mapping aborted at src/common/creators.ts, line 162, column 19, with message: unexpected null wasm backtrace: 0: 0x80cd - <unknown>!src/common/creators/createWithdraw 1: 0x833e - <unknown>!src/mappings/pool/handleBurn in handler `handleBurn` at block #11014326 (69e39fcb23c85c462f832a35e84f506424a97538721993451872cada7cbc6089)
```

While it shows the block number and the block hash, it has no info about the relevant transaction, which is super helpful for debugging the issue.

Transaction hash should be included in the error message

```
Mapping aborted at src/common/creators.ts, line 162, column 19, with message: unexpected null wasm backtrace: 0: 0x80cd - <unknown>!src/common/creators/createWithdraw 1: 0x833e - <unknown>!src/mappings/pool/handleBurn in handler `handleBurn` at block #11014326 (69e39fcb23c85c462f832a35e84f506424a97538721993451872cada7cbc6089) at transaction xxxxxx
```

Also, `SubgraphError` entity in the health endpoint https://api.thegraph.com/index-node/graphql should include transaction.

```graphql
type SubgraphError {
  message: String!
  transaction: Transaction
  block: Block
  handler: String
  deterministic: Boolean!
}

type Transaction {
  hash: String!
  // ...
}
```

### LINKED ISSUE #4437: Add transaction hash to error logs
Currently we provide block number and block hash when logging errors, but this can make it hard to find the event in blocks with lots of transactions.

```
2023-03-09T10:39:02.044376451Z stderr F Mar 09 10:39:02.044 ERRO Subgraph error 1/1, code: SubgraphSyncingFailure, error: Mapping aborted at src/mappings/poolController.ts, line 396, column 14, with message: unexpected null	wasm backtrace:	    0: 0x5310 - <unknown>!src/mappings/poolController/handleTransfer	 in handler `handleTransfer` at block #40095408 (d668119485ad07dd601c413cb6104eea3e6e43c6ce292399bdd728c2664ad90e), block_hash: 0xd668119485ad07dd601c413cb6104eea3e6e43c6ce292399bdd728c2664ad90e, block_number: 40095408, sgd: 513442, subgraph_id: QmXsooSgsCT4AbWvhyaAwu2T4F7js7avHr8buxriiyV7J3, component: SubgraphInstanceManager
```

incrypto32:
Closes #4437 , Closes #3694

This PR allows details like transaction hash to be added to an `error_context()` method on `MappingTrigger`  which can then be logged or returned with `SubgraphError`

### DIFF
--- a/runtime/wasm/src/module/mod.rs
+++ b/runtime/wasm/src/module/mod.rs
@@ -207,8 +207,10 @@ impl<C: Blockchain> WasmInstance<C> {
         let handler_name = trigger.handler_name().to_owned();
         let gas = self.gas.clone();
         let logging_extras = trigger.logging_extras().cheap_clone();
+        let error_context = trigger.trigger.error_context();
         let asc_trigger = trigger.to_asc_ptr(self.instance_ctx_mut().deref_mut(), &gas)?;
-        self.invoke_handler(&handler_name, asc_trigger, logging_extras)
+
+        self.invoke_handler(&handler_name, asc_trigger, logging_extras, error_context)
     }
 
     pub fn take_ctx(&mut self) -> WasmInstanceContext<C> {
@@ -238,6 +240,7 @@ impl<C: Blockchain> WasmInstance<C> {
         handler: &str,
         arg: AscPtr<T>,
         logging_extras: Arc<dyn SendSyncRefUnwindSafeKV>,
+        error_context: String,
     ) -> Result<(BlockState<C>, Gas), MappingError> {
         let func = self
             .instance
@@ -287,7 +290,8 @@ impl<C: Blockchain> WasmInstance<C> {
         };
 
         if let Some(deterministic_error) = deterministic_error {
-            let message = format!("{:#}", deterministic_error).replace('\n', "\t");
+            let message =
+                format!("{:#}, {}", deterministic_error, error_context).replace('\n', "\t");
 
             // Log the error and restore the updates snapshot, effectively reverting the handler.
             error!(&self.instance_ctx().ctx.logger,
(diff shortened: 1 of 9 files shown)


runtime/wasm/src/module/mod.rs
```diff
@@ -207,8 +207,10 @@ impl<C: Blockchain> WasmInstance<C> {
         let handler_name = trigger.handler_name().to_owned();
         let gas = self.gas.clone();
         let logging_extras = trigger.logging_extras().cheap_clone();
+        let error_context = trigger.trigger.error_context();
```
lutter, line 210: This will create a `String` every time a handler is invoked, which is a lot; we should try to avoid this allocation overhead on every invocation. 

One way to do that is to pass a `&dyn MappingTriggerTrait` into `invoke_handler` instead of the `error_context: Option<String>` and then call `error_context()` only if an error actually happens.

lutter (changes requested):

leoyvens:
I also gave this a look, it lgtm

runtime/wasm/src/module/mod.rs
lutter: Looking some more into this, the fact that `to_asc_ptr` takes ownership of the trigger makes what I said above impossible; it's not clear to me that `to_asc_ptr` has to take ownership, but changing that would be quite a bit more work. So I think we should go with this as-is, and plan on optimizing this later on.

lutter (approved):

PR #4642, opened 2023-05-24
Add ability to pause and resume subgraphs

### LINKED ISSUE #4255: Support pause/resume of subgraphs
Currently, subgraphs get paused and resumed by exploiting a side-effect of how `graph-node` handles assignments: assigning to a nonexistent node pauses a subgraph and assigning it back to a real node resumes it. That is clunky UX and makes it impossible for `graph-node` to decide what to do when a paused subgraph is deployed again (as in issue #4242)

A better design is to make `graph-node` aware of pausing/resuming subgraphs. To this end, we should make the following changes:

* add commands `graphman pause`/`graphman resume`
* change the assignment messages we send internally to make sure subgraphs stop indexing when paused and start when resumed (this can probably happen with the existing add/remove messages)
* have index nodes ignore paused subgraphs when they start up

Backing this would require a small schema change in `subgraph_deployment_assignment` to add a column `paused_at timstamptz` where a `null` value means 'the subgraph should be indexed'. While we're at it, we should also add a column `assigned_at` - it's not really needed for this functionality, but can be useful for auditing.

In addition, the `subgraph_deploy` request should get a new parameter that controls whether redeploying an existing subgraph will resume it if it exists and is paused.

incrypto32:
Closes #4255 

- [x] Schema changes to `subgraph_deployment_assignment` to include `paused_at` and `assigned_at`
- [x] Add pause_subgraph and  resume_subgraph methods to store
- [x] Graphman commands `pause` and `resume` 
- [x] Send StoreEvents `EntityChangeOperation::Set` and `EntityChangeOperation::Remove` 
- [x] Ignore paused subgraphs on startup

### DIFF
--- a/graph/src/components/store/traits.rs
+++ b/graph/src/components/store/traits.rs
@@ -106,6 +106,9 @@ pub trait SubgraphStore: Send + Sync + 'static {
 
     fn assignments(&self, node: &NodeId) -> Result<Vec<DeploymentLocator>, StoreError>;
 
+    /// Returns assignments that are not paused
+    fn active_assignments(&self, node: &NodeId) -> Result<Vec<DeploymentLocator>, StoreError>;
+
     /// Return `true` if a subgraph `name` exists, regardless of whether the
     /// subgraph has any deployments attached to it
     fn subgraph_exists(&self, name: &SubgraphName) -> Result<bool, StoreError>;
--- a/graph/src/components/store/traits.rs
+++ b/graph/src/components/store/traits.rs
@@ -95,6 +95,15 @@ pub trait SubgraphStore: Send + Sync + 'static {
 
     fn assigned_node(&self, deployment: &DeploymentLocator) -> Result<Option<NodeId>, StoreError>;
 
+    /// Returns Option<(node_id,is_paused)> where `node_id` is the node that
+    /// the subgraph is assigned to, and `is_paused` is true if the
+    /// subgraph is paused.
+    /// Returns None if the deployment does not exist.
+    fn assignment_status(
+        &self,
+        deployment: &DeploymentLocator,
+    ) -> Result<Option<(NodeId, bool)>, StoreError>;
+
     fn assignments(&self, node: &NodeId) -> Result<Vec<DeploymentLocator>, StoreError>;
 
     /// Return `true` if a subgraph `name` exists, regardless of whether the
--- a/node/src/manager/commands/assign.rs
+++ b/node/src/manager/commands/assign.rs
@@ -85,15 +85,23 @@ pub fn pause_or_resume(
         .locate_site(locator.clone())?
         .ok_or_else(|| anyhow!("failed to locate site for {locator}"))?;
 
-    if pause {
-        println!("pausing {locator}");
-        conn.pause_subgraph(&site)?;
-        println!("paused {locator}")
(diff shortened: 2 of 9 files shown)


store/postgres/src/primary.rs
```diff
@@ -981,6 +981,44 @@ impl<'a> Connection<'a> {
...
+        use subgraph_deployment_assignment as a;
+
+        let conn = self.conn.as_ref();
+
+        let updates = update(a::table.filter(a::id.eq(site.id)))
+            .set(a::paused_at.eq(sql("now()")))
+            .execute(conn)?;
+        match updates {
+            0 => Err(StoreError::DeploymentNotFound(site.deployment.to_string())),
+            1 => Ok(()),
```
lutter: If you do what `unassign_subgraph` does here, namely
```
let change =
                    EntityChange::for_assignment(site.into(), EntityChangeOperation::Removed);
                Ok(vec![change])
```

things should just work

node/src/manager/commands/assign.rs
```diff
@@ -69,3 +69,31 @@ pub fn reassign(
...
+
+    if pause {
+        println!("pausing {locator}");
+        conn.pause_subgraph(&site)?;
+        println!("paused {locator}")
+    } else {
+        println!("resuming {locator}");
+        conn.resume_subgraph(&site)?;
+        println!("resumed {locator}")
+    }
```
lutter: Assign the return value from `pause_subgraph`/`resume_subgraph` to a variable here and call `conn.send_store_event` similarly to what `reassign` in this file does.

store/postgres/src/primary.rs
```diff
@@ -626,6 +627,22 @@ mod queries {
             .transpose()
     }
 
+    pub(super) fn assignment_status(
+        conn: &PgConnection,
+        site: &Site,
+    ) -> Result<Option<(String, bool)>, StoreError> {
```
lutter, line 633: It would be clearer if this returned a `Option<(NodeId, bool)>`; if the conversion from `String` to `NodeId` fails, you can just return a `constraint_violation!` since that shouldn't happen.

graph/src/components/store/traits.rs
```diff
@@ -95,6 +95,11 @@ pub trait SubgraphStore: Send + Sync + 'static {
 
     fn assigned_node(&self, deployment: &DeploymentLocator) -> Result<Option<NodeId>, StoreError>;
 
+    fn assignment_status(
+        &self,
+        deployment: &DeploymentLocator,
+    ) -> Result<Option<(String, bool)>, StoreError>;
```
lutter, line 101: This method should have a comment explaining what the return values mean

lutter (commented):
Looks good so far, some small comments

lutter (approved):
Nice, this will be very useful

paymog:
if I'm understanding this PR correctly, graphman directly interacts with postgres when pausing/resuming. Is that correct? If so, it would be great if this functionality was also exposed as an [endpoint](https://github.com/graphprotocol/graph-node/blob/ce693d8893e9fa7cd5b17bdeffcae5a100b6cdb9/server/json-rpc/src/lib.rs#L50-L70) on the graph node itself so that other applications could use this feature.

Issue #4482, opened 2023-03-20
[Feature] Expose `history_blocks` setting in index-node API


lutter:
### Description

The `history_blocks` setting which governs/will govern how much history a subgraph retains should be exposed through the index-node API.

### Are you aware of any blockers that must be resolved before implementing this feature? If so, which? Link to any relevant GitHub issues.

Merge #4429 

### Some information to help us out

- [ ] Tick this box if you plan on implementing this feature yourself.
- [X] I have searched the issue tracker to make sure this issue is not a duplicate.

PR #4662, opened 2023-05-31
graph, store - expose history_blocks in index-node API

incrypto32:
Closes [#4482](https://github.com/graphprotocol/graph-node/issues/4482)


lutter (approved):
Nice! Looks great!

PR #4675, opened 2023-06-06
build: minify html as a pre-build step

YassinEldeeb:
The HTML file for `graphiql` once lands the repository, it's never touched.

So this adds a `build.rs` script to minify that `index.html` file to be served faster in the `http` sever for GraphQL.

Cargo will only run the minification when the file is touched, otherwise it'd skip it.


neysofu:
Hi @YassinEldeeb,
This is pretty cool, but I'm thinking the size savings are not necessarily worth (1) the extra compilation time during clean builds and (2) the maintenance burden, as small as it is, of a custom `build.rs`.

How about we update this PR and minify it manually?

YassinEldeeb:
@neysofu Yeah, for sure! These are good arguments, I can switch it as a `bin` that's ran as a script whenever we need to do a minification manually, what do you think?

leoyvens:
@YassinEldeeb That sounds good, if you want to pick this back up.

azf20:
@YassinEldeeb checking on this one?

PR #4676, opened 2023-06-06
graphql: update introspection schema

saihaj:
Updating the introspection schema as per October 2021 spec update - http://spec.graphql.org/October2021/#sec-Schema-Introspection.Schema-Introspection-Schema


graph/src/schema/introspection.graphql
```diff
@@ -73,7 +79,8 @@ type __Directive {
   name: String!
   description: String
   locations: [__DirectiveLocation!]!
-  args: [__InputValue!]!
+  args(includeDeprecated: Boolean = false): [__InputValue!]!
+  isRepeatable: Boolean!
```
dotansimha, line 83: Do we need to change anything in the Rust code to implement these new fields?

saihaj: I think it should work. We do seem to even have a test for it https://github.com/graphprotocol/graph-node/blob/363f4c52955a06acaf17bd02c04baafcd36c5e40/store/test-store/tests/graphql/introspection.rs#L587-L837

dotansimha (approved):

leoyvens:
Do the tests cover the added fields? We should make sure that they do, particularly because non-nullable fields were added.

leoyvens:
@saihaj didn't know that, is that in the graphql spec somewhere?

lutter:
I just added a commit to this PR that fixes the test failures.

When we execute a query, the schema we use for that is a combination of the subgraph schema and the introspection schema. When we mix the introspection schema into the subgraph schema, we add the two introspection fields to the subgraph schema's `Query` type. Before this PR, the introspection schema did not have a `Query` type; by adding a `Query` type to the introspection schema, the the subgraph schema's `Query` type gets replaced with the introspection schema's `Query` type, effectively erasing all the root query fields of the subgraph schema.

PR #4679, opened 2023-06-07
Add new table to track subgraph features

### LINKED ISSUE #4542: Track deployment features & data source types
It is difficult to identify which live deployments are using which types of subgraph features and data source types, making it challenging to monitor feature usage and performance.

These could be stored on as metadata on the deployment. This could be stored in the primary, to enable usage in sharded setups.

Relevant metadata:
- ApiVersion and SpecVersion
- Data source types (ethereum/contract, file/ipfs, substreams, templates)
- Features under `features` in `subgraph.yaml`

For existing deployments a migration / one-off will be required to populate this data

incrypto32:
Closes #4542


store/postgres/src/primary.rs
```diff
@@ -1098,6 +1108,30 @@ impl<'a> Connection<'a> {
...
+        let conn = self.conn.as_ref();
+        insert_into(f::table)
+            .values((
+                f::id.eq(id),
+                f::spec_version.eq(spec_version),
+                f::api_versions.eq(api_versions),
+                f::features.eq(features),
+                f::data_sources.eq(data_source_kinds),
+            ))
+            .on_conflict_do_nothing()
```
leoyvens, line 1130: Lets overwrite on conflict:
```
    .on_conflict(f::id)
    .do_update()
```

store/postgres/src/primary.rs
```diff
@@ -78,6 +78,16 @@ table! {
...
 }
 
+table! {
+    subgraphs.subgraph_features (id) {
+        id -> Text,
+        spec_version -> Text,
+        api_versions -> Array<Text>,
+        features -> Array<Text>,
+        data_sources -> Array<Text>,
+    }
```
leoyvens, line 88: I like this schema a lot, with the string arrays, a good balance between structure and flexibility.

I'd make `api_version` a nullable `Text`, which should be the `unified_api_version` or `null` if the subgraph is legacy and has mixed versions.

incrypto32:
@leoyvens  Ah forgot, one more thing, will add tests and request a re-review.

runtime/wasm/src/module/mod.rs
```diff
@@ -1170,6 +1170,15 @@ impl<C: Blockchain> WasmInstanceContext<C> {
...
         let field: String = asc_get(self, field_ptr, gas)?;
+
+        error!(
+            &self.ctx.logger,
+            "store_load_related";
+            "type" => &entity_type,
+            "id" => &id,
+            "field" => &field,
+        );
+
```
leoyvens, line 1181: This log looks unrelated.

store/postgres/src/primary.rs
```diff
@@ -1108,7 +1108,29 @@ impl<'a> Connection<'a> {
         }
     }
 
-    pub fn create_deployment_features(
+    pub fn get_subgraph_features(
+        &self,
+        id: String,
+    ) -> Result<Option<(String, String, Option<String>, Vec<String>, Vec<String>)>, StoreError>
```
leoyvens, line 1114: Could this return a `DeploymentFeatures`?

leoyvens (changes requested):
Nice work on the test! Just a couple comments.

store/postgres/src/primary.rs
incrypto32: done

runtime/wasm/src/module/mod.rs
incrypto32: oops, put it there while i was debugging another issue, removed it. TY

PR #4681, opened 2023-06-07
add a retry(5) when refetch_firehose_block fails

sduchesneau:
* `refetch_firehose_block` is called when using filters and a dynamic datasource is added.
* If it fails, the whole subgraph fails and retries after 2 minutes
* This downtime is out of proportions with a single "failed http fetch", so a few retries make sense.


leoyvens (approved):

leoyvens:
Thanks!

PR #4684, opened 2023-06-12
Reduce CI flakiness: disable TCP checksum offloading

neysofu:
We're not the only ones experiencing flaky networking on Github Actions hosted runners: https://github.com/actions/runner-images/issues/1187

Someone suggested disabling TCP checksum offloading as a possible solution as hosted runners' implementation may be buggy. Performance impact should be minimal and error rates go down significantly.


incrypto32 (approved):
Looks good

Issue #4670, opened 2023-06-05
[Bug] Going back from firehose ingestor to rpc-based ingestor is broken


sduchesneau:
## Steps to reproduce:
1) launch graph-node with a subgraph, without any rpc provider, so that firehose is used as the blockIngestor
2) restart graph-node without firehose, using rpc provider as the blockingestor and stream source for the subgraph
3) observe the following debug logs: `DEBG Block stream produced a non-fatal error, error: missing field `parentHash`,`

## Workaround
* Flushing the block cache fixes the issue

## Explanation
The "firehose block ingestor" is implemented differently that the "rpc block ingestor". It only saves the block hash and number to the database:

https://github.com/graphprotocol/graph-node/blob/3d6645ed3a65772dfa7ba522e8d8908ceb7cd393/store/postgres/src/chain_store.rs#L1740-L1745 
(called from 
https://github.com/graphprotocol/graph-node/blob/master/graph/src/blockchain/firehose_block_ingestor.rs#L153)

The "rpc block ingestor" saves more data in the database:
https://github.com/graphprotocol/graph-node/blob/3d6645ed3a65772dfa7ba522e8d8908ceb7cd393/store/postgres/src/chain_store.rs#L1570-L1574
(called from ingest_block() in https://github.com/graphprotocol/graph-node/blob/master/chain/ethereum/src/ingestor.rs#L181)

## Proposed solution
either:
1) fix the firehose-based ingestor so that it adds all the required fields (namely, parent_hash)
2) add a check on graph-node start: if the subgraphs are being served from RPC for a given chain, validate that the blocks are complete in the database or flush the cache)

### Some information to help us out

- [ ] Tick this box if this bug is caused by a regression found in the latest release.
- [ ] Tick this box if this bug is specific to the hosted service.
- [X] I have searched the issue tracker to make sure this issue is not a duplicate.

### OS information

None

sduchesneau:
@neysofu interesting. Maybe this is a regression, then ? Or maybe it does not affect subgraphs that have only a BlockHandler.

sduchesneau:
I can reproduce the issue locally, like this:
1) launch with firehose provider
2) deploy a subgraph with initialBlock very close to HEAD and let it sync
3) unassign the subgraph 
4) leave it running for a bit, so that firehose blocks come in
5) restart with rpc provider
6) reassign the subgraph
[...]

Issue #4402, opened 2023-02-27
How do manually delete firehose ingested blocks in v0.30.0?


paymog:
The news for version 0.30.0 say the following:

> In order to downgrade, indexers must manually delete all blocks accumulated by Firehose in the database. For this reason, we suggest caution when switching over from RPC to Firehose

However there isn't guidance on how to do this correctly. What is the mechanism to correctly delete firehose ingested blocks when switching from firehose providers to rpc providers?

neysofu:
@mangas I'm not really sure myself. Would be able to write a small paragraph for me to include in the release notes?

mangas:
I think we should add this to graphman rather than add docs to manually change the db data.

paymog:
@mangas adding this to graphman would be wonderful but that would require another image to be released, right? While we wait for that graphman command to be implemented and released, do you mind sharing, at a high level, how to do the rollback? I'd guess we need to delete specific rows from the `chainX.blocks` table(s) but having a bit of confirmation around that would be wonderful.

tilacog:
If you don't mind clearing the whole block cache, `graphman truncate` is already available.

tilacog:
If you prefer deleting a specific block range, I'd advise doing it in chunks (of about 5k blocks). Deleting a large range at once might cause a lot of congestion in the database.

But yes, you would delete from the `chainX.blocks` table. You can use the `block_number` column as a chunk delimiter.

paymog:
Is there a way to be more surgical and only delete the blocks which were ingested by firehose? What is it about the firehose ingested blocks that makes them different from the rpc ingested blocks? Will one of the fields be null?

paymog:
Though I'm now also wondering what the correct order of operations would be when falling back to only rpc providers. Should the old blocks be truncated first and then do the rollback (which risks adding more unusable blocks to the cache before the rollback finishes)? Or would it be better to disable firehose and then perform the rollback which would potentially cause issues with the invalid block cache being used in rpc-only mode?

paymog:
I found that adding the following index

```
CREATE INDEX CONCURRENTLY on chainX.blocks USING BTREE ((data->'block'->'data'));
```

significantly speeds up the following query which can be used to delete just the firehose ingested blocks

```
DELETE FROM chainX.blocks WHERE data->'block'->'data' = 'null'::jsonb';
```

[...]

PR #4691, opened 2023-06-13
delete 'shallow blocks' in block cache when not using firehose

sduchesneau:
* Should fix this: https://github.com/graphprotocol/graph-node/issues/4670
* And this: https://github.com/graphprotocol/graph-node/issues/4402

### DIFF
--- a/store/postgres/migrations/2023-06-03_050000_index_ethereum_nulldata/down.sql
+++ b/store/postgres/migrations/2023-06-03_050000_index_ethereum_nulldata/down.sql
@@ -1,1 +1,0 @@
-DROP INDEX ethereum_blocks_expr_idx;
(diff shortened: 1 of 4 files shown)


incrypto32:
@leoyvens I tested it out locally , all the firehose blocks are getting deleted

leoyvens:
@sduchesneau Thanks for raising the index issue. I think that's going to be a problem, to trigger a full scan of all the RPC block caches on startup. It is actually possible to write a migration that adds indexes to those tables, [see this example](https://github.com/graphprotocol/graph-node/tree/master/store/postgres/migrations/2022-03-09-07000_call_cache_block_number_idx).

sduchesneau:
@leoyvens could you review now that I've incorporated your suggestion for indexes ?

leoyvens (approved):
Nice, I'll just make sure we run the migration ahead of time on hosted to avoid any hang up if the index creation turns out slower than expected.

sduchesneau:
Let me know if there's anything pending or if you expect me to merge it ^_^

incrypto32:
@sduchesneau its a code freeze currently, will merge it after next week

paymog:
Any updates on this? We just had to disable a firehose and the manual process of deleting firehose ingested blocks is quite cumbersome

PR #4692, opened 2023-06-13
don't fail ingestor on net_version changed from zero

sduchesneau:
If a chain was first set up using firehose and we then try to switch to an RPC provider, we get an error on a failing safety check for net_version.

Before this fix:
```
Jun 13 09:37:26.175 ERRO the net version for chain mainnet has changed from 0 to 1 since the last time we ran, component: BlockStore
Jun 13 09:37:26.186 INFO Creating LoadManager in disabled mode, component: LoadManager
Jun 13 09:37:26.187 ERRO Failed to create block ingestor Not starting block ingestor (chain is defective), network_name mainnet, network_name: mainnet
```

After this fix:
```Jun 13 11:11:03.201 WARN the net version for chain mainnet has changed from 0 to 1 since the last time we ran, ignoring difference because 0 means UNSET and firehose does not provide it, component: BlockStore```

### DIFF
--- a/store/postgres/src/block_store.rs
+++ b/store/postgres/src/block_store.rs
@@ -7,7 +7,7 @@ use std::{
 use graph::{
     blockchain::ChainIdentifier,
     components::store::BlockStore as BlockStoreTrait,
-    prelude::{error, BlockNumber, BlockPtr, Logger, ENV_VARS},
+    prelude::{error, warn, BlockNumber, BlockPtr, Logger, ENV_VARS},
 };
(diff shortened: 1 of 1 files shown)


leoyvens (approved):

yohanelly95:
getting this error at the moment, but its showing hex to number chain id changes. How do we resolve this? this apparently effects grafting as well, we graft to a new sgd it copies all previous data but stops indexing from the block it initially stopped working with this error `Not starting block ingestor (chain is defective)`. hey @leoyvens any insights here?

PR #4695, opened 2023-06-14
Allow redeployment of grafted subgraph even when graft_base is not available

### LINKED ISSUE #4534: [Feature] Allow redeployment of grafted subgraphs even if the graft base is not found
### Description

Subgraph validation fails if the graft base is not found when attempting a redeployment.

The error message currently displayed is this: 
```
"message": "subgraph validation error: [the graft base is invalid: deployment not found: Qm....]"
```

It looks like Graph Node attempts to validate the subgraph manifest before it realizes it already has the deployment in place.

Ideally, it should skip the validation (or just some of its steps) if the operation concerns redeployment.

_No response_

- [X] Tick this box if you plan on implementing this feature yourself.

incrypto32:
Closes #4534

### DIFF
--- a/core/src/subgraph/registrar.rs
+++ b/core/src/subgraph/registrar.rs
@@ -566,7 +566,7 @@ async fn create_subgraph_version<C: Blockchain, S: SubgraphStore>(
 ) -> Result<DeploymentLocator, SubgraphRegistrarError> {
     let raw_string = serde_yaml::to_string(&raw).unwrap();
     let unvalidated = UnvalidatedSubgraphManifest::<C>::resolve(
-        deployment,
+        deployment.clone(),
         raw,
         resolver,
         logger,
@@ -574,9 +574,9 @@ async fn create_subgraph_version<C: Blockchain, S: SubgraphStore>(
     )
     .map_err(SubgraphRegistrarError::ResolveError)
     .await?;
-
+    let exists = store.is_deployed(&deployment)?;
     let manifest = unvalidated
-        .validate(store.cheap_clone(), true)
+        .validate(store.cheap_clone(), !exists)
         .await
         .map_err(SubgraphRegistrarError::ManifestValidationError)?;
 
--- a/core/src/subgraph/registrar.rs
+++ b/core/src/subgraph/registrar.rs
@@ -575,9 +575,9 @@ async fn create_subgraph_version<C: Blockchain, S: SubgraphStore>(
     .map_err(SubgraphRegistrarError::ResolveError)
     .await?;
     let exists = store.is_deployed(&deployment)?;
-    let graft_pending = store.graft_pending(&deployment)?;
+    let should_validate = !exists || store.graft_pending(&deployment)?;
     let manifest = unvalidated
-        .validate(store.cheap_clone(), !exists || graft_pending)
+        .validate(store.cheap_clone(), should_validate)
         .await
         .map_err(SubgraphRegistrarError::ManifestValidationError)?;
 
--- a/core/src/subgraph/registrar.rs
+++ b/core/src/subgraph/registrar.rs
@@ -575,8 +575,9 @@ async fn create_subgraph_version<C: Blockchain, S: SubgraphStore>(
     .map_err(SubgraphRegistrarError::ResolveError)
     .await?;
     let exists = store.is_deployed(&deployment)?;
+    let graft_pending = store.graft_pending(&deployment)?;
     let manifest = unvalidated
-        .validate(store.cheap_clone(), !exists)
+        .validate(store.cheap_clone(), !exists || graft_pending)
         .await
         .map_err(SubgraphRegistrarError::ManifestValidationError)?;
 
--- a/core/src/subgraph/registrar.rs
+++ b/core/src/subgraph/registrar.rs
@@ -574,8 +574,16 @@ async fn create_subgraph_version<C: Blockchain, S: SubgraphStore>(
     )
     .map_err(SubgraphRegistrarError::ResolveError)
     .await?;
-    let exists = store.is_deployed(&deployment)?;
-    let should_validate = !exists || store.graft_pending(&deployment)?;
+
+    // Determine if the graft_base should be validated.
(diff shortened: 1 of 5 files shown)


store/postgres/src/subgraph_store.rs
```diff
@@ -424,6 +424,21 @@ impl SubgraphStoreInner {
...
+    pub(crate) fn deployment_exists(&self, id: &DeploymentHash) -> Result<bool, StoreError> {
+        let (store, site) = match self.store(id) {
+            Ok(pair) => pair,
+            Err(err) => {
+                if let StoreError::DeploymentNotFound(_) = err {
+                    return Ok(false);
+                } else {
+                    return Err(err);
+                }
+            }
```
tilacog, line 436: I think we can use a shorter form here:

```suggestion
            Err(StoreError::DeploymentNotFound(_)) => return Ok(false)
            Err(err) => return Err(err)
```

core/src/subgraph/registrar.rs
```diff
@@ -566,17 +566,17 @@ async fn create_subgraph_version<C: Blockchain, S: SubgraphStore>(
...
+        deployment.clone(),
         raw,
         resolver,
         logger,
         ENV_VARS.max_spec_version.clone(),
     )
     .map_err(SubgraphRegistrarError::ResolveError)
     .await?;
-
+    let exists = store.is_deployed(&deployment)?;
```
leoyvens, line 577: There is an `fn graft_pending` which we perhaps should additionally check, because it checks if the subgraph has actually finished the graft process. It would be nice if we were able to use that here.

store/postgres/src/deployment_store.rs
```diff
@@ -1441,6 +1441,11 @@ impl DeploymentStore {
         .await
     }
 
+    pub(crate) fn exists(&self, id: Arc<Site>) -> Result<bool, StoreError> {
+        let conn = self.get_conn()?;
```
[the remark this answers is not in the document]
incrypto32: @leoyvens i tried that initially wouldn't that mean i would have to make all the callers also async function to await on it?

store/postgres/src/subgraph_store.rs
```diff
@@ -516,11 +526,19 @@ impl SubgraphStoreInner {
...
+
+        let schema_version = if exists {
+            let layout = self.layout(deployment_hash)?;
+            layout.site.schema_version
+        } else if let Some(graft_base) = graft_base {
+            let layout = self.layout(graft_base)?;
+            layout.site.schema_version
+        } else {
+            DeploymentSchemaVersion::LATEST
+        };
```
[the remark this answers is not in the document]
incrypto32: got it, will make that change

store/postgres/src/deployment_store.rs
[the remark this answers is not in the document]
incrypto32: [fn create_deployment_internal](https://github.com/graphprotocol/graph-node/blob/a54409d11485b75e90d1d6ce240ff31bbe40fd04/store/postgres/src/subgraph_store.rs#L530) is not async but its callers can be made async ( cause callers all the way is async ). Should i make them all async in that case?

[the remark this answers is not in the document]
leoyvens: I see now that this is a bigger problem since `fn create_deployment_internal` is doing many sync DB operations, but is called from an async context. Ok we can leave this `exists` as sync for now, and later find a solution for the whole function.

incrypto32:
Integration tests, failing.. checking it out.

core/src/subgraph/registrar.rs
```diff
@@ -575,8 +575,13 @@ async fn create_subgraph_version<C: Blockchain, S: SubgraphStore>(
     .map_err(SubgraphRegistrarError::ResolveError)
     .await?;
 
+    let should_validate = match store.graft_pending(&deployment) {
+        Ok(graft_pending) => graft_pending,
+        Err(StoreError::DeploymentNotFound(_)) => true,
+        Err(e) => return Err(SubgraphRegistrarError::StoreError(e)),
+    };
```
leoyvens, line 582: Would be great to have comments here.

store/postgres/src/primary.rs
```diff
@@ -1247,18 +1247,33 @@ impl<'a> Connection<'a> {
     }
 
     /// Create a site for a brand new deployment.
+    /// If it already exists, return the existing site.
+    /// and a boolean indicating whether a new site was created.
+    /// false means the site already existed.
```
leoyvens, line 1252: Nice comment, just need to check ponctuation and capitalization.

leoyvens (approved):
Looks great! Thanks for bearing with the reviews

store/postgres/src/primary.rs
incrypto32: ah thanks for catching that. I'll fix that

core/src/subgraph/registrar.rs
```diff
@@ -575,8 +575,17 @@ async fn create_subgraph_version<C: Blockchain, S: SubgraphStore>(
     .map_err(SubgraphRegistrarError::ResolveError)
     .await?;
 
+    // Determine if the subgraph should be validated.
+    // Validate the subgraph if there is a pending graft, indicating the presence of a graft base.
+    // If the subgraph is new (DeploymentNotFound), it should also be validated.
+    // If the subgraph already exists and there is no pending graft, validation is not required.
```
leoyvens, line 581: The wording here is not very precise, this boolean is not deciding if the whole subgraph should be validated, but if the graft base should be validated.

incrypto32: // Determine if the graft_base should be validated.
// Validate the graft_base if there is a pending graft, ensuring its presence.
// If the subgraph is new (indicated by DeploymentNotFound), the graft_base should be validated.
// If the subgraph already exists and there is no pending graft, graft_base validation is not required.


@leoyvens how does this look?

leoyvens (approved):

PR #4699, opened 2023-06-15
 Move create_subgraph_features to build_subgraph_runner

incrypto32:


leoyvens:
@incrypto32 Glad to see this wasn't too much rework. In terms of tests, you're right, there isn't a good way to unit test the subgraph runner. We have to resort to the integration tests. This might not warrant its own integration test, but perhaps you could add a check to an existing test that the features were written correctly.

store/postgres/src/subgraph_store.rs
```diff
@@ -1320,6 +1320,14 @@ impl SubgraphStoreTrait for SubgraphStore {
         self.mirror.subgraph_exists(name)
     }
 
+    fn subgraph_features(
+        &self,
+        deployment: &DeploymentHash,
+    ) -> Result<Option<DeploymentFeatures>, StoreError> {
+        self.primary_conn()?
+            .get_subgraph_features(deployment.to_string())
```
leoyvens, line 1328: This should be made async. We should never try to get a DB connection from async code, as that is a blocking operation, so `with_conn` must be used.

PR #4713, opened 2023-06-22
Allow offchain datasource creation in offchain handlers

incrypto32:


leoyvens:
Looks good! Lets also add a test that onchain data sources cannot be created from an offchain data source.

incrypto32:
@leoyvens I've added the suggested test.

PR #4714, opened 2023-06-22
firehose/substreams: add 'bearer' in authorization field

sduchesneau:
correct header syntax is: `Authorization: bearer AAAaaBBbbbCCCcccDDddd....`

It works without the word "bearer" currently, but it does not follow the specs.


leoyvens (approved):
🐻

Issue #3221, opened 2022-02-07
Initialization handlers


azf20:
Subgraph developers often want to initialise a specific state for their subgraph.

Currently this needs to be done across handlers, or on the basis of some other trigger.

TBD: could parts of this use case be handled declaratively in the manifest e.g. via context?

azf20:
This was proposed separately in the Forum: https://forum.thegraph.com/t/grp-0001-decorator-hooks-cron-jobs-and-initialisation-handlers/3347

azf20:
This could be a `type` of block handler, similar to #959:

```
blockHandlers
  - handler: initializeState
    filter:
      kind: once
```

leoyvens:
@azf20 I like that, we'd just need to make sure it runs before any other triggers in the first block, rather than after as would be the default for a block handler.

azf20:
Interesting - I wonder if that could be a separate blockHandler configuration, unless that seems like too much of a footgun?

leoyvens:
It does seem interesting to have a separate configuration to make any block handler be handled first. Like a field `order: first|last`, defaulting to `last`.

Data-Nexus:
> @azf20 I like that, we'd just need to make sure it runs before any other triggers in the first block, rather than after as would be the default for a block handler.

Could we make it start 1 block before the earliest startblocks stated in `subgraph.yaml` thus ensuring no other handlers are triggered?

adamazad:
Hey, is there an ETA on this?

PR #4725, opened 2023-06-27
Feature : Polling blockHandler and intialization handler

### LINKED ISSUE #959: Add a polling block handler
Currently, we have blockHandlers with and without filter. Additionally, it would be great to have a polling filter that runs every nth block (or time window).

E.g.

```yml
blockHandlers
  - handler: handleFundShutDown
    every: 24h
```

incrypto32:
Closes #959, #3221

### DIFF
--- a/chain/ethereum/src/adapter.rs
+++ b/chain/ethereum/src/adapter.rs
@@ -706,7 +706,7 @@ impl EthereumBlockFilter {
             polling_intervals,
             contract_addresses,
             trigger_every_block,
-            ..
+            wildcard_filters,
         } = other;
 
         self.trigger_every_block = self.trigger_every_block || trigger_every_block;
@@ -743,6 +743,14 @@ impl EthereumBlockFilter {
                 ));
             }
         }
+
+        self.wildcard_filters = match (self.wildcard_filters.take(), wildcard_filters) {
+            (Some(mut current), Some(other)) => {
+                current.polling = current.polling || other.polling;
+                Some(current)
+            }
+            (current, other) => current.or(other),
+        };
     }
 
     fn requires_traces(&self) -> bool {
@@ -1097,7 +1105,6 @@ mod tests {
                 wildcard_signatures: HashSet::new(),
             },
             block: EthereumBlockFilter {
-                // TODO: add polling intervals
                 polling_intervals: HashSet::default(),
                 contract_addresses: HashSet::from_iter([
                     (100, address(1000)),
--- a/chain/ethereum/src/adapter.rs
+++ b/chain/ethereum/src/adapter.rs
@@ -158,7 +158,7 @@ impl bc::TriggerFilter<Chain> for TriggerFilter {
 
     fn to_firehose_filter(self) -> Vec<prost_types::Any> {
         let EthereumBlockFilter {
-            polling_intervals: _polling_intervals,
+            polling_intervals,
             contract_addresses: _contract_addresses,
             trigger_every_block,
         } = self.block.clone();
@@ -174,7 +174,7 @@ impl bc::TriggerFilter<Chain> for TriggerFilter {
         let combined_filter = CombinedFilter {
             log_filters,
             call_filters,
-            send_all_block_headers: trigger_every_block,
+            send_all_block_headers: trigger_every_block || !polling_intervals.is_empty(),
         };
 
         vec![Any {
@@ -648,7 +648,7 @@ impl EthereumBlockFilter {
                         .into_iter()
                         .filter_map(|block_handler| match block_handler.filter {
                             Some(BlockHandlerFilter::Polling { every }) => {
-                                Some((data_source.start_block, every))
+                                Some((data_source.start_block, every.get() as i32))
                             }
                             Some(BlockHandlerFilter::Once) => Some((data_source.start_block, 0)),
                             _ => None,
@@ -1066,7 +1066,7 @@ mod tests {
                 wildcard_signatures: HashSet::new(),
             },
             block: EthereumBlockFilter {
-                polling_intervals: HashSet::default(),
+                polling_intervals: HashSet::from_iter(vec![(1, 10), (3, 24)]),
                 contract_addresses: HashSet::from_iter([
                     (100, address(1000)),
                     (200, address(2000)),
@@ -1170,7 +1170,7 @@ mod tests {
             filter.event_signatures.sort();
         }
         assert_eq!(expected_log_filters, actual_log_filters);
-        assert_eq!(false, actual_send_all_block_headers);
+        assert_eq!(true, actual_send_all_block_headers);
     }
 
     #[test]
@@ -1390,7 +1390,7 @@ mod tests {
     }
 
     #[test]
-    fn extending_ethereum_block_filter_conflict_picks_lowest_block_from_base() {
+    fn extending_ethereum_block_filter_conflict_doesnt_include_both_copies() {
         let mut base = EthereumBlockFilter {
             polling_intervals: HashSet::from_iter(vec![(2, 3)]),
             contract_addresses: HashSet::from_iter(vec![(2, address(1))]),
@@ -1398,7 +1398,7 @@ mod tests {
         };
 
         let extension = EthereumBlockFilter {
-            polling_intervals: HashSet::from_iter(vec![(3, 3)]),
+            polling_intervals: HashSet::from_iter(vec![(3, 3), (2, 3)]),
             contract_addresses: HashSet::from_iter(vec![(10, address(1))]),
             trigger_every_block: false,
         };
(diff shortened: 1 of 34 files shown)


chain/ethereum/src/data_source.rs
```diff
@@ -372,6 +383,17 @@ impl DataSource {
...
+                .mapping
+                .block_handlers
+                .iter()
+                .find(move |handler| match handler.filter {
+                    Some(BlockHandlerFilter::Polling { every: interval }) => {
+                        interval == *trigger_interval
+                    }
+                    _ => false,
+                })
+                .cloned(),
```
leoyvens, line 396: Seems to me the correct matching logic would be to check that `(trigger.block - self.start_block) % handler.filter.every == 0`.

chain/ethereum/src/trigger.rs
```diff
@@ -294,6 +295,9 @@ impl EthereumTrigger {
 
             // Unfiltered block triggers match any data source address.
             EthereumTrigger::Block(_, EthereumBlockTriggerType::Every) => None,
+            EthereumTrigger::Block(_, EthereumBlockTriggerType::Polling(address, _)) => {
+                Some(address)
+            }
```
leoyvens, line 300: In principle, I think `EthereumBlockTriggerType` should not carry the address, and `None` should be returned here. As a block trigger, in principle, should not be associated to any specific address since it represents the whole block. `EthereumBlockTriggerType::WithCallTo` is an exception to this principle, and I'd prefer not to add another one.

chain/ethereum/src/ethereum_adapter.rs
```diff
@@ -1604,6 +1679,74 @@ pub(crate) fn parse_block_triggers(
...
+    }
+
+    let block_ptr = BlockPtr::from(&block.ethereum_block);
+    let block_ptr3 = block_ptr.cheap_clone();
+    let block_number = block_ptr.number;
+
+    let mut triggers = vec![];
+
+    match &block_filter.wildcard_filters {
+        Some(WildcardBlockFilter { once: true, .. }) => triggers.push(EthereumTrigger::Block(
```
incrypto32, line 1726: NOTE: This seems that it can be further optimized. These once triggers are only triggered at the startBlock so this seems very unncessary to send Every triggers in every block

chain/ethereum/src/adapter.rs
```diff
@@ -158,8 +158,10 @@ impl bc::TriggerFilter<Chain> for TriggerFilter {
 
     fn to_firehose_filter(self) -> Vec<prost_types::Any> {
         let EthereumBlockFilter {
+            polling_intervals: _polling_intervals,
```
incrypto32, line 161: @leoyvens can something be done here for `to_firehose_filter`, can we optimize the firehose filters based on polling intervals if `trigger_on_every_block` is false and only certain blocks are required?

Here it seems `send_all_block_headers` is only true when trigger_on_every_block is true. Does that mean this feature wont work with firehose?

```
 let combined_filter = CombinedFilter {
            log_filters,
            call_filters,
            send_all_block_headers: trigger_every_block,
        };
 ```

Comment (leoyvens) [reply chain/ethereum/src/adapter.rs:161]: Thanks for raising attention this point, I don't think we could optimize it without adding support for this on the Firehose side.

It does look like the current code would have a bug when syncing with Firehose that would result in missed block triggers. If any `polling_intervals` exist, then we should set `send_all_block_headers` in the Firehose filter.

File: chain/ethereum/src/adapter.rs [hunk 5]
```diff
@@ -158,8 +158,10 @@ impl bc::TriggerFilter<Chain> for TriggerFilter {
...
             contract_addresses: _contract_addresses,
             trigger_every_block,
+            ..
```
leoyvens, line 164: We should not use `..` here, as it makes it easy to add a field in the future without considering the consequences to this fn.

chain/ethereum/src/adapter.rs
```diff
@@ -583,6 +585,8 @@ impl From<&EthereumBlockFilter> for EthereumCallFilter {
 
 #[derive(Clone, Debug, Default)]
 pub(crate) struct EthereumBlockFilter {
+    /// Used for polling block handlers, a hashset of (start_block, address, polling_interval)
+    pub polling_intervals: HashSet<(BlockNumber, Address, i32)>,
```
leoyvens, line 589: I don't think this needs to carry the `Address`.

chain/ethereum/src/data_source.rs
```diff
@@ -242,23 +244,53 @@ impl blockchain::DataSource<Chain> for DataSource {
...
+            && (call_filtered_block_handler_count > 0
+                || polling_filtered_block_handler_count > 0
+                || initialization_handler_count > 0);
+
+        if has_both_filtered_and_non_filtered {
+            errors.push(anyhow!(
+                "data source has both filtered and non-filtered block handlers, \
+                     this is not allowed"
+            ));
+        }
```
leoyvens, line 284: We currently allow having a call filter and an unfiltered block handler in the same ds. So this would be a breaking change, I think we can continue to allow that combination.

But do we really want to allow combining polling and initialization handlers?

chain/ethereum/src/data_source.rs
```diff
@@ -358,13 +404,40 @@ impl DataSource {
...
+                })
+                .cloned(),
+            // End matches all handlers without a filter or with a `polling` filter
+            EthereumBlockTriggerType::End => self
+                .mapping
+                .block_handlers
+                .iter()
+                .find(move |handler| match handler.filter {
+                    Some(BlockHandlerFilter::Polling { every }) => {
+                        // POLLING_TODO: Get creation block if start_block is zero
```
leoyvens, line 427: This is very well noticed, and we need to fix before merging. There are two places where we set `start_block: 0` for dynamic data sources, maybe we should just set that to `start_block: creation_block` and this code would work fine.

chain/ethereum/src/data_source.rs
```diff
@@ -358,13 +404,40 @@ impl DataSource {
...
+                        let start_block = self.start_block;
+                        let should_trigger = match every {
+                            // 0 is an invalid polling interval
+                            // A polling interval of 0 is used internally to indicate
+                            // that the block handler should be triggered once
+                            0 => false,
```
leoyvens, line 433: Considering we now have the `Once` filter, we shouldn't need to set this to `0` for triggering only once. One cool thing you could try is using a [NonZeroU32](https://doc.rust-lang.org/stable/std/num/struct.NonZeroU32.html) to indicate that this is never zero.

chain/ethereum/src/ethereum_adapter.rs
```diff
@@ -1337,6 +1405,30 @@ pub(crate) async fn blocks_with_triggers(
     let trigger_futs: FuturesUnordered<BoxFuture<Result<Vec<EthereumTrigger>, anyhow::Error>>> =
         FuturesUnordered::new();
 
+    // This is for `start` triggers which can be initialization handlers which needs to be run
+    // before all other triggers
+    if filter.block.trigger_every_block {
+        let block_future = eth
+            .block_range_to_ptrs(logger.clone(), from, to)
```
leoyvens, line 1412: When `trigger_every_block` is set, this is now calling `block_range_to_ptrs` twice, once for Start triggers and then again for End triggers, which is very inefficient. Lets call it only once, and use the result both for the Start and End triggers.

leoyvens (changes requested):

chain/ethereum/src/adapter.rs
maoueh: Only possibility I could see would be to use the `HeaderOnly` filter, if you only need the `BlockHeader`, it's a good idea to use to avoid the network transfer of many useless bytes.

But like Leo said, Firehose would need to be adapted to send block each X block.

chain/ethereum/src/data_source.rs
incrypto32: > We currently allow having a call filter and an unfiltered block handler in the same ds

@leoyvens do people actually use that combination? I don't see the point in having any filtered handler when there is a non-filtered block handler. is it only because its a breaking change that we want to allow that?

leoyvens: If we can use some tooling to determine that no subgraph uses this combination, we could disallow it. But unless we can prove otherwise, [this usually happens](https://xkcd.com/1172/).

chain/ethereum/src/adapter.rs
incrypto32: I've modified this to set `send_all_block_headers` true when `polling_intervals` are not empty

chain/ethereum/src/data_source.rs
incrypto32: @leoyvens done!

leoyvens (approved):
Approved! We can merge after an integration test run

incrypto32:
Rebased with master

PR #4731, opened 2023-06-29
Restricted Substreams accepted types, updated to latest `substreams-sink-entity-changes` model

maoueh:
The Substreams handler should restrict its output to know message urls. Trying to decode an invalid url should be a fatal error. The only 2 acceptable types for now are `type.googleapis.com/sf.substreams.entity.v1.EntityChanges` (renamed to next accepted package) and `type.googleapis.com/sf.substreams.sink.entity.v1.EntityChanges`. The package id recently changed to be `sf.substreams.sink.entity.v1`, it has the same binary layout (and will remain the same through `v1` model), so it's not a problem accepting both.

This also brings in latest `substreams-entity-change` proto definition.

On the fatal error handling, I did something so that the stream can returns a "fatal error" and stop the block stream. This seems weirdly handled in the consuming side. First, it logs a message that error is non-fatal. But the code seems to continue on error but then the subgraph is stopped and nothing happens anymore. I'm pretty sure this is invalid and the subgraph should switch to determinisic error state but it does not seems to be the case (but I could be wrong).

What I see:

```
Jun 29 15:16:03.819 ERRO Mapping message to BlockStreamEvent failed: Invalid message type url, got type.googleapis.com/sf.sink.entity.v1.EntityChanges but accepting only type.googleapis.com/sf.substreams.entity.v1.EntityChanges, type.googleapis.com/sf.substreams.sink.entity.v1.EntityChanges, provider: substreams, deployment: QmejrrfbjjM4RTcZBFq3LuWBcsXtRcnFzvHA4doWaQBCbu, sgd: 6, subgraph_id: QmejrrfbjjM4RTcZBFq3LuWBcsXtRcnFzvHA4doWaQBCbu, component: SubstreamsBlockStream
Jun 29 15:16:03.819 DEBG Block stream produced a non-fatal error, error: Mapping message to BlockStreamEvent failed, sgd: 6, subgraph_id: QmejrrfbjjM4RTcZBFq3LuWBcsXtRcnFzvHA4doWaQBCbu, component: SubgraphInstanceManager
Jun 29 15:16:03.819 INFO Stopping subgraph, sgd: 6, subgraph_id: QmejrrfbjjM4RTcZBFq3LuWBcsXtRcnFzvHA4doWaQBCbu, component: SubgraphInstanceManager
<Nothing more after that until I Ctrl-C>
```

Which is handled https://github.com/graphprotocol/graph-node/blob/6637691999961ad4909470e0c53a12915427cc88/core/src/subgraph/runner.rs#L1026-L1043 but as you can see, the code says `Ok(Action::Continue)` but `Stopping subgraph` is called somewhere after that. Also the log says non-fatal error but don't do nothing, so either the handling needs to retry or the log needs to change.

### DIFF
--- a/chain/substreams/proto/codec.proto
+++ b/chain/substreams/proto/codec.proto
@@ -1,6 +1,8 @@
 syntax = "proto3";
 
-package substreams.entity.v1;
+package sf.substreams.sink.entity.v1;
+
+option go_package = "github.com/streamingfast/substreams-sink-entity-changes/pb/sf/substreams/sink/entity/v1;pbentity";
 
 message EntityChanges {
   repeated EntityChange entity_changes = 5;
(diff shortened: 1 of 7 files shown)


azf20:
@maoueh is this still valid?

maoueh:
It is yes but I haven't took time to rebase it. I should probably also split it in 2 PRs actually. I will try to find some time to finish this week :)

PR #4752, opened 2023-07-12
Revert "delete 'shallow blocks' in block cache when not using firehose (#4691)"

leoyvens:
This has a few of issues that need to be solved:
1. Even with the index in place, the query is slow to run for tables with a lot of blocks. Some mitigations for this:
   - Run it in a background thread, so it doesn't block the node startup.
   - In a multi-node setup, make sure only one node runs it. One way to achieve this would be to have the block ingestor do the cleanup.
2. The cleanup is being run for all non-ethereum chains, even though it should never be run for those since they can only be Firehose.


leoyvens:
I'm keeping the migration since assuming the points above are resolved before the next release, we'll still want it

PR #4757, opened 2023-07-13
use expose instead of ports

DaMandal0rian:
This PR updates the docker compose file for graph node to expose ports on the bridge network and not the host network for docker.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -2,7 +2,7 @@ version: '3'
 services:
   graph-node:
     image: graphprotocol/graph-node
-    ports:
+    expose:
       - '8000:8000'
       - '8001:8001'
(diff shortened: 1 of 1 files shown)


docker/docker-compose.yml
```diff
@@ -2,7 +2,7 @@ version: '3'
 services:
   graph-node:
     image: graphprotocol/graph-node
-    ports:
+    expose:
```
mangas, line 5: I would expect the whole point of running these is to expose the ports to the host. People should not be running production setups from this file. For this reason I do not think we should make this change, it would make it useless for development

DaMandal0rian:
Please include a disclaimer on the purpose of this docker-compose setup and intended audience.

Issue #4577, opened 2023-04-27
Extend substreams data source definition to pass `params` to parameterised substreams


azf20:
Substreams modules can be parameterised: https://substreams.streamingfast.io/developers-guide/modules/inputs#input-type-params

Substreams datasources could be extended to include `params`, which could then be passed to the substream. This would provide an easy way to leverage parameterised substreams in subgraphs without requiring any changes to the substream package

PR #4759, opened 2023-07-17
substreams: Add support for overriding module params

mangas:
If params is provided on the substreams-based subgraph manifest then replace all the existing params on the output module with the provided list.

FIxes https://github.com/graphprotocol/graph-node/issues/4577


neysofu (approved):
LGTM. Good job with those tests 😁

PR #4773, opened 2023-07-20
Fix loadRelated not working with `Bytes` as ids

### LINKED ISSUE #4701: [Bug] Can't query related entity if current entity has id of type Bytes
### Bug report

My schema looks like this:

```
interface Account {
  id: Bytes!
}

type User implements Account @entity {
  id: Bytes!
}

type Oracle @entity {
  id: Bytes!
  owner: Account!
}
```

When I try to query for the oracles owner, I get the following error: "Failed to get entities from store: operator does not exist: bytea = text"

If I change the oracle id type to ID!, then the query runs fine. Was trying to switch all of my id's over to Bytes per this article: https://thegraph.com/blog/two-simple-subgraph-performance-improvements/

I think it should be working as I have things setup? If not, or if I'm doing something wrong please let me know!

### Relevant log output

_No response_

### IPFS hash

_No response_

https://thegraph.com/hosted-service/subgraph/0xsplits/splits-subgraph-goerli-testing

### OS information

None

incrypto32:
Closes #4701

### DIFF
--- a/graph/src/components/store/entity_cache.rs
+++ b/graph/src/components/store/entity_cache.rs
@@ -199,13 +199,14 @@ impl EntityCache {
         &mut self,
         eref: &LoadRelatedRequest,
     ) -> Result<Vec<Entity>, anyhow::Error> {
-        let (base_type, field) = self.schema.get_field_related(eref)?;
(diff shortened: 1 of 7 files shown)


dotansimha (approved):
LGTM 🚀

Issue #4323, opened 2023-01-27
Duplicated provider without any configuration


maoueh:
**Do you want to request a *feature* or report a *bug*?**

I would say more a mis-behavior

**What is the current behavior?**

I was doing some testing with Arweave, in my config I have a single provider defined, but it seems `graph-node` is actually auto duplicating it to become like 20 providers, all with the same endpoint:

```
Jan 27 13:12:41.566 DEBG Cleaning up large notifications after about 300s, channel: chain_head_updates, component: ChainHeadUpdateListener > NotificationListener
Jan 27 13:12:41.714 INFO Connecting to Firehose to get chain identifier, provider: arweave-mainnet-0, provider: arweave-mainnet-0
Jan 27 13:12:41.714 INFO Requesting genesis block from firehose, provider: arweave-mainnet-0
Jan 27 13:12:41.715 DEBG Connecting to firehose to retrieve block for number 0, provider: arweave-mainnet-0
Jan 27 13:12:41.715 INFO Connecting to Firehose to get chain identifier, provider: arweave-mainnet-1, provider: arweave-mainnet-1
Jan 27 13:12:41.715 INFO Requesting genesis block from firehose, provider: arweave-mainnet-1
Jan 27 13:12:41.715 DEBG Connecting to firehose to retrieve block for number 0, provider: arweave-mainnet-1
...
Jan 27 13:12:41.717 DEBG Connecting to firehose to retrieve block for number 0, provider: arweave-mainnet-19
Jan 27 13:12:42.088 DEBG Retrieving block(s) from firehose, provider: arweave-mainnet-14
...
Jan 27 13:12:42.308 DEBG Retrieving block(s) from firehose, provider: arweave-mainnet-8
Jan 27 13:12:42.550 INFO Connected to Firehose, genesis_block: #900000 (d750464ec1fb149c031a4ed76a1b3bfc324f800b5b5ad2b66ade5906e8520ecd95e351a8b2fa1c673f83f0594775dabb), provider: arweave-mainnet-14, provider: arweave-mainnet-14
...
Jan 27 13:12:42.879 INFO Connected to Firehose, genesis_block: #900000 (d750464ec1fb149c031a4ed76a1b3bfc324f800b5b5ad2b66ade5906e8520ecd95e351a8b2fa1c673f83f0594775dabb), provider: arweave-mainnet-6, provider: arweave-mainnet-6
```

**What is the expected behavior?**

[...]

PR #4778, opened 2023-07-24
Handle conn pool initialization gracefully

mangas:
- Duplicated labels no longer allowed on config
- Endpoints get tested/connected to per provider label instead of per endpoint


neysofu (approved):

maoueh:
So that this avoid doing N firehose endpoint check and does only one per provider is that correct?

maoueh:
If it's true, I think we can safely close https://github.com/graphprotocol/graph-node/issues/4323, we should still probably change the default `conn_pool` to 1 IMO, but that and the improved balancing across endpoints, but they can be left for another brand new issue.

PR #4779, opened 2023-07-24
Add subgraph paused status to index node API

### LINKED ISSUE #4667: [Feature] Add subgraph status (paused/active) to index-node API and graphman info
### Description

After #4642 , subgraphs can be paused and resumed using `graphman` it would be good to add the status (paused/active) to the index-node API and `graphman info`

_No response_

- [X] Tick this box if you plan on implementing this feature yourself.

incrypto32:
Closes #4667

### DIFF
--- a/node/src/manager/deployment.rs
+++ b/node/src/manager/deployment.rs
@@ -200,6 +200,7 @@ impl Deployment {
         ];
         if !statuses.is_empty() {
             rows.extend(vec![
+                "paused",
                 "synced",
                 "health",
                 "earliest block",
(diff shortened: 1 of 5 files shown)


node/src/manager/deployment.rs
```diff
@@ -228,6 +229,10 @@ impl Deployment {
             if let Some(status) = status {
                 let chain = &status.chains[0];
                 rows.extend(vec![
+                    status
+                        .paused
+                        .map(|b| b.to_string())
+                        .unwrap_or("---".to_string()),
```
mangas, line 235: What is this?

mangas (approved):

node/src/manager/deployment.rs
mangas: I still dont understand why --- rather than defaulting to a sensible value like "false"?

mangas: I'm approving to unblock but I'd rather change this, however I'll leave that decision to you

incrypto32: Its a placeholder for `None` the value is a `Option<bool>`

PR #4787, opened 2023-07-27
[Feature] Add support for `endBlock` in data sources

incrypto32:


core/src/subgraph/runner.rs
```diff
@@ -144,12 +144,59 @@ where
...
+                let static_data_sources = self.ctx.instance().static_data_sources();
+
+                let mut filter = C::TriggerFilter::from_data_sources(
+                    static_data_sources
+                        .iter()
+                        .filter_map(|ds| ds.as_onchain())
+                        // Filter out data sources that have reached their end block.
+                        .filter(|ds| match ds.end_block() {
+                            Some(end_block) => {
+                                current_ptr.as_ref().map(|ptr| ptr.number).unwrap_or(0) > end_block
```
incrypto32, line 171: @leoyvens we need to somehow know that the block is final before the end_block reached datasource is removed when reconstructing the filter, how do we know which block is final and which is not at this point?

chain/ethereum/src/data_source.rs
```diff
@@ -217,6 +225,8 @@ impl blockchain::DataSource<Chain> for DataSource {
             manifest_idx,
             address,
             start_block: 0,
+            // EBTODO: Re-evaluate if this needs to be set to done_at,
+            end_block: done_at,
```
mangas, line 229: As far as I understand, done_at is only used for file data sources, it does not make sense to use it here as there is no "done" for a Dynamic Data Source in other scenarios. Would be good to confirm with @leoyvens as he may have intended this to be used in some other and add a comment to explain what it is and what it is used for to avoid confusion in the future.

mangas: This is to say I don't think we should use this field as that will make it quite difficult to understand the meaning on different contexts, I'd rather just add an explicit field end_block or maybe consider deleting the data source once it's no longer needed?

core/src/subgraph/context.rs
```diff
@@ -177,10 +174,13 @@ impl<C: Blockchain, T: RuntimeHostBuilder<C>> IndexingContext<C, T> {
         self.instance.causality_region_next_value()
     }
 
-    #[cfg(debug_assertions)]
     pub fn instance(&self) -> &SubgraphInstance<C, T> {
         &self.instance
     }
+
+    pub fn hosts(&self) -> &[Arc<T::Host>] {
```
mangas, line 181: There's already an instance fn, why not just make the field public?

core/src/subgraph/context/instance.rs
```diff
@@ -176,6 +180,7 @@ where
             return vec![];
         }
 
+        // EBTODO
```
mangas, line 183: ??

chain/ethereum/src/data_source.rs
```diff
@@ -146,6 +152,8 @@ impl blockchain::DataSource<Chain> for DataSource {
             address,
             mapping,
             context,
+            // EBTODO: Re-evaluate if endBlock need to be considered
```
mangas, line 155: no need for this comment

chain/ethereum/src/data_source.rs
```diff
@@ -217,6 +225,8 @@ impl blockchain::DataSource<Chain> for DataSource {
             manifest_idx,
             address,
             start_block: 0,
+            // EBTODO: Re-evaluate if this needs to be set to done_at,
```
mangas, line 228: comment

core/src/subgraph/context/instance.rs
```diff
@@ -215,10 +220,17 @@ where
         self.causality_region_seq.next_val()
     }
 
-    #[cfg(debug_assertions)]
     pub fn hosts(&self) -> &[Arc<T::Host>] {
         &self.hosts.hosts()
     }
+
+    pub fn static_data_sources(&self) -> Arc<Vec<DataSource<C>>> {
```
mangas, line 227: Why?

core/src/subgraph/context/instance.rs
```diff
@@ -215,10 +220,17 @@ where
...
+        self.static_data_sources.clone()
+    }
+
+    pub fn templates(&self) -> Arc<Vec<DataSourceTemplate<C>>> {
```
mangas, line 231: same here

core/src/subgraph/context/instance.rs
```diff
@@ -36,6 +37,8 @@ where
     pub fn from_manifest(
         logger: &Logger,
         manifest: SubgraphManifest<C>,
+        static_data_sources: Vec<DataSource<C>>,
```
mangas, line 40: the terminology we use is data sources and dynamic data source, please keep this for consistency

core/src/subgraph/context/instance.rs
```diff
@@ -59,7 +63,7 @@ where
         // Create a new runtime host for each data source in the subgraph manifest;
         // we use the same order here as in the subgraph manifest to make the
         // event processing behavior predictable
-        for ds in manifest.data_sources {
+        for ds in data_sources {
```
mangas, line 66: If manifest already has all the necessary information why pass these separate?

core/src/subgraph/runner.rs
```diff
@@ -144,12 +144,59 @@ where
...
+
+            let static_filters = self.inputs.static_filters
+                || self.ctx.instance().hosts().len() > ENV_VARS.static_filters_threshold;
+
+            // if static_filters is enabled, build a minimal filter with the static data sources and
+            // add the necessary filters based on templates.
+            // if not enabled we just stick to the filter based on all the data sources.
+            // This specifically removes dynamic data sources based filters because these can be derived
+            // from templates AND this reduces the cost of egress traffic by making the payloads smaller.
+            let filter = if static_filters {
```
mangas, line 157: Why was this moved to the runner?

graph/src/data_source/mod.rs
```diff
@@ -158,9 +158,19 @@ impl<C: Blockchain> DataSource<C> {
...
     ) -> Result<Option<TriggerWithHandler<MappingTrigger<C>>>, Error> {
         match (self, trigger) {
-            (Self::Onchain(ds), TriggerData::Onchain(trigger)) => ds
-                .match_and_decode(trigger, block, logger)
-                .map(|t| t.map(|t| t.map(MappingTrigger::Onchain))),
+            (Self::Onchain(ds), TriggerData::Onchain(trigger)) => {
+                // If the data source has an `endBlock` field, check that the trigger block is
+                // within the range of blocks that the data source is supposed to handle.
+                // Otherwise, ignore the trigger.
+                if let Some(end_block) = ds.end_block() {
```
mangas, line 165: Maybe this can be added as another pattern with a guard, feels like that would be a lot easier read. something like: 

```
(Self::Onchain(ds), _ ) if ds.has_expired(block.number())
```

store/postgres/src/dynds/mod.rs
```diff
@@ -47,7 +47,7 @@ pub(crate) fn revert(
     }
 }
 
-pub(crate) fn update_offchain_status(
+pub(crate) fn update_processed_datasources_status(
```
mangas, line 50: Why? What does "processed" add to this name that it didn't have before?

store/test-store/tests/chain/ethereum/manifest.rs
```diff
@@ -369,6 +369,44 @@ specVersion: 0.0.2
...
+specVersion: 0.0.2
+";
+
+    let manifest = resolve_manifest(YAML, SPEC_VERSION_0_0_4).await;
+    // Check if end block is parsed correctly
+    let data_source = manifest.data_sources.first().unwrap();
+    let end_block = data_source.as_onchain().unwrap().end_block;
+
+    assert_eq!(Some(9562481), end_block);
+    assert_eq!("Qmmanifest", manifest.id.as_str());
```
mangas, line 407: What is this actually adding to the test?

store/postgres/src/dynds/mod.rs
incrypto32: Ah nice catch, this was done because initially the plan was to use done_at for on-chain data sources as well, but midway we decided that is not neccessary, will change it back

core/src/subgraph/context/instance.rs
incrypto32: What manifest has is the static data sources defined in the manifest, earlier the manifest was mutated to add in created dynamic datasources, this is no longer the case, the `data_sources` vec now has the complete static_data_source + dynamic_data_sources in it, and `manifest.data_sources` only contains the static data sources

core/src/subgraph/context/instance.rs
incrypto32: ty for pointing out, actually that vec is not at all needed there since manifest carries it, so i just removed it. the reasoning behind `static_data_sources` naming was that in the same code `data_sources` is used to indicate a vec containing both static and dynamic ds's

core/src/subgraph/runner.rs
incrypto32: We now need to rebuild the TriggerFilter every time the Blockstream is restarted, this is because  its not possible to remove a datasource from TriggerFilter when its endBlock is reached, As per my discussion with @leoyvens we decided reconstructing the trigger filter is a more robust way to deal with this

core/src/subgraph/context/instance.rs
incrypto32: @mangas I dont understand, are you suggesting making the field public instead of exposing it in a method?

store/test-store/tests/chain/ethereum/manifest.rs
incrypto32: This was just a test i initially wrote to test the parsing of the endBlock field, if you think its not relevant i'll remove it

core/src/subgraph/context.rs
incrypto32: done

chain/ethereum/src/data_source.rs
incrypto32: I just set the field to `None` for now, but if we decide to have a way to have an `endBlock` for dynamic data sources, we would need to use `done_at` at some point and then set `endBlock` here.

graph/src/data_source/mod.rs
incrypto32: This is done

store/test-store/tests/chain/ethereum/manifest.rs
mangas: No, I'm all for the test, I think that's great, I meant the assertion of Qmmanifest :)

incrypto32: Yes, it's not relevant was a residue of a copy paste, removed it

chain/near/src/data_source.rs
```diff
@@ -75,6 +75,13 @@ impl blockchain::DataSource<Chain> for DataSource {
         self.source.start_block
     }
 
+    fn has_expired(&self, block: BlockNumber) -> bool {
+        match self.source.end_block {
```
mangas, line 79: if you add end_block() to the trait, then this could be implemented in the trait itself as default and wouldn't need to be copied into all the trait implementations. Might not be a bad trade-off

core/src/subgraph/context/instance.rs
mangas: We are still passing the templates around, so I'm not sure about this code, if we don't need the entire manifest anymore then could we just pass the bits that are needed?

core/src/subgraph/context/instance.rs
mangas: I don't think we should be adding getter methods unless they are absolutely needed. I'm also not too happy about introducing "static_data_sources" as a term, historically we have referred to these as "data_sources" and "dynamic_data_sources" if we need this distinction then perhaps it should be done with types

core/src/subgraph/runner.rs
```diff
@@ -144,12 +144,58 @@ where
...
```
mangas, line 157: I think this needs to go into some functions rather than being inlined here

core/src/subgraph/context/instance.rs
incrypto32: The reason for passing `datasources` separately from the manifest was to avoid mutating the manifest to add in the dynamic data sources (which was also a TODO in a comment)

> We are still passing the templates around, so I'm not sure about this code
Templates are passed on to `SubgraphInstance` which is used in `HostBuilder`

The manifest is still used and is needed for passing down `manifest.data_sources` and `manifest.templates` to the Instance.

core/src/subgraph/context/instance.rs
incrypto32: makes sense, so better making it `pub` than exposing it in a method, got it.

Also i think yes we can rename it to `data_sources` instead, This is true for `manifest.data_sources` where only "static" datasources where stored. 

We can keep that naming, i was trying to make a clear distinction because, Although the `SubgraphManifest.data_sources` was meant to be just the "static" datasources, it used to carry both static and dynamic datasources when its mutated to add dynamic sources in `build_subgraph_runner`.

chain/near/src/data_source.rs
incrypto32: @mangas then we'd still have to go around implementing `end_block()` on all Datasources right? its either `end_block()` or `has_expired()`. Lets go with `has_expired` then?

core/src/subgraph/runner.rs
```diff
@@ -657,11 +717,6 @@ where
             );
             block_state.persist_data_source(data_source.as_stored_dynamic_data_source());
         }
-
-        // Merge filters from data sources into the block stream builder
-        self.ctx
-            .filter
-            .extend(data_sources.iter().filter_map(|ds| ds.as_onchain()));
```
incrypto32, line 664: We dont need to Merge the filters anymore since we rebuild the filters each time the block stream is restarted. And block stream is restarted everytime an onchain data source is created

core/src/subgraph/runner.rs
```diff
@@ -1041,7 +1096,22 @@ where
...
+            .is_some();
+
+        let action = if needs_restart {
+            Action::Restart
+        } else {
+            Action::Continue
+        };
+
+        Ok(action)
     }
```
incrypto32, line 1115: This the part where we determine if the blockstream is to be restarted on a reorg. If there is atleast one datasource which reached its endBlock during the block range for which the reorg is happening then we need to restart

incrypto32:
Rebased with master, Addressed the blockstream restart issue, We now restart the block stream on a reorg when there is an endBlock reached datasource in the reorg range

core/src/subgraph/runner.rs
```diff
@@ -106,6 +106,57 @@ where
...
+        let current_ptr = self.inputs.store.block_ptr();
+        let static_filters = self.inputs.static_filters
+            || self.ctx.instance().hosts().len() > ENV_VARS.static_filters_threshold;
+
+        // if static_filters is enabled, build a minimal filter with the static data sources and
+        // add the necessary filters based on templates.
+        // if not enabled we just stick to the filter based on all the data sources.
+        // This specifically removes dynamic data sources based filters because these can be derived
+        // from templates AND this reduces the cost of egress traffic by making the payloads smaller.
+        let filter = if static_filters {
```
mangas, line 119: you can get rid of this nesting, either `if static_filters { ... return filter}` and remove the else or just return the result of the if expression.

tests/tests/runner_tests.rs
```diff
@@ -309,6 +309,52 @@ async fn derived_loaders() {
     );
 }
 
+#[tokio::test]
+async fn end_block() -> anyhow::Result<()> {
```
mangas, line 313: is it possible to add a test for the revert to make sure we have coverage of the filter rebuild?

incrypto32: Just made the changes for making the filter rebuild testable

chain/cosmos/src/data_source.rs
```diff
@@ -505,6 +509,8 @@ pub struct MappingMessageHandler {
 pub struct Source {
     #[serde(rename = "startBlock", default)]
     pub start_block: BlockNumber,
+    #[serde(rename = "endBlock")]
```
mangas, line 512: https://serde.rs/container-attrs.html#rename_all  <- at this point might as well add this

chain/near/src/data_source.rs
```diff
@@ -498,5 +502,7 @@ pub(crate) struct Source {
     pub(crate) account: Option<String>,
     #[serde(rename = "startBlock", default)]
     pub(crate) start_block: BlockNumber,
+    #[serde(rename = "endBlock")]
```
mangas, line 505: prolly same here https://serde.rs/container-attrs.html#rename_all

core/src/subgraph/runner.rs
```diff
@@ -144,12 +197,17 @@ where
 
             let block_stream_canceler = CancelGuard::new();
             let block_stream_cancel_handle = block_stream_canceler.handle();
+            // TriggerFilter needs to be rebuilt eveytime the blockstream is restarted
```
mangas, line 200: Exactly this, if the build returns the trigger filter and replace it then we know what mutations it needs to make and the build_filter function can then take a reference instead of mutable ref

core/src/subgraph/runner.rs
```diff
@@ -144,12 +197,17 @@ where
...
+            self.build_filter();
 
-            let mut block_stream =
-                new_block_stream(&self.inputs, &self.ctx.filter, &self.metrics.subgraph)
-                    .await?
-                    .map_err(CancelableError::Error)
-                    .cancelable(&block_stream_canceler, || Err(CancelableError::Cancel));
+            let mut block_stream = new_block_stream(
+                &self.inputs,
+                self.ctx.filter.as_ref().unwrap(), // Safe to unwrap as we just called `build_filter` in the previous line
```
mangas, line 205: we would also still have the filter here so maybe we could just pass that in instead of unwrapping, mostly a read improvement

core/src/subgraph/runner.rs
```diff
@@ -1041,7 +1102,22 @@ where
 
         self.revert_state(subgraph_ptr.number)?;
 
-        Ok(Action::Continue)
+        // If any datasource has reached its end block in the range [revert_to_ptr, subgraph_ptr]
+        // then we need to restart the blockstream to reset the TriggerFilter.
+        let needs_restart: bool = self
```
mangas, line 1107: I think this should be a function to improve readability

graph/src/blockchain/mod.rs
```diff
@@ -255,15 +255,31 @@ pub trait DataSource<C: Blockchain>: 'static + Sized + Send + Sync + Clone {
...
     fn kind(&self) -> &str;
     fn network(&self) -> Option<&str>;
     fn context(&self) -> Arc<Option<DataSourceContext>>;
     fn creation_block(&self) -> Option<BlockNumber>;
     fn api_version(&self) -> semver::Version;
+
+    /// **NOTE: DO NOT OVERRIDE THIS METHOD.**
+    /// Use `min_spec_version_inner` for custom logic instead.
+    /// This method is used to determine the minimum spec version required by the data source.
     fn min_spec_version(&self) -> semver::Version {
```
mangas, line 269: This does not look like the right place for this sort of contraint, maybe the manifest validation would be the right place. This here is leaking the details about spec into data sources

incrypto32: Got it, i moved this logic to manifest validation

core/src/subgraph/runner.rs
incrypto32: To prevent the unwrapping, i would need to clone the filter to once to assign to `self.ctx.filter` and then to pass it to be blockstream, so just going with unwrap for now

core/src/subgraph/runner.rs
incrypto32: Thanks, just did that

chain/near/src/data_source.rs
incrypto32: done

core/src/subgraph/runner.rs
incrypto32: Yeah the logic is same but the context in which `has_expired` is used is different.

core/src/subgraph/runner.rs
mangas: This doesn't seem to be addressed

mangas (approved):
Would be great to see the improvements we discussed privately done before merging but will leave that up to you

PR #4790, opened 2023-07-31
delete 'shallow' fh eth blks on rpc+ingestor start

sduchesneau:
when launching grpah-node with an ingestor on an eth chain with an RPC provider (i.e. not firehose), this will delete the 'shallow' blocks in chainX.blocks, so they do not cause issues to the RPC provider.

It only runs on ethereum chains where firehose is not used and when ingestor is active.


sduchesneau:
@leoyvens I am not quite sure how to delete the shallow blocks in a background thread without causing a race condition with the indexer...
So I have two questions:
1) Do you believe there are risks associated with the deletion while the ingestor will be ingesting more blocks ?
2) If there are none, could you give me some pointers on what to use for this ? :)

store/postgres/src/block_store.rs
```diff
@@ -456,9 +456,15 @@ impl BlockStore {
                     continue;
                 };
             }
+            match store.chain_head_block(&&store.chain).unwrap_or(None) {
```
leoyvens, line 459: How about:

`if let Some(head) = store.chain_head_block(&&store.chain)?`

store/postgres/src/block_store.rs
```diff
@@ -456,9 +456,15 @@ impl BlockStore {
...
+                Some(head) => {
+                    let lower_bound = head - ENV_VARS.reorg_threshold;
```
leoyvens, line 461: Check for overflow. Also I'd give some slack here so we don't have to worry about off by one, race conditions or whatever, maybe use `2 * ENV_VARS.reorg_threshold`.

leoyvens:
I did a review of the code to check if it's safe to keep the final blocks ingested by Firehose in the cache when switching to RPC. The concern being that those blocks could have a `data` column in a format incompatible with what the RPC block stream accepts. It really is incompatible, but the RPC block stream is robust to this because:
- In `chain_store.rs`, the data json value is returned only in `fn ancestor_block` and `fn blocks`.
- The RPC block stream uses `fn ancestor_block` only for non-final blocks.
- When it uses `fn blocks`, it ignores parsing errors:
https://github.com/graphprotocol/graph-node/blob/aeacbebd09603fae485430d9ce169cee94fe6f90/chain/ethereum/src/ethereum_adapter.rs#L1271

So while the situation with json formats for the `data` field is a bit of mess, it seems deleting just the blocks within the reorg threshold will work.

If my review is wrong and something complains of deserialization failure, we can re-revisit this and resort to the alternative of truncating the whole blocks table when a switch from Firehose to RPC is detected.

leoyvens (approved):

store/postgres/src/block_store.rs
```diff
@@ -449,7 +449,8 @@ impl BlockStore {
     // table but RPC provider requires those blocks to be full.
     //
     // - This issue only affects ethereum chains.
-    // - This issue only happens when switching providers from firehose back to RPC
+    // - This issue only happens when switching providers from firehose back to RPC. it is gated by
+    // the presence of a cursor in the chainX.blocks table for a chain configured without firehose.
```
leoyvens, line 453: The cursor is actually stored in the `ethereum_networks` table.

store/postgres/src/block_store.rs
```diff
@@ -469,18 +470,20 @@ impl BlockStore {
...
-            if let Some(head) = store.chain_head_block(&&store.chain)? {
-                let lower_bound = head.saturating_sub(ENV_VARS.reorg_threshold * 2);
 
-                info!(&self.logger, "Cleaning shallow blocks and cursor on non-firehose chain"; "network" => &store.chain, "lower_bound" => lower_bound);
-                let ret = store.remove_cursor();
-                if ret.is_err() {
-                    return ret;
+            match store.remove_cursor(&&store.chain) {
+                Ok(res) => {
+                    if let Some(head_block) = res {
```
leoyvens, line 476: Style:
```
if let Some(head_block) = store.remove_cursor(&&store.chain)? {
 // ...
}
```

PR #4799, opened 2023-08-09
Fix `loadDerived` not taking entity cache into consideration when loading derived entities

### LINKED ISSUE #4727: [Bug] creating related entity and loading in the same transaction not working
### Bug report

I have two call handlers, one is creating a related entity and the other is loading it. That is working as expected except for the case where both handlers are running on the same transaction. In that case the first call handler runs and creates the entity. Then the second handler runs but when it goes to load the related field it's not finding the entity that was just created.

Here's an example [query](https://api.thegraph.com/subgraphs/name/0xsplits/splits-subgraph-goerli-testing/graphql?query=%7B%0A++user%28id%3A%220xfaa66ecff431321498be86ebfaa92b8a3f9b604a%22%29%7B%0A++++contractEarnings%7B%0A++++++contract+%7B%0A++++++++id%0A++++++%7D%0A++++++internalBalances+%7B%0A++++++++token+%7B%0A++++++++++id%0A++++++++%7D%0A++++++++amount%0A++++++%7D%0A++++++withdrawals%7B%0A++++++++token+%7B%0A++++++++++id%0A++++++++%7D%0A++++++++amount%0A++++++%7D%0A++++%7D%0A++%7D%0A%7D)
In that example, the first call handler creates a `ContractEarnings` entity and sets an internal balance for it. The second call handler loads all `ContractEarnings` on the user object and moves the internal balances over to the withdrawals. I did one contract (id `0x021b97710febaf68ed56349acd46eb9e27d29a00`) in two separate transactions. The other contract (id `0x35530ee9bda38fdc0877f93d946908fb473552fd`) I ran a multicall so both steps were performed in the same transaction. You can see in the latter case there's still a balance in the internal balances, and the withdrawals are empty. Seems like the `.load` isn't finding the newly created `ContractEarnings` related entity on the user object in that case.

### Relevant log output

_No response_

### IPFS hash

_No response_

https://thegraph.com/hosted-service/subgraph/0xsplits/splits-subgraph-goerli-testing

### OS information

None

incrypto32:
Closes #4727


graph/src/components/store/entity_cache.rs
```diff
@@ -209,11 +209,59 @@ impl EntityCache {
...
+            }
+        });
+
+        // Retrieve the derived entities from the cache by filtering the cached entities.
+        // Include entities if either a matching value is found in the specified entity field
+        // as indicated by `query`, or if the entity is None.
+        let entities = self.current.iter().filter(|(key, entity)| {
+            key.entity_type == query.entity_type
+                && entity.as_ref().map_or(true, |e| {
+                    e.get(&query.entity_field.to_camel_case())
```
lutter, line 243: Why is `to_camel_case` needed here?

graph/src/components/store/entity_cache.rs
```diff
@@ -209,11 +209,59 @@ impl EntityCache {
...
+                        .map(|v| v.to_string() == query.value.to_string())
```
lutter, line 244: Is there any way to avoid the `to_string()` here? It seems a bit sketchy to convert these values to strings, and also causes a lot of allocations.

graph/src/components/store/entity_cache.rs
```diff
@@ -209,11 +209,59 @@ impl EntityCache {
...
+                self.current.insert(k.clone(), None);
+            }
         });
-        let entities: Vec<Entity> = entities.values().cloned().collect();
+
+        self.handler_updates.iter().for_each(|(k, _)| {
+            if !self.current.contains_key(k) {
+                self.current.insert(k.clone(), None);
+            }
+        });
```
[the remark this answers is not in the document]
lutter: Actually, I think I understand why you are inserting the `None` here - so that you can drive the lookup loop below by iterating over `self.current`. Wouldn't it be better to iterate over `self.updates` and `self.handler_updates` and merge whatever is found there into the entities from the store? That way, you don't need to iterate over `self.current` and therefore don't need to insert `None`

lutter (changes requested):

graph/src/components/store/entity_cache.rs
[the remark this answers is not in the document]
incrypto32: Ok, i was intially unaware of updates and handlerUpdates so thought everything thats updated is in self.current. Good to know. Will make those changes

graph/src/components/store/entity_cache.rs
incrypto32: query.entity_field is in snake case and the same field is in snake case in the type `Entity`

graph/src/components/store/entity_cache.rs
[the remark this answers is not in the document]
incrypto32: On second thoughts i think this is actually the best approach i could think of, its something similar that happens in `store.get` besides that if i were to iterate over `self.updates` and `handler_updates` i would still need to create a `None` entity and then call EntityOp.apply_to() on the None entity in the case where the entities are newly created entities within the block / handler.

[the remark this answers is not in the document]
incrypto32: Same thing happens with store.get() if the entity is not present in the store a `None` value is getting inserted into `self.current` [here](https://github.com/graphprotocol/graph-node/blob/d97ca7c6e326f3f4eb032794fd81335006641c17/graph/src/components/store/entity_cache.rs#L172)

Then later entity operations is being applied to this `None` value [here](https://github.com/graphprotocol/graph-node/blob/d97ca7c6e326f3f4eb032794fd81335006641c17/graph/src/components/store/entity_cache.rs#L188)

graph/src/components/store/entity_cache.rs
```diff
@@ -209,11 +209,59 @@ impl EntityCache {
...
 
+        // Get the entities from the store and insert it into the cache if it's not already there
         let entities = self.store.get_derived(&query)?;
-        entities.iter().for_each(|(key, e)| {
-            self.current.insert(key.clone(), Some(e.clone()));
+
+        entities.iter().for_each(|(k, v)| {
+            // Only insert to the cache if it's not already there
+            if !self.current.contains_key(k) {
+                self.current.insert(k.clone(), Some(v.clone()));
```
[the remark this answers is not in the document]
incrypto32: yes `v` is of type &Entity and it needs to be converted to Entity and it cannot be dereferenced as well

graph/src/components/store/entity_cache.rs
incrypto32: How does this sound?

 ```
   .map(|v| match v {
                            Value::String(s) => s.as_str() == query.value.as_str(),
                            Value::Bytes(b) => b.as_slice() == query.value.as_bytes(),
                            _ => false,
     })
```

File: graph/src/components/store/entity_cache.rs (hunk 3 above)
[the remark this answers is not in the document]
Comment (lutter) [reply graph/src/components/store/entity_cache.rs:235]: My point is that we don't really know here whether the entity is in the store or not; and it seems more logical to me to compose the list of entities we return of entities from the store + what we find in `self.updates` and `self.handler_updates`.

File: graph/src/components/store/entity_cache.rs (hunk 2 above)
Comment (lutter) [reply graph/src/components/store/entity_cache.rs:244]: Is that different from `v == query.value` ?

File: graph/src/components/store/entity_cache.rs (hunk 1 above)
Comment (lutter) [reply graph/src/components/store/entity_cache.rs:243]: Ugh .. that's pretty wild. Definitely an oversight when we first merged this. Stuff called `_field` should use GraphQL names.

File: graph/src/components/store/entity_cache.rs (hunk 2 above)
Comment (incrypto32) [reply graph/src/components/store/entity_cache.rs:244]: yes query.value is of type `Word` and v is of type `Value`

File: graph/src/components/store/entity_cache.rs (hunk 3 above)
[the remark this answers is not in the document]
Comment (incrypto32) [reply graph/src/components/store/entity_cache.rs:235]: Ok, i see your point now. That was an oversight from my part thanks for pointing. Since we insert all the entities from `store.get_derived()` to  `self.current` i thought we can be sure that if its not in `self.current` its not both in the store and cache, oversaw the part where updates could contain entities that are not related to the ones we are searching for

File: graph/src/components/store/entity_cache.rs (hunk 1 above)
Comment (incrypto32) [reply graph/src/components/store/entity_cache.rs:243]: Ah makes sense, there was a bug that i fixed for the derived loaders and i remember the fix was using to_snake_case() somewhere in the code.

File: graph/src/components/store/entity_cache.rs (hunk 3 above)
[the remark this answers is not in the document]
Comment (incrypto32) [reply graph/src/components/store/entity_cache.rs:235]: I think we still would need to iterate over `self.current` since when we move on the next block and a new blockstate is created `updates` becomes empty, only `self.current` now has the updated entities, neither store nor `updates` has them

File: graph/src/components/store/entity_cache.rs (hunk 1 above)
Comment (incrypto32) [reply graph/src/components/store/entity_cache.rs]: https://github.com/graphprotocol/graph-node/issues/4682

That happened because the store expected snake_case and needed to convert it, I've changed this back and only call `to_snake_case` at the point of constructing the sql query

File: graph/src/components/store/entity_cache.rs [hunk 5]
```diff
@@ -203,18 +202,94 @@ impl EntityCache {
...
+            }
+
+            if let Some(op) = self.handler_updates.get(key).cloned() {
+                op.apply_to(&mut entity_cow)
+                    .map_err(|e| key.unknown_attribute(e))?;
+            }
+
+            if let Some(updated_entity) = entity_cow {
+                *entity = updated_entity.into_owned();
+            }
```
lutter, line 237: If `entity_cow` is `None` here, shouldn't we remove it from `entity_map`? AFAICT, that indicates that the entity exists in the store but was deleted.

graph/src/components/store/entity_cache.rs
```diff
@@ -203,18 +202,94 @@ impl EntityCache {
...
+            match op {
+                EntityOp::Update(entity) | EntityOp::Overwrite(entity)
+                    if query.matches(key, entity) =>
+                {
+                    Ok(Some(entity.clone()))
+                }
+                EntityOp::Remove => Ok(None),
+                _ => Ok(None),
+            }
+        }
```
lutter, line 255: I think the comment and the name of the function are what's confusing me - we are not applying an update. We use this only on entities that do not exist in the store. That means this is either the creation of an entity or the deletion of a nonexistent entity. We can ignore the deletion, and for the creation we want to add it to `entity_map` assuming it matches the query.

Maybe `is_matching_creation` or `matches_query` would be a better name for this function. And the comment should also be changed.

graph/src/components/store/entity_cache.rs
```diff
@@ -203,18 +202,94 @@ impl EntityCache {
...
+        // Iterate over self.updates to find entities that:
+        // - Aren't already present in the entity_map
+        // - Match the query
+        // If these conditions are met:
+        // - Check if there's an update for the same entity in handler_updates and apply it.
+        // - Add the entity to entity_map.
+        for (key, op) in self.updates.iter() {
+            if !entity_map.contains_key(key) {
+                if let Some(entity) = apply_update(op, &query, key)? {
+                    if let Some(handler_op) = self.handler_updates.get(key) {
```
lutter, line 266: If an entity is in both `self.updates` and `self.handler_updates`, that means it was created and then modified. In that case, don't we have to apply the `op` from `self.handler_updates` to the entity from `self.updates` just as we do in the case for entities that already exist in the store? I think that's where the name `apply_update` is misleading.

graph/src/components/store/entity_cache.rs
```diff
@@ -203,18 +202,94 @@ impl EntityCache {
...
+                        // If there's an update from handler_updates, apply it
+                        if let Some(updated_entity) = apply_update(handler_op, &query, key)? {
+                            entity_map.insert(key.clone(), updated_entity);
+                            continue; // Move to the next iteration since the entity is already updated and inserted
```
lutter, line 270: This `continue` is odd; shouldn't the `insert` below just go into an `else` branch?

lutter (changes requested):

graph/src/components/store/entity_cache.rs
incrypto32: `apply_update` oh thats right, exactly that bad naming that i did confused me as well. Just modified the name, comment and addresses the update from `handler_op` in this case

graph/src/components/store/entity_cache.rs
```diff
@@ -203,18 +202,111 @@ impl EntityCache {
...
+        for (key, handler_op) in self.handler_updates.iter() {
+            if !entity_map.contains_key(key) && !self.updates.contains_key(key) {
+                if let Some(entity) = matches_query(handler_op, &query, key)? {
+                    entity_map.insert(key.clone(), entity);
+                }
+            }
+        }
+
+        // Remove keys that were marked for removal
+        for key in keys_to_remove {
```
incrypto32, line 305: The removal is placed at last because for the stuff that has been removed we don't need to reprocess it in the `updates` and `handler_updates` section (which only care about stuff that is not in `entity_map`. So if the removal is immediatly after the processing entities from store `updates` section will reapply the changes unneccessarily. The alternate is to convert `keys_to_remove` a `HashSet` and have another condition `!keys_to_remove.contains()` in the `updates` and `handler_updates` section when iterating over them

graph/src/components/store/entity_cache.rs
```diff
@@ -203,18 +202,111 @@ impl EntityCache {
...
-        entities.iter().for_each(|(key, e)| {
-            self.current.insert(key.clone(), Some(e.clone()));
-        });
-        let entities: Vec<Entity> = entities.values().cloned().collect();
-        Ok(entities)
+        let mut entity_map = self.store.get_derived(&query)?;
+
+        for (key, entity) in entity_map.iter() {
+            // Only insert to the cache if it's not already there
+            // This is to avoid overwriting updates
```
lutter, line 215: Since `self.current` only contains entities that have already been written to the store, there's no danger of clobbering an update. It's still good to add those entities to `self.current`, it's just that the comment is misleading.

graph/src/components/store/entity_cache.rs
lutter: I agree with the reasoning; it would be good to capture some of that in the comment, something like
```
// Remove entities that are in the store but have been removed by an update. 
// We do this last since the loops over updates and handler_updates are only
// concerned with entities that are not in the store yet and by leaving removed 
// keys in entity_map we avoid processing these updates a second time when we
// already looked at them when we went through entity_map
```

lutter (approved):

graph/src/components/store/entity_cache.rs
incrypto32: in that case should we just insert it without checking ? Is there a reason to keep the check?

incrypto32:
@lutter Any idea why the runner tests are failing with `PoisonError` ? it failed every single time for this PR. But works fine when i run locally

incrypto32:
The runners test issue was a yarn workspace issue. fixed it

Issue #4466, opened 2023-03-16
Mappings can set non-existent Entity attributes


lutter:
Mappings can store entries in an entity that do not correspond to attributes on the underlying type. These extra entries will be silently discarded when the entity is stored, but they do influence the POI calculation. That makes it possible that a subgraph generates a divergent POI depending on whether such an entity is fetched from the internal cache (where these attributes are still present) or from the database.

To fix this, we can take two approaches:

1. Fail the subgraph when an entity has extraneous attributes
2. Silently remove these attributes before doing the POI calculation

I would prefer (1) because this generally indicates some error in the subgraph's logic, but we'd need to figure out what level of backwards compatibility we need to provide (cc @azf20)

azf20:
Is there any way to identify whether a subgraph is doing this, without checking the mappings themselves?

lutter:
No, it requires looking at the source and seeing if they use `store.set` directly, i.e. without going through the generated wrappers.

PR #4807, opened 2023-08-16
Validate field exists in schema before setting entity

### LINKED ISSUE #4791: [Bug] Validate entities before creating Proof of Indexing
### Bug report

When investigating a POI divergence between two graph-node versions, a polygon subgraph `QmWTuqjkyGogHMBuGvQHZjkb189CN4EtGi8ps1yT76Vj9V` diverges at block `45504938`, On further investigation by checking logs the only difference found was a `Proof of Indexing` event log 

Screenshot 2023-08-04 at 9 05 05 PM

The `HourlyVolumeBySource` does not have  a field named `marginAndLiquidation` but still that field is being set to 0 in this case.

Here is the [schema](https://api.thegraph.com/ipfs/api/v0/cat?arg=QmfBAQwYcabyiQ9BH5txv8yQiMnBR9V7bba1nLyXRqiRqt) of this subgraph. There is another entity `FeeStat` which has `marginAndLiquidation` entity

### Relevant log output

_No response_

### IPFS hash

QmWTuqjkyGogHMBuGvQHZjkb189CN4EtGi8ps1yT76Vj9V

_No response_

### OS information

None

incrypto32:
Closes #4791 #4466

### DIFF
--- a/graph/src/data/store/mod.rs
+++ b/graph/src/data/store/mod.rs
@@ -9,15 +9,16 @@ use crate::{
 };
 use crate::{data::subgraph::DeploymentHash, prelude::EntityChange};
 use anyhow::{anyhow, Error};
+use graphql_parser::schema::ObjectType;
 use itertools::Itertools;
 use serde::de;
 use serde::{Deserialize, Serialize};
 use stable_hash::{FieldAddress, StableHash, StableHasher};
-use std::borrow::Cow;
 use std::convert::TryFrom;
 use std::fmt;
 use std::str::FromStr;
 use std::sync::Arc;
+use std::{borrow::Cow, collections::HashSet};
 use strum_macros::IntoStaticStr;
 
 use super::{
@@ -845,6 +846,31 @@ impl Entity {
             }
         }
 
+        fn validate_missing_fields<'a>(
+            field_names_iter: impl Iterator<Item = &'a str>,
+            object_type: &ObjectType<'_, String>,
+        ) -> Result<(), Error> {
+            let fields_set: HashSet<_> =
+                object_type.fields.iter().map(|f| f.name.as_str()).collect();
+
+            // Fields that are not in `object_type.fields`
+            let missing_fields: Vec<_> = field_names_iter
+                .filter(|field_name| !fields_set.contains(field_name))
+                .collect();
+
+            // If `missing_fields` is not empty that means this entity
+            // has fields that are not defined in the schema
+            if !missing_fields.is_empty() {
+                Err(anyhow!(
+                    "The provided entity has fields not defined in the schema for entity `{}`: {}",
+                    object_type.name,
+                    missing_fields.join(", "),
+                ))
+            } else {
+                Ok(())
+            }
+        }
+
         if key.entity_type.is_poi() {
             // Users can't modify Poi entities, and therefore they do not
             // need to be validated. In addition, the schema has no object
@@ -857,6 +883,8 @@ impl Entity {
                 key.entity_type, key.entity_id, key.entity_type
             )
         })?;
+
+        validate_missing_fields(self.0.iter().map(|(k, _)| k), object_type)?;
 
         for field in &object_type.fields {
             let is_derived = field.is_derived();
--- a/graph/src/data/store/mod.rs
+++ b/graph/src/data/store/mod.rs
@@ -14,11 +14,11 @@ use itertools::Itertools;
 use serde::de;
 use serde::{Deserialize, Serialize};
 use stable_hash::{FieldAddress, StableHash, StableHasher};
+use std::borrow::Cow;
 use std::convert::TryFrom;
 use std::fmt;
 use std::str::FromStr;
 use std::sync::Arc;
-use std::{borrow::Cow, collections::HashSet};
 use strum_macros::IntoStaticStr;
 
 use super::{
@@ -850,12 +850,14 @@ impl Entity {
             field_names_iter: impl Iterator<Item = &'a str>,
             object_type: &ObjectType<'_, String>,
         ) -> Result<(), Error> {
-            let fields_set: HashSet<_> =
-                object_type.fields.iter().map(|f| f.name.as_str()).collect();
-
             // Fields that are not in `object_type.fields`
             let missing_fields: Vec<_> = field_names_iter
-                .filter(|field_name| !fields_set.contains(field_name))
+                .filter(|field_name| {
+                    !object_type
+                        .fields
+                        .iter()
+                        .any(|field| field.name.as_str() == *field_name)
+                })
                 .collect();
 
             // If `missing_fields` is not empty that means this entity
--- a/graph/src/data/store/mod.rs
+++ b/graph/src/data/store/mod.rs
@@ -649,27 +649,59 @@ pub trait TryIntoEntityIterator<E>: IntoIterator<Item = Result<(Word, Value), E>
 
 impl<E, T: IntoIterator<Item = Result<(Word, Value), E>>> TryIntoEntityIterator<E> for T {}
 
-#[derive(Debug, Error, PartialEq, Eq)]
+#[derive(Debug, Error, PartialEq, Eq, Clone)]
 pub enum EntityValidationError {
-    #[error("The provided entity has fields not defined in the schema for entity `{0}`: {1}")]
-    FieldsNotDefined(String, String),
-
-    #[error("Entity {0}[{1}]: unknown entity type `{0}`")]
-    UnknownEntityType(String, String),
-
-    #[error("Entity {0}[{1}]: field `{2}` is of type {3}, but the value `{4}` contains a {5} at index {6}")]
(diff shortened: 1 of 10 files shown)


lutter (approved):

graph/src/data/store/mod.rs
```diff
@@ -846,6 +847,31 @@ impl Entity {
             }
         }
 
+        fn validate_fields<'a>(
```
mangas, line 850: IMO this name is ambiguous, would prefer something like check_missing_fields or validate_missing_fields, since it's all it does and this way you can easily glean this information from the name

graph/src/data/store/mod.rs
```diff
@@ -846,6 +847,31 @@ impl Entity {
...
+            field_names_iter: impl Iterator<Item = &'a str>,
+            object_type: &ObjectType<'_, String>,
+        ) -> Result<(), Error> {
+            let fields_set: HashSet<_> =
+                object_type.fields.iter().map(|f| f.name.as_str()).collect();
```
mangas, line 855: I think this might be overkill for most entities which are probably small (a bit arbitrary) but this function seems to be called potentially twice per `set` operation on the entity cache . I think it may be better to either consider storing the fields on a hashset to begin with at the entity level, or ensure they are ordered by name and just lookup on the vec which *should* be a lot cheaper for smaller entities.

mangas: cc @lutter for an opinion

runtime/test/src/test.rs
```diff
@@ -1389,3 +1390,143 @@ async fn test_store_set_id() {
...
+    let err = host
+        .store_set(
+            USER,
+            UID,
+            vec![("id", "u1"), ("name", "user1"), ("test3", "invalid_field")],
+        )
+        .err()
+        .unwrap();
+
+    err_says(
```
mangas, line 1528: Could be worth creating a ValidationError with the current anonymous errors and check on types if you feel up for that

graph/src/data/store/mod.rs
```diff
@@ -846,6 +847,31 @@ impl Entity {
...
+
+            // Fields that are not in `object_type.fields`
+            let missing_fields: Vec<_> = field_names_iter
```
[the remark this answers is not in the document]
incrypto32: We are failing anyways. So why not fail with a better error message? 
If there are no missing fields we would have to iterate through the entire thing anyways.

graph/src/data/store/mod.rs
incrypto32: thanks will make that change

graph/src/data/store/mod.rs
mangas: it may be a more complicated change but possibly sorting entries in the Object from intern.rs could be useful for this since we could have `field_names_iter` sorted which means that looking up is amortized and probably cheaper than creating a HashSet + a Vec for each validation

mangas: nitpick: The naming here could also be improved. Something like `set_fields(_iter)` instead of `field_names_iter` and `object_schema` rather than `object_type`would feel a bit easier IMO but obviously it's subjective and nitpicky

graph/src/data/store/mod.rs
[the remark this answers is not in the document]
mangas: That's a good point 👍

incrypto32:
Refactored the test a bit to avoid code duplication for the implementation of the mocked `Host`

graph/src/data/store/mod.rs
lutter: I am not sure how much of an issue this will be in the end, but I am also a little concerned about the impact of creating a `HashSet` all the time .. I think using `.contains()` will be preferable since object types generally have a small number of fields and so its likely that a quadratic algorithm here won't hurt much.

Another option would be to precompute the `HashSet` when the `InputSchema` is created, store them in a map `type name -> HashSet` and use those for this check.

A third option is to store a `Vec` of atoms for each object type in the `InputSchema` and compare against that - since you are only comparing integers at that point, and the `Vec` will fit into a cache line in almost all cases, that's probably the fastest way to do that.

incrypto32: How would option 3 work @lutter?
add a new Map<ObjectType, Atom> and store the Atom corresponding to the fields.
Then wouldn't we need to look up the Atom from the pool first with the `field_name` only then we could compare right? 

I'd go with 1 for now and make the changes to use the `.contains()` let me know if 3 would be worth it. I'll do that if so.

graph/src/data/store/mod.rs
```diff
@@ -647,6 +649,38 @@ pub trait TryIntoEntityIterator<E>: IntoIterator<Item = Result<(Word, Value), E>
 
 impl<E, T: IntoIterator<Item = Result<(Word, Value), E>>> TryIntoEntityIterator<E> for T {}
 
+#[derive(Debug, Error, PartialEq, Eq)]
+pub enum EntityValidationError {
```
incrypto32, line 653: @mangas I added new `EntityValidationError` enum and migrated all the validation error to use it.

graph/src/data/store/mod.rs
```diff
@@ -647,6 +649,38 @@ pub trait TryIntoEntityIterator<E>: IntoIterator<Item = Result<(Word, Value), E>
...
+    #[error("The provided entity has fields not defined in the schema for entity `{0}`: {1}")]
+    FieldsNotDefined(String, String),
```
mangas, line 655: in most of these, I feel like having named fields would be a lot better, 

```
    #[error(
        "The provided entity has fields not defined in the schema for entity `{entity}`: {field}"
    )]
    FieldsNotDefined { entity: String, field: String },

```

graph/src/data/store/mod.rs
```diff
@@ -647,6 +649,38 @@ pub trait TryIntoEntityIterator<E>: IntoIterator<Item = Result<(Word, Value), E>
...
+    #[error("Entity {0}[{1}]: unknown entity type `{0}`")]
+    UnknownEntityType(String, String),
+
+    #[error("Entity {0}[{1}]: field `{2}` is of type {3}, but the value `{4}` contains a {5} at index {6}")]
+    MismatchedElementTypeInList(String, String, String, String, String, String, usize),
+
+    #[error(
+        "Entity {0}[{1}]: the value `{2}` for field `{3}` must have type {4} but has type {5}"
+    )]
+    InvalidFieldType(String, String, String, String, String, String),
```
mangas, line 666: especially this one xD

runtime/test/src/test.rs
mangas: Here you should now be able to check the error returned rather than the message error, this is why the enum is useful

store/postgres/src/fork.rs
```diff
@@ -220,7 +221,11 @@ query Query ($id: String) {{
             map
         };
 
-        Ok(Some(schema.make_entity(map)?))
+        Ok(Some(
+            schema
+                .make_entity(map)
+                .map_err(|e| StoreError::Unknown(anyhow!(e)))?,
```
mangas, line 227: I think it's probably worth adding something like an EntityValidationError value 

something along the lines of 

#[error("Entity validation failed: {0}")}
EntityValidationError(#[from] graph::whatever::EntityValidationError)

mangas: since we do know what kind of error it is in this case

graph/src/data/store/mod.rs
incrypto32: Thanks for catching those, I was blindly following how error struct from other parts of the [code](https://github.com/graphprotocol/graph-node/blob/48e0958e505b95a0d59ee6b3ef3d874cacda97eb/graph/src/schema/mod.rs#L69).

But this one clearly needs to be named. Thanks will keep an eye for such things going forward

runtime/test/src/test.rs
incrypto32: The problem is its no longer `EntityValidationError` anymore it get anonymised in between an only the message gets passed into `HostExportError:Deterministic`

mangas: 🤦‍♂️ okay, oh well

store/postgres/src/fork.rs
incrypto32: changed

graph/src/data/store/mod.rs
lutter: yeah, I think going with (1) will be fine for now; what I had in mind for (3) was a `Map<ObjectType, Vec<Atom>>` where the `Vec<Atom>` contains all the field names for the object type.

lutter: One other small complication: the object type will also have the names of derived fields, even though they can't be set directly. Maybe we should precompute that map for (3) and only include non-derived fields for each type.

incrypto32:
Rebased with master

mangas (approved):

Issue #4794, opened 2023-08-07
[Bug] Regression: Only Last 2 Fulltext Tags Work


Faolain:
### Bug report

Hey all, 

Our users reported that our search functionality no longer worked on our frontend https://dev.marketplace.dclimate.net/ around May 31 which was strange to us as we had not touched that part of our codebase for some time and our subgraph had also not changed. Worst of all we were not seeing this issue locally. We had not gotten a chance to investigate more deeply until recently whereupon @eloramirez1356 did a deep dive and discovered that this issue was not being seen locally (deploying local graph and using local graph node) because our local docker graph node version was fixed to `image: graphprotocol/graph-node:6499d38` and that a change to the graph node since that image has caused this regression to surface on our deployed graph on the hosted service (and this image likely is used on the decentralized service too?) where **Only the Last 2 Fulltext Tags Work on a Schema**.

## Background:
We have the following schema with 4 @fulltext tags which powers our search/filtering functionality, basically the title and description of assets which get published to the graph. 

```
type _Schema_
  @fulltext(
    name: "assetSearch"
    language: en
    algorithm: rank
    include: [
      { entity: "Asset", fields: [{ name: "title" }, { name: "description" }] }
    ]
  )
    @fulltext(
    name: "subscriptionAssetSearch"
    language: en
    algorithm: rank
    include: [
      { entity: "SubscriptionAsset", fields: [{ name: "title" }, { name: "description" }] }
    ]
  )
  @fulltext(
    name: "geohashSearch"
    language: en
    algorithm: rank
    include: [{ entity: "Asset", fields: [{ name: "geohash" }] }]
  )
  @fulltext(
    name: "subscriptionGeohashSearch"
    language: en
    algorithm: rank
    include: [{ entity: "SubscriptionAsset", fields: [{ name: "geohash" }] }]
  )
  ```

This was previously working when deployed to the *hosted service*. This was also working locally. At some point around May 31st this stopped working on the hosted service. After some debugging by @eloramirez1356 it was discovered that

> the first 2 tags (fulltext searches) don't work, but the last 2 works as expected, and if I vary the order of these tags in the schema, doing a new deployment of the schema with the different order, only the last 2 fulltext tags are the ones that works as expected.

So essentially it seems that a graph node image update which occurred around May 31st creates a regression where only the last 2 Fulltext tags work. 

Old Graph Image (Working with 4 Fulltext Queries):
image: graphprotocol/graph-node:6499d38

[...]

PR #4808, opened 2023-08-17
store: Populate all search indexes for an entity

lutter:
In commit ad1c6ead, we inadvertently made it so that only one search index per entitiy would be populated. With these changes, all of them will be populated again.

Fixes https://github.com/graphprotocol/graph-node/issues/4794


incrypto32 (approved):
LGTM

mangas (approved):

PR #4809, opened 2023-08-17
Make `LoadManager` take shards into account

### LINKED ISSUE #4262: Make the load manager work for sharded setups
The [`LoadManager`](https://github.com/graphprotocol/graph-node/blob/master/graph/src/data/graphql/effort.rs#L189) tries to determine when the system enters an overloaded state and drops queries in that case. Unfortunately, the data it bases its decisions on was never fully updated to take multiple shards into account and it seems that it therefore never notices when a shard has high load.

lutter:
The `LoadManager` was never updated for sharded setups. It now tracks query effort for each shard separately, and makes decisions about dropping queries per shard, too.

In addition, queries are now tracked by deployment and shape hash rather than just by shape hash so that identical queries for different deployments, which can vary greatly in the effort they cause, are not confused with each other

Fixes #4262

### DIFF
--- a/graph/src/data/graphql/load_manager.rs
+++ b/graph/src/data/graphql/load_manager.rs
@@ -212,7 +212,8 @@ pub struct LoadManager {
     /// there is no way for a query to get out of jail other than
     /// restarting the process
     jailed_queries: RwLock<HashSet<QueryRef>>,
-    kill_state: RwLock<KillState>,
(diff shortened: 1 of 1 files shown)


mangas (approved):

lutter:
This is merged - somehow github got confused and marks this as closed

PR #4811, opened 2023-08-21
Remove the entry from `subgraph_features` table when a subgraph is removed

incrypto32:
Closes https://github.com/graphprotocol/graph-node/issues/4793

### DIFF
--- a/store/postgres/src/primary.rs
+++ b/store/postgres/src/primary.rs
@@ -1334,10 +1334,11 @@ impl<'a> Connection<'a> {
             .map(|_| ())
     }
 
-    /// Remove all subgraph versions and the entry in `deployment_schemas` for
-    /// subgraph `id` in a transaction
+    /// Remove all subgraph versions, the entry in `deployment_schemas` and the entry in
+    /// `subgraph_features` for subgraph `id` in a transaction
     pub fn drop_site(&self, site: &Site) -> Result<(), StoreError> {
         use deployment_schemas as ds;
+        use subgraph_features as f;
         use subgraph_version as v;
         use unused_deployments as u;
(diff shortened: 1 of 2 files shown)


lutter:
A better way to do that would be to have a foreign key constraint on `subgraph_features(id) references deployment_schemas(subgraph) on delete cascade` in the database. That way, Postgres will clean up.

Also, in the hosted service there are 40k entries in the `subgraph_features` table but only 28k entries in `deployment_schemas` - this change should therefore also come with a migration that removes unneeded entries before creating the fk constraint, something like (untested)
```
delete from subgraph_features f 
where not exists (select 1 from deployment_schemas ds where ds.subgraph = f.id)
```

It also seems that the `subgraph_features` table still needs to be documented in the [metadata docs](https://github.com/graphprotocol/graph-node/blob/master/docs/implementation/metadata.md)

incrypto32:
> better way to do that would be to have a foreign key constraint on subgraph_features(id) references deployment_schemas(subgraph) on delete cascade

@lutter the `subgraph` column on `deployment_schemas` is not a unique field right so it wont be possible to create a fk constraint? Since same deployment_hash can be on different shards under different id its not possible to add a unique constraint too?

Would it be worth creating a postgres trigger instead ?

incrypto32:
> it also seems that the subgraph_features table still needs to be documented in the metadata docs

I added the docs in [#4820](https://github.com/graphprotocol/graph-node/pull/4820)

lutter (approved):
Forget what I said about a fk constraint; you are right, since we don't have a unique index on `deployment_schemas.subgraph`, we can't do that here.

Issue #4724, opened 2023-06-27
[Bug] Spurious issue with subgraph deployments


paymog:
### Bug report

We have an end to end test suite which does many subgraph deployments. We're find that our tests spuriously fail with `subgraph validation error: [the specified block must exist on the Ethereum network]`. In our test suite, we always deploy subgraphs with against mainnet with start blocks in the range of 0-100. 

We can't reliably reproduce the issue but do find it happens often on different tests. Our API calls the `subgraph_deploy` on graph node and this is the full message that's returned: `subgraph validation error: [the specified block must exist on the Ethereum network]`.

We're running graph node v0.31.0. 

### Some information to help us out

- [ ] Tick this box if this bug is caused by a regression found in the latest release.
- [ ] Tick this box if this bug is specific to the hosted service.
- [X] I have searched the issue tracker to make sure this issue is not a duplicate.

### OS information

None

paymog:
Found some relevant logs from our e2e tests:

```
e2e-tests-graph-node-1   | Jul 07 07:50:18.580 INFO Received subgraph_create request, params: SubgraphCreateParams { name: SubgraphName("cljsa19l3001jc6joh17r1gq4") }, component: JsonRpcServer
e2e-tests-graph-node-1   | Jul 07 07:50:18.588 DEBG Created subgraph, subgraph_name: cljsa19l3001jc6joh17r1gq4, component: SubgraphRegistrar
e2e-tests-graph-node-1   | Jul 07 07:50:18.590 INFO Received subgraph_deploy request, params: SubgraphDeployParams { name: SubgraphName("cljsa19l3001jc6joh17r1gq4"), ipfs_hash: DeploymentHash("QmX2wBD6mB5CcqXkp9XP1HfF9LrdKwFSA5eddXVj8sYSTB"), node_id: None, debug_fork: None, history_blocks: None }, component: JsonRpcServer
e2e-tests-graph-node-1   | Jul 07 07:50:19.164 ERRO subgraph_deploy failed, params: SubgraphDeployParams { name: SubgraphName("cljsa19l3001jc6joh17r1gq4"), ipfs_hash: DeploymentHash("QmX2wBD6mB5CcqXkp9XP1HfF9LrdKwFSA5eddXVj8sYSTB"), node_id: None, debug_fork: None, history_blocks: None }, error: ManifestValidationError([BlockNotFound("100")]), component: JsonRpcServer
```

[...]

PR #4812, opened 2023-08-21
Retry block_pointer_from_number on deploy

mangas:
Add retry to avoid a single failure while making the call prevent the subsgraph from being deployed Add warning for failed attempts

Closes https://github.com/graphprotocol/graph-node/issues/4724

### DIFF
--- a/core/src/subgraph/registrar.rs
+++ b/core/src/subgraph/registrar.rs
@@ -13,6 +13,9 @@ use graph::prelude::{
     CreateSubgraphResult, SubgraphAssignmentProvider as SubgraphAssignmentProviderTrait,
     SubgraphRegistrar as SubgraphRegistrarTrait, *,
 };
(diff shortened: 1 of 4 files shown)


core/src/subgraph/registrar.rs
```diff
@@ -518,18 +521,38 @@ async fn resolve_start_block(
...
+    // Apply limit (maximum retry count)
+    match limit_opt {
+        Some(limit) => {
+            // Items are delays *between* attempts,
+            // so subtract 1 from limit.
+            Box::new(backoff.take(limit - 1))
+        }
+        None => Box::new(backoff),
+    }
+}
```
neysofu, line 555: This looks identical to `graph::util::futures::retry_strategy`, correct? If so, we can make that public and reuse it here, and also avoid `pub use tokio_retry`.

neysofu (approved):

PR #4820, opened 2023-08-23
Track handlers in subgraph_features table

### LINKED ISSUE #4772: [Feature] Track handlers in `subgraph_features` table
### Description

It would be nice to have insights into which handlers are used by subgraphs.
eg: If a subgraph uses a call handler or not, How many subgraphs use blockHandlers with call filter etc.

It would also be beneficial to know if the handlers are in a template or a normal datasource.

_No response_

incrypto32:
Closes #4772

### DIFF
--- a/chain/arweave/src/data_source.rs
+++ b/chain/arweave/src/data_source.rs
@@ -12,13 +12,15 @@ use graph::{
     },
     semver,
 };
+use std::collections::HashSet;
 use std::sync::Arc;
 
 use crate::chain::Chain;
 use crate::trigger::ArweaveTrigger;
 
 pub const ARWEAVE_KIND: &str = "arweave";
-
+const BLOCK_HANDLER_KIND: &str = "block";
+const TRANSACTION_HANDLER_KIND: &str = "transaction";
 /// Runtime representation of a data source.
 #[derive(Clone, Debug)]
 pub struct DataSource {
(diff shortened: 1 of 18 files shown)


chain/arweave/src/data_source.rs
```diff
@@ -47,6 +49,20 @@ impl blockchain::DataSource<Chain> for DataSource {
         self.source.start_block
     }
 
+    fn handler_kinds(&self) -> HashSet<&str> {
+        let mut kinds = HashSet::new();
```
incrypto32, line 53: Wondering if this should be a `HashSet` or just a `Vec`. But in `ethereum` we have blockHandlers with filters so its needed to iterate over it and add it to a HashSet to avoid duplication.

At the end way up in the call path, we convert this to a HashSet anyway.

chain/substreams/src/data_source.rs
```diff
@@ -80,6 +80,11 @@ impl blockchain::DataSource<Chain> for DataSource {
         None
     }
 
+    fn handler_kinds(&self) -> HashSet<&str> {
+        // This is placeholder, substreams do not have a handler kind.
+        vec![SUBSTREAMS_HANDLER_KIND].into_iter().collect()
```
incrypto32, line 85: This was added as a placeholder to prevent having using an `Option`

store/postgres/migrations/2023-08-23-143628_add_handlers_to_subgraph_features/up.sql
```diff
@@ -0,0 +1,6 @@
+truncate table subgraphs.subgraph_features;
```
incrypto32, line 1: truncating the table so that the new columns values will get populated, `subgraph_features` table is populated during `build_subgraph_runner` so it will repopulate all the entries on startup

docs/implementation/metadata.md
```diff
@@ -141,3 +141,16 @@ correctly across index node restarts.
 
 The table `subgraphs.table_stats` stores which tables for a deployment
 should have the 'account-like' optimization turned on.
+
+### `subgraphs.subgraph_features`
+
+Details about features that a deployment uses
```
lutter, line 147: This should indicate that that table is stored in the primary (I think that's where it lives, right?)

incrypto32: Indicated that.

lutter (approved):
Looks good!

Issue #4261, opened 2022-12-20
Optimize lookup of certain child ids


lutter:
For GraphQL queries that only retrieve the id of children where the parents store the child ids, we should change query generation to optimize that query away entirely.

Those are queries of the form
```
parent(..) {
  children { id }
}
```
where the `children` in the parent type are not derived.

lutter:
This is actually not possible because of a subtlety of how we store relationships and don't enforce foreign key constraints: with a schema like
```
type Thing {
  id Bytes!
  stuff Stuff
}

type Stuff {
  id Bytes!
}
```

it is possible to store a `Thing` whose `stuff` field points to `Stuff` that doesn't actually exist. That issue is only detected at query time where we'd either return `stuff: null` or, if the `stuff` field is actually of type `Stuff!` fail the query with an error `null value resolved for non-null field`

Because of that, we can't just return the content of `Thing.stuff` as the `id` for the related `Stuff` field - we do have to go to the database and check whether there is `Stuff` with that id, making it impossible to optimize that part of the query away.

PR #4827, opened 2023-08-24
Avoid redundant child queries

lutter:
If a query only needs to get the `id` of children, and we already have them  in memory, do not run a query to fetch them again. Instead, construct the appropriate objects in memory.
    
Fixes https://github.com/graphprotocol/graph-node/issues/4261


mangas (approved):

lutter:
@mangas Thanks for approving - I added one more commit to have a feature flag to turn all this off. Not sure if you want to have another look.

PR #4830, opened 2023-08-28
Allow creating partial indexes with graphman

### LINKED ISSUE #4581: [Feature] Create partial indexes
Since a lot of queries happen close to the subgraph head, queries can benefit from partial indexes that only include data from recent blocks. In SQL, such an index is created by adding a clause `where coalesce(upper(block_range), 2147483647) > $recent_block` to the `create index` command. For some queries, a partial index can improve performance considerably, similar to the effect that pruning has.

The `graphman index create` command should be enhanced to take a flag `--after $recent_block` to create a partial index instead of an index on the whole table.

To make that index usable, queries against such table need to have a clause `coalesce(upper(block_range), 2147483647) > $block` added to them, similar how the account-like optimization changes query generation. We also need to make sure that this addition to the query does not worsen query performance.

### LINKED ISSUE #4665: [Feature] `graphman index create` does not respect prefixed fields
### Description

For fields of type `String` and `Bytes` we only index a prefix of the value (except for the `id` attribute) 

With the GraphQL schema
```
type Thing {
  id Bytes!
  name Bytes!
  age Int!
}
```
filters that involve `name` will actually use `substring(name, 1, 64)` in queries. The default attribute index on `name` takes that into account and is an index on `substring(name, 1, 64)`

But when trying to build a combined index with `graphman index create sgdNN thing name age` the index will be built on `sgdNNN.thing(name, age)` but that index is useless for queries and has to be built on `sgdNNN.thing(substring(name, 1, 64), age)`

It is also not possible to run something like `graphman index create sgdNN thing 'substring(name, 1, 64)' age` as `graphman` will reject that.

_No response_

incrypto32:
- [x] Allow creating partial indexes with graphman Closes #4581 
- [x] Index only a prefix for types `String` and `Bytes` Closes #4665

### DIFF
--- a/store/postgres/src/deployment_store.rs
+++ b/store/postgres/src/deployment_store.rs
@@ -726,7 +726,15 @@ impl DeploymentStore {
             let column_names_sep_by_underscores = column_names.join("_");
             let index_exprs = resolve_index_exprs(column_names_to_types);
             let table_name = &table.name;
-            let index_name = format!("manual_{table_name}_{column_names_sep_by_underscores}");
+            let index_name = format!(
+                "manual_{table_name}_{column_names_sep_by_underscores}{}",
+                if let Some(after_value) = after {
+                    format!("_{}", after_value)
+                } else {
+                    String::new()
+                }
+            );
+
             let mut sql = format!(
                 "create index concurrently if not exists {index_name} \
                  on {schema_name}.{table_name} using {index_method} \
@@ -740,6 +748,11 @@ impl DeploymentStore {
                         " where coalesce(upper({}), 2147483647) > {}",
                         BLOCK_RANGE_COLUMN, after
                     ));
+                } else {
+                    return Err(CancelableError::Error(StoreError::Unknown(anyhow!(
+                        "Partial index not allowed on immutable table `{}`",
+                        table_name
+                    ))));
                 }
             }
 
--- a/store/postgres/src/deployment_store.rs
+++ b/store/postgres/src/deployment_store.rs
@@ -709,6 +709,7 @@ impl DeploymentStore {
         entity_name: &str,
         field_names: Vec<String>,
         index_method: Method,
+        after: Option<BlockNumber>,
     ) -> Result<(), StoreError> {
         let store = self.clone();
         let entity_name = entity_name.to_owned();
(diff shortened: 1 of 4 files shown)


store/postgres/src/deployment_store.rs
```diff
@@ -704,23 +709,40 @@ impl DeploymentStore {
...
+            let mut sql = format!(
                 "create index concurrently if not exists {index_name} \
                  on {schema_name}.{table_name} using {index_method} \
-                 ({column_names_sep_by_commas})"
+                 ({index_exprs}) ",
             );
+
+            // If 'after' is provided and the table is not immutable, add a WHERE clause for partial indexing
+            if let Some(after) = after {
+                if !table.immutable {
```
incrypto32, line 738: Not sure what needs to be done for entities that are immutable? Return an error? or just ignore the partial indexing and log something that suggests --after is being ignored?

lutter: I think for immutable tables we want to produce an error - these partial indexes are meant to reduce the amount of data the db has to look at to find entities that are visible after block `after`, but immutable entities are visible forever, and such an index doesn't make sense to them. We'd just want `graphman` to provide some error to tell the user that they tried something nonsensical.

store/postgres/src/deployment_store.rs
```diff
@@ -1853,21 +1878,37 @@ fn resolve_column_names<'a, T: AsRef<str>>(
...
+                format!("left({}, {})", name, STRING_PREFIX_SIZE)
+            }
+            Some(ColumnType::Bytes) => {
+                format!("substring({}, 1, {})", name, BYTE_ARRAY_PREFIX_SIZE)
+            }
+            _ => name.to_string(),
+        })
+        .collect::<Vec<String>>()
+        .join(",")
+}
```
lutter, line 1910: Nice!

lutter (commented):
This looks really good so far!

store/postgres/src/deployment_store.rs
```diff
@@ -704,23 +709,40 @@ impl DeploymentStore {
...
+            let column_names = column_names_to_types
+                .iter()
+                .map(|(name, _)| *name)
+                .collect::<Vec<&SqlName>>();
+
             let column_names_sep_by_underscores = column_names.join("_");
-            let column_names_sep_by_commas = column_names.join(", ");
+            let index_exprs = resolve_index_exprs(column_names_to_types);
             let table_name = &table.name;
             let index_name = format!("manual_{table_name}_{column_names_sep_by_underscores}");
```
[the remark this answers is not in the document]
incrypto32: Made those changes

store/postgres/src/deployment_store.rs
incrypto32: Ty, Made the change

lutter (approved):

PR #4839, opened 2023-08-29
Add substreams block ingestor

mangas:
- Substreams ingestor not currently allowed for supported chains
- Chains using substreams protocol can only have substreams endpoints
- Add substream-head-tracker to the project
- Embed substreams-head-tracker in substreams block ingestor


substreams-head-tracker/rust-toolchain.toml
```diff
@@ -0,0 +1,4 @@
+[toolchain]
```
maoueh, line 1: Would we have a way to use the top-level one directly. Only thing I'm wondering is the `wasm32-unknown-unknown` target.

node/src/config.rs
```diff
@@ -530,6 +530,31 @@ impl Chain {
...
+            if has_only_substreams_providers {
+                bail!(
+                    "{} protocol requires an rpc or firehose endpoint defined",
+                    self.protocol
+                );
+            }
+        }
+
+        // When using substreams protocol, only substreams endpoints are allowed
+        if matches!(self.protocol, BlockchainKind::Substreams) {
```
neysofu, line 548: This is equivalent to an `else`, correct?

neysofu (approved):
I don't think I have the necessary context to thoroughly review this, so it's great to have other eyes on this. This is pretty in line with what I would expect though, LGTM.

node/src/config.rs
mangas: Yes, this was mostly to make it clearer, if you think it isn't I can refactor to improve

neysofu: No need, it's fine!

PR #4840, opened 2023-08-30
set sslmode paramater in dockerfile

DaMandal0rian:
Enable ssl mode for postgres to be configurable with other options in connection string

### DIFF
--- a/docker/start
+++ b/docker/start
@@ -65,7 +65,7 @@ run_graph_node() {
     else
         unset GRAPH_NODE_CONFIG
         postgres_port=${postgres_port:-5432}
-        postgres_url="postgresql://$postgres_user:$postgres_pass@$postgres_host:$postgres_port/$postgres_db?sslmode=prefer"
+        postgres_url="postgresql://$postgres_user:$postgres_pass@$postgres_host:$postgres_port/$postgres_db?$postgres_args"
 
         wait_for_ipfs "$ipfs"
         echo "Waiting for Postgres ($postgres_host:$postgres_port)"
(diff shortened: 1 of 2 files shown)


docker/start
```diff
@@ -65,7 +65,7 @@ run_graph_node() {
     else
         unset GRAPH_NODE_CONFIG
         postgres_port=${postgres_port:-5432}
-        postgres_url="postgresql://$postgres_user:$postgres_pass@$postgres_host:$postgres_port/$postgres_db?sslmode=prefer"
+        postgres_url="postgresql://$postgres_user:$postgres_pass@$postgres_host:$postgres_port/$postgres_db?sslmode=${postgres_sslmode}"
```
mangas, line 68: can you please add a general purpose args env var instead of one specific to the sslmode? 

my suggestion would be something like 
`ENV postgres_args "sslmode=prefer"`

this way it can be overwritten with any number of options available for the postgres conn string 

similar to [this usage](https://github.com/graphprotocol/graph-node/blob/master/.github/workflows/ci.yml#L71)

DaMandal0rian: That is for a postgres image, we are simply providing the connection string. Extra arguments should ne set in postgres image.

mangas: I understand what that is for, I'm referring to the pattern of allowing the user to set more than one parameter in a single var. The user can, for example, set: 

postgres_args=`sslmode=require,requirepeer=postgres` (or any other options, like the one related to certificates) without having to change anything else.

mangas: the way it's done now can be misused to do so by doing postgres_sslmode="require,requirepeer=postgres", it's just not very user friendly

DaMandal0rian: @mangas gotcha!

mangas (approved):

Issue #4719, opened 2023-06-26
[Bug] Cross shard graft never starts


paymog:
### Bug report

I have a subgraph deployed in my primary shard. A new subgraph was deployed into my secondary shard which grafts off the first subgraph. The copy operation for the graft never starts and no relevant logs are printed as far as I can tell. When I check `graphman copy list` I see the following:

```
deployment           | QmR4EG5LiCR1zjHFw1x8o6ExkUbYKa1GMNRWyBLzapj9ED
action               | sgd18355 -> sgd19620 (second)
started              | 2023-06-26T12:58:00+00:00
progress             | 0.00% done, 0/1532
```

I've checked my replication slot lag and I have very little lag.

I'm running v0.31.0.

### Relevant log output

EDIT: I reproduced this in a staging environment after trying to [reproduce locally](https://github.com/graphprotocol/graph-node/issues/4719#issuecomment-1609243299).

```
Jun 27 10:51:46.548 INFO Received subgraph_create request, params: SubgraphCreateParams { name: SubgraphName("clje6444w0igt49ri4v9w8skj") }, component: JsonRpcServer
Jun 27 10:51:46.552 DEBG Created subgraph, subgraph_name: clje6444w0igt49ri4v9w8skj, component: SubgraphRegistrar
Jun 27 10:51:46.556 INFO Received subgraph_deploy request, params: SubgraphDeployParams { name: SubgraphName("clje6444w0igt49ri4v9w8skj"), ipfs_hash: DeploymentHash("QmTHV6jX365VjoZf6g19bNrXR2AfPEQgqtw474hDXRZk5s"), node_id: None, debug_fork: None, history_blocks: None }, component: JsonRpcServer
Jun 27 10:51:46.568 DEBG Connecting to firehose to retrieve block for number 17570214, sgd: 0, subgraph_id: QmTHV6jX365VjoZf6g19bNrXR2AfPEQgqtw474hDXRZk5s, component: SubgraphRegistrar
Jun 27 10:51:46.630 DEBG Retrieving block(s) from firehose, sgd: 0, subgraph_id: QmTHV6jX365VjoZf6g19bNrXR2AfPEQgqtw474hDXRZk5s, component: SubgraphRegistrar
[...]

PR #4845, opened 2023-09-01
store: Map other shards into a new shard on startup

lutter:
Fixes https://github.com/graphprotocol/graph-node/issues/4719

### DIFF
--- a/store/postgres/src/connection_pool.rs
+++ b/store/postgres/src/connection_pool.rs
@@ -1011,13 +1011,7 @@ impl PoolInner {
         let result = pool
             .configure_fdw(coord.servers.as_ref())
             .and_then(|()| migrate_schema(&pool.logger, &conn))
-            .and_then(|had_migrations| {
-                if had_migrations {
-                    coord.propagate_schema_change(&self.shard)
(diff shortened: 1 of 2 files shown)


store/postgres/src/connection_pool.rs
```diff
@@ -1107,12 +1101,31 @@ impl PoolInner {
 
 embed_migrations!("./migrations");
 
+struct MigrationCount {
```
mangas, line 1104: IMO this struct could be a bit more clear , both`had_migrations` and `is_new` sort of require reading the code and usage to understand what is going, I wonder if an enum of NoChanges, NewShard, MigrationsApplied(Int) would make it a bit more clear. It's private and super critical so I will approve and leave it with you to decide if you think there is a benefit or not 🤓

mangas (approved):

store/postgres/src/connection_pool.rs
lutter: I agree that the naming isn't great. I like the idea of making this an `enum` - I'll do that next time I have to touch something in there.

PR #4848, opened 2023-09-04
Support DataSourceContext in manifest

### LINKED ISSUE #2981: Support dataSource context on regular dataSources
A dataSource can have context, but this is not supported as part of the `subgraph.yaml`. Users should be able to add context to their manifest which can be accessed in the runtime

incrypto32:
Closes #2981


incrypto32:
The huge diff is the `yarn.lock`

mangas (approved):

Issue #4831, opened 2023-08-28
[Feature] `graph-node` Substreams powered Subgraph ingestion time could be improved


maoueh:
### Description

When we did the RETH benchmarking a few weeks ago, I had anecdotally tested `graph-node` finding it was really slow. At that time, I think the batched writes was not yet fully active. 

Friday, I decided to make a more formal test of how long it takes for a `graph-node` to sync a Substreams powered Subgraph for which the Substreams is already cached. A cached Substreams is one for which the endpoint already fully processed the `.spkg` and cached the output, so essentially we are simply streaming pre-processed files out on the gRPC stream.

#### Experience

Using https://github.com/streamingfast/substreams-eth-reth-benchmark/blob/master/reth-erc20-rocket-v1.0.6.spkg and ensuring it was cached, validated by verifying the server's cache files for the specific module ensuring all outputs files were actually present.

##### `substreams-sink-postgres`

Using the instructions outlined at https://github.com/streamingfast/substreams-eth-reth-benchmark#instructions, I got the following ingestion time:

- Started at `2023-08-25T14:34:39.519-0400`
- Finished at `2023-08-25T14:41:57.111-0400`
- Took `7m 17s`

This test ran on my laptop, a 1 Gbps connection over WiFi (I'm hitting ~100 Mbps download speed). The database is Postgres 14 ran through Docker Compose on my machine.

The endpoint used was `mainnnet.eth.streamingfast.io:443`.

##### `graph-node`

Using latest master branch (https://github.com/graphprotocol/graph-node/commit/837948a0b193c9ec75908c981984d568ae9ae160), I ran the same experiment deploying the Substreams powered Subgraph using manifest https://github.com/streamingfast/substreams-eth-reth-benchmark/blob/master/subgraph.yaml.

The endpoint configured for the provider in `graph-node` is pointing at `mainnnet.eth.streamingfast.io:443`.

- Started at `Aug 25 14:51:36.341`
- Finished at `Aug 26 02:06:56.169`
- Tool `11h 15m 19s`

##### `substreams-sink-rust`

This is a small Rust project that essentially extracted https://github.com/graphprotocol/graph-node/blob/master/graph/src/blockchain/substreams_block_stream.rs and `endpoint.rs` into a simple tutorial for people willing to consume their Substreams with Rust code.

The code can be checked at https://github.com/streamingfast/substreams-sink-rust.

It took me `8m 33s` using this code to stream back the `.spkg` above

[...]

PR #4851, opened 2023-09-06
Move substreams processing to block stream

mangas:
Improves perfromance based on [this issue](https://github.com/graphprotocol/graph-node/issues/4831)

### DIFF
--- a/chain/substreams/src/chain.rs
+++ b/chain/substreams/src/chain.rs
@@ -3,12 +3,11 @@ use crate::{data_source::*, EntityChanges, TriggerData, TriggerFilter, TriggersA
 use anyhow::Error;
 use graph::blockchain::client::ChainClient;
 use graph::blockchain::{
-    BasicBlockchainBuilder, BlockIngestor, BlockchainBuilder, EmptyNodeCapabilities,
-    NoopRuntimeAdapter,
+    BasicBlockchainBuilder, BlockIngestor, EmptyNodeCapabilities, NoopRuntimeAdapter,
 };
-use graph::components::store::DeploymentCursorTracker;
+use graph::components::store::{DeploymentCursorTracker, EntityKey};
 use graph::firehose::FirehoseEndpoints;
-use graph::prelude::{BlockHash, CheapClone, LoggerFactory, MetricsRegistry};
+use graph::prelude::{BlockHash, CheapClone, Entity, LoggerFactory, MetricsRegistry};
 use graph::{
     blockchain::{
         self,
@@ -20,13 +19,29 @@ use graph::{
     prelude::{async_trait, BlockNumber, ChainStore},
     slog::Logger,
 };
+
 use std::sync::Arc;
+
+// ParsedChanges are an internal representation of the equivalent operations defined on the
+// graph-out format used by substreams.
+// Unset serves as a sentinel value, if for some reason an unknown value is sent or the value
+// was empty then it's probably an unintended behaviour. This code was moved here for performance
+// reasons, but the validation is still performed during trigger processing so while Unset will
+// very likely just indicate an error somewhere, as far as the stream is concerned we just pass
+// that along and let the downstream components deal with it.
+#[derive(Debug, Clone)]
+pub enum ParsedChanges {
+    Unset,
+    Delete(EntityKey),
+    Upsert { key: EntityKey, entity: Entity },
+}
 
 #[derive(Default, Debug, Clone)]
 pub struct Block {
     pub hash: BlockHash,
     pub number: BlockNumber,
     pub changes: EntityChanges,
+    pub parsed_changes: Vec<ParsedChanges>,
 }
 
 impl blockchain::Block for Block {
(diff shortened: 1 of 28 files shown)


chain/substreams/src/trigger.rs
```diff
@@ -189,95 +183,37 @@ where
         _subgraph_metrics: &Arc<graph::prelude::SubgraphInstanceMetrics>,
         _instrument: bool,
     ) -> Result<BlockState<Chain>, MappingError> {
-        for entity_change in block.changes.entity_changes.iter() {
-            match entity_change.operation() {
-                Operation::Unset => {
+        for parsed_change in block.parsed_changes.clone().into_iter() {
```
maoueh, line 186: This `clone` seems at first sight superflous. In each arm of the match, you either use a reference or clone, I didn't see any move in there.

mangas: I will check but I think this is because block is a &Arc so you can't move it out of there to call into_iter

maoueh: Seems that `state.entity_cache.remove(entity_key);` is the only move. I think cloning just this one would be better. I leave to you the decision.

maoueh: Scratch all my comment above, I just saw `state.entity_cache.set(key, entity)?;` which is a full move, I missed it because it was part of non-modified code. Sorry :)

mangas: Yeah that's it I was just re-checking and the set requires the move so we might as well clone it all ahead of time. No problem at all, thanks for checking!

chain/substreams/src/mapper.rs
```diff
@@ -1,13 +1,27 @@
...
+use graph::data_source::CausalityRegion;
+use graph::prelude::BigDecimal;
+use graph::prelude::{async_trait, BigInt, BlockHash, BlockNumber, BlockPtr, Logger, Value};
+use graph::schema::InputSchema;
 use graph::substreams::Clock;
 use graph::substreams_rpc::response::Message as SubstreamsMessage;
 use prost::Message;
 
-pub struct Mapper {}
+pub struct Mapper {
```
lutter, line 22: It would be good to explain here what this is for, and how setting `schema` to `None` influences things

chain/substreams/src/mapper.rs
```diff
@@ -89,3 +113,237 @@ impl SubstreamsMapper<Chain> for Mapper {
...
+                        .or_insert(Value::Null) = value;
+                }
+                let entity = schema
+                    .make_entity(parsed_data)
+                    .map_err(anyhow::Error::from)?;
+
+                ParsedChanges::Upsert { key, entity }
+            }
+            entity_change::Operation::Delete => ParsedChanges::Delete(key),
+            entity_change::Operation::Unset => ParsedChanges::Unset,
```
lutter, line 176: I am not familiar with substreams vocab, but what does `Unset` mean in this context? Do we even need to keep that in the parsed changes? Might be good to add that as a comment to `ParsedChanges`

chain/substreams/src/chain.rs
```diff
@@ -20,13 +19,22 @@ use graph::{
...
     slog::Logger,
 };
+
 use std::sync::Arc;
 
+#[derive(Debug, Clone)]
+pub enum ParsedChanges {
+    Unset,
+    Delete(EntityKey),
+    Upsert { key: EntityKey, entity: Entity },
```
lutter, line 29: No more intermediate `HashMap` .. yay!

core/src/subgraph/runner.rs
```diff
@@ -723,7 +725,7 @@ where
         for trigger in triggers {
             // Using an `EmptyStore` and clearing the cache for each trigger is a makeshift way to
             // get causality region isolation.
-            let schema = self.inputs.store.input_schema();
+            let schema = ReadStore::input_schema(&self.inputs.store);
```
lutter, line 728: Why was that change necessary?

lutter (approved):

core/src/subgraph/runner.rs
mangas: I added the `input_schema` function to another trait which caused this call to be ambiguous.

chain/substreams/src/mapper.rs
mangas: will do, Unset is just to preserve the code logic that's applied to protobuf. Unset there is the default value when a new unsupported value is used. Say you have an enum with 3 values,  if the server has a 4th value then when parsing it will default to Unset, works as a sentinel value of sorts. 

In this case, at parsing we don't "care" about what operation it is, we're just transforming so there is a 1:1 mapping of supported operations. Once we run the trigger processing then there will be an error if the operation is `Unset` as that's not a valid operation. I didn't want to change the responsibility of the block stream so this was the "easiest" way to maintain it

chain/substreams/src/mapper.rs
mangas: will do!

chain/substreams/src/mapper.rs
mangas: I will also add a comment here to make it clearer

PR #4853, opened 2023-09-06
graph, graphql: Expand which fields we optimize away

lutter:
This is a follow-on to https://github.com/graphprotocol/graph-node/pull/4827 to cover the important case where no ordering is specified (for example, for single-value child fields)


incrypto32 (approved):

lutter:
Closing this for the reasons explained in #4261

PR #4856, opened 2023-09-10
Add rootless Docker container support

computeronix:
By default the container will run as root

This will force the container, in prod, to run as rootless user, (default user is 1000)

feel free to change the name of the user as needed

Please further test, this has worked for me for over a year but please have others test to ensure the execution is unaffected by permissions

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -97,9 +97,9 @@ EXPOSE 8020
 EXPOSE 8030
 
 RUN apt-get update \
-    && apt-get install -y libpq-dev ca-certificates netcat \
+    && apt-get install -y libpq-dev ca-certificates netcat whois \
     && useradd -m graphuser \
-    && echo "graphuser:graphuser" | chpasswd
+    && echo "graphuser:$(mkpasswd -m sha-512 -s somesalt4325)" | chpasswd
 
 ADD docker/wait_for docker/start /usr/local/bin/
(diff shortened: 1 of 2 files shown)


docker/Dockerfile
```diff
@@ -97,11 +97,14 @@ EXPOSE 8020
 EXPOSE 8030
 
 RUN apt-get update \
-    && apt-get install -y libpq-dev ca-certificates netcat
+    && apt-get install -y libpq-dev ca-certificates netcat whois \
+    && useradd -m graphuser \
+    && echo "graphuser:$(mkpasswd -m sha-512 -s somesalt4325)" | chpasswd
```
mangas, line 102: This is not necessary right?

docker/docker-compose.yml
```diff
@@ -2,6 +2,7 @@ version: '3'
 services:
   graph-node:
     image: graphprotocol/graph-node
+    user: "1000:1000"
```
mangas, line 5: also not sure this is going to work until the upstream build was produced by the file above

computeronix:
Corrected the password change as your right that is not needed.  The user will be added without a password locking the account (but allowing the account to be used on start-up with the USER command from docker)

This is better security this way too

For the compose file -> if the previous file is not there and you specify 1000:1000 it will show noname as the username but still run in context without root

PR #4868, opened 2023-09-15
add trace_id to substreams logger

mangas:


chain/substreams/src/mapper.rs
```diff
@@ -13,10 +14,14 @@ pub struct Mapper {}
...
     async fn to_block_stream_event(
         &self,
-        logger: &Logger,
+        logger: &mut Logger,
         message: Option<SubstreamsMessage>,
     ) -> Result<Option<BlockStreamEvent<Chain>>, SubstreamsError> {
         match message {
+            Some(SubstreamsMessage::Session(session_init)) => {
+                *logger = logger.new(o!("trace_id" => session_init.trace_id));
+                return Ok(None);
```
maoueh, line 23: Let's add right away also a log line saying that `session is initialized` logging every pieces found in `SessionInit`.

maoueh: I just saw that some new fields in `SessionInit` are not re-surfaced, let's not delay that more and will had in another PR.

Issue #4756, opened 2023-07-13
[Feature] Do not cache invalid eth_call results in case node returned invalid results.


SozinM:
### Description

During the implementation of https://github.com/graphprotocol/graph-node/pull/1270 it was discussed that we probably should not cache invalid data.
We experienced this issue with the following graph-node logs:
```
 Jul 11 15:11:32.995 DEBG eth_call, data: a290bf38000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000005ab, address: d2863157539b1d11f39ce23fc4834b62082f6874, sgd: 271, subgraph_id: QmW7j8ipQ8cbf8agCLYjRgJkxwUaS9F8Xa46XkCfSFPUxW, component: SubgraphInstanceManager
 Jul 11 15:11:33.010 INFO Contract call reverted, reason: empty response, sgd: 271, subgraph_id: QmW7j8ipQ8cbf8agCLYjRgJkxwUaS9F8Xa46XkCfSFPUxW, component: SubgraphInstanceManager
```
The eth call is on block 44939006.
This was caused by the node outage (that happened at the same time) which caused eth_call to return an invalid result.
After that eth_call was cached and we needed to clean the eth_call call for this block.

CC @leoyvens @Jannis 

### Some information to help us out

- [X] Tick this box if you plan on implementing this feature yourself.
- [X] I have searched the issue tracker to make sure this issue is not a duplicate.

balakhonoff:
Just my two cents, it is really hard to track that some wrong node responses lead to the wrong cache, and we know about it only from the customers, because only they know that the code which shouldn't raise any errors is actually raising them. And the descriptions of errors always pointing to wrong reasons which is confusing.

SozinM:
Curl with request:
```
curl -H "Content-Type: application/json" -d '{"method":"eth_call","params": [{"data":"0xa290bf38000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000005ab","gas":"0x2faf080","to":"0xd2863157539b1d11f39ce23fc4834b62082f6874"},{"blockHash":"0xf0862352a99eab45a31d5952419c7004e4b6da9fe441ca94782499107977abc0"}],"id":1,"jsonrpc":"2.0"}' <endpoint>
```

azf20:
hey @SozinM what chain is that on?

[...]

PR #4879, opened 2023-09-19
Cache only correct eth_call data

SozinM:
This PR closes https://github.com/graphprotocol/graph-node/issues/4756 
It changes caching logic to cache only correct and decodable eth_calls to prevent cache from providing faulty data

### DIFF
--- a/chain/ethereum/src/ethereum_adapter.rs
+++ b/chain/ethereum/src/ethereum_adapter.rs
@@ -1303,14 +1303,21 @@ impl EthereumAdapterTrait for EthereumAdapter {
                         .map(move |result| {
                             // Don't block handler execution on writing to the cache.
                             let for_cache = result.0.clone();
-                            let _ = graph::spawn_blocking_allow_panic(move || {
-                                cache
-                                    .set_call(call.address, &call_data, call.block_ptr, &for_cache)
-                                    .map_err(|e| {
-                                        error!(logger, "call cache set error";
-                                                   "error" => e.to_string())
-                                    })
-                            });
+                            if !result.0.is_empty() {
+                                let _ = graph::spawn_blocking_allow_panic(move || {
+                                    cache
+                                        .set_call(
+                                            call.address,
+                                            &call_data,
+                                            call.block_ptr,
+                                            &for_cache,
+                                        )
+                                        .map_err(|e| {
+                                            error!(logger, "call cache set error";
+                                                       "error" => e.to_string())
+                                        })
+                                });
+                            }
                             result.0
                         }),
                     )


chain/ethereum/src/ethereum_adapter.rs
```diff
@@ -1300,19 +1300,7 @@ impl EthereumAdapterTrait for EthereumAdapter {
...
-                                cache
-                                    .set_call(call.address, &call_data, call.block_ptr, &for_cache)
-                                    .map_err(|e| {
-                                        error!(logger, "call cache set error";
-                                                   "error" => e.to_string())
-                                    })
-                            });
-                            result.0
-                        }),
+                        .map(move |result| result.0),
```
SozinM, line 1303: I feel like this could take a bit of refactoring, but I'm not sure it's possible to make it more nice looking

chain/ethereum/src/ethereum_adapter.rs
```diff
@@ -1276,7 +1276,8 @@ impl EthereumAdapterTrait for EthereumAdapter {
             "block_hash" => call.block_ptr.hash_hex(),
             "block_number" => call.block_ptr.block_number()
         );
-
+        let cache_logger = logger.clone();
+        let cache = cache.clone();
```
SozinM, line 1280: I would like to move it to more deep scope, but I cannot wrap my head on how to correctly clone it in and_then future.

chain/ethereum/src/ethereum_adapter.rs
```diff
@@ -1326,9 +1314,18 @@ impl EthereumAdapterTrait for EthereumAdapter {
                 } else {
                     // Decode failures are reverts. The reasoning is that if Solidity fails to
                     // decode an argument, that's a revert, so the same goes for the output.
-                    call.function.decode_output(&output).map_err(|e| {
+                    let result = call.function.decode_output(&output).map_err(|e| {
                         EthereumContractCallError::Revert(format!("failed to decode output: {}", e))
-                    })
+                    });
+                    let _ = graph::spawn_blocking_allow_panic(move || {
```
SozinM, line 1320: I think I could move it to dedicated and_then future using https://docs.rs/futures-async-combinators/latest/futures_async_combinators/future/fn.and_then.html
But I didn't see it usage in the graph so I'm hesitant to add this

lutter (changes requested):
First off, congrats on making it through this code - it's really gnarly and would hugely benefit from asyncifying (though that might be quite a bit of an undertaking)

Before merging this, we need to find a way to not call `set_call` for calls that are already in the cache as that will have a bad effect on performance.

azf20:
hey @SozinM, thanks so much for this PR! Did you manage to take a look at David's comments?

SozinM:
Hi!
Yes, I did saw
I will fix them once I return to work (in dec)

PR #4881, opened 2023-09-21
Pass polling triggers only after start_block

incrypto32:


mangas (approved):
As per our conversation I'm approving this so it can be included in the release but I think these conditions are not super easy to visually parse so I'd rather move the logic into a function and add some comments of what it is doing, what's interval and some testing

incrypto32:
https://github.com/graphprotocol/graph-node/issues/4885

Opened this issue for addressing the tests and extracting the condition

PR #4883, opened 2023-09-21
Handle `EntityType` and entity ids more clearly and safely

lutter:
We used to treat both `EntityType` and ids as just strings, which lead to situations where it was very unclear what either of them meant and how it had to be converted. In addition, we had to do lots of lookups for ancillary information against things like the `InputSchema` which made code harder to follow than it has to be. It also has caused hard-to-spot bugs in the past, especially around conversion of strings to bytes and vice versa.

With these changes, entity types and ids are converted into a better internal representation at the point where we take them in from the outside (mappings, substreams, database) with checks that what we take in actually makes sense.

The `EntityType` is now properly checked against the `InputSchema` when it is first constructed, and can be used to look up ancillary information like the type of the `id` field directly without needing to have an `InputSchema` around.

Similarly, the `Id` type now wraps entity ids and carries type information with it making it much less likely that code changes get confused about the actual type they need to deal with.

A lot of the commits in this PR are pretty boring (my apologies to the reviewer), and either just moving methods around or slightly changing method signatures. None of them should cause a functional change.

### DIFF
--- a/graph/src/data/store/mod.rs
+++ b/graph/src/data/store/mod.rs
@@ -931,7 +931,7 @@ impl Entity {
         for field in self.0.atoms() {
             if !schema.has_field(&key.entity_type, field) {
                 return Err(EntityValidationError::FieldsNotDefined {
-                    entity: key.entity_type.clone().into_string(),
+                    entity: key.entity_type.to_string(),
                 });
             }
         }
--- a/graph/src/data/store/mod.rs
+++ b/graph/src/data/store/mod.rs
@@ -28,7 +28,7 @@ use super::{
 
 /// Handling of entity ids
 mod id;
-pub use id::IdType;
+pub use id::{Id, IdList, IdRef, IdType};
 
 /// Custom scalars in GraphQL.
 pub mod scalar;
@@ -806,12 +806,8 @@ impl Entity {
     /// string. If it is `Bytes`, return it as a hex string with a `0x`
     /// prefix. If the ID is not set or anything but a `String` or `Bytes`,
     /// return an error
-    pub fn id(&self) -> Word {
-        match self.get("id") {
-            Some(Value::String(s)) => Word::from(s.clone()),
-            Some(Value::Bytes(b)) => Word::from(b.to_string()),
-            None | Some(_) => unreachable!("we checked the id when constructing this entity"),
-        }
+    pub fn id(&self) -> Id {
+        Id::try_from(self.get("id").unwrap().clone()).expect("the id is set to a valid value")
     }
 
     /// Merges an entity update `update` into this entity.
@@ -1047,7 +1043,7 @@ impl std::fmt::Debug for Entity {
 /// possibly a pointer to its parent if the query that constructed it is one
 /// that depends on parents
 pub struct QueryObject {
-    pub parent: Option<r::Value>,
+    pub parent: Option<Id>,
     pub entity: r::Object,
 }
 
--- a/graph/src/data/store/mod.rs
+++ b/graph/src/data/store/mod.rs
@@ -26,6 +26,10 @@ use super::{
     value::Word,
 };
 
+/// Handling of entity ids
+mod id;
+pub use id::IdType;
+
 /// Custom scalars in GraphQL.
 pub mod scalar;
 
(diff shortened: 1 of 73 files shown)


graphql/src/execution/ast.rs
```diff
@@ -351,7 +350,7 @@ pub(crate) fn resolve_object_types(
         .ok_or_else(|| QueryExecutionError::AbstractTypeError(name.to_string()))?
     {
         s::TypeDefinition::Interface(intf) => {
-            for obj_ty in &schema.types_for_interface()[&EntityType::new(intf.name.to_string())] {
+            for obj_ty in &schema.types_for_interface()[&intf.name] {
```
mangas, line 353: Is there some risk that this could cause a panic?

graphql/src/store/query.rs
```diff
@@ -39,8 +39,7 @@ pub(crate) fn build_query<'a>(
             let selected_columns = column_names.get(object);
             vec![((*object).into(), selected_columns)]
         }
-        ObjectOrInterface::Interface(interface) => types_for_interface
-            [&EntityType::from(*interface)]
+        ObjectOrInterface::Interface(interface) => types_for_interface[&interface.name]
```
mangas, line 48: same here

graph/src/schema/input_schema.rs
```diff
@@ -79,11 +81,13 @@ impl InputSchema {
                         .iter()
                         .map(|field| pool.lookup(&field.name).unwrap())
                         .collect();
-                    (EntityType::from(obj_type), fields)
-                }),
+                    EntityType::new(&pool, &obj_type.name).map(|t| (t, fields))
+                })
+                .collect::<Result<Vec<_>, _>>()
+                .unwrap(),
```
mangas: if this unwrap is just because EntityType::new can't fail then I think it's probably better to just unwrap the new, similar to how the lookup is also unwrapped, IMO it would make it slightly clearer even though it still makes me a bit nervous to see all the unwraps 😅

graph/src/schema/input_schema.rs
```diff
@@ -79,11 +81,13 @@ impl InputSchema {
...
         );
 
-        let poi_type = EntityType::new(POI_OBJECT.to_string());
+        let poi_type = EntityType::new(&pool, POI_OBJECT).unwrap();
```
mangas: maybe it could be better to just have some of these functions separately like `new_poi_object` which uses the const internally, leaks less details and could avoid the unwrap

graph/src/data/store/mod.rs
```diff
@@ -182,6 +182,26 @@ pub enum IdType {
     Bytes,
 }
 
+impl TryFrom<&s::ObjectType> for IdType {
+    type Error = Error;
+
+    fn try_from(obj_type: &s::ObjectType) -> Result<Self, Self::Error> {
+        let base_type = obj_type.field("id").unwrap().field_type.get_base_type();
```
mangas: should "id" just be a const?

mangas (approved):

lutter:
I ran this in the integration cluster which turned up no issues. I also used the query validation tool, which pointed out one issue which I addressed in commit 6a2fa56

graphql/src/execution/ast.rs
lutter: In theory, it could, though we've had this there for a long time w/o issues, and the changes in the code make it less likely to panic.

store/postgres/src/relational_queries.rs
```diff
@@ -548,15 +552,29 @@ impl EntityData {
                                         "query unexpectedly produces parent ids"
                                     )))
                                 }
-                                Some(parent_type) => Some(
-                                    T::Value::from_column_value(parent_type, json)
-                                        .map(|value| (PARENT_ID.clone(), value)),
-                                ),
+                                Some(parent_type) => {
+                                    Some(T::Value::from_column_value(parent_type, json))
+                                }
```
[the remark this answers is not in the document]
lutter: Good point. Fixed

graph/src/data/store/mod.rs
lutter: yep; I changed that in a few places now

PR #4894, opened 2023-09-26
Add new API Version to validate when setting fields not defined in the schema 

incrypto32:


runtime/wasm/src/host_exports.rs
```diff
@@ -199,13 +200,48 @@ impl<C: Blockchain> HostExports<C> {
...
+                    })
+                    .collect();
+
+                return Err(HostExportError::Deterministic(anyhow!(
+                    "Entity `{}` has fields not in schema: {:?}",
+                    key.entity_type,
+                    invalid_fields
+                )));
+            }
+        }
```
lutter, line 241: Could the block starting at line 203 be put into a `check_invalid_fields` helper function? It would make `store_set` more readable.

lutter (approved):
Nice!

lutter:
> @lutter this has conflicts after you merged the recent PR. If its ok to include that in the release I'll rebase ( i have rebased locally ). Else we can merge this into the branch `incrypto32/v0.33.0` after the review.

That's unfortunate - I think you'll have to put this onto a release branch, and then rebase it to merge it into master.

PR #4895, opened 2023-09-27
feat: initial Starknet support

xJonathanLEI:
This PR brings initial support for Starknet into `graph-node`. The implementation comes from [starknet-graph](https://github.com/starknet-graph), and has been used in production by the [zkLend](https://zklend.com/) team for around a year. That said, the support enabled by this PR is not production ready:

- Trigger filter is not implemented, and `graph-node` would always consume the entire Firehose block stream and build all possible triggers. No benchmark has been done yet, but this should hurt performance pretty bad.
- Data source templates are not supported. This hasn't been tested yet, but technically `graph-node` would panic if someone deploys a subgraph with Starknet data source templates defined.
- Substream support has not been added (will take inspiration from #4887).

Nevertheless, this PR is submitted here first to kickstart the integration process. Let's bring `graph-node`'s amazing indexing capabilities to Starknet!

### DIFF
--- a/chain/starknet/Cargo.toml
+++ b/chain/starknet/Cargo.toml
@@ -12,7 +12,8 @@ hex = { version = "0.4.3", features = ["serde"] }
 prost = { workspace = true }
 prost-types = { workspace = true }
 serde = "1.0"
-starknet-core = "0.6.0"
+sha3 = "0.10.8"
+starknet-ff = "0.3.4"
 
 graph-runtime-wasm = { path = "../../runtime/wasm" }
 graph-runtime-derive = { path = "../../runtime/derive" }
--- a/chain/starknet/Cargo.toml
+++ b/chain/starknet/Cargo.toml
@@ -1,0 +1,18 @@
+[package]
+name = "graph-chain-starknet"
+version.workspace = true
+edition.workspace = true
+
+[build-dependencies]
+tonic-build = { workspace = true }
+
+[dependencies]
+graph = { path = "../../graph" }
+hex = { version = "0.4.3", features = ["serde"] }
+prost = { workspace = true }
+prost-types = { workspace = true }
+serde = "1.0"
+starknet-core = "0.6.0"
+
+graph-runtime-wasm = { path = "../../runtime/wasm" }
+graph-runtime-derive = { path = "../../runtime/derive" }
--- a/chain/starknet/src/adapter.rs
+++ b/chain/starknet/src/adapter.rs
@@ -1,0 +1,32 @@
+use graph::blockchain::{EmptyNodeCapabilities, TriggerFilter as TriggerFilterTrait};
+
+use crate::{
+    data_source::{DataSource, DataSourceTemplate},
+    Chain,
+};
+
+// Currently `TriggerFilter` is useless, as it doesn't really filter anything from the block
+// stream, and the complete stream is always consumed, regardless of what the subgraph is indexing.
+// This is very bad for indexing performance and must be implemented for it to be considered
+// production-ready.
+// TODO: implement trigger filter for much better performance
+#[derive(Default, Clone)]
+pub struct TriggerFilter;
+
+impl TriggerFilterTrait<Chain> for TriggerFilter {
+    #[allow(unused)]
+    fn extend_with_template(&mut self, data_source: impl Iterator<Item = DataSourceTemplate>) {
+        todo!()
+    }
+
+    #[allow(unused)]
+    fn extend<'a>(&mut self, data_sources: impl Iterator<Item = &'a DataSource> + Clone) {}
+
+    fn node_capabilities(&self) -> EmptyNodeCapabilities<Chain> {
+        todo!()
+    }
+
+    fn to_firehose_filter(self) -> Vec<prost_types::Any> {
+        todo!()
+    }
+}
--- a/chain/starknet/src/adapter.rs
+++ b/chain/starknet/src/adapter.rs
@@ -5,11 +5,6 @@ use crate::{
     Chain,
 };
 
-// Currently `TriggerFilter` is useless, as it doesn't really filter anything from the block
-// stream, and the complete stream is always consumed, regardless of what the subgraph is indexing.
-// This is very bad for indexing performance and must be implemented for it to be considered
-// production-ready.
-// TODO: implement trigger filter for much better performance
 #[derive(Default, Clone)]
 pub struct TriggerFilter;
 
--- a/chain/starknet/src/chain.rs
+++ b/chain/starknet/src/chain.rs
@@ -55,15 +55,6 @@ pub struct FirehoseMapper {
 
 pub struct TriggersAdapter;
 
-// impl Chain {
-//     pub fn new(
-//         logger_factory: LoggerFactory,
-//         name: String,
-//         registry: Arc<dyn MetricsRegistry>,
-//         firehose_endpoints: FirehoseEndpoints,
-//         chain_store: Arc<dyn ChainStore>,
-//         block_stream_builder: Arc<dyn BlockStreamBuilder<Self>>,
-// ) -> Self {
 impl BlockchainBuilder<Chain> for BasicBlockchainBuilder {
     fn build(self, _config: &Arc<EnvVars>) -> Chain {
         Chain {
@@ -383,7 +374,6 @@ impl TriggersAdapterTrait<Chain> for TriggersAdapter {
         panic!("Should never be called since not used by FirehoseBlockStream")
     }
 
-    // Used for reprocessing blocks when creating a data source.
     #[allow(unused)]
     async fn triggers_in_block(
         &self,
@@ -391,15 +381,13 @@ impl TriggersAdapterTrait<Chain> for TriggersAdapter {
         block: codec::Block,
         filter: &crate::adapter::TriggerFilter,
     ) -> Result<BlockWithTriggers<Chain>, Error> {
-        // We can't do any filtering here as `TriggerFilter` is useless.
-        // TODO: implement trigger filtering once `TriggerFilter` actually does something.
-
         let shared_block = Arc::new(block.clone());
 
         let mut triggers: Vec<_> = shared_block
             .transactions
             .iter()
             .flat_map(|transaction| -> Vec<StarknetTrigger> {
+                let transaction = Arc::new(transaction.clone());
                 transaction
                     .events
                     .iter()
@@ -407,7 +395,7 @@ impl TriggersAdapterTrait<Chain> for TriggersAdapter {
                         StarknetTrigger::Event(StarknetEventTrigger {
                             event: Arc::new(event.clone()),
                             block: shared_block.clone(),
-                            transaction: Arc::new(transaction.clone()),
+                            transaction: transaction.clone(),
                         })
                     })
                     .collect()
--- a/chain/starknet/src/chain.rs
+++ b/chain/starknet/src/chain.rs
@@ -1,0 +1,439 @@
+use graph::{
+    anyhow::Result,
+    blockchain::{
+        block_stream::{
+            BlockStream, BlockStreamBuilder, BlockStreamEvent, BlockWithTriggers, FirehoseCursor,
+            FirehoseError, FirehoseMapper as FirehoseMapperTrait,
+            TriggersAdapter as TriggersAdapterTrait,
+        },
+        client::ChainClient,
+        firehose_block_ingestor::FirehoseBlockIngestor,
+        firehose_block_stream::FirehoseBlockStream,
+        BasicBlockchainBuilder, Block, BlockIngestor, BlockPtr, Blockchain, BlockchainBuilder,
+        BlockchainKind, EmptyNodeCapabilities, IngestorError, NoopRuntimeAdapter,
+        RuntimeAdapter as RuntimeAdapterTrait,
+    },
+    cheap_clone::CheapClone,
+    components::store::{DeploymentCursorTracker, DeploymentLocator},
+    data::subgraph::UnifiedMappingApiVersion,
+    env::EnvVars,
+    firehose::{self, FirehoseEndpoint, ForkStep},
+    prelude::{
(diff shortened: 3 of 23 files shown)


xJonathanLEI:
Just force pushed the PR branch to breaking changes from #4887. The branch is now up to date again.

chain/starknet/src/adapter.rs
```diff
@@ -0,0 +1,32 @@
...
+use crate::{
+    data_source::{DataSource, DataSourceTemplate},
+    Chain,
+};
+
+// Currently `TriggerFilter` is useless, as it doesn't really filter anything from the block
+// stream, and the complete stream is always consumed, regardless of what the subgraph is indexing.
+// This is very bad for indexing performance and must be implemented for it to be considered
+// production-ready.
+// TODO: implement trigger filter for much better performance
```
mangas, line 12: Probably not worth leaving these TODOs in here

chain/starknet/src/chain.rs
```diff
@@ -0,0 +1,439 @@
...
+pub struct StarknetStreamBuilder;
+
+pub struct FirehoseMapper {
+    adapter: Arc<dyn TriggersAdapterTrait<Chain>>,
+    filter: Arc<TriggerFilter>,
+}
+
+pub struct TriggersAdapter;
+
+// impl Chain {
```
mangas, line 58: can you please remove the comment?

chain/starknet/src/chain.rs
```diff
@@ -0,0 +1,439 @@
...
+    ) -> Result<BlockWithTriggers<Chain>, Error> {
+        // We can't do any filtering here as `TriggerFilter` is useless.
+        // TODO: implement trigger filtering once `TriggerFilter` actually does something.
+
+        let shared_block = Arc::new(block.clone());
+
+        let mut triggers: Vec<_> = shared_block
+            .transactions
+            .iter()
+            .flat_map(|transaction| -> Vec<StarknetTrigger> {
```
mangas, line 402: Could you please move the Arc::new(transaction) up so that the clone in every step is a clone of the Arc instead of a clone of the transaction?

chain/starknet/src/chain.rs
```diff
@@ -0,0 +1,439 @@
...
+    // Used for reprocessing blocks when creating a data source.
+    #[allow(unused)]
+    async fn triggers_in_block(
+        &self,
+        logger: &Logger,
+        block: codec::Block,
+        filter: &crate::adapter::TriggerFilter,
+    ) -> Result<BlockWithTriggers<Chain>, Error> {
+        // We can't do any filtering here as `TriggerFilter` is useless.
+        // TODO: implement trigger filtering once `TriggerFilter` actually does something.
```
mangas, line 395: I think this can also be removed

chain/starknet/src/chain.rs
```diff
@@ -0,0 +1,439 @@
...
+    async fn scan_triggers(
+        &self,
+        _from: BlockNumber,
+        _to: BlockNumber,
+        _filter: &crate::adapter::TriggerFilter,
+    ) -> Result<Vec<BlockWithTriggers<Chain>>, Error> {
+        panic!("Should never be called since not used by FirehoseBlockStream")
+    }
+
+    // Used for reprocessing blocks when creating a data source.
```
mangas, line 386: I think this comment is mislead, perhaps taken from runner.rs, where this is somewhat true but more widely this is used to transform the Block data into chain-specific triggers.

chain/starknet/src/codec.rs
```diff
@@ -0,0 +1,32 @@
+#[allow(clippy::all)]
```
mangas, line 1: Do we really need this?

chain/starknet/src/codec.rs
```diff
@@ -0,0 +1,32 @@
...
+#[rustfmt::skip]
+#[path = "protobuf/zklend.starknet.r#type.v1.rs"]
+pub mod pbcodec;
```
mangas, line 4: I think this could be private since we are already exporting the types a few lines below. (which is preferable IMO)

chain/starknet/src/data_source.rs
```diff
@@ -0,0 +1,320 @@
...
+        trigger: &StarknetTrigger,
+        block: &Arc<codec::Block>,
+        _logger: &Logger,
+    ) -> Result<Option<TriggerWithHandler<Chain>>, Error> {
+        if self.start_block() > block.number() {
+            return Ok(None);
+        }
+
+        let handler = match trigger {
+            StarknetTrigger::Block(_) => match self.mapping.block_handlers.first() {
```
mangas, line 130: If there can only be 1 block handler maybe we could change this from a vec to a single handler? Including in the validations?

chain/starknet/src/data_source.rs
```diff
@@ -0,0 +1,320 @@
...
+        Some(self.mapping.runtime.clone())
+    }
+}
+
+impl DataSource {
+    /// Returns event trigger if an event.key matches the handler.key and optionally
+    /// if event.fromAddr matches the source address. Note this only supports the default
+    /// Starknet behavior of one key per event.
+    fn handler_for_event(&self, event: &StarknetEventTrigger) -> Option<MappingEventHandler> {
+        let event_key = FieldElement::from_byte_slice_be(event.event.keys.first()?).ok()?;
```
mangas, line 215: if the event only matches against 1 key maybe we could either filter the rest out on ingress (when decoding)

chain/starknet/src/data_source.rs
```diff
@@ -0,0 +1,320 @@
...
+                    .collect::<Result<Vec<_>, Error>>()?,
+                runtime: Arc::new(module_bytes),
+            },
+        })
+    }
+}
+
+// Starknet data source templates are not supported at the moment. This should be implemented for
+// the Starknet support to be considered production ready.
+// TODO: support Starknet data source template
```
mangas, line 276: remove TODO comment

chain/starknet/src/data_source.rs
```diff
@@ -0,0 +1,320 @@
...
+    }
+
+    fn kind(&self) -> &str {
+        todo!()
+    }
+}
+
+// Starknet data source templates are not supported at the moment. This should be implemented for
+// the Starknet support to be considered production ready.
+// TODO: support Starknet data source template
```
mangas, line 301: remove todo comment

graph/src/blockchain/mod.rs
```diff
@@ -395,6 +395,9 @@ pub enum BlockchainKind {
     Cosmos,
 
     Substreams,
+
+    /// StarkNet chains
```
mangas, line 399: This is not needed

chain/starknet/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+[build-dependencies]
+tonic-build = { workspace = true }
+
+[dependencies]
+graph = { path = "../../graph" }
+hex = { version = "0.4.3", features = ["serde"] }
+prost = { workspace = true }
+prost-types = { workspace = true }
+serde = "1.0"
+starknet-core = "0.6.0"
```
mangas, line 15: Could we get away with using starknet-ff and duplicating the util function? Would be great if we could minimize the amount of deps brought in

chain/starknet/src/codec.rs
xJonathanLEI: Oops must have been mixed into this branch when working on #4169.

chain/starknet/src/codec.rs
xJonathanLEI: Hmmm was following the `near` counterpart haha. Sure I will make it private.

chain/starknet/Cargo.toml
xJonathanLEI: Since we're calculating hashes we will also need `starknet-crypto`. I will update this part too.

chain/starknet/src/data_source.rs
xJonathanLEI: I think this was originally copied from the `near` implementation. The `near` code has been refactored a bit but you can still see this line taking only the first block handler:

https://github.com/graphprotocol/graph-node/blob/53dbe9356ff4515cee08f9a0b92c70161dee11fa/chain/near/src/data_source.rs#L304-L306

and in turn called in `match_and_decode()`:

https://github.com/graphprotocol/graph-node/blob/53dbe9356ff4515cee08f9a0b92c70161dee11fa/chain/near/src/data_source.rs#L139-L142

I'm not sure why it's implemented this way, but I figured I should better follow suit.

chain/starknet/src/data_source.rs
xJonathanLEI: I agree that this is not optimal but I'm not sure what's the best approach for this is. `Event` is managed by `prost` and generated from `graph-as-to-rust`. We can technically change the `firehose-starknet` protobuf format to only emit 1 element, but that seems to be sub-optimal too, as the network technically allows multiple keys.

An alternative would be to create a new struct that resembles `Event`, but instead of using a `Vec` for `keys`, it uses `Option<>` to capture only the first element if there's any. For all other fields we can just move from the old `Event` struct.

But then we're just changing from checking whether the first element exists, to whether the `Option<>` is `Some`. I'm not sure whether this is an improvement.

chain/starknet/src/adapter.rs
xJonathanLEI: Removed.

chain/starknet/src/chain.rs
xJonathanLEI: Removed.

chain/starknet/src/chain.rs
xJonathanLEI: Fixed.

chain/starknet/src/chain.rs
xJonathanLEI: Removed.

chain/starknet/src/chain.rs
xJonathanLEI: Removed.

chain/starknet/src/codec.rs
xJonathanLEI: Removed.

chain/starknet/src/codec.rs
xJonathanLEI: Made private.

chain/starknet/src/data_source.rs
xJonathanLEI: Removed.

chain/starknet/src/data_source.rs
xJonathanLEI: Removed.

graph/src/blockchain/mod.rs
xJonathanLEI: Removed.

chain/starknet/Cargo.toml
xJonathanLEI: Updated with code copied from `starknet-rs`.

chain/starknet/src/data_source.rs
mangas: yeah I saw near does the same, I also don't understand why that decision was made. The main issue is that if this is not validated it's easy for people to think they can define more than one handler and then it will just not run. The UnresolvedMapping has a `block_handlers: Vec...` property which will allow someone to define multiple handlers, if that turns into a single String handler then I think this is fine to keep internally if you prefer that way, although it would be better to just use Option internally in that case.

xJonathanLEI: Hmm for the unresolved part I can totally see that we can't really fix it for chains like `near` without bumping manifest version. For `starknet` though we can totally fix it, both the unresolved and resolved types. Do you prefer if I fix it now and use `Option` for both unresolved and resolved?

chain/starknet/src/data_source.rs
mangas: As far as I can tell the events are not actually passed down to WASM so maybe we could just drop them at the firehose level. Since The AscTransaction type only cares about type and hash then I think ideally we should avoid sending the unnecessary data

mangas: We don't need to hold the PR for that, just ensure we follow up on this

xJonathanLEI: By dropping them at the firsthose level, do you mean that we simply never emit it in `firehose-starknet` in the first place? I always had this impression that a Firehose implementation should be as faithful to the actual network state as possible, due to the quote I read somewhere that "it should be technically feasible to reconstruct an entire archive node from Firehose data along". Dropping them from `firehose-starknet` seems to go against this principle.

Of course, I'm not an expert on this and would make changes as you see fit. Please let me know your final decision on this. Thanks!

xJonathanLEI: Also I don't understand the "events are not sent to WASM" part. Do you mean "event keys"? Cuz I'm quite certain event data are indeed passed?

xJonathanLEI: > We don't need to hold the PR for that, just ensure we follow up on it

Just saw this. Got it. But if we need to change the Firehose representation now would be the best time lol. Once people start using it that becomes a breaking change.

mangas: You are correct, the trigger is indeed passed fully, I was looking for an implement `ToAsc...` for event and didn't see one. Since it's passed to WASM then I think it's fine to keep all the keys, you may match on one but want to use the rest in mappings so let's leave it as is.

xJonathanLEI: Sure. Thanks for confirming!

chain/starknet/src/data_source.rs
mangas: We can merge the PR for now so there's less of a chance for conflicts and then you can fix and add some tests on a separate PR, you can check near/chain.rs's tests, just making sure we have some examples in order to avoid breaking things

xJonathanLEI: Sounds good! Will definitely follow up with another PR on this. As long as we do it before the next release we aren't really breaking anyone after all.

mangas (approved):

PR #4899, opened 2023-09-28
Make it impossible to construct an `InputSchema` that does not validate

lutter:
With these changes, an `InputSchema` is validated as soon as it is constructed, ensuring that any `InputSchema` is actually valid.

This PR also does:
* move the validations for an `InputSchema` from `Schema` to `InputSchema` (they were in `Schema` for bad historical reasons)
* fixes all the schemas we use in tests to be actually valid
* adds a validation for the `id` field; in practice, all schemas pass this validation, but it's good to spell the requirement out explicitly
* adds a small tool to bulk validate `InputSchemas`; that was used on the schemas from the hosted service to make sure that all existing schemas pass the new `id` validation


lutter:
Rebase to latest master

mangas (approved):

PR #4926, opened 2023-10-18
 Allow querying subgraphs with more than one '/' in their name

### LINKED ISSUE #4882: Allow querying subgraphs with more than one '/' in their name
The code [here](https://github.com/graphprotocol/graph-node/blob/master/server/http/src/service.rs#L381) restricts subgraph names to the form `a/b`. That should be changed so that those routes use the rest of a path `/subgraphs/name/<rest>` as the subgraph name. AFAIK, this is the only place in the code where we pay attention to the structure of the name. Everywhere else, we treat it as an opaque string.

This is particularly important as we will start using names with more `/` in them.

incrypto32:
Closes #4882

### DIFF
--- a/server/http/src/service.rs
+++ b/server/http/src/service.rs
@@ -378,14 +378,14 @@ where
             (Method::POST, &["subgraphs", "name", subgraph_name]) => self
                 .handle_graphql_query_by_name(subgraph_name.to_owned(), req)
                 .boxed(),
-            (Method::POST, ["subgraphs", "name", subgraph_name_part1, subgraph_name_part2]) => {
-                let subgraph_name = format!("{}/{}", subgraph_name_part1, subgraph_name_part2);
+            (Method::POST, ["subgraphs", "name", ..]) => {
+                let subgraph_name = path_segments[2..].join("/");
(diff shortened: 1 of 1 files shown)


server/http/src/service.rs
```diff
@@ -378,14 +378,14 @@ where
...
+                let subgraph_name = path_segments[2..].join("/");
                 self.handle_graphql_query_by_name(subgraph_name, req)
                     .boxed()
             }
-            (Method::POST, ["subgraphs", "network", subgraph_name_part1, subgraph_name_part2]) => {
-                let subgraph_name =
-                    format!("network/{}/{}", subgraph_name_part1, subgraph_name_part2);
+            (Method::POST, ["subgraphs", "network", ..]) => {
+                let subgraph_name = path_segments[2..].join("/");
+                let subgraph_name = format!("network/{}", subgraph_name);
```
lutter, line 388: I think that we can get rid of the `network/` routes - they are left over from when we wanted to create a block explorer. But that should happen as a separate PR

lutter (approved):
Nice!

PR #4930, opened 2023-10-20
Graphman deploy command

### LINKED ISSUE #4842: [Feature] Add a `graphman deploy`
### Description

It is sometimes useful to make a deployment available under a different name. Administrators should be able to do this by running `graphman deploy [--create] <name> <ipfs hash>`

If the `--create` flag is passed, the command should fail when `name` already exists, and create a new name and deploy to it if it does not.

_No response_

incrypto32:
Closes #4842

### DIFF
--- a/node/src/bin/manager.rs
+++ b/node/src/bin/manager.rs
@@ -347,12 +347,16 @@ pub enum Command {
 
     // Deploy a subgraph
     Deploy {
+        name: DeploymentSearch,
         deployment: DeploymentSearch,
 
-        #[clap(long, short)]
-        name: String,
+        /// The url of the graph-node
         #[clap(long, short, default_value = "http://localhost:8020")]
         url: String,
+
+        /// Create the subgraph name if it does not exist
+        #[clap(long, short)]
+        create: bool,
     },
 }
 
@@ -1528,7 +1532,13 @@ async fn main() -> anyhow::Result<()> {
             deployment,
             name,
             url,
-        } => commands::deploy::run(deployment, name, url).await,
+            create,
+        } => {
+            let store = ctx.store();
+            let subgraph_store = store.subgraph_store();
+
+            commands::deploy::run(subgraph_store, deployment, name, url, create).await
+        }
     }
 }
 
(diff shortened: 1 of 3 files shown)


node/src/bin/manager.rs
```diff
@@ -344,6 +344,20 @@ pub enum Command {
...
     },
+
+    // Deploy a subgraph
+    Deploy {
+        name: DeploymentSearch,
+        deployment: DeploymentSearch,
+
+        /// The url of the graph-node
+        #[clap(long, short, default_value = "http://localhost:8020")]
+        url: String,
```
lutter, line 355: It would be awesome if we would _not_ need a `url`, but getting away from that might be pretty tough as we would have to instantiate our own `SubgraphRegistrar`; if the default is ok in most cases, I think it's not a big deal.

node/src/manager/commands/deploy.rs
```diff
@@ -0,0 +1,101 @@
...
+            }
+
+            // Send the subgraph_deploy request
+            println!("Deploying subgraph `{}` to `{}`", hash, name);
+            send_deploy_request(&name, &hash, &url).await?;
+            println!("Subgraph `{}` deployed to `{}`", name, url);
+        }
+        Ok(())
+    } else {
+        bail!("Deployment not found");
```
lutter, line 99: Using `DeploymentSearch` here is very clever!

I think the error message is a little terse - we should tell people what exactly they did wrong with the arguments; so maybe something like
```rust
let hash = match deployment {
  DeploymentSearch::Hash { hash, _ } => hash,
  _ => bail!("The `deployment` must be a valid IPFS hash")
};
let name = match name { .. };
if create { .. }
..
```

lutter (approved):

node/src/bin/manager.rs
incrypto32: That didn't cross my mind, just took a look at how that would turn out. We still would need to provide a NodeID when creating a new SubgraphRegistrar. 
And the default_value for the url i believe is almost always ok unless the indexer change it explicitly. So we can go with this now and update in the future?

node/src/manager/commands/deploy.rs
incrypto32: Updated

node/src/bin/manager.rs
lutter: Yes, let's not hold this up because of the URL; just go ahead and merge

PR #4933, opened 2023-10-24
graph, store: Do not create GIN indexes on array attributes

lutter:
Experience has shown that these indexes are very expensive to update and can have a very bad impact on the write performance of the database, but are hardly ever used or needed by queries.


leoyvens (approved):
So long GIN indexes! We should probably try dropping existing ones on hosted based on this insight.

Issue #4863, opened 2023-09-12
[Bug] Various reports that substream-powered-subgraphs silently stop processing blocks


chriswessels:
### Bug report

Reports from StakeSquid, Pinax, Data Nexus all confirm that substream-powered-subgraphs periodically stop processing blocks and silently fail. Reports that restarting graph-node fixes it.

### Some information to help us out

- [ ] Tick this box if this bug is caused by a regression found in the latest release.
- [ ] Tick this box if this bug is specific to the hosted service.
- [X] I have searched the issue tracker to make sure this issue is not a duplicate.

### OS information

None

colindickson:
In my experience as an indexer operator:

We used to have an issue where the substreams-based-subgraphs would stop syncing.  The graph-node would seemingly fail to reconnect if there was a disconnection and blocks would stop flowing indefinitely. We would have to restart the graph-node. 

This seems to fit the description of the bug reported here.

I have alerts set up to warn me whenever this happens, and they used to happen frequently - about 2 or 3 times a week.

**However**, ever since we upgraded our indexer graph-nodes to v0.32.0-rc.0 and now v0.32.0, we have not had this issue at all.

We are running our indexer with the StreamingFast endpoints for both firehose and substreams.

------------

The StreamingFast team is ready to assist in any debugging for this issue.   

If you are encountering this issue, can you please provide the following:
* graph-node version 
* substreams version 
* substreams endpoint 
* any logs whatsoever around the time of the incident indicating any disconnection

mindstyle85:
Hi, I am adding this here although our bug is a bit different. To first explain, we have 2 indexers, same exact setup on both in terms of config for graph-node, and both have version 32.

When we added substreams to the graph-node config on our primary indexer, it stopped syncing about 1/3 of our ETHEREUM only subgraphs, while the substream subgraph was syncing normally. A restart of the graph-node solved the stalling for a short while (less than 5min), but then those same subgraphs got stuck again. When we removed the susbtreams from the graph-node config and just left ethereum there, all the ethereum subgraphs started to sync again normally. When i switched back again, they stalled again. Note that the config is exactly the same as our other indexer, where it all works normally. Not one thing is different.

Unfortunately i cant let you know about versions etc since we just tested it and used a friends endpoint. I was also scouring logs but did not find nothing at all that would indicate any warning, error or anything that would explain why the ethereum subgraphs stopped syncing.

maoueh:
[...]

PR #4935, opened 2023-10-25
Log the progress of substreams

zorancv:
Better logging functionality in order to resolve problems like the ones in https://github.com/graphprotocol/graph-node/issues/4863


graph/src/blockchain/block_stream.rs
```diff
@@ -410,6 +418,18 @@ pub trait SubstreamsMapper<C: Blockchain>: Send + Sync {
                 )))
             }
 
+            Some(SubstreamsMessage::Progress(progress)) => {
+                if last_progress.elapsed() > Duration::from_secs(30) {
+                    info!(
+                        &logger,
+                        "Received progress update";
+                        "progress" => format!("{:?}", progress),
```
maoueh, line 426: The `progress` message in full is a bit verbose and hard to work with when printed using `Debug` trait.

I see different possibilities. Quickest one would be to print as JSON, we would need to tweak a bit the generated code so that `ModulesProgress` can be serialized to JSON. See commit https://github.com/graphprotocol/graph-node/commit/d66e348bd4b87fcdc636f36123496641fa9cacc2 for details.

My preferred approach would be that in `INFO`, we present a condensed "view" of the `ModulesProgress`, I would see:

```
Substreams backend `graph_out` last block is <last_seen_block>, Stages: [#S up to B (<first(reversed modules)>, +M), ...], Jobs: [#J on Stage S @ <start_block> | +<processed_blocks>|<end_block - start_block> elapsed <formatted duration>, ...]
```

The rendering would look like this:

```
Substreams backend `graph_out` last block is 18332001, Stages: [#5 (graph_out), #4 up to 18332000 (store_derived_factory_tvl), #3 up to 18332000 (store_max_windows, +3), #2 up to 16332000 (store_eth_prices), #1 up to 18332000 (store_ticks_liquidities, +10), #0 up to 18332000 (store_pool_count, +3)], Jobs: [#0 on Stage 4 @ 18332000 +517/1000 elapsed 3m 53s], [#1 on Stage 3 @ 18332000 +747/1000 elapsed 3m 53s], [#2 on Stage 2 @ 18332000 +877/1000 elapsed 3m 53s], [#3 on Stage 1 @ 18332000 +898/1000 elapsed 3m 53s], [#4 on Stage 0 @ 18332000 +894/1000 elapsed 3m 53s], [#5 on Stage 0 @ 18333000 +0/2 elapsed 3m 53s]
```

In `DEBUG` it could be a more expanded view (for example all modules instead of +N) and in `TRACE` could be the JSON view.

If you guys decide to make it, I can define more specifically how and where to get the info I have in my formatted view.

maoueh (changes requested):

graph/src/blockchain/block_stream.rs
zorancv: I don't see a reason why it wouldn't be done that way - each level with increased details. Maybe the info could be even more condensed...

graph/src/blockchain/block_stream.rs
```diff
@@ -356,9 +357,16 @@ pub trait SubstreamsMapper<C: Blockchain>: Send + Sync {
...
         message: Option<Message>,
+        last_progress: &mut Instant,
     ) -> Result<Option<BlockStreamEvent<C>>, SubstreamsError> {
         match message {
             Some(SubstreamsMessage::Session(session_init)) => {
+                info!(
+                    &logger,
+                    "Received session init";
+                    "trace_id" => session_init.trace_id.clone(),
+                    "session" => format!("{:?}", session_init),
```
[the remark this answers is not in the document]
zorancv: Yeah, trace_id was duplicated that way - removed it. Not sure of manual listing the fields. Why would you prefer it? Are some of them superfluous?

graph/src/blockchain/block_stream.rs
zorancv: @maoueh can you tell me where to get the info for the more concise log levels, please?

maoueh: Using format and assuming Rust struct progress message of type [ModulesProgress](https://github.com/graphprotocol/graph-node/pull/4935/files#diff-c67185a22602db5b79b5ccf8449efd8f79530a243ae472729665b54c753c3dcdR174):

> `Substreams backend `graph_out` last block is <last_seen_block>, Stages: [#S up to B (<first(reversed modules)>, +M), ...], Jobs: [#J on Stage S @ <start_block> | +<processed_blocks>|<end_block - start_block> elapsed <formatted duration>, ...]`

- `<last_seen_block>` Would be kept within `substreams_block_stream` struct directly, would be initialized to start block and then would be updated as the stream receives block (would update on `BlockScopedData` and `BlockUndoSignal` messages).
- `Stages` array would be printed in reverse order as found in `ModulesProgress.stages` field of the Rust struct.
- `#S` represents the **original** index in the `stages` struct (I say original so that index of reversed array is not used)
- `up to B` this should be displayed if the `Stage.completed_ranges.len()` is > 0 where the `B` value is (pseudo-code here) `Stage.completed_ranges.map(x.end_block).min()`.
- `(<first(reversed modules)>, +M)`, my definition here is a bit complicate, so feel free to change. I would render the `Stage.modules.reverse()` dynamically based on the module count. If module count is 1, render as `(<module_name>)` otherwise renders as `(first(reverse_modules), +{reverse_modules.len()-1})`. 
- `Jobs` would be the in order `ModulesProgress.running_jobs`
- `#J` would be the index of the Job within `running_jobs`
- `on Stage S` would refer to `Job.stage`.
- `<start_block>` would be `Job.start_block`
- `+<processed_blocks>` would be `Job.processed_blocks`
- `<end_block - start_block>` is actually the block count to be executed by the job, so `Job.end_block - Job.start_block`.
- `<formatted duration>` would be humanized value of `Job.duration_ms`.

Hopefully this is clear enough, happy to chat over Slack if further details is needed on what I had in mind.

We could think of a summary for `INFO`, what I propose above in `DEBUB` and JSON for `TRACE`. I leave you guys the decision on those.

graph/src/blockchain/block_stream.rs
```diff
@@ -403,13 +417,89 @@ pub trait SubstreamsMapper<C: Blockchain>: Send + Sync {
...
+                    };
+                    info!(&logger, "Substreams backend graph_out last block is {}, \
+                                    {} stages up to {}, {} jobs",
+                                    log_data.last_seen_block, len, up_to.unwrap_or(0), jlen);
+                    debug!(&logger, "Substreams backend graph_out last block is {},{}{}", 
+                        log_data.last_seen_block,
+                        stage_str,
+                        job_str,
+                    );
+                    trace!(
```
maoueh, line 493: I think we should log conditional to the logger actual level, if possible. What I mean is that if `GRAPH_LOG=trace` then you get the 3 log line where I think it's better to have a single line outputted based on the level.

Can we determine what level is active here and emit only the correct one? Pseudo-code:

```
match logger.level {
   trace: trace!(...),
   debug: debug!(...),
   _: info!(...) 
}
```

This have the added benefits also that in `info`, we would pay a small computation price. 

I would also personally move the formatting code implementation into `impl ModulesProgress { ... }` directly to avoid cluttering `block_stream` with such details.

maoueh (commented):
The rest seems like a good start.

graph/src/blockchain/block_stream.rs
zorancv: This doesn't seem to be possible with the logging mechanism we currently use (`slog`). There is no reliable way to request current maximum logging level - both `.is_enabled()` and the static fn `slog::__slog_static_max_level() `always return TRACE. I think it's not too bad to have the lower level printed out. Not sure computation would be substantial for debug case compared to the info, worst part are couple more memory allocations.

maoueh: Ok indeed there is nothing reliable on `slog` to achieve that, sad thing. To achieve this, we would need to have a global access to [pub fn logger_with_levels(show_debug: bool, levels: Option<&str>) -> Logger {](https://github.com/graphprotocol/graph-node/blob/866da3f0337d6b33c390a232426781ff76c59229/graph/src/log/mod.rs#L38-L58).

That would make it work as it's where we decide what to set the logger level to and we have no dynamic update. But could be seen as far fetched.

I'm changing my review to approve and let you guys decide what to do about this duplication. Don't forget to look at my comment about moving the formatting code somewhere else then in `block_stream.rs`.

maoueh (approved):

graph/src/blockchain/block_stream.rs
zorancv: I still am not sure how one sets logging to trace level in the `logger_with_levels()` function, not there is a command line switch for that...

Regarding the name/code pollution of the block_stream I completely agree - moved the string generation all to the `SubstreamsLogData` strucure which is now inside `substreams_block_stream.rs`. I feel it belongs there.

graph/src/blockchain/block_stream.rs
```diff
@@ -403,13 +412,28 @@ pub trait SubstreamsMapper<C: Blockchain>: Send + Sync {
...
                 Ok(Some(BlockStreamEvent::ProcessBlock(
                     block,
                     FirehoseCursor::from(cursor.clone()),
                 )))
             }
 
+            Some(SubstreamsMessage::Progress(progress)) => {
+                if log_data.last_progress.elapsed() > Duration::from_secs(30) {
+                    info!(&logger, "{}", log_data.info_string(&progress));
+                    debug!(&logger, "{}", log_data.debug_string(&progress));
```
mangas, line 426: This means you get all 3 logs if debug level is tracing, maybe it would be worth making sure you'd only get the appropriate information once?

graph/src/blockchain/block_stream.rs
zorancv: @mangas here we had a longer discussion on the subject and more context. I feel that printing all 3 levels is not so bad as info is just one line, debug is more or less detailed information, where the trace_id is very different as a structure (a JSON) compared to the debug and it might be useful to have both at this level. WDYT?

graph/src/blockchain/block_stream.rs
```diff
@@ -356,9 +358,15 @@ pub trait SubstreamsMapper<C: Blockchain>: Send + Sync {
...
+        log_data: &mut SubstreamsLogData,
     ) -> Result<Option<BlockStreamEvent<C>>, SubstreamsError> {
         match message {
             Some(SubstreamsMessage::Session(session_init)) => {
+                info!(
+                    &logger,
+                    "Received session init";
+                    "session" => format!("{:?}", session_init),
+                );
                 *logger = logger.new(o!("trace_id" => session_init.trace_id));
```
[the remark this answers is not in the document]
zorancv: I looked into it somewhat. So far my conclusion is that there are multiple calls or some recursion. Solving it on the level of macros would be all local but was somewhat challenging and I gave up, but can give it another try.

graph/src/blockchain/block_stream.rs
zorancv: So looking more into `logger_with_levels()` I'd either have to implement a lite version of [parse_logging_spec](https://github.com/slog-rs/envlogger/blob/b8c7081592b7088b296f46283d896c8f2f4b251c/src/lib.rs#L259) as it is private and combine the result with the `show_debug` flag. Or `std::mem::transmute` the result of `LogBuilder`s `build()` call. Neither seems very elegant to me, so I feel it's not worth it.

zorancv: Closing this as discussed on slack.

PR #4937, opened 2023-10-25
Track per query connection and permit wait times

lutter:
There are other places during query execution where we might be waiting for connections, but this at least records the times for getting the actual data.


leoyvens (approved):

leoyvens:
Nice!

PR #4940, opened 2023-10-26
Metrics for DIPS experiments

incrypto32:


chain/ethereum/src/runtime/runtime_adapter.rs
```diff
@@ -103,7 +103,8 @@ fn ethereum_call(
     abis: &[Arc<MappingABI>],
     eth_call_gas: Option<u32>,
 ) -> Result<AscEnumArray<EthereumValueKind>, HostExportError> {
-    ctx.gas.consume_host_fn(ETHEREUM_CALL)?;
+    ctx.gas
+        .consume_host_fn_with_metrics(ETHEREUM_CALL, "ethereum_call")?;
```
incrypto32, line 107: There is a macro crate `function_name!` https://docs.rs/function_name/latest/function_name/

Which can be potentially used here instead of hardcoding then name. Thoughts?

leoyvens: The nicest would be an enum for the gas ops, but you don't need to spend time refactoring that right now.

leoyvens (approved):
LGTM! Since this is off by default we don't have to think too hard about any performance impact.

PR #4950, opened 2023-10-27
Rewrite integration tests

lutter:
With these changes, we do not use truffler/ganache anymore. Tests are now written as normal Rust tests. See `tests/README.md` for details


lutter:
This is now ready for review - it's not clear to me why the runner tests fail, but I believe it's unrelated to the changes in integration tests.

.github/workflows/ci.yml
```diff
@@ -130,39 +149,32 @@ jobs:
...
+      - name: Run integration tests
         id: integration-tests-1
         uses: actions-rs/cargo@v1
         env:
-          N_CONCURRENT_TESTS: "4"
-          TESTS_GANACHE_HARD_WAIT_SECONDS: "30"
-          # Reduce how long a batch will stick around to avoid timeouts
-          GRAPH_STORE_WRITE_BATCH_DURATION: 30
+          # Basically, unlimted concurrency
+          N_CONCURRENT_TESTS: "1000"
```
mangas, line 171: Do we still need this env var then?

tests/docker-compose.yml
```diff
@@ -0,0 +1,32 @@
+version: '3'
```
mangas, line 1: could we just use the docker-compose.yaml from the docker folder and add anvil there? Would also be happy with replacing the other one with this.

mangas (approved):
This looks great !

.github/workflows/ci.yml
lutter: I wasn't sure if we need to tweak this for GH actions, and wanted to call that out as a tunable, but if we don't need to adjust it, we can take it out.

tests/docker-compose.yml
lutter: The `docker-compose.yml` from uses very specific and somewhat idiosyncratic port numbers that I chose to that they would not conflict with other things a dev might have running. I think it's easier to keep them as separate files.

lutter:
> This looks great !

Thanks for the review - I agree there are ways to improve this further, but I think it's a lot better than what we had, so merging this. We still need to figure out what's ailing the runner tests.

PR #4954, opened 2023-10-30
tests: Update runner-tests package.json

lutter:
This gets rid of some of the JS errors from the runner-tests, but it looks like some of the tests are failing now.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -84,6 +84,14 @@ jobs:
         run: sudo ethtool -K eth0 tx off rx off
       - name: Checkout sources
         uses: actions/checkout@v2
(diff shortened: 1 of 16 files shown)


lutter:
I celebrated too early - even though those yarn install errors don't happen locally, they still happen in CI. Not sure how to proceed.

lutter:
@saihaj Could you have a look at the complaints from yarn in the 'Subgraph Runner integration tests' ? I thought I had updated all the `package.json` files correctly and that part works locally, but somehow they fail in CI. Not sure how I would have to massage `package.json` to get past that.

lutter:
This has been addressed by #4957

PR #4957, opened 2023-10-31
fix(tests): yarn workspace to install all subfolders

saihaj:
building on top of #4954

1. Make sure we are installing everything for the `yarn workspace` not all the test folders were listed part of workspace and it is much easier to have it install all dependencies for the test folders
2. I saw that for `runner-tests` we are not setting Node.js version, so set that up
3. Upgraded to use Node 20 so we are on LTS version


lutter (approved):
Thanks for doing that! Things are definitely much better with this.

PR #4958, opened 2023-10-31
runer-tests: Fix up the graft base for grafted.yaml

lutter:


leoyvens:
Fixed in https://github.com/graphprotocol/graph-node/pull/4962

PR #4961, opened 2023-11-01
test: Update truffle to latest version

lutter:
The previous version didn't play well with node v20

The test `poi-for-failed-subgraph` still fails locally, but the other tests succeed at least locally

### DIFF
--- a/tests/integration-tests/api-version-v0-0-4/package.json
+++ b/tests/integration-tests/api-version-v0-0-4/package.json
@@ -15,12 +15,12 @@
     "solc": "^0.8.2"
   },
   "dependencies": {
-    "@truffle/contract": "^4.3",
-    "@truffle/hdwallet-provider": "^1.2",
+    "@truffle/contract": "^4.6.31",
+    "@truffle/hdwallet-provider": "^2.1.15",
     "apollo-fetch": "^0.7.0",
     "babel-polyfill": "^6.26.0",
     "babel-register": "^6.26.0",
     "gluegun": "^4.6.1",
-    "truffle": "^5.2"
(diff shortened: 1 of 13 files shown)


saihaj (approved):

lutter:
@saihaj I have another JS headscratcher with this: the tests now fail with the error
```
Compiling your contracts...
===========================
> Compiling ./../../common/Migrations.sol
> Compiling ./../../common/SimpleContract.sol

No input sources specified.

Compilation failed. See above.
Truffle v5.11.5 (core: 5.11.5)
Node v20.8.1
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

I can't really figure out where that error is coming from; the only thing in all the node_modules that has the string `No input sources specified.` is `tests/integration_tests/node_modules/solc/wrapper.js` but it doesn't look like that is being run - I tried putting `console.log` statements in there as well as adding syntax errors and there's no sign that anything is amiss.

Do you have any idea how to further investigate this? According to the truffle docs, our setup should be fine and running `yarn truffle compile` in, for example, `tests/integration-tests/int8` should work.

leoyvens:
I tried debugging this and also had no luck. Maybe we need to go straight to HardHat or another alternative?

azf20:
Might be time for Foundry? https://github.com/foundry-rs/foundry

lutter:
> Might be time for Foundry? https://github.com/foundry-rs/foundry

Yeah, that's what I have been working with in my rewrite-the-world attempt

lutter:
Not needed anymore - #4950 is a better solution to this problem

Issue #4943, opened 2023-10-26
[Feature] Add more labels to `deployment_sync_secs`


lutter:
### Description

The `deployment_sync_secs` metric measures indexing speed. We should add the following labels to it:

* `sgd`: the internal identifier of the deployment to disambiguate copies of the same IPFS hash
* `shard`: the shard in which the deployment lives so that we can aggregate by shard and hopefully use that to better understand when indexing in a specific shard becomes slow

### Some information to help us out

- [ ] Tick this box if you plan on implementing this feature yourself.
- [X] I have searched the issue tracker to make sure this issue is not a duplicate.

PR #4965, opened 2023-11-02
Add more labels to `deployment_sync_secs`

zorancv:
Addressing https://github.com/graphprotocol/graph-node/issues/4943

### DIFF
--- a/graph/src/components/metrics/stopwatch.rs
+++ b/graph/src/components/metrics/stopwatch.rs
@@ -46,6 +46,7 @@ impl StopwatchMetrics {
         subgraph_id: DeploymentHash,
         stage: &str,
         registry: Arc<MetricsRegistry>,
+        shard: String
     ) -> Self {
         let stage = stage.to_owned();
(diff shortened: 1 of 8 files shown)


graph/src/components/metrics/stopwatch.rs
```diff
@@ -128,7 +132,7 @@ impl StopwatchInner {
             // Register the current timer.
             let elapsed = self.timer.elapsed().as_secs_f64();
             self.counter
-                .get_metric_with_label_values(&[section, &self.stage])
+                .get_metric_with_label_values(&[section, &self.stage, &&self.shard])
```
lutter, line 135: The double `&&` shouldn't be needed, just one `&` should be enough.

lutter (changes requested):
This looks good. There are some compile/test failures that need to be addressed.

zorancv:
@lutter I believe that the issues are addressed.

lutter (approved):

PR #4968, opened 2023-11-02
fix(graphql): include introspection fields in error

### LINKED ISSUE #2443: Introspection queries stop working after a subgraph has encountered an indexing error
bug

Introspection queries fail after a subgraph has encountered an indexing error.

This potentially also happens if the subgraph flat fails (not sure) but we encountered it whilst using the nonFatalErrors feature. Hence:

1. Deploy a failing subgraph using the nonFatalErrors feature
2. Wait for it to fail and then try to run a schema introspection or type query
3. The query returns `message: indexing_error`

Even if a subgraph has failed, introspection queries should always work (both when using the nonFatalErrors feature or not). The schema is immutable and not affected by any indexing errors.

The `__schema` and `__type` fields are part of the spec and used by various tooling. Hence, adding a `subgraphErrors: allow` input field to those fields wouldn't work well. Instead, both `__type` and `__schema` should simply always work.

saihaj:
include `__schema` and `__type` part of the result when there is an indexing error and the error policy is deny.

Fixes #2443

----

There is another issue I found (could just be how tests are setup) but I wasn't really able to fix. In the example below if there is an indexing error this will return `null` but I would imagine that we return partial result i.e. `__schema` is returned in the `data` 
```graphql
query { musician(id: "m1") { id } __schema { queryType { name } } }
```

### DIFF
--- a/graphql/src/store/resolver.rs
+++ b/graphql/src/store/resolver.rs
@@ -385,33 +385,31 @@ impl Resolver for StoreResolver {
                 let mut data = result.take_data();
 
                 // Only keep the _meta, __schema and __type fields from the data
-                let meta_fields = data
-                    .as_mut()
-                    .and_then(|d| {
-                        let meta_field = d.remove("_meta");
-                        let schema_field = d.remove("__schema");
-                        let type_field = d.remove("__type");
+                let meta_fields = data.as_mut().and_then(|d| {
+                    let meta_field = d.remove("_meta");
+                    let schema_field = d.remove("__schema");
+                    let type_field = d.remove("__type");
 
-                        // combine the fields into a vector
-                        let mut meta_fields = Vec::new();
+                    // combine the fields into a vector
+                    let mut meta_fields = Vec::new();
 
-                        if let Some(meta_field) = meta_field {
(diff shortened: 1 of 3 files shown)


graphql/src/store/resolver.rs
```diff
@@ -382,12 +382,36 @@ impl Resolver for StoreResolver {
...
+                let meta_fields = data.as_mut().and_then(|d| {
+                    let meta_field = d.remove("_meta");
+                    let schema_field = d.remove("__schema");
+                    let type_field = d.remove("__type");
+
+                    // combine the fields into a vector
+                    let mut meta_fields = Vec::new();
+
+                    if let Some(meta_field) = meta_field {
+                        meta_fields.push((Word::from("_meta"), meta_field));
```
mangas, line 397: can we move these magic values to constants?

mangas:
LGTM but I'd rather @lutter also reviews

lutter (approved):

lutter:
@saihaj Can you rebase once more? CI is now actually stable and we should then be able to merge this finally

PR #4977, opened 2023-11-08
Reset substreams/firehose block ingestor backoff

### LINKED ISSUE #4976: [Bug] Substreams Block Ingestor sleep times
### Bug report

Noticed that [Substreams](https://github.com/graphprotocol/graph-node/blob/139f3144d8d878b528416fcd4f9a34353dc3dfb1/chain/substreams/src/block_ingestor.rs#L134-L135) Block Ingestor `backoff` instance is never reset in the loop. So the sleep time drops to 30 seconds eventually on an instance with connectivity issues, and never recovers.

Reproduce:
1. Start indexing a substreams subgraph locally
2. Shutdown Postgres docker container and wait until time between retries drops to 30 seconds
3. Start Postgres container back up
4. Shut it down again. It will continue to sleep for 30 seconds between retries even though it would make sense to recover back to 250ms wait times

Should be an easy fix - just reset it every time new cursor is different from `latest_cursor`

### Relevant log output

_No response_

### IPFS hash

_No response_

_No response_

### OS information

None

YaroShkvorets:
Fixes #4976


maoueh (approved):
Seems good

YaroShkvorets:
Will re-implement on top of the current master branch with integration tests and resubmit

PR #4978, opened 2023-11-10
Optimise `cached_head_ptr` calls within process block

### LINKED ISSUE #4972: Break down `process_block` section in `deployment_sync_secs` metrics to smaller sections
### Description

_No response_

_No response_

incrypto32:
Closes #4972

### DIFF
--- a/core/src/subgraph/runner.rs
+++ b/core/src/subgraph/runner.rs
@@ -71,6 +71,7 @@ where
                     env_vars.subgraph_error_retry_jitter,
                 ),
                 entity_lfu_cache: LfuCache::new(),
+                cached_head_ptr: None,
             },
             logger,
             metrics,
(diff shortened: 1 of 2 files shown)


incrypto32:
Let me know if the sections for process_block could be better, i was unclear how to divide it into more sections

core/src/subgraph/runner.rs
```diff
@@ -899,12 +900,14 @@ where
                 .observe(block.trigger_count() as f64);
         }
 
+        let ptr_updates_section = self.metrics.stream.stopwatch.start_section("ptr_updates");
+
         if block.trigger_count() == 0
```
leoyvens, line 905: An idea, lets wrap this whole if block in an outer if with the condition:
`if self.state.skip_ptr_updates_timer.elapsed() && !self.state.synced`
This way, if the timer hasn't elapsed or the subgraph is synced, we won't even bother with resetting the timer and updating the cached head.

incrypto32: Just pushed a commit, hope the change made is what you suggested

incrypto32: hmm, intereseting i thought when the variable that is assigned to the section goes out of scope the timer is reset.

zorancv: I think that the check for `!self.state.synced` should be separated and outer one, and then the else branch of the check for `self.state.skip_ptr_updates_timer.elapsed()` should reset the timer too.

zorancv (approved):
Other than the timer handling I've commented it looks good to me.

PR #4983, opened 2023-11-13
runner tests: Do not clobber api-version tests

lutter:
Two different tests are run in the same directory; make sure that the files they generate and use do not clobber each other.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -100,11 +100,9 @@ jobs:
       - name: Run runner tests
         id: runner-tests-1
         uses: actions-rs/cargo@v1
-        env:
-          TESTS_GANACHE_HARD_WAIT_SECONDS: "60"
         with:
           command: test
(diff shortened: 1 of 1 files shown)


tests/runner-tests/api-version/package.json
```diff
@@ -2,12 +2,12 @@
   "name": "api-version",
   "version": "0.1.0",
   "scripts": {
-    "codegen": "graph codegen --skip-migrations",
+    "codegen": "graph --version # dummy, we need a  'codegen'",
```
azf20, line 5: Not sure what is happening here?

azf20 (approved):

tests/runner-tests/api-version/package.json
lutter: The Rust framework in `tests/tests/runner_tests.rs` that is running these tests always runs `yarn codegen` so we need something there that succeeds; doing that was easier than modifying that framework

PR #4984, opened 2023-11-13
fix: Increase RPC base backoff

leoyvens:
Graph Node tends to spam failing network requests of different kinds, often giving the impression that we have no backoff at all. Starting at a slightly higher base backoff might help.

Also add a newly discovered deterministic Geth error.


lutter (approved):

sduchesneau:
LGTM ! thanks @leoyvens

PR #4985, opened 2023-11-13
bump rust version

mangas:


runtime/test/src/test/abi.rs
```diff
@@ -1,42 +1,46 @@
...
+
+// #[tokio::test(flavor = "multi_thread")]
+// async fn unbounded_loop_v0_0_4() {
+//     test_unbounded_loop(API_VERSION_0_0_4).await;
+// }
+
+// #[tokio::test(flavor = "multi_thread")]
+// async fn unbounded_loop_v0_0_5() {
+//     test_unbounded_loop(API_VERSION_0_0_5).await;
+// }
```
lutter, line 40: IIRC, this will be addressed in a separate PR, right? Could you leave a comment why that's commented out and roughly what needs to be done here?

chain/near/src/trigger.rs
```diff
@@ -156,365 +159,389 @@ pub struct ReceiptWithOutcome {
...
-        fn asc_type_id(
-            &mut self,
-            type_id_index: graph::runtime::IndexForAscTypeId,
-        ) -> Result<u32, HostExportError> {
-            // Not totally clear what is the purpose of this method, why not a default implementation here?
-            Ok(type_id_index as u32)
-        }
-    }
-}
+// #[cfg(test)]
```
lutter, line 162: There should be a comment why these tests are commented out and what needs to happen to run them again

core/src/subgraph/runner.rs
```diff
@@ -674,20 +675,32 @@ where
...
+                .manifest_idx()
+                .ok_or_else(|| anyhow!("Expected template to have an idx"))?;
+            let x = self
+                .inputs
+                .templates
+                .iter()
+                .find(|t| t.manifest_idx() == manifest_idx)
+                .ok_or_else(|| {
+                    anyhow!("Expected to find a template for this dynamic data source")
+                })?;
```
lutter, line 695: Could we have a more descriptive name for `x` ?

graph/src/runtime/wasm/module/context.rs
```diff
@@ -0,0 +1,1334 @@
...
+            .store_get(
+                &mut self.as_mut().ctx.state,
+                entity_type.clone(),
+                id.clone(),
+                gas,
+                scope,
+            )?
+            // This is not great in a hot path but otherwise the self ref would not
+            // be released for the next block. Would be good to find a better pattern here.
+            .map(|e| e.into_owned());
```
lutter, line 163: This change builds for me without requiring a clone:

```suggestion
        let block_state = &mut self.as_mut().ctx.state;
        let entity_option =
            host_exports.store_get(block_state, entity_type.clone(), id.clone(), gas, scope)?;
```

lutter (changes requested):
Kudos! This is a huge amount of work. The main thing I'd like to see addressed are a couple comments and the clone in `store_get_scoped`. Other than that, I think it's ready to merge, but I am no wasm expert, and it would be great for @leoyvens to look at this, too.

chain/near/src/trigger.rs
mangas: Good catch, I've added all of these to https://github.com/graphprotocol/graph-node/issues/5079 so I can address them with mroe time. This api changed and the utility functions to create the wasm instance were not importable from test-store and wasm-runtime-test modules but I didn't want to spend more time to fix it here.

core/src/subgraph/runner.rs
mangas: good catch 🤦

leoyvens:
Is there any protection here against a mapping maxing out a CPU by running an infinite loop?

mangas:
> Is there any protection here against a mapping maxing out a CPU by running an infinite loop?

That will be added on a following PR, that's why the comments were disabled

graph/src/runtime/wasm/module/context.rs
mangas: it compiles but breaks the tests because of the runtime borrow checking so I had to put it back

leoyvens:
We might not want to rollout, and surely not release, without some type of protection against that in place. It's preferable that master is always in a state which we'd be comfortable releasing.

mangas:
> We might not want to rollout, and surely not release, without some type of protection against that in place. It's preferable that master is always in a state which we'd be comfortable releasing.

Agreed, yet this is a specific care for which I proposed we make an exception since the fix should be shortly along but should not be added in this PR, first because of the need to maintain the PR open in the meantime and second we are not going to release before this is will be fixed so that's not an issue as discussed and agreed by the team

leoyvens:
You can put the fix in a separate PR that's based on this branch, the fix doesn't need to be blocked on this being merged.

graph/src/runtime/wasm/module/context.rs
lutter: I found a way around the clone, but it's ugly as hell. The clone basically undoes PR #4624. Do we really need interior mutability for `WasmInstanceContext`? Can't we get away with just plain `&mut` as we used to?

```diff
diff --git a/graph/src/runtime/wasm/module/context.rs b/graph/src/runtime/wasm/module/context.rs
index cc3733214..6502209c3 100644
--- a/graph/src/runtime/wasm/module/context.rs
+++ b/graph/src/runtime/wasm/module/context.rs
@@ -140,6 +140,11 @@ impl WasmInstanceContext {
         scope: GetScope,
     ) -> Result<AscPtr<AscEntity>, HostExportError> {
         let host_exports = self.as_ref().ctx.host_exports.cheap_clone();
+        let instrument = self.as_ref().ctx.instrument;
+        let host_metrics = self.as_ref().host_metrics.cheap_clone();
+        let debug_fork = self.as_ref().ctx.debug_fork.cheap_clone();
+        let logger = self.as_ref().ctx.logger.cheap_clone();
+
         let _timer = self
             .as_ref()
             .host_metrics
@@ -150,37 +155,29 @@ impl WasmInstanceContext {
         let entity_type: String =
             asc_get(&read_store, self.as_mut().asc_heap_mut(), entity_ptr, gas)?;
         let id: String = asc_get(&read_store, self.as_mut().asc_heap_mut(), id_ptr, gas)?;
+        let (mut heap_mut, mut state_mut) = RefMut::map_split(
+            self.as_mut(),
+            |WasmInstanceContextInner { ctx, asc_heap, .. }| {
+                (asc_heap.as_mut().unwrap(), &mut ctx.state)
+            },
+        );
         let entity_option = host_exports
-            .store_get(
-                &mut self.as_mut().ctx.state,
-                entity_type.clone(),
-                id.clone(),
-                gas,
-                scope,
-            )?
+            .store_get(&mut state_mut, entity_type.clone(), id.clone(), gas, scope)?
             // This is not great in a hot path but otherwise the self ref would not
             // be released for the next block. Would be good to find a better pattern here.
             .map(|e| e.into_owned());
 
-        if self.as_ref().ctx.instrument {
-            debug!(self.as_ref().ctx.logger, "store_get";
+        if instrument {
+            debug!(logger, "store_get";
                     "type" => &entity_type,
                     "id" => &id,
                     "found" => entity_option.is_some());
         }
-        let host_metrics = self.as_ref().host_metrics.cheap_clone();
-        let debug_fork = self.as_ref().ctx.debug_fork.cheap_clone();
 
         let ret = match entity_option {
             Some(entity) => {
                 let _section = host_metrics.stopwatch.start_section("store_get_asc_new");
-                // let entity = entity.to_owned().sorted_ref();
-                asc_new(
-                    store,
-                    self.as_mut().asc_heap_mut(),
-                    &entity.sorted_ref(),
-                    gas,
-                )?
+                asc_new(store, &mut *heap_mut, &entity.sorted_ref(), gas)?
             }
             None => match &debug_fork {
                 Some(fork) => {
@@ -194,12 +191,7 @@ impl WasmInstanceContext {
                         Some(entity) => {
                             let _section =
                                 host_metrics.stopwatch.start_section("store_get_asc_new");
-                            let entity = asc_new(
-                                store,
-                                self.as_mut().asc_heap_mut(),
-                                &entity.sorted(),
-                                gas,
-                            )?;
+                            let entity = asc_new(store, &mut *heap_mut, &entity.sorted(), gas)?;
                             self.store_set(store, gas, entity_ptr, id_ptr, entity)?;
                             entity
                         }
```

leoyvens:
Taking a look at the use of interior mutability in `WasmInstanceContext`, but after hacking a bit on this I do think we can avoid it.

The fundamental issue seems to be that when trying to call `AscHeap`, we end up needing a mutable borrow to _both_ a `WasmInstanceContext` and a `StoreContextMut`. But one is inside the other with `StoreContextMut<WasmInstanceContext>`, so this results in a double mutable borrow.

One way around a double mutable borrow would be the interior mutability, but a more idiomatic solution might be:
`impl AscHeap for StoreContextMut<'_, WasmInstanceContext> { ... }`

I experimented with that [here](https://github.com/graphprotocol/graph-node/commit/b67d5946eb375b47ad112a7921bfca545e214b19) and it seemed possible. It also seems to avoid changing the definition of the `AscHeap` trait, and as a consequence hopefully means we can avoid merging the `runtime` crate with `graph`.

leoyvens:
To be able to keep the crate separation, we might in `graph` the impl:
`impl<H: AscHeap> AscHeap for StoreContextMut<'_, H> { ... }`

And then in `runtime`:
`impl AscHeap for WasmInstanceContext { ... }`

leoyvens:
The above might not quite work, unless we also either:
1. Make one change to `trait AscHeap`, which is to make `fn raw_new` take `&self` instead of `&mut self`. This will require using some atomics in `struct AscHeap` but should work.
2. 'Flatten' the fields of `struct AscHeap` into `WasmInstanceContext`, removing `struct AscHeap` as a separate struct.

So there are some details we need to hash out, but I'm happy to pair on this

graph/src/runtime/wasm/module/context.rs
leoyvens: Even removing the RefCell this borrow conflict persisted (but now at compile time). I fixed it by switching from `Cow` to `Arc`.

leoyvens:
This is ready for another review, all comments addressed. I'll run it on an integration cluster.

leoyvens:
This has been running in the integration cluster for a few days, no discrepancies.

lutter (approved):

PR #4990, opened 2023-11-16
More `InputSchema` refactorings

lutter:
This is all groundwork for aggregations; in an effort to not produce a gigantic PR, I am putting up somewhat smaller ones. The biggest change this PR introduces is our own structs for object and interface type that are easier/saner to work with than the raw GraphQL AST

### DIFF
--- a/graph/src/data/store/id.rs
+++ b/graph/src/data/store/id.rs
@@ -19,7 +19,7 @@ use crate::{
     schema::EntityType,
 };
 
-use super::{scalar, Value, ID};
+use super::{scalar, Value, ValueType, ID};
 
 /// The types that can be used for the `id` of an entity
 #[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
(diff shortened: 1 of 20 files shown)


graph/src/data/store/id.rs
```diff
@@ -19,13 +19,14 @@ use crate::{
...
 
-use super::{scalar, Value, ID};
+use super::{scalar, Value, ValueType, ID};
 
 /// The types that can be used for the `id` of an entity
 #[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
 pub enum IdType {
     String,
     Bytes,
+    Int8,
```
mangas, line 29: Isn't int8 short for an ID?

mangas (approved):

lutter:
Thinking about `Int8` as `id` more, I will merge this PR only up to the introduction of `Int8` and rework the rest without `Int8`; we'll leave the `Int8` issue for another PR.

lutter:
Rebased to latest master and removed everything after the `Int8` addition

PR #4993, opened 2023-11-16
tests: Make how long runner tests wait for sync configurable

lutter:
Set it to 10 minutes for github actions. Default is 1 minute


zorancv (approved):
Looks good to me.

PR #4995, opened 2023-11-16
Refactor index creation and fix unncessary prefix

### LINKED ISSUE #4982: [Bug] `graphman index create` indexes prefixes when it shouldn't
### Bug report

The command `graphman index create` indexes prefixes of columns in some cases (e.g., when indexing `id` it creates and index on `left(id,..)`) when the whole column should be indexed. The problem is that `deployment_store::resolve_index_exprs` indexes prefixes for all `String` and `Bytes` columns when it should use the entire column if the column is a reference. 

The logic in `resolve_index_exprs` needs to match the logic [here](https://github.com/graphprotocol/graph-node/blob/master/store/postgres/src/relational/ddl.rs#L228-L270) It would probably be cleanest to factor this logic into a helper and use it in both places, not sure how complicated that will be. 

### Relevant log output

_No response_

### IPFS hash

_No response_

_No response_

- [X] Tick this box if this bug is caused by a regression found in the latest release.

### OS information

None

incrypto32:
Closes #4982

### DIFF
--- a/store/postgres/src/relational/ddl.rs
+++ b/store/postgres/src/relational/ddl.rs
@@ -198,6 +198,56 @@ impl Table {
         }
     }
 
+    /// Calculates the indexing method and expression for a database column.
+    ///
+    /// ### Parameters
+    /// * `immutable`: A boolean flag indicating whether the table is immutable.
+    /// * `column`: A reference to the `Column` struct, representing the database column for which the index method and expression are being calculated.
+    ///
+    /// ### Returns
+    /// A tuple `(String, String)` where:
+    /// - The first element is the indexing method ("btree", "gist", or "gin"),
+    /// - The second element is the index expression as a string.
+    pub fn calculate_index_method_and_expression(
+        immutable: bool,
+        column: &Column,
+    ) -> (String, String) {
+        if column.is_reference() && !column.is_list() {
+            if immutable {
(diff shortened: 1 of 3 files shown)


store/postgres/src/relational/ddl.rs
```diff
@@ -198,6 +198,56 @@ impl Table {
...
+    /// * `column`: A reference to the `Column` struct, representing the database column for which the index method and expression are being calculated.
+    ///
+    /// ### Returns
+    /// A tuple `(String, String)` where:
+    /// - The first element is the indexing method ("btree", "gist", or "gin"),
+    /// - The second element is the index expression as a string.
+    pub fn calculate_index_method_and_expression(
+        immutable: bool,
+        column: &Column,
+    ) -> (String, String) {
```
lutter, line 214: There's an enum for index methods [here](https://github.com/graphprotocol/graph-node/blob/master/store/postgres/src/relational/index.rs#L17) - it would be better to use that. I wonder if that should return the whole `CreateIndex` from that file instead of the second string, but that might become a bit too unwieldy. Maybe some or all of this logic should go into a new constructor for `CreateIndex`

lutter: Actually, I think what I said above would be good for another refactor, but what you have here is ok for now.

lutter (approved):

PR #5001, opened 2023-11-20
Fix: substream datasources missing in TriggerFilter

incrypto32:


tests/runner-tests/end-block/substreams-test-v1.0.1.spkg
mangas, line 1: Why was this changed?

mangas (approved):

PR #5010, opened 2023-11-21
perf(store): Avoid the BRIN index for queries that filter by id

leoyvens:
A query has a where clause which filters by id is best served by the GiST index on `(id, block_range)`. However when the table is set to account-like, the BRIN index is sometimes chosen by the query planner even though it can have much worse performance.

We've encountered this repeatedly on some large subgraphs and even had to turn off `account-like` on some tables to resolve the situation, which is of course not ideal because some queries do benefit from it. This happened even with BRIN indexes using using `*_multi_ops`.

This proposes addressing this by not adding the BRIN-friendly clauses on queries that we know are filtering by id. Which include type C and D queries, and top-level queries explicitly filtering by `id`.

Impact to be tested in real-world scenarios.

### DIFF
--- a/store/postgres/src/block_range.rs
+++ b/store/postgres/src/block_range.rs
@@ -167,8 +167,11 @@ impl<'a> BlockRangeColumn<'a> {
 }
 
 impl<'a> BlockRangeColumn<'a> {
-    /// Output SQL that matches only rows whose block range contains `block`
-    pub fn contains(&self, out: &mut AstPass<Pg>) -> QueryResult<()> {
+    /// Output SQL that matches only rows whose block range contains `block`.
+    ///
+    /// `filters_by_id` has no impact on correctness. It is a heuristic to determine
+    /// whether the brin index should be used. If `true`, the brin index is not used.
+    pub fn contains(&self, out: &mut AstPass<Pg>, filters_by_id: bool) -> QueryResult<()> {
         out.unsafe_to_cache_prepared();
 
         match self {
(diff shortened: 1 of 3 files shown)


store/postgres/src/block_range.rs
```diff
@@ -167,20 +167,26 @@ impl<'a> BlockRangeColumn<'a> {
 }
 
 impl<'a> BlockRangeColumn<'a> {
-    /// Output SQL that matches only rows whose block range contains `block`
-    pub fn contains(&self, out: &mut AstPass<Pg>) -> QueryResult<()> {
+    /// Output SQL that matches only rows whose block range contains `block`.
+    ///
+    /// `filters_by_id` has no impact on correctness. It is a heuristic to determine
+    /// whether the brin index should be used. If `true`, the brin index is not used.
+    pub fn contains(&self, out: &mut AstPass<Pg>, filters_by_id: bool) -> QueryResult<()> {
```
lutter, line 174: Maybe `skip_brin_index` would be a better name than `filters_by_id` in this function, but it's not a big deal.

lutter (approved):
LGTM! I would change this so that it's possible to restore the old behavior by setting an environment variable in case this turns out to cause trouble under load as changing an env var is a much simpler operation than rolling back. Once this has been in production for a while and works well, we can get rid of the env var.

leoyvens:
Thanks for the review, I've added the env var

lutter:
I still approve ;)

PR #5015, opened 2023-11-23
Optimize log filtering in blockstream

incrypto32:


chain/ethereum/src/trigger.rs
```diff
@@ -291,8 +358,12 @@ impl EthereumTrigger {
...
             }
             EthereumTrigger::Call(call) => Some(&call.to),
-            EthereumTrigger::Log(log, _) => Some(&log.address),
-
+            EthereumTrigger::Log(log_ref) => match log_ref {
+                LogRef::FullLog(log, _) => Some(&log.address),
+                LogRef::LogPosition(index, receipt) => {
+                    Some(&receipt.logs.get(*index).unwrap().address)
+                }
+            },
```
leoyvens, line 366: We should not be matching on `LogRef`, it needs an `address()` method. Same goes for `block_hash` and `block_number` above.

chain/ethereum/src/trigger.rs
```diff
@@ -229,11 +229,75 @@ impl ToAscPtr for MappingTrigger {
...
+        match self {
+            LogRef::FullLog(log, _) => log.clone(),
+            LogRef::LogPosition(index, receipt) => {
+                Arc::new(receipt.logs.get(*index).unwrap().clone())
+            }
+        }
+    }
+}
+
+impl PartialEq for LogRef {
```
leoyvens, line 282: Where is this impl being relied on exactly?

chain/ethereum/src/trigger.rs
```diff
@@ -229,11 +229,75 @@ impl ToAscPtr for MappingTrigger {
...
+            LogRef::LogPosition(_, receipt) => Some(receipt.transaction_index),
+        }
+    }
+
+    fn transaction_hash(&self) -> Option<H256> {
+        match self {
+            LogRef::FullLog(log, _) => log.transaction_hash,
+            LogRef::LogPosition(_, receipt) => Some(receipt.transaction_hash),
+        }
+    }
```
leoyvens, line 263: To be extra paranoid that the behaviour is equivalent between variants, I'd suggest we have `log_index`, `transaction_index`, `trasaction_hash` and all similar methods call `self.log()` first and then access the field.

leoyvens (changes requested):
Nice! We should put this through an integration cluster run.

chain/ethereum/src/trigger.rs
incrypto32: https://github.com/graphprotocol/graph-node/blob/b1fa9cecd4f6951788bb32dc3e151db0eb66acba/chain/ethereum/src/trigger.rs#L303

Its relied in `impl PartialEq for EthereumTrigger`

leoyvens: I find it weird that this compares the `receipt` field given that it is optional, seems wrong really. Lets not have this impl, and inline the logic into `EthereumTrigger` as it was before. That code is critical for correct ordering of the triggers so I find that less indirection makes it easier to read. Also lets stop comparing the `receipt`, because it's optional, and instead compare only on `transaction_hash` and `log_index`.

incrypto32: @leoyvens clarifying i understand this right, should we remove the previously existing comparison of the optional receipt right? Or should we just inline the logic and keep the original comparison of the optional receipt?

incrypto32: done! 

I'll run it through the integration cluster

chain/ethereum/src/data_source.rs
```diff
@@ -399,7 +399,11 @@ impl DataSource {
 
     fn handlers_for_log(&self, log: &Log) -> Result<Vec<MappingEventHandler>, Error> {
```
leoyvens, line 400: Can this no longer return `Result`?

PR #5027, opened 2023-11-28
Add docker instruction on main Readme

marc-aurele-besner:
## Add docker instruction on main Readme

Since nowhere on the main README of this repo it mentions docker, when I had first look at it, I did not realize it already had a docker compose file. It's only when someone point out to me that it had a docker folder that I realize it was there. 🤦‍♂️ 

So I'm suggesting simply adding a small mention about it


lutter (approved):
Nice catch! Yes, the toplevel README leads a lot more edits.

PR #5029, opened 2023-11-29
Allow (autogenerated) `Int8` id values

### LINKED ISSUE #5018: [Feature] Auto-incrementing ids
### Description

In a lot of situations, the `id` of an entity doesn't really carry any meaning, and it's only used to distinguish between different entities of the same type. Unfortunately, it's also often not easy to come up with unique id's from within the mappings because the environment is very restricted. That often leads to very long `id` values that are simply a combination of other attributes of the entity that duplicate data and therefore waste space.

To address that, we should do the following:
* Allow `Int8` as the type of an `id` field
* When `store.set(entity_type, id, data)` is called and the `id` argument is the string `"auto"`, set the `id` value of the entity to the next value from a sequence that `graph-node` maintains internally
* Calling `store.get(entity_type, id)` with `id = "auto"` will return `null`

The implementation for the internal sequence could use a Postgres sequence, but that will require a lot of roundtrips to the database. Instead, we should do the following:
* When a subgraph is started, query the `max(id) + 1` for all entities that have an `id` of type `Int8` and use that as initial values for a map of sequences in `IndexingState`
* Whenever we need a new `id` for an `"auto"` entity, get a new value from `IndexingState` and update the sequence there appropriately

Most of this can be implemented entirely within `graph-node`. The main thing that needs to change in `graph-tooling` is that we accept `id: Int8` in subgraph schemas and do a similar automatic conversion to `String` for the various `store.*` calls that we do for `Bytes`.

lutter:
With this change, it is possible to use `Int8` as the type of an entity's `id`. In addition, when `store.set` is called with the id argument set to `"auto"`, an id is auto-generated (see the comment on `IdType::generate_id` for details)

Auto-generation is also possible if the type of `id` is `Bytes`

For now, this is really only useful for immutable entities as it's not possible to find out what the id was set to from the mappings; we should address that in a future PR.

Fixes #5018


graph/src/schema/api.rs
```diff
@@ -624,6 +624,7 @@ fn id_type_as_scalar(
     }?;
     let scalar_type = id_type.map(|id_type| match id_type {
         IdType::String | IdType::Bytes => ScalarType::new(String::from("String")),
+        IdType::Int8 => ScalarType::new(String::from("Int")),
```
leoyvens, line 627: Shouldn't this be `"Int8"`?

lutter: Nice catch, and a fun way to spend a few hours why this works with `Int` but not with `Int8` - the upshot is that `Int8` gets treated as a string internally in too many places, and queries just blow up in that case. Addressing that is a little scary and I've opened the separate PR #5034 to do that.

For this PR, I just left a big comment why it's `Int` and not `Int8` for now.

leoyvens (approved):

PR #5030, opened 2023-11-29
feat: Starknet data source validation with improvements

xJonathanLEI:
Long overdue follow-up PR of #4895 to resolve [this comment](https://github.com/graphprotocol/graph-node/pull/4895#discussion_r1341301815) from @mangas.

Main changes:

1. renames `block_handlers` to `block_handler` (and type changed from `Vec<_>` to `Option<_>`) because it makes more sense.
2. implements data source validation with `validate()` (this function previously always returns an empty array).

Also adds data source validation tests to `chain/starknet/src/chain.rs`, similar to those found in `chain/near/src/chain.rs`. This does not include trigger filter tests in `chain/near/src/chain.rs` though - trigger filter hasn't been implemented for Starknet, as noted in the original PR, and should be addressed with another follow-up PR.

In the process of implementing validation, I noticed that the `DataSource` trait has a method `address()` that returns a `Option<&[u8]>`, which is impossible to implement with the previous use of `FieldElement` in the data source, as this type doesn't have a `[u8]` backing. The type uses the Montgomery representation internally for fast finite field mulitplication, at the cost of slower (de)serialization. However, in `graph-node`, this type is only used for equality comparison, which calls for a type that's backed by a `u8` array instead.

Therefore, this PR also adds a new type `Felt` that's a wrapper around `[u8; 32]`, and replaces any `FieldElement` occurrence. The `starknet-ff` dependency is also removed.


mangas (approved):

xJonathanLEI:
Thanks @mangas for the review and approval. I've rebased the branch to the latest `master` (`f016503e3`) due to conflicts introduced by dependabot bumping `starknet-ff`, as we're removing it completely here. No changes were made other than resolving this specific conflict.

xJonathanLEI:
Hi @incrypto32, I noticed that the base branch has been changed to the `v0.34.0` tree. Will this get ported to `master` once it's reviewed and merged? Asking cuz I have a follow-up PR that depends on this one, so it would be nice if these changes are available on `master` soon too. Thanks!

xJonathanLEI:
Thanks for confirming!

PR #5032, opened 2023-11-30
Add history_blocks to manifest

### LINKED ISSUE #3667: Provide a control for subgraph pruning in the subgraph manifest
With #3666, operators can control pruning of individual subgraphs, but it requires user action for each subgraph that should be pruned.

There should be a control that automatically prunes subgraphs without the need to explicitly run `graphman prune`. There are a few options for this:

* Make pruning part of the subgraph manifest so that subgraph authors can indicate how much history should be retained
* Allow operators to set how much history to retain through configuration, e.g., for each chain in `graph-node.toml`, or through additional rules in `graph-node.toml` based on subgraph names

Making pruning part of the manifest has the downside that subgraphs that only differ in how much history they retain will have different IPFS hashes, even though they are really the same subgraph, and can all be used to serve the same queries, assuming that the query is for a block height within the subgraph's available history. As an example, a subgraph that retains all history and a subgraph that only retains the last 1000 blocks of history can both be used to respond to a query at a block that is less than 1000 blocks behind the subgraph head.

incrypto32:
Closes #3667

### DIFF
--- a/graph/src/data/subgraph/mod.rs
+++ b/graph/src/data/subgraph/mod.rs
@@ -542,6 +542,19 @@ pub struct BaseSubgraphManifest<C, S, D, T> {
     pub templates: Vec<T>,
     #[serde(skip_serializing, default)]
     pub chain: PhantomData<C>,
+    pub indexer_hints: Option<IndexerHints>,
+}
+
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct IndexerHints {
+    pub prune: Option<PruneConfig>,
+}
+
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct PruneConfig {
+    pub history_blocks: Option<BlockNumber>,
 }
 
 /// SubgraphManifest with IPFS links unresolved
@@ -674,6 +687,13 @@ impl<C: Blockchain> SubgraphManifest<C> {
             .collect()
     }
 
+    pub fn history_blocks(&self) -> Option<BlockNumber> {
+        self.indexer_hints
+            .as_ref()
+            .and_then(|h| h.prune.as_ref())
+            .and_then(|p| p.history_blocks)
+    }
+
     pub fn api_versions(&self) -> impl Iterator<Item = semver::Version> + '_ {
         self.templates
             .iter()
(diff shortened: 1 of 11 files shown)


graph/src/data/subgraph/mod.rs
```diff
@@ -542,6 +542,13 @@ pub struct BaseSubgraphManifest<C, S, D, T> {
     pub templates: Vec<T>,
     #[serde(skip_serializing, default)]
     pub chain: PhantomData<C>,
+    pub indexer_hints: Option<IndexerHints>,
+}
+
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct IndexerHints {
+    pub history_blocks: Option<BlockNumber>,
```
incrypto32, line 551: @leoyvens how does this look? If we have ongoing pruning do we need to the `prune` hint to do the manual pruning?

```
indexingHints:
   prune:
    historyBlocks: 100
  ```

File: store/test-store/tests/chain/ethereum/manifest.rs [hunk 2]
```diff
@@ -187,6 +187,58 @@ specVersion: 0.0.2
...
+  block: 12345
+specVersion: 0.0.2
+indexerHints:
+  prune:
+";
+
+    let manifest = resolve_manifest(YAML, SPEC_VERSION_0_0_4).await;
+
+    assert_eq!("Qmmanifest", manifest.id.as_str());
+    assert!(manifest.indexer_hints.unwrap().prune.is_some());
```
lutter: I don't understand this test - what is this trying to ensure?

lutter (approved):
LGTM.

store/test-store/tests/chain/ethereum/manifest.rs
incrypto32: Oh i used that test to test out the indexer_hints parsing is working properly.

incrypto32: Will remove it.

incrypto32: its was removed already. merging

lutter: Just a comment what the intent of the test was would have been fine, too - it was just a bit confusing.

PR #5034, opened 2023-11-30
graph, graphql, store: Coerce Int8 values to ints

lutter:
Previously, they were coerced to strings which ultimately leads to errors when lists of int8 are used in queries

@dotansimha Do you remember why `Int8` was initially coerced into `String` in a lot of places? The whole topic of coercion is really confusing to me; part of it is because the `Value` of `graphql-parser` can't represent all the values we deal with internally (like bytes or int4 vs int8) and so you are forever trying to figure out if a string is really a string or whether it's a byte array. The other part is that coercion happens all over the place and it's really hard to draw a line between 'here we are dealing with values that have fairly random types and aren't checked yet' vs 'these values have been properly checked and express the proper type'. I wonder if we should expand the `r::Value` type to faithfully represent all value types, and take the conversion from `q::Value` to `r::Value` as the point where we go from unchecked to checked.

### DIFF
--- a/graph/src/data/store/mod.rs
+++ b/graph/src/data/store/mod.rs
@@ -518,7 +518,7 @@ impl From<Value> for r::Value {
         match value {
             Value::String(s) => r::Value::String(s),
             Value::Int(i) => r::Value::Int(i as i64),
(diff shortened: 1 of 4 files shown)


lutter:

PR #5038, opened 2023-12-04
Make `function not found` a deterministic

### LINKED ISSUE #3059: Make `function not found` a deterministic failure
Bug

If a subgraph handler calls a function that doesn't exist, the subgraph fails with `deterministic: false`, and the following error log:
```
failed to process trigger: block #13380099 (0x38d4…b3aa), transaction ca32b1581b837c6ca539cf6c2fd542e9bcacbdc345c6a424382b4f0e48b96153: function handleSwapped not found
```
[Example subgraph](https://ipfs.io/ipfs/QmWXF3jTLo6Wy9MxDZz6WKhET3GFcSFqn6vmrDDYPF7UHK)

This handler will never run successfully, so this error should be deterministic

incrypto32:
Closes #3059

### DIFF
--- a/runtime/wasm/src/module/mod.rs
+++ b/runtime/wasm/src/module/mod.rs
@@ -244,54 +244,73 @@ impl<C: Blockchain> WasmInstance<C> {
         let func = self
             .instance
             .get_func(handler)
-            .with_context(|| format!("function {} not found", handler))?;
-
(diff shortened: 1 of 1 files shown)


leoyvens (approved):
Thanks for refactoring this code, it isn't pretty.

PR #5048, opened 2023-12-05
Fix typo at sharding.md

rootwarp:


lutter (approved):
Thanks a ton for the fix!

PR #5049, opened 2023-12-05
graph: Exclude Self from ObjectType.share_interfaces()

lutter:
The conflicting entity check in DeploymentStore expects that the type itself is not in the list of types that share an interface; if it is, the check causes spurious failures when entities are updated.


incrypto32 (approved):
LGTM

PR #5051, opened 2023-12-06
Fix offchain created from offchain

leoyvens:
This is blocking the update to the network subgraph, which now uses file data sources within file data sources.

This is more easily reviewed commit by commit. The first four commits are strictly refactors.


core/src/subgraph/context/instance.rs
```diff
@@ -89,34 +90,31 @@ where
...
                 // monitor data source only if it's not processed.
                 if !ds.is_processed() {
                     offchain_monitor.add_source(ds.source.clone())?;
                 }
             }
 
-            let host = this.new_host(logger.cheap_clone(), ds, module_bytes)?;
-            this.hosts.push(Arc::new(host));
+            let Some(host) = this.new_host(logger.cheap_clone(), ds)? else { continue };
+            this.hosts.push(host);
```
lutter: I think this would be cleaner as a `if let Some(..)` so we can avoid the `continue`

core/src/subgraph/context/instance/hosts.rs
```diff
@@ -103,3 +111,90 @@ impl<C: Blockchain, T: RuntimeHostBuilder<C>> Hosts<C, T> {
         )
     }
 }
+
+/// Note that unlike `OnchainHosts`, this does not maintain the order of insertion. Ultimately, the
+/// processing order should not matter because as each offchain ds has its own causality region.
```
lutter, line 125: Small typo: `s/because as each/because each/`

core/src/subgraph/context/instance/mod.rs
```diff
@@ -24,9 +24,11 @@ pub(super) struct SubgraphInstance<C: Blockchain, T: RuntimeHostBuilder<C>> {
     /// The hosts represent the data sources in the subgraph. There is one host per data source.
     /// Data sources with no mappings (e.g. direct substreams) have no host.
     ///
-    /// The runtime hosts must be created in increasing order of block number.
-    /// `fn hosts_for_trigger` will return the hosts in the same order as their were inserted.
-    hosts: Hosts<C, T>,
+    /// Onchain hosts must be created in increasing order of block number. `fn hosts_for_trigger`
+    /// will return the onchain hosts in the same order as their were inserted.
```
lutter, line 28: `s/as their were/as they were/`

core/src/subgraph/context/instance/mod.rs
```diff
@@ -127,28 +130,44 @@ where
...
         let Some(host) = self.new_host(logger.clone(), data_source)? else { return Ok(None) };
 
-        Ok(if self.hosts.contains(&host) {
-            None
-        } else {
-            self.hosts.push(host.cheap_clone());
-            Some(host)
-        })
+        // Check for duplicates and add the host.
+        match is_onchain {
```
lutter, line 144: Wouldn't a simple `if` be clearer here?

core/src/subgraph/context/instance/mod.rs
```diff
@@ -127,28 +130,44 @@ where
...
+            true => {
+                // `onchain_hosts` will remain ordered by the creation block.
+                // See also 8f1bca33-d3b7-4035-affc-fd6161a12448.
+                assert!(
+                    self.onchain_hosts
+                        .last()
+                        .and_then(|h| h.creation_block_number())
+                        <= host.data_source().creation_block()
+                );
```
lutter, line 153: I am always a bit queasy with asserts since it's not that easy to spot panics in the logs; could we return an error here? That way, the subgraph will fail and the failure can tell use what exactly is wrong.

lutter (approved):

tests/tests/runner_tests.rs
```diff
@@ -180,6 +180,63 @@ async fn data_source_revert() -> anyhow::Result<()> {
...
     );
 
+    // This is an entirely different test, but running it here conveniently avoids race conditions
+    // since it uses the same deployment id.
+    data_source_long_revert().await.unwrap();
+
+    Ok(())
+}
+
+async fn data_source_long_revert() -> anyhow::Result<()> {
```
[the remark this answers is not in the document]
leoyvens: Yhea I'll squash merge

core/src/subgraph/context/instance.rs
leoyvens: This code changes in a further commit.

leoyvens:
@lutter thanks for the review! I've addressed it and removed the stray `dbg!`s I forgot to remove in https://github.com/graphprotocol/graph-node/pull/5083.

PR #5053, opened 2023-12-06
fix ethereum codec: should not be checking block header parent with all 0s

sduchesneau:
When checking the block parent header, ignore header with 32 * [0] (it is actually not length 0 when on genesis block)
This could panic if the head block of the chain happens to be the genesis, ex: on a local dev chain

### DIFF
--- a/chain/ethereum/src/codec.rs
+++ b/chain/ethereum/src/codec.rs
@@ -395,6 +395,7 @@ impl BlockHeader {
     pub fn parent_ptr(&self) -> Option<BlockPtr> {
         match self.parent_hash.len() {
             0 => None,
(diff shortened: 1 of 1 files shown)


chain/ethereum/src/codec.rs
```diff
@@ -395,6 +395,7 @@ impl BlockHeader {
     pub fn parent_ptr(&self) -> Option<BlockPtr> {
         match self.parent_hash.len() {
             0 => None,
+            _ if self.parent_hash.iter().all(|x| *x == 0) => None,
```
leoyvens, line 398: We can now remove the above `0 => None` branch, noting in a comment that this check will also return `None` if the len is 0.

leoyvens (changes requested):

PR #5057, opened 2023-12-08
Fix an issue with API schema generation for interfaces

lutter:
The code that determined the id type of an interface was completely wrong, and panicked to boot.

In general, the API schema generation code needs a good rewrite as it's a complete mess of hard-to-follow navigation of a raw GraphQL AST when it should make use of the more accessible information in `InputSchema`


leoyvens (approved):

lutter:
This was merged - I just screwed up some git commands and confused github

PR #5065, opened 2023-12-12
graph, store: Batch writes for synced subgraphs when possible

lutter:


lutter:
Closing this as PR #5252 should achieve something very similar by outright turning batching back on

PR #5067, opened 2023-12-12
Make the RecentBlocksCache very dumb

lutter:
This changes the `RecentBlocksCache` to be just a map from block number to block that keeps the `capacity` last entries.

The current data structure, a `BTreeMap` is probably not ideal for this seeing how few entries this cache usually has, and a `Vec<(BlockNumber, Box<JsonBlock>)>` would probably perform better since the whole cache would fit into a few cache lines.


leoyvens (approved):
Ignorance is strength

PR #5069, opened 2023-12-13
fix(graphql): remove duplicated input type in meta schema

saihaj:
I was trying to import this `meta.graphql` schema remotely in GraphQL CLI for some validation work and saw that we are re-declaring the same input type which is invalid.

### DIFF
--- a/graph/src/schema/meta.graphql
+++ b/graph/src/schema/meta.graphql
@@ -56,12 +56,6 @@ input BlockChangedFilter {
   number_gte: Int!
 }
 
-input Block_height {
-  hash: Bytes
-  number: Int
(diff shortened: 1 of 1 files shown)


lutter (approved):
Nice catch!

lutter:
Hmm .. introspection tests are failing. Something probably changed in the introspection result

mangas:
What is the status on this PR @dotansimha @saihaj ? Is it still needed? The tests were not passing so it can't be merged as is.

saihaj:
we should clean this up but not sure why tests are failing, I was using it for something which I no longer am so not critical at the moment for what I was doing

PR #5070, opened 2023-12-13
Change health type

zorancv:


lutter:
If all else fails, we could just make that column `text`; it would require a proper migration, something along the lines of
- add a `health_new text` column
- update `subgraph_deployment` and set `health_new = health::text`
- drop column `health`
- rename column `health_new` -> `health`

zorancv:
> If all else fails, we could just make that column `text`; it would require a proper migration, something along the lines of
> 
> * add a `health_new text` column
> * update `subgraph_deployment` and set `health_new = health::text`
> * drop column `health`
> * rename column `health_new` -> `health`

I see. Thanks for the tips. Not sure how reliable down migration would be.

PR #5076, opened 2023-12-14
Make the recent blocks cache more effective for concurrent lookups

lutter:
With these changes, we make sure that only one of possibly multiple pending attempts to lookup blocks by hash queries the database and the other attempts use cached data.

This is done by keying off the list of hashes that are to be looked up; we would really like to do this for each hash individually, but that would require a lot more implementation effort. This will help for cases where a lot of subgraphs are close to the chain head and only look up a single block hash, the chain head.


store/postgres/src/chain_store.rs
```diff
@@ -1894,10 +1908,40 @@ impl ChainStoreTrait for ChainStore {
...
+                };
+                let lookup_herd = self.lookup_herd.cheap_clone();
+                let logger = self.logger.cheap_clone();
+                let (BlocksLookupResult(res), _) =
+                    lookup_herd.cached_query(hash, lookup_fut, &logger).await;
+                // Try to avoid cloning a non-concurrent lookup; it's not
+                // entirely clear whether that will actually avoid a clone
+                // since it depends on a lot of the details of how the
+                // `HerdCache` is implemented
+                let res = Arc::try_unwrap(res).unwrap_or_else(|arc| (*arc).clone());
```
leoyvens, line 1933: My impression is also that this will work, clever!

store/postgres/src/chain_store.rs
```diff
@@ -1894,10 +1908,40 @@ impl ChainStoreTrait for ChainStore {
...
+                let stored = match res {
+                    Ok(blocks) => {
+                        for block in &blocks {
+                            self.recent_blocks_cache.insert_block(block.clone());
```
leoyvens, line 1937: Why must we clone again here, is `blocks` not owned?

lutter: We need one copy to stick into the cache, and one copy to return (`blocks()` returns owned values) I thought about sticking that all into an `Arc`, but the use of this in `etherum_adapter` needs an owned value for `serde::from_value`, so one way or another we'll end up cloning

leoyvens (approved):

PR #5077, opened 2023-12-14
Migration changing the type of health column to text

zorancv:
It is needed in the context of the conversion to diesel2, since the later is not able to properly handle health columns enum.


store/postgres/src/deployment.rs
```diff
@@ -40,6 +40,7 @@ use crate::{block_range::BLOCK_RANGE_COLUMN, primary::Site};
 use graph::constraint_violation;
 
 #[derive(DbEnum, Debug, Clone, Copy)]
+#[PgType = "text"]
```
lutter, line 43: Does leaving the `DbEnum` there automatically convert the `text` to `enum SubgraphHealth`? If so, nice!

store/postgres/migrations/2023-12-14-160628_change_health_column_format/down.sql
```diff
@@ -0,0 +1,23 @@
...
+
+alter table
+    subgraphs.subgraph_deployment
+drop column
+    health;
+
+alter table
+    subgraphs.subgraph_deployment
+rename column
+    health_new to health;
```
lutter, line 23: To put things back the way they were, I think this also needs a `not null` constraint.

lutter (approved):

store/postgres/src/deployment.rs
zorancv: Yep. I find this approach much nicer than the alternative where we'd do it ourselves like [this](https://github.com/graphprotocol/graph-node/pull/5125/files#diff-4f0442fb210dcfd363ebb2af35c9b706b76d1663cf35adb945ae49054c85f2e9), which I think is much more verbose and error prone.

store/postgres/migrations/2023-12-14-160628_change_health_column_format/down.sql
zorancv: Yes, that's correct. It is fixed now.

PR #5082, opened 2023-12-15
Aggregations (write side)

lutter:
This PR introduces aggregations (see docs/aggregations.md for details) It only implements the write side, i.e., with this pull request, we populate aggregation tables but there is not yet a way to query aggregations.

Until we have a native `Timestamp` type, timestamps are represented as an `Int8` as seconds since the Unix epoch.

An example subgraph that aggregates ethereum block numbers can be found [here](https://github.com/graphprotocol/common-subgraphs/tree/main/aggregation-example) To use it, `graph-node` must be run with `GRAPH_MAX_SPEC_VERSION="0.1.1"`. To deploy it, `graph-cli` needs to be patched:
* Check out the `graph-tooling` repo, using the `lutter/agg-minimal` branch ([this PR](https://github.com/graphprotocol/graph-tooling/pull/1548))
* Run `pnpm install && pnpm build` in the `graph-tooling` checkout
* Run `yarn install && pnpm link <path to graph-tooling/packages/cli>` in the checkout of the subgraph
* Deploy the subgraph against a local `graph-node`


chain/substreams/src/chain.rs
```diff
@@ -57,6 +57,10 @@ impl blockchain::Block for Block {
     fn parent_ptr(&self) -> Option<BlockPtr> {
         None
     }
+
+    fn timestamp(&self) -> BlockTime {
+        BlockTime::NONE
```
mangas, line 62: Should this type be called something like MaybeBlockTime or something just to indicate it may be none even though it's not an option? 

I guess the substreams notion of blocks doesn't fit the aggregations model anyway

docs/aggregations.md
```diff
@@ -0,0 +1,178 @@
...
+aggregations, and it doesn't seem like it used in practice, we won't
+initially support it. (same for variance, stddev etc.)
+
+**TODO** The timeseries type can be simplified for some situations if
+aggregations can be done over expressions, for example over `priceUSD *
+amount` to track `totalVolumeUSD`
+
+**TODO** It might be necessary to allow `@aggregate` fields that are only
+used for some intervals. We could allow that with syntax like
+`@aggregate(fn: .., arg: .., interval: "day")`
```
mangas, line 96: feels like more complicated aggregations could benefit from a mapping to do them? Essentially providing a (x, x) -> x function to generalise.

graph/src/schema/input_schema.rs
```diff
@@ -395,6 +435,332 @@ impl EnumMap {
     }
 }
 
+#[derive(PartialEq, Debug, Clone)]
+pub enum AggregateFn {
+    Sum,
+    Max,
+    Min,
+    Cnt,
```
mangas, line 443: I think we don't lose much by calling this Count

graph/src/schema/input_schema.rs
```diff
@@ -395,6 +435,332 @@ impl EnumMap {
...
+impl AggregationMapping {
+    pub fn source_type(&self, schema: &InputSchema) -> EntityType {
+        let source = self.aggregation(schema).source;
+        EntityType::new(schema.cheap_clone(), source)
+    }
+
+    pub fn aggregation<'a>(&self, schema: &'a InputSchema) -> &'a Aggregation {
+        schema.inner.type_infos[self.aggregation]
+            .aggregation()
+            .expect("the aggregation source is an object types")
```
mangas, line 588:
```suggestion
            .expect("the aggregation source is an object type")
```

graph/src/schema/input_schema.rs
```diff
@@ -395,6 +435,332 @@ impl EnumMap {
...
+    }
+
+    pub fn agg_type(&self, schema: &InputSchema) -> EntityType {
```
mangas, line 591: I think "aggregation_type" is also more readable

graph/src/schema/input_schema.rs
```diff
@@ -395,6 +435,332 @@ impl EnumMap {
...
+            func,
+            arg,
+            field_type: field_type.clone(),
+            value_type: field_type.get_base_type().parse().unwrap(),
+        }
+    }
+
+    /// The field needed for the finalised aggregation for hourly/daily
+    /// values
+    fn as_agg_field(&self) -> Field {
```
mangas, line 653: same here

graph/src/schema/input_schema.rs
```diff
@@ -395,6 +435,332 @@ impl EnumMap {
...
+    pub aggregates: Box<[Aggregate]>,
+    /// The object types for the aggregated data, one for each interval, in
+    /// the same order as `intervals`
+    obj_types: Box<[ObjectType]>,
+}
+
+impl Aggregation {
+    pub fn new(schema: &Schema, pool: &AtomPool, agg_type: &s::ObjectType) -> Self {
+        let name = pool.lookup(&agg_type.name).unwrap();
+        let id_type = IdType::try_from(agg_type).expect("validation caught any issues here");
```
mangas, line 680: Should this return an error instead?

graph/src/schema/input_schema.rs
```diff
@@ -395,6 +435,332 @@ impl EnumMap {
...
+    pub fields: Box<[Field]>,
+    pub aggregates: Box<[Aggregate]>,
+    /// The object types for the aggregated data, one for each interval, in
+    /// the same order as `intervals`
+    obj_types: Box<[ObjectType]>,
+}
+
+impl Aggregation {
+    pub fn new(schema: &Schema, pool: &AtomPool, agg_type: &s::ObjectType) -> Self {
+        let name = pool.lookup(&agg_type.name).unwrap();
```
mangas, line 679: I think these cases could benefit from a lookup_unchecked or something which does the unwrap. There's no scenario where this lookup can fail? I guess the ObjectType parsing probably adds it here but would probably be worth adding a comment. 

There are a couple of similar cases around

mangas (approved):

chain/substreams/src/chain.rs
lutter: The `NONE` thing should really go away (maybe except for tests) I just put it there for the few cases where I wasn't sure what's happening (like substreams)

graph/src/schema/input_schema.rs
lutter: Heh, yeah, not sure why I wrote that. Changed it

graph/src/schema/input_schema.rs
lutter: No, we should never get here since validations reject a subgraph where the type of `id` would fail this conversion. The problem is that we first do validations, and then extract things from the validated AST. It would be much better to do the extraction at the same time as validation, but that's a pretty big change that I didn't want to also shoehorn into this PR.

A big part of this and a couple previous PR's is to isolate these kinds of conversions by only doing them while we construct the `InputSchema`; previously, a lot of this happened everywhere where the AST was used (since there were no internal data structures to represent our idea of schema information)

graph/src/schema/input_schema.rs
lutter: Any failure here is a programming error since the `agg_type.name` should have been interned already. All a `lookup_unchecked` could do is `lookup(..).unwrap()` anyway, but the knowledge that this lookup can't fail (except for genuine bugs) resides here, in the usage of the pool.

chain/substreams/src/mapper.rs
```diff
@@ -59,8 +60,16 @@ impl BlockStreamMapper<Chain> for WasmBlockMapper {
 
         let block_data = block.into_boxed_slice();
 
+        // This is not a great idea: we really always need a timestamp; if
+        // substreams doesn't give us one, we use a fixed one which will
+        // lead to all kinds of strange behavior
+        let timestamp = timestamp
```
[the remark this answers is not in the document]
lutter: Since it should always be present, I just made it missing an error (which gets rid of one of the iffy uses of `BlockTime::NONE`)

PR #5090, opened 2023-12-21
core, graph: Add a metric to capture entire processing time for a block

lutter:
We don't have a metric that accurately reflects the time it takes to process a block. This PR introduces two new metrics for each deployment, one for the total time taken to process a block, and one for the number of blocks that should make it possible to calculate an overall processing speed. The metrics we have only cover a specific section of processing (and I am not sure how useful the histograms in `SubgraphInstanceMetrics` actually are)

There's a bit of handwaving since we can't account for the time a task is not actively running with this, but it should give us a reasonably close idea of overall timing.


zorancv (approved):
Looks good to me.

PR #5092, opened 2023-12-22
fix(fds): Avoid constraint violation for fds-in-fds

leoyvens:


store/postgres/src/dynds/private.rs
```diff
@@ -160,7 +160,9 @@ impl DataSourcesTable {
                     causality_region,
                 } = ds;
 
-                if creation_block != &Some(block) {
+                // Nested offchain data sources might not pass this check, as their `creation_block`
+                // will be their parent's `creation_block`, not necessarily `block`.
+                if causality_region == &CausalityRegion::ONCHAIN && creation_block != &Some(block) {
```
mangas, line 165: Should a test be added to prevent regression?

mangas (approved):
As per my comment, I think we should add a test so perhaps you can create an issue to come back to it when you can

leoyvens:
Thanks, I've opened an issue https://github.com/graphprotocol/graph-node/issues/5105

PR #5107, opened 2024-01-04
fix(graphman): Allow rewinding past the graft block

leoyvens:
When manually rewinding, sometimes we need to do strange things such as rewinding past a graft block. This is very likely to result in an incorrect PoI, but we have a situation right now where we just need to do this to work around a bug in grafting.

The sanity check is kept for normal reverts.


mangas (approved):
This should have a test to ensure there isn't a regression. At least an issue should be created so it doesn't get forgotten. Approved due for expediency

Issue #5111, opened 2024-01-05
[Bug] Deduplication logic for Cosmos Event is faulty leading to missed event


betterclever:
### Bug report

Hi team!
I am an engineer at Persistence Labs where we maintain a self-hosted deployment of Graph Node for usage with Persistence Core-1 chain (a Cosmos SDK based L1). 

I noticed that a few events were being missed in our subgraph which were similar but not exactly duplicates. Only first of them was being indexed, and the rest were being ignored and the handler actually never reached the subgraph.

On going deep, I realized this is happening due to a bug in the dedup logic for the events for the Cosmos Blockchain.

The dedup call happens here:

https://github.com/graphprotocol/graph-node/blob/8c6edc195102ee11a0767c77201a043aac45366b/graph/src/blockchain/block_stream.rs#L225

Rust stdlib dedup function uses `PartialEq` implementation for the dedup.

The `PartialEq` for BlockStream is defined here:

https://github.com/graphprotocol/graph-node/blob/8c6edc195102ee11a0767c77201a043aac45366b/chain/cosmos/src/trigger.rs#L75-L100

Particularly, the line 90 has an issue where only the source and event name is checked for duplicate. There's a possibility that events with the same name are emitted multiple times in the same block in Cosmos in the same transaction with different attribute values.

## Error Analysis and reproducible scenario

I have made a fork branch with a test-case where I downloaded a block from the PersistenceCore Core-1 Firehose and copied it's base64 data. I later processed the triggers exactly as in the current 0.33 release, and then ran the dedup operation to demonstrate that non-duplicate data is being removed leading to ignored events during indexing.

Link to test case in the fork: 
https://github.com/graphprotocol/graph-node/blob/01f3f435341ba124a38b420895d05e9c1f42df0b/chain/cosmos/src/chain.rs#L470

Run the test using:
`cargo test --package graph-chain-cosmos --lib -- chain::test::test_decode_block --exact --nocapture`

## Fix

The fix is pretty straightforward. We need to update the `PartialEq` logic also include block data scenario. Only the name and source doesn't suffice. 

### Possible remediation for older releases where wrong data was indexed

As I can imagine, a lot of past indexed data must have been missed due to this issue. All those subgraphs would have to be re-indexed on a fixed Graph-node release binary to have data without any missing pieces.

### Some information to help us out

- [X] Tick this box if this bug is caused by a regression found in the latest release.
- [ ] Tick this box if this bug is specific to the hosted service.
[...]

PR #5112, opened 2024-01-05
fix dedup issue for cosmos attributes

betterclever:
fixes issue: #5111

Introduces additional checks in the event PartialEq comparison check so that no event is missed due to the dedup operation.


leoyvens (approved):

leoyvens:
This needs to be rustfmt'd

betterclever:
Done.

leoyvens:
Looks great, thanks!

PR #5113, opened 2024-01-05
Attempted fix for `fn fatal_error` test spurious failure

leoyvens:
I think this is happening because we have two runner tests using the same subgraph, `fatal_error` and `poi_for_deterministically_failed_sg`. I ended up addressing this by merging the tests.

The first commit is a refactor that ended up not being relevant, but is maybe still worth it as a refactor for readability. The second commit is the actual fix.

I'll run this a few times in CI to check if we don't get the spurious failure anymore.


leoyvens:
Five green runs in a row, I declare victory over the spurious failures.

PR #5114, opened 2024-01-05
migration: Fix up any inconsistent data_sources tables

leoyvens:
This is a script by @avandras which we've tested with manual runs. Because the bug it fixes is present in version 0.33, it's important that we get this in for version 0.34.

### DIFF
--- a/store/postgres/migrations/2024-01-05-170000_ds_corruption_fix_up/up.sql
+++ b/store/postgres/migrations/2024-01-05-170000_ds_corruption_fix_up/up.sql
@@ -1,0 +1,96 @@
+/*
+
+
+This migration exists to fix up DB inconsistencies caused by a bug that was already fixed in this PR:
+https://github.com/graphprotocol/graph-node/pull/5083
+
+What it does is:
+1. select latest_ethereum_block_number from subgraphs.subgraph_deployment where deployment = '$Qm..';
+2. With that value, check if there are any data sources with a higher block:
+   select count(*) from sgd*.data_sources$ where lower(block_range) > $latest_block;
+3. If there are, then we need to intervene in that subgraph by removing all such data sources:
+   delete from sgd*.data_sources$ where lower(block_range) > $latest_block;
+4. It also unclamps any data sources that need unclamping (even though I don't think we ever clamp them).
+*/
+
+CREATE TEMPORARY TABLE temp_sgd_last_block (
+    deployment_schema text,
+    last_block_from_registry numeric,
+    is_ok boolean,
+    is_upper_limit_too_high boolean
+);
+
+-- collect the latest block number from SG deployment details
+INSERT INTO temp_sgd_last_block (deployment_schema, last_block_from_registry)
+SELECT d.name, latest_ethereum_block_number
+  FROM subgraphs.subgraph_deployment AS sd
+  JOIN deployment_schemas AS d ON sd.deployment = d.subgraph
+ WHERE d.name IN (SELECT relnamespace::regnamespace::name
+                    FROM pg_class
+                   WHERE relname LIKE 'data_sources$%' AND relkind = 'r'
+                 );
+
+-- check if the block numbers in the tables are OK
+-- if not, it can be
+-- - either the lower bound is higher than the last block
+-- - or the lower is fine, but the higher is past the last block
+DO $do$
+DECLARE
+    s text;
+BEGIN
+    FOR s IN SELECT deployment_schema FROM temp_sgd_last_block
+    LOOP
+        EXECUTE format($$
+        UPDATE temp_sgd_last_block
+           SET is_ok = NOT EXISTS (SELECT 1
+                                     FROM %1$I."data_sources$"
+                                    WHERE lower(block_range) > last_block_from_registry
+                                    LIMIT 1)
+                       AND NOT EXISTS (SELECT 1
+                                         FROM %1$I."data_sources$"
+                                        WHERE upper(block_range) > last_block_from_registry 
+                                          AND lower(block_range) <= last_block_from_registry
+                                        LIMIT 1),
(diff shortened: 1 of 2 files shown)


store/postgres/migrations/2024-01-05-170000_ds_corruption_fix_up/up.sql
```diff
@@ -0,0 +1,96 @@
...
+    last_block_from_registry numeric,
+    is_ok boolean,
+    is_upper_limit_too_high boolean
+);
+
+-- collect the latest block number from SG deployment details
+INSERT INTO temp_sgd_last_block (deployment_schema, last_block_from_registry)
+SELECT d.name, latest_ethereum_block_number
+  FROM subgraphs.subgraph_deployment AS sd
+  JOIN deployment_schemas AS d ON sd.deployment = d.subgraph
```
lutter, line 27: You want to use `primary_public.deployment_schemas` here as that's the authoritative source. In shards other than `primary`, `public.deployment_schemas` is a backup copy that gets refreshed occasionally.

store/postgres/migrations/2024-01-05-170000_ds_corruption_fix_up/up.sql
```diff
@@ -0,0 +1,96 @@
...
+ WHERE d.name IN (SELECT relnamespace::regnamespace::name
+                    FROM pg_class
+                   WHERE relname LIKE 'data_sources$%' AND relkind = 'r'
+                 );
```
lutter, line 31: I don't quite follow what's going on here - this is to find all deployments that have a `data_sources$` table? Might be simpler to join on `information_schema.tables.table_schema where information_schema.tables.table_name = 'data_sources$'`

store/postgres/migrations/2024-01-05-170000_ds_corruption_fix_up/up.sql
```diff
@@ -0,0 +1,96 @@
...
+        UPDATE temp_sgd_last_block
+           SET is_ok = NOT EXISTS (SELECT 1
+                                     FROM %1$I."data_sources$"
+                                    WHERE lower(block_range) > last_block_from_registry
+                                    LIMIT 1)
+                       AND NOT EXISTS (SELECT 1
+                                         FROM %1$I."data_sources$"
+                                        WHERE upper(block_range) > last_block_from_registry 
+                                          AND lower(block_range) <= last_block_from_registry
+                                        LIMIT 1),
```
lutter, line 53: Don't these two conditions just boil down to `where upper(block_range) > last_block_from_registry`? In the first branch if `lower(block_range) > last_block_from_registry`, then the same condition holds with `upper`. Basically, we see if there's a data source that's visible beyond `last_block_from_registry`

store/postgres/migrations/2024-01-05-170000_ds_corruption_fix_up/up.sql
```diff
@@ -0,0 +1,96 @@
...
+                                                  WHERE upper(block_range) > last_block_from_registry 
+                                                    AND lower(block_range) <= last_block_from_registry
+                                                  LIMIT 1)
+         WHERE deployment_schema = '%1$s'  
+                $$, s);
+    END LOOP;
+END;
+$do$;
+
+SELECT * FROM temp_sgd_last_block WHERE NOT is_ok;
```
lutter, line 65: What is this for? Just to show which subgraphs are not ok?

store/postgres/migrations/2024-01-05-170000_ds_corruption_fix_up/up.sql
```diff
@@ -0,0 +1,96 @@
...
+        GET DIAGNOSTICS cnt = ROW_COUNT;
+        RAISE NOTICE 'DELETEd % rows from %', cnt, schema;
+    END LOOP;
+
+    FOR schema, last_block_from_registry IN SELECT deployment_schema, t.last_block_from_registry 
+                                           FROM temp_sgd_last_block AS t
+                                          WHERE NOT is_ok AND is_upper_limit_too_high
+    LOOP
+        EXECUTE format($$UPDATE %I."data_sources$" 
+                            SET block_range = range_merge(block_range, (%2$s,))
```
lutter, line 89: I _think_ this is setting the `block_range` to `int4range(lower(block_range), null)`, right? Why not just write that here?

lutter (changes requested):

leoyvens:
Thanks for the review, deferring to @avandras on the fix.

store/postgres/migrations/2024-01-05-170000_ds_corruption_fix_up/up.sql
avandras: Exactly.  TBH `information_schema` is something I never use, as it's limited in most places, and otherwise I'm way more familiar with the catalog tables and views.

store/postgres/migrations/2024-01-05-170000_ds_corruption_fix_up/up.sql
avandras: I think it was just left there from the process of building the queries, going to remove it.

store/postgres/migrations/2024-01-05-170000_ds_corruption_fix_up/up.sql
avandras: They do, I just left it like that for expliciteness' sake (to match the explanation above the statement)

store/postgres/migrations/2024-01-05-170000_ds_corruption_fix_up/up.sql
avandras: One can do that as well, indeed.  This must have been some atavism on my side, coming from Python coding, I guess :)

lutter (approved):

PR #5117, opened 2024-01-08
Add more controls on historyBlocks in manifest

incrypto32:


graph/src/env/mod.rs
```diff
@@ -391,6 +401,11 @@ struct Inner {
     enable_gas_metrics: EnvVarBoolean,
     #[envconfig(from = "GRAPH_EXPERIMENTAL_TIMESERIES")]
     enable_timeseries: Option<bool>,
+    #[envconfig(from = "GRAPH_HISTORY_BLOCKS_OVERRIDE")]
+    history_blocks_override: Option<BlockNumber>,
+    // TODO: Find a good default value
```
incrypto32, line 406: We need to decide what a good default value for min is

leoyvens: I'd propose `2*ETHEREUM_REORG_THRESHOLD`, short but safe. @lutter ideas?

graph/src/env/mod.rs
```diff
@@ -391,6 +401,11 @@ struct Inner {
...
+    #[envconfig(from = "GRAPH_MIN_HISTORY_BLOCKS", default = "10000")]
+    min_history_blocks: BlockNumber,
```
leoyvens, line 408: No default here, and an empty value means never override.

graph/src/env/mod.rs
lutter: Yeah, anything bigger than reorg threshold should be safe. Some small multiple sounds good.

graph/src/data/subgraph/mod.rs
```diff
@@ -548,7 +551,83 @@ pub struct BaseSubgraphManifest<C, S, D, T> {
...
+pub enum HistoryBlocks {
+    All,
+    Min,
+    Blocks(BlockNumber),
+}
+
+impl HistoryBlocks {
+    pub fn history_blocks(&self) -> BlockNumber {
+        match self {
+            HistoryBlocks::All => BLOCK_NUMBER_MAX,
```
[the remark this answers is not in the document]
lutter: What's the issue with `BLOCK_NUMBER_MAX` here? That's what's used in other places for 'unlimited history'

graph/src/env/mod.rs
incrypto32: I dont understand> No default for GRAPH_MIN_HISTORY_BLOCKS ?

graph/src/data/subgraph/mod.rs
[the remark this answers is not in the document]
incrypto32: @leoyvens what do you mean by only two variants? All variants of `HistoryBlocks::Enum` needs to be converted to i32 in the end. This `history_blocks()` method is just a helper to do that.

https://github.com/graphprotocol/graph-node/blob/b002d3d73a758d7fb2dbde5d10df22ad4b935bd5/graph/src/data/subgraph/schema.rs#L190-L207

```rust
impl SubgraphManifestEntity {
    pub fn new(
        raw_yaml: String,
        manifest: &super::SubgraphManifest<impl Blockchain>,
        entities_with_causality_region: Vec<EntityType>,
    ) -> Self {
        Self {
            spec_version: manifest.spec_version.to_string(),
            description: manifest.description.clone(),
            repository: manifest.repository.clone(),
            features: manifest.features.iter().map(|f| f.to_string()).collect(),
            schema: manifest.schema.document_string(),
            raw_yaml: Some(raw_yaml),
            entities_with_causality_region,
       -     history_blocks: BLOCK_NUMBER_MAX,   
       +     history_blocks: manifest.history_blocks(),
        }
    }

```

Previously it was BLOCK_NUMBER_MAX or the overrided value from `DeploymentCreate` (set with an RPC param in the `subgraph_deploy` rpc request) that went into the `subgraph_manifest.history_blocks` column in the DB. That column is a non nullable integer which is set to max value of int when there is no historyBlocks specified.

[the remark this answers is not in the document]
leoyvens: I was just afraid of it breaking something in the store somewhere. But if @lutter has confirmed that `BLOCK_NUMBER_MAX` already has the semantics of 'unlimited history' in the store, then we're good and no change is needed.

graph/src/env/mod.rs
leoyvens: Ah I confused this with `history_blocks_override`. Sorry, ignore this comment.

graph/src/env/mod.rs
incrypto32: Made that change to use 2*reorg threshold

leoyvens:
@incrypto32 looking good, one last thing, can we change the `grapman prune` default history blocks to also be `2 * reorg_threshold`? For consistency across ways of pruning.

leoyvens:
Chatting to @Data-Nexus, we don't want the syntax to spook subgraphs if they understand 'historical' as meaning entities that are just old, and that we'd delete their entities just because they are old. An alternative syntax could be:
```yaml
indexerHints:
  prune: [auto|never|<number>]
```  

This way the default
```yaml
indexerHints:
  prune: auto
```
Looks as friendly and reassuring as a piece of YAML can.

schmidsi:
I agree that "prune" is probably the better word. See: https://dictionary.cambridge.org/us/dictionary/english/prune. 

> to reduce something by removing things that are not necessary

node/src/bin/manager.rs
```diff
@@ -311,8 +312,8 @@ pub enum Command {
         #[clap(long, short)]
         delete_threshold: Option<f64>,
         /// How much history to keep in blocks
-        #[clap(long, short = 'y', default_value = "10000")]
-        history: usize,
+        #[clap(long, short = 'y')]
+        history: Option<usize>,
```
leoyvens, line 316: Lets add a CLI doc string here explaining the default value.

leoyvens (approved):

Issue #5131, opened 2024-01-10
[Bug] Disallow grafts that are within the reorg threshold of the base subgraph


leoyvens:
### Bug report

Currently this is allowed, but if the base subgraph reverts it can result in an inconsistent graft. This should be instead rejected at deployment time.

### Relevant log output

```Shell
2024-01-02T14:22:30.164887291Z stderr F Jan 02 14:22:30.164 WARN Firehose selected first streamed block's parent should match subgraph start block, reverting to last know final chain segment, firehose_start_block: #18746373 (5ca99c16915d2cd025e4fc1b3962c08eb75d171840d70116d2ff9204d3a1c5cc), subgraph_current_block: #18746373 (ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff), provider: mainnet-firehose-sf, deployment: QmXYAmabbEmnSDNR5GWJhmm2NQKXnPkSJEBC3EsMwjXr4e, sgd: 567032, subgraph_id: QmXYAmabbEmnSDNR5GWJhmm2NQKXnPkSJEBC3EsMwjXr4e, component: FirehoseBlockStream

2024-01-02T14:22:30.772869846Z stderr F Jan 02 14:22:30.772 INFO Reverting block to get back to main chain, revert_to_ptr: #18746174 (b6fc3f8b01773770b6b4e90c691df73d8baea33b49cbf49b0adffa6936d13367), subgraph_ptr: #18746373 (ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff), sgd: 567032, subgraph_id: QmXYAmabbEmnSDNR5GWJhmm2NQKXnPkSJEBC3EsMwjXr4e, component: SubgraphInstanceManager
```

### Some information to help us out

- [ ] Tick this box if this bug is caused by a regression found in the latest release.
- [ ] Tick this box if this bug is specific to the hosted service.
- [X] I have searched the issue tracker to make sure this issue is not a duplicate.

### OS information

None

PR #5135, opened 2024-01-12
graph: Disallow grafts within the reorg threshold

lutter:
Fixes https://github.com/graphprotocol/graph-node/issues/5131


leoyvens (approved):

lutter:
I fixed the test failure in the nastiest way possible, but can't think of a better way to address it short of rewriting a lot of the subgraph runner tests.

leoyvens:
Truly heinous, I love it

PR #5155, opened 2024-01-23
graphman: Change default index method

### LINKED ISSUE #5130: [Bug] graphman sometimes creates the wrong index
### Bug report

The following graphman command resulted in an unexpected index:
```
graphman index create sgd647572 pool_hour_data pool
Index creation started. Please wait.
Index creation completed.
```
Listing information about the table in the DB showed the following index having been created:
```
=> \d sgd647572.pool_hour_data
                                          Table "sgd647572.pool_hour_data"
         Column          |   Type    | Collation | Nullable |                        Default                        
-------------------------+-----------+-----------+----------+-------------------------------------------------------
 vid                     | bigint    |           | not null | nextval('sgd647572.pool_hour_data_vid_seq'::regclass)
 block_range             | int4range |           | not null | 
 id                      | text      |           | not null | 
 period_start_unix       | integer   |           | not null | 
 pool                    | text      |           | not null | 
 [...]
Indexes:
    "pool_hour_data_pkey" PRIMARY KEY, btree (vid)
    "attr_15_0_pool_hour_data_id" btree (id)
[...]
    "attr_15_9_pool_hour_data_fee_growth_global_1x128" btree (fee_growth_global_1x128)
    "brin_pool_hour_data" brin (lower(block_range) int4_minmax_multi_ops, COALESCE(upper(block_range), 2147483647) int4_minmax_multi_ops, vid int8_minmax_multi_ops)
    "manual_pool_hour_data_pool" btree (pool, block_range)
    "pool_hour_data_block_range_closed" btree (COALESCE(upper(block_range), 2147483647)) WHERE COALESCE(upper(block_range), 2147483647) < 2147483647
    "pool_hour_data_id_block_range_excl" gist (id, block_range)
```
I saw this behaviour earlier on other SGs as well, in one case on a `token` table.

### Relevant log output

_No response_

### IPFS hash

QmZeCuoZeadgHkGwLwMeguyqUKz1WPWQYKcKyMCeQqGhsF

_No response_

- [X] Tick this box if this bug is specific to the hosted service.

### OS information

None

incrypto32:
Closes #5130

### DIFF
--- a/store/postgres/src/deployment_store.rs
+++ b/store/postgres/src/deployment_store.rs
@@ -693,7 +693,7 @@ impl DeploymentStore {
         site: Arc<Site>,
         entity_name: &str,
         field_names: Vec<String>,
-        index_method: Method,
+        index_method: Option<Method>,
         after: Option<BlockNumber>,
     ) -> Result<(), StoreError> {
         let store = self.clone();
@@ -702,6 +702,17 @@ impl DeploymentStore {
             let schema_name = site.namespace.clone();
             let layout = store.layout(conn, site)?;
             let table = resolve_table_name(&layout, &entity_name)?;
+
+            // If no index method is specified, we default to BTree if the
+            // table is not immutable, and GIST otherwise
+            let index_method = index_method.unwrap_or_else(|| {
+                if table.immutable {
+                    Method::BTree
+                } else {
+                    Method::Gist
+                }
+            });
+
             let (column_names, index_exprs) =
                 resolve_column_names_and_index_exprs(table, &field_names)?;
 
--- a/node/src/bin/manager.rs
+++ b/node/src/bin/manager.rs
@@ -674,12 +674,14 @@ pub enum IndexCommand {
         /// case (as its SQL colmun name).
         #[clap(min_values = 1, required = true)]
         fields: Vec<String>,
-        /// The index method. Defaults to `btree`.
+        /// The index method.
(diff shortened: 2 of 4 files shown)


store/postgres/src/deployment_store.rs
```diff
@@ -702,6 +702,17 @@ impl DeploymentStore {
...
+
+            // If no index method is specified, we default to BTree if the
+            // table is not immutable, and GIST otherwise
+            let index_method = index_method.unwrap_or_else(|| {
+                if table.immutable {
+                    Method::BTree
+                } else {
+                    Method::Gist
+                }
+            });
```
lutter, line 714: Wouldn't that also create a GiST index for an index on two numbers? We should only create a GiST index when one of the indexed columns is a range type (i.e., `block_range`)

lutter (changes requested):

store/postgres/src/deployment_store.rs
incrypto32: Should i be reusing this method? Or should i just check if any of the fields is `block_range` and just set Gist if thats the case?

https://github.com/graphprotocol/graph-node/blob/cea3ad277ffd481f0c6b27fa77390876d79345c3/store/postgres/src/relational/ddl.rs#L202-L250

incrypto32:
Closing in favor of #5292

Issue #4973, opened 2023-11-07
[Feature] Allow users to change the gas limit for eth_call


refl3ction:
### Description

The gas limit for `eth_call` is currently hard coded in the code. This is preventing Graph Node to index Polygon Edge networks. I'm running a zero gas L2 chain and Graph Node is causing a lot of errors on the RPC node, because the gas limit for this call on Edge should be zero.

Ref:
- https://github.com/graphprotocol/graph-node/issues/3881
- https://github.com/0xPolygon/polygon-edge/issues/2039
- https://github.com/graphprotocol/graph-node/blob/139f3144d8d878b528416fcd4f9a34353dc3dfb1/chain/ethereum/src/runtime/runtime_adapter.rs#L37

### Some information to help us out

- [ ] Tick this box if you plan on implementing this feature yourself.
- [X] I have searched the issue tracker to make sure this issue is not a duplicate.

PR #5157, opened 2024-01-24
ethereum: Restore GRAPH_ETH_CALL_GAS environment variable

nup9151f:
Solves #4973 
Basically, reverts changes introduced [here](https://github.com/graphprotocol/graph-node/commit/06a6abab734db3b8f91482196e4bb6fb0bbd2c80)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+### Environment Variables
+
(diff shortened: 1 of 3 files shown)


lutter:
The code changes look fine, but making this configurable has repercussions for the network; I think @leoyvens is in a better position to make a call on how to deal with that.

Issue #5137, opened 2024-01-12
[Feature] Allow for changing the block cache shard


trader-payne:
### Description

Currently there's no way of managing the shard location of a block cache once it has been deployed. As in, the cache lives in shard A and I want to move it to shard B. The problem is even if they try to delete it, it asks to get rid of all the subgraphs first.

David suggested the following:
```
In graph-node.toml, change the shard for the block cache (don't restart yet)
In your primary, do something like update chains set name = 'thechain-old' where name = 'thechain'
In the shard where the old block cache lives do update ethereum_networks set name = 'thechain-old' where name = 'thechain'
Restart all graph-node processes
That will make graph-node think that thechain is a new chain and it will set up the cache in the shard you specified in (1)
If you want to get rid of the old block cache, you could then do that with graphman chain remove thechain-old
```

But unfortunately this doesn't work, as per following logs:
```
ERROR:  update or delete on table "chains" violates foreign key constraint "deployment_schemas_network_fkey" on table "deployment_schemas"
DETAIL:  Key (name)=(mainnet) is still referenced from table "deployment_schemas".
```

### Some information to help us out

- [ ] Tick this box if you plan on implementing this feature yourself.
- [X] I have searched the issue tracker to make sure this issue is not a duplicate.

PR #5169, opened 2024-01-29
Add graphman command to change block cache shard

incrypto32:
Closes [#5137](https://github.com/graphprotocol/graph-node/issues/5137)

### DIFF
--- a/node/src/manager/commands/chain.rs
+++ b/node/src/manager/commands/chain.rs
@@ -21,16 +21,9 @@ use graph_store_postgres::BlockStore;
 use graph_store_postgres::ChainStatus;
 use graph_store_postgres::ChainStore;
 use graph_store_postgres::Shard;
-use graph_store_postgres::Storage;
 use graph_store_postgres::{
     command_support::catalog::block_store, connection_pool::ConnectionPool,
 };
-
-#[derive(QueryableByName, Debug)]
-struct ChainIdSeq {
-    #[sql_type = "diesel::sql_types::BigInt"]
-    last_value: i64,
-}
 
 pub async fn list(primary: ConnectionPool, store: Arc<BlockStore>) -> Result<(), Error> {
     let mut chains = {
@@ -185,18 +178,9 @@ pub fn change_block_cache_shard(
         let ident = chain_store.chain_identifier.clone();
         let shard = Shard::new(shard.to_string())?;
 
-        // Fetch the current last_value from the sequence
-        let result = sql_query("SELECT last_value FROM chains_id_seq").get_result::<ChainIdSeq>(&conn)?;
+        let chain = BlockStore::allocate_chain(&conn, &chain_name, &shard, &ident)?;
 
-        let last_val = result.last_value;
-
-        let next_val = last_val + 1;
-        let namespace = format!("chain{}", next_val);
-        let storage= Storage::new(namespace.to_string()).map_err(|e| {
-            anyhow!("Failed to create storage: {}", e)
-        })?;
-
-        store.add_chain_store_inner(&chain_name, &shard, &ident, storage,ChainStatus::Ingestible, true)?;
+        store.add_chain_store(&chain,ChainStatus::Ingestible, true)?;
 
         // Drop the foreign key constraint on deployment_schemas
         sql_query(
@@ -206,7 +190,7 @@ pub fn change_block_cache_shard(
 
         // Update the current chain name to chain-old
         update_chain_name(&conn, &chain_name, &new_name)?;
-        chain_store.update_name(&new_name)?;
+
 
         // Create a new chain with the name in the destination shard
         let _=  add_chain(&conn, &chain_name, &ident, &shard)?;
@@ -218,6 +202,8 @@ pub fn change_block_cache_shard(
          .execute(&conn)?;
         Ok(())
     })?;
+
+    chain_store.update_name(&new_name)?;
 
     println!(
         "Changed block cache shard for {} from {} to {}",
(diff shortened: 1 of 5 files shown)


node/src/manager/commands/chain.rs
```diff
@@ -143,3 +152,55 @@ pub fn remove(primary: ConnectionPool, store: Arc<BlockStore>, name: String) ->
...
+pub fn change_block_cache_shard(
+    primary_store: ConnectionPool,
+    store: Arc<BlockStore>,
+    chain_name: String,
+    shard: String,
+) -> Result<(), Error> {
+    println!("Changing block cache shard for {} to {}", chain_name, shard);
+
+    let conn = primary_store.get()?;
+    let chain = find_chain(&conn, &chain_name)?.expect("no such chain");
```
lutter, line 165: Instead of `expect`, `.ok_or_else(|| anyhow!("unknown chain: {}", name))?;` would produce a friendlier error

node/src/manager/commands/chain.rs
```diff
@@ -143,3 +152,55 @@ pub fn remove(primary: ConnectionPool, store: Arc<BlockStore>, name: String) ->
...
+
+    let chain_store = store
+        .chain_store(&chain_name)
+        .ok_or_else(|| anyhow!("unknown chain: {}", chain_name))?;
+    let new_name = format!("{}-old", &chain_name);
+
+    conn.transaction(|| -> Result<(), StoreError> {
+        let ident = chain_store.chain_identifier.clone();
+        let shard = Shard::new(shard.to_string())?;
+        // Drop the foriegn key constraint on deployment_schemas
```
lutter, line 178: small typo in `foreign`

node/src/manager/commands/chain.rs
```diff
@@ -143,3 +152,55 @@ pub fn remove(primary: ConnectionPool, store: Arc<BlockStore>, name: String) ->
...
+        let chain=  add_chain(&conn, &chain_name, &ident, &shard)?;
+        store.add_chain_store(&chain, ChainStatus::Ingestible, true)?;
+
+        // Re-add the foreign key constraint
+        sql_query(
+            "alter table deployment_schemas add constraint deployment_schemas_network_fkey foreign key (network) references chains(name);",
+        )
+         .execute(&conn)?;
+        Ok(())
+    })?;
```
lutter, line 198: One thing I am wondering about is what happens when this block gets interrupted (e.g., because the user presses ^C) Since there are three shards involved, it may leave everything in a messy state, possibly in a state where `graph-node` won't start again. I think it's right to use one transaction for the primary, at least that way, we won't lose the constraint.

Maybe this order of operations is safer:

1. Add the new chain store to the new shard (no change to primary yet) It's a little sucky since you can't use the `add_chain` helper for that as that modifies the primary
2. Update the `chains` table in the primary (in one txn: update `chains` entry to `-old`, create new entry for chain name pointing to new shard)
3. Update the name in the old shard to `-old`

I am not 100% sure that that leaves everything at least in a state where `graph-node` will start again, but I think it's close.

lutter (changes requested):

node/src/manager/commands/chain.rs
incrypto32: Im not entirely sure of what you suggested, i've made some changes now, could you take a look.

Changed the order, but had to get the next value for the chains_id_seq when doing this before calling `add_chain`

https://github.com/graphprotocol/graph-node/blob/8cded389543e972753baafa258814aa60d7f478e/node/src/manager/commands/chain.rs#L183-L221

lutter: What I am trying to get at is that the `conn.transaction` block puts changes in the primary into a transaction. If that gets interrupted, the transaction in the primary gets rolled back, but changes made in other shards will stick around. My concern then is that we could end up in a state where the primary still points at the chain in the old shard (in `chains`), but we've already updated the chain name in that shard (in `ethereum_networks`), and so when `graph-node` starts up again, it won't find that chain since it first looks at the `chains` table to figure out which shard the block cache is in, but can't find the corresponding entry in `ethereum_networks`.

I _think_ the only change that's needed from what you have right now is to move the call to `chain_store.update_name(&new_name)?;` out of the transaction block. When the txn commits we have an entry for `chain` in the primary that points to the new shard and is correct. We'd also have an entry for `chain-old` that points at the old shard and is incorrect (`ethereum_networks` there still has the entry for `chain`), but that's tolerable since `graph-node` _should_ still start up though I am not 100% confident about that.

Unfortunately, this is really tricky, and testing all possible points where an interrupt (or error, e.g., because the db crashed) could happen is very hard.

lutter (approved):
Please make the change to the transaction block (moving `chain_store.update_name(&new_name)?` to after it) before merging. The other things are more suggestions.

node/src/manager/commands/chain.rs
```diff
@@ -162,20 +169,36 @@ pub fn change_block_cache_shard(
...
+        .ok_or_else(|| anyhow!("unknown chain: {}", &chain_name))?;
     let new_name = format!("{}-old", &chain_name);
 
     conn.transaction(|| -> Result<(), StoreError> {
         let ident = chain_store.chain_identifier.clone();
         let shard = Shard::new(shard.to_string())?;
-        // Drop the foriegn key constraint on deployment_schemas
+
+        // Fetch the current last_value from the sequence
+        let result = sql_query("SELECT last_value FROM chains_id_seq").get_result::<ChainIdSeq>(&conn)?;
```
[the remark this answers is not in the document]
incrypto32: Wouldn't using `next_value` wouldn't that increment the sequence too than not? I manually incremented because i don't want the increment in the sequence myself but just predict the next value.

incrypto32:
@lutter ok to merge this?

lutter (approved):
Yes, I think that's really good now!

PR #5182, opened 2024-01-31
increase blockstream buffer size for substreams

mangas:


graph/src/blockchain/block_stream.rs
```diff
@@ -23,7 +23,7 @@ use crate::{prelude::*, prometheus::labels};
 
 pub const BUFFERED_BLOCK_STREAM_SIZE: usize = 100;
 pub const FIREHOSE_BUFFER_STREAM_SIZE: usize = 1;
-pub const SUBSTREAMS_BUFFER_STREAM_SIZE: usize = 1;
+pub const SUBSTREAMS_BUFFER_STREAM_SIZE: usize = 100;
```
lutter, line 26: Should that be configurable with an environment variable?

lutter (approved):

PR #5183, opened 2024-01-31
store: Fix change_health_column migration for sharded setup

lutter:


store/postgres/migrations/2024-01-25-133200_change_health_column_format/up.sql
```diff
@@ -24,4 +24,20 @@ alter table
...
+  begin
+    for r in select foreign_table_schema as nsp
+               from information_schema.foreign_tables
+              where foreign_table_schema like 'shard_%'
+                and foreign_table_name = 'subgraph_deployment'
+    loop
+      execute 'drop foreign table ' || r.nsp || '.subgraph_deployment';
+    end loop;
+  end
+$$;
```
zorancv, line 41: I think we need to do the same in the down migration too.

zorancv (approved):
looks good to me

store/postgres/migrations/2024-01-25-133200_change_health_column_format/up.sql
lutter: I looked at the down migration, and I think that is fine, since when we run the down migration, the imported tables don't reference anything.

lutter:
On my local test system, I had to manually run `graphman database remap` to rebuild the table imports. It's not clear why that is, but I think it's in one of my shards the migration failed because of a local change I had that shouldn't have been there.

PR #5186, opened 2024-02-02
perf(prune): Only analyze when rebuilding

leoyvens:
When pruning by delete, it's overzealous to analyze. When combined with auto-prune on large tables, I've seen shards struggle with one long analyze after the other.

It can also conflict with the lock for autovacuum and result in a long running transaction waiting for the lock.

### DIFF
--- a/store/postgres/src/relational/prune.rs
+++ b/store/postgres/src/relational/prune.rs
@@ -449,6 +449,10 @@ impl Layout {
                         Ok(())
                     })?;
                     reporter.finish_switch();
+
+                    // Analyze the new tables
+                    let tables = prunable_tables.iter().map(|(table, _)| *table).collect();
+                    self.analyze_tables(conn, reporter, tables, cancel)?;
                 }
                 PruningStrategy::Delete => {
                     // Delete all entity versions whose range was closed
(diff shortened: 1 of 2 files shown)


store/postgres/src/relational/prune.rs
```diff
@@ -495,10 +499,6 @@ impl Layout {
             catalog::set_last_pruned_block(conn, &self.site, &table.name, req.earliest_block)?;
         }
 
-        // Analyze the new tables
-        let tables = prunable_tables.iter().map(|(table, _)| *table).collect();
-        self.analyze_tables(conn, reporter, tables, cancel)?;
-
```
lutter, line 501: With default settings, `Delete` will be used if we think we'll delete between 5% and 50% of the table, where the latter is a lot and statistics from before deletion will likely be junk. It would be better to analyze the tables in the `Delete` branch for which `catalog::needs_autoanalyze` says that an analyze is needed. That would basically cut out waiting for the autoanalyze daemon to get around to analyzing these tables.

leoyvens:
@lutter thanks for reviewing, I've tried another approach on the latest commit, given I've learned about the very convenient `analyze skip_locked`.

lutter (approved):
Nice! I really like this solution.

lutter:
> @lutter thanks for reviewing, I've tried another approach on the latest commit, given I've learned about the very convenient `analyze skip_locked`. This is a Postgres 12 feature, do we document our minimum Postgres? Postgres 11 has hit end-of-life already anyways.

I don't think we have a minimum version documented, but not worrying about EOL versions seems reasonable to me.

PR #5198, opened 2024-02-09
#5085 optimize subgraph synced check

### LINKED ISSUE #5085: Optimize subgraph synced check
Every time we process a block, we check if the subgraph has reached the chain head and can be considered synced. Right now this requires a roundtrip to the DB to check the chain head, here:
https://github.com/graphprotocol/graph-node/blob/fe96516fb6dd8300d9eafd9e40a7eb6eb7ee7395/core/src/subgraph/runner.rs#L753

This overhead is unnecessary, since if a subgraph checks the chain head once and sees that it is very far, there is no need to check again on every block processed.

seem-less:
Closes #5085

### DIFF
--- a/core/src/subgraph/instance_manager.rs
+++ b/core/src/subgraph/instance_manager.rs
@@ -451,12 +451,15 @@ impl<S: SubgraphStore> SubgraphInstanceManager<S> {
             stream: block_stream_metrics,
         };
 
+        let is_deployment_synced = inputs.store.is_deployment_synced().await?;
+
         Ok(SubgraphRunner::new(
             inputs,
             ctx,
             logger.cheap_clone(),
             metrics,
             env_vars,
+            is_deployment_synced,
         ))
     }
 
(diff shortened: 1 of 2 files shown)


leoyvens:
Hey @seem-less, I have to say I'm impressed, this code is subtle and your PR is to the point! Thank you for this contribution. Mind if I ask how you tested this?

And mind if I also ask what has brought you to contribute?

leoyvens:
This code looks good and tests pass so I'm merging.

seem-less:
thanks @leoyvens !
Apologies for the late reply. I only really tested this by using the [example-subgraph](https://github.com/graphprotocol/example-subgraph) and logging the amount of times that the `chain_head_ptr()` method was called before and after my changes. I was considering writing unit/integration tests but instead just pushed to see what the feedback would be.

I contributed because I am exploring opportunities to contribute to web3 infrastructure, and this project seems cool :)

PR #5201, opened 2024-02-10
feat(starknet): trigger filter

xJonathanLEI:
As mentioned in the [original Starknet support PR](https://github.com/graphprotocol/graph-node/pull/4895):

> Trigger filter is not implemented, and `graph-node` would always consume the entire Firehose block stream and build all possible triggers. No benchmark has been done yet, but this should hurt performance pretty bad.

It would no longer be the case with this PR! This PR implements event and block trigger filter for Starknet. The immediate benefit would be that `graph-node` would no longer waste resources paring and matching irrelevant triggers.

However, it would _still_ be the case that an entire, unfiltered Firehose stream would be consumed for each subgraph. We need Firehose filter for that, which builds on what this PR enables. I will submit a follow-up PR for that once we're done with this one. Thanks!

### DIFF
--- a/chain/starknet/src/adapter.rs
+++ b/chain/starknet/src/adapter.rs
@@ -1,3 +1,5 @@
+use std::collections::{hash_map::Entry, HashMap};
+
 use graph::{
     blockchain::{EmptyNodeCapabilities, TriggerFilter as TriggerFilterTrait},
     components::store::BlockNumber,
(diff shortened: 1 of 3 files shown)


xJonathanLEI:
Weird. CI is successful on the fork. Is this the same cache corruption issue as before?

xJonathanLEI:
Just did a rebase on latest `master`.

xJonathanLEI:
Superseded by #5322.

PR #5202, opened 2024-02-12
Support eth_getBalance in mapping

### LINKED ISSUE #895: Support for calling Ether Balances in mappings
Right now there is no way to call an ether balance of an account or a contract. 

This functionality is very useful in protocols that use ether - this includes Uniswap and Compound, and likely also Dharma and Maker. 

This might block me on Uniswap (figuring out if I can work around it right now). With compound we were able to get around it, because they added a getter function that allowed us to read the ether balance of their smart contracts. 

It should return the Eth balance in WEI.

incrypto32:
Closes #895

### DIFF
--- a/chain/ethereum/src/runtime/runtime_adapter.rs
+++ b/chain/ethereum/src/runtime/runtime_adapter.rs
@@ -1,5 +1,6 @@
 use std::{sync::Arc, time::Instant};
 
+use crate::adapter::EthereumGetBalanceError;
 use crate::data_source::MappingABI;
 use crate::{
     capabilities::NodeCapabilities, network::EthereumNetworkAdapters, Chain, DataSource,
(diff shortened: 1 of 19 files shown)


chain/ethereum/src/runtime/runtime_adapter.rs
```diff
@@ -132,6 +147,35 @@ fn ethereum_call(
     }
 }
 
+fn eth_get_balance(
+    eth_adapter: &EthereumAdapter,
+    ctx: HostFnCtx<'_>,
+    wasm_ptr: u32,
+) -> Result<AscPtr<AscBigInt>, HostExportError> {
```
lutter, line 154: Do we need to charge gas for this call? I am not sure about the mechanics/reasoning of gas, but it feels wrong to make this free.

lutter (approved):

lutter:
For the failing unit tests, if you rebase onto latest master, the tests should pass.

lutter:
Oh, one more thing: do we need a new spec or API version because this adds a host fn?

chain/ethereum/src/runtime/runtime_adapter.rs
incrypto32: Right, this shouldn't be gas free. @leoyvens how would we determine a good gas value for `eth_getBalance`?

leoyvens: I suppose we can start with charging the same we charge for `eth_call`.

PR #5210, opened 2024-02-14
graph: Make the ES index we log to configurable

lutter:


graph/src/env/mod.rs
```diff
@@ -326,6 +330,8 @@ struct Inner {
     elastic_search_flush_interval_in_secs: u64,
     #[envconfig(from = "GRAPH_ELASTIC_SEARCH_MAX_RETRIES", default = "5")]
     elastic_search_max_retries: usize,
+    #[envconfig(from = "GRAPH_ELASTIC_SEARCH_INDEX", default = "subgraph-logs")]
```
DaMandal0rian, line 333: can we make the alias the default - `subgraph`

lutter: Just changed that

leoyvens (approved):

PR #5213, opened 2024-02-18
Refactor `graphman rewind` to use pause and resume logic

itsjerryokolo:
This PR refactors the `graphman rewind` command to use the pause and resume function.

### DIFF
--- a/node/src/bin/manager.rs
+++ b/node/src/bin/manager.rs
@@ -1204,9 +1204,8 @@ async fn main() -> anyhow::Result<()> {
             deployments,
             start_block,
         } => {
-            let ct = ctx;
-            let notification_sender = ct.notification_sender();
-            let (store, primary) = ct.store_and_primary();
+            let notification_sender = ctx.notification_sender();
+            let (store, primary) = ctx.store_and_primary();
 
             commands::rewind::run(
                 primary,
--- a/node/src/bin/manager.rs
+++ b/node/src/bin/manager.rs
@@ -923,19 +923,6 @@ impl Context {
         pools
     }
 
-    fn clone(&self) -> Self {
-        Self {
-            logger: self.logger.clone(),
-            node_id: self.node_id.clone(),
-            config: self.config.clone(),
-            ipfs_url: self.ipfs_url.clone(),
-            arweave_url: self.arweave_url.clone(),
-            fork_base: self.fork_base.clone(),
-            registry: self.registry.clone(),
-            prometheus_registry: self.prometheus_registry.clone(),
-        }
-    }
-
     async fn store_builder(&self) -> StoreBuilder {
         StoreBuilder::new(
             &self.logger,
@@ -1194,7 +1181,7 @@ async fn main() -> anyhow::Result<()> {
             let locator = &deployment.locate_unique(&pool)?;
             commands::assign::pause_or_resume(pool, &sender, locator, true)
         }
-        }
+
         Resume { deployment } => {
             let sender = ctx.notification_sender();
             let pool = ctx.primary_pool();
--- a/node/src/bin/manager.rs
+++ b/node/src/bin/manager.rs
@@ -1185,23 +1185,16 @@ async fn main() -> anyhow::Result<()> {
         Resume { deployment } => {
             let sender = ctx.notification_sender();
             let pool = ctx.primary_pool();
-            commands::assign::pause_or_resume(
-                pool.clone(),
-                &sender,
-                &deployment.locate_unique(&pool).unwrap(),
-                false,
-            )
+            let locator = &deployment.locate_unique(&pool).unwrap();
+
+            commands::assign::pause_or_resume(pool, &sender, locator, false)
         }
         Restart { deployment, sleep } => {
-            let ctx_clone = ctx.clone();
             let sender = ctx.notification_sender();
             let pool = ctx.primary_pool();
-            commands::assign::restart(
-                ctx_clone.primary_pool(),
-                &sender,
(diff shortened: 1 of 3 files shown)


node/src/manager/commands/rewind.rs
```diff
@@ -104,30 +107,24 @@ pub async fn run(
...
-                let loc = deployment.locator();
-                let node =
-                    NodeId::new(format!("{}{}", PAUSED, node)).expect("paused_ node id is valid");
-                subgraph_store.reassign_subgraph(&loc, &node)?;
-                println!("  ... paused {}", loc);
-                paused = true;
-            }
-        }
-    }
+        let search = match searches.iter().find(|s| s.locate_unique(&primary).is_ok()) {
```
incrypto32, line 111: The type `DeploymentSearch` can correspond to different deployments. `locate_unique` just pops one deployment that correspond to a search, we want all of them. 

So a cleaner way to do it would be to modify `pause_or_resume` to take a `DeploymentLocator` instead of `DeploymentSearch1 this way we could just only call `pause_or_resume(deployment)` inside this loop,

itsjerryokolo: Okay, I'lll revise.

node/src/manager/commands/rewind.rs
```diff
@@ -104,30 +106,19 @@ pub async fn run(
...
 
-    if paused {
-        // There's no good way to tell that a subgraph has in fact stopped
-        // indexing. We sleep and hope for the best.
-        println!(
-            "\nWaiting {}s to make sure pausing was processed",
-            sleep.as_secs()
-        );
-        thread::sleep(sleep);
+        if paused.is_ok() {
```
incrypto32, line 112: is_ok() doesn't mean the subgraph is paused, it only means the result is the Ok variant. 

We want to check the boolean value of paused, So handle the result from `pause_or_resume` with a `?` and then do 
``` if paused {}```

File: node/src/bin/manager.rs [hunk 3]
```diff
@@ -1165,16 +1179,37 @@ async fn main() -> anyhow::Result<()> {
...
+
             let sender = ctx.notification_sender();
-            commands::assign::pause_or_resume(ctx.primary_pool(), &sender, &deployment, true)
+            let pool = ctx.primary_pool();
+            commands::assign::pause_or_resume(
+                ctx_clone.primary_pool(),
+                &sender,
+                &deployment.locate_unique(&pool).unwrap(),
+                true,
+            )
```
incrypto32, line 1191:
```suggestion
            let pool = ctx.primary_pool();
            let locator = &deployment.locate_unique(&pool)?;
            commands::assign::pause_or_resume(pool, &sender, locator, true)
        }
```

incrypto32: We can avoid cloning ctx

node/src/bin/manager.rs
```diff
@@ -1165,16 +1179,37 @@ async fn main() -> anyhow::Result<()> {
...
             let sender = ctx.notification_sender();
-            commands::assign::restart(ctx.primary_pool(), &sender, &deployment, sleep)
+            let pool = ctx.primary_pool();
+            commands::assign::restart(
+                ctx_clone.primary_pool(),
+                &sender,
+                &deployment.locate_unique(&pool).unwrap(),
+                sleep,
+            )
         }
```
incrypto32, line 1213: This is also needs similar changes like `Pause`

incrypto32 (changes requested):
@itsjerryokolo I just added some comments and suggestions

node/src/bin/manager.rs
```diff
@@ -912,6 +913,19 @@ impl Context {
...
+            node_id: self.node_id.clone(),
+            config: self.config.clone(),
+            ipfs_url: self.ipfs_url.clone(),
+            arweave_url: self.arweave_url.clone(),
+            fork_base: self.fork_base.clone(),
+            registry: self.registry.clone(),
+            prometheus_registry: self.prometheus_registry.clone(),
+        }
+    }
+
```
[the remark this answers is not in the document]
itsjerryokolo: Will remove.

node/src/bin/manager.rs
```diff
@@ -1176,15 +1177,31 @@ async fn main() -> anyhow::Result<()> {
...
+            let pool = ctx.primary_pool();
+            commands::assign::pause_or_resume(
+                pool.clone(),
+                &sender,
+                &deployment.locate_unique(&pool).unwrap(),
+                false,
+            )
         }
         Restart { deployment, sleep } => {
+            let ctx_clone = ctx.clone();
```
incrypto32, line 1196: Clones can be avoided here too

node/src/bin/manager.rs
itsjerryokolo: It's done

node/src/bin/manager.rs
itsjerryokolo: Removed

node/src/manager/commands/rewind.rs
```diff
@@ -104,21 +104,9 @@ pub async fn run(
...
-                paused = true;
-            }
-        }
-    }
+        pause_or_resume(primary.clone(), &sender, &deployment.locator(), true)?;
 
-    if paused {
         // There's no good way to tell that a subgraph has in fact stopped
         // indexing. We sleep and hope for the best.
         println!(
```
incrypto32, line 112: This wait should not be inside the loop

incrypto32 (changes requested):

incrypto32 (approved):

PR #5215, opened 2024-02-19
Track DIPS gas metrics in csv

incrypto32:


graph/src/components/metrics/gas.rs
```diff
@@ -1,64 +1,112 @@
...
+use csv::Writer;
+use slog::{error, info, Logger};
+use std::sync::RwLock;
+use std::{collections::HashMap, sync::Arc};
 
-#[derive(Clone)]
 pub struct GasMetrics {
-    pub gas_counter: CounterVec,
-    pub op_counter: CounterVec,
+    pub gas_counter_map: Arc<RwLock<HashMap<String, u64>>>,
```
incrypto32, line 13: Not sure if we should use something else instead of a `RwLock`

leoyvens:
Lets use [object-store](https://crates.io/crates/object_store), instead of `cloud-storage`, it's support multiple cloud providers and very importantly supports local files.

graph/Cargo.toml
```diff
@@ -71,6 +71,9 @@ itertools = "0.12.0"
 # Without the "arbitrary_precision" feature, we get the error `data did not match any variant of untagged enum Response`.
 web3 = { git = "https://github.com/graphprotocol/rust-web3", branch = "graph-patches-onto-0.18", features = ["arbitrary_precision"] }
 serde_plain = "1.0.2"
+csv = "1.3.0"
+object_store = { version = "0.9.1", features = ["gcp"] }
+cloud-storage = { version = "0.11.1", features = ["global-client"] }
```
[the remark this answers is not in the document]
incrypto32: Noted, Yes right, still a work in progress, have some more local changes to be pushed

leoyvens:
What if we left `GasMetrics` as it was, since those Prometheus metrics may be useful for other purposes. And then added this new collector under the `BlockState`? From a Rust ownership perspective it may be nicer, we may be able to avoid the locks, and have `flush_metrics_to_store` consume `self`, such that `fn reset_counters` wouldn't be necessary.

graph/src/components/subgraph/instance.rs
```diff
@@ -76,6 +88,222 @@ pub struct BlockState {
 
     // Marks whether a handler is currently executing.
     in_handler: bool,
+
+    pub metrics: BlockStateMetrics,
+}
+
+#[derive(Debug)]
+pub struct BlockStateMetrics {
```
leoyvens, line 96: I'd put this in a submodule in a separate file, just to keep things organized. But up to you.

leoyvens (approved):
This code is very elegant now, great work! Just one comment.

leoyvens:
I don't know if one file per block will be ok or not, but we can move forward with this and evaluate after testing a bit in real-life scenarios.

graph/src/components/subgraph/instance.rs
incrypto32: done

PR #5241, opened 2024-02-26
monitoring: Downgrade 'Entity cache statistics' log to trace

leoyvens:
It has interesting info but it logs very frequently and isn't actually read often. Better left outside the default logs.


fordN (approved):
👍

PR #5252, opened 2024-02-28
Make batching conditional on caught-up status

leoyvens:
This takes a different approach to PR https://github.com/graphprotocol/graph-node/pull/5198. Importantly, it will properly disable batching if the subgraph is caught up with the chain, regardless of the 'synced' flag status, fixing a regression.

It separates the related concepts of synced and caught-up, and moves handling that into transact_block_operations`. This nicely allows us to resume batching once caught up.

### DIFF
--- a/graph/src/components/store/traits.rs
+++ b/graph/src/components/store/traits.rs
@@ -338,16 +338,15 @@ pub trait WritableStore: ReadStore + DeploymentCursorTracker {
         deterministic_errors: Vec<SubgraphError>,
         offchain_to_remove: Vec<StoredDynamicDataSource>,
         is_non_fatal_errors_active: bool,
+        is_caught_up_with_chain_head: bool,
     ) -> Result<(), StoreError>;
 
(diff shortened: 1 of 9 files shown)


lutter (approved):
LGTM. I am a little concerned whether we will turn batching on too often, and users will think their subgraph is stuck, though I _think_ it won't be observable by users, but we need to keep an eye out for that.

graph/src/components/store/traits.rs
```diff
@@ -338,16 +338,15 @@ pub trait WritableStore: ReadStore + DeploymentCursorTracker {
         deterministic_errors: Vec<SubgraphError>,
         offchain_to_remove: Vec<StoredDynamicDataSource>,
         is_non_fatal_errors_active: bool,
+        is_caught_up_with_chain_head: bool,
```
lutter, line 341: Maybe worth commenting what behavior this triggers

leoyvens: Documented

PR #5254, opened 2024-02-29
Support Indexed argument filtering

### LINKED ISSUE #2052: Feature Request: Support indexed argument filtering
Feature

Currently TheGraph pulls all logs matching the specified `topic0`. In certain cases, specifically ours, we do not need to process every event, but a subset of these events. 

For example, picture a subgraph which wishes to process a subset of Uniswap `Swap` events which only pertain to their dApp. There are many `Swap` events in almost every block, though not all of these are relevant to the dApp.

We would like to make use of the additional topics to specify the additional `indexed` fields, allowing us to skip over events which do not relate to our dApp. This would significantly save on syncing time and reduce excess processing by nodes.

Currently we are restricted to the following:

```yaml
          eventHandlers:
              - event: Swap(indexed address,uint256,uint256,uint256,uint256,indexed address)
                handler: handleGenericSwap
```

We would like the ability to add additional filtering to the declaration:

**Single exact filter**

```yaml
          eventHandlers:
              - event: Swap(indexed address,uint256,uint256,uint256,uint256,indexed address)
                topic1: "0xabcd...." # filter only events where the sender is the 0xabcd address
                handler: handleSwap
```

**Or filter on the first indexed argument** 
```yaml
          eventHandlers:
              - event: Swap(indexed address,uint256,uint256,uint256,uint256,indexed address)
                topic1: ["0xabcd....", "0x1234"] # filter only events where the sender is the 0xabcd or "0x1234" address
                handler: handleSwap
```

**Or filter on the second indexed argument** 
```yaml
          eventHandlers:
              - event: Swap(indexed address,uint256,uint256,uint256,uint256,indexed address)
                topic2: ["0xabcd....", "0x1234"] # filter only events where the recipient is the 0xabcd or "0x1234" address
                handler: handleSwap
```

I'll leave a reference to how an ethereum node is supposed to respect the `topics` argument in the RPC request.

```
A transaction with a log with topics [A, B] will be matched by the following topic filters:
  [] “anything”
  [A] “A in first position (and anything after)”
  [null, B] “anything in first position AND B in second position (and anything after)”
  [A, B] “A in first position AND B in second position (and anything after)”
  [[A, B], [A, B]] “(A OR B) in first position AND (A OR B) in second position (and anything after)”
```

This change would allow us to sync our subgraph much quicker and reduce the overhead of processing unrelated (to our purposes) events. As an example, it currently takes us a solid 24 hours to process our subgraph and as a result we only store approximately 100,000 entities.

incrypto32:
Closes #2052


incrypto32:
The large diff is the yarn.lock file for the intgration tests

chain/ethereum/src/adapter.rs
```diff
@@ -35,6 +35,53 @@ use crate::{Chain, Mapping, ENV_VARS};
 pub type EventSignature = H256;
 pub type FunctionSelector = [u8; 4];
 
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub struct EventSignatureWithTopics {
```
mangas, line 39: I think this matches in a somewhat unexpected way (for me at least) so i think ti would be worth adding some rust docs here to explain what this is used for and how it works. Maybe add some illustrative examples

chain/ethereum/src/adapter.rs
```diff
@@ -220,6 +306,8 @@ pub struct EthereumLogFilter {
     /// Event sigs with no associated address, matching on all addresses.
     /// Maps to a boolean representing if a trigger requires a transaction receipt.
     wildcard_events: HashMap<EventSignature, bool>,
+    /// Events with any of the topic filters set
+    events_with_topic_filters: HashMap<EventSignatureWithTopics, bool>,
```
mangas, line 310: do you use the bool here for anything? Should this be a hashset?

mangas: seems to be used to mean "requires_receipt", can you add some clarification around it with a comment?

chain/ethereum/src/data_source.rs
```diff
@@ -1173,6 +1174,30 @@ impl MappingEventHandler {
         self.topic0
             .unwrap_or_else(|| string_to_h256(&self.event.replace("indexed ", "")))
     }
+
+    pub fn matches(&self, log: &Log) -> bool {
+        if let Some(topic0) = log.topics.get(0) {
+            return self.topic0() == *topic0
```
mangas, line 1180: I think it would be great if we could turn these repeated conditions into something with a name.

mangas (approved):
See comments but overall LGTM

PR #5256, opened 2024-02-29
graphman: pause subgraph before reassigning

### LINKED ISSUE #5253: graphman reassign should pause the subgraph first
Otherwise it can cause race conditions.

roaminro:
Resolves #5253 

The subgraph being reassigned is now paused if it wasn't and is then resumed once the reassigment process is complete.


lutter:
Sorry this has been sitting for so long. 

I think the change is totally fine, but I am curious if that really addresses the problem in practice - my experience has been that it can take a while for a subgraph to actually process the pause request. Have you seen that? Do you still get a race when that happens?

The basic issue is that the current mechanism for handling assignments doesn't provide a reliable way to say "node x is done with a subgraph so node y can start working on it" which opens the dorr to these races.

PR #5261, opened 2024-03-04
fix(pruning): Fix version stats for huge number of versions

### LINKED ISSUE #5257: [Bug] `graphman stats show` fails with very large tables
### Bug report

With very large tables, `graphman stats show` fails with `integer out of range` because it's converting the estimate of rows [here](https://github.com/graphprotocol/graph-node/blob/master/store/postgres/src/catalog.rs#L736) to an `int4`. 

That query needs to be changed to use an `int8` where ever it explicitly converts `c.reltuples` to an integer.

### Relevant log output

_No response_

### IPFS hash

_No response_

_No response_

### OS information

None

leoyvens:
Resolves #5257

### DIFF
--- a/graph/src/components/store/mod.rs
+++ b/graph/src/components/store/mod.rs
@@ -1054,8 +1054,8 @@ impl ReadStore for EmptyStore {
 /// in a database table
 #[derive(Clone, Debug)]
 pub struct VersionStats {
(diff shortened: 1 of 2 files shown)


lutter (approved):
Nice!

Issue #5262, opened 2024-03-04
[Feature] Declarative eth calls


lutter:
### Description

Subgraphs that use `ethereum.call` a lot can become extremely slow because of these calls; even when a subgraph makes a moderate amount of calls, the overhead of `eth_call` tends to dominate indexing time. Currently, `eth_calls` need to be executed serially, as they are embedded in the WASM code and `graph-node` does not know about these calls until handlers try to execute them.

It should therefore become possible to declare `eth_calls` ahead of time so that `graph-node` can execute them in parallel, and ideally as part of the indexing pipeline so that when handlers run, the results of the calls are known already. To that end, the manifest syntax should be extended. It should be possible to annotate handlers with `eth_calls` with the following syntax:

```yaml
      eventHandlers:
        - event: Swap(indexed address,indexed address,int256,int256,uint160,uint128,int24)
          handler: handleSwap
          calls:
            - global0X128: Pool[event.address].feeGrowthGlobal0X128()
            - global1X128: Pool[event.address].feeGrowthGlobal1X128()
```

With these declarations, the two calls will be executed before the handler is invoked. The declaration `global0X128` has the same effect as this code in the handler

```
let poolContract = PoolABI.bind(event.address)
let feeGrowthGlobal0X128 = poolContract.try_feeGrowthGlobal0X128()
```

Without a call declaration, this causes an actual `eth_call` (or a lookup from the call cache if we made this call before) With the call declaration, the above code will simply get the results from an in-memory cache that was populated before the handler ran.

Declared calls can access the `event.address` of the underlying event as well as all the `event.params`, so that a declared call could also look like
```yaml
calls:
  - ERC20DecimalsToken0: ERC20[event.params.token0].decimals()
```

### Some information to help us out

- [ ] Tick this box if you plan on implementing this feature yourself.
- [X] I have searched the issue tracker to make sure this issue is not a duplicate.

PR #5264, opened 2024-03-05
Declarative calls

lutter:
Addresses #5262 

This PR is concerned with two things:

* refactoring trigger decoding and processing so that we get the decoded triggers all at once _before_ any of them get processed
* making eth calls for a block once before running handlers and plumbing things in a way that eth calls from handlers retrieve the results from memory

The results of `eth_calls` are injected back into the mappings via a `BufferedCallCache` which keeps the call results for the current block in memory. That way, mapping code does not need to change, and users can reap the benefit of declarative calls simply by, well, declaring them in the manifest. There's a small overhead in this approach as it requires calling a host function; compared to making an actual RPC call or just a database lookup, that's negligible.

When we do eth calls in bulk, we do one query against the store for all the calls we need (rather than one per call) For the calls for which we do not have anything in the database, we make RPC calls in parallel. By declaring eth calls, the time for all calls goes from the sum of the time the calls take to the max. Depending on the number of calls a subgraph makes for a block, that can be a considerable speedup. For one subgraph which did 10-20 eth calls per block, the time to process a block went from ~ 1.2s to ~ 200ms, a sixfold speedup. The environment where this was tested took 70-80ms per call.

Done:
- [x] Actually parallelize the calls
- [x] Deduplicate calls
- [x] Cache failed calls in memory to avoid duplicate calls
- [x] Better error messages when an `eth_call` fails

Still todo:
- Charge gas for the calls (#5331)
- Handle overloaded functions (#5329)
- Allow declaring `eth_calls` for other handlers, not just event handlers. (#5330)

This PR is already pretty big; I will file the remaining open items as separate issues before merging this.

### DIFF
--- a/core/src/subgraph/context/mod.rs
+++ b/core/src/subgraph/context/mod.rs
@@ -76,7 +76,7 @@ where
     pub offchain_monitor: OffchainMonitor,
     pub filter: Option<C::TriggerFilter>,
     pub(crate) trigger_processor: Box<dyn TriggerProcessor<C, T>>,
-    pub(crate) decoder: Box<Decoder>,
+    pub(crate) decoder: Box<Decoder<C, T>>,
 }
 
 impl<C: Blockchain, T: RuntimeHostBuilder<C>> IndexingContext<C, T> {
@@ -88,7 +88,7 @@ impl<C: Blockchain, T: RuntimeHostBuilder<C>> IndexingContext<C, T> {
         instances: SubgraphKeepAlive,
         offchain_monitor: OffchainMonitor,
         trigger_processor: Box<dyn TriggerProcessor<C, T>>,
-        decoder: Box<Decoder>,
+        decoder: Box<Decoder<C, T>>,
     ) -> Self {
         let instance = SubgraphInstance::new(
             manifest,
--- a/core/src/subgraph/context/mod.rs
+++ b/core/src/subgraph/context/mod.rs
@@ -10,7 +10,6 @@ use graph::{
     components::{
         store::{DeploymentId, SubgraphFork},
         subgraph::{HostMetrics, MappingError, RuntimeHost as _, SharedProofOfIndexing},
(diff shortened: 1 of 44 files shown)


core/src/subgraph/context/mod.rs
```diff
@@ -99,36 +103,10 @@ impl<C: Blockchain, T: RuntimeHostBuilder<C>> IndexingContext<C, T> {
             offchain_monitor,
             filter: None,
             trigger_processor,
+            decoder,
         }
     }
 
-    pub async fn process_trigger(
```
mangas, line 105: :fire:

mangas (approved):

lutter:
Using a reliable eth provider, this ran cleanly in the integration cluster with no PoI divergences (test ran for 3 days)

kasparkallas:
Hi @lutter

This looks like an useful feature that could increase our subgraph indexing speed a ton. So thanks!

Question, **if a declarative call is defined in the manifest, are these RPC calls always executed eagerly, regardless if the handler actually called it?** The reason I ask this is because some of our handlers have conditional logic to decide whether to execute the RPC call or not.

kasparkallas:
Link to "declared `eth_call`" docs to add more detail to this PR: https://thegraph.com/docs/en/subgraphs/developing/creating/advanced/#declared-eth_call

lutter:
> Question, **if a declarative call is defined in the manifest, are these RPC calls always executed eagerly, regardless if the handler actually called it?**
> 
> The reason I ask this is because some of our handlers have conditional logic to decide whether to execute the RPC call or not. I'm just wondering if I'm now in the position of choosing whether to optimize the RPC calls' performance or the number of RPC calls.

Yes, the declared calls are made no matter what. Whether declared calls in your situation will speed things up or not will depend on how likely it is that a call is made/not made, but since declared calls are all made in parallel, the cost for unnecessary calls (within reason) is very small

PR #5265, opened 2024-03-05
graphql: Fix _meta query failure due to incorrect selection set use

incrypto32:


lutter (approved):
Wow, nice find! This issue has been present for a long time, I wonder why it didn't cause problems before.

PR #5272, opened 2024-03-11
Support ethereum.getCode in mappings

### LINKED ISSUE #2917: Check whether code is present at a given address
Feature

Graph Node is not able to distinguish whether an Ethereum address is a contract, or a EOA

**What is the desired behavior?**
Add a means to check whether code is present at a given address (via `eth_getCode`, docs [here](https://github.com/ethereum/execution-apis/blob/bea0266c42919a2fb3ee524fb91e624a23bc17c5/src/eth/state.json#L92))

If code is present, it is a smart contract

Considerations: this will need to check for code at a particular block, as contracts can be destroyed. `eth_getCode` seems to support block number, not block hash which introduces determinism problems.

Discussed [here](https://forum.thegraph.com/t/a-framework-to-build-any-data-applications-entirely-on-the-decentralized-web3-infrastructures/1979/14)

incrypto32:
Closes #2917


lutter:
One thought: do mappings actually need the code or is it enough for them to know that there is code at that address? If the latter is enough, we should only add a `hasCode`. At some point, we will need to cache these lookups, and with `hasCode`, we only need to cache a boolean rather than the whole code.

chain/ethereum/src/ethereum_adapter.rs
```diff
@@ -414,22 +414,62 @@ impl EthereumAdapter {
...
+
+        let block_id = self.block_ptr_to_id(&block_ptr);
+        let retry_log_message = format!("eth_getCode RPC call for block {}", block_ptr);
+
+        retry(retry_log_message, &logger)
+            .when(|result| match result {
+                Ok(_) => false,
+                Err(_) => true,
+            })
+            .limit(ENV_VARS.request_retries)
```
mangas, line 445: I think we shouldn't be adding more references to ENV_VARS instead of having those as values in structs. This makes it basically impossible to test and we should be removing these patterns instead of adding to them.

tests/tests/integration_tests.rs
```diff
@@ -253,7 +259,7 @@ async fn test_int8(subgraph: Subgraph) -> anyhow::Result<()> {
 * the `cases` variable in `integration_tests`.
 */
 
-async fn test_timestamp(subgraph: Subgraph) -> anyhow::Result<()> {
+async fn test_timestamp(subgraph: Subgraph, _ctx: TestContext) -> anyhow::Result<()> {
```
mangas, line 262: If Testcontext is passed to all of these then it would probably be easier to just pass move the Subgraph argument to the Context struct and then all is needed is the Context

mangas (approved):
See comments

tests/tests/integration_tests.rs
incrypto32: Well noted, just changed

PR #5273, opened 2024-03-11
resolve #5200

### LINKED ISSUE #5200: Allow for parallelized query execution on dedicated/subgraph specific nodes

leoyvens:


lutter (approved):

lutter:
For the `limiter`, we need to think about whether that makes any sense and whether we want that sort of constraint for all db access. I am not sure what the intent behind it is/was.

PR #5276, opened 2024-03-12
graph: Memoize Batch.indirect_weight

lutter:


graph/src/components/store/write.rs
```diff
@@ -772,11 +782,15 @@ impl Batch {
     pub fn groups<'a>(&'a self) -> impl Iterator<Item = &'a RowGroup> {
         self.mods.groups.iter()
     }
+
+    fn weigh(&mut self) {
```
leoyvens, line 786: Got an english lesson with this one.

leoyvens (approved):

PR #5279, opened 2024-03-13
Allow generating a map of section nesting in debug builds

lutter:
With this PR, debug builds (and only debug builds) pay attention to an environment variable `GRAPH_SECTION_MAP` which will make it print a list of the parent/child relations of stopwatch sections. That can then be turned into an image, e.g. with `dot` to visualize the nesting.

sections

### DIFF
--- a/graph/src/components/metrics/stopwatch.rs
+++ b/graph/src/components/metrics/stopwatch.rs
@@ -150,6 +150,8 @@ impl StopwatchInner {
     }
 
     fn start_section(&mut self, id: String) {
+        #[cfg(debug_assertions)]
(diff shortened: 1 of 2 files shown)


graph/src/components/metrics/stopwatch.rs
```diff
@@ -150,6 +150,8 @@ impl StopwatchInner {
     }
 
     fn start_section(&mut self, id: String) {
+        #[cfg(debug_assertions)]
```
zorancv, line 153: We can disable one statement, nice!

zorancv (approved):
Works on my machine.

Issue #5087, opened 2023-12-20
IPFS: Stop doing files/stat calls, use only cat


leoyvens:
Right now we rely on the `files/stat` ipfs route to fetch file sizes before processing them. However in practice we've found that this API tends to error when `cat` still succeeds, and generally the more exotic IPFS API methods aren't super well documented or necessarily stable across Kubo versions. It would be more robust to rely on `cat` as our sole interface to IPFS.

Of course `files/stat` is serving a purpose here of preventing us of trying to download huge files, but we can achieve the same with `cat`, because we can `cat` the files in chunks and stop if we've passed our file size threshold.

PR #5284, opened 2024-03-15
ipfs: Stop using the stat API

leoyvens:
To reduce the total number of calls to IPFS, and to generally simplify our interface on IPFS, this stops using the `stat` API to check existence and control file size. Both can be achieved efficiently with the more usual cat api.

Relying on files/stat was questionable anyways, since we don't know to what extent IPFS clients actually enforce that the declared metadata size value matches the real file size.

Resolves https://github.com/graphprotocol/graph-node/issues/5087.

### DIFF
--- a/graph/src/components/link_resolver/ipfs.rs
+++ b/graph/src/components/link_resolver/ipfs.rs
@@ -2,6 +2,7 @@ use std::sync::{Arc, Mutex};
 use std::time::Duration;
 
 use crate::env::EnvVars;
+use crate::ipfs_client::IpfsError;
 use crate::util::futures::RetryConfigNoTimeout;
 use anyhow::anyhow;
 use async_trait::async_trait;
@@ -13,7 +14,7 @@ use lru_time_cache::LruCache;
 use serde_json::Value;
 
 use crate::{
-    ipfs_client::{IpfsClient, StatApi},
+    ipfs_client::IpfsClient,
     prelude::{LinkResolver as LinkResolverTrait, *},
 };
 
@@ -21,7 +22,7 @@ fn retry_policy<I: Send + Sync>(
     always_retry: bool,
     op: &'static str,
     logger: &Logger,
-) -> RetryConfigNoTimeout<I, crate::prelude::reqwest::Error> {
+) -> RetryConfigNoTimeout<I, IpfsError> {
     // Even if retries were not requested, networking errors are still retried until we either get
     // a valid HTTP response or a timeout.
     if always_retry {
@@ -29,8 +30,9 @@ fn retry_policy<I: Send + Sync>(
     } else {
         retry(op, logger)
             .no_limit()
-            .when(|res: &Result<_, reqwest::Error>| match res {
+            .when(|res: &Result<_, IpfsError>| match res {
                 Ok(_) => false,
+                Err(IpfsError::FileTooLarge(..)) => false,
                 Err(e) => !(e.is_status() || e.is_timeout()),
             })
     }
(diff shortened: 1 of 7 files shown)


graph/src/components/link_resolver/ipfs.rs
```diff
@@ -43,70 +45,51 @@ fn retry_policy<I: Send + Sync>(
...
+                return Ok(clients[index].cheap_clone());
             }
             Err(e) => err = Some(e.into()),
         }
     }
 
     Err(err.unwrap_or_else(|| {
         anyhow!(
-            "No IPFS clients were supplied to handle the call to object.stat. File: {}",
+            "No IPFS clients were supplied to handle the call. File: {}",
```
lutter, line 91: That error message might be a little clearer if it said 'Could not find an IPFS client that has file {path}'

graph/src/components/link_resolver/ipfs.rs
```diff
@@ -246,22 +227,26 @@ impl LinkResolverTrait for IpfsResolver {
...
-            StatApi::Files,
             path.to_string(),
             self.timeout,
             self.retry,
         )
         .await?;
 
         let max_file_size = self.env_vars.mappings.max_ipfs_map_file_size;
-        restrict_file_size(path, size, max_file_size)?;
+        let mut cummulative_file_size = 0;
```
lutter, line 246: typo: `cumulative_file_size`

graph/src/ipfs_client.rs
```diff
@@ -12,6 +14,37 @@ use std::fmt::Display;
...
+
+impl IpfsError {
+    pub fn is_timeout(&self) -> bool {
+        match self {
+            Self::Request(e) => e.is_timeout(),
+            _ => false,
+        }
+    }
+
+    pub fn is_status(&self) -> bool {
```
lutter, line 33: What does `is_status` indicate? Would be good to have a comment on what that means

graph/src/ipfs_client.rs
```diff
@@ -139,41 +145,42 @@ impl IpfsClient {
...
-        let res = self.call(url, None, Some(timeout)).await?;
-        match api {
-            StatApi::Files => Ok(res.json::<FilesStatResponse>().await?.cumulative_size),
-            StatApi::Block => Ok(res.json::<BlockStatResponse>().await?.size),
-        }
+    /// To check the existence of a cid, we do a cat of a single byte.
+    pub async fn exists(&self, cid: &str, timeout: Option<Duration>) -> Result<(), IpfsError> {
+        self.call(self.cat_url("cat", cid, Some(1)), None, timeout)
+            .await?;
+        Ok(())
```
lutter, line 152: I haven't thought through the code changes this requires, but wouldn't it be better to try and get the first block of a file, assuming optimistically that it exists, and only continuing with the client that gets that block first? They could maybe share an `AtomicBool` which gets set by the first one to get the block and the other clients then just stop doing work when the flag becomes `true`

graph/src/ipfs_client.rs
```diff
@@ -139,41 +145,42 @@ impl IpfsClient {
...
+        max_file_size: usize,
+    ) -> Result<Bytes, IpfsError> {
+        let byte_stream = self.cat_stream(cid, timeout).await?;
+        let bytes = byte_stream
+            .err_into()
+            .try_fold(BytesMut::new(), |mut acc, chunk| async move {
+                acc.extend_from_slice(&chunk);
+
+                // Check size limit
+                if acc.len() > max_file_size as usize {
```
lutter, line 168: very minor but isn't `max_file_size` already `usize`?

graph/src/ipfs_client.rs
```diff
@@ -139,41 +145,42 @@ impl IpfsClient {
...
+    }
+    pub async fn cat_stream(
         &self,
         cid: &str,
         timeout: Option<Duration>,
-    ) -> Result<impl Stream<Item = Result<Bytes, reqwest::Error>>, reqwest::Error> {
+    ) -> Result<impl Stream<Item = Result<Bytes, reqwest::Error>> + 'static, reqwest::Error> {
         Ok(self
-            .call(self.url("cat", cid), None, timeout)
+            .call(self.cat_url("cat", cid, None), None, timeout)
```
lutter, line 183: I've looked at the reqwest docs and trawled through the code a bit, but none of it indicates that this will actually read and return the response in chunks. I assume it does, but it's a bit strange that none of the underlying docs guarantee that we don't read everything into memory before getting a response.

lutter (approved):
LGTM! Some minor comments inline

graph/src/ipfs_client.rs
leoyvens: I'm not aware of anyone actually configuring multiple clients so this won't be called much, but getting the first block would be interesting, particularly if the file fits in a single block.

graph/src/ipfs_client.rs
leoyvens: Good catch

graph/src/ipfs_client.rs
leoyvens: Reqwest really isn't that well documented. But I'm assuming `bytes_stream` does what we'd expect and gives us partial data as soon as it is available.

PR #5292, opened 2024-03-21
Refactor manual index creation & add ddl tests

incrypto32:


store/postgres/src/relational/ddl.rs
```diff
@@ -225,31 +225,35 @@ impl Table {
...
+        } else {
+            column.name.quoted()
+        };
 
-            let method = if column.is_list() || column.is_fulltext() {
-                "gin".to_string()
-            } else {
-                "btree".to_string()
-            };
+        let method = if column.is_list() || column.is_fulltext() {
```
avandras, line 250: We abandoned GIN indexes altogether.  Also, having more than one column in an index should not automatically mean non-btree (would be GiST in this case), GiST should be the exception when `block_range` is in the list specified.  The original issue was about `block_range` appearing in the column list despite not having been specified on the `graphman` command line.

lutter: We abandoned GIN for lists/arrays, but we do still need it for fulltext search, right?

avandras: Indeed, @incrypto32 please ignore that part.

incrypto32: Ok, so for the default index im going with this
Screenshot 2024-04-08 at 7 17 35 PM

node/src/bin/manager.rs
```diff
@@ -690,10 +690,10 @@ pub enum IndexCommand {
         fields: Vec<String>,
         /// The index method. Defaults to `btree`.
```
lutter, line 691: That comment also needs to be updated. Something like 
```
/// The index method. Defaults to `btree` in general, and to `gist` when the index includes the `block_range` column
```

node/src/manager/commands/index.rs
```diff
@@ -34,12 +35,23 @@ pub async fn create(
...
+
+    // If the fields contain the block range column, we use GIN
+    // indexes. Otherwise we default to B-tree indexes.
+    let index_method = index_method.unwrap_or_else(|| {
+        if field_names.contains(&BLOCK_RANGE_COLUMN.to_string()) {
+            "gist".to_string()
+        } else {
+            "btree".to_string()
+        }
+    });
```
lutter, line 53: This is really minor, but do we need all the `to_string()` here? (If so, that's ok)

store/postgres/src/deployment_store.rs
```diff
@@ -1880,6 +1855,49 @@ fn resolve_table_name<'a>(layout: &'a Layout, name: &'_ str) -> Result<&'a Table
...
+
+    let mut sql = format!(
+        "create index concurrently if not exists {index_name} \
+         on {schema_name}.{table_name} using {index_method} \
+         ({index_exprs_joined}) ",
+    );
+
+    // If 'after' is provided and the table is not immutable, add a WHERE clause for partial indexing
+    if let Some(after) = after {
+        if !table.immutable {
```
lutter, line 1885: This would be a tad more readable if it was `if table.immutable` and the branches were swapped. My brain has trouble with double negatives

store/postgres/src/relational/ddl_tests.rs
```diff
@@ -60,6 +60,75 @@ fn check_eqv(left: &str, right: &str) {
     }
 }
 
+#[test]
+fn test_manual_index_creation_ddl() {
+    let layout = Arc::new(test_layout(BOOKS_GQL));
+
+    fn assert_generated_sql(
```
lutter, line 67: If you add a `#[track_caller]` before the `fn`, failures will point to the line where this was called from, which can help with fixing failing tests.

store/postgres/src/relational/ddl_tests.rs
```diff
@@ -60,6 +60,75 @@ fn check_eqv(left: &str, right: &str) {
...
+    );
+
+    assert_generated_sql(
+        layout.clone(),
+        "Book",
+        vec!["page_count".to_string(), "title".to_string()],
+        BTREE,
+        "create index concurrently if not exists manual_book_page_count_title on {namespace}.book using btree (\"page_count\", left(\"title\", 256))",
+    );
+}
```
lutter, line 130: Very nice! It would be good to also add a test for an index on `Book.writtenBy` (should be GiST over `(written_by, block_range)`, a test for a explicitly including `block_range` like an index on `vec!["content", "block_range"]` and one for a partial index.

lutter (approved):
Very nice! All my comments are pretty minor; it would be good to add a few more tests as I mentioned, but other than that, it's good to go.

PR #5293, opened 2024-03-21
store: Check interface uniqueness in bulk

lutter:
We used to run one query for each entity that was created or updated if it was of a type that implemented an interface. These single-entity queries can become quite time consuming during syncing.

With this change, we check all the entities of a given type with a single query, so that uniqueness checking for interfaces causes as many queries as there are types (with new/updated entities) rather than as many as there are entities of such types.

### DIFF
--- a/graph/src/data/store/id.rs
+++ b/graph/src/data/store/id.rs
@@ -306,6 +306,16 @@ impl<'a> IdRef<'a> {
     }
 }
 
+impl<'a> From<&'a Id> for IdRef<'a> {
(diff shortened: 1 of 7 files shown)


lutter:
Rebased to latest master and fixed a test failure. The test `postgres::graft::graft` was failing because it used an entity type from the source schema, where `User` implements an interface, but in the destination subgraph, `User` doesn't implement any interfaces, and the test failed because it was checking for a non-existent interface

zorancv (approved):

PR #5294, opened 2024-03-22
Handle null blocks from Lotus

incrypto32:


graph/src/components/store/traits.rs
```diff
@@ -471,14 +471,16 @@ pub trait ChainStore: Send + Sync + 'static {
     ) -> Result<Vec<serde_json::Value>, Error>;
 
     /// Get the `offset`th ancestor of `block_hash`, where offset=0 means the block matching
-    /// `block_hash` and offset=1 means its parent. Returns None if unable to complete due to
-    /// missing blocks in the chain store.
+    /// `block_hash` and offset=1 means its parent. If `root` is passed, short-circuit upon finding
+    /// a child of `root`. Returns None if unable to complete due to missing blocks in the chain
+    /// store.
```
incrypto32, line 476: The comments here should be clearer, we need short circuiting on root because  FEVM can skip blocks and relying on offset can be inaccurate, so we use root as a reference and whenever we find a block whose parent is the root, we know that this is the ancestor we are looking for. The usual value of  root is `subgraph_ptr`

store/postgres/src/chain_store.rs
```diff
@@ -908,38 +908,68 @@ mod data {
...
                             b::table
-                                .filter(b::hash.eq(&hash.hash))
+                                .filter(b::hash.eq(&block.hash))
                                 .select(b::data)
                                 .first::<json::Value>(conn)?,
-                            BlockHash::from_str(&hash.hash)?,
+                            BlockPtr::new(
+                                BlockHash::from_str(&block.hash)?,
+                                i32::try_from(block.number).unwrap(),
+                            ),
```
incrypto32, line 972: @lutter could you take a look at this query generation part, looks good to me, but with the recursive cte's just wanted to double check if everythings alright.

chain/arweave/src/chain.rs
```diff
@@ -1,3 +1,5 @@
+// Portions copyright (2023) Vulcanize, Inc.
+
```
lutter: Can we please remove these copyright notices? That goes for all the copyright notices that this PR introduces.

chain/ethereum/src/adapter.rs
```diff
@@ -963,14 +965,8 @@ pub trait EthereumAdapter: Send + Sync + 'static {
...
-    /// Load block pointer for the specified `block number`.
-    fn block_pointer_from_number(
-        &self,
-        logger: &Logger,
-        block_number: BlockNumber,
-    ) -> Box<dyn Future<Item = BlockPtr, Error = bc::IngestorError> + Send>;
-
-    /// Find a block by its number, according to the Ethereum node.
+    /// Find a block by its number, according to the Ethereum node. If `retries` is passed, limits
+    /// the number of attempts.
```
lutter: The comment mentions retries, but the method doesn't accept any.

chain/ethereum/src/adapter.rs
```diff
@@ -985,6 +981,17 @@ pub trait EthereumAdapter: Send + Sync + 'static {
...
+    /// Finds the hash and number of the lowest non-null block with height greater than or equal to
+    /// the given number.
+    ///
+    /// Note that the same caveats on reorgs apply as for `block_hash_by_block_number`, and must
+    /// also be considered for the resolved block, in case it is higher than the requested number.
+    async fn nearest_block_hash_to_number(
+        &self,
+        logger: &Logger,
+        block_number: BlockNumber,
+    ) -> Result<BlockPtr, Error>;
```
lutter, line 990: I would call this `next_existing_block_ptr` or some such; 'nearest' is ambiguous in whether we are looking for something before or after that number

chain/ethereum/src/chain.rs
```diff
@@ -746,11 +748,12 @@ impl TriggersAdapterTrait<Chain> for TriggersAdapter {
         &self,
         ptr: BlockPtr,
         offset: BlockNumber,
+        root: Option<BlockHash>,
```
lutter, line 749: What does `root` mean here? Should be explained in a comment what that is for

chain/ethereum/src/ethereum_adapter.rs
```diff
@@ -45,6 +45,7 @@ use std::iter::FromIterator;
 use std::pin::Pin;
 use std::sync::Arc;
 use std::time::Instant;
+// use back_to_the_future::futures_await;
```
lutter, line 48: This can probably be deleted.

chain/ethereum/src/ethereum_adapter.rs
```diff
@@ -728,13 +738,12 @@ impl EthereumAdapter {
...
-            .compat()
+        // TODO: This considers null blocks, but we could instead bail if we encounter one as a
+        // small optimization.
+        let canonical_block = self
+            .nearest_block_hash_to_number(logger, block_ptr.number)
             .await?;
-        block_hash
-            .ok_or_else(|| anyhow!("Ethereum node is missing block #{}", block_ptr.number))
-            .map(|block_hash| block_hash == block_ptr.hash_as_h256())
+        Ok(canonical_block == block_ptr)
```
lutter, line 746: This check will always be `true` for chains that are not Lotus, right? And this check requires that we make an RPC call, i.e., introduces overhead for chains that don't need it. I would much prefer an approach where Lotus gets its own EthereumAdapter, something like

```rust
struct LotusAdapter {
  inner: chain::ethereum::EthereumAdapter
}
```

where the impl of trait `EthereumAdapter` just forwards calls to `inner` when that's appropriate and has its own Lotus-specific impl like for this method when needed. That way, existing chains don't pay a price for the additional checks that Lotus requires.

graph/src/components/store/traits.rs
```diff
@@ -474,14 +474,16 @@ pub trait ChainStore: Send + Sync + 'static {
...
```
lutter, line 479: I don't understand what 'short-circuiting' here means. Does that mean that when `root` is passed that the search stops if we either get to the offset-th ancestor of `block_ptr` or find a block whose parent is `root`?

store/postgres/src/chain_store.rs
```diff
@@ -952,38 +954,67 @@ mod data {
...
               from ancestors a, ethereum_blocks b
              where a.block_hash = b.hash
                and a.block_offset < $2
+               {}
         )
-        select a.block_hash as hash
+        select a.block_hash as hash, b.number as number
           from ancestors a
-         where a.block_offset = $2;";
+               inner join ethereum_blocks b on a.block_hash = b.hash
```
lutter: This join is probably not a big deal, but couldn't it be avoided if the recursive CTE also selected the block number, i.e. became `with recursive ancestors (block_hash, block_number, block_offset) ..` ?

lutter (changes requested):
I would like to see changes made to make sure we don't introduce additional overhead for existing chains.

chain/arweave/src/chain.rs
lutter: Somehow github showed me an old version of the file.

graph/src/components/store/traits.rs
incrypto32: Yes that is right

chain/ethereum/src/adapter.rs
incrypto32: I had removed that, i think you were looking at a previous commit

chain/ethereum/src/ethereum_adapter.rs
incrypto32: > This check will always be true for chains that are not Lotus, right? 

I dont understand this, this check can be false on other chains as well i believe, and even before the PR it still needed an RPC call the difference now is that for Lotus it **might** need more than one call, but for other chains it will still need the one call which was present even before this PR.

Am i missing something?

chain/ethereum/src/adapter.rs
incrypto32: done

chain/ethereum/src/ethereum_adapter.rs
incrypto32: done

chain/ethereum/src/chain.rs
incrypto32: Added comments

graph/src/components/store/traits.rs
lutter: I looked at it again, and I think I misread the changes the first time around. You are right, it's not adding an RPC call, just changed how a call is handled. So looks good to me.

One minor nit is that the method name `nearest_block_hash_to_number` sounds like it's translating a hash to a number when it is doing the reverse, maybe calling it `nearest_block_hash_for_number` would be clearer.

lutter (approved):

PR #5296, opened 2024-03-24
Improve query traces

lutter:
Query tracing did not handle multiple block constraints properly.


mangas:
LGTM, only thing that jumped out to me was removing `conn_wait`, not sure if that's used/exposed anywhere. Im also not very familiar with this part of the code so you may want a review from @leoyvens

mangas (approved):

PR #5301, opened 2024-03-26
docs: Update aggregation docs

lutter:
Removed a bunch of more speculative todos - if these are actually needed by users they will let us know ;)

Contrary to what's described in the doc, queries can actually filter on all attributes; I am still not 100% sure if this is a good idea in terms of performance or if we should restrict filtering in the way described in the doc.


azf20 (approved):
Thanks so much!

PR #5305, opened 2024-03-27
Resolve block constraints in bulk

lutter:
The lookup needed for block constraints by hash is now done with one query  instead of one query for each such constraint. None of the other types of  constraints need a query any more.


incrypto32 (approved):
Nice!

PR #5317, opened 2024-04-04
Optimize subgraph ptr update

leoyvens:
And also drastically reduce the number of unfail calls.

### DIFF
--- a/store/postgres/src/deployment.rs
+++ b/store/postgres/src/deployment.rs
@@ -418,7 +418,6 @@ pub fn transact_block(
     firehose_cursor: &FirehoseCursor,
     count: i32,
 ) -> Result<BlockNumber, StoreError> {
-    use crate::diesel::BoolExpressionMethods;
     use subgraph_deployment as d;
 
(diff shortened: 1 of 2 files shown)


store/postgres/src/deployment.rs
```diff
@@ -418,47 +418,49 @@ pub fn transact_block(
...
-            Some(block_ptr_from) if block_ptr_from.number >= ptr.number => Err(
-                StoreError::DuplicateBlockProcessing(site.deployment.clone(), ptr.number),
-            ),
-            None | Some(_) => Err(StoreError::Unknown(anyhow!(
-                "unknown error forwarding block ptr"
-            ))),
-        },
+        // No matching rows were found. This is logically impossible, as the `block_ptr` would have
+        // caught a non-existing deployment.
+        0 => Err(StoreError::Unknown(anyhow!(
```
lutter, line 461: Minor note, but for internal errors of the "can't happen" type, I usually use `StoreError::ConstraintViolation`. Looks good otherwise though.

PR #5321, opened 2024-04-06
Use `Timestamp` type for the `timestamp` attribute of aggregations

lutter:
Timeseries and aggregations now use `Timestamp` as the type of the mandatory `timestamp` field

### DIFF
--- a/graph/src/schema/input/mod.rs
+++ b/graph/src/schema/input/mod.rs
@@ -2626,7 +2626,6 @@ mod validations {
         }
 
         /// Aggregations must have a `timestamp` field of type `Timestamp`
-        /// FIXME: introduce a timestamp type and use that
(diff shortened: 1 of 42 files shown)


graph/src/schema/input/mod.rs
```diff
@@ -2625,7 +2625,7 @@ mod validations {
             errors
         }
 
-        /// Aggregations must have a `timestamp` field of type Int8
+        /// Aggregations must have a `timestamp` field of type `Timestamp`
         /// FIXME: introduce a timestamp type and use that
```
incrypto32, line 2629: We can remove this FIXME

incrypto32 (approved):
LGTM

graph/src/schema/input/mod.rs
lutter: Nice catch. Removed

PR #5322, opened 2024-04-08
feat(starknet): firehose trigger filter

xJonathanLEI:
Supersedes #5201.

This PR adds 2 trigger filters to Starknet:

- block trigger that specifies a list of ranges
- event trigger that's capable of filtering by contract address, event topic, _and_ block ranges

When used without turning on the `filters` feature on the Firehose provider, these filters only work on the `graph-node` side - meaning that full blocks are still always streamed to `graph-node`. Filters are applied _only to avoid excessive trigger parsing_. This part was what #5201 enabled.

What's new in this new PR is the additional implementation of Firhose triggers (i.e. `to_firehose_filter`), allowing `graph-node` to send a `transforms` field in its Firehose stream subscription requests when the `filters` feature is enabled for the provider. Note that all blocks are still always streamed - the filters only affects the list of transactions that are included.

`firehose-starknet` has already [implemented](https://github.com/starknet-graph/firehose-starknet/commit/ec0ae1e3df841635923ff15f99150c8f03fbb950) support for filters. This PR has been tested against that implementation and it works as expected - blocks without matching events have empty transactions, and those with matching ones only come with relevant transactions.

### DIFF
--- a/chain/starknet/src/adapter.rs
+++ b/chain/starknet/src/adapter.rs
@@ -1,3 +1,5 @@
+use std::collections::{hash_map::Entry, HashMap};
+
 use graph::{
     blockchain::{EmptyNodeCapabilities, TriggerFilter as TriggerFilterTrait},
     components::store::BlockNumber,
@@ -5,12 +7,21 @@ use graph::{
 
 use crate::{
     data_source::{DataSource, DataSourceTemplate},
+    felt::Felt,
     Chain,
 };
 
+type TopicWithRanges = HashMap<Felt, Vec<BlockRange>>;
+
 #[derive(Default, Clone)]
 pub struct TriggerFilter {
+    pub(crate) event: StarknetEventFilter,
     pub(crate) block: StarknetBlockFilter,
+}
+
+#[derive(Default, Clone)]
+pub struct StarknetEventFilter {
+    pub contract_addresses: HashMap<Felt, TopicWithRanges>,
 }
 
 #[derive(Default, Clone)]
@@ -28,6 +39,8 @@ impl TriggerFilterTrait<Chain> for TriggerFilter {
     fn extend_with_template(&mut self, _data_source: impl Iterator<Item = DataSourceTemplate>) {}
 
     fn extend<'a>(&mut self, data_sources: impl Iterator<Item = &'a DataSource> + Clone) {
+        self.event
+            .extend(StarknetEventFilter::from_data_sources(data_sources.clone()));
         self.block
             .extend(StarknetBlockFilter::from_data_sources(data_sources));
     }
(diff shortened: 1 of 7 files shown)


chain/starknet/src/adapter.rs
```diff
@@ -1,27 +1,219 @@
...
+const TRANSACTION_EVENT_FILTER_TYPE_URL: &str =
+    "type.googleapis.com/zklend.starknet.transform.v1.TransactionEventFilter";
 
 use crate::{
     data_source::{DataSource, DataSourceTemplate},
+    felt::Felt,
     Chain,
 };
 
+type TopicWithRanges = HashMap<Felt, Vec<BlockRange>>;
```
mangas, line 27: This type is quite opaque in how it's supposed to work, what does Felt represent here? Elsewhere it seems to be an address, here seems to be a "topic". 

Can you please make sure the Felt type comment covers what is represents? 

How many topics do we expect to represent in these? HashMap is probably overkill if it's only handling a handful of values.

chain/starknet/src/adapter.rs
```diff
@@ -1,27 +1,219 @@
...
+
+#[derive(Default, Clone)]
+pub struct TriggerFilter {
+    pub(crate) event: StarknetEventFilter,
+    pub(crate) block: StarknetBlockFilter,
+}
+
+#[derive(Default, Clone)]
+pub struct StarknetEventFilter {
+    pub contract_addresses: HashMap<Felt, TopicWithRanges>,
```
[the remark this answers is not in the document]
mangas: Also somewhat related would be great to have test coverage here

chain/starknet/src/adapter.rs
```diff
@@ -1,27 +1,219 @@
...
+    }
 
     fn node_capabilities(&self) -> EmptyNodeCapabilities<Chain> {
-        todo!()
+        Default::default()
     }
 
     fn to_firehose_filter(self) -> Vec<prost_types::Any> {
-        todo!()
+        // An empty event filter list means that the subgraph is not interested in events at all.
```
mangas, line 66: can you please add some unit tests here

chain/starknet/src/chain.rs
```diff
@@ -391,20 +392,61 @@ impl TriggersAdapterTrait<Chain> for TriggersAdapter {
                 transaction
                     .events
                     .iter()
-                    .map(|event| {
-                        StarknetTrigger::Event(StarknetEventTrigger {
-                            event: Arc::new(event.clone()),
-                            block: shared_block.clone(),
-                            transaction: transaction.clone(),
-                        })
+                    .filter_map(|event| {
```
mangas, line 395: As mentioned in a different comment, would be great if you could move the matching logic into the filter itself leave this implementation focused on extract the triggers from the block rather than dealing with internals from the filter

mangas: Would also be great if we can have some tests, near has some examples you can follow.

mangas (changes requested):
See other comments

chain/starknet/src/adapter.rs
xJonathanLEI: Oh sorry for the confusion. `Felt` is the universal primitive type in Starknet. Everything is a `Felt`: from addresses to contract call params to event logs, literally everything. I will make sure to cover this fact in the `Felt` type comment itself.

As for "topic", probably bad naming but what it means is the first element in any event `keys`, similar to Ethereum event signatures (which are the first element of any log "topics"). I should probably name this `signature` instead?

mangas: maybe something like EventPrefix or EventFirstKey something along those lines, signature could make sense too if that's how it's treated, I think adding rust comments to the types will make that clearer regardless of the name :) I'm terrible at naming things eheh

xJonathanLEI:
Pushed changes:

- encapsulated trigger filter logic
- renamed a few type aliases to reduce ambiguity
- added trigger filter matching tests

Notably, since the trigger filter type is now completely encapsulated, it's now rather difficult to construct a `TriggerFilter` instance from `chain.rs`. So I just tested everything inside `adapter.rs` instead.

mangas (approved):

mangas:
As I understand it this integration was taken in a different direction, I'll close the PR for now, re-open if needed. Also if not needed anymore would be great to remove the existing stakrnet code.

PR #5323, opened 2024-04-08
Upgrade hyper to version 1.2

lutter:
This PR updates hyper to the 1.0 API. As a necessity, it also further decouples ourselves from diesel's version of `BigDecimal` through some ugly copy&paste, but these changes ensure that we are no longer dependent on other crates using our preferred version of `BigDecimal`.

### DIFF
--- a/graph/Cargo.toml
+++ b/graph/Cargo.toml
@@ -8,10 +8,9 @@ anyhow = "1.0"
 async-trait = "0.1.74"
 async-stream = "0.3"
 atomic_refcell = "0.1.13"
-# Make sure no newer version of bigdecmal library is used in diesel dependencies reflected in the Cargo.lock file.
-# Automatic reneratiion of the Cargo.lock file would break compilation as only diesel dependencies would be updated.
-# Using a higher version of bigdecimal in complete graph-node project would change the PoI, hence break the consensus.
-# BigDecimals should only by used through graph::store::scalar and not directly from this crate
+# We require this precise version of bigdecimal. Updating to later versions
+# has caused PoI differences; if you update this version, you will need to
+# make sure that it does not cause PoI changes
 old_bigdecimal = { version = "=0.1.2", features = ["serde"], package = "bigdecimal" }
 bytes = "1.0.1"
 cid = "0.11.0"


graph/Cargo.toml
```diff
@@ -24,21 +25,25 @@ reqwest = { version = "0.11.18", features = ["json", "stream", "multipart"] }
...
 num-traits = "=0.2.17"
 rand = "0.8.4"
 regex = "1.5.4"
 semver = { version = "1.0.21", features = ["serde"] }
-serde = { version = "1.0.126", features = ["rc"] }
-serde_derive = "1.0.125"
-serde_json = { version = "1.0", features = ["arbitrary_precision"] }
-serde_regex = "1.1.0"
-serde_yaml = "0.9.21"
+serde = { workspace = true }
```
mangas, line 42: unless we're using this longer form for a reason I think `serde.workspace = true` is a lot nicer.

mangas (approved):
Very nice improvements :fire:

graph/src/data/store/scalar/bigint.rs
```diff
@@ -0,0 +1,391 @@
...
+    impl Display for BigInt {
+        fn fmt(&self, f: &mut Formatter) -> Result<(), fmt::Error> {
+            self.0.fmt(f)
+        }
+    }
+
+    impl BigInt {
+        // Postgres `numeric` has a limit documented here [https://www.postgresql.org/docs/current/datatype-numeric.htm]:
+        // "Up to 131072 digits before the decimal point; up to 16383 digits after the decimal point"
+        // So based on this we adopt a limit of 131072 decimal digits for big int, converted here to bits.
```
zorancv, line 37: Maybe 16383 is more suitable limit here, not that any of those is likely to be reached anyway.

graph/Cargo.toml
```diff
@@ -11,7 +11,8 @@ atomic_refcell = "0.1.13"
 # Make sure no newer version of bigdecmal library is used in diesel dependencies reflected in the Cargo.lock file.
 # Automatic reneratiion of the Cargo.lock file would break compilation as only diesel dependencies would be updated.
 # Using a higher version of bigdecimal in complete graph-node project would change the PoI, hence break the consensus.
```
zorancv, line 13: I guess we can remove the comment in the three lines above, since this PR also solves exactly that technical debt.

graph/Cargo.toml
lutter: I personally like the `{ workspace = true }` notation better because it's more in line with the notation for setting features etc. but either is fine by me

graph/Cargo.toml
lutter: I rewrote that comment

graph/src/data/store/scalar/bigint.rs
lutter: I'll leave that as a fix for another time ...

lutter:
I screwed up when I merged this - it is actually merged, not just closed

graph/Cargo.toml
zorancv: That's even better!

Issue #3937, opened 2022-09-14
Make `graph-node` tolerate chains not being available during startup


lutter:
Right now, when `graph-node` starts, it does some checks against each RPC endpoint; if those checks fail or time out, `graph-node` will mark the chain as not working and not use it anymore. That will also make all subgraphs that use that chain fail at startup. If the problem with the endpoint is transient, the only way to get `graph-node` to use it again is to restart it (at the danger that now some other chain has a transient issue)

The code needs to be changed such that `graph-node` is much more tolerant to such transient issues and automatically retries using a chain that caused trouble during startup. As part of solving the issue, we should also produce documentation that describes what is expected of an endpoint before we will use it, and a `graphman` command that allows checking any given endpoint by going through its startup sequence. Additionally, there needs to be some way to figure out which of the configured endpoints `graph-node` considers usable/not usable at any given point in time.

matthewdarwin:
Sounds like a great improvement.

Please also expose this knowledge via Prometheus monitoring wherever possible/reasonable.

paymog:
I just filed #4115 which seems possibly related to this. Does graph-node mark a chain as not working if a firehose provider goes down after successful startup?

leoyvens:
#4754 should help with firehose chains, by allowing them to retry for 30 secs before giving up.

azf20:
@paymog I think it will help, but this is more targeted https://github.com/graphprotocol/graph-node/pull/4778

PR #5337, opened 2024-04-11
Remove provider checks at startup

mangas:
Fixes https://github.com/graphprotocol/graph-node/issues/3937

- New chain now created with `ChainIdentifier::default()` on start
- Adapter genesis+net_version now checked by `ProviderManager`
- Adapter checks will be disabled temporarily after an intermittent failure
- Adapter won't be returned by the `ProviderManager` if the ident returned is not the expected value.
- BlockIngestor now takes a `ChainClient` and tries to get an adapter with each poll, this means it won't stop if the first poll fails
- node/src/main.rs refactored so that most of the init code is automatically executed by the new `Networks` type.
- graphman run now uses the same init code as the node, should be able to run using any chain
- EthereumNetworks type removed
- FirehoseNetworks type removed 
- New chains require a working adapter for creation

Future improvements:
- Cache genesis once it's set
- Cache chain ident once it's set

### DIFF
--- a/chain/ethereum/src/ethereum_adapter.rs
+++ b/chain/ethereum/src/ethereum_adapter.rs
@@ -1905,7 +1905,8 @@ pub(crate) async fn get_calls(
             } else {
                 client
                     .rpc()?
-                    .cheapest_with(capabilities)?
+                    .cheapest_with(capabilities)
+                    .await?
                     .calls_in_block(
                         &logger,
                         subgraph_metrics.clone(),
--- a/chain/ethereum/src/ingestor.rs
+++ b/chain/ethereum/src/ingestor.rs
@@ -1,5 +1,11 @@
-use crate::{chain::BlockFinality, EthereumAdapter, EthereumAdapterTrait, ENV_VARS};
-use graph::futures03::compat::Future01CompatExt;
+use crate::{chain::BlockFinality, ENV_VARS};
+use crate::{EthereumAdapter, EthereumAdapterTrait as _};
+use graph::blockchain::client::ChainClient;
+use graph::blockchain::BlockchainKind;
+use graph::components::adapter::ChainId;
+use graph::futures03::compat::Future01CompatExt as _;
+use graph::slog::o;
+use graph::util::backoff::ExponentialBackoff;
 use graph::{
     blockchain::{BlockHash, BlockIngestor, BlockPtr, IngestorError},
     cheap_clone::CheapClone,
@@ -13,25 +19,25 @@ use std::{sync::Arc, time::Duration};
 pub struct PollingBlockIngestor {
     logger: Logger,
     ancestor_count: i32,
-    eth_adapter: Arc<EthereumAdapter>,
+    chain_client: Arc<ChainClient<crate::chain::Chain>>,
     chain_store: Arc<dyn ChainStore>,
     polling_interval: Duration,
-    network_name: String,
+    network_name: ChainId,
 }
 
 impl PollingBlockIngestor {
     pub fn new(
         logger: Logger,
         ancestor_count: i32,
-        eth_adapter: Arc<EthereumAdapter>,
+        chain_client: Arc<ChainClient<crate::chain::Chain>>,
         chain_store: Arc<dyn ChainStore>,
         polling_interval: Duration,
-        network_name: String,
+        network_name: ChainId,
     ) -> Result<PollingBlockIngestor, Error> {
         Ok(PollingBlockIngestor {
             logger,
             ancestor_count,
-            eth_adapter,
+            chain_client,
             chain_store,
             polling_interval,
             network_name,
@@ -59,8 +65,12 @@ impl PollingBlockIngestor {
         }
     }
 
-    async fn do_poll(&self) -> Result<(), IngestorError> {
-        trace!(self.logger, "BlockIngestor::do_poll");
+    async fn do_poll(
+        &self,
+        logger: &Logger,
+        eth_adapter: Arc<EthereumAdapter>,
+    ) -> Result<(), IngestorError> {
+        trace!(&logger, "BlockIngestor::do_poll");
 
         // Get chain head ptr from store
         let head_block_ptr_opt = self.chain_store.cheap_clone().chain_head_ptr().await?;
@@ -68,7 +78,7 @@ impl PollingBlockIngestor {
         // To check if there is a new block or not, fetch only the block header since that's cheaper
         // than the full block. This is worthwhile because most of the time there won't be a new
         // block, as we expect the poll interval to be much shorter than the block time.
-        let latest_block = self.latest_block().await?;
+        let latest_block = self.latest_block(logger, &eth_adapter).await?;
 
         if let Some(head_block) = head_block_ptr_opt.as_ref() {
             // If latest block matches head block in store, nothing needs to be done
@@ -80,7 +90,7 @@ impl PollingBlockIngestor {
                 // An ingestor might wait or move forward, but it never
                 // wavers and goes back. More seriously, this keeps us from
                 // later trying to ingest a block with the same number again
-                warn!(self.logger,
+                warn!(&logger,
                     "Provider went backwards - ignoring this latest block";
                     "current_block_head" => head_block.number,
                     "latest_block_head" => latest_block.number);
@@ -92,7 +102,7 @@ impl PollingBlockIngestor {
         match head_block_ptr_opt {
             None => {
                 info!(
-                    self.logger,
+                    &logger,
                     "Downloading latest blocks from Ethereum, this may take a few minutes..."
                 );
             }
@@ -108,7 +118,7 @@ impl PollingBlockIngestor {
                 };
                 if distance > 0 {
                     info!(
-                        self.logger,
+                    &logger,
                         "Syncing {} blocks from Ethereum",
                         blocks_needed;
                         "current_block_head" => head_number,
@@ -125,7 +135,9 @@ impl PollingBlockIngestor {
         // Might be a no-op if latest block is one that we have seen.
         // ingest_blocks will return a (potentially incomplete) list of blocks that are
         // missing.
-        let mut missing_block_hash = self.ingest_block(&latest_block.hash).await?;
+        let mut missing_block_hash = self
+            .ingest_block(&logger, &eth_adapter, &latest_block.hash)
+            .await?;
 
         // Repeatedly fetch missing parent blocks, and ingest them.
         // ingest_blocks will continue to tell us about more missing parent
@@ -146,29 +158,27 @@ impl PollingBlockIngestor {
         //   iteration will have at most block number N-1.
         // - Therefore, the loop will iterate at most ancestor_count times.
         while let Some(hash) = missing_block_hash {
-            missing_block_hash = self.ingest_block(&hash).await?;
+            missing_block_hash = self.ingest_block(&logger, &eth_adapter, &hash).await?;
         }
         Ok(())
     }
 
     async fn ingest_block(
         &self,
+        logger: &Logger,
+        eth_adapter: &Arc<EthereumAdapter>,
         block_hash: &BlockHash,
     ) -> Result<Option<BlockHash>, IngestorError> {
         // TODO: H256::from_slice can panic
         let block_hash = H256::from_slice(block_hash.as_slice());
 
         // Get the fully populated block
-        let block = self
-            .eth_adapter
-            .block_by_hash(&self.logger, block_hash)
+        let block = eth_adapter
+            .block_by_hash(logger, block_hash)
             .compat()
             .await?
             .ok_or(IngestorError::BlockUnavailable(block_hash))?;
-        let ethereum_block = self
-            .eth_adapter
-            .load_full_block(&self.logger, block)
-            .await?;
+        let ethereum_block = eth_adapter.load_full_block(&logger, block).await?;
 
         // We need something that implements `Block` to store the block; the
         // store does not care whether the block is final or not
@@ -188,31 +198,62 @@ impl PollingBlockIngestor {
             .await
             .map(|missing| missing.map(|h256| h256.into()))
             .map_err(|e| {
-                error!(self.logger, "failed to update chain head");
+                error!(logger, "failed to update chain head");
                 IngestorError::Unknown(e)
             })
     }
 
-    async fn latest_block(&self) -> Result<BlockPtr, IngestorError> {
-        self.eth_adapter
-            .latest_block_header(&self.logger)
+    async fn latest_block(
+        &self,
+        logger: &Logger,
+        eth_adapter: &Arc<EthereumAdapter>,
+    ) -> Result<BlockPtr, IngestorError> {
+        eth_adapter
+            .latest_block_header(&logger)
             .compat()
             .await
             .map(|block| block.into())
+    }
+
+    async fn eth_adapter(&self) -> anyhow::Result<Arc<EthereumAdapter>> {
+        self.chain_client
+            .rpc()?
+            .cheapest()
+            .await
+            .ok_or_else(|| graph::anyhow::anyhow!("unable to get eth adapter"))
     }
 }
 
 #[async_trait]
 impl BlockIngestor for PollingBlockIngestor {
     async fn run(self: Box<Self>) {
+        let mut backoff =
+            ExponentialBackoff::new(Duration::from_millis(250), Duration::from_secs(30));
+
         loop {
-            match self.do_poll().await {
-                // Some polls will fail due to transient issues
+            let eth_adapter = match self.eth_adapter().await {
+                Ok(adapter) => {
+                    backoff.reset();
+                    adapter
+                }
                 Err(err) => {
                     error!(
-                        self.logger,
-                        "Trying again after block polling failed: {}", err
+                        &self.logger,
+                        "unable to get ethereum adapter, backing off... error: {}",
+                        err.to_string()
                     );
+                    backoff.sleep_async().await;
+                    continue;
+                }
+            };
+            let logger = self
+                .logger
+                .new(o!("provider" => eth_adapter.provider().to_string()));
+
+            match self.do_poll(&logger, eth_adapter).await {
+                // Some polls will fail due to transient issues
+                Err(err) => {
+                    error!(logger, "Trying again after block polling failed: {}", err);
                 }
                 Ok(()) => (),
             }
@@ -225,7 +266,11 @@ impl BlockIngestor for PollingBlockIngestor {
         }
     }
 
-    fn network_name(&self) -> String {
+    fn network_name(&self) -> ChainId {
         self.network_name.clone()
+    }
+
+    fn kind(&self) -> BlockchainKind {
+        BlockchainKind::Ethereum
     }
 }
--- a/chain/ethereum/src/network.rs
+++ b/chain/ethereum/src/network.rs
@@ -1,15 +1,14 @@
 use anyhow::{anyhow, bail};
-use graph::cheap_clone::CheapClone;
+use graph::blockchain::ChainIdentifier;
+use graph::components::adapter::{ChainId, NetIdentifiable, ProviderManager, ProviderName};
 use graph::endpoint::EndpointMetrics;
 use graph::firehose::{AvailableCapacity, SubgraphLimit};
 use graph::prelude::rand::seq::IteratorRandom;
 use graph::prelude::rand::{self, Rng};
-use std::cmp::Ordering;
-use std::collections::HashMap;
 use std::sync::Arc;
 
 pub use graph::impl_slog_value;
-use graph::prelude::Error;
+use graph::prelude::{async_trait, Error};
 
 use crate::adapter::EthereumAdapter as _;
 use crate::capabilities::NodeCapabilities;
@@ -29,7 +28,32 @@ pub struct EthereumNetworkAdapter {
     limit: SubgraphLimit,
 }
 
+#[async_trait]
+impl NetIdentifiable for EthereumNetworkAdapter {
+    async fn net_identifiers(&self) -> Result<ChainIdentifier, anyhow::Error> {
+        self.adapter.net_identifiers().await
+    }
+    fn provider_name(&self) -> ProviderName {
+        self.adapter.provider().into()
+    }
+}
+
 impl EthereumNetworkAdapter {
+    pub fn new(
+        endpoint_metrics: Arc<EndpointMetrics>,
+        capabilities: NodeCapabilities,
+        adapter: Arc<EthereumAdapter>,
+        limit: SubgraphLimit,
+    ) -> Self {
+        Self {
+            endpoint_metrics,
+            capabilities,
+            adapter,
+            limit,
+        }
+    }
+
+    #[cfg(debug_assertions)]
     fn is_call_only(&self) -> bool {
         self.adapter.is_call_only()
     }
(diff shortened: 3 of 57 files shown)


graph/src/blockchain/mod.rs
```diff
@@ -510,18 +512,36 @@ impl BlockchainKind {
 
 /// A collection of blockchains, keyed by `BlockchainKind` and network.
 #[derive(Default, Debug, Clone)]
-pub struct BlockchainMap(HashMap<(BlockchainKind, String), Arc<dyn Any + Send + Sync>>);
+pub struct BlockchainMap(pub HashMap<(BlockchainKind, ChainId), Arc<dyn Any + Send + Sync>>);
```
lutter, line 515: Do we need to make this `pub`? Could we get by with adding some accessor method to `BlockchainMap` instead?

graph/src/components/adapter.rs
```diff
@@ -0,0 +1,762 @@
...
+    blockchain::{BlockHash, ChainIdentifier},
+    cheap_clone::CheapClone,
+    data::value::Word,
+    prelude::error,
+    tokio::sync::RwLock,
+};
+
+use crate::components::store::{BlockStore as BlockStoreTrait, ChainStore as ChainStoreTrait};
+
+const VALIDATION_ATTEMPT_TTL_SECONDS: i64 = 60 * 5;
```
lutter, line 24: Would be better to make that a `Duration`. That'll also save you the `_SECONDS` postfix ;)

graph/src/components/adapter.rs
```diff
@@ -0,0 +1,762 @@
...
+        &self,
+        chain_id: &ChainId,
+        ident: &ChainIdentifier,
+    ) -> Result<(), IdentValidatorError> {
+        let network_chain = self
+            .chain_store(&chain_id)
+            .ok_or_else(|| IdentValidatorError::UnavailableStore(chain_id.clone()))?;
+        let store_ident = network_chain.chain_identifier();
+
+        if store_ident == &ChainIdentifier::default() {
```
lutter: This should be something like `store_ident.is_unset()`, and would ideally be implemented by making `ChainIdentifier` an enum that has an `Unset` variant; or even better make `ChainStore.chain_identifier` return a `Result<Option<ChainIdentifier>, Error>`. Relying on the convention that the default means 'not really there' is very brittle.

node/src/config.rs
```diff
@@ -100,7 +122,335 @@ fn validate_name(s: &str) -> Result<()> {
...
+
+impl Networks {
+    // noop is important for query_nodes as it shortcuts a lot of the process.
+    fn noop() -> Self {
+        Self {
+            adapters: vec![],
+            rpc_provider_manager: ProviderManager::new(
+                Logger::root(Discard, o!()),
+                vec![].into_iter(),
+                Arc::new(MockIdentValidator),
```
lutter, line 201: Can't this be the real validator? If there's no endpoints, the validator won't have to do anything. I don't like having all these traits floating around where whether they do anything or not depends on setup and you're never quite sure if the setup did things correctly.

node/src/config.rs
```diff
@@ -100,7 +122,335 @@ fn validate_name(s: &str) -> Result<()> {
...
+            chain_id,
+            self.rpc_provider_manager.cheap_clone(),
+            eth_adapters,
+            None,
+        )
+    }
+}
+
 impl Config {
+    pub async fn networks(
```
lutter, line 417: This module was really only meant for handling the configuration file and any checks that can be performed on the config file without any other setup. It feels like some of the headaches this needs to address is that, because `Config` has very little setup around it, the code needs to work both for when nothing has been set up and for when we actually have things like a store available.

It might be better to restructure this so that all this network setup code gets run from, say, a `NetworkSetup` struct that gets a config and whatever else is needed (like the `BlockStore`) and can then do its thing knowing that the store exists already.

node/src/main.rs
```diff
@@ -1,58 +1,36 @@
 use clap::Parser as _;
-use ethereum::chain::{
-    EthereumAdapterSelector, EthereumBlockRefetcher, EthereumRuntimeAdapterBuilder,
-    EthereumStreamBuilder,
-};
-use ethereum::{BlockIngestor, EthereumNetworks};
```
lutter, line 6: Nice! I like that this tightens up `main` overall - it had way too much control flow mixed with intricate but somewhat boring detail.

lutter (commented):
Not a full review (it's Friday evening and I am running out of time) but I wanted to provide some feedback for the PR. It feels like this could be simplified quite a bit by restructuring how it works and changing code so that it can rely on basic assumptions like 'the store is there and we can talk to it'

chain/ethereum/src/network.rs
```diff
@@ -123,20 +185,23 @@ impl EthereumNetworkAdapters {
...
     }
 
-    pub fn remove(&mut self, provider: &str) {
-        self.adapters
-            .retain(|adapter| adapter.adapter.provider() != provider);
-    }
+    // pub fn remove(&mut self, provider: &str) {
+    //     self.manager.get_all_sync(&self.chain_id).unwrap_or_default()
+    //         .retain(|adapter| adapter.adapter.provider() != provider);
+    // }
```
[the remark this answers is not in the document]
mangas: Was left over from another layer around endpoints that was remove :stuck_out_tongue:

chain/ethereum/src/network.rs
```diff
@@ -48,57 +71,96 @@ impl EthereumNetworkAdapter {
...
     fn default() -> Self {
-        Self::new(None)
+        Self {
+            chain_id: "".into(),
+            manager: ProviderManager::default(),
+            call_only_adapters: vec![],
+            retest_percent: DEFAULT_ADAPTER_ERROR_RETEST_PERCENT,
+        }
     }
 }
```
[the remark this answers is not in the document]
mangas: I usually add these types of defaults when the setup is not super straightforward so that anyone using it doesn't have to dive into how the struct works to use. The main issue with ChainClient is that in some cases we do need to use both, I could probably add a third option that explicitly uses firehose AND rpc to remove that bound

graph/src/blockchain/mod.rs
mangas: added an iter accessor

graph/src/components/adapter.rs
mangas: image

graph/src/components/adapter.rs
mangas: The reason I went with this is basically because the store does not have a way of expressing "unset", it won't be null, since the field is not nullable. This means that to return the Option we'd just hide this assumption further down in the store. The main issue with that IMO is that the store should not know anything about set/unset unless that can be expressed in the data model. 

While this is not strictly true because the store sets the default (since it isn't passed one) I feel like the better place fro all these decisions is in the IdentValidator which would be responsible for all of this. I agree relying on default values is a bit brittle but in this case is a complexity trade-off to avoid a migration in a somewhat critical place.

node/src/config.rs
mangas: This is basically what Networks is, I could move it to it's own module and add a from_config instead of having config.networks(), this was just for convenience, I don't feel strongly either way.

mangas:
> Not a full review (it's Friday evening and I am running out of time) but I wanted to provide some feedback for the PR. It feels like this could be simplified quite a bit by restructuring how it works and changing code so that it can rely on basic assumptions like 'the store is there and we can talk to it'
> 
> To elaborate a little more what I mean: my basic intuition is that whatever structs we use today to represent endpoints now become enums that roughly look like
> 
> ```rust
> enum Endpoint {
>     NotReady(details)
>    Ready(details)
>    // .. maybe other states to help with bookkeeping
> }
> ```
> 
> and methods on `Endpoint` that, for example, make a request, might become fallible, i.e. go from `fn foo(&self) -> Something` to `fn foo(&self) -> Result<Something, Error>` where `foo` then needs to check for the `NotReady` case if the endpoint looks ok, and either return an error if it doesn't or transition to the `Ready` case when it does. That's a little too simplistic since it involves mutation.
> 
> But most of these operations happen on a pool of endpoints, and the manager for that pool, like `ProviderManager` can mediate the mutation. Instead of calling methods directly on the endpoint, call it on the manager which can then call `Endpoint.get_ready(self) -> Result<Self, Error>` and hide the complexity of readiness checking and going from not ready -> ready from the outside world. In that world, you'd call `foo` on the provider manager, maybe with some hints on how to select an appropriate endpoint, and it would succeed if it could find or make an endpoint ready, and fail if none of them are ready.

Thanks for the review! 

There are a couple of nuances on how we use the adapters, for example, we have different types for firehose and for rpc, one of these layers take care of stuff like falling over to a different adapter if one has a high error rate or handling a usage cap per adapter, these also handle some details like turning a single adapter into a pool and things like that. 

These 2 types (firehose and rpc) have very different interfaces so I think merging those is not something I would want to tackle on a PR like this, however I think longer term there could be some benefits of being able to go back and forth between rpc and firehose where that is possible.

The notion of readiness is already encapsulated on the ProviderManager, when calling get_all, the caller already gets ready to use adapters so the rest of the code has no notion of readiness. The reason I decided to encapsulate this on a deeper layer instead of a decorator is because on firehose we turn 1 definition into a connection pool to avoid hitting some limits(can find the issue later but it's not relevant for this conversation why we do it). Hence, the idea of having a decorator approach would then cause the same check to happen 20 (default) times for each adapter instead of once per adapter. This is also similar to how we handle the connection failover on firehose and rpc for the same-ish reason. 

If I understand correctly, the suggestion would be the expose the ProviderManager, essentially invert the layers in relation to where they are now. Ie

As a user of an adapter
```
fn do_something_with_firehose(pm: ProviderManager)  -> FirehoseEndpoint

pm.get_firehose()
```

instead of 
```
fn do_something_with_firehose(fe: FirehoseEndpoints) -> FirehoseEndpoint

fe.endpoint()
```

I like that this unifies the types for adapters behind a single manager type, on the other hand this may have other implications like how it is implemented, right now it uses async and there are some cases where we can't use that (which is why I added some escape hatches). The fact it's a bigger interface is not great but it may open some possibilities to some behaviour we can't mock with the current implementation. My gut feeling is this is quite a large change but probably worth exploring further. Hopefully this is what you meant :smile:

graph/src/components/adapter.rs
lutter: I see two options for addressing that:
1. Change the data model to allow `null` in those columns. That then requires that all code handles the possibility that this is `null`
2. Insist that for new chains we can actually talk to an endpoint to set these fields; if not the chain is ignored like it is today. I think that would be a reasonable requirement, that adding a new chain requires a bit more oversight. It's also not that big a deal if the new chain gets disabled as there can not be any subgraphs for it.

My preference would be (2)

node/src/config.rs
lutter: Yes, I think putting it into its own module would be better. It also seems to me that requiring having a `BlockStore` here would simplify the logic.

lutter:
> I like that this unifies the types for adapters behind a single manager type, on the other hand this may have other implications like how it is implemented, right now it uses async and there are some cases where we can't use that (which is why I added some escape hatches). The fact it's a bigger interface is not great but it may open some possibilities to some behaviour we can't mock with the current implementation. My gut feeling is this is quite a large change but probably worth exploring further. Hopefully this is what you meant 😄

I was a little unclear in my initial message: I didn't mean that we necessarily only have one `Endpoint` and `ProviderManager` struct - it might be that it's better to split that between Rpc and Firehose and have two different structs each. It would be great if we could unify this all the way, but there might be things we want from one that make no sense for the other. I was more driving at that it should be possible to encapsulate the logic around provider testing, failover etc. in something like the `ProviderManager` so that users of these do not need to know about how that happens.

mangas:
> > I like that this unifies the types for adapters behind a single manager type, on the other hand this may have other implications like how it is implemented, right now it uses async and there are some cases where we can't use that (which is why I added some escape hatches). The fact it's a bigger interface is not great but it may open some possibilities to some behaviour we can't mock with the current implementation. My gut feeling is this is quite a large change but probably worth exploring further. Hopefully this is what you meant 😄
> 
> I was a little unclear in my initial message: I didn't mean that we necessarily only have one `Endpoint` and `ProviderManager` struct - it might be that it's better to split that between Rpc and Firehose and have two different structs each. It would be great if we could unify this all the way, but there might be things we want from one that make no sense for the other. I was more driving at that it should be possible to encapsulate the logic around provider testing, failover etc. in something like the `ProviderManager` so that users of these do not need to know about how that happens.

I agree we can have a similar approach for those and just remove them from the specific implementations. I don't think I will try to tackle that on this PR but happy to chase that separately :stuck_out_tongue:

node/src/config.rs
mangas: I've only implemented this for BlockStore and then the different test/mock implementations, I could add a Noop implementation but otherwise we don't have the necessary information to create it without arguments. This struct is not supposed to be mutated manually so the fact it it generated as noop is just intended for situations where it is expected to be used as noop/empty. Not sure yet another empty implementation is really useful, what do you think?

graph/src/components/adapter.rs
mangas: I looked into it and I think it's needs to change a bit again because we will need remove the assumption one will always be returned and then expose a way of adding the store. So I think I will do that on a following PR, I feel like this implementation is ok enough as a first step I think it could be merge as is to make the follow update easier to review and get merged ?

mangas: I decided to give ti a go here since I had some extra time so it's already on the last commit, if you could double-check that it would be great

chain/ethereum/src/network.rs
```diff
@@ -29,7 +28,32 @@ pub struct EthereumNetworkAdapter {
...
+    ) -> Self {
+        Self {
+            endpoint_metrics,
+            capabilities,
+            adapter,
+            limit,
+        }
+    }
+
+    #[cfg(debug_assertions)]
```
zorancv, line 56: Is this not used in release?

graph/src/components/adapter.rs
```diff
@@ -0,0 +1,886 @@
...
+        Ok(self
+            .inner
+            .adapters
+            .get(chain_id)
+            .map(|v| v.iter().map(|v| &v.1).collect())
+            .unwrap_or_default())
+    }
+
+    /// get_all will trigger the verification of the endpoints for the provided chain_id, hence the
+    /// async. If this is undesirable, check `get_all_verified` as an alternatives that does not
```
zorancv, line 321: Spelling: `get_all_unverified`

server/index-node/src/resolver.rs
```diff
@@ -593,7 +593,7 @@ impl<S: Store> IndexNodeResolver<S> {
             }
             BlockchainKind::Starknet => {
                 let unvalidated_subgraph_manifest =
-                    UnvalidatedSubgraphManifest::<graph_chain_substreams::Chain>::resolve(
+                    UnvalidatedSubgraphManifest::<graph_chain_starknet::Chain>::resolve(
```
zorancv, line 596: Nice fix.

zorancv (approved):
Nice that you did refactoring along with functional improvements.

node/src/main.rs
```diff
@@ -323,186 +256,58 @@ async fn main() {
...
+                        })
+                        .collect()
+                }) {
+                Some(eth_network_names) => {
+                    network_store
+                        .block_store()
+                        .cleanup_ethereum_shallow_blocks(eth_network_names)
+                        .unwrap();
+                }
+                None => todo!(),
```
incrypto32, line 307: Will this ever be reached? Is unreachable!() more suitable here?

mangas: yep good catch I've updated and added an explanation

node/src/network_setup.rs
```diff
@@ -0,0 +1,412 @@
...
+            .or_else(|_| {
+                get_identifier(
+                    self.substreams_provider_manager.cheap_clone(),
+                    logger,
+                    chain_id,
+                    "substreams",
+                )
+            })
+        })
+        .await
```
incrypto32, line 178: Minor nitpick
```suggestion
        get_identifier(
            self.rpc_provider_manager.cheap_clone(),
            logger,
            chain_id,
            "rpc",
        )
        .or_else(|_| {
            get_identifier(
                self.firehose_provider_manager.cheap_clone(),
                logger,
                chain_id,
                "firehose",
            )
        })
        .or_else(|_| {
            get_identifier(
                self.substreams_provider_manager.cheap_clone(),
                logger,
                chain_id,
                "substreams",
            )
        })
        .await
```

chain/ethereum/src/network.rs
```diff
@@ -48,64 +72,129 @@ impl EthereumNetworkAdapter {
...
+    pub fn new(
+        chain_id: ChainId,
+        manager: ProviderManager<EthereumNetworkAdapter>,
+        call_only_adapters: Vec<EthereumNetworkAdapter>,
+        retest_percent: Option<f64>,
+    ) -> Self {
+        #[cfg(debug_assertions)]
+        call_only_adapters.iter().for_each(|a| {
+            a.is_call_only();
+        });
```
incrypto32, line 128: Any specific reason we need this here?

chain/ethereum/src/network.rs
```diff
@@ -48,64 +72,129 @@ impl EthereumNetworkAdapter {
...
+            .manager
+            .get_all(&self.chain_id)
+            .await
+            .unwrap_or_default();
+
+        Self::available_with_capabilities(all, required_capabilities)
+    }
+
+    // get all the adapters, don't trigger the ProviderManager's validations because we want this
+    // this function to remain sync
```
incrypto32, line 167:
```suggestion
    // get all the adapters, don't trigger the ProviderManager's validations because we want
    // this function to remain sync
```

incrypto32 (approved):

chain/ethereum/src/network.rs
mangas: just debug time sanity check

node/src/chain.rs
```diff
@@ -528,20 +366,272 @@ pub async fn create_ethereum_networks_for_chain(
...
-    parsed_networks.sort();
-    Ok(parsed_networks)
+    adapters.sort_by(|a, b| {
+        a.capabilities
+            .partial_cmp(&b.capabilities)
+            // We can't define a total ordering over node capabilities,
+            // so incomparable items are considered equal and end up
+            // near each other.
+            .unwrap_or(Ordering::Equal)
+    });
```
lutter, line 384: This is the second time this makes an appearance. Would be better to move that to some helper method. And add a comment what this ordering is and why it is important.

node/src/chain.rs
```diff
@@ -528,20 +366,272 @@ pub async fn create_ethereum_networks_for_chain(
...
+
+    Ok(AdapterConfiguration::Rpc(EthAdapterConfig {
+        chain_id: network_name.into(),
+        adapters,
+        call_only: call_only_adapters,
+        polling_interval: Some(chain.polling_interval),
+    }))
```
lutter, line 391: Why doesn't `EthAdapterConfig` have a constructor that sorts the adapters? I assume the ordering is important for the inner workings of it, and that would be a good place to do the sorting.

node/src/chain.rs
```diff
@@ -427,19 +427,22 @@ pub async fn networks_as_chains(
...
+        let chain_store = match store.chain_store(chain_id) {
+            Some(c) => c,
+            None => {
+                let ident = networks
+                    .chain_identifier(&logger, chain_id)
+                    .await
+                    .expect("must be able to get chain identity to create a store");
+                store
+                    .create_chain_store(chain_id, ident)
+                    .expect("must be able to create store if one is not yet setup for the chain")
```
lutter, line 442: These expects should be errors; otherwise, if you configure a new chain and the provider is down, it won't be possible to start `graph-node` until either the config is changed or the provider comes back. It's fine if the response to the error is to act as if that chain wasn't in the config, but blocking startup entirely in this case is too debilitating.

chain/ethereum/src/network.rs
```diff
@@ -48,64 +72,129 @@ impl EthereumNetworkAdapter {
...
-    pub fn cheapest_with(
+    pub async fn all_cheapest_with(
         &self,
         required_capabilities: &NodeCapabilities,
+    ) -> impl Iterator<Item = &EthereumNetworkAdapter> + '_ {
+        let all = self
+            .manager
+            .get_all(&self.chain_id)
+            .await
+            .unwrap_or_default();
```
lutter, line 161: Is the intent behind the `.unwrap_or_default()` that we'll return an empty iterator?  It might be better if this returned `Result<impl Iterator<Item = &EthereumNetworkAdapter> + '_, SomeError>` to force the caller to handle that condition. If not, the method should at least have a comment on the circumstances under which this returns an empty iterator.

chain/ethereum/src/network.rs
```diff
@@ -48,64 +72,129 @@ impl EthereumNetworkAdapter {
...
+    // get all the adapters, don't trigger the ProviderManager's validations because we want
+    // this function to remain sync
+    pub(crate) fn all_unverified_cheapest_with(
+        &self,
+        required_capabilities: &NodeCapabilities,
+    ) -> impl Iterator<Item = &EthereumNetworkAdapter> + '_ {
+        let all = self
+            .manager
+            .get_all_unverified(&self.chain_id)
+            .unwrap_or_default();
```
lutter, line 175: Same comment on the `unwrap_or_default` as above

graph/src/firehose/endpoints.rs
```diff
@@ -1,826 +1,960 @@
...
+
+                if used_percent < threshold_percent {
+                    return AvailableCapacity::High;
+                }
+
+                AvailableCapacity::Low
+            }
+            _ => AvailableCapacity::High,
+        }
+    }
```
lutter, line 288: Why does this discretize the `SubgraphLimit`? I would have thought you want to select the provider with the biggest available capacity and/or the lowest usage. But it seems this message first classifies providers into 3 groups, and presumably something than picks one from one of the groups.

graph/src/firehose/endpoints.rs
```diff
@@ -1,826 +1,960 @@
...
+                let bearer_token = format!("bearer {}", token);
+                bearer_token.parse::<MetadataValue<Ascii>>().map(Some)
+            })
+            .expect("Firehose token is invalid");
+
+        let key: Option<MetadataValue<Ascii>> = key
+            .map_or(Ok(None), |key| {
+                key.parse::<MetadataValue<Ascii>>().map(Some)
+            })
+            .expect("Firehose key is invalid");
```
lutter, line 342: All these expect and panics will cause a configuration error to stop `graph-node` from starting, and AFAICT, we don't even have tooling to check that part of the configuration. So a typo like `url = "htps://some.where.com/"` will make it impossible to start `graph-node`. These issues should result in an error, and that part of the config should be ignored, but the rest of `graph-node` should start up successfully.

node/src/chain.rs
```diff
@@ -528,20 +366,272 @@ pub async fn create_ethereum_networks_for_chain(
...
+        let adapters: Vec<&AdapterConfiguration> = adapters.into_iter().collect();
+        let kind = adapters
+            .iter()
+            .map(|a| a.kind())
+            .reduce(|a1, a2| match (a1, a2) {
+                (BlockchainKind::Substreams, k) => k,
+                (k, BlockchainKind::Substreams) => k,
+                (k, _) => k,
+            })
+            .expect("each chain should have at least 1 adapter");
```
lutter, line 429: It would be better to log an error and ignore that chain instead of aborting startup of the entire process.

lutter (changes requested):

lutter:
> I agree we can have a similar approach for those and just remove them from the specific implementations. I don't think I will try to tackle that on this PR but happy to chase that separately 😛

When do you think that will happen? We also still have [this issue](https://github.com/graphprotocol/graph-node/pull/4916#discussion_r1413723254) outstanding. So I am a bit hesitant to approve this also on the promise of a future improvement.

mangas:
> > I agree we can have a similar approach for those and just remove them from the specific implementations. I don't think I will try to tackle that on this PR but happy to chase that separately 😛
> 
> When do you think that will happen? We also still have [this issue](https://github.com/graphprotocol/graph-node/pull/4916#discussion_r1413723254) outstanding. So I am a bit hesitant to approve this also on the promise of a future improvement.

The way I see it, we can either learn to live with these "out of scope" improvements in a way that will be addressed somewhere in the future, since they are, after all, out of the scope of this particular PR or we need to essentially accept that whenever we write any PR we will need to fix the world in one go. I think it's already a really bad practice to have a PR large as this as the standard and would prefer not to make that the norm. 

I also don't think whoever works on a specific piece should be "owning" any future improvements since most of those require time that someone needs to spend but anyone can really spend that time. 

In this case if you are unwilling to accept the compromise then I guess I would like to understand what you are willing to accept from this PR so that I can get the needed functionality merged in.

node/src/chain.rs
mangas: If this is a valid state of configuration where an empty section can be provided then sure, I would also like to know what we should do with that setup. 

Do you have a suggestion for the error message?

graph/src/firehose/endpoints.rs
mangas: This code wasn't modified by this, it was just moved, probably because of a rebase and I don't really want to open the scope of the review to that code since it's not part of the PR

graph/src/firehose/endpoints.rs
mangas: same as the previous comment, this wasn't changed, this code has been reviewed elsewhere and hasn't been modified

chain/ethereum/src/ethereum_adapter.rs
```diff
@@ -129,6 +129,43 @@ impl EthereumAdapter {
...
+    pub async fn check_block_receipt_support(
+        web3: Arc<Web3<impl web3::Transport>>,
+        provider: &str,
+        supports_eip_1898: bool,
+        call_only: bool,
+        logger: &Logger,
+    ) -> Result<(), Error> {
+        if call_only {
+            warn!(logger, "Call only providers not supported"; "provider" => provider);
+            return Err(anyhow!("Call only providers not supported"));
```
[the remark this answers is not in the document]
mangas: This code hasn't been added in this PR, I need to check why this difference is showing, might have been part of a rebase

chain/ethereum/src/ingestor.rs
```diff
@@ -188,31 +198,62 @@ impl PollingBlockIngestor {
...
             .map(|block| block.into())
     }
+
+    async fn eth_adapter(&self) -> anyhow::Result<Arc<EthereumAdapter>> {
+        self.chain_client
+            .rpc()?
+            .cheapest()
+            .await
+            .ok_or_else(|| graph::anyhow::anyhow!("unable to get eth adapter"))
+    }
```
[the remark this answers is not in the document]
mangas: do_pool will loop inside IIRC and will only exit if there was an error, So if this will only be called on a retry. Before, if the adapter was broken it would never be replaced. Where are these methods taken an eth_adapter? They are probably called after selection, most place that do adapter selection should have a ChainClient IIRC and then inner methods just get the simpler eth_adapter so they don't need to deal with error. That's the general pattern, this was one of the few places where this behavior hadn't been implemented.

chain/ethereum/src/runtime/runtime_adapter.rs
```diff
@@ -111,7 +111,7 @@ impl blockchain::RuntimeAdapter<Chain> for RuntimeAdapter {
         let ethereum_get_balance = HostFn {
             name: "ethereum.getBalance",
             func: Arc::new(move |ctx, wasm_ptr| {
-                let eth_adapter = eth_adapters.cheapest_with(&NodeCapabilities {
+                let eth_adapter = eth_adapters.unverified_cheapest_with(&NodeCapabilities {
```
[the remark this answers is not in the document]
mangas: If it gets the broken one it will retry, yes it's because of async that is hard to get into this part of the codebase

docker/docker-compose.yml
```diff
@@ -28,7 +28,7 @@ services:
     volumes:
       - ./data/ipfs:/data/ipfs:Z
   postgres:
-    image: postgres:14
+    image: postgres
```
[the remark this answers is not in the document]
mangas: it';s not necessary, I don't think we are using 14 anymore anywhere, do you want to pin it to another version?

graph/src/blockchain/types.rs
```diff
@@ -333,6 +333,21 @@ pub struct ChainIdentifier {
...
+}
+
+impl Default for ChainIdentifier {
+    fn default() -> Self {
+        Self {
+            net_version: String::default(),
+            genesis_block_hash: BlockHash::from(H256::zero()),
+        }
+    }
+}
```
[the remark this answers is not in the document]
mangas: it doesn't mean anything special, it's just easier to build than repeating the same code everywhere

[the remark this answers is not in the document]
mangas: What changes?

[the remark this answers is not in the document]
mangas: If you are referring to [these](https://github.com/graphprotocol/graph-node/pull/5337/files/1623b904c455504b58d6fbcdbe89d2ba09b46b9f#diff-5fc766188c2b64f33c6143f3377c76c8d0f9994c5f93e151f60985c973422691R2306) then they are not strictly necessary. I'm not a big fan of implicit behavior in the sense that the current implementations writes some warning to update the net_version if there are some special values (again somewhat implicit except for a comment). 

I thought it would be more correct to update the value to the latest used. If you prefer not to then I'll revert those changes.

chain/ethereum/src/network.rs
mangas: added a comment

chain/ethereum/src/network.rs
mangas: same

chain/ethereum/src/network.rs
```diff
@@ -48,64 +72,129 @@ impl EthereumNetworkAdapter {
...
 
-impl Default for EthereumNetworkAdapters {
-    fn default() -> Self {
-        Self::new(None)
-    }
-}
-
 impl EthereumNetworkAdapters {
-    pub fn new(retest_percent: Option<f64>) -> Self {
+    pub fn empty() -> Self {
```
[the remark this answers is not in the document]
mangas: Had to revert to renaming because of the build in release mode step on CI that builds the test crate for some reason 
image

node/src/chain.rs
lutter: If reading the config file validates that there is at least one provider, logs an error and removes the chain from the config otherwise, it would be ok to leave the `expect` there to say something like `validation should have checked we have at least one provider`. Otherwise, the error should be something like `Chain {name} does not have any providers configured. Ignoring it`

graph/src/firehose/endpoints.rs
lutter: I didn't realize that this code was moved over. I went through the PR to look at places where it can produce a panic, as we've had several instances now where panics can cause huge operational issues and we need to avoid them as much as possible.

chain/ethereum/src/ingestor.rs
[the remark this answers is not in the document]
lutter: I meant methods like `latest_block` or `ingest_block` just preceding this.

docker/docker-compose.yml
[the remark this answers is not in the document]
lutter: ugh .. we should fix CI then, too. I was just surprised to find this change in this PR. There are definitely people running on PG14, and that's what we should test against until we can figure out how to raise the minimum Postgres version.

lutter:
> > > I agree we can have a similar approach for those and just remove them from the specific implementations. I don't think I will try to tackle that on this PR but happy to chase that separately 😛
> > 
> > 
> > When do you think that will happen? We also still have [this issue](https://github.com/graphprotocol/graph-node/pull/4916#discussion_r1413723254) outstanding. So I am a bit hesitant to approve this also on the promise of a future improvement.
> 
> The way I see it, we can either learn to live with these "out of scope" improvements in a way that will be addressed somewhere in the future, since they are, after all, out of the scope of this particular PR or we need to essentially accept that whenever we write any PR we will need to fix the world in one go. I think it's already a really bad practice to have a PR large as this as the standard and would prefer to avoid it when possible.

I agree that such a large PR is bad practice, and it would have been better to split it into multiple PRs focused on specific aspects (like 'restructure code, no functional change', 'Delay provider checking') especially when the individual commits are not really reviewable by themselves and the commit messages give no hint at what the commit tries to achieve (a message of 'fix tests' does not give me confidence that the implementation followed a plan)

Fixing 'out of scope' improvements is part and parcel of working with a large, mature codebase; there will always be things we understand better now than we did when the code was first written. The work for aggregations for example was 70-80% just refactoring how we deal with the subgraph schema to bring some sanity to that code; the actual functionality was a relatively small amount of the overall work.

> I also don't think whoever works on a specific piece should be "owning" any future improvements since most of those require time that someone needs to spend but anyone can really spend that time.

I think this is a little facile, and sounds a lot like "I don't want to deal with it, somebody else should do it", especially in light of PR 4916 which copy-pasted some very intricate code and which I approved with the understanding that you would fix that - it's been 6 months.

The attitude of "let's commit this and we'll fix issues with it later (never)" is what makes the code around ingestion from providers so complex: the code wasn't in the greatest shape when we had just RPC, we then shoehorned Firehose and substreams on top of it, and now how any of that works is incredibly hard to follow from the code as it is scattered across a million places. This code needs an owner who has a keen interest in improving it - that requires rethinking how we do that, writing up a plan, and actually doing it. Without somebody who understands this code at a deep level, fixing things in isolation will only make matters worse.

> In this case if you are unwilling to accept the compromise then I guess I would like to understand what you are willing to accept from this PR so that I can get the needed functionality merged in.

At a minimum, please file tickets for the various issues I pointed out (like crashing on misconfiguration) We also need to figure out who will own the whole front side of the house so that we can make progress on improving this code.

chain/ethereum/src/ingestor.rs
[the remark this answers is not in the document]
mangas: sorry I'm not sure I follow, once the ingestor fetches the adapter it will re-use the same one until there is an error, in which case we would get a new one which goes through the selection that's already implemented everywhere else. Did I miss some part of it?

node/src/chain.rs
mangas: updated

lutter (approved):
Approving this with the understanding that we will schedule a more thorough revamp of the whole ingestion side of `graph-node` soon.

mangas:
> Approving this with the understanding that we will schedule a more thorough revamp of the whole ingestion side of `graph-node` soon.

Turned both cards into issues so I can link them here: 

https://github.com/graphprotocol/graph-node/issues/5499
https://github.com/graphprotocol/graph-node/issues/5500

They are, however, still not prioritised

PR #5340, opened 2024-04-12
Added Support for Fuel Network in Graph Node

Salka1988:
This pull request adds all the necessary components and functionalities for Graph Node to support the Fuel network. With these changes, Graph Node is now fully equipped to handle interactions with the Fuel network.


abourget:
We've been clarifying the integration point for next gen chains being integrated into `graph-node`, and have come to the conclusion that integrations would come through Substreams, and triggers designed within Substreams modules.. which gives much more flexibility in terms of what can be brought inside `graph-node`, and supports any chain right off the bat.

With the landing of block indexes (soon), Substreams will have the same superpowers that the Firehose has with its indexes. There'll be nothing missing, and we'll be able to leverage a bunch of caches for common modules that wasn't possible with Firehose.

The goal is to shape triggers from Substreams, and most of this will be able to be done with no-code interactions, cookie-cutting Substreams modules, or simply parameterizing them, as the source of a _subgraph_. The final aggregations and layout of the data would be done in AssemblyScript, with Protobuf as a single input message.

This would mean deep integrations within `graph-node` like this wouldn't be necessary nor exist anymore.

Salka1988:
Nice! We support Substreams also. I'll close  fhis PR.

azf20:
Thanks so much for the PR @Salka1988! Are you already running Firehoses / Graph Nodes with these changes applied in your own setup?
As @abourget says, there are some ongoing conversations about the recommended approach for new chain integrations. Would love to get your take from a Fuel perspective, have you considered or experimented with Substreams?

Salka1988:
Yes, the code in this PR also supports Substreams. I implemented `Firehose-core` and `Substreams` for our needs and managed to run `Substreams supported Subgraphs`.

Both configurations work, and if I'm not mistaken, you will remove the first one and everything will go through `substreams'.

```
protocol = "fuel"
provider = [
    { label = "substreams", details = { type = "substreams", url = "http://localhost:10016" }},
    { label = "fuel", details = { type = "firehose", url = "http://localhost:10015" }},
]
```
```
protocol = "substreams"
provider = [
    { label = "substreams", details = { type = "substreams", url = "http://localhost:10016" }},
]
```

I also think that it is faster and simpler to use Substreams.

PR #5357, opened 2024-04-18
store: Select specific attributes for query_window_one_entity

lutter:
The 'select by specific attributes' code had overlooked one place where we do a `select *`


lutter:
I ran a query comparison test of this, and it showed no differences except for the ones caused by issue #5359

mangas (approved):

fordN (approved):
👍

PR #5360, opened 2024-04-20
firehose: Try to be more consistent in parsing to address

leoyvens:
Use the same logic in both places where we parse it. An empty `to` is now assumed to be a contract creation.


sduchesneau (approved):
LGTM! makes https://github.com/graphprotocol/graph-node/pull/5377 redundant :)

PR #5362, opened 2024-04-22
feat/erc725

richtera:
WIP: refactoring out unnecessary code

Implement erc725 namespace
Prepare for https download (API not fully hooked in)
Add standard schema decoding using erc725.decode_key_value
The erc725.rs is a prototype for a real crate.


mangas:
Is there some additional context on this PR? Was there an issue where this was previous discussed?

richtera:
We are trying to index erc725y metadata. This allows retrieval and verification of ipfs and https and arweave URLs (this is probably now possible natively) decoding erc725 data is really hard to do in assemblyscript and we have an available module is rust, go and typescript. In addition there are some tools to read big endian numbers. One problem we have right now is erigon which is required for trace_filter does not allow one of our expensive static calls making the index currently fail

richtera:
This is currently a WIP progress collecting changes for a simplified PR. The goal is to move the erc725 changes to either a substream or as assemblyscript. The remaining things should stay here or be split into multiple PRs depending on it's final size.

Issue #5140, opened 2024-01-15
Graphman copy indexing optimizations


leoyvens:
To reduce the size of a DB shard without downtime, `graphman copy`ing deployments out is a good option. But it needs a few optimizations to be usable en masse, including on index creation. This track two of those optimizations:
1. Create the indexes based on the index creation DDL of the source, not based on the default.
2. Create the BTree indexes after the copy is done, and only the GiST afterwards.

@incrypto32 will mentor @itsjerryokolo on this task.

leoyvens:
We can also apply this for grafts, and copy over custom statistics in addition to the indexes.

lutter:
Here's my take on this:
* In general, we should defer creating BTree attribute indexes on all subgraphs until the subgraph is synced. What we really want is to kick off creating those indexes so that they are ready when the subgraph is synced, but not much before then, but as a first step it's ok to create them when the subgraph becomes synced.
* All other indexes (not BTree or not an attribute index) should be created when the subgraph is first deployed as creating them later can be very very slow with bad consequences for the rest of the system.
* For copies and grafts, we do not want to create the indexes that we create by default for new subgraphs. Instead, we should look at the source of the graft/copy, and recreate all the indexes that actually exist on the source

PR #5386, opened 2024-05-03
Check if the indexes are correct

zorancv:
Somewhat related to the https://github.com/graphprotocol/graph-node/issues/5140 . Check all the attribute indexes at subgraph start and recreate the invalid ones.

### DIFF
--- a/store/postgres/src/deployment_store.rs
+++ b/store/postgres/src/deployment_store.rs
@@ -49,7 +49,6 @@ use graph::schema::{ApiSchema, EntityKey, EntityType, InputSchema};
 use web3::types::Address;
 
 use crate::block_range::{BLOCK_COLUMN, BLOCK_RANGE_COLUMN};
-use crate::catalog::{pg_class, pg_index, pg_namespace};
 use crate::deployment::{self, OnSync};
 use crate::detail::ErrorDetail;
 use crate::dynds::DataSourcesTable;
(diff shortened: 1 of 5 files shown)


store/postgres/src/deployment_store.rs
```diff
@@ -1579,10 +1589,72 @@ impl DeploymentStore {
...
+                    .bind::<Text, _>(index_name.clone())
+                    .get_results::<IndexInfo>(&mut conn)?
+                    .into_iter()
+                    .map(|ii| ii.into())
+                    .collect::<Vec<IndexInfo>>();
+                // TODO: pick one of the above
+                assert!(ii_vec.len() <= 1);
+                // Check if the index is valid. If either isvalid or isready flag of pg_index table
+                // isn't true, drop it and recreate it.
+                if ii_vec.len() == 0 || !ii_vec[0].isvalid || !ii_vec[0].isready {
```
zorancv, line 1644: Criteria for recreating the index is that either `inisvalid` or `indisready`  is false from `pg_index` table. If the respective row is completely missing I'd skip dropping the old index and proceed with creating the new one directly. WDYT @lutter

store/postgres/src/deployment_store.rs
```diff
@@ -1579,10 +1589,72 @@ impl DeploymentStore {
...
             })?;
         }
+
+        //check if all indexes are valid and recreate them
+        let namespace = site.namespace.as_str();
+        let mut conn = self.get_conn()?;
+        for table in dst.tables.values() {
+            for (index_name, create_query) in table.create_postponed_indexes() {
+                let table_name = table.name.clone();
+                // Version 1 - SQL generated with diesel
```
zorancv, line 1599: I am wondering is a diesel version of the query is more maintainable compared to the raw string SQL bellow. WDYT @lutter?

store/postgres/src/deployment_store.rs
zorancv: I guess we decided to pick the raw SQL string.

PR #5399, opened 2024-05-07
chain/ethereum: fix block receipts call failing for some providers

incrypto32:


chain/ethereum/src/ethereum_adapter.rs
```diff
@@ -145,14 +145,20 @@ impl EthereumAdapter {
...
+                )
             })
-            .unwrap_or(false); // Default to false if there's an error in fetching receipts.
+            .unwrap_or_else(|err| {
+                // Store the error message and default to false
+                (false, Some(err.to_string()))
+            });
 
-        info!(logger, "Checked if provider supports eth_getBlockReceipts"; "provider" => provider, "supports_block_receipts" => supports_block_receipts);
+        info!(logger, "Checked if provider supports eth_getBlockReceipts"; "provider" => provider, "supports_block_receipts" => supports_block_receipts, "error" => error_message.unwrap_or_else(|| "".to_string()));
```
lutter, line 161: Minor nit: it would be nicer in the logs if it said `error => none` if there was no error than just an empty string

lutter (approved):

PR #5409, opened 2024-05-13
Check if subgraph head exists before trying to unfail

leoyvens:
Otherwise, looking for the parent ptr will fail and the subgraph simply won't start, logging `Trying again after load block 0x2a74…e47e failed`. When a chain does a re-genensis or re-hashing this can happen en masse so we need to deal with it.


incrypto32 (approved):
LGTM

PR #5419, opened 2024-05-14
store: Make sure transact_block_operations does not go backwards

lutter:


store/postgres/src/writable.rs
```diff
@@ -1658,6 +1658,16 @@ impl WritableStoreTrait for WritableStore {
             self.writer.start_batching();
         }
 
+        if let Some(block_ptr) = self.block_ptr.lock().unwrap().as_ref() {
+            if block_ptr_to.number <= block_ptr.number {
+                return Err(constraint_violation!(
+                    "transact_block_operations called for block {} but its head is alread at {}",
```
zorancv, line 1664: Spelling 'already'.

zorancv (approved):
Assuming that reorgs are handled elsewhere seems a reasonable thing to do.

store/postgres/src/writable.rs
lutter: woops .. fixed

PR #5420, opened 2024-05-15
graph, store: Improve error message when write fails

lutter:
When a write fails because Postgres is unhappy with it, e.g., because of an invalid block_range, it is very hard to understand what happened from the error message. This change improves that by including the entity type and the block at which the write failed as well as some other detail in the error message.


zorancv (approved):
Conversion of errors, so no risk of breaking changes.

PR #5425, opened 2024-05-17
Graphman copy indexing improvements

zorancv:
Addresses the https://github.com/graphprotocol/graph-node/issues/5140 issue. The tests are still pending.

### DIFF
--- a/store/postgres/src/copy.rs
+++ b/store/postgres/src/copy.rs
@@ -13,6 +13,7 @@
 //! `graph-node` was restarted while the copy was running.
 use std::{
     convert::TryFrom,
+    ops::DerefMut,
     sync::Arc,
     time::{Duration, Instant},
 };
@@ -24,8 +25,7 @@ use diesel::{
     pg::Pg,
     r2d2::{ConnectionManager, PooledConnection},
     select,
-    serialize::Output,
-    serialize::ToSql,
+    serialize::{Output, ToSql},
     sql_query,
     sql_types::{BigInt, Integer},
     update, Connection as _, ExpressionMethods, OptionalExtension, PgConnection, QueryDsl,
@@ -36,11 +36,13 @@ use graph::{
     prelude::{info, o, warn, BlockNumber, BlockPtr, Logger, StoreError, ENV_VARS},
     schema::EntityType,
 };
+use itertools::Itertools;
 
 use crate::{
     advisory_lock, catalog,
     dynds::DataSourcesTable,
     primary::{DeploymentId, Site},
+    relational::index::IndexList,
 };
 use crate::{connection_pool::ConnectionPool, relational::Layout};
 use crate::{relational::Table, relational_queries as rq};
@@ -761,7 +763,7 @@ impl Connection {
         Ok(())
     }
 
-    pub fn copy_data_internal(&mut self) -> Result<Status, StoreError> {
+    pub fn copy_data_internal(&mut self, index_list: IndexList) -> Result<Status, StoreError> {
         let src = self.src.clone();
         let dst = self.dst.clone();
         let target_block = self.target_block.clone();
@@ -806,6 +808,46 @@ impl Connection {
             progress.table_finished(&table.batch);
         }
 
+        // Create indexes for all the attributes that were postponed at the start of
+        // the copy/graft operations.
+        // First recreate the indexes that existed in the original subgraph.
+        let conn = self.conn.deref_mut();
+        for table in state.tables.iter() {
+            let arr = index_list.indexes_for_table(
+                &self.dst.site.namespace,
+                &table.batch.src.name.to_string(),
+                &table.batch.dst,
+                true,
+                true,
+            )?;
+
+            for (_, sql) in arr {
+                let query = sql_query(format!("{};", sql));
+                query.execute(conn)?;
+            }
+        }
+
+        // Second create the indexes for the new fields.
+        // Here we need to skip those created in the first step for the old fields.
+        for table in state.tables.iter() {
+            let orig_colums = table
+                .batch
+                .src
+                .columns
+                .iter()
+                .map(|c| c.name.to_string())
+                .collect_vec();
+            for sql in table
+                .batch
+                .dst
+                .create_postponed_indexes(orig_colums)
+                .into_iter()
+            {
+                let query = sql_query(sql);
+                query.execute(conn)?;
+            }
+        }
+
         self.copy_private_data_sources(&state)?;
 
         self.transaction(|conn| state.finished(conn))?;
@@ -820,6 +862,8 @@ impl Connection {
     /// block is guaranteed to not be subject to chain reorgs. All data up
     /// to and including `target_block` will be copied.
     ///
+    /// The parameter index_list is a list of indexes that exist on the `src`.
+    ///
     /// The copy logic makes heavy use of the fact that the `vid` and
     /// `block_range` of entity versions are related since for two entity
     /// versions `v1` and `v2` such that `v1.vid <= v2.vid`, we know that
@@ -828,7 +872,7 @@ impl Connection {
     /// lower(v1.block_range) => v2.vid > v1.vid` and we can therefore stop
     /// the copying of each table as soon as we hit `max_vid = max { v.vid |
     /// lower(v.block_range) <= target_block.number }`.
-    pub fn copy_data(&mut self) -> Result<Status, StoreError> {
+    pub fn copy_data(&mut self, index_list: IndexList) -> Result<Status, StoreError> {
         // We require sole access to the destination site, and that we get a
         // consistent view of what has been copied so far. In general, that
         // is always true. It can happen though that this function runs when
@@ -842,7 +886,7 @@ impl Connection {
             "Obtaining copy lock (this might take a long time if another process is still copying)"
         );
         advisory_lock::lock_copying(&mut self.conn, self.dst.site.as_ref())?;
-        let res = self.copy_data_internal();
+        let res = self.copy_data_internal(index_list);
         advisory_lock::unlock_copying(&mut self.conn, self.dst.site.as_ref())?;
         if matches!(res, Ok(Status::Cancelled)) {
             warn!(&self.logger, "Copying was cancelled and is incomplete");
--- a/store/postgres/src/deployment_store.rs
+++ b/store/postgres/src/deployment_store.rs
@@ -29,8 +29,8 @@ use lru_time_cache::LruCache;
 use rand::{seq::SliceRandom, thread_rng};
 use std::collections::{BTreeMap, HashMap};
 use std::convert::Into;
-use std::ops::Bound;
 use std::ops::Deref;
+use std::ops::{Bound, DerefMut};
 use std::str::FromStr;
 use std::sync::{atomic::AtomicUsize, Arc, Mutex};
 use std::time::{Duration, Instant};
@@ -52,7 +52,7 @@ use crate::deployment::{self, OnSync};
 use crate::detail::ErrorDetail;
 use crate::dynds::DataSourcesTable;
 use crate::primary::DeploymentId;
-use crate::relational::index::{CreateIndex, Method};
+use crate::relational::index::{CreateIndex, IndexList, Method};
 use crate::relational::{Layout, LayoutCache, SqlName, Table};
 use crate::relational_queries::FromEntityData;
 use crate::{advisory_lock, catalog, retry};
@@ -172,6 +172,11 @@ impl DeploymentStore {
         DeploymentStore(Arc::new(store))
     }
 
+    // Parameter index_def is used to copy over the definition of the indexes from the source subgraph
+    // to the destination one. This happens when it is set to Some. In this case also the BTree attribude
+    // indexes are created later on, when the subgraph has synced. In case this parameter is None, all
+    // indexes are created with the default creation strategy for a new subgraph, and also from the very
+    // start.
     pub(crate) fn create_deployment(
         &self,
         schema: &InputSchema,
@@ -180,6 +185,7 @@ impl DeploymentStore {
         graft_base: Option<Arc<Layout>>,
         replace: bool,
         on_sync: OnSync,
+        index_def: Option<IndexList>,
     ) -> Result<(), StoreError> {
         let mut conn = self.get_conn()?;
         conn.transaction(|conn| -> Result<_, StoreError> {
@@ -212,6 +218,7 @@ impl DeploymentStore {
                     site.clone(),
                     schema,
                     entities_with_causality_region.into_iter().collect(),
+                    index_def,
                 )?;
                 // See if we are grafting and check that the graft is permissible
                 if let Some(base) = graft_base {
@@ -744,6 +751,13 @@ impl DeploymentStore {
             Ok(indexes.into_iter().map(CreateIndex::parse).collect())
         })
         .await
+    }
+
+    pub(crate) fn load_indexes(&self, site: Arc<Site>) -> Result<IndexList, StoreError> {
+        let store = self.clone();
+        let mut binding = self.get_conn()?;
+        let conn = binding.deref_mut();
+        IndexList::load(conn, site, store)
     }
 
     /// Drops an index for a given deployment, concurrently.
@@ -1483,12 +1497,12 @@ impl DeploymentStore {
         &self,
         logger: &Logger,
         site: Arc<Site>,
-        graft_src: Option<(Arc<Layout>, BlockPtr, SubgraphDeploymentEntity)>,
+        graft_src: Option<(Arc<Layout>, BlockPtr, SubgraphDeploymentEntity, IndexList)>,
     ) -> Result<(), StoreError> {
         let dst = self.find_layout(site.cheap_clone())?;
 
         // If `graft_src` is `Some`, then there is a pending graft.
-        if let Some((src, block, src_deployment)) = graft_src {
+        if let Some((src, block, src_deployment, index_list)) = graft_src {
             info!(
                 logger,
                 "Initializing graft by copying data from {} to {}",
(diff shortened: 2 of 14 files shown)


store/postgres/src/copy.rs
```diff
@@ -828,7 +871,7 @@ impl Connection {
     /// lower(v1.block_range) => v2.vid > v1.vid` and we can therefore stop
     /// the copying of each table as soon as we hit `max_vid = max { v.vid |
     /// lower(v.block_range) <= target_block.number }`.
-    pub fn copy_data(&mut self) -> Result<Status, StoreError> {
+    pub fn copy_data(&mut self, index_list: IndexList) -> Result<Status, StoreError> {
```
lutter, line 874: It would be good to have a comment here that says that `index_list` is the list of indexes that currently exist on the source

store/postgres/src/deployment_store.rs
```diff
@@ -735,6 +738,25 @@ impl DeploymentStore {
...
+        };
+        let schema_name = site.namespace.clone();
+        let layout = store.layout(conn, site)?;
+        for (_, table) in &layout.tables {
+            let table_name = table.name.as_str();
+            let indexes = catalog::indexes_for_table(conn, schema_name.as_str(), table_name)
+                .map_err(StoreError::from)?;
+            let collect: Vec<CreateIndex> = indexes.into_iter().map(CreateIndex::parse).collect();
+            list.indexes.insert(table_name.to_string(), collect);
+        }
```
lutter, line 756: It would probably be nicer to move the code from line 745 to here into a `IndexList::load(conn: &mut PgConnection, layout: &Layout) -> Result<Self, StoreError>` to encapsulate this a little more.

store/postgres/src/deployment_store.rs
```diff
@@ -1577,10 +1607,62 @@ impl DeploymentStore {
...
+                            if ii_vec.len() > 0 {
+                                let drop_query =
+                                    sql_query(format!("DROP INDEX {}.{};", namespace, index_name));
+                                conn.transaction(|conn| drop_query.execute(conn))?;
+                            }
+                            sql_query(create_query).execute(&mut conn)?;
+                        }
+                    }
+                }
+            }
```
lutter, line 1660: This whole `for` loop could also be a method, maybe `IndexList::recreate_invalid_indexes(&self, conn: &mut PgConnection, layout: &Layout)`

store/postgres/src/deployment_store.rs
```diff
@@ -2004,3 +2086,38 @@ impl PruneReporter for OngoingPruneReporter {
         )
     }
 }
+
+#[derive(Debug)]
+pub struct IndexList {
```
lutter, line 2091: This struct might fit better into `store/postgres/src/relational/index.rs`

store/postgres/src/relational.rs
```diff
@@ -1436,6 +1439,11 @@ pub struct Table {
     /// aggregations, this is the object type for a specific interval, like
     /// `Stats_hour`, not the overall aggregation type `Stats`.
     pub object: EntityType,
+
+    /// The name of the database table for this type ('thing'), snakecased
+    /// version of `object`
+    pub namespace: SqlName,
```
lutter, line 1445: I don't think this is needed; the one place where this is read is in `Table::as_ddl` where you have a `Catalog` and can get the namespace from its `site`. If it truly is needed, the comment needs to be adjusted.

store/postgres/src/relational/index.rs
```diff
@@ -180,6 +180,22 @@ impl Expr {
         }
     }
 
+    fn is_same_kind_columns(current: &Vec<Expr>, orig: &Vec<Expr>) -> bool {
```
lutter, line 183: Is this checking if `current` and `orig` are the same expressions? I am not entirely clear on that. It feels a little weird that the comparison is then done on the string representation of the `Expr`. It would seem more logical to me if this was a method on `Expr`; since that implements `PartialEq`, would that be the right comparison so that you could just say `current == orig`? If that's not the right comparison, maybe having a method `Expr::is_same(&self, other: &Expr) -> bool` would be clearer.

In that case, this method would boil down to `current.iter().zip(orig.iter()).all(|(c,o)| c.is_same(o))` (I think)

I would also call this method `same_columns` - that name seems a bit clearer.

store/postgres/src/relational/index.rs
```diff
@@ -411,6 +427,32 @@ impl CreateIndex {
         }
     }
 
+    pub fn with_nsp(&self, nsp2: String) -> Self {
+        let s = self.clone();
+        match s {
+            CreateIndex::Unknown { defn } => CreateIndex::Unknown { defn },
```
lutter, line 433: This is a tough one: since we failed to parse the create index statement, we have no idea what's in it, and I think `with_nsp` should return an error. The danger is that `defn` will talk about some namespace, but we will then run it thinking we changed the namespace, which will be sure to produce the wrong result.

store/postgres/src/relational/index.rs
```diff
@@ -517,13 +563,137 @@ impl CreateIndex {
...
+        }
+        match self {
+            CreateIndex::Unknown { .. } => false,
+            CreateIndex::Parsed {
+                name,
+                columns,
+                method,
+                ..
+            } => {
+                if *method != Method::BTree && *method != Method::Brin {
```
lutter, line 603: I don't think we want to postpone BRIN indexes. They are small and cheap to build; we probably don't do that much, but if we turned the account-like optimization on during indexing, the BRIN index could help speed up lookups of mutable entities.

store/postgres/src/relational/index.rs
```diff
@@ -517,13 +563,137 @@ impl CreateIndex {
...
                     && method == o_method
-                    && columns == o_columns
+                    && Expr::is_same_kind_columns(columns, o_columns)
                     && cond == o_cond
                     && with == o_with
             }
         }
     }
 
+    pub fn is_id_immutable_table(&self, immutable: bool) -> bool {
```
lutter, line 573: Passing `immutable` in here feels a little strange. I think it would be better to just have this as `is_id(&self) -> bool` and have callers do `immutable && ci.is_id()`

store/postgres/src/relational/index.rs
```diff
@@ -517,13 +563,137 @@ impl CreateIndex {
...
+
+    pub fn fields_exist_in_dest(&self, dest_table: &Table) -> bool {
+        fn column_exists(columns: &Vec<String>, column_name: &String) -> bool {
+            for c in columns.iter() {
+                if *c == *column_name {
+                    return true;
+                }
+            }
+            false
+        }
```
lutter, line 629: Isn't this just `columns.iter().any(|c| c == column_name)`? I am all for leaving this a helper for readability, though

store/postgres/src/relational/index.rs
```diff
@@ -517,13 +563,137 @@ impl CreateIndex {
...
+
+        fn some_column_contained(expresion: &String, columns: &Vec<String>) -> bool {
```
lutter, line 631: Typo in `expression`

store/postgres/src/relational/index.rs
```diff
@@ -517,13 +563,137 @@ impl CreateIndex {
...
+                }
+            }
+            false
+        }
+
+        let cols = &dest_table
+            .columns
+            .iter()
+            .map(|i| i.name.to_string())
+            .collect::<Vec<String>>();
```
lutter, line 644: This is very minor here, and can stay like this, since this code isn't run very often, but in general it is better to avoid allocations like that; if this code was called more often, the helpers above should take a `Table` and iterate over its columns without allocating

store/postgres/src/relational/index.rs
```diff
@@ -517,13 +563,137 @@ impl CreateIndex {
...
+
+        match self {
+            CreateIndex::Unknown { defn: _ } => (),
```
lutter, line 647: Is it correct that we return `true` for an index that we couldn't parse? It would be clearer if this arm had a `return` in it.

store/postgres/src/relational/prune.rs
```diff
@@ -94,7 +94,9 @@ impl TablePair {
         if catalog::table_exists(conn, dst_nsp.as_str(), &dst.name)? {
             writeln!(query, "truncate table {};", dst.qualified_name)?;
         } else {
-            dst.as_ddl(schema, catalog, &mut query)?;
+            // In case of pruning we don't do delayed creation of indexes,
+            // as the asumption is that there is not that much data inserted.
+            dst.as_ddl(schema, catalog, None, &mut query)?;
```
lutter, line 99: Nice!

lutter (dismissed):

store/postgres/src/deployment_store.rs
```diff
@@ -180,6 +180,7 @@ impl DeploymentStore {
         graft_base: Option<Arc<Layout>>,
         replace: bool,
         on_sync: OnSync,
+        is_copy_op: bool,
```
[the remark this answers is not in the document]
zorancv: Meanwhile I removed this flag altogether and replaced if with `Option<IndexList>`. Do you think that way it's readable enough or I should still go for an explicit enum?

store/postgres/src/copy.rs
```diff
@@ -806,6 +808,47 @@ impl Connection {
             progress.table_finished(&table.batch);
         }
 
+        // Create indexes for all the attributes that were postponed at the start of
+        // the copy/graft operations.
+        // First recreate the indexes that existed in the original subgraph.
+        let conn = self.conn.deref_mut();
+        let namespace = self.dst.site.namespace.as_str().to_string();
```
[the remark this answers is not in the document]
zorancv: Seems reasonable. The code in `ddl.rs` becomes a little awkward:
`Namespace::new(self.namespace.to_string()).map_err(|_| fmt::Error)?;`, but it's not so bad I think.

store/postgres/src/deployment_store.rs
zorancv: Great suggestion!

store/postgres/src/relational/index.rs
zorancv: Right! Same with the next one. Need to get better grip of the idiomatic Rust. ;-)

store/postgres/src/relational.rs
zorancv: Yes, it's not needed. The value in catalog is the same. Removed it.

store/postgres/src/relational/index.rs
zorancv: Addressed it, thanks!

store/postgres/src/relational/index.rs
zorancv: Addressed it, thanks!

store/postgres/src/relational/index.rs
zorancv: The motivation was to make the logic conditions simpler with helper, but I agree it looks silly. Changed it.

store/postgres/src/relational/index.rs
zorancv: It's not exactly a comparison for equality but is it "same kind" of expression. For example the `Vid` and `Unknown("vid int8_minmax_multi_ops")` are considered "same kind", also `BlockRangeLower` and `Unknown("lower(block_range) int4_minmax_multi_ops)"`. The current implementation checks if the string representation of each expression of the original is a prefix of the current one. Hence the operation is not commutative either. For sure e comment of the functionality would help here so I am adding the summary of this paragraph.

I guess a proper parsing of `Vid`, `BlockRangeLower`, etc. that would keep the optional operator class as a part of the enum, and then compare the enum types and ignore the presence/absence of the operator class would be more robust, and then also `PartialEq` would work, but I took a shortcut that I believe is good enough here. What do you think?

store/postgres/src/relational/index.rs
```diff
@@ -558,6 +725,118 @@ impl CreateIndex {
...
+        let namespace = &layout.catalog.site.namespace;
+        for table in layout.tables.values() {
+            for (ind_name, create_query) in
+                self.indexes_for_table(namespace, &table.name.to_string(), table, true, true)
+            {
+                if let Some(index_name) = ind_name {
+                    let table_name = table.name.clone();
+                    let query = r#"
+                        SELECT  x.indisvalid           AS isvalid,
+                                x.indisready           AS isready
```
zorancv, line 803: @lutter do you think that check for `indisready` from `pg_index` table is needed, or should I just check the `indisvalid` only? I decided to ignore the `indislive` here...

store/postgres/src/relational/index.rs
zorancv: Right. Added an error result here.

store/postgres/src/relational/ddl_tests.rs
```diff
@@ -155,31 +155,31 @@ fn test_manual_index_creation_ddl() {
 #[test]
 fn generate_ddl() {
     let layout = test_layout(THING_GQL);
-    let sql = layout.as_ddl().expect("Failed to generate DDL");
+    let sql = layout.as_ddl(false).expect("Failed to generate DDL");
```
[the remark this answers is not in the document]
zorancv: Some tests are definitely needed and I plan a separate PR for that. I had them in the initial [PR](https://github.com/graphprotocol/graph-node/pull/5364/files) but now I also need to mock the `IndexList` struct.

store/postgres/src/deployment_store.rs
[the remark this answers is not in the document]
lutter: I like that a lot better! There should be a comment on `create_deployment` explaining what `Some(..)` means and what happens when `None` is passed.

store/postgres/src/relational/index.rs
lutter: Oh .. I see. Yes, that's the downside of this homegrown parsing that it fails for some of the newer things we do because the code was never updated. We should definitely update the code in a follow-on PR and can then hopefully reduce this method to an equality comparison. Or maybe it's time to replace the homegrown code with `sqlparser`

store/postgres/src/relational/index.rs
lutter: Looking at the [docs](https://www.postgresql.org/docs/current/catalog-pg-index.html) I think checking `indisvalid` is enough, though I don't think we should ever see one where `indisready` ir `indislive` are `false`

store/postgres/src/relational/index.rs
zorancv: Here I decided to err on the side of allowing more indexes to be created. Now in the context of with_nsp() returning error for this case I think we can keep it true (explicitly), so that the issue get's logged later on.

store/postgres/src/relational/index.rs
zorancv: Looking again the the docs it seems to me that `inisvalid` is the most strict criteria of the 3. So it feels that when it's true the other two must also be true. So I'll keep that one only.

store/postgres/src/deployment_store.rs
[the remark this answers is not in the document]
zorancv: Added it.

store/postgres/src/relational/ddl_tests.rs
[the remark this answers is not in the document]
zorancv: I added one case where the initial skip is activated. Also added a test with generating the SQL statements for the postponed indexes.

store/postgres/src/relational/index.rs
zorancv: I agree that updating the parsing would be really good. Even if we don't end up with equality comparison, it would be much cleaner than the current comparison against a prefix of a string...

Regarding the homegrown code vs sqlparser I was initially all for improving the homegrown code and it kind of works with some minor changes, but now I am slightly in favour for a sqlparser. Implementing the test code made me stumble upon a few edge cases (need for a space after the indexing method, fixed spaces, trailing semicolon, and support for quotes in fieldnames and namespace), while all those are fixable too, I feel that we would benefit from more robust solution than a regex expression.

store/postgres/src/relational/index.rs
```diff
@@ -558,6 +729,115 @@ impl CreateIndex {
...
+        postponed: bool,
+        concurent_if_not_exist: bool,
+    ) -> Result<Vec<(Option<String>, String)>, Error> {
+        let mut arr = vec![];
+        if let Some(vec) = self.indexes.get(table_name) {
+            for ci in vec {
+                if ci.fields_exist_in_dest(dest_table)
+                    && !ci.is_default_non_attr_index()
+                    && !(ci.is_id() && dest_table.immutable)
+                    && postponed == ci.to_postpone()
```
incrypto32, line 771: Small suggestion, a comment here explaining why all this conditions is neccessary would be great

store/postgres/src/relational/index.rs
```diff
@@ -558,6 +729,115 @@ impl CreateIndex {
...
+        Ok(list)
+    }
+
+    pub fn indexes_for_table(
+        &self,
+        namespace: &Namespace,
+        table_name: &String,
+        dest_table: &Table,
+        postponed: bool,
+        concurent_if_not_exist: bool,
```
incrypto32, line 763:
```suggestion
        concurrent_if_not_exist: bool,
```
typo

incrypto32 (approved):
LGTM, left some minor comments

store/postgres/src/relational/index.rs
zorancv: addressed

store/postgres/src/relational/index.rs
zorancv: Created an [issue](https://github.com/graphprotocol/graph-node/issues/5512) for this discussion.

store/postgres/src/copy.rs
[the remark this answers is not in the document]
zorancv: Oh, correct. I was confused. Removed it.

PR #5427, opened 2024-05-20
Use latest block hash for `check_block_receipts`

incrypto32:


chain/ethereum/src/ethereum_adapter.rs
```diff
@@ -107,20 +107,26 @@ impl EthereumAdapter {
...
-        .await
-        .is_ok();
-
-        info!(logger, "Checked if provider supports eth_getBlockReceipts"; "provider" => &provider, "supports_block_receipts" => supports_block_receipts);
+        let block_receipts_support_result =
+            Self::check_block_receipt_support(web3.clone(), supports_eip_1898, call_only, &logger)
+                .await;
+
+        // Error message to log if the check fails.
+        let error_message = block_receipts_support_result
```
mangas, line 118: this here seems unnecessarily verbose, printing stuff like "call only providers no supported" is not really something we care about and may cause operators to be confused. Maybe a simple thing would be to have `check_block_receipt_support` return Result<bool>, this if it's just a "no" then we don't need to print anything, if there's an actual error we can print it. Alternatively we can define the error (instead of anyhow) and for specific types of errors we can print, although my gut feeling is that you will prolly end up with `enum Error { Unsupported, Unknown(#[from] anyhow::Error) ` which may not be too useful

chain/ethereum/src/ethereum_adapter.rs
```diff
@@ -107,20 +107,26 @@ impl EthereumAdapter {
...
-            supports_eip_1898,
-            call_only,
-            &logger,
-        )
-        .await
-        .is_ok();
-
-        info!(logger, "Checked if provider supports eth_getBlockReceipts"; "provider" => &provider, "supports_block_receipts" => supports_block_receipts);
+        let block_receipts_support_result =
+            Self::check_block_receipt_support(web3.clone(), supports_eip_1898, call_only, &logger)
```
mangas, line 114: the fact this happens at the start means we probably need to set an upper limit on how long we wait, if this is a 1m timeout on start it will add minutes to the total start-up time if a provider is down. Maybe we need to check this lazily?

chain/ethereum/src/ethereum_adapter.rs
incrypto32: Good point that was the case in my initial commit, but decided against it since we might be interested in the actual error message? If the operator sees that a provider which used to support block receipts stopped supporting it, the would be interested in knowing the reason to report to the provider, without manually reproducing? Wdyt?

chain/ethereum/src/ethereum_adapter.rs
incrypto32: noted, making that change, Also should we add a few retries as well?

chain/ethereum/src/ethereum_adapter.rs
mangas: maybe what we should do is have sort of endpoint on the status server that shows the status of these things? It could be useful for other stuff as well as these seem to be stateful decisions made once and kept for the duration of the process. I think could be very useful for the status of endpoints as well as genesis validation which is coming soon. I think logs might not be the right medium for this type of information

chain/ethereum/src/ethereum_adapter.rs
incrypto32: Makes sense, a lazy check or a stateful check is the long term solution for sure. Let me look into making this a lazy loading one.

chain/ethereum/src/ethereum_adapter.rs
incrypto32: Ah, yes that sounds better. That would probably mean us needing store this state somewhere in the DB right?
Especially when we have a multi graph-node setup

chain/ethereum/src/ethereum_adapter.rs
```diff
@@ -1332,14 +1350,20 @@ impl EthereumAdapterTrait for EthereumAdapter {
             })));
         }
         let hashes: Vec<_> = block.transactions.iter().map(|txn| txn.hash).collect();
-        let receipts_future = fetch_receipts_with_retry(
-            web3,
-            hashes,
+
+        let supports_block_receipts_future = self.check_block_receipt_support_and_update_cache(
```
incrypto32, line 1354: There’s probably a bit more optimization that we can do by reusing the receipt from the check for the actual receipt call, but that would require a bit more refactor and didn’t feel worth it

chain/ethereum/src/env.rs
```diff
@@ -157,6 +162,8 @@ struct Inner {
     max_block_range_size: BlockNumber,
     #[envconfig(from = "GRAPH_ETHEREUM_JSON_RPC_TIMEOUT", default = "180")]
     json_rpc_timeout_in_secs: u64,
+    #[envconfig(from = "GRAPH_ETHEREUM_BLOCK_RECEIPTS_TIMEOUT", default = "180")]
```
mangas, line 165: I think this waaaaayyyy too long for startup, I would probably go with something like 5 or 10 seconds

mangas (approved):

Issue #5418, opened 2024-05-14
Track more features on `subgraph_features` table


incrypto32:
Track the below features in subgraph features table

- immutable: true in the schema
- id: Bytes in the schema
- `@aggregation` in the schema
- declared calls (have calls: in the manifest)

PR #5479, opened 2024-06-11
Track more features in subgraph features table 

incrypto32:
Closes [#5418](https://github.com/graphprotocol/graph-node/issues/5418)


graph/src/data/subgraph/mod.rs
```diff
@@ -531,6 +531,10 @@ pub struct DeploymentFeatures {
     pub data_source_kinds: Vec<String>,
     pub network: String,
     pub handler_kinds: Vec<String>,
+    pub has_declared_calls: bool,
```
mangas, line 534: this feels like the wrong way to go about keeping track of the features :rofl:

graph/src/blockchain/mock.rs
```diff
@@ -84,6 +84,10 @@ impl<C: Blockchain> DataSource<C> for MockDataSource {
             .collect()
     }
 
+    fn has_declared_calls(&self) -> bool {
+        true
```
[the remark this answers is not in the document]
incrypto32: Its just there for tests, wanted to just check with `true` since the default implementation for the trait uses `false`

graph/src/data/subgraph/mod.rs
incrypto32: What do you suggest instead?

azf20: Couldn't we put these `bool` in the [features array](https://github.com/graphprotocol/graph-node/blob/0426f3c693586dc2e82046ae52137ffa59af08be/graph/src/data/subgraph/mod.rs#L530)?

incrypto32: `features` array is String array, if we do that we'd need to have names for these features and would need to hardcode those names in the code (we've done something similar for handlers)

incrypto32: We could have json object `others` or something like that if thats cleaner, the advantage is just that we don't need a migration each time we change the schema

lutter: I don't have very strong feelings about any of the approaches; modeling these things is always pretty icky. As long as the reading from/writing to the database is tightly controlled (i.e., a few functions that do the actual db interaction and then translate that into an appropriate data structure at the Rust level) JSONB seems a bit like overkill to me, especially since that would make storing all kinds of things possible, and code would then have to be prepared to deal with reading these in a backwards compatible way.

It might be good to define an enum for the possible boolean features and then store them in an array of that enum; that way, at least there's decent control over what gets stored and we are somewhat protected from typos. I think no matter what we do, changes here will involve a migration unless we go full-on schemaless with JSONB which will then require handling all that in code. 

In my experience, there's a lot of value in limiting what can be stored in the db (eventually, anything that can be stored according to the schema will be stored); I'd be totally fine with leaving these flags as their own columns - as I said, it's going to be a bit awkward no matter what we do.

incrypto32: Thanks @lutter.

@mangas based on David's comments and our discussion in the chat could we go ahead with the PR as it is and revisit options when the table gets more complicated?

This table gets easily rebuilt when graph-node starts up, so its safe to just truncate the table next time when we want to make significant schema changes.

mangas (approved):

PR #5483, opened 2024-06-13
Increase firehose grpc max decode size

mangas:
Increase max decoding size after near testnet failure


incrypto32 (approved):
Worth moving that to a constant?

Issue #3728, opened 2022-07-12
 New `graphman` command to rewind all data related to a chain


tilacog:
We could create a new `graphman chain rewind` command to rewind all subgraphs from a given chain, targeting a specific block.

The command could be invoked in a similar fashion as `graphman rewind`:
```
graphman chain rewind <chain-name> <block-hash>
```

Additionally, this command could also clear the block and call cache up to the target block.

alex-pakalniskis:
Needs to be refactored to run in background, will revisit soon™

PR #5486, opened 2024-06-14
graphman: Add chain rewind subcommand

encalypto:
This rewinds all subgraphs on a chain to a specified block. It has most of the same arguments as `graphman rewind`, but takes a chain name instead of a list of deployments.

Remaining work:

- [ ] Clear block cache up to target block
- [ ] Clear call cache up to target block

#3728

### DIFF
--- a/node/src/bin/manager.rs
+++ b/node/src/bin/manager.rs
@@ -210,16 +210,16 @@ pub enum Command {
         sleep: Duration,
         /// The block hash of the target block
         #[clap(
-            required_unless_present = "start-block",
-            conflicts_with = "start-block",
+            required_unless_present = "start_block",
+            conflicts_with = "start_block",
             long,
             short = 'H'
         )]
         block_hash: Option<String>,
         /// The block number of the target block
         #[clap(
-            required_unless_present = "start-block",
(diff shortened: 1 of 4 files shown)


node/src/bin/manager.rs
```diff
@@ -1391,6 +1428,31 @@ async fn main() -> anyhow::Result<()> {
...
+                    block_number,
+                    sleep,
+                    force,
+                    start_block,
+                } => {
+                    let notification_sender = ctx.notification_sender();
+                    let (store, primary) = ctx.store_and_primary();
+                    let deployments = vec![DeploymentSearch::AllOnChain { chain_name }];
+
+                    commands::rewind::run(
```
incrypto32, line 1443: Correct me if i'm wrong: But the logic here seems to be that we get all the deployments on a chain and then loop over it and rewind them one by one? 
That might be an issue when we have 1000's of deployments and the process would just run for hours.
And if it get cancelled midway, we'd need to start over.

In my opinion we need some way to manage this large number of deployments. Maybe this process should run in the background and we can come back and check the progress with another command. Something like what `graphman copy` has

PR #5488, opened 2024-06-17
Derived fields should respect causality region + Improve tests

### LINKED ISSUE #5267: [Bug] Derived loaders does not respect causality regions
### Bug report

Derived loaders are not respecting causality regions.
Entities created by file data sources are not allowed to be loaded from an onchain handler but it can be loaded when entities are linked via a derived field in another entity. 

### Relevant log output

_No response_

### IPFS hash

_No response_

_No response_

### OS information

None

incrypto32:
Closes #5267

### DIFF
--- a/graph/src/components/store/mod.rs
+++ b/graph/src/components/store/mod.rs
@@ -90,6 +90,7 @@ impl DerivedEntityQuery {
     /// Checks if a given key and entity match this query.
     pub fn matches(&self, key: &EntityKey, entity: &Entity) -> bool {
         key.entity_type == self.entity_type
+            && key.causality_region == self.causality_region
             && entity
                 .get(&self.entity_field)
                 .map(|v| &self.value == v)
--- a/tests/runner-tests/file-data-sources/src/mapping.ts
+++ b/tests/runner-tests/file-data-sources/src/mapping.ts
@@ -107,44 +107,40 @@ export function handleTestEvent(event: TestEvent): void {
 }
 
 export function handleFile(data: Bytes): void {
-  log.info("handleFile {}", [dataSource.stringParam()]);
+  log.info('handleFile {}', [dataSource.stringParam()]);
   let context = dataSource.context();
 
-  if (context.isSet("command")) {
-    let contextCommand = context.getString("command");
-
-    if (contextCommand == SPAWN_FDS_FROM_OFFCHAIN_HANDLER) {
-      let hash = context.getString("hash");
-      log.info("Creating file data source from handleFile: {}", [hash]);
-      dataSource.createWithContext("File", [hash], new DataSourceContext());
-    }
-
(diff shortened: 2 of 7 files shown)


graph/src/components/store/mod.rs
```diff
@@ -90,6 +90,7 @@ impl DerivedEntityQuery {
     /// Checks if a given key and entity match this query.
     pub fn matches(&self, key: &EntityKey, entity: &Entity) -> bool {
         key.entity_type == self.entity_type
+            && key.causality_region == self.causality_region
```
zorancv, line 93: Love short implementations!

tests/runner-tests/file-data-sources/src/mapping.ts
```diff
@@ -4,130 +4,148 @@ import {
...
+  let context = dataSource.context();
+
+  if (context.isSet("command")) {
+    let contextCommand = context.getString("command");
+
+    if (contextCommand == SPAWN_FDS_FROM_OFFCHAIN_HANDLER) {
+      let hash = context.getString("hash");
+      log.info("Creating file data source from handleFile: {}", [hash]);
+      dataSource.createWithContext("File", [hash], new DataSourceContext());
+    }
```
zorancv, line 120: Not sure what's idiomatic for TypeScript bu wouldn't `else if` look more like a switch statement?

zorancv (approved):
Cool that we can do a test with a blockchain!

tests/runner-tests/file-data-sources/src/mapping.ts
```diff
@@ -4,130 +4,148 @@ import {
...
+
+    if (contextCommand == CREATE_FOO) {
+      log.info("Creating FileEntity with relation to Foo", []);
+      let entity = new FileEntity(dataSource.stringParam());
+      entity.foo = dataSource.stringParam();
+      entity.content = data.toString();
+      entity.save();
+    }
+  } else {
```
mangas, line 141: Can remove a nesting level by doing this block if not command and returning. Then you know all the rest are commands. as zoran said it's prolly more of a switch case, mostly for making it a bit easier to maintain and read if you think it's worth. But also none of the ifs return when match so that could create some weird test situations :laughing:

tests/runner-tests/file-data-sources/src/mapping.ts
incrypto32: Agree, it should have been an else if.

We cannot use switch cases though since switch-case in assemblyscript does not work with string
https://github.com/AssemblyScript/assemblyscript/issues/1643

tests/runner-tests/file-data-sources/src/mapping.ts
incrypto32: Agreed, making the change.

PR #5504, opened 2024-06-21
Add arbitrum-sepolia chain ID to GRAPH_ETH_CALL_NO_GAS default value

ziyadonji:
This PR adds  arbitrum-sepolia chain ID to GRAPH_ETH_CALL_NO_GAS default value

### DIFF
--- a/chain/ethereum/src/env.rs
+++ b/chain/ethereum/src/env.rs
@@ -184,6 +184,6 @@ struct Inner {
     target_triggers_per_block_range: u64,
     #[envconfig(from = "GRAPH_ETHEREUM_GENESIS_BLOCK_NUMBER", default = "0")]
     genesis_block_number: u64,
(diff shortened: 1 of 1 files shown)


chain/ethereum/src/env.rs
```diff
@@ -182,7 +182,7 @@ struct Inner {
         default = "100"
     )]
     target_triggers_per_block_range: u64,
-    #[envconfig(from = "GRAPH_ETHEREUM_GENESIS_BLOCK_NUMBER", default = "0")]
+    #[envconfig(from = "GRAPH_ETH_CALL_NO_GAS", default = "421613,421614")]
```
incrypto32, line 185: I think you meant to change the line below

incrypto32 (changes requested):

chain/ethereum/src/env.rs
ziyadonji: yeah right.

incrypto32 (approved):

PR #5505, opened 2024-06-24
store: truncate subgraph_features table before altering

incrypto32:


zorancv (approved):
The row is recreated on each subgraph when I start it, so LGTM.

PR #5517, opened 2024-06-28
graphman config check providers

mangas:


store/postgres/src/chain_store.rs
```diff
@@ -2308,15 +2309,25 @@ impl ChainStoreTrait for ChainStore {
...
+                ))
+                .execute(conn)?;
+
+            diesel::update(c::table.filter(c::name.eq(&self.chain)))
+                .set((
+                    c::genesis_block_hash.eq(ident.genesis_block_hash.hash_hex()),
+                    c::net_version.eq(&ident.net_version),
+                ))
+                .execute(conn)
+        })?;
```
incrypto32, line 2330: Bit confused about the location of these tables. Isn't the `chains` table on the `primary` shard and `ethereum_network` on the shard of the block cache for the chain is, so in this transaction we are only updating the block cache shard right?
Don't we need to update in primary also?

node/src/manager/commands/config.rs
```diff
@@ -16,10 +16,41 @@ use graph::{
...
+pub async fn providers(networks: &Networks, store: Arc<BlockStore>) -> Result<(), Error> {
+    println!("Checking providers");
+    for (chain_id, ids) in networks.all_chain_identifiers().await.into_iter() {
+        let (_oks, errs): (Vec<_>, Vec<_>) = ids
+            .into_iter()
+            .map(|(provider, id)| {
+                id.map_err(IdentValidatorError::from)
+                    .and_then(|id| store.check_ident(chain_id, &id).map(|_| (provider, id)))
+            })
+            .partition_result();
```
incrypto32, line 36: Minor nitpick: Felt like this bit could be made a bit more readable, feel free to leave as it is if there's time constraint

node/src/manager/commands/config.rs
```diff
@@ -16,10 +16,41 @@ use graph::{
...
+            .dedup()
+            .collect::<Vec<IdentValidatorError>>();
+
+        if errs.is_empty() {
+            println!("chain_id: {}: status: OK", chain_id);
+            continue;
+        }
+
+        println!("chain_id: {}: status: NOK", chain_id);
+        println!("errors: {:?}", errs);
```
incrypto32, line 48: I think it would be helpful to printout the provider label of the specific provider which failed the test. Not sure if it already does that, the errors seemed to be only printing the chain name

node/src/network_setup.rs
```diff
@@ -131,6 +134,48 @@ impl Networks {
...
+    )> {
+        let mut out = vec![];
+        for chain_id in self.adapters.iter().map(|a| a.chain_id()).sorted().dedup() {
+            // let mut adapters: Vec<dyn NetIdentifiable> =
+            //     self.rpc_provider_manager.get_all_unverified(chain_id);
+            // adapters.extend(self.firehose_provider_manager.get_all_unverified(chain_id));
+            // adapters.extend(
+            //     self.substreams_provider_manager
+            //         .get_all_unverified(chain_id),
+            // );
```
incrypto32, line 157: Should we remove these comments?

fordN (approved):
👍

PR #5544, opened 2024-07-17
Subgraph Composition: Reading the entities for subgraph as a datasource

zorancv:
Contains initial implementation of the database access functionality for fetching a single entity in a range of blocks. Will later be extended to support multiple entities.

TODOs in the code are addressed in later PRs.


store/postgres/src/relational_queries.rs
```diff
@@ -1971,6 +2006,54 @@ impl<'a> Query for FindQuery<'a> {
...
+        // Generate
+        //    select '..' as entity, to_jsonb(e.*) as data
+        //      from schema.table e where id = $1
+        out.push_sql("select ");
+        out.push_bind_param::<Text, _>(self.table.object.as_str())?;
+        out.push_sql(" as entity, to_jsonb(e.*) as data\n");
+        out.push_sql("  from ");
+        out.push_sql(self.table.qualified_name.as_str());
+        out.push_sql(" e\n where ");
+        // TODO: do we need to care about it?
```
zorancv, line 2035: Yes, we do. Enabled in later [PR](https://github.com/graphprotocol/graph-node/pull/5614).

store/test-store/tests/postgres/writable.rs
```diff
@@ -286,3 +342,47 @@ fn restart() {
...
+    };
+    let e = writable.get_range(et, br).unwrap();
+    e.iter().map(|(_, v)| v.iter()).flatten().count()
+}
+
+#[test]
+fn read_range_mutable() {
+    run_test(|store, writable, deployment| async move {
+        let num_entities = read_range(store, writable, deployment, true).await;
+        assert_eq!(num_entities, 6) // TODO: fix it - it should be 4 as the range is open
```
zorancv, line 378: Will be reworked later on.

graph/src/components/store/traits.rs
```diff
@@ -227,6 +228,13 @@ pub trait ReadStore: Send + Sync + 'static {
         keys: BTreeSet<EntityKey>,
     ) -> Result<BTreeMap<EntityKey, Entity>, StoreError>;
 
+    /// Looks up entities using the given store key for a range of blocks.
```
lutter, line 231: I would reword this as

> Returns all versions of entities of the given `entity_type` that were changed in the given `block_range`

store/postgres/src/block_range.rs
```diff
@@ -132,6 +132,66 @@ impl<'a> QueryFragment<Pg> for BlockRangeUpperBoundClause<'a> {
...
 
+/// Helper for generating SQL fragments for selecting entities in a specific block range
+#[derive(Debug, Clone, Copy)]
+pub enum EntityBlockRange {
+    Mutable(BlockRange), // TODO: check if this is a proper type here (maybe Range<BlockNumber>?)
+    Immutable(BlockRange),
+}
+
+impl EntityBlockRange {
+    pub fn new(immutable: bool, block_range: std::ops::Range<BlockNumber>) -> Self {
```
lutter, line 143: IMHO, it would be nicer to pass a `&Table` here so that the caller says 'make a block range for this table' and doesn't need to worry about what that entails

store/postgres/src/relational.rs
```diff
@@ -514,6 +515,32 @@ impl Layout {
...
+        let table = self.table_for_entity(entity_type)?;
+        let mut entities: BTreeMap<BlockNumber, Vec<Entity>> = BTreeMap::new();
+        if let Some(vec) = FindRangeQuery::new(table.as_ref(), block_range)
+            .get_results::<EntityData>(conn)
+            .optional()?
+        {
+            for e in vec {
+                let block = e.clone().deserialize_block_number::<Entity>()?;
+                let en = e.deserialize_with_layout::<Entity>(self, None)?;
+                match entities.get_mut(&block) {
```
lutter, line 533: This `match` statement would be more idiomatic as `entities.entry(&block).or_default().push(en)`

store/postgres/src/relational_queries.rs
```diff
@@ -479,6 +482,38 @@ impl EntityData {
         schema.entity_type(&self.entity).unwrap()
     }
 
+    pub fn deserialize_block_number<T: FromEntityData>(self) -> Result<BlockNumber, StoreError> {
+        use serde_json::Value as j;
+        match self.data {
+            j::Object(map) => {
+                let mut entries = map.into_iter().filter_map(move |(key, json)| {
+                    if key == "block_range" {
```
lutter, line 490: There's a constant `BLOCK_RANGE_COLUMN` that should be used here

store/postgres/src/relational_queries.rs
```diff
@@ -479,6 +482,38 @@ impl EntityData {
...
+                        let r = json.as_str().unwrap();
+                        let rx = Regex::new("\\[(?P<start>[0-9]+),([0-9]+)?\\)").unwrap();
+                        let cap = rx.captures(r).unwrap();
+                        let start = cap
+                            .name("start")
+                            .map(|mtch| mtch.as_str().to_string())
+                            .unwrap();
+                        let n = start.parse::<i32>().unwrap();
+                        Some(n)
+                    } else if key == "block$" {
```
lutter, line 500: There's a constant `BLOCK_COLUMN` that should be used here

store/postgres/src/relational_queries.rs
```diff
@@ -1971,6 +2006,54 @@ impl<'a> Query for FindQuery<'a> {
 
 impl<'a, Conn> RunQueryDsl<Conn> for FindQuery<'a> {}
 
+#[derive(Debug, Clone)]
+pub struct FindRangeQuery<'a> {
```
lutter, line 2010: Have you thought about doing it the way `Layout::find` now does it? That would save a roundtrip through JSON. For the block, you'd want to add either `block$` or `lower(block_range) as block$` into the selected columns.

store/postgres/src/writable.rs
```diff
@@ -1217,6 +1228,14 @@ impl Queue {
         Ok(map)
     }
 
+    fn get_range(
+        &self,
+        entity_type: &EntityType,
+        block_range: Range<BlockNumber>,
+    ) -> Result<BTreeMap<BlockNumber, Vec<Entity>>, StoreError> {
+        self.store.get_range(entity_type, block_range)
```
lutter, line 1236: This will return the wrong result if `block_range` covers blocks that are still in the queue - those aren't in the database yet. You'd need to use the `BlockTracker` to go through the queue and find entities for those blocks. You also only want to look in the store for the part of `block_range` that isn't in the queue.

lutter (changes requested):

store/postgres/src/relational.rs
zorancv: Beautiful. I remember pondering if it can become one-liner with no luck. Thanks!

store/postgres/src/writable.rs
zorancv: I've removed it from the `Queue` and from the `Writer` and left if only in the `WritableStore` with additional comment. Would that be acceptable? Alternative is to create `ReaderStore` struct.

store/postgres/src/relational_queries.rs
zorancv: It crossed my mind a couple times before but as discussed offline, I would prefer to postpone it for a later moment in order not to increase the complexity here.

store/postgres/src/writable.rs
zorancv: I decided to go with a new trait since the methods of readStore would need to be implemented in WritableStore too.

zorancv: Right now the `WritableStore` is a private member of the `SourceableStore` struct. It is only used for implementing the `DeploymentCursorTracker` methods of the `SourceableStore` trait, which replaces ReadStore in this context. I hope that addresses your concerns sufficiently.

zorancv: Ok finally removed all traces of `WritableStore` from the `SourceableStore`. Feel free to review it with the other two PRs. Sorry for the confusion before.

graph/src/blockchain/mod.rs
```diff
@@ -189,7 +191,7 @@ pub trait Blockchain: Debug + Sized + Send + Sync + Unpin + 'static {
         deployment: DeploymentLocator,
         store: impl DeploymentCursorTracker,
         start_blocks: Vec<BlockNumber>,
-        source_subgraph_stores: Vec<(DeploymentHash, Arc<dyn ReadStore>)>,
+        source_subgraph_stores: Vec<(DeploymentHash, Arc<dyn SourceableStore>)>,
```
lutter, line 194: You don't necessarily need the `DeploymentHash` here since `SourceableStore.input_schema().id()` will return it (and there could be a convenience method on `SourceableStore` for it)

Actually, not having it explicit lowers the risk that there is a mismatch between the two.

core/src/subgraph/instance_manager.rs
```diff
@@ -206,48 +205,27 @@ impl<S: SubgraphStore> SubgraphInstanceManager<S> {
 
     pub async fn hashes_to_read_store<C: Blockchain>(
```
lutter, line 206: Maybe this should be renamed now that it returns `SourceableStore` not `ReadStore`

graph/src/util/lfu_cache.rs
```diff
@@ -179,7 +179,7 @@ impl<K: Clone + Ord + Eq + Hash + Debug + CacheWeight, V: CacheWeight + Default>
         })
     }
 
-    pub fn iter<'a>(&'a self) -> impl Iterator<Item = (&K, &V)> {
+    pub fn iter<'a>(&'a self) -> impl Iterator<Item = (&'a K, &'a V)> {
```
lutter, line 182: Did `rustc` start complaining about the lifetimes here?

lutter (changes requested):
Nice! I think this is almost ready to merge, but it would be good to address the redundant passing of `(DeploymentHash, SourceableStore)` and just use `SourceableStore` there.

graph/src/util/lfu_cache.rs
zorancv: Yes. Already [merged](https://github.com/graphprotocol/graph-node/pull/5725) in master as our CI updates automatically to the the latest version of rustc.

lutter (approved):

PR #5549, opened 2024-07-17
Improve net_identifiers call with timeout

mangas:
- Add timeout when checking providers
- When running graphman patch the pool-size to avoid unnecessary errors


fordN (approved):
Nice, looks good.

PR #5551, opened 2024-07-17
fix: resolves #5550 - make graphql  value nullable

### LINKED ISSUE #5550: [Bug] `paused_at` null values lead to error querying status endpoint
### Bug report

`paused_at` values in the db lead to errors when querying the status endpoint.

Test case:
```
curl -X POST http://localhost:8030/graphql \
-H "Content-Type: application/json" \
-H "Authorization: Bearer YOUR_ACCESS_TOKEN" \
-d '{  "query": "{ indexingStatuses { node paused } }"  }'
```

Errors observed in-situ:
```
{"errors":[{"locations":[{"column":27,"line":1}],"message":"Null value resolved for non-null field `paused`"},{"locations":[{"column":27,"line":1}],"message":"Null value resolved for non-null field `paused`"},{"locations":[{"line":1,"column":27}],"message":"Null value resolved for non-null field `paused`"},{"locations":[{"line":1,"column":27}],"message":"Null value resolved for non-null field `paused`"},{"locations":[{"column":27,"line":1}],"message":"Null value resolved for non-null field `paused`"},{"locations":[{"line":1,"column":27}],"message":"Null value resolved for non-null field `paused`"},{"locations":[{"column":27,"line":1}],"message":"Null value resolved for non-null field `paused`"},{"locations":[{"column":27,"line":1}],"message":"Null value resolved for non-null field `paused`"},{"locations":[{"line":1,"column":27}],"message":"Null value resolved for non-null field `paused`"},{"locations":[{"line":1,"column":27}],"message":"Null value resolved for non-null field `paused`"},{"locations":[{"column":27,"line":1}],"message":"Null value resolved for non-null field `paused`"},{"locations":[{"column":27,"line":1}],"message":"Null value resolved for non-null field `paused`"}]}
```

### Relevant log output

_No response_

### IPFS hash

_No response_

_No response_

### OS information

None

dwerner:
This PR fixes #5550 by making the `paused` field in the status graphql nullable, which matches the struct field of `Option<bool>` and allows the indexer-agent to handle the `None/null` case when appropriate.

### DIFF
--- a/server/index-node/src/schema.graphql
+++ b/server/index-node/src/schema.graphql
@@ -70,8 +70,12 @@ type SubgraphIndexingStatus {
   nonFatalErrors: [SubgraphError!]!
   chains: [ChainIndexingStatus!]!
   entityCount: BigInt!
+
+  "null if deployment is not assigned to an indexing node"
   node: String
-  paused: Boolean!
+  "null if deployment is not assigned to an indexing node"
+  paused: Boolean
+
   historyBlocks: Int!
 }
 


dwerner:
indexer-agent PR https://github.com/graphprotocol/indexer/pull/958

incrypto32:
@dwerner we'd need to address the root cause here, the paused field should never be null, it should either be `true` or `false`, The only reason i can see that a null field is possible there is when the `deployment_schemas` table doesn’t have the subgraph_hash we are looking for, im not sure if thats even possible.
If that is the case then the paused field gets populated with a None

https://github.com/graphprotocol/graph-node/blob/0530ce13796bde0259910f96c15cecb3696650b4/store/postgres/src/primary.rs#L632

To address the root cause we'll need to inspect the database if this is reproduced somewhere

dwerner:
> The only reason i can see that a null field is possible there is when the `deployment_schemas` table doesn’t have the subgraph_hash we are looking for, im not sure if thats even possible.

This was observed in a user's database, and the reason behind the bug I submitted. It's clearly possible, in fact it did happen.

Can you tell me why the field (to quote you) "should never be null", if the field is `Option<bool>`? We should use the type system to express what's possible, and not rely on the happy path here.

For the indexer-agent's purposes, we can handle this field being null (which we now do because this bug existed), and it has meaning: it is the same as `paused == false`.

dwerner:
Closing this as a root cause fix is preferred

dwerner:
In a conversation with @lutter and @fordN, deployments can exist where assignments do not. Then this being `bool!` fails where an assignment does not exist. So this is the correct way to express this. Just like the `node` field above it.

lutter (approved):
Looks good. Could you add comments to the `node` and `paused` field in the schema to say that a `null` means that the deployment is not assigned to an indexing node?

dwerner:
image

Issue #3729, opened 2022-07-12
Expose `graphman` commands through `graph-node`'s admin GraphQL interface


tilacog:
We could expose all `graphman`'s commands as GraphQL mutations over the indexer management API endpoint _(port `18000`, by default)_.

In this design, `graph-node` would act as a service, exposing all necessary routes and returned types for other tools to interface with. 

This concept is similar to `docker`, which is implemented as a daemon.

chriswessels:
This is how the `indexer` components work. `indexer-agent` exposes an indexer management GraphQL API, and `indexer-cli` presents a CLI interface over this strongly typed and inspectable API.

lutter:
The main reason why `graphman` doesn't work that way is authentication: with the CLI tool, authn is punted to the config file where you proved that you can get to the DB passwords. With an API, we'd need to make sure that that API is not accessible to unauthorized parties, especially since `graphman` misuse can be catastrophic for an indexer (e.g., delete all subgraphs)

chriswessels:
In both instances, there's a client and a server. In both instances, the connection between them must be secured.

Today the client is `graphman` and the server is PostgreSQL. With this proposal, the server becomes a `graph-node` instance (which in turn writes to Postgres). IMO this is much easier to secure than access to the database. This is also more reliable in the event that the connection between the client and the server is spotty.

Moving to an API approach also makes automating the tasks `graphman` does much easier. Instead of spawning an executable (`graphman`) that has access to the database and its password, you could interact with an API that provides a constrained set of management tasks and an independent auth mechanism that doesn't risk leaking the DB pass.

YassinEldeeb:
# RFC: Integration of Graphman Commands into GraphQL Server for Graph-node

## Summary

This RFC proposes the integration of core functions from the Graphman CLI into a new GraphQL server for Graph-node that runs on port `8050` spawned as a child process.

The core functions, which have been successfully extracted and reused across the CLI can be reused within the GraphQL server.

Starting with the `stats` subcommands as a starting point will be useful:

- `account-like`: Toggle whether a table is account-like
- `analyze`: Perform a SOL ANALYZE in a Entity table
- `set-target`: Print this message or the help of the given subcommand (s)
- `show`: Show statistics for the tables of a deployment
- `target`: Set the statistics targets for the statistics collector

These functions are essential for the operation of the Graphman CLI and they provide valuable capabilities for the GraphQL server.

## Motivation

[...]

PR #5554, opened 2024-07-18
graphman: create GraphQL API to execute commands

isum:
This PR creates a new graphman GraphQL API with initial support for deployment info, pause, resume and restart commands. It also provides the building blocks needed to support more complex graphman commands.

Part of #3729

### Note

The GraphQL API will be extended incrementally, so there will be a transition period when graphman CLI modules will have weird type conversions and potentially duplicate code. That will go away when all commands are extracted into the core crate.

### Todos

- [x] Integrate extracted commands back into graphman CLI
- [x] Add tests to the API using a real database connection
- [x] Add documentation for the new graphman API
- [x] Add support for long running graphman commands
- [x] Fix metrics initialization error
- [x] ~Extend the API with the remaining commands~ _Will be in the scope of another PR_
- [x] Add documentation on how to extend the GraphQL API with new commands


encalypto (approved):
Generally LGTM! 👍 

It might be worth getting a second opinion for the command creation and extension layering APIs

node/src/bin/manager.rs
```diff
@@ -1585,3 +1560,222 @@ async fn main() -> anyhow::Result<()> {
...
+    let store = Arc::new(GraphmanStore::new(ctx.pool.clone()));
+    let ctx = GraphmanExtensionContext::new(ctx);
+
+    let _restarted = command
+        .layer(IdentifyCommand::new(CommandKind::RestartDeployment.into()))
+        .layer(HandleBrokenExecutions::new(store.clone()))
+        .layer(PreventDuplicateExecutions::new(store.clone()))
+        .layer(TrackExecution::new(store))
+        .execute(ctx)
+        .await?;
```
lutter, line 1665: For the CLI, that's not really what you want: you want the execution to be synchronous so that execution stops immediately when the user hits ^C.

Cargo.toml
```diff
@@ -24,25 +27,49 @@ repository = "https://github.com/graphprotocol/graph-node"
...
 
 [workspace.dependencies]
-diesel = { version = "2.1.3", features = ["postgres", "serde_json", "numeric", "r2d2", "chrono"] }
+anyhow = "1.0"
+async-graphql = { version = "7.0.6", features = ["chrono", "uuid"] }
+async-graphql-axum = "7.0.6"
+axum = "0.7.5"
+chrono = "0.4.38"
+clap = { version = "4.5.4", features = ["derive", "env"] }
+constant_time_eq = "0.3.0"
```
lutter, line 36: Do we really need this? Where it's used, to compare tokens, doesn't seem all that time critical. And looking at the source, I don't see how it compares strings in constant time

core/graphman_extensions/src/handle_broken_executions.rs
```diff
@@ -0,0 +1,145 @@
...
+use crate::GraphmanExtensionStore;
+
+error_builder!("HandleBrokenExecutions");
+
+// Command executions will be considered broken if they have
+// not received any updates for this duration.
+const DEFAULT_MAX_INACTIVE_TIME: Duration = Duration::from_secs(300);
+
+/// Marks command executions as failed if they did not receive any
+/// updates for the specified duration.
```
lutter, line 24: There's a setup for running jobs periodically in `store/postgres/src/jobs.rs`. Couldn't we just add a job that periodically goes through executions and marks them as failed?

docs/graphman-graphql-api.md
```diff
@@ -0,0 +1,213 @@
+# Graphman GraphQL API
+
+When starting the `graph-node` optionally, a graphman GraphQL server can be started, which makes it possible to execute
+graphman commands via GraphQL.
+
+To start the graphman GraphQL server, the `GRAPHMAN_SERVER_AUTH_TOKEN` environment variable should be set. It is
+required that the authentication token be exactly 64 characters long. This is equivalent to an SHA-256 hash represented
```
lutter, line 7: Do we need to require that? Why can't we leave it up to the person running `graph-node` how big/secure they want to make their token?

graph/src/env/mod.rs
```diff
@@ -212,6 +212,24 @@ pub struct EnvVars {
...
+
+    /// Set the maximum query depth for the graphman GraphQL server.
+    ///
+    /// The default value is 255.
+    pub graphman_server_query_max_depth: u8,
+
+    /// Set the maximum query complexity for the graphman GraphQL server.
+    ///
+    /// The default value is 1,000,000.
+    pub graphman_server_query_max_complexity: u32,
```
lutter, line 232: We've tried to use query complexity for the 'normal' GraphQL API, and found it's pretty much useless in determining how hard a query is. Do we really need a setting for max complexity at all?

node/src/bin/manager.rs
```diff
@@ -1585,3 +1560,222 @@ async fn main() -> anyhow::Result<()> {
...
+        status,
+    };
+
+    let ctx = make_graphman_context(ctx);
+    let deployments = command.execute(ctx).await?;
+
+    print_deployments(deployments);
+
+    Ok(())
+}
```
lutter, line 1618: IMHO, it would be better to stick with the current convention of keeping the implementation of CLI commands in their own files instead of putting them into `manager.rs`

lutter (changes requested):
The code is very well written, and very clean. Nice! I am still not convinced we need all the machinery around layering and that background execution could be simplified - I'll leave a longer comment about that and some other things I noticed.

lutter:
I noticed a few things in the GraphQL schema (which is well documented):
* the schema has a `QueryRoot` and `MutationRoot`. IIRC, at least for queries, that must be called `Query`
* for `info` to filter by namespace, you have to say `info(deployment: { namespace: { namespace: "sgd86"}} )`, i.e., mention `namespace` twice. Couldn't that just be `info(deployment: { namespace: "sgd86"})` ? Similarly, when looking for a subgraph by name, that could be shortened a bit, too.
* very minor, but I don't think the dpeloyment info should expose the `id`, that's an implementation detail.

lutter:
For the general approach how to expose the same business logic in the CLI and API, if you have a function `biz` that implements the business logic (I'll write as if this was a thunk, but it should probably be a trait), you want to do one of two things:

### Run `biz` in the CLI
* parameters come from the command line (clap) and possibly some general setup like the store
* the function `biz()` should run synchronously, even if the command might take a long time
* output should go straight to the terminal (ideally, there's access to stdout _and_ stderr, though I don't think we use that today)
* success/failure is indicated by the exit status
* hitting ^C should immediately abort the command
* being conscious about how much users have to type to run a command is a consideration

### Run `biz` in the API
* parameters come from the GraphQL query and might use existing setup like the store
* the function `biz()` should run as part of a larger setup where a long-running `biz()` runs in the background, i.e., not synchronously with the GraphQL request.
* output should be logged, though what should be logged might be different from what the CLI prints on the terminal
* success/failure of the request just indicates whether the task to run `biz()` has been accepted

For the API case, it seems to me that it would be enough to have something that wraps `biz()` in the additional logic that background execution requires, something like

```
fn run(biz: Biz, ctx: Context) -> Result <_,_> {
  // The main thing that cmd_ctx allows is to change the status (or progress) of the command
  // for very long-running things
  // This also glosses over preventing duplicate execution, and how to tell generically whether two commands are the same
  // but that would happen here, too
  let cmd_ctx = add_row_to_graphman_command_executions(..)
  let res = biz.execute(ctx, cmd_ctx);
  cmd_ctx.command_finished();
  res
}
```
This might need to be split up a bit to make the right pieces async, but it could be fairly linear code with little need for extension points.

server/graphman/src/resolvers/deployment_mutation.rs
```diff
@@ -0,0 +1,84 @@
...
+            deployment: deployment.into(),
+            delay: Duration::from_secs(delay_seconds),
+        };
+
+        let store = ctx.data::<Arc<GraphmanStore>>()?.to_owned();
+
+        let ctx = GraphmanExtensionContext::new(make_graphman_context(ctx)?);
+
+        let output = command
+            .layer(IdentifyCommand::new(CommandKind::RestartDeployment.into()))
```
lutter, line 74: Doesn't that mean we only allow one restart command at a time? It's totally fine to restart several deployments concurrently. What we want to avoid is concurrent restarts of the same deployment.

store/postgres/src/graphman_store/mod.rs
```diff
@@ -0,0 +1,161 @@
+mod models;
```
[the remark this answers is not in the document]
isum: Fixed

store/postgres/migrations/2024-07-23-100427_create_graphman_command_executions_table/up.sql
```diff
@@ -0,0 +1,20 @@
+create table public.graphman_command_executions
+(
+    id             uuid primary key,
```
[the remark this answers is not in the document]
isum: In the original implementation, it was easier to generate and use IDs before writing data to the database.

Refactored to use `bigserial`

core/graphman/src/commands/deployment/pause.rs
```diff
@@ -0,0 +1,68 @@
...
+            GraphmanError::Datastore(anyhow!("failed to locate site for '{locator}'"))
+        })?;
+
+        let Some((_, is_paused)) = conn.assignment_status(&site)? else {
+            info!(ctx.logger, "Deployment '{}' not found", locator);
+            return Ok(false);
+        };
+
+        if is_paused {
+            info!(ctx.logger, "Deployment '{}' is already paused", locator);
```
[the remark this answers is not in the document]
isum: Refactored to allow the CLI to print useful information to stdout and removed unnecessary messages

core/graphman/src/deployment_search.rs
```diff
@@ -0,0 +1,178 @@
...
+use graph::data::subgraph::DeploymentHash;
+use graph_store_postgres::command_support::catalog;
+use graph_store_postgres::connection_pool::ConnectionPool;
+use itertools::Itertools;
+
+use crate::GraphmanError;
+
+#[derive(Clone)]
+pub struct DeploymentSearch {
+    pool: ConnectionPool,
```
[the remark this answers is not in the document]
isum: Renamed the `pool` to `primary_pool` where necessary

core/graphman/src/deployment_search.rs
```diff
@@ -0,0 +1,178 @@
...
+                GraphmanError::Datastore(anyhow!("expected exactly one deployment, found {count}"))
+            })?;
+
+        Ok(deployment)
+    }
+
+    fn included_versions_filter(
+        deployments: impl Iterator<Item = Deployment>,
+        included_versions: Option<DeploymentVersionFilter>,
+    ) -> impl Iterator<Item = Deployment> {
```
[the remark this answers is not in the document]
isum: This was the existing behavior, refactored to filter in the database

Cargo.toml
```diff
@@ -24,25 +27,49 @@ repository = "https://github.com/graphprotocol/graph-node"
...
-diesel_derives = "2.1.4"
 diesel-dynamic-schema = "0.2.1"
+diesel_derives = "2.1.4"
 diesel_migrations = "2.1.0"
+graph = { path = "./graph" }
+graph-core = { path = "./core" }
+graph-store-postgres = { path = "./store/postgres" }
+graphman = { path = "./core/graphman" }
+graphman-extensions = { path = "./core/graphman_extensions" }
+graphman-primitives = { path = "./core/graphman_primitives" }
```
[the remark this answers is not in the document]
isum: After the refactor there are still 2 crates, one specific to graphman commands and one to avoid cyclic dependencies between graphman commands and postgres store.

core/graphman/src/context.rs
```diff
@@ -0,0 +1,31 @@
...
+
+use graph_store_postgres::connection_pool::ConnectionPool;
+use graph_store_postgres::NotificationSender;
+use graph_store_postgres::Store;
+use slog::Logger;
+
+/// The data that is needed to execute graphman commands.
+#[derive(Clone)]
+pub struct GraphmanContext {
+    pub pool: ConnectionPool,
```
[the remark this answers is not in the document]
isum: This type no longer exists

core/graphman/src/context.rs
```diff
@@ -0,0 +1,31 @@
...
+    pub notification_sender: Arc<NotificationSender>,
+    pub store: Arc<Store>,
+    pub logger: Logger,
+}
+
+// Allows the commands to not depend on the exact type of the context,
+// while allowing it as an argument.
```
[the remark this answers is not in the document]
isum: Fixed by the refactor

core/graphman/src/error.rs
```diff
@@ -0,0 +1,19 @@
+use thiserror::Error;
+
+#[derive(Debug, Error)]
+pub enum GraphmanError {
+    #[error("datastore error: {0}")]
```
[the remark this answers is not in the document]
isum: Updated the message

core/graphman_primitives/src/dynamic_context.rs
```diff
@@ -0,0 +1,86 @@
...
+
+use thiserror::Error;
+
+use crate::ExtensibleGraphmanContext;
+
+/// Provides a way for command extensions to share additional
+/// command execution data with other extensions.
+pub struct DynamicContext {
+    inner: HashMap<TypeId, Box<dyn Any + Send + Sync + 'static>>,
+}
```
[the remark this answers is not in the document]
isum: Fixed by the refactor

core/graphman_extensions/src/prevent_duplicate_executions.rs
```diff
@@ -0,0 +1,135 @@
...
+            let other_executions_in_progress = store
+                .any_executions_in_progress(kind.to_owned())
+                .map_err(|err| e!(Datastore, err))?;
+
+            if other_executions_in_progress {
+                return Err(e!(
+                    ExtensionFailed,
+                    anyhow!("other executions of kind '{kind}' are in progress")
+                ));
+            }
```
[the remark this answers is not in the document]
isum: This type no longer exists

core/graphman_extensions/src/prevent_duplicate_executions.rs
```diff
@@ -0,0 +1,135 @@
...
+
+            inner
+                .execute(ctx)
+                .await
+                .map_err(|err| e!(CommandFailed, anyhow!(err.into())))
```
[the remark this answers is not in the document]
isum: This type no longer exists

Cargo.toml
isum: Fixed by the refactor

core/graphman_extensions/src/handle_broken_executions.rs
isum: That's a nice idea, but after the refactor I removed that functionality for now because it's not strictly necessary

docs/graphman-graphql-api.md
isum: Removed that requirement

graph/src/env/mod.rs
isum: Removed these configuration options for now

node/src/bin/manager.rs
isum: Fixed

server/graphman/src/resolvers/deployment_mutation.rs
isum: Removed that restriction

docs/graphman-graphql-api.md
```diff
@@ -0,0 +1,212 @@
+# Graphman GraphQL API
+
+When starting the `graph-node` optionally, a graphman GraphQL server can be started, which makes it possible to execute
+graphman commands via GraphQL.
+
+To start the graphman GraphQL server, the `GRAPHMAN_SERVER_AUTH_TOKEN` environment variable should be set. The token is
+used to authenticate graphman GraphQL requests.
```
lutter, line 7: I would rewrite this as:

> The Graphman API provides functionality to manage various aspects of `graph-node` through GraphQL operations. It is only started when the environment variable `GRAPHMAN_SERVER_AUTH_TOKEN` is set. The token is used to authenticate graphman GraphQL requests. Even with the token, the server should not be exposed externally as it provides operations that an attacker can use to severely impede the functioning of an indexer. The server listens on the port `GRAPHMAN_PORT`, port `8050` by default.

docs/graphman-graphql-api.md
```diff
@@ -0,0 +1,212 @@
...
+
+Other environment variables that can be set:
```
lutter, line 9: I would just say `Environment variables to control the graphman API` here and also list the var for the token (again) That makes it easier to see all of them for reference

lutter (approved):
Very nice! I really like this and how it is put together.

core/graphman/src/execution_tracker.rs
```diff
@@ -0,0 +1,84 @@
...
+    id: ExecutionId,
+    heartbeat_stopper: Arc<Notify>,
+    store: Arc<S>,
+}
+
+impl<S> GraphmanExecutionTracker<S>
+where
+    S: GraphmanStore + Send + Sync + 'static,
+{
+    /// Creates a new execution tracker that spawns a separate process that keeps
```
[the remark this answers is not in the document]
isum: Replaced the word `process` with `background task`

docs/graphman-graphql-api.md
isum: Updated with the proposed version

docs/graphman-graphql-api.md
isum: Updated with the proposed version

core/graphman_store/src/lib.rs
```diff
@@ -0,0 +1,127 @@
+//! This crate allows graphman commands to store data in a persistent storage.
```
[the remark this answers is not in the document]
isum: If it is not a big problem, let's leave it as is for now, and if it causes some confusion when new commands are added, we will move it to the `graph` module.

Issue #5514, opened 2024-06-27
[Bug] handle reverted eth_calls on declared eth_calls


Shiyasmohd:
### Bug report

Declared eth_calls executes before eth_calls on handlers triggered. But when declared eth_calls get reverted, subgraph get stuck in syncing.

Repo that can reproduce this error - [Link](https://github.com/Shiyasmohd/declared-eth-call-err)
Deployment ID: `QmfH5xCioLUqd1zM2b2QGE6Br5uZdks2tx2A1NPB1dZQSK`

Subgraph that sync when declared eth_call is removed
Deployment ID: `QmaWD3fz3Y6PZdEmZ9QQ79fkVwTzh49JBqLcbzG85wDMrr`

### Relevant log output

```Shell
Error: Ethereum node returned an error for an eth_call: RPC error: Error { code: ServerError(-32015), message: "VM execution error.", data: Some(String("Reverted ERC721URIStorage: URI query for nonexistent token")) }
```

### IPFS hash

QmfH5xCioLUqd1zM2b2QGE6Br5uZdks2tx2A1NPB1dZQSK

### Some information to help us out

- [ ] Tick this box if this bug is caused by a regression found in the latest release.
- [ ] Tick this box if this bug is specific to the hosted service.
- [X] I have searched the issue tracker to make sure this issue is not a duplicate.

### OS information

None

incrypto32:
@Shiyasmohd can i close it? this is non-issue now right?

Shiyasmohd:
Yes. Closing this issue.

PR #5563, opened 2024-07-23
Fix declared calls revert

incrypto32:
Closes [#5514](https://github.com/graphprotocol/graph-node/issues/5514)


lutter:
I am not convinced this is the right approach for fixing this - the code now assumes that any error in `EthereumAdapter.call_and_cache` is a call revert; but the code (in `EthereumAdapter::call`) goes through great length to identify reverts and returns `Ok(Retval::Null)` which the rest of the code handles appropriately.

Could #5514 just be a case of an RPC returning an error that we don't recognize as a revert?

incrypto32:
Ah, thanks for the context, just seeing those error message parsing to detect revert, i'll try that out

PR #5565, opened 2024-07-23
disable genesis validation by default

mangas:


zorancv (approved):
Nice to have the feature optional.

PR #5566, opened 2024-07-23
Store timestamp when marking subgraph as synced

encalypto:
We migrate `synced` from a boolean value to a nullable `synced_at` timestamp on `subgraphs.subgraph_deployment` and `unused_deployments`. The default timestamp used in the migration is the unix epoch, `1970-01-01 00:00:00+00`.

Note that in the down migration we skip removal of the `DEFAULT false` on `unused_deployments.`


lutter (approved):
Nice! I've been wanting to have that for a long time. Would be nice to do the same thing for `failed` at some point.

PR #5568, opened 2024-07-25
timeout when trying to get net_identifiers at startup

mangas:
1. If an error occurs during startup net_identifier calls, just ignore an use a default value 
2. If validation is enabled, this will cause the adapter to eventually fail because whatever genesis the adapter returns won't match the default (0x00) genesis value. 

This change prevents the graph-node from failing to start due to broken adapters but if genesis validation is enabled it should eventually update the store

### DIFF
--- a/node/src/manager/commands/config.rs
+++ b/node/src/manager/commands/config.rs
@@ -177,6 +177,7 @@ pub async fn provider(
         registry,
         metrics,
         Arc::new(NoopIdentValidator),
+        false,
     )
     .await?;
     let network: ChainId = network.into();
--- a/graph/src/components/adapter.rs
+++ b/graph/src/components/adapter.rs
@@ -290,7 +290,6 @@ impl<T: NetIdentifiable + Clone + 'static> ProviderManager<T> {
             .unwrap_or_default()
     }
 
-    #[cfg(debug_assertions)]
     pub async fn mark_all_valid(&self) {
         for (_, status) in self.inner.status.iter() {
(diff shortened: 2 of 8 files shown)


zorancv (approved):
So we don't try to enforce specific genesis block, right? In that case it LGTM.

mangas:
> So we don't try to enforce specific genesis block, right? In that case it LGTM.

Yeah unfortunately the erroring part can cause an outage easily so I think the best course is to try to keep running and then if something fails it should happen for the specific chain

zorancv:
> > So we don't try to enforce specific genesis block, right? In that case it LGTM.
> 
> Yeah unfortunately the erroring part can cause an outage easily so I think the best course is to try to keep running and then if something fails it should happen for the specific chain

I guess more conservative approach would be to allow failures for the 'sloppy' chains and be strict for the 'good' ones. However not sure if even that is possible given that firehose can't provide it, for instance... At any rate merging this seems very reasonable in order to solve the pressing issue. and it's already strictly better than the original behaviour.

node/src/manager/commands/config.rs
```diff
@@ -177,6 +177,7 @@ pub async fn provider(
         registry,
         metrics,
         Arc::new(NoopIdentValidator),
+        false,
```
zorancv, line 180: Don't you want it dependant on the config in graphman too?

mangas: only for run, the others don't depend on the ProviderManager component

PR #5569, opened 2024-07-29
only start substreams if no other block investor is available

mangas:


zorancv (approved):
substream ingestors now seems to actually be created later.

PR #5578, opened 2024-07-31
Sanitise columns in SQL

zorancv:


store/postgres/src/relational/ddl_tests.rs
```diff
@@ -221,8 +221,6 @@ fn generate_ddl() {
     let il = IndexList::mock_thing_index_list();
     let layout = test_layout(THING_GQL);
     let sql = layout.as_ddl(Some(il)).expect("Failed to generate DDL");
-    println!("SQL: {}", sql);
```
mangas, line 224: was this intentionally removed?

mangas (approved):

store/postgres/src/relational/ddl_tests.rs
zorancv: yeah. used to be one of my debugging statements...

Issue #5535, opened 2024-07-15
[Bug] Subgraph with endBlock on all datasources keeps synching with unfiltered firehose request


sduchesneau:
### Bug report

When running a subgraph with a single datasource bounded by an 'endBlock', the subgraph will continue indexing forever passed that endBlock.

This is a costly operation if the graph-node is using firehose, because the filter will be EMPTY.

Either we should stop the indexing of the subgraph at that point, or just the blockstream, or at least, make sure that the firehose filters are set to "not match anything".

To reproduce:

* On a local graph-node, take an existing subgraph like this one: 
https://github.com/graphprotocol/graph-tooling/blob/main/examples/ethereum-basic-event-handlers/subgraph.yaml

* Apply these few changes:
```
+++ b/examples/ethereum-basic-event-handlers/subgraph.yaml
@@ -1,4 +1,4 @@
-specVersion: 0.0.5
+specVersion: 0.0.9
 description: Gravatar for Ethereum
 repository: https://github.com/graphprotocol/graph-tooling
 schema:
@@ -6,9 +6,11 @@ schema:
 dataSources:
   - kind: ethereum/contract
     name: Gravity
-    network: test
+    network: mainnet
     source:
-      address: '0x5FbDB2315678afecb367f032d93F642f64180aa3'
+      address: '0x2e645469f354bb4f5c8a05b3b30a929361cf77ec'
+      startBlock: 6175244
+      endBlock: 6175300
       abi: Gravity
     mapping:
       kind: ethereum/events
```

* stop it after a while and restart it.
* Observe this log:
```Blockstream disconnected, connecting, provider_err_count: 0, cursor: (...), start_block: 6243715, endpoint_uri: firehose, provider: firehose, deployment: Qmc5kz5vA4kXWDHmrvMXkxyTgEUX4xHNsNRWX5pnjSVcST, sgd: 1, subgraph_id: Qmc5kz5vA4kXWDHmrvMXkxyTgEUX4xHNsNRWX5pnjSVcST, component: FirehoseBlockStream```

The issue with the firehose filter is here:
https://github.com/graphprotocol/graph-node/blob/v0.35.1/chain/ethereum/src/adapter.rs#L316-L318

A way to get around the issue would be to apply a "never-matching" filter, example by having a "log filter" with a single address of "0x0000000000000..."

But I'm wondering if the solution should be around the handling of the "end block" situation ?

### Some information to help us out

[...]

PR #5583, opened 2024-08-03
Stop subgraphs passing max endBlock

shuaibbapputty:
Closes https://github.com/graphprotocol/graph-node/issues/5535.

Previously, subgraphs continued scanning even when all the data sources had passed the end block.

This update finds the maximum end block across the data sources and stops the subgraph from scanning once the `max_end_block` is reached.

``` from runner test
Aug 03 02:51:00.634 DEBG Starting or restarting subgraph, sgd: 8, test: end_block, subgraph_id: QmPL5J13p2652oit5umZZPQiNAvfV4E2KXkLNayguL9cJd
Aug 03 02:51:00.634 DEBG Starting block stream, sgd: 8, test: end_block, subgraph_id: QmPL5J13p2652oit5umZZPQiNAvfV4E2KXkLNayguL9cJd
Aug 03 02:51:00.635 INFO Stopping subgraph as we reached maximum endBlock, current_block: 9, max_end_block: 8, sgd: 8, test: end_block, subgraph_id: QmPL5J13p2652oit5umZZPQiNAvfV4E2KXkLNayguL9cJd

```


shuaibabubakker:
@fordN @leoyvens would appreciate a review on this.😊

incrypto32:
Nice! I think it'd be a bit more cleaner if we just set the exisiting `stop_block` value instead of creating a new `max_end_block` And also stop the subgraph on start up by checking stop block early on, just like you did for max_end_block

alex-pakalniskis:
@shuaibbapputty bump 😄

core/src/subgraph/instance_manager.rs
```diff
@@ -331,6 +331,13 @@ impl<S: SubgraphStore> SubgraphInstanceManager<S> {
             })
             .collect();
 
+        let max_end_block: Option<BlockNumber> = if manifest.data_sources.len() == end_blocks.len()
```
incrypto32, line 334: A subgraph can also have offchain data sources, so we cant just compare it like this since offchain datasources dont have an end block

incrypto32 (changes requested):

shuaibbapputty:
Sorry for the delay @alex-pakalniskis 

As per my discussion with @incrypto32 on discord, I will address the review changes and update the PR asap. Thankyou!

core/src/subgraph/instance_manager.rs
```diff
@@ -331,6 +331,19 @@ impl<S: SubgraphStore> SubgraphInstanceManager<S> {
             })
             .collect();
 
+        // We can set `max_end_block` to the maximum of `end_blocks` and stop the subgraph
+        // at that point only when there are no dynamic data sources present and offchain
+        // data sources exist. This is because:
+        // - Dynamic data sources do not have a defined `end_block`, so we can't determine
+        //   when to stop processing them.
+        // - Offchain data sources might still require processing beyond the end block of
+        //   onchain data sources, necessitating the continuation of the subgraph.
```
incrypto32, line 340:
```suggestion
        // We can set `max_end_block` to the maximum of `end_blocks` and stop the subgraph
        // at that point only when there are no dynamic data sources present and offchain
        // data sources exist. This is because:
        // - Dynamic data sources do not have a defined `end_block`, so we can't determine
        //   when to stop processing them.
        // - Offchain data sources might still require processing beyond the end block of
        //   onchain data sources, necessitating the continuation of the subgraph.
```

incrypto32 (approved):

Issue #5197, opened 2024-02-08
[Bug] Change subgraph runner tests to not graft onto subgraph head


lutter:
Some of the subgraph runner tests graft directly onto the subgraph head. PR #5135 made that very verboten, which would make those tests fail. For expediency, the PR just disables the check that we graft at least reorg_threshold blocks back for debug builds.

The right fix for this would be to change the subgraph runner tests to set a small reorg threshold and to graft only outside that reorg threshold so that the check can be enabled for debug builds, too.

PR #5584, opened 2024-08-06
Enable graft validation checks for debug

shuaibbapputty:
Closes https://github.com/graphprotocol/graph-node/issues/5197

Fix graft validation checks

- Removed the conditional compilation for reorg threshold checks in debug builds

This will be good to go as we have `reorg_threshold` set to 0 on debug builds.
 https://github.com/graphprotocol/graph-node/blob/b0c89526fd932393f3e43350ee664404eb858040/graph/src/env/mod.rs#L238-L241

### DIFF
--- a/graph/src/data/subgraph/mod.rs
+++ b/graph/src/data/subgraph/mod.rs
@@ -500,13 +500,7 @@ impl Graft {
             // The graft point must be at least `reorg_threshold` blocks
             // behind the subgraph head so that a reorg can not affect the
             // data that we copy for grafting
-            //
-            // This is pretty nasty: we have tests in the subgraph runner
-            // tests that graft onto the subgraph head directly. We
-            // therefore skip this check in debug builds and only turn it on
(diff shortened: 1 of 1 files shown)


shuaibabubakker:
@lutter @incrypto32 would appreciate a review of this PR.

lutter:
Interesting. This used to fail CI; do you have any idea where the reorg threshold is set to 0 when we run tests?

shuaibabubakker:
Yep, this used to fail before. The reason for the change in behavior is that at the time the issue was opened, the `cfg!(debug_assertions)` macro wasn't being used to set the reorg threshold.

And now, while running tests (which are typically in debug mode), the cfg!(debug_assertions) macro is being applied, and the reorg threshold is set to 0 by default. unless explicitly set by `"ETHEREUM_REORG_THRESHOLD"` env var

lutter (approved):

PR #5590, opened 2024-08-08
fix: ErrorPolicy::Allow flow should not add an indexing error

dwerner:
fix https://github.com/graphprotocol/indexer/issues/910


dwerner:
Chatting with @leoyvens, the existence of the error in the response should be preserved, but we should make it `attestable` somehow.

PR #5595, opened 2024-08-12
Subgraph composition : Add Integration Tests

zorancv:
The branch is intended to be a base for further PRs of the subgraph composition feature. It adds integration tests.


tests/tests/integration_tests.rs
```diff
@@ -439,6 +517,35 @@ async fn test_eth_api(ctx: TestContext) -> anyhow::Result<()> {
...
+            {  "number": "0" },
+            {  "number": "1" },
+            { "number": "2" },
+            {  "number": "3" },
+            {  "number": "4" },
+            {  "number": "5" },
+            { "number": "6" },
+            {  "number": "7" },
+            {  "number": "8" },
+            {  "number": "9" },
```
lutter, line 534: The formatting here is off, and #5599 needs to correct that. Would be better to adjust the formatting here

zorancv: It's again changed in [#5602](https://github.com/graphprotocol/graph-node/pull/5602), so neither is actually needed.

lutter (approved):

PR #5600, opened 2024-08-13
IPFS: use IPFS Gateway API

### LINKED ISSUE #5573: Use IPFS gateway API
Gateway API docs: https://docs.ipfs.tech/concepts/ipfs-gateway/

isum:
This PR implements a new IPFS client that connects and sends requests to the IPFS Gateway. The existing IPFS RPC API client has been refactored to reuse new code.

The `--ipfs` CLI argument accepts both IPFS Gateway URLs and RPC API URLs, and the appropriate client is automatically detected and created.

Closes #5573

### Todos

- [x] Clean up tests affected by the changes
- [x] Add unit tests for new clients


graph/src/ipfs/gateway_client.rs
```diff
@@ -0,0 +1,702 @@
...
+            logger: logger.to_owned(),
+            test_request_timeout: TEST_REQUEST_TIMEOUT,
+        })
+    }
+
+    pub fn into_boxed(self) -> Box<dyn IpfsClient> {
+        Box::new(self)
+    }
+
+    async fn send_test_request(&self) -> anyhow::Result<()> {
```
mangas, line 79: Not sure what this is trying to achieve, is this only used for tests or is this a check for something specific? If it's only used for tests, could it be marked with cfg(test) ?

mangas (approved):
LGTM overall, I think this should be tested in the integration cluster before merging

graph/src/ipfs/gateway_client.rs
isum: Test requests are used to check which type of API a server address supports (RPC or Gateway) in order to initialize the appropriate client.

lundgrenalex:
@isum it is not working with RPC API URLs, I moved to ipfs-api-gateway, and it works. I used Kubo v0.30.0

PR #5605, opened 2024-08-16
Subgraph composition: TriggersAdapter  changes

zorancv:


incrypto32:
This PR introduces a wrapper around the TriggersAdapter to support subgraph triggers originating from other subgraphs. Previously, the TriggersAdapter was designed to be chain-specific, handling triggers for individual chains. However, with subgraph data sources, we need a higher-level adapter capable of managing cross-subgraph triggers. This new wrapper enables fetching entity triggers from source subgraphs in a way that is independent of any specific chain, streamlining the trigger logic for broader subgraph interoperability.

PR #5606, opened 2024-08-19
Subgraph composition: Use block cache to get blocks for subgraph triggers

incrypto32:


graph/src/blockchain/client.rs
```diff
@@ -11,13 +11,17 @@ use anyhow::anyhow;
 // Substreams only requires the FirehoseEndpoints.
 #[derive(Debug)]
 pub enum ChainClient<C: Blockchain> {
-    Firehose(FirehoseEndpoints),
+    Firehose(FirehoseEndpoints, Option<C::Client>),
```
mangas, line 14: I think we had this code before and was removed, did you intentionally add the client back in?

store/postgres/src/chain_store.rs
```diff
@@ -2094,12 +2250,15 @@ impl ChainStoreTrait for ChainStore {
...
                 let lookup_fut = async move {
                     let res = this.blocks_from_store(hashes).await;
-                    BlocksLookupResult(Arc::new(res))
+                    BlocksLookupResult::ByHash(Arc::new(res))
                 };
                 let lookup_herd = self.lookup_herd.cheap_clone();
                 let logger = self.logger.cheap_clone();
-                let (BlocksLookupResult(res), _) =
-                    lookup_herd.cached_query(hash, lookup_fut, &logger).await;
+                //TODO(krishna): Add comments explaining the return value of cached_query
```
mangas, line 2257: resolve todo

store/postgres/src/chain_store.rs
```diff
@@ -2094,12 +2250,15 @@ impl ChainStoreTrait for ChainStore {
...
+                let res = match lookup_herd.cached_query(hash, lookup_fut, &logger).await {
+                    (BlocksLookupResult::ByHash(res), _) => res,
+                    _ => unreachable!(),
```
mangas, line 2260: is this actually unreachable? Probably an error would be better just so we don't panic in case of a logic error

graph/src/blockchain/client.rs
incrypto32: Will remove it

store/postgres/src/chain_store.rs
incrypto32: Its actually unreachable. Since earlier in the code `lookup_fut` is defined as

```rust
 let lookup_fut = async move {
                    let res = this.blocks_from_store(hashes).await;
                    BlocksLookupResult::ByHash(Arc::new(res))
                };
```

So it always returns `ByHash`. But i'll just go ahead and return an error since thats defenetly cleaner

store/postgres/src/chain_store.rs
incrypto32: Addressed

store/postgres/src/chain_store.rs
incrypto32: Addressed in latest commit

graph/src/blockchain/client.rs
incrypto32: Removed

mangas (approved):

PR #5610, opened 2024-08-19
Store `synced_at_block_number` when a deployment syncs

encalypto:
This is a follow up to #5566 

We use both `INTEGER` and `NUMERIC` types in the database to match how other block numbers are stored in the same table.

### DIFF
--- a/store/postgres/migrations/2024-08-14-205601_store_synced_at_block/up.sql
+++ b/store/postgres/migrations/2024-08-14-205601_store_synced_at_block/up.sql
@@ -1,0 +1,2 @@
+ALTER TABLE subgraphs.subgraph_deployment ADD COLUMN synced_at_block_number NUMERIC;
+ALTER TABLE unused_deployments            ADD COLUMN synced_at_block_number INTEGER;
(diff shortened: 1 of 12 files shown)


store/postgres/migrations/2024-08-14-205601_store_synced_at_block/up.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE subgraphs.subgraph_deployment ADD COLUMN synced_at_block_number NUMERIC;
```
lutter, line 1: It would be better to make that an `int4`, too. The fact that `latest_ethereum_block_number` is `numeric` is more a historical accident than anything we should continue doing for other places where we store block numbers. In an ideal world, we'd change al the columns that store block numbers and use `numeric` to `int4` but that might be a bit delicate and nobody's done that, obviously.

store/postgres/src/deployment.rs
```diff
@@ -731,15 +732,25 @@ pub fn state(conn: &mut PgConnection, id: DeploymentHash) -> Result<DeploymentSt
...
-pub fn set_synced(conn: &mut PgConnection, id: &DeploymentHash) -> Result<(), StoreError> {
+pub fn set_synced(
+    conn: &mut PgConnection,
+    id: &DeploymentHash,
+    block_ptr: BlockPtr,
+) -> Result<(), StoreError> {
     use subgraph_deployment as d;
 
+    // Work around a Diesel issue with serializing BigDecimals to numeric
+    let number = format!("{}::int4", block_ptr.number);
```
zorancv, line 743: interesting

zorancv (approved):
Looks good!

PR #5614, opened 2024-08-20
Subgraph composition sql more entities

zorancv:
Add support for querying more entities with a single SQL query.

### DIFF
--- a/graph/src/blockchain/block_stream.rs
+++ b/graph/src/blockchain/block_stream.rs
@@ -1,5 +1,5 @@
 use crate::blockchain::SubgraphFilter;
-use crate::data_source::subgraph;
+use crate::data_source::{subgraph, CausalityRegion};
 use crate::substreams::Clock;
 use crate::substreams_rpc::response::Message as SubstreamsMessage;
 use crate::substreams_rpc::BlockScopedData;
@@ -433,9 +433,19 @@ async fn scan_subgraph_triggers<C: Blockchain>(
     }
 }
 
+#[derive(Debug)]
+pub enum EntitySubgraphOperation {
+    Create,
+    Modify,
+    Delete,
+}
+
+#[derive(Debug)]
 pub struct EntityWithType {
+    pub entity_op: EntitySubgraphOperation,
     pub entity_type: EntityType,
     pub entity: Entity,
+    pub vid: i64,
 }
 
 async fn get_entities_for_range(
@@ -445,32 +455,12 @@ async fn get_entities_for_range(
     from: BlockNumber,
     to: BlockNumber,
 ) -> Result<BTreeMap<BlockNumber, Vec<EntityWithType>>, Error> {
-    let mut entities_by_block = BTreeMap::new();
-
-    for entity_name in &filter.entities {
-        let entity_type = schema.entity_type(entity_name)?;
-
-        let entity_ranges = store.get_range(&entity_type, from..to)?;
-
-        for (block_number, entity_vec) in entity_ranges {
-            let mut entity_vec = entity_vec
-                .into_iter()
-                .map(|e| EntityWithType {
-                    entity_type: entity_type.clone(),
-                    entity: e,
-                })
-                .collect();
-
-            entities_by_block
-                .entry(block_number)
-                .and_modify(|existing_vec: &mut Vec<EntityWithType>| {
-                    existing_vec.append(&mut entity_vec);
-                })
-                .or_insert(entity_vec);
-        }
-    }
-
-    Ok(entities_by_block)
+    let entity_types: Result<Vec<EntityType>> = filter
+        .entities
+        .iter()
+        .map(|name| schema.entity_type(name))
+        .collect();
+    Ok(store.get_range(entity_types?, CausalityRegion::ONCHAIN, from..to)?)
 }
 
 impl<C: Blockchain> TriggersAdapterWrapper<C> {
--- a/store/postgres/src/block_range.rs
+++ b/store/postgres/src/block_range.rs
@@ -132,36 +132,52 @@ impl<'a> QueryFragment<Pg> for BlockRangeUpperBoundClause<'a> {
     }
 }
 
+#[derive(Debug, Clone, Copy)]
+pub enum BoundSide {
+    Lower,
+    Upper,
+}
+
 /// Helper for generating SQL fragments for selecting entities in a specific block range
 #[derive(Debug, Clone, Copy)]
 pub enum EntityBlockRange {
-    Mutable(BlockRange), // TODO: check if this is a proper type here (maybe Range<BlockNumber>?)
+    Mutable((BlockRange, BoundSide)),
     Immutable(BlockRange),
 }
 
 impl EntityBlockRange {
-    pub fn new(table: &Table, block_range: std::ops::Range<BlockNumber>) -> Self {
+    pub fn new(
+        immutable: bool,
+        block_range: std::ops::Range<BlockNumber>,
+        bound_side: BoundSide,
+    ) -> Self {
         let start: Bound<BlockNumber> = Bound::Included(block_range.start);
         let end: Bound<BlockNumber> = Bound::Excluded(block_range.end);
         let block_range: BlockRange = BlockRange(start, end);
-        if table.immutable {
+        if immutable {
             Self::Immutable(block_range)
         } else {
-            Self::Mutable(block_range)
+            Self::Mutable((block_range, bound_side))
         }
     }
 
-    /// Output SQL that matches only rows whose block range contains `block`.
+    /// Outputs SQL that matches only rows whose entities would trigger a change
+    /// event (Create, Modify, Delete) in a given interval of blocks. Otherwise said
+    /// a block_range border is contained in an interval of blocks. For instance
+    /// one of the following:
+    /// lower(block_range) >= $1 and lower(block_range) <= $2
+    /// upper(block_range) >= $1 and upper(block_range) <= $2
+    /// block$ >= $1 and block$ <= $2
     pub fn contains<'b>(&'b self, out: &mut AstPass<'_, 'b, Pg>) -> QueryResult<()> {
         out.unsafe_to_cache_prepared();
         let block_range = match self {
-            EntityBlockRange::Mutable(br) => br,
+            EntityBlockRange::Mutable((br, _)) => br,
             EntityBlockRange::Immutable(br) => br,
         };
         let BlockRange(start, finish) = block_range;
 
         self.compare_column(out);
-        out.push_sql(" >= ");
+        out.push_sql(">= ");
         match start {
             Bound::Included(block) => out.push_bind_param::<Integer, _>(block)?,
             Bound::Excluded(block) => {
@@ -170,9 +186,9 @@ impl EntityBlockRange {
             }
             Bound::Unbounded => unimplemented!(),
         };
-        out.push_sql(" AND ");
+        out.push_sql(" and");
         self.compare_column(out);
-        out.push_sql(" <= ");
+        out.push_sql("<= ");
         match finish {
             Bound::Included(block) => {
                 out.push_bind_param::<Integer, _>(block)?;
@@ -186,7 +202,12 @@ impl EntityBlockRange {
 
     pub fn compare_column(&self, out: &mut AstPass<Pg>) {
         match self {
-            EntityBlockRange::Mutable(_) => out.push_sql(" lower(block_range) "),
+            EntityBlockRange::Mutable((_, BoundSide::Lower)) => {
+                out.push_sql(" lower(block_range) ")
+            }
+            EntityBlockRange::Mutable((_, BoundSide::Upper)) => {
+                out.push_sql(" upper(block_range) ")
+            }
             EntityBlockRange::Immutable(_) => out.push_sql(" block$ "),
         }
     }
--- a/store/postgres/src/relational.rs
+++ b/store/postgres/src/relational.rs
@@ -28,6 +28,7 @@ use diesel::{connection::SimpleConnection, Connection};
 use diesel::{
     debug_query, sql_query, OptionalExtension, PgConnection, QueryDsl, QueryResult, RunQueryDsl,
 };
+use graph::blockchain::block_stream::{EntitySubgraphOperation, EntityWithType};
 use graph::blockchain::BlockTime;
 use graph::cheap_clone::CheapClone;
 use graph::components::store::write::{RowGroup, WriteChunk};
@@ -57,8 +58,8 @@ use std::time::{Duration, Instant};
 
 use crate::relational::value::{FromOidRow, OidRow};
 use crate::relational_queries::{
-    ConflictingEntitiesData, ConflictingEntitiesQuery, FindChangesQuery, FindDerivedQuery,
-    FindPossibleDeletionsQuery, ReturnedEntityData,
+    ConflictingEntitiesData, ConflictingEntitiesQuery, EntityDataExt, FindChangesQuery,
+    FindDerivedQuery, FindPossibleDeletionsQuery, ReturnedEntityData,
 };
 use crate::{
     primary::{Namespace, Site},
@@ -75,7 +76,7 @@ use graph::prelude::{
     QueryExecutionError, StoreError, StoreEvent, ValueType, BLOCK_NUMBER_MAX,
 };
 
-use crate::block_range::{BLOCK_COLUMN, BLOCK_RANGE_COLUMN};
+use crate::block_range::{BoundSide, BLOCK_COLUMN, BLOCK_RANGE_COLUMN};
 pub use crate::catalog::Catalog;
 use crate::connection_pool::ForeignServer;
 use crate::{catalog, deployment};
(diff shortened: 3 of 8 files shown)


store/postgres/src/relational_queries.rs
```diff
@@ -586,6 +558,49 @@ impl EntityData {
...
+    pub vid: i64,
+}
+
+impl Ord for EntityDataExt {
+    fn cmp(&self, other: &Self) -> Ordering {
+        let ord = self.block_number.cmp(&other.block_number);
+        if ord != Ordering::Equal {
+            ord
+        } else {
+            // TODO: check if this and the next cmp for string match postgress C localle
```
zorancv, line 581: Rust string comparison is basically the same as memcmp from libc (byte per byte binary comparison). The LC_COLLATE=C in Postgress should be the same.

store/postgres/src/writable.rs
```diff
@@ -1217,6 +1233,16 @@ impl Queue {
...
 
+    fn get_range(
+        &self,
+        entity_types: Vec<EntityType>,
+        causality_region: CausalityRegion,
+        block_range: Range<BlockNumber>,
+    ) -> Result<BTreeMap<BlockNumber, Vec<EntityWithType>>, StoreError> {
+        self.store
+            .get_range(entity_types, causality_region, block_range)
+    }
```
lutter, line 1244: Maybe I lost track of what's happening, but didn't we talk about that this shouldn't go into the `WritableStore` and instead be part of a separate read interface? Sadly, I can't find my comment to that effect anymore; might be a victim of rebasing/force pushing

store/postgres/src/relational_queries.rs
```diff
@@ -586,6 +558,49 @@ impl EntityData {
...
+    pub data: serde_json::Value,
+    #[diesel(sql_type = Integer)]
+    pub block_number: i32,
+    #[diesel(sql_type = Text)]
+    pub id: String,
+    #[diesel(sql_type = BigInt)]
+    pub vid: i64,
+}
+
+impl Ord for EntityDataExt {
```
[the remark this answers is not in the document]
zorancv: The order in the query is exactly that one, but this comparison is needed in order to process the types of operations for each entity change here: https://github.com/graphprotocol/graph-node/blob/b892793cc09cc61929fcee611899f7a38fb260c0/store/postgres/src/relational.rs#L563-L606

store/postgres/src/writable.rs
zorancv: Yeah, it was addressed in a different PR: https://github.com/graphprotocol/graph-node/pull/5544#discussion_r1870924333 . Unfortunately not merged/rebased downstream just yet.

zorancv: All should be up to date now.

graph/src/blockchain/block_stream.rs
```diff
@@ -441,32 +451,12 @@ async fn get_entities_for_range(
...
-
-            entities_by_block
-                .entry(block_number)
-                .and_modify(|existing_vec: &mut Vec<EntityWithType>| {
-                    existing_vec.append(&mut entity_vec);
-                })
-                .or_insert(entity_vec);
-        }
+        entity_types.push(entity_type);
     }
```
lutter, line 458: It would be more idiomatic to write this as `let entity_types = filter.entities.iter().map(|name| schema.entity_type(name)).collect()?`

graph/src/blockchain/block_stream.rs
```diff
@@ -429,9 +429,19 @@ async fn scan_subgraph_triggers<C: Blockchain>(
...
 }
 
+#[derive(Debug)]
+pub enum EntitySubgraphOperation {
```
lutter, line 433: I am not sure if this makes sense, but might be better to stick the entity into the operation, i.e. to have `Create(Entity)`, `Modify(Entity)`, `Delete(Entity)`; that way, users of this enum are forced to consider each of these cases, and it would make it easy to, e.g., change to `Delete(Id)`, i.e., not load the entire entity if that's useful.

But I don't understand how all of composition fits together, so I am not sure if this suggestions is really an improvement.

store/postgres/src/block_range.rs
```diff
@@ -135,33 +135,37 @@ impl<'a> QueryFragment<Pg> for BlockRangeUpperBoundClause<'a> {
 /// Helper for generating SQL fragments for selecting entities in a specific block range
 #[derive(Debug, Clone, Copy)]
 pub enum EntityBlockRange {
-    Mutable(BlockRange), // TODO: check if this is a proper type here (maybe Range<BlockNumber>?)
+    Mutable((BlockRange, bool)),
```
lutter, line 138: There should be a comment explaining what the `bool` is for

Actually, since this is to indicate whether we want to compare with the upper or lower of the block range, it might be worth having an enum `{ Lower, Upper }` for this

store/postgres/src/block_range.rs
```diff
@@ -170,9 +174,9 @@ impl EntityBlockRange {
             }
```
lutter, line 174: This could just push the bind param `block + 1`

store/postgres/src/block_range.rs
```diff
@@ -135,33 +135,37 @@ impl<'a> QueryFragment<Pg> for BlockRangeUpperBoundClause<'a> {
...
+        if immutable {
             Self::Immutable(block_range)
         } else {
-            Self::Mutable(block_range)
+            Self::Mutable((block_range, is_uppper_range))
         }
     }
 
     /// Output SQL that matches only rows whose block range contains `block`.
     pub fn contains<'b>(&'b self, out: &mut AstPass<'_, 'b, Pg>) -> QueryResult<()> {
```
lutter, line 159: I find this very confusing. It would help if the comment described what the SQL text is that can be generated by this method

store/postgres/src/relational.rs
```diff
@@ -518,21 +519,105 @@ impl Layout {
...
-            .get_results::<EntityData>(conn)
+    ) -> Result<BTreeMap<BlockNumber, Vec<EntityWithType>>, StoreError> {
+        let mut tables = vec![];
+        let mut et_map: HashMap<String, Arc<EntityType>> = HashMap::new();
+        for et in entity_types {
+            tables.push(self.table_for_entity(&et)?.as_ref());
+            et_map.insert(et.to_string(), Arc::new(et));
+        }
+        let mut entities: BTreeMap<BlockNumber, Vec<EntityWithType>> = BTreeMap::new();
+        let lower_vec = FindRangeQuery::new(&tables, causality_region, false, block_range.clone())
```
lutter, line 533: It would be helpful if there was a comment about what `lower_vec` and `upper_vec` represent

store/postgres/src/relational.rs
```diff
@@ -518,21 +519,105 @@ impl Layout {
...
+            let vid = ede.vid;
+            let ewt = EntityWithType {
+                entity_op,
+                entity_type,
+                entity,
+                vid,
+            };
+            Ok((ewt, block))
+        };
+        while lower_now.is_some() || upper_now.is_some() {
```
lutter, line 563: The body of this is pretty intricate. There should be a comment about what the strategy here is and how it works.

store/postgres/src/relational.rs
```diff
@@ -518,21 +519,105 @@ impl Layout {
...
+            let (ewt, block) = if lower_now.is_some() {
```
lutter, line 564: This whole chain of `if .. else` would probably be clearer if it was turned into a `match (lower_now, upper_now) { .. }` as that's more symmetric than nested ifs

store/postgres/src/relational_queries.rs
```diff
@@ -2008,37 +2023,84 @@ impl<'a, Conn> RunQueryDsl<Conn> for FindQuery<'a> {}
...
+            // the immutable entities don't have upper range and also can't be modified or deleted
+            if !(self.is_upper_range && table.immutable) {
+                if first {
+                    first = false;
+                } else {
+                    out.push_sql("\nunion all\n");
+                }
+
+                // Generate
+                //    select '..' as entity, to_jsonb(e.*) as data, block$ as block_number
```
lutter, line 2067: For mutable tables, what is `block_number` set to? (Looking at the code below, I think I can work it out, but would be nice to have it in the comment)

store/postgres/src/relational_queries.rs
```diff
@@ -2008,37 +2023,84 @@ impl<'a, Conn> RunQueryDsl<Conn> for FindQuery<'a> {}
...
+                } else {
+                    self.mut_range.contains(&mut out)?;
+                }
+            }
+        }
+
+        if first {
+            // In case we have only immutable entities, the upper range will not create any
+            // select statement. So here we have to generate an empty SQL statement.
+            out.push_sql("select 1");
```
lutter, line 2098: Does this work? If we have clauses in the query, it returns `(text, jsonb, int)`, but this fallback just returns `(int)` and is therefore the wrong shape

store/postgres/src/relational_queries.rs
```diff
@@ -2008,37 +2023,84 @@ impl<'a, Conn> RunQueryDsl<Conn> for FindQuery<'a> {}
...
+            if !(self.is_upper_range && table.immutable) {
+                if first {
+                    first = false;
+                } else {
+                    out.push_sql("\nunion all\n");
+                }
+
+                // Generate
+                //    select '..' as entity, to_jsonb(e.*) as data, block$ as block_number
+                //      from schema.table e where id = $1
```
lutter, line 2068: Is this left over from something else? I don't see a clause `id = $1` being generated.

lutter (changes requested):
I don't fully follow how this loads entities and classifies them as create/modify/delete, partly because the interplay of mutable/immutable and `is_upper_range` is very intricate. I am not sure if code changes can clarify that (would be great); if not, this needs more comments explaining what is going on.

graph/src/blockchain/block_stream.rs
zorancv: That worked baring the trailing error propagation operator.

store/postgres/src/block_range.rs
zorancv: I wasn't able to make the compiler borrower happy...

store/postgres/src/block_range.rs
zorancv: Added some sensible comment. Hope it's clear now.

store/postgres/src/relational.rs
zorancv: I [did](https://github.com/graphprotocol/graph-node/pull/5614/commits/4136889f06a96b0257dd4b639c75f202b31f2663?diff=split&w=1) it with a couple of matches. Not sure that it's more readable, but for sure one level of block nesting is spared and is probably more idiomatic too.

store/postgres/src/relational_queries.rs
zorancv: I believe I tested it. The intention was to return an empty set. Now when I run it on the command line it returns one row, so maybe diesel silently substituted wrong types with empty rows, but its a bad style for sure. Needs fixing.

zorancv:
@lutter I believe that the comments in the `find_range()` make the matching algorithm for detecting the type of the operation more clear. Not sure if there is an easy way to make it simpler. Also I feel that splitting mutable and immutable cases won't improve things much, as the immutable case is a special case (and simpler one) of the mutable one.

graph/src/blockchain/block_stream.rs
zorancv: I guess here the naming is poor. The enum `EntitySubgraphOperation` is a kind of operation more akin to the C style enums. Where the struct just bellow it `EntityWithType` contains the whole 'entity event' with all the fields including this operation kind and the entity itself.

store/postgres/src/relational_queries.rs
zorancv: [Addressed](https://github.com/graphprotocol/graph-node/pull/5614/commits/5e4df3863532eceb8316c3ac28b148c86aeef34e) it and added a test for that case.

graph/src/blockchain/block_stream.rs
zorancv: I did some renaming which I believe is closer to the semantics. I done it in a separate [PR](https://github.com/graphprotocol/graph-node/pull/5728/files) in order to save on rebasing.

store/postgres/src/relational.rs
```diff
@@ -518,21 +519,131 @@ impl Layout {
...
+                vid,
+            };
+            Ok((ewt, block))
+        };
+
+        // The algorithm advances simultaneously entities from both lower_vec and upper_vec and tries
+        // to match entities that have entries in both vectors for a particular block. The match is
+        // successfull if an entry in one array has same values for the number of the block, entity
+        // name and the entity id. The comparison operation over the EntityDataExt fullfils that check.
+        // In addition to that, it also helps to order the elements so the algorith can detect if one
```
lutter, line 580: I don't quite understand the sentence `it also helps ..`. I think it should also be mentioned somewhere that this algorithm relies on the fact that `lower_vec` and `upper_vec` are ordered by `(block, entity type, entity id)` which allows for mergesort-like processing of the two vecs.

store/postgres/src/relational.rs
```diff
@@ -518,21 +519,131 @@ impl Layout {
...
+                            lower = lower_now.unwrap_or(&EntityDataExt::default()).clone();
+                            (ewt, block)
+                        }
+                        std::cmp::Ordering::Equal => {
+                            let (ewt, block) = transform(lower, EntitySubgraphOperation::Modify)?;
+                            // advance both lower_vec and upper_vec pointers
+                            lower_now = lower_iter.next();
+                            lower = lower_now.unwrap_or(&EntityDataExt::default()).clone();
+                            upper_now = upper_iter.next();
+                            upper = upper_now.unwrap_or(&EntityDataExt::default()).clone();
```
lutter, line 610: What happens here if `lower_vec` contains multiple entries for the same entity? Advancing `upper_now` might move us to the next entity, and so we think that `lower_now` which has another version of the entity we just looked at will look like a create.

lutter: I wonder whether a better strategy for all this might be to throw `upper_vec` and `lower_vec` into one vec, sort that by `(block, entity type, entity id)` and scan the resulting vec. You'd emit an `EntityWithType` if `block` is in the desired block range, and the operation will be determined with

* if previous entry had different type/id, it's a create
* if next entry has different type/id, it's a delete
* otherwise it's an update

That sorta requires a 3 element window to pass over the Vec, but since it's a Vec, that could be done with just using indices, like `vec[i-1]`, `vec[i]`, `vec[i+1]` with some special care for `i=0` and `i=vec.len()-1`, i.e. best to hide those intricacies in a helper functions.

lutter (changes requested):

lutter:
One other thing I am wondering about with this strategy: if you are looking for changes in `[10,20)` and you find an entity version with a range of `[11, 200)` will this properly check if there is a version `[200,_)` to tell whether this is a delete or an update?

zorancv:
> One other thing I am wondering about with this strategy: if you are looking for changes in `[10,20)` and you find an entity version with a range of `[11, 200)` will this properly check if there is a version `[200,_)` to tell whether this is a delete or an update?

We don't reason about the situation around 200 since it's > 20. We only check if there is another entry ending at 11 in order to decide if it's creation or modification at 11.

store/postgres/src/relational.rs
zorancv: Not sure it it's possible to have multiple modifications of the entity in a block. That would entail a range of a type [10,10). In that case we might end up with wrong modification since we would have ranges like [3,10), [10,10), [10,15) and there is no guarantee that [10, 10) will be before [10,15). I guess if that's the case the DB request has to be extended with a 4. sorting criteria (for the lower_vec the end of the range, and for the upper_vec the start I believe)... Still if there are more than two changes in a block the order of the inner ones is not guaranteed to be correct, only the final result for a block...

zorancv: Not sure if merging the two vectors would make the algorithm simpler. It would entail having double entries for each entity that is modified. Also likely we would need a flag if it's a start or end of a range in order to deduce if there is a creation or deletion for single entries, also to avoid false creations at the first entry and false deletion at the last one. And to deduce the proper entry for update. Probably with that flag and 3 element window one can create such algorithm, but I believe it would have at least 6 checks if not more, where the current one has 5. Not to mention the additional complexity of the start/end boundaries in the helper function and the additional complexity needed for the SQL query.

store/postgres/src/block_range.rs
zorancv: Yeah. Enum seems reasonable here.

store/postgres/src/relational.rs
```diff
@@ -518,21 +519,131 @@ impl Layout {
...
+            BoundSide::Lower,
+            block_range.clone(),
+        )
+        .get_results::<EntityDataExt>(conn)
+        .optional()?
+        .unwrap_or_default();
+        // collect all entities that have their 'upper(block_range)' attribute in the
+        // interval of blocks defined by the variable block_range. For the immutable
+        // entities no entries are returned.
+        let upper_vec =
```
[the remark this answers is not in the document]
zorancv: I'll just copy this text.

store/postgres/src/relational.rs
zorancv: I did rewording. Hope it's clear now.

store/postgres/src/relational.rs
```diff
@@ -518,21 +519,131 @@ impl Layout {
...
+        // The algorithm advances simultaneously entities from both lower_vec and upper_vec and tries
+        // to match entities that have entries in both vectors for a particular block. The match is
+        // successfull if an entry in one array has same values for the number of the block, entity
+        // name and the entity id. The comparison operation over the EntityDataExt fullfils that check.
+        // In addition to that, it also helps to order the elements so the algorith can detect if one
+        // side of the range exists and the other is missing. That way a creation and deletion are
+        // deduced. For immutable entities the entries in upper_vec are missing hence they are considered
+        // having a lower bound at particular block and upper bound at infinity.
+        while lower_now.is_some() || upper_now.is_some() {
+            let (ewt, block) = match (lower_now.is_some(), upper_now.is_some()) {
```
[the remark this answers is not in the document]
zorancv: That's beautiful! Did it.

store/postgres/src/relational_queries.rs
```diff
@@ -586,6 +558,48 @@ impl EntityData {
...
+        } else {
+            let ord = self.entity.cmp(&other.entity);
+            if ord != Ordering::Equal {
+                ord
+            } else {
+                self.id.cmp(&other.id)
+            }
+        }
+    }
+}
```
lutter, line 589: One thing that confuses me about the sort order `(block_number, entity type, entity id)` is that entries for different entities can be interleaved, like if there are two entities `a` and `b`, you might end up with `[ (2, a), (2,b), (3, a), (3, b)]` and since `find_range` processes the entries in the vecs in order, can't you end up comparing an entry for `a` with one for `b`?

zorancv: You have to keep in mind that matching is always done between elements in the different vectors, and never between elements in the same one. Also a missing element in `upper_vec` is a `creation` and missing element in `lower_vec` is `deletion`, where a match (a presence of the equivalent respective values in both vectors) is a `modification`. And finally the two pairs of cases where there is either no matching element on one side or no remaining element at all on that same side, are in essence equivalent and produce the same result. Advancements of iterators are done on the already evaluated element(s). 

Probably a simple example would be more explanatory. 
Imagine you have elements in `lower_vec` like this: 
`[ (2, a), (2, c), (3, a) (3, b), (3, c) ]`, 

and in the `upper_vec` something like: 
`[ (2, a), (2, b), (3, a), (3, c), (4, a) ]`,

the strategy will first find a match of `(2, a)` in both vectors so modification of `a` in block `2`,  then `(2, b) `is missing in `lower_vec` so deletion of `b` in `2`, then `(2, c)` is missed in `upper_vec` so creation there, then `(3, a)` matched again so another modification, `(3, b)` only in `lower_vec` - creation, `(3, c)` on both sides - modification, `(4, a)` only in `upper_vec` - deletion.

Looking again at the comments at the start of the big while loop I tried to enter the same content as the first paragraph here, but also referred to some other aspects and that might had made it less readable. Should I try to improve it?

lutter (approved):
Had a discussion with Zoran, and since I have been unable to come up with counterexamples to support my worries, this is good to go.

PR #5616, opened 2024-08-21
Use correct store when loading indexes for graft base

encalypto:
Previously, we were trying to load indexes from the store corresponding to the subgraph being deployed. This was the wrong shard, resulting in deployment failures when a subgraph is on a different shard from its graft based.

This updates the call to use the correct store.


zorancv (approved):
Thanks for fixing this!

yash251:
Any timeline for when this will be merged and released?

encalypto:
@yash251 
> Any timeline for when this will be merged and released?

Confirming this has been released to production, and deploying grafts now works correctly.

PR #5627, opened 2024-08-30
Add "substreams/triggers" kind for substreams datasource mapping

YaroShkvorets:
We had a discussion with @abourget on Slack that `substreams/graph-entities` kind doesn't make sense for trigger-based SbS.
This PR adds new kind `substreams/triggers`.
Data source mapping of this kind requires both `file` and `handler` defined in the manifest


sduchesneau (approved):
LGTM!

mangas:
What is the difference between this new type and the old type conceptually? Just that this one requires a handler that is optional on the other type? Is that worth maintaining (and documenting, etc) yet another type? 

Has support for this been added to graph-cli? That should be the first step for changes like this since ideally any validation should happen during compilation time. Unfortunately for the time being this validation has to be implemented in both places. 

All that being said, I don't really see a benefit to introducing this definition, plus it should be properly documented and validated by all components if it is to be created.

mangas (changes requested):
See comment

PR #5630, opened 2024-09-03
Subgraph composition: Use block ptrs instead of full blocks in blockstream

incrypto32:


graph/src/blockchain/types.rs
```diff
@@ -330,6 +341,174 @@ impl From<BlockPtr> for BlockNumber {
...
+        let s = s.trim_start_matches("0x");
+        i32::from_str_radix(s, 16).map_err(serde::de::Error::custom)
+    } else {
+        i32::from_str(&s).map_err(serde::de::Error::custom)
+    }
+}
+
+#[derive(Clone, PartialEq, Eq, Hash, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct BlockPtrExt {
```
mangas, line 360: I don't know if this is a nitpick but I think we should prefer more explicit names, in this case ExternalBlockPtr if that's what it means. Would also be great to have a comments on why and how it's used because this terminology is very overloaded in some parts of the code.

incrypto32: Agreed! i meant ExtendedBlockPtr, i'll go with the full name

incrypto32: Resolved in latest commit

PR #5631, opened 2024-09-03
Subgraph Composition: Cache blocks processed by firehose block stream

incrypto32:


graph/src/blockchain/firehose_block_stream.rs
```diff
@@ -359,11 +364,46 @@ async fn process_firehose_response<C: Blockchain, F: FirehoseMapper<C>>(
         .await
         .context("Mapping block to BlockStreamEvent failed")?;
 
+    if let BlockStreamEvent::ProcessBlock(block, _) = &event {
+        info!(logger, "Inserting block to cache"; "block_number" => block.block.number(), "block_hash" => format!("{:?}", block.block.hash()));
+
+        let start_time = Instant::now();
+
+        let result = chain_store
```
mangas, line 372: Isn't this the exact responsibility of the Block Ingestor? I also don't think this should be in here because there will always be 1 stream per subgraph so this would be amplifying any write attempts or any backoffs, on top of adding more time to processing. Caching blocks is already done on the ingestor, if we need more data that's where I think it should go or if we need a healing mechanism it seems like something different.

incrypto32: Agreed, Im closing this PR as upon evaluating, this caching is no longer required since the use of cached block ptr's which is done in a later PR

PR #5636, opened 2024-09-09
Subgraph Composition: Entity Ops Detection in Handlers

incrypto32:


runtime/wasm/src/module/mod.rs
```diff
@@ -76,7 +76,7 @@ impl ToAscPtr for subgraph::TriggerData {
         heap: &mut H,
         gas: &GasCounter,
     ) -> Result<AscPtr<()>, HostExportError> {
-        asc_new(heap, &self.entity.sorted_ref(), gas).map(|ptr| ptr.erase())
+        asc_new(heap, &self.entity, gas).map(|ptr| ptr.erase())
```
mangas, line 79: Is sorting no longer needed? Will that break existing behavior ?

mangas: or both

mangas (approved):
Just the one comment, otherwise LGTM

runtime/wasm/src/module/mod.rs
incrypto32: it is indeed the only entry point, calling `to_asc_ptr` on `subgraph::TriggerData` is the only way in which the mapping is going to get the entity trigger, which internally calls the `EntityWithType.to_asc_obj` which gets the sorted_ref() of the entity as you mentioned happens in `external.rs` 

Hope that answers that.

PR #5639, opened 2024-09-09
"count" aggregations fix 

### LINKED ISSUE #5634: [Bug] Single "count" aggregation fails
### Bug report

If I have a single "count" aggregation:
```
type Block @entity(timeseries: true) {
  id: Int8!
  timestamp: Timestamp!
  hash: Bytes!
  number: Int!
}

type Stats @aggregation(intervals: ["hour", "day"], source: "Block") {
  id: Int8!
  timestamp: Timestamp!

  count: Int! @aggregate(fn: "count")
}

```

graph-node fails with invalid SQL expression error:
```
postgres-defi    | 2024-09-07 20:45:32.788 UTC [91] ERROR:  syntax error at or near "from" at character 205
postgres-defi    | 2024-09-07 20:45:32.788 UTC [91] STATEMENT:  insert into "sgd1"."stats_hour"(id, timestamp, block$, "count") select max(id) as id, timestamp, $3, count(*) as "count" from (select id, date_bin('3600s', timestamp, 'epoch'::timestamptz) as timestamp,  from "sgd1"."block" where "sgd1"."block".timestamp >= $1 and "sgd1"."block".timestamp < $2 order by "sgd1"."block".timestamp) data group by timestamp
```
Note "," before "from" that shouldn't be there.

Subgraph example: https://github.com/YaroShkvorets/common-subgraphs/tree/main/aggregation-broken

### Relevant log output

_No response_

### IPFS hash

_No response_

_No response_

### OS information

None

YaroShkvorets:
Fixes #5634 

- [x] fix extra comma
- [x] add test case
- [x] refactor sql query construction to avoid same bug in other cases

### DIFF
--- a/store/postgres/src/relational/rollup.rs
+++ b/store/postgres/src/relational/rollup.rs
@@ -336,20 +336,14 @@ impl<'a> RollupSql<'a> {
         if with_block {
             write!(w, ", $3")?;
         }
-        if !self.dimensions.is_empty() {
-            write!(w, ", ")?;
-            write_dims(self.dimensions, w)?;
-        }
+        write_dims(self.dimensions, w)?;
         comma_sep(self.aggregates, false, w, |w, agg| agg.aggregate("id", w))?;
         let secs = self.interval.as_duration().as_secs();
         write!(
             w,
             " from (select id, date_bin('{secs}s', timestamp, 'epoch'::timestamptz) as timestamp"
         )?;
-        if !self.dimensions.is_empty() {
-            write!(w, ", ")?;
-            write_dims(self.dimensions, w)?;
-        }
+        write_dims(self.dimensions, w)?;
         let agg_srcs: Vec<&str> = {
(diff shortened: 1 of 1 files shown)


lutter (changes requested):
Nice catch! It seems that with these changes `comma_sep` is always called with `first == false` and we could therefore get rid of that argument and simplify the code a little.

YaroShkvorets:
Yes, I'll drop it.

YaroShkvorets:
Should be good now.

lutter (approved):
Thanks for the PR!

matthewdarwin:
@lutter can you merge it as well?  @YaroShkvorets doesn't have merge permissions.

lutter:
> @lutter can you merge it as well? @YaroShkvorets doesn't have merge permissions.

Woops .. sorry, overlooked that. Merged now.

PR #5645, opened 2024-09-16
graphman: Add simpler `chain rewind` command to delete blocks

encalypto:
This is a simpler version of #5486, which adds a command to delete blocks past a certain point:

```
$ graphman chain rewind sepolia --block-hash 0xabc123 --block-number 123
```

Note that deployments are not rewound here.

### DIFF
--- a/store/postgres/src/chain_store.rs
+++ b/store/postgres/src/chain_store.rs
@@ -1123,6 +1123,32 @@ mod data {
             }
         }
 
+        pub(super) fn delete_blocks_after(
+            &self,
+            conn: &mut PgConnection,
+            chain: &str,
+            block: i64,
+        ) -> Result<usize, Error> {
+            match self {
+                Storage::Shared => {
+                    use public::ethereum_blocks as b;
+
+                    diesel::delete(b::table)
+                        .filter(b::network_name.eq(chain))
+                        .filter(b::number.gt(block))
+                        .execute(conn)
+                        .map_err(Error::from)
+                }
+                Storage::Private(Schema { blocks, .. }) => {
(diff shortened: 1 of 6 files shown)


store/postgres/src/chain_store.rs
```diff
@@ -1870,6 +1896,29 @@ impl ChainStore {
...
+
+            self.storage
+                .delete_blocks_after(conn, &self.chain, block_ptr_to.number as i64)?;
+
+            // Update the chain head
+            diesel::update(
+                ethereum_networks::table.filter(ethereum_networks::name.eq(&self.chain)),
+            )
+            .set((
+                ethereum_networks::head_block_number.eq(block_ptr_to.number as i64),
```
mangas, line 1914: I think there is somewhere we keep the firehose cursor, I think this table has a firehose cursor too, that should be set to empty string.

store/postgres/src/chain_store.rs
```diff
@@ -1123,6 +1123,32 @@ mod data {
...
+                Storage::Shared => {
+                    use public::ethereum_blocks as b;
+
+                    diesel::delete(b::table)
+                        .filter(b::network_name.eq(chain))
+                        .filter(b::number.gt(block))
+                        .execute(conn)
+                        .map_err(Error::from)
+                }
+                Storage::Private(Schema { blocks, .. }) => {
```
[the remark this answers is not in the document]
mangas: I think it does work because the ChainStore should already always have the right pool when it gets created. If you already tested that's great otherwise maybe rewind something in staging

store/postgres/src/chain_store.rs
mangas: same here regarding tests

PR #5651, opened 2024-09-29
Add json-rpc api support for dropping subgraphs

mangas:
Add json-rpc method for dropping subgraphs. Using this endpoint will remove all the subgraphs for the deployment_id and will delete the underlying data. `docker/bin/drop` is a helper script to use the new endpoint.

Adding some sane defaults as discussed: 

```
 ./docker/bin/drop Qmd3vU6y6pxxXPrvVWRZMN9soNB8AFQCEnqPa9jMSZZDEG
Dropping deployment Qmd3vU6y6pxxXPrvVWRZMN9soNB8AFQCEnqPa9jMSZZDEG
{"jsonrpc":"2.0","error":{"code":1,"message":"subgraph registrar error with store: store error: multiple subgraphs exist fo
r this hash, please remove them before trying to drop"},"id":"1"}
```

```
./docker/bin/remove y
{"jsonrpc":"2.0","result":{"type":"Null"},"id":"1"}
```

```
./docker/bin/drop Qmd3vU6y6pxxXPrvVWRZMN9soNB8AFQCEnqPa9jMSZZDEG
Dropping deployment Qmd3vU6y6pxxXPrvVWRZMN9soNB8AFQCEnqPa9jMSZZDEG
{"jsonrpc":"2.0","result":{"type":"Null"},"id":"1"}%
```

### DIFF
--- a/core/src/subgraph/registrar.rs
+++ b/core/src/subgraph/registrar.rs
@@ -455,6 +455,14 @@ where
         Ok(())
     }
 
+    async fn remove_deployment(&self, id: &DeploymentHash) -> Result<(), SubgraphRegistrarError> {
+        self.store.drop_subgraph(id)?;
+
+        debug!(self.logger, "Removing deployment(s)"; "hash" => id.to_string());
+
+        Ok(())
+    }
+
     /// Reassign a subgraph deployment to a different node.
     ///
     /// Reassigning to a nodeId that does not match any reachable graph-nodes will effectively pause the
--- a/store/postgres/src/subgraph_store.rs
+++ b/store/postgres/src/subgraph_store.rs
@@ -1351,6 +1351,23 @@ impl SubgraphStoreTrait for SubgraphStore {
         })
     }
 
+    fn drop_subgraph(&self, hash: &DeploymentHash) -> Result<(), StoreError> {
+        let deployments = self.subgraphs_for_deployment_hash(hash)?;
+        for (deployment, _) in deployments.into_iter() {
+            self.remove_subgraph(
+                SubgraphName::new(deployment.clone())
+                    .map_err(|_| StoreError::DeploymentNotFound(deployment))?,
+            )?;
+        }
+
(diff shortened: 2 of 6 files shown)


mangas:
> Perhaps explore Indexer Agent-based solution to avoid `graphman drop` danger zone?

Not sure I understand the argument here. This is a private endpoint, it's as dangerous as letting someone add subgraphs unboundedly and fill up the disks, taking down the graph-node. It is also a basic operational task that currently cannot be automated because it's not exposed in any api.

store/postgres/src/subgraph_store.rs
```diff
@@ -1351,6 +1351,23 @@ impl SubgraphStoreTrait for SubgraphStore {
         })
     }
 
+    fn drop_subgraph(&self, hash: &DeploymentHash) -> Result<(), StoreError> {
+        let deployments = self.subgraphs_for_deployment_hash(hash)?;
+        for (deployment, _) in deployments.into_iter() {
+            self.remove_subgraph(
+                SubgraphName::new(deployment.clone())
+                    .map_err(|_| StoreError::DeploymentNotFound(deployment))?,
+            )?;
```
incrypto32, line 1360: Dont this delete both the pending and current version of a subgraph that belongs to the same name as `hash` provided? Thought we are moving away from that behaviour?

core/src/subgraph/registrar.rs
```diff
@@ -455,6 +455,14 @@ where
         Ok(())
     }
 
+    async fn remove_deployment(&self, id: &DeploymentHash) -> Result<(), SubgraphRegistrarError> {
+        self.store.drop_subgraph(id)?;
+
+        debug!(self.logger, "Removing deployment(s)"; "hash" => id.to_string());
+
+        Ok(())
+    }
```
incrypto32, line 464: Was the debug logs supposed to be before the  `self.store.drop_subgraph(id)?;` ?

PR #5672, opened 2024-10-17
Use the new Firehose info endpoint

### LINKED ISSUE #5673: Use new firehose & substreams /info path
Check firehose endpoints on startup for detail level, and genesis block. FIrehose has a new endpoint for reporting the detail level back to graph-node `/info`. 

References:
- firehose block detail https://github.com/graphprotocol/graph-node/pull/5571/files
- provider startup validations https://github.com/graphprotocol/graph-node/pull/5517

isum:
This PR introduces the use of the new Firehose Info endpoint. The primary use case for this is provider validation. As a result, the existing provider manager responsible for validating providers has been refactored to support multiple checks. It has also been optimized to rely less on locks, and the number of allocations has been greatly reduced. Also, a new extended blocks check has been added, which requires providers to support extended block details if enabled.

Closes #5673

### Todos

- [x] Add more tests where necessary
- [x] Test new features locally with a running graph-node
- [x] Extend graphman with extended block checks


chain/ethereum/src/network.rs
```diff
@@ -109,15 +117,14 @@ impl EthereumNetworkAdapters {
         let provider = ProviderManager::new(
             Logger::root(Discard, o!()),
             vec![(chain_id.clone(), adapters)].into_iter(),
-            Arc::new(NoopIdentValidator),
+            &[],
```
mangas, line 120: from usability perspective I think I'd rather have a type you pass here like NoChecks or a AndChecks(Box<[]>) or OneOf(Box<[]>), in this case I think we'd only need the first 2 obviosuly. I think this probably more of a nitpick though

graph/src/env/mod.rs
```diff
@@ -311,6 +319,10 @@ impl EnvVars {
             genesis_validation_enabled: inner.genesis_validation_enabled.0,
             genesis_validation_timeout: Duration::from_secs(inner.genesis_validation_timeout),
             graphman_server_auth_token: inner.graphman_server_auth_token,
+            firehose_require_extended_blocks_for_chains:
+                Self::firehose_require_extended_blocks_for_chains(
```
mangas, line 323: I think this should be the opposite, it should IMO require extended by default with the flag to authorize "downgrading" since the default option should be the safest option possible

mangas (approved):
Really good work on this PR. Super clean! :tada:

mangas (approved):

mangas (approved):

PR #5674, opened 2024-10-17
Graphman restart: Handle cases where multiple subgraph names reference a deployment ID

### LINKED ISSUE #5670: Improve Graphman restart: handle cases where multiple subgraph names reference the deployment ID

isum:
There was an issue that prevented `graphman restart` from running if a deployment ID was referenced by multiple subgraph names. This PR fixes that.

Closes #5670

### DIFF
--- a/core/graphman/src/commands/deployment/pause.rs
+++ b/core/graphman/src/commands/deployment/pause.rs
@@ -39,12 +39,11 @@ pub fn load_active_deployment(
 ) -> Result<ActiveDeployment, PauseDeploymentError> {
     let mut primary_conn = primary_pool.get().map_err(GraphmanError::from)?;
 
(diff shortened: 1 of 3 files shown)


zorancv:
It actually pauses and resumes the first of the two deployments. The second one is not accessible. Both names would point to the whichever happens to be first. If that's desired behaviour then is fine IMO.

zorancv (approved):

lutter:
In `node/src/manager/deployment.rs`, there's a similar helper `locate_unique`. I don't think this duplication is a big deal, just wanted to point that out.

isum:
@lutter The code duplication will eventually be eliminated once more commands are added to Graphman GraphQL API.

Issue #5530, opened 2024-07-11
Subgraph using Aggregations⁠ failing randomly


itsjerryokolo:
### Bug report

This [subgraph](https://github.com/team-usd/ltusd-index) fails randomly with the error below. 
```
"Failed to transact block operations: store error: duplicate key value violates unique constraint \"transfer_stats_hour_id_key\"",
```
The same block `6284464`  which it failed in (QmQZadt6wd8wo8U6opsVnjpAjx2Ysh3iLiPXuydJ65eqT3) is successfully processed in a new deployment (QmYktZYrbaENqm9nS8kvBxnzsRUyh4ueuQ679mYwMdNLpz)

### Some information to help us out

- [ ] Tick this box if this bug is caused by a regression found in the latest release.
- [ ] Tick this box if this bug is specific to the hosted service.
- [X] I have searched the issue tracker to make sure this issue is not a duplicate.

### OS information

None

george-openformat:
Running into the same issue:
[subgraph](https://github.com/open-format/metrics-subgraph) | [playground](https://api.studio.thegraph.com/query/82634/metrics-arbitrum-sepolia/v0.0.4)

First deployment failed on block 64040656 after deploying a new version failed on block 64603832

silvercondor:
getting same error on multiple subgraphs. 

note that it can also happen for daily aggregations

tinypell3ts:
I'm also getting the same error in my subgraph. 

```
ERRO Subgraph failed with non-deterministic error: Failed to transact block operations: store error: duplicate key value violates unique constraint "transaction_stat_hour_id_key", retry_delay_s: 3732, attempt: 40, sgd: 5389, subgraph_id: QmZFm5y76cnjiF5TBfCCufVyR9sQAaWV6Ygm9Fg3cTUSed, component: SubgraphInstanceManager
```

tsudmi:
Running into the same issue. Any updates on this?

gperezalba:
Same issue here. Also would be nice to include a working example of the mapping in https://github.com/graphprotocol/graph-node/blob/master/docs/aggregations.md

lutter:
Digging into one such example on Avalanche, this seems to be caused by blocks having the same timestamps. For example, on Avalanche blocks 51482738 and 51482739 both have the timestamp `0x67040580` which corresponds to `2024-10-07 16:00:00+00` That would trigger a rollup for the hour starting at 15:00.

[...]

PR #5675, opened 2024-10-18
Do not repeat a rollup after restart in some corner cases

lutter:
When graph-node is restarted, we need to determine for subgraphs with aggregations when the last rollup was triggered to ensure aggregations get filled without gaps or duplication.

The code used the `block_time` column in the PoI table for that but that is not correct as the PoI table only records blocks and times for which the  subgraph actually has writes. When the subgraph scans through a largish number of blocks without changes, we only update the head pointer but also do rollups as aggregation intervals pass. Because of that, we might perform a rollup without a corresponding entry in the PoI table.
    
With this change, we actually find the maximum timestamp from all aggregation tables to tell us when the last rollup was triggered as that data reflects when rollups happened accurately.
    
For safety, this new behavior can be turned off by setting  `GRAPH_STORE_LAST_ROLLUP_FROM_POI=true` to return to the old buggy behavior in case the new behavior causes some other unexpected problems.
    
Fixes https://github.com/graphprotocol/graph-node/issues/5530

### DIFF
--- a/graph/src/blockchain/types.rs
+++ b/graph/src/blockchain/types.rs
@@ -435,3 +435,9 @@ impl ToSql<Timestamptz, Pg> for BlockTime {
         <Timestamp as ToSql<Timestamptz, Pg>>::to_sql(&self.0, out)
     }
 }
+
+impl FromSql<Timestamptz, Pg> for BlockTime {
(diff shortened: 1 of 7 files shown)


incrypto32 (approved):
LGTM

PR #5677, opened 2024-10-18
Create SECURITY.md

martyb1666:


lutter:
This is just boilerplate that has zero `graph-node` specific information. Feel free to reopen if you have content that is pertinent to the security of running `graph-node`

PR #5678, opened 2024-10-18
Add graphman unassign/reassign commands to graphman graphql api

shiyasmohd:


server/graphman/src/resolvers/deployment_mutation.rs
```diff
@@ -65,4 +70,45 @@ impl DeploymentMutation {
...
+    ) -> Result<Response> {
+        let ctx = GraphmanContext::new(ctx)?;
+        let deployment = deployment.try_into()?;
+
+        unassign::run(&ctx, &deployment)?;
+
+        Ok(Response::new(
+            true,
+            format!("Unassigned {}", deployment.as_str()),
+        ))
```
isum, line 88: We can replace this with `EmptyResponse` because the general expectation in APIs is that if a request is successful, it's done what it was supposed to do.

isum (commented):
The code looks good, but there are a few things to consider before I can complete my review:

- I would suggest rethinking the approach with warnings, so that we are not hiding potentially important details from users. It might be a good idea to break the commands into smaller pieces to allow more flexibility in the interfaces that assemble the commands _(it is important not to allow assembling wrong commands)_;
- When I introduced the GraphQL API, we agreed that all implemented commands should have tests, and there are no tests for the new GraphQL commands;
- The primary use case for extracting the Graphman commands into a separate crate and breaking them into smaller pieces was to share the functionality between the CLI and the GraphQL API, but this PR does not share the extracted code with the CLI;

server/graphman/src/resolvers/deployment_mutation.rs
```diff
@@ -81,4 +93,42 @@ impl DeploymentMutation {
...
+    ) -> Result<ReassignResponse> {
+        let ctx = GraphmanContext::new(ctx)?;
+        let deployment = deployment.try_into()?;
+        let node = NodeId::new(node.clone()).map_err(|()| anyhow!("illegal node id `{}`", node))?;
+        reassign::run(&ctx, &deployment, &node)?;
+
+        let mirror = catalog::Mirror::primary_only(ctx.primary_pool);
+        let count = mirror.assignments(&node)?.len();
+        if count == 1 {
+            let warning_msg = format!("warning: this is the only deployment assigned to '{}'. Are you sure it is spelled correctly?",node.as_str());
```
isum, line 126: We can remove the "warning" prefix from the message because the type only expects warnings. And it does not seem right to ask a question in the message.

```suggestion
            let warning_msg = format!("This is the only deployment assigned to '{}'. Please make sure that the node ID is spelled correctly.",node.as_str());
```

server/graphman/tests/deployment_mutation.rs
```diff
@@ -390,3 +390,197 @@ fn graphql_cannot_remove_subgraph_with_invalid_name() {
...
+
+        let subgraph_node_id = send_graphql_request(
+            json!({
+                "query": r#"query Deployment {
+                    deployment {
+                        info(deployment: { hash: "subgraph_1" }) {
+                            nodeId
+                        }
+                    }
+                }"#
```
isum, line 432: Named query does not seem to be necessary.

```suggestion
                "query": r#"{
                    deployment {
                        info(deployment: { hash: "subgraph_1" }) {
                            nodeId
                        }
                    }
                }"#
```

isum (approved):
It looks good, but I would suggest you use the extracted functionality of reassign and unassign in the CLI modules to remove some code duplication. The goal is to eventually share as much functionality as possible between the CLI and the GraphQL API.

core/graphman/src/commands/deployment/reassign.rs
```diff
@@ -0,0 +1,117 @@
...
+
+    #[error(transparent)]
+    Common(#[from] GraphmanError),
+}
+
+#[derive(Clone, Debug)]
+pub enum ReassignResult {
+    EmptyResponse,
+    CompletedWithWarnings(Vec<String>),
+}
```
isum, line 42: Just a note. It is not very common for a Rust function to return `T::EmptyResponse` on success _(it makes sense for GraphQL types because of GraphQL style and conventions)_. Maybe a better name for this variant is `ReassignResult::Ok`, but that's not really important right now

isum (approved):

PR #5685, opened 2024-10-22
Add graphman create/remove commands to graphql api

shiyasmohd:


server/graphman/src/resolvers/deployment_mutation/remove.rs
```diff
@@ -0,0 +1,26 @@
+use crate::resolvers::context::GraphmanContext;
```
isum, line 1: Very minor, and you can ignore it, but the general style is to have `std` imports first, then third-party imports, and then `super` & `crate` imports in separate groups.

server/graphman/src/resolvers/deployment_mutation/remove.rs
```diff
@@ -0,0 +1,26 @@
...
+    let primary_pool = ctx.primary_pool.get().map_err(GraphmanError::from)?;
+    let mut catalog_conn = catalog::Connection::new(primary_pool);
+
+    let name = match SubgraphName::new(name) {
+        Ok(name) => name,
+        Err(_) => {
+            return Err(GraphmanError::Store(anyhow!(
+                "Subgraph name must contain only a-z, A-Z, 0-9, '-' and '_'"
+            ))
+            .try_into()?)
```
isum, line 18: You can just use `.into()` instead of `.try_into()?`.

server/graphman/src/resolvers/deployment_mutation/create.rs
```diff
@@ -0,0 +1,25 @@
+use crate::resolvers::context::GraphmanContext;
```
isum, line 1: Very minor, and you can ignore it, but the general style is to have `std` imports first, then third-party imports, and then `super` & `crate` imports in separate groups.

server/graphman/src/resolvers/deployment_mutation/create.rs
```diff
@@ -0,0 +1,25 @@
...
+    let primary_pool = ctx.primary_pool.get().map_err(GraphmanError::from)?;
+    let mut catalog_conn = catalog::Connection::new(primary_pool);
+
+    let name = match SubgraphName::new(name) {
+        Ok(name) => name,
+        Err(_) => {
+            return Err(GraphmanError::Store(anyhow!(
+                "Subgraph name must contain only a-z, A-Z, 0-9, '-' and '_'"
+            ))
+            .try_into()?)
```
isum, line 18: You can just use `.into()` instead of `.try_into()?`.

isum (commented):
Could you please add some tests for the new GraphQL mutations?

isum (approved):

PR #5695, opened 2024-11-05
Subgraph composition : TriggersAdapter wrapper

incrypto32:
This PR introduces a wrapper around the TriggersAdapter to support subgraph triggers from other subgraphs. Previously, the TriggersAdapter was chain-specific, handling triggers only for individual chains. With subgraph data sources, we need a higher-level adapter to fetch entity triggers from source subgraphs without depending on any specific chain.

This PR lays the groundwork for enabling the new subgraph data source type. The changes include:

Allow subgraph datasource in manifest
A wrapper for TriggersAdapter
A wrapper for TriggerFilter
A new trigger type specific to subgraph data sources

### DIFF
--- a/chain/ethereum/src/chain.rs
+++ b/chain/ethereum/src/chain.rs
@@ -3,15 +3,16 @@ use anyhow::{Context, Error};
 use graph::blockchain::client::ChainClient;
 use graph::blockchain::firehose_block_ingestor::{FirehoseBlockIngestor, Transforms};
 use graph::blockchain::{
-    BlockIngestor, BlockTime, BlockchainKind, ChainIdentifier, TriggersAdapterSelector,
+    BlockIngestor, BlockTime, BlockchainKind, ChainIdentifier, TriggerFilterWrapper,
+    TriggersAdapterSelector,
 };
 use graph::components::network_provider::ChainName;
-use graph::components::store::DeploymentCursorTracker;
+use graph::components::store::{DeploymentCursorTracker, ReadStore};
 use graph::data::subgraph::UnifiedMappingApiVersion;
 use graph::firehose::{FirehoseEndpoint, ForkStep};
 use graph::futures03::compat::Future01CompatExt;
 use graph::prelude::{
-    BlockHash, ComponentLoggerConfig, ElasticComponentLoggerConfig, EthereumBlock,
+    BlockHash, ComponentLoggerConfig, DeploymentHash, ElasticComponentLoggerConfig, EthereumBlock,
     EthereumCallCache, LightEthereumBlock, LightEthereumBlockExt, MetricsRegistry,
 };
 use graph::schema::InputSchema;
@@ -61,6 +62,7 @@ use crate::{BufferedCallCache, NodeCapabilities};
 use crate::{EthereumAdapter, RuntimeAdapter};
 use graph::blockchain::block_stream::{
     BlockStream, BlockStreamBuilder, BlockStreamError, BlockStreamMapper, FirehoseCursor,
+    TriggersAdapterWrapper,
 };
 
 /// Celo Mainnet: 42220, Testnet Alfajores: 44787, Testnet Baklava: 62320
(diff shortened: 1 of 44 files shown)


graph/src/blockchain/block_stream.rs
```diff
@@ -256,6 +308,144 @@ impl<C: Blockchain> BlockWithTriggers<C> {
...
+    }
+
+    pub async fn triggers_in_block(
+        &self,
+        logger: &Logger,
+        block: C::Block,
+        filter: &C::TriggerFilter,
+    ) -> Result<BlockWithTriggers<C>, Error> {
+        self.adapter.triggers_in_block(logger, block, filter).await
+    }
```
incrypto32, line 370: Both `scan_triggers` and `triggers_in_block` is completely implemented in the upcoming PR's

graph/src/blockchain/block_stream.rs
```diff
@@ -256,6 +308,144 @@ impl<C: Blockchain> BlockWithTriggers<C> {
...
+                (
+                    "number".into(),
+                    Value::BigInt(scalar::BigInt::from(block.block_number())),
+                ),
+                ("hash".into(), hash),
+            ])
+            .unwrap();
+
+        data
+    }
```
incrypto32, line 448: This is a mock implementation which is properly implemented in later PR's

graph/src/blockchain/mod.rs
```diff
@@ -370,6 +404,76 @@ pub trait UnresolvedDataSource<C: Blockchain>:
...
+{
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        match (self, other) {
+            (Trigger::Chain(data1), Trigger::Chain(data2)) => data1.cmp(data2),
+            (Trigger::Subgraph(_), Trigger::Chain(_)) => std::cmp::Ordering::Greater,
+            (Trigger::Chain(_), Trigger::Subgraph(_)) => std::cmp::Ordering::Less,
+            (Trigger::Subgraph(_), Trigger::Subgraph(_)) => std::cmp::Ordering::Equal,
+        }
+    }
+}
```
incrypto32, line 469: In a later PR the `Subgraph` Triggers are ordered by vid (which is generated in the order of creation of the entiites)

graph/src/blockchain/mod.rs
```diff
@@ -247,6 +250,37 @@ impl From<web3::Error> for IngestorError {
     }
 }
 
+#[derive(Debug)]
+pub struct TriggerFilterWrapper<C: Blockchain> {
```
mangas, line 254: Can you please add some comments on what this is, what it does and how it's supposed to be used?

chain/ethereum/src/chain.rs
```diff
@@ -409,17 +437,35 @@ impl Blockchain for Chain {
         deployment: DeploymentLocator,
         store: impl DeploymentCursorTracker,
         start_blocks: Vec<BlockNumber>,
-        filter: Arc<Self::TriggerFilter>,
+        source_subgraph_stores: Vec<(DeploymentHash, Arc<dyn WritableStore>)>,
```
mangas, line 440: Do we need a writable store to be used here? Is there another trait we could use with less access?

core/src/subgraph/context/instance/mod.rs
```diff
@@ -29,6 +29,9 @@ pub(crate) struct SubgraphInstance<C: Blockchain, T: RuntimeHostBuilder<C>> {
     /// will return the onchain hosts in the same order as they were inserted.
     onchain_hosts: OnchainHosts<C, T>,
 
+    // TODO(krishna): Describe subgraph_hosts
+    subgraph_hosts: OnchainHosts<C, T>,
```
mangas, line 33: Can you resolve the todo before merging please?

graph/src/data_source/subgraph.rs
```diff
@@ -0,0 +1,305 @@
...
+    pub api_version: String,
+    pub language: String,
+    pub file: Link,
+    pub handlers: Vec<EntityHandler>,
+    pub entities: Vec<String>,
+}
+
+impl UnresolvedDataSource {
+    #[allow(dead_code)]
+    pub(super) async fn resolve(
```
mangas, line 168: Does this need some test coverage?

mangas:
Would be good to address some of the comments, particularly around docs and comments. I also didn't see much in the way of testing. Would be great to see some test coverage around the new paths.

incrypto32:
Just addressed the comments, and the suggestion for not using the WriteStore

chain/ethereum/src/chain.rs
incrypto32: Addressed in latest commits

PR #5698, opened 2024-11-06
Optimize IPFS retries

### LINKED ISSUE #5696: Reduce unnecessary IPFS retries

isum:
This PR reduces the number of unnecessary IPFS retries by requiring an explicit retry policy, so that different components of the graph-node can decide to use different types of retries, or not use retries at all.

Closes #5696

**Other improvements:**

IPFS client pool now reuses the initial response instead of doubling the number of requests.

**Future improvements:**

This PR includes the first steps in defining deterministic IPFS errors so that some retries can be eliminated altogether. This functionality is not part of this PR as it requires more thought and testing and will be added in a future PR.

### DIFF
--- a/core/src/polling_monitor/ipfs_service.rs
+++ b/core/src/polling_monitor/ipfs_service.rs
@@ -101,9 +101,14 @@ mod test {
     use graph::ipfs::test_utils::add_files_to_local_ipfs_node_for_testing;
     use graph::ipfs::IpfsRpcClient;
     use graph::ipfs::ServerAddress;
+    use graph::log::discard;
     use graph::tokio;
     use tower::ServiceExt;
     use uuid::Uuid;
+    use wiremock::matchers as m;
+    use wiremock::Mock;
+    use wiremock::MockServer;
+    use wiremock::ResponseTemplate;
 
     use super::*;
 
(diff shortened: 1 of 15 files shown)


graph/src/ipfs/gateway_client.rs
```diff
@@ -353,108 +257,13 @@ mod tests {
...
-    }
-
     #[tokio::test]
     async fn new_unchecked_creates_the_client_without_checking_the_gateway() {
         let server = mock_server().await;
 
         IpfsGatewayClient::new_unchecked(server.uri(), &discard()).unwrap();
     }
 
-    #[tokio::test]
```
mangas, line 377: did you replace these with something else?

core/src/polling_monitor/ipfs_service.rs
```diff
@@ -50,12 +51,17 @@ impl IpfsServiceInner {
 
         let res = self
             .client
-            .cat(&path, self.max_file_size, Some(self.timeout))
+            .cat(
+                &path,
+                self.max_file_size,
+                Some(self.timeout),
+                RetryPolicy::None,
```
mangas, line 58: Any way we can make sure this is tested to prevent regression?

mangas (approved):
See comments. Like the fact you simplified some of the traits and cleared a good amount of code.

graph/src/ipfs/gateway_client.rs
isum: These tests were for the `.can_provide()` method used by the IPFS client pool to select the appropriate client for requests, but this is no longer needed, so the method and its tests have been removed.

I also removed tests for non-retriable errors, because from the perspective of IPFS clients, all errors are now retriable, except for deterministic errors that are not implemented yet.

core/src/polling_monitor/ipfs_service.rs
isum: The IPFS client's internal retries caused the high number of IPFS requests here, because the polling monitor has it's own infinite retries with priorities. By disabling the internal retries of the IPFS client, we allow the polling monitor to behave as expected and rely on it's retry functionality.

mangas: Understood. It's kind of hard to differentiate the different layers of retries. I'm happy to leave it as is since now we know it won't retry on the client level.

mangas (approved):

PR #5704, opened 2024-11-18
Subgraph Composition: Support ethereum host functions in dependant subgraphs

incrypto32:


mangas (approved):

maoueh:
Would this work be re-usable to also enable Ethereum hosts fn on Substreams datasource? This was requested before and when I saw the PR, it felt it could be applied to Substreams data sources too.

incrypto32:
@maoueh it can be applied but what purpose does it serve? Substreams cannot call these host functions on the graph-node anyways (i believe).

maoueh:
@incrypto32 

There is two Substreams integration now:
- EntityChanges where we feed graph-node with the right entity changes (no AssemblyScript involved)
- Substreams sources where a Subgraph dataSources input is the output module of a particular Substreams, this enables AssemblyScript linear processing.

It's on the second case that it make sense to add it. People have been request it to be able to make `eth_call` in this mode where they run AssemblyScript and they couldn't make `eth_call`.

incrypto32:
@maoueh got it, yes i think it makes sense to add them and i believe there's not much additional work other than just adding them as host_fns. Feel free to open an issue.

maoueh:
https://github.com/graphprotocol/graph-node/issues/5727

PR #5706, opened 2024-11-18
Subgraph Composition: Use Firehose endpoint to load blocks for subgraph triggers

incrypto32:


chain/ethereum/src/chain.rs
```diff
@@ -734,6 +741,55 @@ pub struct TriggersAdapter {
     unified_api_version: UnifiedMappingApiVersion,
 }
 
+/// Fetches blocks from the cache based on block numbers, excluding duplicates
+/// (i.e., multiple blocks for the same number), and identifying missing blocks that
+/// need to be fetched via RPC/Firehose. Returns a tuple of the found blocks and the missing block numbers.
+async fn fetch_unique_blocks_from_cache(
```
mangas, line 747: I think this would be a good candidate to add some unit test coverage

chain/ethereum/src/chain.rs
```diff
@@ -763,23 +819,96 @@ impl TriggersAdapterTrait<Chain> for TriggersAdapter {
         logger: Logger,
         block_numbers: HashSet<BlockNumber>,
     ) -> Result<Vec<BlockFinality>> {
-        use graph::futures01::stream::Stream;
+        // Common function to handle block loading, regardless of source
+        async fn load_blocks<F, Fut>(
```
mangas, line 823: is there a reason this couldn't be a function on triggers adapter instead of inside the function? This is more of a nitpick but this usually means it's code that cannot be tested and personally I think it makes code less readable. I understand the code locality here and I think it does help understand the full context (in this case). Just wary of doing this sort of thing as a norm

chain/ethereum/src/chain.rs
```diff
@@ -840,13 +969,25 @@ impl TriggersAdapterTrait<Chain> for TriggersAdapter {
...
 
     async fn is_on_main_chain(&self, ptr: BlockPtr) -> Result<bool, Error> {
-        self.chain_client
-            .rpc()?
-            .cheapest()
-            .await
-            .ok_or(anyhow!("unable to get adapter for is_on_main_chain"))?
-            .is_on_main_chain(&self.logger, ptr.clone())
-            .await
+        match &*self.chain_client {
```
mangas, line 972: wondering if this should be on the ChainClient, since it's implemented already for rpc adapters, perhaps it would make sense to get the block from the adapter (whichever) and then have the logic on the layer above that makes the decision on whether it is on the main chain.

incrypto32: Would be hard to move since, the `adapter.cheapest()` method is only implemented for `EthereumNetworkAdapters` its not on the trait. Its not part of the `Blockchain::Client` which is what is available in the `ChainClient` implmentation.

chain/ethereum/src/chain.rs
incrypto32: The reason is that its a very specific function and it is not at all likely to be used anywhere else since this method takes closure which `fetch_missing`. Overall its a very specific function and wouldn't be used anywhere else. But yes feels like this could be moved outside, I'll just go with an external function, instead of inline

chain/ethereum/src/chain.rs
incrypto32: Added in latest commit

PR #5720, opened 2024-11-29
Add `deployment_status` metric

### LINKED ISSUE #5405: Update deployment_failed to better reflect syncing status
The deployment_failed metric is only created/updated once the block stream has been set up, but not when starting up a deployment fails, e.g. [because it references a non-existent chain](https://github.com/graphprotocol/graph-node/blob/4869e73d37a287979774d6f909b7000183124c7c/core/src/subgraph/instance_manager.rs#L315)

This makes it harder to detect failed subgraphs. We should update (and maybe rename) this metric so that it better captures syncing status

starting - a node figured that it should run the deployment and is working on starting it; that should be set as early as possible
running - everything's fine, we are processing blocks
failed - something went wrong

isum:
This PR adds a new `deployment_status` metric with the following behavior:
- Once graph-node has figured out that it should index a deployment, `deployment_status` is set to `1` _(starting)_;
- When the block stream is created and blocks are ready to be processed, `deployment_status` is set to `2` _(running)_;
- When a deployment is unassigned, `deployment_status` is set to `3` _(stopped)_;
- If a temporary or permanent failure occurs, `deployment_status` is set to `4` _(failed)_;
    - If indexing manages to recover from a temporary failure, the `deployment_status` is set back to `2` _(running)_;

_No other values for `deployment_status` should be expected._

Closes #5405

## Breaking changes

The `deployment_failed` metric is removed and the failures are reported by the new `deployment_status` metric.

Node operators that relied on `deployment_failed` should switch to using `deployment_status`.

## Testing

The new metric was tested on a local graph-node and Prometheus setup. I have tested various scenarios such as errors, pauses, restarts and the metric behaves as expected and reports correct values.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,24 @@
 # NEWS
 
+## v0.36.1
+
+### What's new
+
+- A new `deployment_status` metric is added [(#5720)](https://github.com/graphprotocol/graph-node/pull/5720) with the
+  following behavior:
+    - Once graph-node has figured out that it should index a deployment, `deployment_status` is set to `1` _(starting)_;
+    - When the block stream is created and blocks are ready to be processed, `deployment_status` is set to `2` _(
+      running)_;
+    - When a deployment is unassigned, `deployment_status` is set to `3` _(stopped)_;
+    - If a temporary or permanent failure occurs, `deployment_status` is set to `4` _(failed)_;
+        - If indexing manages to recover from a temporary failure, the `deployment_status` is set back to `2` _(
+          running)_;
+
+### Breaking changes
+
+- The `deployment_failed` metric is removed and the failures are reported by the new `deployment_status`
+  metric. [(#5720)](https://github.com/graphprotocol/graph-node/pull/5720)
+
 ## v0.36.0
 
 ### Note on Firehose Extended Block Details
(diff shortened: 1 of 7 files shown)


NEWS.md
```diff
@@ -1,5 +1,24 @@
 # NEWS
 
+## Unreleased
```
mangas, line 3: Nobody will remember to change this I think might as well use the next point release here

core/src/subgraph/runner.rs
```diff
@@ -244,6 +244,8 @@ where
 
             debug!(self.logger, "Starting block stream");
 
+            self.metrics.subgraph.deployment_status.running();
```
mangas, line 247: I think you may be missing some other options here like when the runner stops for any reason (pause, stop, etc) where it's not an error.

mangas: probably worth checking with pause resume locally for example.

NEWS.md
isum: Updated to `v0.36.1`, but it feels a bit odd that a patch update introduces a breaking change.

core/src/subgraph/runner.rs
isum: I have tested multiple new cases locally and the new metric behaves as expected.

mangas (approved):

PR #5721, opened 2024-12-02
Subgraph Composition: Support declared calls for subgraph datasources

incrypto32:


lutter (approved):
Nice! This was a lot more involved than I would have thought, though since it adds a whole different kind of declared calls, it's maybe not that surprising.

PR #5731, opened 2024-12-16
Subgraph Composition: Support multiple subgraph datasources

incrypto32:


graph/src/data_source/subgraph.rs
```diff
@@ -354,18 +354,38 @@ pub struct MappingEntityTrigger {
...
     }
 
     pub fn entity_type(&self) -> &str {
         self.entity.entity_type.as_str()
     }
 }
 
+impl Ord for TriggerData {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        match self.source_idx.cmp(&other.source_idx) {
```
mangas, line 376: This is somewhat important, I think would be worth some comments/documentation about the decision and why it is working this way. Also probably a simple test to ensure this assumption is not broken going forward

mangas:
LGTM, really nice that you got added an integration test with the aggregation.

zorancv (approved):

PR #5737, opened 2024-12-18
Rework generation of the VID

zorancv:
This PR changes the generation of the VID in such a way that entities have an order determined by the order of the blockchain transactions from which they originate, rather than to be ordered by the current implementation of the graph-node and the VID autogenerated by the database. This in turn will ensure that the dependant subgraphs would have a deterministic order of their input entities and it would be consistent across different graph-node deployments producing the same POI.

### DIFF
--- a/core/src/subgraph/runner.rs
+++ b/core/src/subgraph/runner.rs
@@ -1670,6 +1670,7 @@ async fn update_proof_of_indexing(
         key: EntityKey,
         digest: Bytes,
         block_time: BlockTime,
+        block: BlockNumber,
     ) -> Result<(), Error> {
         let digest_name = entity_cache.schema.poi_digest();
         let mut data = vec![
@@ -1684,11 +1685,12 @@ async fn update_proof_of_indexing(
             data.push((entity_cache.schema.poi_block_time(), block_time));
         }
         let poi = entity_cache.make_entity(data)?;
-        entity_cache.set(key, poi, None)
+        entity_cache.set(key, poi, block, None)
     }
 
     let _section_guard = stopwatch.start_section("update_proof_of_indexing");
 
+    let block_number = proof_of_indexing.get_block();
     let mut proof_of_indexing = proof_of_indexing.take();
 
     for (causality_region, stream) in proof_of_indexing.drain() {
@@ -1724,6 +1726,7 @@ async fn update_proof_of_indexing(
             entity_key,
             updated_proof_of_indexing,
             block_time,
+            block_number,
         )?;
     }
 
--- a/graph/src/components/store/entity_cache.rs
+++ b/graph/src/components/store/entity_cache.rs
@@ -1,4 +1,4 @@
-use anyhow::anyhow;
+use anyhow::{anyhow, bail};
 use std::borrow::Borrow;
 use std::collections::HashMap;
 use std::fmt::{self, Debug};
@@ -16,6 +16,10 @@ use crate::util::lfu_cache::{EvictStats, LfuCache};
 use super::{BlockNumber, DerivedEntityQuery, LoadRelatedRequest, StoreError};
 
 pub type EntityLfuCache = LfuCache<EntityKey, Option<Arc<Entity>>>;
+
+// Number of VIDs that are reserved outside of the generated ones here.
+// Currently none is used, but lets reserve a few more.
+const RESERVED_VIDS: u32 = 100;
 
 /// The scope in which the `EntityCache` should perform a `get` operation
 pub enum GetScope {
@@ -105,6 +109,10 @@ pub struct EntityCache {
     /// generated IDs, the `EntityCache` needs to be newly instantiated for
     /// each block
     seq: u32,
+
+    // Sequence number of the next VID value for this block. The value written
+    // in the database consist of a block number and this SEQ number.
+    pub vid_seq: u32,
 }
 
 impl Debug for EntityCache {
@@ -132,6 +140,7 @@ impl EntityCache {
             schema: store.input_schema(),
             store,
             seq: 0,
+            vid_seq: RESERVED_VIDS,
         }
     }
 
@@ -152,6 +161,7 @@ impl EntityCache {
             schema: store.input_schema(),
             store,
             seq: 0,
+            vid_seq: RESERVED_VIDS,
         }
     }
 
@@ -353,6 +363,7 @@ impl EntityCache {
         &mut self,
         key: EntityKey,
         entity: Entity,
+        block: BlockNumber,
         write_capacity_remaining: Option<&mut usize>,
     ) -> Result<(), anyhow::Error> {
         // check the validate for derived fields
@@ -360,7 +371,6 @@ impl EntityCache {
 
         if let Some(write_capacity_remaining) = write_capacity_remaining {
             let weight = entity.weight();
-
             if !self.current.contains_key(&key) && weight > *write_capacity_remaining {
                 return Err(anyhow!(
                     "exceeded block write limit when writing entity `{}`",
@@ -369,6 +379,21 @@ impl EntityCache {
             }
 
             *write_capacity_remaining -= weight;
+        }
+
+        // The next VID is based on a block number and a sequence within the block
+        let vid = ((block as i64) << 32) + self.vid_seq as i64;
+        self.vid_seq += 1;
+        let mut entity = entity;
+        let old_vid = entity.set_vid(vid).expect("the vid should be set");
+        // Make sure that there was no VID previously set for this entity.
+        if let Some(ovid) = old_vid {
+            bail!(
+                "VID: {} of entity: {} with ID: {} was already present when set in EntityCache",
+                ovid,
+                key.entity_type,
+                entity.id()
+            );
         }
 
         self.entity_op(key.clone(), EntityOp::Update(entity));
@@ -507,7 +532,7 @@ impl EntityCache {
                 // Entity was removed and then updated, so it will be overwritten
                 (Some(current), EntityOp::Overwrite(data)) => {
                     let data = Arc::new(data);
-                    self.current.insert(key.clone(), Some(data.clone()));
+                    self.current.insert(key.clone(), Some(data.cheap_clone()));
                     if current != data {
                         Some(Overwrite {
                             key,
--- a/graph/src/data/store/mod.rs
+++ b/graph/src/data/store/mod.rs
@@ -3,7 +3,7 @@ use crate::{
     derive::CacheWeight,
     prelude::{lazy_static, q, r, s, CacheWeight, QueryExecutionError},
     runtime::gas::{Gas, GasSizeOf},
-    schema::{EntityKey, EntityType},
+    schema::{input::VID_FIELD, EntityKey, EntityType},
     util::intern::{self, AtomPool},
     util::intern::{Error as InternError, NullValue, Object},
 };
@@ -908,6 +908,29 @@ impl Entity {
     /// return an error
     pub fn id(&self) -> Id {
         Id::try_from(self.get("id").unwrap().clone()).expect("the id is set to a valid value")
+    }
+
+    /// Return the VID of this entity and if its missing or of a type different than
+    /// i64 it panics.
+    pub fn vid(&self) -> i64 {
+        self.get(VID_FIELD)
+            .expect("the vid must be set")
+            .as_int8()
+            .expect("the vid must be set to a valid value")
+    }
+
+    /// Sets the VID of the entity. The previous one is returned.
+    pub fn set_vid(&mut self, value: i64) -> Result<Option<Value>, InternError> {
+        self.0.insert(VID_FIELD, value.into())
+    }
+
+    /// Sets the VID if it's not already set. Should be used only for tests.
+    #[cfg(debug_assertions)]
+    pub fn set_vid_if_empty(&mut self) {
+        let vid = self.get(VID_FIELD);
+        if vid.is_none() {
+            let _ = self.set_vid(100).expect("the vid should be set");
+        }
     }
 
     /// Merges an entity update `update` into this entity.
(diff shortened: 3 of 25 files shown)


graph/src/components/store/entity_cache.rs
```diff
@@ -105,6 +105,10 @@ pub struct EntityCache {
     /// generated IDs, the `EntityCache` needs to be newly instantiated for
     /// each block
     seq: u32,
+
+    // Sequence number of the next VID value for this block. The value written
+    // in the database consist of a block number and this SEQ number.
+    pub vid_seq: i32,
```
lutter, line 111: Why is this an `i32`? Shouldn't it be a `u32`?

store/postgres/src/relational/ddl.rs
```diff
@@ -116,19 +116,26 @@ impl Table {
             Ok(cols)
         }
 
+        let vid_type = if self.object.is_poi() || !self.object.is_object_type() {
```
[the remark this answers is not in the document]
lutter: Strictly speaking, for POI it doesn't matter, and there's never instances of interfaces (only objects that implement an interface), but I think for simplicity, it might be best to autogenerate a vid for every kind of entity we deal with.

lutter (changes requested):
A general question: we already require that every `Entity` has an `id`, if it doesn't, things will fail in funky ways. Wouldn't it be simpler to just treat the `vid` as another attribute of `Entity`, i.e., require that `entity.get("vid")` always returns something instead of keeping it separate from the entity?

graph/src/components/store/entity_cache.rs
zorancv: My mistake. Good catch!

store/postgres/src/relational/ddl.rs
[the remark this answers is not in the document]
zorancv: So you would prefer that all is generated by us rather than from the database? Is there inconsistency concern regarding the aggregation (that one is more involved)?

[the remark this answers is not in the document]
zorancv: There is a PR https://github.com/graphprotocol/graph-node/pull/5770 that disables aggregations entities as source entities for now.

zorancv:
> A general question: we already require that every `Entity` has an `id`, if it doesn't, things will fail in funky ways. Wouldn't it be simpler to just treat the `vid` as another attribute of `Entity`, i.e., require that `entity.get("vid")` always returns something instead of keeping it separate from the entity?

@lutter I've decided to go that route. It simplifies the code a lot. Also I've added a check in the `EntityCache`  `set()` that there was no `vid` previously set.

graph/src/components/store/entity_cache.rs
```diff
@@ -349,10 +359,23 @@ impl EntityCache {
...
         // check the validate for derived fields
         let is_valid = entity.validate(&key).is_ok();
 
+        // The next VID is based on a block number and a sequence within the block
+        let vid = ((block as i64) << 32) + self.vid_seq as i64;
+        self.vid_seq += 1;
+        let mut entity = entity;
+        let old_vid = entity.set_vid(vid).expect("the vid should be set");
+        // Make sure that there was no VID previously set for this entity.
+        assert!(old_vid.is_none());
```
lutter, line 377: This shouldn't be an assertion - tripping that would take the whole server down. Instead, it should return an error so that only this subgraph is affected.

graph/src/data/store/mod.rs
```diff
@@ -910,6 +910,29 @@ impl Entity {
         Id::try_from(self.get("id").unwrap().clone()).expect("the id is set to a valid value")
     }
 
+    /// Return the VID of this entity and if its missing or of a type different than
+    /// i64 it panics.
+    pub fn vid(&self) -> i64 {
+        self.get("vid")
```
lutter, line 916: It would be better to have a constant `VID: &str = "vid"` to guard against typos. There might even be one defined somewhere already

graph/src/data/subgraph/api_version.rs
```diff
@@ -55,10 +55,12 @@ pub const SPEC_VERSION_1_1_0: Version = Version::new(1, 1, 0);
 pub const SPEC_VERSION_1_2_0: Version = Version::new(1, 2, 0);
 
 // Enables subgraphs as datasource
+// Change the way the VID field is generated. It used to be autoincrement. Now its
+// based on block number and the sequence of the entities in a block.
```
lutter, line 59: This should also mention that ordering by `vid` now, across all entity types for a subgraph, returns entities in the order in which they were written.

graph/src/schema/input/mod.rs
```diff
@@ -35,6 +35,7 @@ pub(crate) const POI_OBJECT: &str = "Poi$";
 const POI_DIGEST: &str = "digest";
 /// The name of the PoI attribute for storing the block time
 const POI_BLOCK_TIME: &str = "blockTime";
+const VID_FIELD: &str = "vid";
```
lutter, line 38: There is a constant, yay!

store/postgres/src/relational/ddl.rs
```diff
@@ -116,19 +116,26 @@ impl Table {
             Ok(cols)
         }
 
+        let vid_type = if !self.object.is_object_type() {
```
lutter, line 119: This would be easier to read if it was not negated. And maybe has a comment that the `vid` of aggregations are not necessarily in insertion order

store/postgres/src/relational/prune.rs
```diff
@@ -252,13 +237,6 @@ impl TablePair {
                   "src" => src_nsp.as_str(), "error" => e.to_string());
         }
 
-        // Make sure the vid sequence
-        // continues from where it was
-        writeln!(
-            query,
-            "select setval('{dst_nsp}.{vid_seq}', nextval('{src_nsp}.{vid_seq}'));"
-        )?;
-
```
lutter, line 261: Is this also correct for aggregations since they still have a vid sequence?

store/postgres/src/relational_queries.rs
```diff
@@ -539,7 +539,14 @@ impl EntityData {
                     // table column; those will be things like the
                     // block_range that `select *` pulls in but that we
                     // don't care about here
-                    if let Some(column) = table.column(&SqlName::verbatim(key)) {
+                    if key == "vid" {
```
lutter, line 542: There's a lot of literal `"vid"` strings here, it would be better to use a constant.

graph/src/components/store/entity_cache.rs
```diff
@@ -17,6 +17,10 @@ use super::{BlockNumber, DerivedEntityQuery, LoadRelatedRequest, StoreError};
 
 pub type EntityLfuCache = LfuCache<EntityKey, Option<Arc<Entity>>>;
 
+// Number of VIDs that are reserved ourside of the generated ones here.
```
lutter, line 20: typo: s/ourside/outside/

core/src/subgraph/runner.rs
```diff
@@ -1617,11 +1618,12 @@ async fn update_proof_of_indexing(
             data.push((entity_cache.schema.poi_block_time(), block_time));
         }
         let poi = entity_cache.make_entity(data)?;
-        entity_cache.set(key, poi)
+        entity_cache.set(key, poi, block)
     }
 
     let _section_guard = stopwatch.start_section("update_proof_of_indexing");
 
+    let block_number = proof_of_indexing.get_block();
```
lutter, line 1626: I didn't quite understand what's happening here with the PoI block number. Why is this needed?

lutter (commented):
This looks good. I'd mostly just like to understand better what is happening with the explicit block number for the PoI.

Also, I don't think there are any tests that ensure that the vid's are indeed set by insertion order. It would be good to have a test for the `EntityCache` that ensures that.

graph/src/data/store/mod.rs
```diff
@@ -910,6 +910,29 @@ impl Entity {
...
+            .expect("the vid is set")
+            .as_int8()
+            .expect("the vid is set to a valid value")
+    }
+
+    /// Sets the VID of the entity. The previous one is returned.
+    pub fn set_vid(&mut self, value: i64) -> Result<Option<Value>, InternError> {
+        self.0.insert("vid", value.into())
+    }
```
lutter, line 925: One additional thought: it would be great if we could guarantee that an entity is always constructed with the `vid` set, similar to how that's done for the `id`. I haven't checked how much work that is, but it would make the various expects sprinkled around here easier to accept.

core/src/subgraph/runner.rs
zorancv: PIO like other entities has a VID generated by us, and for that we also need a block number.

graph/src/data/store/mod.rs
zorancv: I like the suggestion but would like to postpone it together with the testing request above for a later stage. WDYT?

zorancv: The test for the VID order is implemented now: https://github.com/graphprotocol/graph-node/pull/5737/commits/166c9098c4c9a6ae2464706b6d4e4dc79746e6d3

lutter: I am fine with postponing the enforcement of having a `vid`

graph/src/components/store/entity_cache.rs
```diff
@@ -374,7 +374,9 @@ impl EntityCache {
         let mut entity = entity;
         let old_vid = entity.set_vid(vid).expect("the vid should be set");
         // Make sure that there was no VID previously set for this entity.
-        assert!(old_vid.is_none());
+        if old_vid.is_some() {
+            bail!("VID was already present when set in EntityCache");
```
lutter, line 378: This is very very minor but in these error messages it's very helpful if it includes the entity that is involved (type and id) and what the old vid was.

graph/src/data/subgraph/api_version.rs
```diff
@@ -54,9 +54,10 @@ pub const SPEC_VERSION_1_1_0: Version = Version::new(1, 1, 0);
 // Enables eth call declarations and indexed arguments(topics) filtering in manifest
 pub const SPEC_VERSION_1_2_0: Version = Version::new(1, 2, 0);
 
-// Enables subgraphs as datasource
-// Change the way the VID field is generated. It used to be autoincrement. Now its
-// based on block number and the sequence of the entities in a block.
+// Enables subgraphs as datasource.
+// Changes the way the VID field is generated. It used to be autoincrement. Now its
+// based on block number and the order of the entities in a block. The later is
+// representing the writting order of all entity types in a subgraph.
```
lutter, line 60: Typo: "The latter represents the write order across all entity types in the subgraph"

store/postgres/src/relational/prune.rs
```diff
@@ -228,6 +239,11 @@ impl TablePair {
         let src_qname = &self.src.qualified_name;
         let dst_qname = &self.dst.qualified_name;
         let src_nsp = &self.src_nsp;
+        let dst_nsp = &self.dst_nsp;
+
+        let vid_seq = format!("{}_{VID_COLUMN}_seq", self.src.name);
+
+        let old_vid_form = !self.src.object.is_object_type();
```
lutter, line 246: This is also very minor: since we are now using `!EntityType.is_object_type()` in a few places to indicate whether the type has/should have a `vid` sequence, it would be cleaner if there was a helper `EntityType.has_vid_seq()` since that makes it clear what this check is asking for. (the impl of the method would just be `!self.is_object_type()`)

store/postgres/src/relational/prune.rs
```diff
@@ -237,6 +253,15 @@ impl TablePair {
                   "src" => src_nsp.as_str(), "error" => e.to_string());
         }
 
+        // Make sure the vid sequence
+        // continues from where it was
+        if old_vid_form {
```
lutter, line 258: Also minor, but this could just be
```rust
if self.src.object.has_vid_seq() {
    let vid_seq = format!("{}_{VID_COLUMN}_seq", self.src.name);
    writeln!(..)
}
```

store/postgres/src/relational_queries.rs
```diff
@@ -2548,6 +2548,8 @@ impl<'a> QueryFragment<Pg> for InsertQuery<'a> {
         let out = &mut out;
         out.unsafe_to_cache_prepared();
 
+        let new_vid_form = self.table.object.is_object_type();
```
lutter, line 2551: Similarly, this could say `self.table.object.has_vid_seq()` which is also a little clearer than trying to remember what old and new vid forms are

lutter (approved):
Nice! I left some minor stylistic comments, would be great to address them, but I am totally fine with merging this.

store/postgres/src/relational/prune.rs
zorancv: Addressed in the next PR in https://github.com/graphprotocol/graph-node/pull/5782/commits/522d7f79aab981188c3fedf5521d1d5db212a8dd

store/postgres/src/relational/prune.rs
zorancv: Addressed in the next PR in https://github.com/graphprotocol/graph-node/pull/5782/commits/522d7f79aab981188c3fedf5521d1d5db212a8dd

store/postgres/src/relational_queries.rs
zorancv: Addressed in the next PR in https://github.com/graphprotocol/graph-node/pull/5782/commits/522d7f79aab981188c3fedf5521d1d5db212a8dd

store/postgres/src/relational/prune.rs
lutter: Not sure where best to leave that comment (and maybe I am missing something with all the rebasing etc.) but you'll need two different helpers on `EntityType`: `strict_vid_order()` is just a check based on the spec version, and `has_vid_seq()` is a check that uses both `strict_vid_order()` and whether we are dealing with a normal object type or an aggregation. In the commit I saw, it looked like `strict_vid_order()` was now being used in places that should use `has_vid_seq()`

PR #5740, opened 2024-12-28
fix graphman deploy command logic when creating subgraph

### LINKED ISSUE #5739: [Bug] graphman deploy shows deployed even if subgraph was not deployed
### Bug report

Here `busd2` subgraph doesn't exist. So it should fail with error and ask to use `--create` flag instead of showing successfuly deployed
Screenshot 2024-12-28 at 7 26 47 AM

### Relevant log output

_No response_

### IPFS hash

_No response_

_No response_

### OS information

None

shiyasmohd:
Fixes #5739

### DIFF
--- a/node/src/manager/commands/deploy.rs
+++ b/node/src/manager/commands/deploy.rs
@@ -80,20 +80,22 @@ pub async fn run(
         _ => bail!("The `name` must be a valid subgraph name"),
     };
 
-    if create {
+    let subgraph_name =
+        SubgraphName::new(name.clone()).map_err(|_| anyhow!("Invalid subgraph name"))?;
+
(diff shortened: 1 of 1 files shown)


lutter:
I wonder if the current or the new behavior is really what users want. IMHO, names are mostly annoying, and I think it would be nicer/easier if `graphman deploy <name> <hash>` just created the name if it doesn't exist yet and we'd get rid of the `--create` flag altogether. Thoughts?

shiyasmohd:
Yes. That would be better. Thanks for the suggestion. I'll update the PR

incrypto32 (approved):

shiyasmohd:
I've move this PR to another PR #6077 due to signatures required

Issue #5741, opened 2024-12-28
[Bug] graphman rewind to start block not working


shiyasmohd:
### Bug report

The `startBlock` of this subgraph is `7087455`, But its taken as `0` for `startBlock`
Deployment ID: `QmeKaYmqjUSTNHPynX1yCshHK4cTV3RCpphx51tqHE9Y2P`

<img width="857" alt="Screenshot 2024-12-28 at 11 42 21 PM" src="https://github.com/user-attachments/assets/4f42b7bf-9b1c-49f7-8c0e-a08fbf37aca5" />

### IPFS hash

QmeKaYmqjUSTNHPynX1yCshHK4cTV3RCpphx51tqHE9Y2P

### Some information to help us out

- [ ] Tick this box if this bug is caused by a regression found in the latest release.
- [ ] Tick this box if this bug is specific to the hosted service.
- [X] I have searched the issue tracker to make sure this issue is not a duplicate.

### OS information

None

PR #5742, opened 2024-12-28
fix graphman rewind to start block

shiyasmohd:
Fixes: #5741


mitchhs12:
Why was this closed?
Shouldn't this PR allow rewinding to the `start_block` of a subgraph?

shiyasmohd:
PR Moved to #6091

PR #5755, opened 2025-01-06
Fix what seems to be a bad SQL join

King-witcher:


lutter (approved):
Oooh .. nice find! That's a very subtle bug. Did you run into that?

lutter:
Somehow our workflows are misconfigured and don't kick off on this, but I just ran it [manually](https://github.com/graphprotocol/graph-node/actions/runs/13271965057) and it passed, so merging this.

PR #5760, opened 2025-01-09
Subgraph Composition: Fix Entity Ordering with different Idtype

incrypto32:


store/postgres/src/relational_queries.rs
```diff
@@ -558,45 +558,78 @@ impl EntityData {
...
+            return Ok(ord);
+        }
+
+        match idtype {
+            IdType::String => {
+                let self_str = String::from_utf8(self.id.clone()).map_err(IdCompareError::InvalidUtf8)?;
+                let other_str = String::from_utf8(other.id.clone()).map_err(IdCompareError::InvalidUtf8)?;
+                Ok(self_str.cmp(&other_str))
+            }
+            IdType::Bytes => Ok(self.id.cmp(&other.id)),
```
zorancv, line 619: The same check could be used for the string too. We don't care what is the content of the ID, only if it's equal to the other one. And I would go so far to say that it holds for Int8 case too as we don't mind if the order is ascending or descending, we only need the `cmp()` to be able to detect equality and have a consistent ordering otherwise which is achieved with byte comparison. So I would keep the `Ord` a it is and only change the type of the `id` field in `EntityDataExt`.

store/test-store/tests/postgres/writable.rs
```diff
@@ -406,3 +420,71 @@ fn read_immutable_only_range_test() {
         assert_eq!(e.len(), 4);
     })
 }
+
+#[test]
+fn read_range_pool_created_test() {
```
zorancv, line 425: Very nice to add a test case!

zorancv (changes requested):

store/postgres/src/relational.rs
```diff
@@ -576,6 +576,13 @@ impl Layout {
             Ok((ewt, block))
         };
 
+        fn compare_entity_data_ext(a: &EntityDataExt, b: &EntityDataExt) -> std::cmp::Ordering {
+            a.block_number
+                .cmp(&b.block_number)
+                .then_with(|| a.entity.cmp(&b.entity))
```
zorancv, line 582: elegant!

zorancv (approved):

PR #5770, opened 2025-01-14
Subgraph Compositions: Validations

incrypto32:


zorancv (approved):
LGTM

tests/tests/runner_tests.rs
```diff
@@ -1091,54 +1090,6 @@ async fn parse_data_source_context() {
     );
 }
 
-#[tokio::test]
```
zorancv, line 1094: Taking a second look at this PR, why is this test removed?

zorancv: Nice! Thanks for the explanation!

graph/src/data_source/subgraph.rs
```diff
@@ -211,8 +215,62 @@ pub struct UnresolvedMapping {
...
+            .context("Failed to parse source subgraph manifest")?;
+
+        source_manifest
+            .resolve(resolver, logger, LATEST_VERSION.clone())
+            .await
+            .context("Failed to resolve source subgraph manifest")
+            .map(Arc::new)
+    }
+
     #[allow(dead_code)]
```
mangas, line 272: do we still need this tag?

graph/src/data_source/subgraph.rs
```diff
@@ -224,7 +282,38 @@ impl UnresolvedDataSource {
...
+
+        if source_spec_version < &SPEC_VERSION_1_3_0 {
+            return Err(anyhow!(
+                "Source subgraph manifest spec version {} is not supported, minimum supported version is {}",
+                source_spec_version,
+                SPEC_VERSION_1_3_0
+            ));
+        }
+
+        let pruning_enabled = match source_manifest.indexer_hints.as_ref() {
```
mangas, line 297: pruning can also be set outside of the manifest right? Is there are a way we check that too?

graph/src/data_source/subgraph.rs
```diff
@@ -211,8 +215,62 @@ pub struct UnresolvedMapping {
 }
 
 impl UnresolvedDataSource {
+    fn validate_mapping_entities<C: Blockchain>(
```
mangas, line 218: Could you add some tests for this as well?

mangas (approved):
Approve to unblock. As discussed, there are a few considerations: 

1. Pruning can also be run through graphman so we should try and avoid running composition if a subgraph was pruned (while not having that hint on manifest)
2. The manifest hints, are hints. Therefore, it could (maybe should?) be possible to deploy the source subgraph WITHOUT pruning even though it has an indexing hint that suggests it be used. 
3. As this is a corner case I think it would be ok to document the issues and get back to them once we have the bulk of the work merged and working for the general case.

PR #5780, opened 2025-01-22
Add tracking for new metrics in CSV (`entity_changes`, `storage_size`)

MoonBoi9001:
**This PR introduces tracking for entity count changes and current storage size of entities in BlockStateMetrics.**

While the tracking mechanism follows a similar pattern to other metrics (`gas`, `op`, `read_bytes`, `write_bytes`), there's an important distinction in how the newly introduced metrics `entity_changes`  and `storage_size` values are recorded:
- The other metrics use monotonic counters that accumulate totals over time (e.g., total bytes read)
- The new `entity_count_changes` metric tracks the net change in entity counts per block (e.g., +5 id_1 added, -2 id_2 removed)
- The new `storage_size` metric tracks the actual current size of stored entities as a live readout. If the entity increases/reduces in size due to a write event, then this metric will reflect the new size of the entity after the write event.

For `entity_count_changes`, each block's measurement represents changes that occurred within that block, not cumulative totals or absolute counts. To determine the total number of entities at a given block, these changes would need to be summed across all previous blocks.

### DIFF
--- a/graph/src/components/metrics/block_state.rs
+++ b/graph/src/components/metrics/block_state.rs
@@ -10,7 +10,7 @@ use url::Url;
 use crate::{
     blockchain::BlockPtr,
     components::store::{DeploymentId, Entity},
-    data::store::Id,
+    data::{store::Id, value::Word},
     env::ENV_VARS,
     runtime::gas::Gas,
     schema::EntityType,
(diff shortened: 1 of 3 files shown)


camsam01:
Realised that I should have been more clear in my PR description as I have implied that the entity count also uses a monotonic counter, I was just making a statement that the other metrics use monotonic counters. 


Here is a better PR description:


#### This PR introduces tracking for entity counts in BlockStateMetrics.

#### While the tracking mechanism is consistent with other metrics (gas, op, read_bytes, write_bytes), there's an important distinction in how the values are recorded. The other metrics use monotonic counters that only increase over time, but the new entity count metric tracks the actual number of entities present in the subgraph at each measurement point. The entity_count_changes field represents the live entity count, not a cumulative total of changes.

PR #5782, opened 2025-01-23
Subgraph composition spec version

zorancv:
This PR aims to base `VID` generation on `spec_version` so that both old and new way of generating `VID` will work simultaneously in the single binary. Some code that was removed in https://github.com/graphprotocol/graph-node/pull/5737 is reintroduced here.

### DIFF
--- a/graph/src/components/store/entity_cache.rs
+++ b/graph/src/components/store/entity_cache.rs
@@ -207,9 +207,25 @@ impl EntityCache {
         };
 
         // Always test the cache consistency in debug mode. The test only
-        // makes sense when we were actually asked to read from the store
+        // makes sense when we were actually asked to read from the store.
+        // We need to remove the VID as the one from the DB  might come from
+        // a legacy subgraph that has VID autoincremented while this trait
+        // always creates it in a new style.
         debug_assert!(match scope {
-            GetScope::Store => entity == self.store.get(key).unwrap().map(Arc::new),
+            GetScope::Store => {
+                // Release build will never call this function and hence it's OK
+                // when that implementation is not correct.
+                fn remove_vid(entity: Option<Arc<Entity>>) -> Option<Entity> {
+                    entity.map(|e| {
+                        #[allow(unused_mut)]
+                        let mut entity = (*e).clone();
+                        #[cfg(debug_assertions)]
+                        entity.remove("vid");
+                        entity
+                    })
+                }
+                remove_vid(entity.clone()) == remove_vid(self.store.get(key).unwrap().map(Arc::new))
+            }
             GetScope::InBlock => true,
         });
 
--- a/graph/src/schema/entity_type.rs
+++ b/graph/src/schema/entity_type.rs
@@ -150,6 +150,14 @@ impl EntityType {
     pub fn is_object_type(&self) -> bool {
         self.schema.is_object_type(self.atom)
     }
+
+    // Changes the way the VID field is generated. It used to be autoincrement. Now its
+    // based on block number and the order of the entities in a block. The latter
+    // represents the write order across all entity types in the subgraph.
+    pub fn strict_vid_order(&self) -> bool {
+        // Currently the agregations entities don't have VIDs in insertion order
+        self.schema.strict_vid_order() && self.is_object_type()
+    }
 }
 
 impl fmt::Display for EntityType {
--- a/graph/src/schema/input/mod.rs
+++ b/graph/src/schema/input/mod.rs
@@ -17,6 +17,7 @@ use crate::data::graphql::{DirectiveExt, DocumentExt, ObjectTypeExt, TypeExt, Va
 use crate::data::store::{
     self, EntityValidationError, IdType, IntoEntityIterator, TryIntoEntityIterator, ValueType, ID,
 };
+use crate::data::subgraph::SPEC_VERSION_1_3_0;
 use crate::data::value::Word;
 use crate::derive::CheapClone;
 use crate::prelude::q::Value;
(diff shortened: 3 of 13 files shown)


graph/src/components/store/entity_cache.rs
```diff
@@ -207,9 +207,13 @@ impl EntityCache {
...
         // Always test the cache consistency in debug mode. The test only
-        // makes sense when we were actually asked to read from the store
+        // makes sense when we were actually asked to read from the store.
+        // We need to remove the VID as the one from the DB  might come from
+        // a legacy subgraph that has VID autoincremented while this trait
+        // always creates it in a new style.
         debug_assert!(match scope {
-            GetScope::Store => entity == self.store.get(key).unwrap().map(Arc::new),
+            GetScope::Store =>
+                remove_vid(entity.clone()) == remove_vid(self.store.get(key).unwrap().map(Arc::new)),
```
lutter, line 216: You could actually declare `remove_vid` in the match statement, i.e.,
```rust
   GetScope::Store => {
     fn remove_vid(..) { .. }
     remove_vid(..)
  }
```

Not sure which is better, but it would avoid declaring the function in release builds.

store/postgres/src/relational/prune.rs
```diff
@@ -243,7 +243,8 @@ impl TablePair {
 
         let vid_seq = format!("{}_{VID_COLUMN}_seq", self.src.name);
 
-        let old_vid_form = !self.src.object.is_object_type();
+        let old_vid_form = !(self.src.object.new_vid_form() && self.src.object.is_object_type());
```
lutter, line 246: If there was a `EntityType.has_vid_seq()`, you would have only needed to update that implementation ;) More seriously, the logic is now complicated enough that there should be a helper method for it

graph/src/schema/entity_type.rs
```diff
@@ -150,6 +150,10 @@ impl EntityType {
     pub fn is_object_type(&self) -> bool {
         self.schema.is_object_type(self.atom)
     }
+
+    pub fn new_vid_form(&self) -> bool {
```
lutter, line 154: I find the new/old naming a bit confusing, since in 6 months the new vid form will be old news.

What I think 'new vid form' means is that vids are ordered across entity types, maybe something like `cross_type_vids` or `cross_type_vid_order` or `strict_vid_order` would be better (and a comment here on what that means would help when reading code, similar to the comment on `SPEC_VERSION_1_3_0`) Of course, the same method in `InputSchema` should be renamed, too.

lutter (approved):

store/postgres/src/relational/prune.rs
zorancv: Yeah, at this point is clearly better that way.

graph/src/schema/entity_type.rs
zorancv: I like `strict_vid_order()` the most. ;-)

graph/src/components/store/entity_cache.rs
zorancv: I find it better. Also finally managed to remove two implementations.

graph/src/components/store/entity_cache.rs
```diff
@@ -207,9 +207,25 @@ impl EntityCache {
...
-        // makes sense when we were actually asked to read from the store
+        // makes sense when we were actually asked to read from the store.
+        // We need to remove the VID as the one from the DB  might come from
+        // a legacy subgraph that has VID autoincremented while this trait
+        // always creates it in a new style.
         debug_assert!(match scope {
-            GetScope::Store => entity == self.store.get(key).unwrap().map(Arc::new),
+            GetScope::Store => {
+                // Release build will never call this function and hence it's OK
+                // when that implementation is not correct.
```
lutter, line 217: What do you mean by 'not correct'? It seems correct to me, though a little slow because of the clone, but that's fine for debug builds

graph/src/schema/entity_type.rs
```diff
@@ -150,6 +150,14 @@ impl EntityType {
     pub fn is_object_type(&self) -> bool {
         self.schema.is_object_type(self.atom)
     }
+
+    // Changes the way the VID field is generated. It used to be autoincrement. Now its
+    // based on block number and the order of the entities in a block. The latter
+    // represents the write order across all entity types in the subgraph.
+    pub fn strict_vid_order(&self) -> bool {
```
lutter, line 157: That's the method that I would rename to `has_vid_seq`; the main reason is because `InputSchema.strict_vid_order` and `EntityType.strict_vid_order` mean slightly different things. You could also change the comment to (notice the `///`)
```rust
/// Whether the table for this entity type uses a sequence for the `vid` or whether
/// `graph-node` sets them explicitly. See  also [`InputSchema.strict_vid_order()`]
```

graph/src/schema/input/mod.rs
```diff
@@ -1585,6 +1588,10 @@ impl InputSchema {
         }?;
         Some(EntityType::new(self.cheap_clone(), obj_type.name))
     }
+
+    pub fn strict_vid_order(&self) -> bool {
+        self.inner.spec_version >= SPEC_VERSION_1_3_0
+    }
```
lutter, line 1594: This might be a good comment for this method:

```rust
/// How the values for the VID field are generated. 
/// When this is `false`, this subgraph uses the old way of autoincrementing `vid` in the database. 
/// When it is `true`, `graph-node` sets the `vid` explicitly to a number based on block number 
/// and the order in which entities are written, and comparing by `vid` will order entities by that order
```

lutter (approved):

graph/src/components/store/entity_cache.rs
zorancv: The release build is missing the `remove()` call bellow which is guarded with a debug pragma, hence the function is a NOP.

graph/src/schema/entity_type.rs
zorancv: Addressed this and next remark in a separate [PR](https://github.com/graphprotocol/graph-node/pull/5817) as this one is already merged.

PR #5784, opened 2025-01-30
graphman: validate deployment before activating

### LINKED ISSUE #4460: Disallow activating a deployment that is being copied
`graphman copy activate` should check that the deployment that is being activated is not in the process of being copied. It makes no sense to activate such a deployment (all queries against it will fail) and it can lead to internal errors.

shiyasmohd:
Fixes #4460

### DIFF
--- a/node/src/manager/commands/copy.rs
+++ b/node/src/manager/commands/copy.rs
@@ -154,7 +154,12 @@ pub async fn create(
     Ok(())
 }
 
-pub fn activate(store: Arc<SubgraphStore>, deployment: String, shard: String) -> Result<(), Error> {
+pub fn activate(
+    pools: &HashMap<Shard, ConnectionPool>,
+    store: Arc<SubgraphStore>,
+    deployment: String,
+    shard: String,
+) -> Result<(), Error> {
     let shard = Shard::new(shard)?;
     let deployment =
         DeploymentHash::new(deployment).map_err(|s| anyhow!("illegal deployment hash `{}`", s))?;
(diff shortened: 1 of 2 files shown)


node/src/manager/commands/copy.rs
```diff
@@ -167,8 +172,23 @@ pub fn activate(store: Arc<SubgraphStore>, deployment: String, shard: String) ->
...
-    println!("activated copy {}", deployment);
+
+    let (state, _, _) = match CopyState::find(pools, &shard, deployment.id.0)? {
+        Some((state, tables, on_sync)) => (state, tables, on_sync),
+        None => {
+            println!(
+                "copying is queued but has not started or no copy operation for {} exists",
+                deployment.id.0
+            );
+            return Ok(());
```
lutter, line 183: In that case, don't we want the `activate` to actually do something?

node/src/manager/commands/copy.rs
```diff
@@ -167,8 +172,23 @@ pub fn activate(store: Arc<SubgraphStore>, deployment: String, shard: String) ->
...
+        }
+    };
+    if state.finished_at.is_some() {
+        store.activate(&deployment)?;
+        println!("activated copy {}", deployment);
+    } else {
+        println!("copying is not finished yet");
```
lutter, line 190: Could you add some more information to the error, like
```
the deployment sgdNNN is being copied from sgdNNN and can therefore not be activated yet
```

lutter (changes requested):

Issue #5772, opened 2025-01-16
[Feature] Native support for parsing YAML manifests in subgraphs


alex-pakalniskis:
### Description

Subgraphs (via AssemblyScript) do not provide a robust YAML parsing experience, i.e. extracting useful information from a subgraph manifest is a chore. 

Power use-cases like the network subgraph would benefit from an enhanced YAML parsing experience specifically to extract subgraph manifest information that would power discovery in Graph CLI & Graph Explorer. 

Current high-level proposal is a dedicated Rust crate for the YAML manifest parsing logic.

### Some information to help us out

- [ ] Tick this box if you plan on implementing this feature yourself.
- [x] I have searched the issue tracker to make sure this issue is not a duplicate.

PR #5785, opened 2025-01-31
Add YAML parsing support to WASM runtime

isum:
This PR implements two host functions that are exported to WASM mappings, allowing them to easily parse YAML documents.

The main use case for this feature is to parse subgrah manifests, but it is flexible enough to allow parsing any YAML document.

Part of #5772
AssemblyScript types are defined in https://github.com/graphprotocol/graph-tooling/pull/1935

## Todos

- [x] Add WASM runtime tests
- [x] Open a PR in [graph-tooling](https://github.com/graphprotocol/graph-tooling) repo with AssemblyScript type definitions


runtime/wasm/src/module/context.rs
```diff
@@ -1188,4 +1188,64 @@ impl WasmInstanceContext<'_> {
...
+                );
+
+                true
+            });
+
+        asc_new(self, &result, gas)
+    }
+}
+
+/// For debugging, it might be useful to know exactly which bytes could not be parsed as YAML, but
```
mangas, line 1242: Do we have a better indication of where the error is? If so the error could be more relevant and show the part that broke. Not super important just a thought

mangas (approved):
LGTM

runtime/wasm/src/module/context.rs
isum: There is no straightforward way to do this nicely. All the logging context should be enough to identify the cause of a parsing error, so I will leave it as is for now

PR #5786, opened 2025-01-31
Fix bug in chunk size calculation

lutter:
When inserting entities during subgraph syncing, we split the entities that need to be written into chunks which we want to be as large as possible for best performance. Since Postgres only allows using 65k bind variables, the number of entities we can insert in one chunk is roughly 65k/#columns. The code did not account for the `causality_region` when calculating the number of columns to be inserted.

Since we had issues with this calculations in the past and it's next to impossible for operators to work around it, this also introduces an environment variable `GRAPH_STORE_INSERT_EXTRA_COLS` that operators can use to work around this issue should we ever screw up this calculation again.


zorancv (approved):
LGTM

lutter:
This was actually merged at commit 63ea9d76. Not sure why github is showing this as closed

PR #5789, opened 2025-01-31
Typo fix NEWS.md

MonkeyKing44:


lutter:
That typo is actually correct

PR #5790, opened 2025-02-02
graph, store: Use a much larger fetch_size for fdw

lutter:
Changing the fetch_size to 10,000 from its default of 100 will have a dramatic effect on the time copy operations take.

The fetch_size can be controlled with the environment variable `GRAPH_STORE_FDW_FETCH_SIZE`


zorancv (approved):
Looks good.

PR #5791, opened 2025-02-03
pass list of indexes when pruning a subgraph

### LINKED ISSUE #5787: [Bug] Pruning recreates all indexes
When pruning a subgraph by rebuilding it (i.e., when prune will remove a large part of the subgraph), the rebuilt subgraph will have all the default indexes on it rather than only the indexes that exist on the unpruned version. The reason for that is the `None` [here](https://github.com/graphprotocol/graph-node/blob/master/store/postgres/src/relational/prune.rs#L99) which should be changed to pass the list of indexes that actually exist

shiyasmohd:
Fixes #5787

### DIFF
--- a/store/postgres/src/relational/prune.rs
+++ b/store/postgres/src/relational/prune.rs
@@ -1,4 +1,4 @@
-use std::{fmt::Write, sync::Arc, time::Instant};
+use std::{collections::HashMap, fmt::Write, sync::Arc, time::Instant};
 
 use diesel::{
     connection::SimpleConnection,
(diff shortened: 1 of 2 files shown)


lutter (approved):
Nice! That fix was quick!

lutter:
I just reverted this, because I realized there is a serious issue with this.

store/postgres/src/relational/prune.rs
```diff
@@ -94,9 +97,15 @@ impl TablePair {
         if catalog::table_exists(conn, dst_nsp.as_str(), &dst.name)? {
             writeln!(query, "truncate table {};", dst.qualified_name)?;
         } else {
+            let mut list = IndexList {
+                indexes: HashMap::new(),
+            };
+            let indexes = load_indexes_from_table(conn, &src, dst_nsp.as_str())?;
```
lutter, line 103: This should use `src_nsp` because we want to read the indexes from the source, i.e. the original of the prune.

PR #5792, opened 2025-02-03
store: Try to avoid pathological batch size adjustments

lutter:
If a batch is processed very fast, that might indicate that there are very few rows actually matching the range. If this happens repeatedly, we can end up with a gigantic batch size where processing a batch takes way too long once we get into a region with denser vids.

This patch tries to address that by taking a fast batch as an indication that vids are distributed sparsely and that we should jump directly to the next existing vid rather than adjusting the batch size.


lutter:
Closing this in favor of #5807

PR #5804, opened 2025-02-06
Fix reorg threshold bug

zorancv:


graph/src/env/mod.rs
```diff
@@ -362,6 +356,18 @@ impl EnvVars {
...
+        // For testing purposes, we need to set this threshold to 0 because:
+        // 1. Many tests involve reverting blocks.
+        // 2. Blocks cannot be reverted below the reorg threshold.
+        // Therefore, during tests, we want to set the reorg threshold to 0.
+        if *TEST_WITH_NO_REORG.lock().unwrap() {
+            0
+        } else {
+            self.reorg_threshold
+        }
+    }
```
lutter, line 370: This is pretty minor, but AFAICT, we only need the ability to override the reorg threshold when debug assertions is on. Could we therefore have a variant of this function for `#[cfg(not(debug_assertions))]` that doesn't touch the mutex?

tests/tests/runner_tests.rs
```diff
@@ -109,6 +109,8 @@ fn assert_eq_ignore_backtrace(err: &SubgraphError, expected: &SubgraphError) {
 
 #[tokio::test]
 async fn data_source_revert() -> anyhow::Result<()> {
+    *TEST_WITH_NO_REORG.lock().unwrap() = true;
+
```
lutter, line 113: I am not entirely sure if these tests are run in parallel or not. But it might be better to reset `TEST_WITH_NO_REORG` after this test is done.

lutter (approved):
Nice!

tests/tests/runner_tests.rs
zorancv: The current behaviour is that all test have reorg threshold set to 0. But yeah I reset it after the end of test.

PR #5810, opened 2025-02-11
pass list of indexes when pruning a subgraph

### LINKED ISSUE #5787: [Bug] Pruning recreates all indexes

shiyasmohd:
Fixes #5787 and bug mentioned in previous [PR](https://github.com/graphprotocol/graph-node/pull/5791#discussion_r1950129026)


lutter (approved):
Nice!

PR #5811, opened 2025-02-11
Subgraph Composition: More validations

incrypto32:
This PR
- Sets a hard limit (5) on the number of datasources that can be composed together 
- Prevents nesting of subgraph datasources


zorancv:
Not directly related, but it would be nice to have a test that checks if a source entity that is aggregated would produce error.

zorancv (approved):
LGTM

PR #5815, opened 2025-02-11
Changes to VID generation after rebase with master

zorancv:


store/postgres/src/relational/dsl.rs
```diff
@@ -254,7 +254,10 @@ impl<'a> Table<'a> {
         }
 
         match column_names {
-            AttributeNames::All => cols.extend(self.meta.columns.iter()),
+            AttributeNames::All => {
+                cols.extend(self.meta.columns.iter());
+                cols.push(&*VID_COL);
+            }
```
lutter, line 260: Don't we have to include the `VID_COL` no matter what `column_names` is?

lutter (approved):

store/postgres/src/relational/dsl.rs
zorancv: I was not sure as this match side is the only one used. I've changed it to be always be added. Since this is already merged I've addressed it in another [PR](https://github.com/graphprotocol/graph-node/pull/5817).

PR #5817, opened 2025-02-12
VID related fixes and improvements

zorancv:
Addressing some of the remaining remarks in regards to the VIDs.

### DIFF
--- a/store/postgres/src/relational_queries.rs
+++ b/store/postgres/src/relational_queries.rs
@@ -2331,7 +2331,7 @@ impl<'a> InsertQuery<'a> {
         if table.has_causality_region {
             count += 1;
         }
-        if table.object.strict_vid_order() {
+        if table.object.has_vid_seq() {
             count += 1;
         }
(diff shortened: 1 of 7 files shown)


store/postgres/src/relational_queries.rs
```diff
@@ -2355,7 +2355,7 @@ impl<'a> QueryFragment<Pg> for InsertQuery<'a> {
         let out = &mut out;
         out.unsafe_to_cache_prepared();
 
-        let strict_vid_order = self.table.object.strict_vid_order();
+        let strict_vid_order = self.table.object.has_vid_seq();
```
lutter, line 2358: Very minor, but this variable should also be called `has_vid_seq`

lutter (approved):
Nice! This is good to go after addressing that one minor comment

store/postgres/src/relational_queries.rs
zorancv: Same with the variable bellow.

PR #5819, opened 2025-02-12
Remove separate entity counting step at end of copy

lutter:
We used to count all entities in the destination after all data had been copied; for large subgraphs, that can be a very slow operation, and in some cases makes the copy operation fail because the underlying fdw connection times out.

With this PR, we now update the entity count with every batch we copy.

### DIFF
--- a/store/postgres/src/copy.rs
+++ b/store/postgres/src/copy.rs
@@ -33,7 +33,7 @@ use graph::{
 use itertools::Itertools;
 
 use crate::{
-    advisory_lock, catalog,
+    advisory_lock, catalog, deployment,
     dynds::DataSourcesTable,
     primary::{DeploymentId, Site},
     relational::index::IndexList,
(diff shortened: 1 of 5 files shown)


mangas (approved):

store/postgres/src/copy.rs
```diff
@@ -512,7 +475,17 @@ impl TableState {
...
+            let count = rq::CopyEntityBatchQuery::new(self.dst.as_ref(), &self.src, start, end)?
+                .count_current()
+                .get_result::<i64>(conn)
+                .optional()?;
+            Ok(count.unwrap_or(0) as i32)
+        })?;
+
+        let count = count.unwrap_or(0);
+
+        deployment::update_entity_count(conn, &self.dst_site, count)?;
```
mangas, line 488: I'm not sure if I understand this correctly, I thought that count_current would run the count for the batch with the goal of adding it to previous. Otherwise it would be counting all at the end as it was previously, right?

store/postgres/src/relational_queries.rs
```diff
@@ -4917,6 +4929,40 @@ impl<'a> QueryId for CopyEntityBatchQuery<'a> {
...
+impl<'a> CountCurrentVersionsQuery<'a> {
+    pub fn new(copy: CopyEntityBatchQuery<'a>) -> Self {
+        Self { copy }
+    }
+}
+impl<'a> QueryFragment<Pg> for CountCurrentVersionsQuery<'a> {
+    fn walk_ast<'b>(&'b self, mut out: AstPass<'_, 'b, Pg>) -> QueryResult<()> {
+        // Generate a query
+        // with copy_cte as ( {copy} )
+        // select count(*) from copy_cte where {block_range_current}
```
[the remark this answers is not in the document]
lutter: There are tests in `store/test-store/tests/postgres/graft.rs` that run these queries by copying/grafting test subgraphs

store/postgres/src/copy.rs
lutter: The call to `count_current` wraps the `CopyEntityBatchQuery` so that the overall query returns how many current versions were copied, i.e., we end up running something like
```sql
with copy_cte(current) as (
  insert into dst(...) select * from src where ... returning block_range @> int32::MAX)
select count(*) from copy_cte where current
```
For the `select`, `copy_cte` is a table that has one boolean column, which is `true` if the row we just inserted is a current version.

The query does two things at once: (1) it copies a batch (the inner `insert` statement) and (2) returns how many rows that were copied are considered 'current', i.e. entity versions that should be counted for the entity count of the deployment. Line 488 then takes that count and adds it to the entity count for the deployment. 

With this change, we never run a count query just by itself.

PR #5829, opened 2025-02-18
Add ability to order by fields other than timestamp for aggregation entities

incrypto32:
…ate entities


lutter (approved):
Nice! Surprised by how simple the fix is.

PR #5832, opened 2025-02-18
server/graphman: Fix failing test

lutter:
I am not 100% sure that this is the right fix for that test, but it was failing for me locally.

I would also suggest that the response when it has warnings should contain success data, i.e., that it looks something like
```json
"data": {
  "deployment": {
    "reassign": {
      "success": true,
      "warnings": ["..."],
    }  }  }
```

@isum @shiyasmohd Would love your take on that.


isum:
The issue will be fixed by https://github.com/graphprotocol/graph-node/pull/5834

PR #5833, opened 2025-02-18
Remove Cosmos support

lutter:
Cosmos has been deprecated for a while and is not used anywhere anymore. Remove the support for it.


Cargo.toml
```diff
@@ -4,13 +4,27 @@ members = [
...
+    "server/http",
+    "server/index-node",
+    "server/json-rpc",
+    "server/metrics",
+    "server/websocket",
+    "store/postgres",
+    "store/test-store",
+    "substreams/substreams-head-tracker",
+    "substreams/substreams-trigger-filter",
+    "substreams/trigger-filters",
```
zorancv, line 27: Nice!

runtime/test/src/test_padding.rs
```diff
@@ -212,46 +162,6 @@ async fn test_v4_manual_padding_should_fail() {
...
-}
-
-#[tokio::test]
-async fn test_v5_u32_padding_ok() {
-    u32_padding_ok(super::test::API_VERSION_0_0_5).await
-}
-
-#[tokio::test]
-async fn test_v4_u32_padding_ok() {
-    u32_padding_ok(super::test::API_VERSION_0_0_4).await
```
zorancv, line 252: Was padding only required for cosmos or some other network that was removed?

zorancv (approved):
Cutting stuff is always nice!

runtime/test/src/test_padding.rs
lutter: Good question. I _think_ all these tests were solely for the derive macros that came in with Cosmos, but I'll check again before merging

lutter: I just checked, and the tests I removed all depended on the derive macros that I also removed since they were only used by cosmos, i.e., they were testing that the derive macros behaved properly. I'll therefore merge this PR.

PR #5834, opened 2025-02-18
Graphman: Do not use hardcoded node ID in tests

isum:
This PR ensures that the correct node ID is used by tests, as there are some differences between running tests locally and on the CI.


lutter (approved):
Nice! Thanks for the fix

PR #5836, opened 2025-02-19
Remove GraphQL subscriptions

lutter:
This removes support for GraphQL subscriptions as they are not available on the network, and the code is therefore pretty dead. Besides the actual websocket server, this also removes most of the internal gymnastics that were necessary to feed subscriptions. There's a few more simplifications that would be possible by modernizing the subgraph registrar to make it use `async`, but that's a bit hairier


mangas (approved):
🔥 🔥 🔥 🔥

dr-fusion:
1. Do we have an alternative to WS now? If yes it should be documented.
2. Current docker-compose still points to 8001 port for WS
3. How any blockchain project who need realtime updates from indexed data, going to use subgraphs now?

PR #5838, opened 2025-02-19
Add retries for FirehoseEndpoint::load_blocks_by_numbers

incrypto32:


graph/src/firehose/endpoints.rs
```diff
@@ -435,21 +436,39 @@ impl FirehoseEndpoint {
...
+            let logger_for_retry = logger.clone();
+            let logger_for_error = logger.clone();
+
+            let block = retry(retry_log_message, &logger_for_retry)
+                .limit(ENV_VARS.firehose_block_fetch_retry_limit)
+                .timeout_secs(ENV_VARS.firehose_block_fetch_timeout)
+                .run(move || {
+                    let e = endpoint_for_retry.cheap_clone();
+                    let l = logger_for_retry.clone();
+                    Box::pin(async move { e.get_block_by_number::<M>(number, &l).await })
```
zorancv, line 459: Probably a silly question, but why is pinning needed here?

zorancv (approved):
Besides the uncertainty of a need for pining, the rest looks fine.

graph/src/firehose/endpoints.rs
incrypto32: Thanks for pointing, out. It was a leftover from when i moved the method elsewhere. Updated

PR #5840, opened 2025-02-20
Reduce block range on Monad getLogs RPC call size failures

isum:
This PR white-lists the Monad Network RPC error caused by calling getLogs with a block range that is too large. It allows the block range to be reduced until a successful response is received.


incrypto32 (approved):
We might want to add an env variable or config for this in the future, so we don’t have to handle it separately for every chain with a custom error message. But this works for now.

PR #5851, opened 2025-02-25
fix substreams info endpoint

mangas:


graph/src/firehose/endpoints.rs
```diff
@@ -79,8 +81,13 @@ impl NetworkDetails for Arc<FirehoseEndpoint> {
...
     async fn provides_extended_blocks(&self) -> anyhow::Result<bool> {
         let info = self.clone().info().await?;
+        let pred = if info.chain_name.contains("arbitrum") {
+            |x| x == "extended" || x == "hybrid"
+        } else {
+            |x| x == "extended"
+        };
 
-        Ok(info.block_features.iter().all(|x| x == "extended"))
+        Ok(info.block_features.iter().all(pred))
```
isum, line 90: I am not sure if this will work as expected. For example, `arbone.firehose.pinax.network:443` returns only `block_features: ["hybrid"]` which is fine, but `arb-one.streamingfast.io:443` returns `block_features: ["hybrid", "extended@22207818"]` which will not pass the test.

You can replace the `.all()` call with `.any()`, or update the check to `|x| x == "hybrid" || x == "extended" || x.starts_with("extended@")` but event this may not cover all the differences in the Firehose responses.

isum (approved):

PR #5857, opened 2025-02-27
Make VID element internal to the Entity

zorancv:


graph/src/data/store/mod.rs
```diff
@@ -1062,6 +1062,13 @@ impl Entity {
     }
 }
 
+/// Checks equality of two entities while ignoring the VID fields
+impl PartialEq for Entity {
+    fn eq(&self, other: &Self) -> bool {
+        self.sorted_ref() == other.sorted_ref()
+    }
+}
+
```
incrypto32, line 1071: Using the `sorted_ref()` instead of defaulting the the `PartialEq` implementation of `Object` can be inefficient. The `PartialEq` implentation from `Derive` macro delegates it to the `Object` implementation which compares Atoms directly which is effiecient. With `sorted_ref` is doing some heap allocations and copies which could be significant in our production cluster. Also the sorting isn't important for the comparison when using the implementation from `Object`

incrypto32: So maybe it'd be better to do this in there, instead of Entity.

incrypto32 (dismissed):

graph/src/data/store/mod.rs
```diff
@@ -1062,6 +1062,13 @@ impl Entity {
     }
 }
 
+/// Checks equality of two entities while ignoring the VID fields
+impl PartialEq for Entity {
+    fn eq(&self, other: &Self) -> bool {
+        self.sorted_ref() == other.sorted_ref()
```
fordN, line 1068: I'm not following how this is ignoring VID? Because of the updates to .get() to skip VID fields??

zorancv: `sorted_ref()` filters out VIDs

graph/src/data/store/mod.rs
zorancv: A good catch - made it better performant, on par with the current comparison.

graph/src/data/store/mod.rs
zorancv: With the performance improvements that part is more clear now.

graph/src/data/store/mod.rs
lutter: Yes, I'd be pretty nervous about the overhead here, too since `sorted_ref` clones; at the very least, it should do the non-optimal comparison from `Object`, i.e. something like
```rust
self.iter().all(|(key, value)| other.get(key).map_or(false, |o| o == value))
```

graph/src/data/store/mod.rs
```diff
@@ -740,7 +740,7 @@ lazy_static! {
 }
 
 /// An entity is represented as a map of attribute names to values.
-#[derive(Clone, CacheWeight, PartialEq, Eq, Serialize)]
+#[derive(Clone, CacheWeight, Eq, Serialize)]
 pub struct Entity(Object<Value>);
 
 impl<'a> IntoIterator for &'a Entity {
```
lutter, line 746: this also needs to be changed so that `entity.iter()` and `entity.into_iter()` do not show the `vid`, something like 
```rust
impl<'a> IntoIterator for &'a Entity {
    type Item = (Word, Value);

    type IntoIter = intern::ObjectOwningIter<Value>;

    fn into_iter(self) -> Self::IntoIter {
        self.0.clone().into_iter().filter(|(k,_)| k != VID_FIELD)
    }
}
```

lutter (changes requested):

graph/src/util/intern.rs
```diff
@@ -308,6 +308,45 @@ impl<V> Object<V> {
...
+            self.len()
+        };
+        let len2 = if let Some(to_ignore) = other.pool.lookup(ignore_key) {
+            other.len_ignore_atom(&to_ignore)
+        } else {
+            other.len()
+        };
+        if len1 != len2 {
+            return false;
+        }
```
lutter, line 335: This is very minor, but since you need to traverse the list of key/value pairs to determine the length, there's not much value in trying to short-circuit the comparison based on length since you already need to load everything into cache. I would just leave the length-based optimization out and go straight to comparing based on values.

graph/src/data/store/mod.rs
lutter: You should also add unit tests that show that the various methods behave well when there is a `vid` (e.g., that iteration and `sorted` do not include the `vid`)

zorancv: Best I could do is use `retain()`. Can we postpone the UT for later moment?

graph/src/util/intern.rs
zorancv: On a second thought the size comparison avoids false positive when elements in `self` are subset of the ones ni `other`, so I'll keep it.

graph/src/data/store/mod.rs
zorancv: Beautiful! That removes two copies - the original clone and my retain! BTW added the test.

lutter (approved):
Nice!

encalypto:
Merging this as per offline discussion with @fordN in order to test this sooner, given all issues were resolved & PR has been approved

PR #5863, opened 2025-03-03
Stop pretending that there are subgraphs that don't support PoI

lutter:
Subgraphs have in fact supported proof of indexing for almost 5 years since commit 40fd330c and we can therefore stop pretending that `WritableStore.supports_proof_of_indexing` would ever return `false`


mangas (approved):
Loving the hate driven development

PR #5871, opened 2025-03-05
store: Remove unused SSL support

lutter:
We could, in theory, establish SSL connections for the notification listener. Since that uses the same PG URL as the main connections for data through diesel, it was actually not possible to use SSL as we don't support it for our diesel connections.

Until we are able to set up SSL connections across the board, remove the defunct SSL support in the notification listener.


lutter:
On second thought, let's put a pin into this for now

lutter:
This would undo PR #3503 and should therefore not be merged

PR #5874, opened 2025-03-06
enhance err msg for source subgraph manifest resolution in composition

shiyasmohd:
When getting error messages in cli when deploying composition subgraph, the err msg doesn't exactly point the source subgraph deployment id which is giving error. 
This PR solves this issue.


shiyasmohd:
Got a bit messed up locally with conflict & signature. Moved this PR to #5950

PR #5875, opened 2025-03-06
graphman: disable load management when using graphman

incrypto32:
Closes [#5850](https://github.com/graphprotocol/graph-node/issues/5850)


node/src/bin/manager.rs
```diff
@@ -1030,6 +1031,9 @@ impl Context {
 
 #[tokio::main]
 async fn main() -> anyhow::Result<()> {
+    // Disable load management for graphman commands
+    env::set_var("GRAPH_LOAD_THRESHOLD", "0");
```
incrypto32, line 1035: didn't want to add a boolean flag in `LoadManager::new` just for this, so just hardset the env var here

lutter (approved):
Out of curiosity, how was the load manager interfering with `graphman query`?

PR #5876, opened 2025-03-06
Subgraph Composition: Option to force rpc to fetch block ptrs

incrypto32:
Composable subgraphs need to fetch block ptrs to handle reorgs, when its not in the cache graph-node uses firehose endpoint to fetch the block using a `SingleBlockRequest` but this is currently slower than RPC.
This PR
- Adds an option to force using RPC to fetch block ptrs when an RPC endpoint is available
- Parralize fetching blocks when using firehose
- Added some refactoring in `FirehoseEndpoints` code

### DIFF
--- a/chain/ethereum/src/env.rs
+++ b/chain/ethereum/src/env.rs
@@ -91,7 +91,7 @@ pub struct EnvVars {
     /// This is a comma separated list of chain ids for which the gas field will not be set
     /// when calling `eth_call`.
     pub eth_call_no_gas: Vec<String>,
-    /// Set by the flag `GRAPH_ETHEREUM_FORCE_RPC_FOR_BLOCK_PTRS`. Off by default.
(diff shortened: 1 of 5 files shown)


chain/ethereum/src/env.rs
```diff
@@ -192,4 +197,6 @@ struct Inner {
     genesis_block_number: u64,
     #[envconfig(from = "GRAPH_ETH_CALL_NO_GAS", default = "421613,421614")]
     eth_call_no_gas: String,
+    #[envconfig(from = "GRAPH_ETHEREUM_FORCE_RPC_FOR_BLOCK_PTRS", default = "true")]
```
mangas, line 200: The docs say this should be off by default

incrypto32: Ah thats a mistake in the docs, updated it

Issue #5892, opened 2025-03-16
[Bug] Incorrect version label on graph-node logs


jhjhjh94:
### Bug report

Incorrect version label on graph-node logs during initialization

Graph node version 0.36.1 shows `Mar 12 10:55:09.833 INFO Graph Node version: 0.36.0 (2025-01-28)`
Graph node version 0.37.0 shows `Mar 12 10:58:35.250 INFO Graph Node version: 0.36.0 (2025-03-06)`

### Relevant log output

```Shell

```

### Some information to help us out

- [x] Tick this box if this bug is caused by a regression found in the latest release.
- [ ] Tick this box if this bug is specific to the hosted service.
- [x] I have searched the issue tracker to make sure this issue is not a duplicate.

### OS information

None

fordN:
Do you know if this is this resolved? @DaMandal0rian

PR #5896, opened 2025-03-18
Fix: graphnode version for 0.37.x

DaMandal0rian:
This should fix the graphnode version [bug](https://github.com/graphprotocol/graph-node/issues/5892) in the next release.


lutter:
It looks like this inadvertently downgrades `base64` from 0.22.1 to 0.13.1. Can you regenerate `Cargo.lock` so that CI passes?

PR #5907, opened 2025-03-25
Copy private data sources in batches

lutter:
Currently, private data sources are copied one-by-one which can be very slow for deployments with huge numbers of data sources (say, 100's of thousands)

This PR changes that so that we copy data sources in bulk

### DIFF
--- a/store/postgres/src/dynds/private.rs
+++ b/store/postgres/src/dynds/private.rs
@@ -1,4 +1,4 @@
-use std::ops::Bound;
+use std::{collections::HashMap, ops::Bound};
 
 use diesel::{
     pg::sql_types,
@@ -252,32 +252,11 @@ impl DataSourcesTable {
             .order_by(&self.vid)
             .load::<Tuple>(conn)?;
 
+        let manifest_map =
+            ManifestIdxMap::new(src_manifest_idx_and_name, dst_manifest_idx_and_name);
         let mut count = 0;
-        for (block_range, src_manifest_idx, param, context, causality_region, done_at) in src_tuples
-        {
-            let name = &src_manifest_idx_and_name
-                .iter()
-                .find(|(idx, _)| idx == &src_manifest_idx)
-                .with_context(|| {
-                    anyhow!(
-                        "the source {} does not have a template with index {}",
-                        self.namespace,
-                        src_manifest_idx
-                    )
-                })?
-                .1;
-            let dst_manifest_idx = dst_manifest_idx_and_name
-                .iter()
-                .find(|(_, n)| n == name)
-                .with_context(|| {
-                    anyhow!(
-                        "the destination {} is missing a template with name {}. The source {} created one at block {:?}",
-                        dst.namespace,
-                        name, self.namespace, block_range.0
-                    )
-                })?
-                .0;
-
+        for (block_range, src_idx, param, context, causality_region, done_at) in src_tuples {
+            let dst_idx = manifest_map.dst_idx(src_idx)?;
             let query = format!(
                 "\
              insert into {dst}(block_range, manifest_idx, param, context, causality_region, done_at)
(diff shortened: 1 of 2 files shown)


store/postgres/src/dynds/private.rs
```diff
@@ -361,3 +321,141 @@ impl DataSourcesTable {
...
+impl ManifestIdxMap {
+    fn new(src: &[(i32, String)], dst: &[(i32, String)]) -> Self {
+        let map = src
+            .iter()
+            .map(|(src_idx, src_name)| {
+                (
+                    *src_idx,
+                    (
+                        dst.iter()
+                            .find(|(_, dst_name)| src_name == dst_name)
```
zorancv, line 341: Did you consider converting the dst slice to a HashMap? How realistic is the scenario with the maximum of 10'000+ elements? Is a square algorithm reasonable fast for that size? Maybe lower the upper bound with 1-2 orders of magnitude?

store/postgres/src/dynds/private.rs
```diff
@@ -361,3 +321,141 @@ impl DataSourcesTable {
...
+            Bound::Unbounded => i32::MAX,
+        };
+        self.idx = map.dst_idx(self.idx, src_nsp, src_created, dst_nsp)?;
+        Ok(self)
+    }
+}
+
+struct CopyDsQuery<'a> {
+    dst: &'a DataSourcesTable,
+    dss: &'a [DsForCopy],
```
zorancv, line 410: Very minor, but isn't `dsc` a better name for this one?

store/postgres/src/dynds/private.rs
```diff
@@ -361,3 +321,141 @@ impl DataSourcesTable {
...
+        // unclamp block range if it ends beyond target block
+        match self.block_range.1 {
+            Bound::Included(block) if block > target_block => self.block_range.1 = Bound::Unbounded,
+            _ => { /* use block range as is */ }
+        }
+        // Translate manifest index
+        let src_created = match self.block_range.0 {
+            Bound::Included(block) => block,
+            Bound::Excluded(block) => block + 1,
+            Bound::Unbounded => i32::MAX,
```
zorancv, line 401: I don't follow, why is this not 0?

store/postgres/src/dynds/private.rs
lutter: Good point - `src` and `dst` should be relatively small since they are data source declarations from the manifest, but it won't hurt to build a map here. Just added a commit that does that

store/postgres/src/dynds/private.rs
lutter: I was thinking of this as `DataSourceS`

store/postgres/src/dynds/private.rs
```diff
@@ -361,3 +321,141 @@ impl DataSourcesTable {
...
+    fn src_to_dst(
+        mut self,
+        target_block: BlockNumber,
+        map: &ManifestIdxMap,
+        src_nsp: &Namespace,
+        dst_nsp: &Namespace,
+    ) -> Result<Self, StoreError> {
+        // unclamp block range if it ends beyond target block
+        match self.block_range.1 {
+            Bound::Included(block) if block > target_block => self.block_range.1 = Bound::Unbounded,
```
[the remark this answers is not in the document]
lutter: Yes, you are right. I added a match for that (hopefully with the right `if`, this hurts my brain)

store/postgres/src/dynds/private.rs
lutter: Also true - I was thinking of the upper bound, but lower bound unbounded should mean 0

store/postgres/src/dynds/private.rs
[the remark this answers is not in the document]
zorancv: The `if` seems fine to me.

zorancv (approved):
Nice!

PR #5911, opened 2025-03-25
Validate if the graft chain contains any incompatible spec version

incrypto32:


lutter:
Is this to avoid an error when going from old stable hash to fast stable hash? Could we just convert the old format into the new one and let people graft across these versions? As long as the conversion is deterministic, the resulting PoI should also be deterministic since the graft point is fixed

incrypto32:
@lutter this is a completely unrelated PR, this is to prevent people from using subgraphs that graft on subgraphs with specVersion lower than 1.3.0 as source subgraphs for composed subgraphs to ensure consistency of the vid ordering

graph/src/data_source/subgraph.rs
```diff
@@ -286,15 +335,6 @@ impl UnresolvedDataSource {
         let kind = self.kind.clone();
         let source_manifest = self.resolve_source_manifest::<C>(resolver, logger).await?;
         let source_spec_version = &source_manifest.spec_version;
-
-        if source_manifest
```
zorancv, line 290: Do we enable nested source subgraphs?

graph/src/data_source/subgraph.rs
```diff
@@ -270,6 +270,55 @@ impl UnresolvedDataSource {
             .map(Arc::new)
     }
 
+    /// Recursively verifies that all grafts in the chain meet the minimum spec version requirement
+    async fn verify_graft_chain<C: Blockchain>(
```
zorancv, line 274: How about `verify_graft_chain_sourcable()` ? Also maybe mention in the comment above the purpose of the verification.

graph/src/data_source/subgraph.rs
```diff
@@ -270,6 +270,55 @@ impl UnresolvedDataSource {
...
+        logger: &Logger,
+        graft_chain: &mut Vec<String>,
+    ) -> Result<(), Error> {
+        // Add current manifest to graft chain
+        graft_chain.push(manifest.id.to_string());
+
+        // Check if current manifest meets spec version requirement
+        if manifest.spec_version < SPEC_VERSION_1_3_0 {
+            return Err(anyhow!(
+                "Subgraph manifest spec version {} is not supported, minimum supported version is {}. Graft chain: {}",
```
zorancv, line 286:
```suggestion
                "Subgraph with a spec version {} is not supported for a subgraph source, minimum supported version is {}. Graft chain: {}",
```

graph/src/data_source/subgraph.rs
```diff
@@ -303,15 +343,17 @@ impl UnresolvedDataSource {
             ));
         }
 
-        let pruning_enabled = match source_manifest.indexer_hints.as_ref() {
```
zorancv, line 306: Do you have a immutable entities check in another PR?

lutter:
> @lutter this is a completely unrelated PR, this is to prevent people from using subgraphs that graft on subgraphs with specVersion lower than 1.3.0 as source subgraphs for composed subgraphs to ensure consistency of the vid ordering

Aah .. thanks for explaining. I totally missed that

graph/src/data_source/subgraph.rs
incrypto32: Yes, its in #5909

graph/src/data_source/subgraph.rs
incrypto32: Addressed

graph/src/data_source/subgraph.rs
incrypto32: Made the change

graph/src/data_source/subgraph.rs
incrypto32: We do not

zorancv (approved):
Nice!

PR #5913, opened 2025-03-26
Index creation fixes

lutter:
This fixes two issues with postponed index creation:

(1) When the table has a user-supplied `block` column, we created the index on `block$` twice
(2) We did not create postponed indexes on `(<attr>, block$)`

### DIFF
--- a/store/postgres/src/relational/index.rs
+++ b/store/postgres/src/relational/index.rs
@@ -784,7 +784,8 @@ impl IndexList {
         table_name: &String,
         dest_table: &Table,
         postponed: bool,
-        concurrent_if_not_exist: bool,
+        concurrent: bool,
+        if_not_exists: bool,
     ) -> Result<Vec<(Option<String>, String)>, Error> {
         let mut arr = vec![];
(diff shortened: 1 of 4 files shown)


store/postgres/src/relational/index.rs
```diff
@@ -667,9 +685,7 @@ impl CreateIndex {
                         }
                         Expr::Vid => (),
                         Expr::Block => {
-                            if !column_exists(cols, &"block".to_string()) {
-                                return false;
-                            }
+                            return dest_table.immutable;
```
zorancv, line 688:
```suggestion
                            if !dest_table.immutable {
                                return false;
                            }
```

zorancv: We need to continue with other columns in case the condition is true.

zorancv (changes requested):
Thank you for fixing those issue!

store/postgres/src/relational/index.rs
lutter: Doh .. yes, you are right.

zorancv (approved):

PR #5918, opened 2025-03-28
Timeout copy batches that take too long and copy tables in parallel

lutter:
Our estimation of batch sizes is generally good and stays within the prescribed bounds, but there are cases where proper estimation of the batch size is nearly impossible since the size of the rows in the table jumps sharply at some point        that is hard to predict. This mechanism ensures that if our estimation is wrong, the consequences aren't too severe.

That's what the first three commits do; the rest of this PR changes how we copy so that we can copy the tables for a deployment in parallel. The copying parallelizes opportunistically, i.e., it will always copy at least one table, and more if there are database connections available and the configuration allows copying more than one table

### DIFF
--- a/graph/src/env/store.rs
+++ b/graph/src/env/store.rs
@@ -86,7 +86,7 @@ pub struct EnvVarsStore {
     pub batch_timeout: Option<Duration>,
 
     /// The number of workers to use for batch operations. If there are idle
-    /// connectiosn, each subgraph copy operation will use up to this many
+    /// connections, each subgraph copy operation will use up to this many
     /// workers to copy tables in parallel. Defaults to 1 and must be at
     /// least 1
     pub batch_workers: usize,


graph/src/env/store.rs
```diff
@@ -81,6 +81,16 @@ pub struct EnvVarsStore {
     /// The default is 180s.
     pub batch_target_duration: Duration,
 
+    /// Cancel and reset a batch copy operation if it takes longer than
+    /// this. Set by `GRAPH_STORE_BATCH_TIMEOUT`. Unlimited by default
+    pub batch_timeout: Option<Duration>,
+
+    /// The number of workers to use for batch operations. If there are idle
+    /// connectiosn, each subgraph copy operation will use up to this many
```
zorancv, line 89:
```suggestion
    /// connections, each subgraph copy operation will use up to this many
```

lutter: Fixed the typo

zorancv (approved):
Hope it helps the copies.

store/postgres/src/copy.rs
```diff
@@ -653,78 +853,138 @@ impl Connection {
...
+        let mut workers = Vec::new();
+        while !state.unfinished.is_empty() || !workers.is_empty() {
+            // We usually add at least one job here, except if we are out of
+            // tables to copy. In that case, we go through the `while` loop
+            // every time one of the tables we are currently copying
+            // finishes
+            if let Some(worker) = self.default_worker(&mut state, &progress)? {
+                workers.push(worker);
+            }
+            loop {
```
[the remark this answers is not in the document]
lutter: That wouldn't be correct - we can only call `self.default_worker` when `self.conn.is_some()`, i.e., once per `while` loop. The `loop { .. }` is about trying to get more workers than just the one we always have.

lutter:
Messed up my git commands, this was merged at b3543bbb

PR #5926, opened 2025-04-03
Store startup fixes

lutter:
With multiple nodes, they could race each other and cause database errors.

I've reworked the startup code to be (a) race-proof (I hope) and (b) easier to follow. Any node that wants to run database setup now gets a lock on the primary, and runs all code needed for setup while holding that lock. That way, nodes can't interfere with each other. In the common case, where there are no database changes, be it because of migrations, configuration changes, or code changes that map different tables, the time during which any node holds the lock is very brief. This could be further optimized but let's first see how this performs in practice.

### DIFF
--- a/store/postgres/src/connection_pool.rs
+++ b/store/postgres/src/connection_pool.rs
@@ -435,27 +435,27 @@ impl fmt::Debug for ConnectionPool {
     }
 }
 
-/// The name of the pool, mostly for logging, and what purpose it serves.
+/// The role of the pool, mostly for logging, and what purpose it serves.
 /// The main pool will always be called `main`, and can be used for reading
 /// and writing. Replica pools can only be used for reading, and don't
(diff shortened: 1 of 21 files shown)


store/postgres/src/connection_pool.rs
```diff
@@ -1508,4 +1503,188 @@ impl PoolCoordinator {
...
+        }
+
+        // Everything here happens under the migration lock. Anything called
+        // from here should not try to get that lock, otherwise the process
+        // will deadlock
+        debug!(self.logger, "Waiting for migration lock");
+        let res = with_migration_lock(&mut pconn, |_| async {
+            debug!(self.logger, "Migration lock acquired");
+
+            // While we were waiting for the migration lock, another thread
```
incrypto32, line 1661: NICE!

incrypto32 (approved):

PR #5929, opened 2025-04-07
Check grafting from pre 0.0.6 to 0.0.6 and later

zorancv:
This PR tries to check the consistency of the POI implementation for grafting of subgraph in case there is a transition from a spec version older than 0.0.6 towards the 0.0.6 or newer. The grafting block has a conversion of the legacy hashing algorithm to the new one, and here we test that this conversion did not change.

On order to keep the POI with the expected values, we need to make sure that anvil produces the same genesis block hence a fixed timestamp for starting it. Also both local start and remote one need to have same parameters. In addition to that we need to wait until the last checked grafted block is mined before we start the integration tests.

Moreover there is need for specifying an `index` parameter in the fetching POI query in order to avoid error when a local environment variable `GRAPH_POI_ACCESS_TOKEN` is set as that would set it to zero address and produce inconsistent POI value.


tests/tests/integration_tests.rs
```diff
@@ -915,6 +1001,25 @@ async fn test_multiple_subgraph_datasources(ctx: TestContext) -> anyhow::Result<
...
+    while start.elapsed() < STATUS_WAIT {
+        let latest_block = Contract::latest_block().await;
+        if let Some(latest_block) = latest_block {
+            if let Some(number) = latest_block.number {
+                if number >= block_number.into() {
+                    return true;
+                }
+            }
+        }
+        let one_sec = time::Duration::from_secs(1);
```
lutter, line 1017: This would be better as a constant

tests/tests/integration_tests.rs
```diff
@@ -915,6 +1001,25 @@ async fn test_multiple_subgraph_datasources(ctx: TestContext) -> anyhow::Result<
...
+        thread::sleep(one_sec);
```
lutter, line 1018: Probably not a big deal here, but you shouldn't use `thread::sleep` in async code; use `tokio::time::sleep` instead.

tests/tests/integration_tests.rs
```diff
@@ -796,6 +806,82 @@ async fn test_remove_then_update(ctx: TestContext) -> anyhow::Result<()> {
     Ok(())
 }
 
+async fn test_subgraph_grafting(ctx: TestContext) -> anyhow::Result<()> {
+    async fn get_bloock_hash(block_number: i32) -> Option<String> {
```
lutter, line 810: type 'bloock'

tests/tests/integration_tests.rs
```diff
@@ -796,6 +806,82 @@ async fn test_remove_then_update(ctx: TestContext) -> anyhow::Result<()> {
...
+        "0x6053124bc1042dfc9c153b49c626f11860a1bc34d0fde384526a2dbed909d242",
+        "0xca535e3a5be9b93012f6173b33cf9c053cac8028dda9c398b619b67f53a9e1a4",
+    ];
+
+    for i in 1..4 {
+        let block_hash = get_bloock_hash(i).await.unwrap();
+        // We need to make sure that the preconditions for POI are fulfiled
+        // namely that the anvil produces the proper block hashes for the
+        // blocks of which we will check the POI
+        assert_eq!(block_hash, block_hashes[(i - 1) as usize]);
```
lutter, line 850: I am confused by this - the comment made me think we would check anvil, but we are querying graph-node. Maybe the comment should just not mention anvil

tests/integration-tests/grafted/subgraph.yaml
```diff
@@ -0,0 +1,25 @@
+specVersion: 0.0.5
+description: Grafted Subgraph
```
lutter, line 2: I find this naming confusing - 'grafted subgraph' implies that this subgraph is grafted onto something else. Maybe just calling it 'base subgraph' would be better. And what's currently called 'grafting subgraph' could then be called 'grafted subgraph', but leaving that as 'grafting subgraph' is fine.

lutter (changes requested):
Nice test! Just a few minor comments

tests/tests/integration_tests.rs
zorancv: For current usage is irrelevant as all tests need to wait, but if ever used elsewhere it could matter. Changing it.

tests/tests/integration_tests.rs
zorancv: Right. The 'blockchain' is more appropriate here,

tests/integration-tests/grafted/subgraph.yaml
zorancv: Agree, those names are more intuitive.

lutter (approved):

PR #5936, opened 2025-04-08
graph, store: Remove GRAPH_STORE_LAST_ROLLUP_FROM_POI

lutter:
This flag was only meant as a safety switch in case the fixed behavior caused trouble. Since it's not been needed in several months, it's safe to remove it.


zorancv (approved):
The best PRs remove code

PR #5937, opened 2025-04-09
store: Try to spawn copy workers more frequently

lutter:
We used to only try to spawn a new copy worker when we were finished copying a table. But that can take a long time, and in between some connections might have become available. We now check every few minutes if a connection is available and spawn a new worker if it is

### DIFF
--- a/store/postgres/src/copy.rs
+++ b/store/postgres/src/copy.rs
@@ -32,12 +32,13 @@ use diesel::{
 };
 use graph::{
     constraint_violation,
-    futures03::future::select_all,
+    futures03::{future::select_all, FutureExt as _},
     prelude::{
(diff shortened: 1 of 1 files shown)


store/postgres/src/copy.rs
```diff
@@ -705,18 +721,15 @@ impl CopyTableWorker {
...
-    ) -> Result<Self, StoreError> {
+    async fn run(mut self, logger: Logger, progress: Arc<CopyProgress>) -> WorkerResult {
         let object = self.table.dst.object.cheap_clone();
         graph::spawn_blocking_allow_panic(move || {
             self.result = self.run_inner(logger, &progress);
             self
         })
         .await
         .map_err(|e| constraint_violation!("copy worker for {} panicked: {}", object, e))
+        .into()
```
zorancv, line 732: nice!

zorancv (approved):
nifty!

Issue #5449, opened 2024-05-29
[Bug] Impossible combination of entity operations


paymog:
### Bug report

A subgraph that was running without issues on v0.34.0 suddenly started failing in v0.35.0. This subgraph is deployed across many networks and they all started failing with this issue which suggests that this is a regression in v0.35.0.

### Relevant log output

```Shell
May 28 13:26:21.945 ERRO Subgraph failed with non-deterministic error: Failed to transact block operations: internal constraint violated: impossible combination of entity operations: Remove { key: EntityKey(SplitRecipient[0x7c29ca34b44d388ab031ecce7781f2420e1e5c99-0xfa9aad02ffede509520e27ef329ee28871a76828-5], cr=0), block: 15264023 } and then Remove { key: EntityKey(SplitRecipient[0x7c29ca34b44d388ab031ecce7781f2420e1e5c99-0xfa9aad02ffede509520e27ef329ee28871a76828-5], cr=0), block: 15267303 }, retry_delay_s: 108, attempt: 0, sgd: 1, subgraph_id: QmcpChELh7eJShPHvG5zLBUYBsBQby9KZ8roh7BrT2Yp5B, component: SubgraphInstanceManager
```

### IPFS hash

QmcpChELh7eJShPHvG5zLBUYBsBQby9KZ8roh7BrT2Yp5B

### Some information to help us out

- [X] Tick this box if this bug is caused by a regression found in the latest release.
- [ ] Tick this box if this bug is specific to the hosted service.
- [X] I have searched the issue tracker to make sure this issue is not a duplicate.

### OS information

Linux

paymog:
we also see `Failed to transact block operations: internal constraint violated: Batches must go forward. Can't append a batch with block pointer #114200817` as another issue happening on these subgraphs but this one happens less reliably.

paymog:
Seems like this issue might be related to batching. Trying to bisect and the issue doesn't seem to happen reliably on any commits.

paymog:
Now I'm thinking this might be a subgraph bug that wasn't revealed until we upgraded to v0.35.0

paymog:
Setting `GRAPH_STORE_WRITE_BATCH_SIZE=0` seems to resolve the issue

paymog:
[...]

PR #5940, opened 2025-04-10
graph: Allow remove followed by remove in write batches

lutter:
If subgraphs delete entities without checking if they exist, we get two removes in a row. In the database, the second remove would just lead to a query that changes nothing. We'll do the same when putting a write batch together.

Fixes https://github.com/graphprotocol/graph-node/issues/5449


zorancv (approved):
Makes sense!

PR #5941, opened 2025-04-11
Fix reorg threshold

zorancv:
Addressing the review comments from https://github.com/graphprotocol/graph-node/pull/5804 and squashed previous commits to one verified commit.

### DIFF
--- a/graph/src/env/mod.rs
+++ b/graph/src/env/mod.rs
@@ -369,6 +369,7 @@ impl EnvVars {
             .filter(|x| !x.is_empty())
             .collect()
     }
+    #[cfg(debug_assertions)]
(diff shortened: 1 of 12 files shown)


graph/src/env/mod.rs
```diff
@@ -17,6 +18,7 @@ use crate::{
 
 lazy_static! {
     pub static ref ENV_VARS: EnvVars = EnvVars::from_env().unwrap();
+    pub static ref TEST_WITH_NO_REORG: Mutex<bool> = Mutex::new(false);
```
lutter, line 21: This is minor, but I think we could also compile this out with `#[cfg(not(debug_assertions))]`

lutter (approved):
Nice, I like how this turned out

PR #5942, opened 2025-04-11
store: Do not drop the default copy connection when running out of work

lutter:
When the default worker finished while other workers were still copying, and there was no more work to do, we inadvertently dropped the default connection.


fordN (approved):
👍

lutter:
Screwed up git commands. This is merged at e6f55b174

PR #5943, opened 2025-04-12
Improve error classification in subgraph runner and classify db unique constraint violations as deterministic

lutter:


zorancv (approved):
Nice improvements!

PR #5952, opened 2025-04-17
Split up the store::connection_pool module

lutter:
That module was huge and becoming very unwieldy.

There are no functional changes in this PR, just moving code around


zorancv (approved):
Confirming no code change. Just rename and split in files.

PR #5953, opened 2025-04-18
Simplify the SubgraphRunner a little

lutter:
There's lots more that could and should be done, but it's a start to make the runner more readable


mangas (approved):
Definitely seems better than before, would be great to break into functions that can be tested, ideally that don't require `self` or side effects.

PR #5955, opened 2025-04-21
Asyncify some functions in EthereumAdapter

lutter:


chain/ethereum/src/ethereum_adapter.rs
```diff
@@ -1258,169 +1256,147 @@ impl EthereumAdapterTrait for EthereumAdapter {
...
+
+        retry(retry_log_message, &logger)
+            .redact_log_urls(true)
+            .limit(ENV_VARS.request_retries)
+            .timeout_secs(ENV_VARS.json_rpc_timeout.as_secs())
+            .run(move || {
+                Box::pin(web3.eth().block_with_txs(BlockId::Hash(block_hash)))
+                    .compat()
+                    .from_err()
+                    .compat()
```
incrypto32, line 1350:
```suggestion
                let web3 = web3.clone();
                async move {
                    web3.eth()
                        .block_with_txs(BlockId::Hash(block_hash))
                        .await
                        .map_err(Error::from)
                }
```

incrypto32 (approved):
LGTM! There's still some leftover futures01 usage, but we can tackle that later.

chain/ethereum/src/ethereum_adapter.rs
lutter: Nice! I am incorporating that into the PR

lutter:
Yeah, we'll have to make a few more passes over the codebase to get rid of all the futures01

PR #5968, opened 2025-04-22
store: Drop the vid sequence with 'cascade'

lutter:
Since it is used for the default value, we get an error

```
  sequence sgdNNN.<table>_vid_seq membership in
  replication set default depends on sequence sgdNNN.<table>_vid_seq
```

That makes all pruning by rebuilding tables fail


mangas (approved):

lutter:
> Is there any way to add a regression test for this?

Notreally - this only happened if the database used `pglogical` and if the table had been added to a replication set. So pretty specific and sorta niche setup

PR #5970, opened 2025-04-23
graph, store: Make sure vid batching works with large vids

lutter:
Changing to the new vid scheme of `block_num << 32 + sequence_num` revealed some numerical problems in the batching code.


graph/src/util/ogive.rs
```diff
@@ -102,35 +100,61 @@ impl Ogive {
...
+    //
+    // Numerical consideration: in these calculations, we need to be careful
+    // to never convert one of the points directly to f64 since they can be
+    // so large that the conversion from i64 to f64 loses precision. That
+    // loss of precision can cause the convex combination of `points[j]` and
+    // `points[j+1]` above to lie outside of that interval when `(points[j]
+    // as f64) as i64 < points[j]`
+    //
+    // We therefore try to only convert differences between points to f64
+    // which are much smaller.
```
zorancv, line 144: Here the assumption is that differences of the subsequent points is never more than 2'000'000 blocks, as we have only 21 bits of the mantissa available (53 of f64 - 32 of the shift)? If that's not absolutely guaranteed I would suggest to convert all the calculations to work with `i128`. The `bins_size` could be represented as a ratio of two integers. The only drawback are somewhat slow divisions...

zorancv (approved):
This is clear improvement over the current situation and is equivalent to it from algebraic point of view, so I approved it with one remark for consideration in the future. Nice that you added a regression test case!

graph/src/util/ogive.rs
lutter: That's a really good point. I'll leave this as-is for now, but if I ever need to touch it, I'll adopt your suggestion.

Issue #5977, opened 2025-04-30
[Feature] Graph Node Dev Mode: Minimal Setup Local Subgraph Deployment with --watch Support


incrypto32:
### Description

Graph Node Dev Mode will make local subgraph development way easier. This mode will let devs iterate fast by watching file changes and auto-redeploying subgraphs no more messing around with IPFS or PostgreSQL config.

#### Goal
Let developers develop and deploy subgraphs locally with:
- No IPFS dependency 
- No manual PostgreSQL setup ( Still need to install postgres )
- Super minimal configuration

#### Current Scope
- Add `--watch` flag to gnd
- When used:
  - Watches the build/ directory for any chanes
  - Auto redeploys the subgraph to local graph-node whenever files change
  - Shows the subgraph and GraphiQL URLs for immediate testing
- Subgraphs deploy straight from filesystem (no IPFS)
- Rebuilds still require manual triggering via `gnd build`
- Use pg_temp to:
  - Make a temporary Postgres instance inside build folder
  - Eliminate user database configuration

#### Tasks
- [x] Create seperate graphnode-dev binary
- [x] Refactor existing main function for reuse in new binary
- [x] Make filesystem-based IPFS link resolver for subgraph deployment
- [x] Add file watching code to trigger redeploys when files change
- [x] Add logic to kill/restart node cleanly when files update
- [x] Plug in pg_temp to auto provision temporary databases
- [x] Support for Grafting
- [x] Support for subgraph composition locally
- [ ] Implement build pipeline to build the new binary 
- [ ] Graph CLI to install this binary
- [ ] Build filesystem-based IPFS client (for file data sources)

This will significantly improve the local development experience by removing configuration barriers.

### Some information to help us out

- [x] Tick this box if you plan on implementing this feature yourself.
- [x] I have searched the issue tracker to make sure this issue is not a duplicate.

PR #5981, opened 2025-05-01
Implement a File Link Resolver

incrypto32:
This PR implements a file system based link resolver aimed to replace IpfsLinkResolver for local development, this is in preparation for #5977

### DIFF
--- a/graph/src/components/link_resolver/file.rs
+++ b/graph/src/components/link_resolver/file.rs
@@ -50,8 +50,9 @@ impl FileLinkResolver {
 }
 
 pub fn remove_prefix(link: &str) -> &str {
-    if link.starts_with("/ipfs/") {
-        &link[6..] // Skip the "/ipfs/" prefix (6 characters)
+    const IPFS: &str = "/ipfs/";
+    if link.starts_with(IPFS) {
+        &link[IPFS.len()..]
     } else {
         link
     }
--- a/graph/src/components/link_resolver/file.rs
+++ b/graph/src/components/link_resolver/file.rs
@@ -124,8 +124,6 @@ mod tests {
         let link = Link {
             link: test_file_path.to_string_lossy().to_string(),
         };
-        let resolved_path = resolver.resolve_path(&link.link);
-        println!("Absolute mode - Resolved path: {:?}", resolved_path);
         let result = resolver.cat(&logger, &link).await.unwrap();
         assert_eq!(result, test_content);
 
@@ -133,11 +131,6 @@ mod tests {
         let link = Link {
             link: "/test.txt".to_string(),
         };
-        let resolved_path = resolver.resolve_path(&link.link);
-        println!(
-            "Absolute mode - Path with leading slash: {:?}",
-            resolved_path
-        );
         let result = resolver.cat(&logger, &link).await;
         assert!(
             result.is_err(),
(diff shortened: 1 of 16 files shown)


graph/src/components/link_resolver/file.rs
```diff
@@ -0,0 +1,214 @@
...
+            base_dir: Some(base_dir.as_ref().to_owned()),
+            timeout: Duration::from_secs(30),
+        }
+    }
+
+    fn resolve_path(&self, link: &str) -> PathBuf {
+        let path = Path::new(link);
+
+        // If the path is already absolute or if we don't have a base_dir, return it as is
+        if path.is_absolute() || self.base_dir.is_none() {
```
lutter, line 43: Strictly speaking, you don't need the check for `is_absolute` since `Path.join` will already do the right thing

graph/src/components/link_resolver/file.rs
```diff
@@ -0,0 +1,214 @@
...
+        }
+    }
+
+    async fn get_block(&self, _logger: &Logger, _link: &Link) -> Result<Vec<u8>, Error> {
+        Err(anyhow!("get_block is not implemented for FileLinkResolver").into())
+    }
+
+    async fn json_stream(&self, _logger: &Logger, _link: &Link) -> Result<JsonValueStream, Error> {
+        Err(anyhow!("json_stream is not implemented for FileLinkResolver").into())
+    }
```
lutter, line 96: Making these always return errors is sorta ugly. It would be nicer to have a `trait FileResolver` that doesn't have these two and a `trait LinkResolver` that does and change the code to use the right one. That's going to be a bigger change, so fine to do it in a separate PR, but we shouldn't let this linger for too long.

The danger with leaving this too long is that the code becomes brittle since now every user of `LinkResolver` needs to make sure it gets the right kind as that's not guaranteed by the type alone anymore

graph/src/components/link_resolver/file.rs
```diff
@@ -0,0 +1,214 @@
...
+        let link = Link {
+            link: "missing.txt".to_string(),
+        };
+        let result = resolver.cat(&logger, &link).await;
+        assert!(result.is_err());
+
+        // Clean up
+        let _ = fs::remove_file(test_file_path);
+        let _ = fs::remove_dir(temp_dir);
+    }
```
lutter, line 222: There's a few `println` in here that are probably left over from testing

lutter (approved):

graph/src/components/link_resolver/file.rs
```diff
@@ -49,6 +49,14 @@ impl FileLinkResolver {
     }
 }
 
+pub fn remove_prefix(link: &str) -> &str {
+    if link.starts_with("/ipfs/") {
+        &link[6..] // Skip the "/ipfs/" prefix (6 characters)
```
lutter, line 54: You could just do:
```rust
const IPFS: &str = "/ipfs/";
if link.starts_with(IPFS) {
  &link[IPFS.len()..]
} else {
...
}
```

lutter (approved):

graph/src/components/link_resolver/file.rs
incrypto32: Addressed

graph/src/components/link_resolver/file.rs
incrypto32: Addressed

graph/src/data/subgraph/mod.rs
```diff
@@ -116,21 +116,24 @@ impl DeploymentHash {
     pub fn new(s: impl Into<String>) -> Result<Self, String> {
         let s = s.into();
 
-        // Enforce length limit
-        if s.len() > 46 {
-            return Err(s);
-        }
+        // This section is being temporarily commented out. This is to allow file link resolver to work
+        // TODO(krishna): Figure out how to do this better or remove this check
```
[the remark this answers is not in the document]
incrypto32: Addressed

PR #5982, opened 2025-05-01
Graph node dev mode

incrypto32:
- Uses File based link resolver to run subgraphs from file system
- Listen for file system events to remove and redeploy subgraphs during development
- Uses pg_temp to spin up temporary databases in build dirs of subgraphs


node/src/launcher.rs
```diff
@@ -568,6 +573,23 @@ pub async fn run(opt: Opt, env_vars: Arc<EnvVars>) {
...
+                        &logger,
+                        network_store.subgraph_store(),
+                        subgraph_registrar.clone(),
+                        node_id.clone(),
+                        dev_ctx.updates_rx,
+                    )
+                    .await;
+                });
+            }
+        }
```
lutter, line 592: I find this a little odd to have the optional `dev_ctx` passed into `run`. Another approach would be to have a new function `start` that does most of what `run` does now except for the last line `graph::futures03::future::pending::<()>().await` so that graph-node's `run` function would call `start(..)` and then wait on that pending future whereas `gnd`'s `run` would call `start` and then block on `watch_subgraph_updates`

node/src/bin/dev.rs
```diff
@@ -0,0 +1,169 @@
...
+    let dev_opt = DevOpt::parse();
+
+    let build_dir = get_build_dir(&dev_opt.manifest)?;
+
+    let db = PgTempDBBuilder::new()
+        .with_data_dir_prefix(build_dir.clone())
+        .with_initdb_param("-E", "UTF8")
+        .with_initdb_param("--locale", "C")
+        .start_async()
+        .await;
```
lutter, line 132: It would be a nice enhancement if it was possible to pass a postgres URL in as an option, so that when you already have PG setup you can just use that, and use `pgtemp` if the user doesn't pass a URL in.

lutter (approved):
This all looks great! There are a few improvements that would streamline the code, but this is a really good addition

node/src/dev/watcher.rs
```diff
@@ -0,0 +1,136 @@
...
+            Err(_) => break,
+        };
+
+        if first_event.is_none() {
+            continue;
+        }
+
+        // Once we receive an event, wait for a short period of time to allow for multiple events to be received
+        // This is because running graph build writes multiple files at once
+        // Which triggers multiple events, we only need to react to it once
```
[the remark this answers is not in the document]
incrypto32: Agreed, thats much more robust

node/src/bin/dev.rs
incrypto32: Noted. Working on it

node/src/dev/watcher.rs
```diff
@@ -0,0 +1,136 @@
...
+        }
+    }
+}
+
+/// Determines if an event should be processed based on exclusion patterns
+fn should_process_event(event: &Event, base_dir: &Path, exclusion_set: &GlobSet) -> bool {
+    // If no exclusions, process all events
+    if exclusion_set.is_empty() {
+        return true;
+    }
```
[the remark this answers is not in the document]
incrypto32: A thanks for pointing out, it was leftover from when the logic was different.
Fixed in PR #6000

node/src/launcher.rs
incrypto32: Fixed in PR #6000

node/src/bin/dev.rs
incrypto32: Fixed in PR #6000

PR #5995, opened 2025-05-07
Bump IPFS client max retry delay to 60 seconds

encalypto:
We add a new configurable max delay between retries, which defaults to 60 seconds for requests made via the IPFS client.

### DIFF
--- a/graph/src/ipfs/retry_policy.rs
+++ b/graph/src/ipfs/retry_policy.rs
@@ -1,3 +1,5 @@
+use std::time::Duration;
+
 use slog::Logger;
 
 use crate::ipfs::error::IpfsError;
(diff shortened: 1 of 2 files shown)


lutter (approved):
This effectively bumps the max delay between retries from 1s to 60s. That makes sense, but I don't understand why the description says this is configurable - it's still hardcoded to 60s AFAICT

encalypto:
> This effectively bumps the max delay between retries from 1s to 60s. That makes sense, but I don't understand why the description says this is configurable - it's still hardcoded to 60s AFAICT

Ah this is bad wording; I meant that the `max_delay` value is configurable, not that it's an option for indexers.

PR #5998, opened 2025-05-09
graph: IPFS_MAX_ATTEMPTS ani-smap mechanism env var

rotarur:
This PR changes the hardcoded DEFAULT_MAX_ATTEMPTS to a configurable env var GRAPH_NODE_IPFS_MAX_ATTEMPTS to avoid spamming the IPFS on deterministic errors

### DIFF
--- a/docs/environment-variables.md
+++ b/docs/environment-variables.md
@@ -86,11 +86,11 @@ those.
   may use (in bytes, defaults to 256MB).
 - `GRAPH_MAX_IPFS_CACHE_SIZE`: maximum number of files cached (defaults to 50).
 - `GRAPH_MAX_IPFS_CACHE_FILE_SIZE`: maximum size of each cached file (in bytes, defaults to 1MiB).
-- `GRAPH_IPFS_REQUEST_LIMIT`: Limits the number of requests per second to IPFS for file data sources.
-- `GRAPH_IPFS_MAX_ATTEMPTS`: Safety mechanism to prevent infinite spamming
-  of IPFS servers in the event of logical or unhandled deterministic
-  error (default: 100 000)
-  Defaults to 100.
+- `GRAPH_IPFS_REQUEST_LIMIT`: Limits the number of requests per second to IPFS for file data sources. Defaults to 100.
+- `GRAPH_IPFS_MAX_ATTEMPTS`: This limits the IPFS retry requests in case of a
+  file not found or logical issue working as a safety mechanism to
+  prevent infinite spamming of IPFS servers and network congestion
+  (default: 100 000).
 
 ## GraphQL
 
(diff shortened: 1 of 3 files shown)


fordN:
Is this really just for deterministic errors?  I don't see anything that is classifying the errors as deterministic and only hitting max for those.

sirenko (approved):
LGTM

docs/environment-variables.md
```diff
@@ -87,6 +87,9 @@ those.
 - `GRAPH_MAX_IPFS_CACHE_SIZE`: maximum number of files cached (defaults to 50).
 - `GRAPH_MAX_IPFS_CACHE_FILE_SIZE`: maximum size of each cached file (in bytes, defaults to 1MiB).
 - `GRAPH_IPFS_REQUEST_LIMIT`: Limits the number of requests per second to IPFS for file data sources.
+- `GRAPH_IPFS_MAX_ATTEMPTS`: Safety mechanism to prevent infinite spamming
+  of IPFS servers in the event of logical or unhandled deterministic
+  error (default: 100 000)
```
lutter, line 92: The description should explain what this variable does, not why you would want to use it. Also the next line seems to belong to `GRAPH_IPFS_REQUEST_LIMIT`

graph/src/ipfs/retry_policy.rs
```diff
@@ -3,13 +3,10 @@ use std::time::Duration;
...
 
 use crate::ipfs::error::IpfsError;
+use crate::prelude::*;
 use crate::util::futures::retry;
 use crate::util::futures::RetryConfig;
 
-/// This is a safety mechanism to prevent infinite spamming of IPFS servers
-/// in the event of logical or unhandled deterministic errors.
-const DEFAULT_MAX_ATTEMPTS: usize = 10_0000;
-
```
lutter, line 12: The default for the new env var raises the default from 10k to 100k. Is that intentional?

lutter (changes requested):

graph/src/ipfs/retry_policy.rs
rotarur: i thought that `usize: 10_0000` is 100K as the underscore is ignored. should I use the same `10_000` inside the mappings?

docs/environment-variables.md
rotarur: thanks, will fix that and add a better description

graph/src/ipfs/retry_policy.rs
lutter: Oh wow .. I was fooled by the creative placement of the `_` - yes, `10_0000` is 100k. No need to change anything.

lutter (approved):

PR #6002, opened 2025-05-13
Kill and retry pruning batch operations

lutter:
Uses the same timeout as copying. If a prune operation now takes longer than `GRAPH_STORE_BATCH_TIMEOUT` it is aborted and retried with a batch size of 1


zorancv (approved):
Looks good!

PR #6003, opened 2025-05-13
Split the subgraph_deployment table into two

lutter:
We split the table `subgraphs.subgraph_deployment` into two tables, `subgraphs.head` and `subgraphs.deployment`  where the `head` table only contains the metadata that changes on every block.
    
This should help with situations where the `subgraph_deployment` table gets very bloated since the `head` table that gets bloated through frequent changes has much smaller rows than the current `subgraph_deployment` table. Rows in `subgraph_deployment` can grow as big as 500k, whereas rows in the `head` table will only take about 350 bytes at most.
    
Updates will also be marginally better on the `heads` table since it only has one index rather than the two that `subgraph_deployment` has.

This change has the downstream effect that there is no more `sharded.subgraph_deployment` table. Instead there are now `sharded.head` and `sharded.deployment` tables. This change will also require that all dashboards or other tools that used to access `subgraph_deployment` now access the two new tables.

Besides splitting tables, the new tables also use `int4` columns throughout for block numbers rather than `numeric` since the rest of the system is restricted to `int4` block numbers everywhere else.


store/postgres/src/detail.rs
```diff
@@ -260,15 +310,26 @@ pub(crate) fn deployment_details(
...
-            .filter(d::deployment.eq_any(&deployments))
-            .load::<DeploymentDetail>(conn)?
-    };
+            .inner_join(h::table)
+            .filter(d::subgraph.eq_any(&deployments))
+            .select(cols)
+            .load::<(Deployment, Head)>(conn)?
+    }
+    .into_iter()
+    .map(DeploymentDetail::from)
```
zorancv, line 331: nifty!

zorancv (approved):
Looks good!

PR #6004, opened 2025-05-14
graphql: Add a query test

lutter:
Check the situation where an entity that has empty relations to other entities is followed by one that does have them and ensure that we link up the children correctly in that situation.

Found this change cleaning up old branches, but don't remember what prompted writing it. But another test won't hurt either way

### DIFF
--- a/store/test-store/tests/graphql/query.rs
+++ b/store/test-store/tests/graphql/query.rs
@@ -581,6 +581,7 @@ async fn insert_test_entities(
         vec![
             entity! { is => id: "m3", name: "Tom", mainBand: "b2", bands: vec!["b1", "b2"], favoriteCount: 5, birthDate: timestamp.clone(), vid: 2i64 },
             entity! { is => id: "m4", name: "Valerie", bands: Vec::<String>::new(), favoriteCount: 20, birthDate: timestamp.clone(), vid: 3i64 },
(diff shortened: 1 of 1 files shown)


store/test-store/tests/graphql/query.rs
```diff
@@ -1973,7 +1986,7 @@ fn ambiguous_derived_from_result() {
 }
 
 #[test]
-fn can_filter_by_relationship_fields() {
+fn c_by_relationship_fields() {
```
incrypto32, line 1989: Renamed by mistake?

store/test-store/tests/graphql/query.rs
```diff
@@ -3049,3 +3075,79 @@ fn simple_aggregation() {
         assert_eq!(data, exp);
     })
 }
+
+/// Check that if we have entities where a related entity is null, followed
+/// by one where it is not null that the children are joined correctly to
+/// their respective parent
```
incrypto32, line 3081: Curios, why a null would cause an issue, Was there a bug that recently got fixed?

incrypto32 (approved):
LGTM! apart from the accidental rename of `can_filter_by_relationship_fields`

store/test-store/tests/graphql/query.rs
lutter: Woops .. yes, nice catch

store/test-store/tests/graphql/query.rs
lutter: tbh, I don't remember why I wrote that test, and it passed for me on master on the first try. I just figured it doesn't hurt to have this test either way

PR #6012, opened 2025-05-15
Improve how pruning estimates how much of a table will be removed

lutter:
Pruning can use one of two strategies to get rid of unneeded rows in a table: deleting or rebuilding. Rebuilding is better when we remove more than 50% of a table, otherwise, deletion should be used.

Currently, we estimate how much of a table will be removed by looking at the percentage of historical blocks we prune weighted by the ratio of entity versions to entities. This basically assumes that entity changes are very evenly distributed across blocks. In practice, this leads to serious overestimation of how much of a table will be removed. In one case, the estimate was that 58% of a table would be removed, when in reality only 2% of the table would be removed.

With this change, we use database statistics to improve the estimates which reflect reality much more closely.

### DIFF
--- a/graph/src/components/store/mod.rs
+++ b/graph/src/components/store/mod.rs
@@ -1099,11 +1099,22 @@ impl PruneRequest {
             return None;
         }
 
-        // Estimate how much data we will throw away; we assume that
-        // entity versions are distributed evenly across all blocks so
-        // that `history_pct` will tell us how much of that data pruning
-        // will remove.
-        let removal_ratio = self.history_pct(stats) * (1.0 - stats.ratio);
+        let removal_ratio = if stats.block_range_upper.is_empty()
+            || ENV_VARS.store.prune_disable_range_bound_estimation
+        {
+            // Estimate how much data we will throw away; we assume that
+            // entity versions are distributed evenly across all blocks so
+            // that `history_pct` will tell us how much of that data pruning
+            // will remove.
+            self.history_pct(stats) * (1.0 - stats.ratio)
+        } else {
+            // This estimate is more accurate than the one above since it
+            // does not assume anything about the distribution of entities
+            // and versions but uses the estimates from Postgres statistics.
+            // Of course, we can only use it if we have statistics
+            self.remove_pct_from_bounds(stats)
+        };
+
         if removal_ratio >= self.rebuild_threshold {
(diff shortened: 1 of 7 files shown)


graph/src/components/store/mod.rs
```diff
@@ -928,6 +928,14 @@ pub struct VersionStats {
     pub ratio: f64,
     /// The last block to which this table was pruned
     pub last_pruned_block: Option<BlockNumber>,
+    /// Histograms for the lower and upper bounds of the block ranges in
+    /// this table. Each histogram bucket contains roughly the same number
+    /// of rows; values might be repeated to achieve that. The vectors are
+    /// empty if the table hasn't been analyzed, the subgraph is stored in
+    /// Postgres version 16 or lower, or if the table doesn't have a
+    /// block_range column.
+    pub block_range_lower: Vec<BlockNumber>,
```
zorancv, line 937: Do we need this?

lutter: No, it's not used, I just included it for symmetry; but maybe we just leave it out for now

zorancv: Looking more into it, I would actually like to have the ratio calculated from `remove_pct_from_bounds()` stored here rather than the raw vectors. The `VersionStats` entities are also kept as slice/vector, so the compound structure can potentially grow significantly... The same concern applies to the `block_range_histogram()` and the result that is a vector of vectors, I'd rather have one request per table there.

zorancv (approved):
Voiced a single concern which I can't judge how relevant it is. Other than that the improvement of PR seems to be really good and important.

graph/src/components/store/mod.rs
lutter: I removed the `block_range_lower` field - there's no point in hanging on to it if we don't use it.

As for storing just the `remove_pct_from_bounds`, that gets a little awkward since we'd need to thread the earliest block all the way into `Catalog::stats`. I don't think these arrays will cause a big headache - they will contain as many entries as the stats target for the underlying table is set to, which defaults to 100, and I've never seen anybody set it to more than 1000, so in general those arrays will take ~ 800 bytes but might take up to 8k bytes per table.

zorancv: Few KB is nothing. I would not be concerned even with a few MB (multiplied with a single-double digit number of tables).

PR #6025, opened 2025-05-21
Speed up appending changes to a batch

lutter:
When we finish processing a block, we put all the changes into a `Batch`; that batch in turn is then added to the write queue where it might be combined with already existing batches for earlier blocks. An important aspect of speeding up writes is that block ranges for entities that have changes in several batches in the queue are properly closed before changes are written to the database, making it unnecessary to do that with much slower SQL queries.

The way this was done inside the `Batch` was fairly slow: we kept all changes for entities of a given type in a list, and when we needed to add a new change, we would search the entire list to find the last change for the same entity so we could manipulate the block range of that appropriately.

With this change, we keep a hash map from entity id to most recent change so that looking up those changes doesn't require traversing a list anymore.

The speedup can be seen with the `append_row` example. For example, which is fairly realistic,
```
GRAPH_STORE_WRITE_BATCH_MEMOIZE=false cargo run --release --example append_row -- --ids 4000 -r 9000
```
reports a time of about 110ms to construct the batch, whereas running the same with `GRAPH_STORE_WRITE_BATCH_MEMOIZE=true` takes only about 1.8ms.

### DIFF
--- a/graph/examples/append_row.rs
+++ b/graph/examples/append_row.rs
@@ -1,0 +1,123 @@
+use std::{collections::HashSet, sync::Arc, time::Instant};
+
+use anyhow::anyhow;
+use clap::Parser;
+use graph::{
+    components::store::write::{EntityModification, RowGroupForPerfTest as RowGroup},
+    data::{
+        store::{Id, Value},
+        subgraph::DeploymentHash,
+        value::Word,
+    },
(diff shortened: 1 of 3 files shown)


zorancv (approved):
Very nice!

PR #6031, opened 2025-05-22
Allow caching of IPFS files in file system

lutter:
This change adds a variable `GRAPH_IPFS_CACHE_LOCATION`, which, when set, needs to point at a directory writable by `graph-node`. All files that `graph-node` will fetch from IPFS will be cached in that directory so that subsequent requests to fetch these files, e.g., after a restart, will use those cached files rather than issuing another IPFS request.

Caching files has obvious security implications in that it is possible to modify files in the cache, and thereby change a subgraph. In other words, the cache needs to be in a trusted location with restricted access.

### DIFF
--- a/graph/src/ipfs/cache.rs
+++ b/graph/src/ipfs/cache.rs
@@ -10,13 +10,68 @@ use bytes::Bytes;
 use graph_derive::CheapClone;
 use lru_time_cache::LruCache;
 use object_store::{local::LocalFileSystem, path::Path, ObjectStore};
-use slog::{warn, Logger};
+use redis::{
+    aio::{ConnectionManager, ConnectionManagerConfig},
+    AsyncCommands as _, RedisResult, Value,
+};
+use slog::{debug, info, warn, Logger};
+use tokio::sync::Mutex as AsyncMutex;
 
 use crate::{env::ENV_VARS, prelude::CheapClone};
 
 use super::{
     ContentPath, IpfsClient, IpfsError, IpfsRequest, IpfsResponse, IpfsResult, RetryPolicy,
 };
+
+struct RedisClient {
+    mgr: AsyncMutex<ConnectionManager>,
+}
+
+impl RedisClient {
+    async fn new(logger: &Logger, path: &str) -> RedisResult<Self> {
+        let env = &ENV_VARS.mappings;
+        let client = redis::Client::open(path)?;
+        let cfg = ConnectionManagerConfig::default()
+            .set_connection_timeout(env.ipfs_timeout)
+            .set_response_timeout(env.ipfs_timeout);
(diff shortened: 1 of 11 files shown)


isum (approved):

lutter:
@isum I reworked a good chunk of this, for a couple of reasons:
1. I realized that we have more IPFS accesses than just the ones from `IpfsResolver`; I therefore moved caching lower to `IpfsClient`
2. @rotarur requested that we also support Redis as a cache since that scales better than a filesystem

Sorry for troubling you again, but can you have another look at this PR?

graph/src/ipfs/cache.rs
```diff
@@ -0,0 +1,286 @@
...
+            .await
+            .map(|_: Value| ())
+            .map_err(|e| IpfsError::InvalidCacheConfig {
+                source: anyhow!("Failed to put IPFS object {key} in Redis cache: {e}"),
+            })?;
+        Ok(())
+    }
+
+    fn key(path: &ContentPath) -> String {
+        format!("ipfs:{}", path.cid())
```
isum, line 72: This ensures that we generate the correct key for requests of files within IPFS directories

```suggestion
        format!("ipfs:{path}")
```

graph/src/ipfs/cache.rs
```diff
@@ -0,0 +1,286 @@
...
+    async fn find(&self, logger: &Logger, path: &ContentPath) -> Option<Bytes> {
+        match self {
+            Cache::Memory {
+                cache,
+                max_entry_size: _,
+            } => cache.lock().unwrap().get(path).cloned(),
+            Cache::Disk { store } => {
+                let log_err = |e: &object_store::Error| log_object_store_err(logger, e, false);
+
+                let path = Path::from(path.cid().to_string());
```
isum, line 160: This is necessary to ensure that we retrieve the correct data

```suggestion
                let path = Path::from(path.to_string());
```

graph/src/ipfs/cache.rs
```diff
@@ -0,0 +1,286 @@
...
+            Cache::Memory { cache, .. } => {
+                let mut cache = cache.lock().unwrap();
+
+                if !cache.contains_key(&path) {
+                    cache.insert(path.clone(), data.clone());
+                }
+            }
+            Cache::Disk { store } => {
+                let log_err = |e: &object_store::Error| log_object_store_err(logger, e, true);
+                let path = Path::from(path.cid().to_string());
```
isum, line 188: This is necessary to ensure that we cache the correct data

```suggestion
                let path = Path::from(path.to_string());
```

graph/src/ipfs/cache.rs
lutter: Aah .. yes! Nice catch!

PR #6032, opened 2025-05-26
graph: Log the path we are accessing in IPFS errors

lutter:


isum (approved):

isum:
These changes will be replaced by PR #6058, which adds more context to the error logs

PR #6042, opened 2025-05-29
chain/ethereum: Use the log_index as the transaction_log_index

lutter:
This reverts an inadvertent change that was introduced in PR https://github.com/graphprotocol/graph-node/pull/5867


fordN (approved):
LGTM! 👍

PR #6053, opened 2025-06-05
runtime: Use interior mutability in AscHeapCtx

lutter:


mangas (approved):

runtime/wasm/src/module/mod.rs
```diff
@@ -154,19 +168,15 @@ pub struct AscHeapCtx {
     // is zeroed when initialized or grown.
     memory: Memory,
 
-    // First free byte in the current arena. Set on the first call to `raw_new`.
-    arena_start_ptr: i32,
-
-    // Number of free bytes starting from `arena_start_ptr`.
-    arena_free_size: i32,
+    arena: Mutex<Arena>,
```
incrypto32: I feel like the mutex is not required here, there is no concurrent access to the arena since everything is sequential, and we'd have the locking overhead in such a hot codepath. I can see that without it the compiler would complain about ownership issues. Maybe a RefCell would be better here?

lutter: Nice catch! I just updated it to use a RefCell. The main reason for all this is that we want to mutate the `arena` without passing a `&mut AscHeapCtx` around everywhere

PR #6058, opened 2025-06-12
Add IPFS usage metrics / extend logging / extend supported content path formats

isum:
This PR introduces the following improvements to IPFS:

- Adds 4 new metrics grouped by deployment hashes:
  - `ipfs_request_count` - shows the total number of IPFS requests
  - `ipfs_error_count` - shows the total number of failed IPFS requests
  - `ipfs_not_found_count` - shows the total number of IPFS requests that timed out
  - `ipfs_request_duration` - shows the duration of successful IPFS requests
- Fixes a minor bug that could cause the IPFS gateway client to fail during initialization
- The IPFS content path parser now supports the following formats:
    - `<CID>[/<path>]`
    - `/ipfs/<CID>[/<path>]`
    - `ipfs://<CID>[/<path>]`
    - `http[s]://.../ipfs/<CID>[/<path>]`
    - `http[s]://.../api/v0/cat?arg=<CID>[/<path>]`
    - `http[s]://.../<CID>[/<path>]`
 - IPFS error logs now include more context, such as deployment hash and the requested IPFS path

Closes https://github.com/graphprotocol/graph-node/issues/6036

### DIFF
--- a/chain/substreams/src/data_source.rs
+++ b/chain/substreams/src/data_source.rs
@@ -196,7 +196,7 @@ impl blockchain::UnresolvedDataSource<Chain> for UnresolvedDataSource {
     ) -> Result<DataSource, Error> {
         let content = resolver
             .cat(
-                LinkResolverContext::new(deployment_hash, logger),
+                &LinkResolverContext::new(deployment_hash, logger),
                 &self.source.package.file,
             )
             .await?;
@@ -245,7 +245,7 @@ impl blockchain::UnresolvedDataSource<Chain> for UnresolvedDataSource {
         let handler = match (self.mapping.handler, self.mapping.file) {
             (Some(handler), Some(file)) => {
                 let module_bytes = resolver
-                    .cat(LinkResolverContext::new(deployment_hash, logger), &file)
+                    .cat(&LinkResolverContext::new(deployment_hash, logger), &file)
                     .await
                     .with_context(|| format!("failed to resolve mapping {}", file.link))?;
 
@@ -740,13 +740,13 @@ mod test {
             unimplemented!()
         }
 
-        async fn cat(&self, _ctx: LinkResolverContext, _link: &Link) -> Result<Vec<u8>, Error> {
+        async fn cat(&self, _ctx: &LinkResolverContext, _link: &Link) -> Result<Vec<u8>, Error> {
             Ok(gen_package().encode_to_vec())
         }
 
         async fn get_block(
             &self,
-            _ctx: LinkResolverContext,
+            _ctx: &LinkResolverContext,
             _link: &Link,
         ) -> Result<Vec<u8>, Error> {
             unimplemented!()
@@ -754,7 +754,7 @@ mod test {
 
         async fn json_stream(
             &self,
-            _ctx: LinkResolverContext,
+            _ctx: &LinkResolverContext,
             _link: &Link,
         ) -> Result<JsonValueStream, Error> {
             unimplemented!()
--- a/chain/substreams/src/data_source.rs
+++ b/chain/substreams/src/data_source.rs
@@ -4,7 +4,11 @@ use anyhow::{anyhow, Context, Error};
 use graph::{
     blockchain,
     cheap_clone::CheapClone,
-    components::{link_resolver::LinkResolver, subgraph::InstanceDSTemplateInfo},
+    components::{
+        link_resolver::{LinkResolver, LinkResolverContext},
+        subgraph::InstanceDSTemplateInfo,
+    },
+    data::subgraph::DeploymentHash,
     prelude::{async_trait, BlockNumber, Link},
     slog::Logger,
 };
@@ -184,12 +188,18 @@ pub struct UnresolvedMapping {
 impl blockchain::UnresolvedDataSource<Chain> for UnresolvedDataSource {
     async fn resolve(
         self,
+        deployment_hash: &DeploymentHash,
         resolver: &Arc<dyn LinkResolver>,
         logger: &Logger,
         _manifest_idx: u32,
         _spec_version: &semver::Version,
     ) -> Result<DataSource, Error> {
-        let content = resolver.cat(logger, &self.source.package.file).await?;
+        let content = resolver
+            .cat(
+                LinkResolverContext::new(deployment_hash, logger),
+                &self.source.package.file,
+            )
+            .await?;
 
         let mut package = graph::substreams::Package::decode(content.as_ref())?;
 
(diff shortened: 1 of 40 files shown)


chain/substreams/src/data_source.rs
```diff
@@ -433,7 +444,10 @@ mod test {
         let ds: UnresolvedDataSource = serde_yaml::from_str(TEMPLATE_DATA_SOURCE).unwrap();
         let link_resolver: Arc<dyn LinkResolver> = Arc::new(NoopLinkResolver {});
         let logger = Logger::root(Discard, o!());
-        let ds: DataSource = ds.resolve(&link_resolver, &logger, 0).await.unwrap();
+        let ds: DataSource = ds
+            .resolve(&Default::default(), &link_resolver, &logger, 0)
```
lutter: I kinda dislike using `Default::default` because from just reading the code, I have no idea what gets passed there. Could you write that as `LinkResolverContext::default()` ?

chain/substreams/src/data_source.rs
```diff
@@ -470,7 +484,10 @@ mod test {
             serde_yaml::from_str(TEMPLATE_DATA_SOURCE_WITH_PARAMS).unwrap();
         let link_resolver: Arc<dyn LinkResolver> = Arc::new(NoopLinkResolver {});
         let logger = Logger::root(Discard, o!());
-        let ds: DataSource = ds.resolve(&link_resolver, &logger, 0).await.unwrap();
+        let ds: DataSource = ds
+            .resolve(&Default::default(), &link_resolver, &logger, 0)
```
lutter: Same comment about `Default::default()` here

graph/src/ipfs/content_path.rs
```diff
@@ -214,23 +224,67 @@ mod tests {
...
+        assert_eq!(path, make_path(CID_V0, None));
+
+        let path = ContentPath::new(format!("http://ipfs.com/{CID_V0}/readme.md")).unwrap();
+        assert_eq!(path, make_path(CID_V0, Some("readme.md")));
+
+        let path = ContentPath::new(format!("https://ipfs.com/{CID_V0}")).unwrap();
+        assert_eq!(path, make_path(CID_V0, None));
+
+        let path = ContentPath::new(format!("https://ipfs.com/{CID_V0}/readme.md")).unwrap();
+        assert_eq!(path, make_path(CID_V0, Some("readme.md")));
```
lutter, line 301: Everything in this file looks great, but since we are handling user input here, are there ways to abuse this in security-relevant ways? (I _think_ the answer is 'no', but wanted to double-check)

graph/src/components/link_resolver/mod.rs
```diff
@@ -23,14 +26,43 @@ pub trait LinkResolver: Send + Sync + 'static + Debug {
...
+}
+
+#[cfg(debug_assertions)]
+impl Default for LinkResolverContext {
+    fn default() -> Self {
+        Self {
+            deployment_hash: "test".into(),
+            logger: crate::log::discard(),
+        }
+    }
```
lutter, line 81: This seems petty, but rather than implement `Default`, it might be clearer to make this a method `LinkResolverContext::dummy()` (only in debug builds) It's not really a default implementation, since there's not really a sensible default for this struct.

core/src/polling_monitor/ipfs_service.rs
```diff
@@ -5,13 +5,17 @@ use anyhow::anyhow;
...
-use graph::ipfs::IpfsClient;
-use graph::ipfs::RetryPolicy;
+use graph::ipfs::{ContentPath, IpfsClient, IpfsContext, RetryPolicy};
 use graph::{derive::CheapClone, prelude::CheapClone};
 use tower::{buffer::Buffer, ServiceBuilder, ServiceExt};
 
-pub type IpfsService = Buffer<ContentPath, BoxFuture<'static, Result<Option<Bytes>, Error>>>;
+pub type IpfsService = Buffer<IpfsRequest, BoxFuture<'static, Result<Option<Bytes>, Error>>>;
+
+#[derive(Clone, Debug)]
```
lutter: I got a little hung up on why this needs to be 'Clone'; I think it ultimately boils down to the fact that when we make the request, we want it returned back and we do that in `ReturnRequest::call`; it seems we could avoid cloning if we changed what `IpfsServiceInner::call_inner` returns and have it always include the request in its return value, basically moving ownership of the request through `call_inner`. In this case, I am also not sure how important it is to save on cloning, though we clone on every request.

In any event, this would definitely be something for a different PR.

graph/src/ipfs/client.rs
```diff
@@ -32,21 +32,31 @@ pub trait IpfsClient: Send + Sync + 'static {
     /// The timeout is not propagated to the resulting stream.
     async fn cat_stream(
         self: Arc<Self>,
+        ctx: IpfsContext,
```
lutter: Does this need to take ownership? It seems that just a reference would be enough

graph/src/ipfs/client.rs
```diff
@@ -61,27 +71,33 @@ pub trait IpfsClient: Send + Sync + 'static {
     /// does not return a response within the specified amount of time.
     async fn cat(
         self: Arc<Self>,
+        ctx: IpfsContext,
```
lutter: Same question about ownership

graph/src/ipfs/client.rs
```diff
@@ -95,26 +111,32 @@ pub trait IpfsClient: Send + Sync + 'static {
     /// does not return a response within the specified amount of time.
     async fn get_block(
         self: Arc<Self>,
+        ctx: IpfsContext,
```
lutter: Same question about ownership

lutter:
Do we still want these metrics? If not, can we close this PR?

graph/src/ipfs/content_path.rs
isum: We are not making HTTP requests to arbitrary URLs here, and CIDs are parsed and must be valid. The only part that remains unchanged is the optional path after the CID, but it does not seem to be possible to abuse that it a security-relevant way

graph/src/components/link_resolver/mod.rs
isum: Added a dedicated `test()` method to create a test `LinkResolverContext`

core/src/polling_monitor/ipfs_service.rs
isum: I made the request `CheapClone` by using `Arc` internally in `ContentPath`, and that will help a bit. Optimizing the polling monitor is a good idea and I will look into this as part of different issue / PR.

graph/src/ipfs/client.rs
isum: Updated this to accept `&IpfsContext`

graph/src/ipfs/client.rs
isum: Updated this to accept `&IpfsContext`

graph/src/ipfs/client.rs
isum: Updated this to accept `&IpfsContext`

chain/substreams/src/data_source.rs
isum: Updated to `DeploymentHash::default()`

chain/substreams/src/data_source.rs
isum: Updated to `DeploymentHash::default()`

PR #6063, opened 2025-06-18
Migrate from rust-web3 to alloy

incrypto32:


chain/ethereum/src/runtime/abi.rs
isum: Maybe also rename the `AscH160` type to `AscB160` to have consistent naming.

PR #6069, opened 2025-07-10
emulator: Add support for eth_getBlockReceipts emulation on unsupported chains

SohZHong:
This PR introduces an emulator service that enables `eth_getBlockReceipts` support for EVM-compatible chains that do not natively implement it (e.g. Oasis Sapphire).

Changes:
- Added `emulator/index.js` (Express server that emulates `eth_getBlockReceipts`)
- Added `emulator/Dockerfile` for Docker compatibility
- Modified `docker-compose.yml` to support optional emulator service
- Updated `.gitignore` to exclude `node_modules`
- Updated `README.md` with instructions on enabling the emulator

This is useful for subgraph indexing on networks like Oasis, where the standard receipts RPC method is not available.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,6 @@ lcov.info
 # Docker volumes and debug logs
 .postgres
 logfile
+
+# Node modules
+node_modules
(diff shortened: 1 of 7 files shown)


lutter:
Does this really need to live in `graph-node`? Since it's already a standalone express server, couldn't that just be something that people run themselves?

Also, have you contacted the Oasis Sapphire people? Seeing how simple the  workaround is, why can't they implement that in their client? If a chain claims to be EVM compatible, its clients should provide a complete JSON-RPC API

lutter:
Closing this because I don't think this is needed in `graph-node` Feel free to reopen if there is additional information to consider

PR #6080, opened 2025-07-17
store: ensure consistent id type in FindRangeQuery SQL output

shiyasmohd:
When using subgraph composition and indexing entity of different type `id`, the subgraph wouldn't sync due to the below error
```
UNION types bytea and text cannot be matched
```

This was because the `FindRangeQuery` SQL generator wasn't casting entity IDs to a consistent type, causing `UNION ALL` operations to fail when merging results from subgraphs with different ID types

### DIFF
--- a/store/postgres/src/relational_queries.rs
+++ b/store/postgres/src/relational_queries.rs
@@ -1884,7 +1884,19 @@ impl<'a> QueryFragment<Pg> for FindRangeQuery<'a> {
                 } else {
                     self.mut_range.compare_column(&mut out)
                 }
-                out.push_sql("as block_number, id, vid\n");
+                // Cast id to bytea to ensure consistent types across UNION
+                // The actual id type can be text, bytea, or numeric depending on the entity
+                out.push_sql("as block_number, ");
+                let pk_column = table.primary_key();
+
+                // We only support entity id types of string, bytes, and int8.
+                match pk_column.column_type {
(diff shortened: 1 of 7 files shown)


incrypto32:
@shiyasmohd could you write a unit test to test all cases, also would be great if you could improve the integration tests to also cover this

store/postgres/src/relational_queries.rs
```diff
@@ -1884,7 +1884,19 @@ impl<'a> QueryFragment<Pg> for FindRangeQuery<'a> {
...
+                let pk_column = table.primary_key();
+
+                // We only support entity id types of string, bytes, and int8.
+                match pk_column.column_type {
+                    ColumnType::String => out.push_sql("id::bytea"),
+                    ColumnType::Bytes => out.push_sql("id"),
+                    ColumnType::Int8 => out.push_sql("id::text::bytea"),
+                    _ => out.push_sql("id::bytea"),
+                }
+                out.push_sql(" as id, vid\n");
```
incrypto32, line 1899: @shiyasmohd could you write a unit test to test all cases, also would be great if you could improve the integration tests to also cover this

shiyasmohd: Done.

shiyasmohd:
@incrypto32 Changes:
- Added unit test in `query_tests.rs`
- Covered different entity id for source subgraph case in integration tests

incrypto32 (approved):

PR #6102, opened 2025-08-14
fix cargo check warnings

fubhy:


graph/src/components/subgraph/proof_of_indexing/reference.rs
```diff
@@ -9,17 +9,20 @@ use web3::types::{Address, H256};
 /// well-implemented (without conflicting sequence numbers, or other oddities).
 /// It's just way easier to check that this works, and serves as a kind of
 /// documentation as a side-benefit.
+#[allow(dead_code)]
```
lutter, line 12: This should be moved into the `tests` module in  `graph/src/components/subgraph/proof_of_indexing/mod.rs`.

lutter (approved):
There's a bunch of `allow(dead_code)` that should really be addressed by moving code around. I can do that once this PR is committed.

PR #6113, opened 2025-08-20
all: enhance CLAUDE.md with test verification and release build requirements

fubhy:
## Summary
- Add explicit test verification requirements to prevent false test successes from cargo test commands that match 0 tests
- Add mandatory `cargo build --release` step to catch linking/optimization issues that `cargo check` might miss

## Test plan
- [x] Review changes to CLAUDE.md documentation
- [x] Verify all test sections now include verification requirements
- [x] Confirm release build is added to mandatory pre-commit checks

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -45,6 +45,12 @@ cargo test -p graph data_source::common::tests
 cargo test <specific_test_name>
 ```
 
+**⚠️ Test Verification Requirements:**
+- **ALWAYS verify tests actually ran** - Check the output for "test result: ok. X passed" where X > 0
+- **If output shows "0 passed" or "0 tests run"**, the test filter/path was wrong - fix and re-run
+- **Never trust exit code 0 alone** - Cargo can exit successfully even when no tests matched your filter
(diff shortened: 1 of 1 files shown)


CLAUDE.md
```diff
@@ -103,11 +119,15 @@ cargo fmt --all
 
 # 🚨 MANDATORY: Check code for warnings and errors - MUST have zero warnings
 cargo check
+
+# 🚨 MANDATORY: Build in release mode to catch linking/optimization issues that cargo check misses
+cargo build --release
```
lutter: Release builds are _really_ slow. Wouldn't `cargo check --release` be enough here? That can sometimes fail because of conditional compilation, but I've never seen a linking/optimization error.

lutter (changes requested):

CLAUDE.md
fubhy: I have no idea. I just noticed that I was getting build errors in CI sometimes despite cargo check passing.

fubhy: Will ask claude to adjust with your feedback later

lutter: I just changed it to `cargo check --release` for now so I can merge this

PR #6124, opened 2025-08-21
ci: optimize ci setup

fubhy:
- [ ] Switch to `nextest` as test runner
- [ ] Configure mutex via `nextest` filtersets for tests that do not support concurrency (some unit tests and most integration/runner tests)
- [ ] Use `justfile` to create more accessible aliases for common test & development commands. Use them in CI.
- [ ] Explore how we can refactor integration & runner tests to actually treat test cases as individual tests instead of implementing a custom test runner internally
- [ ] Create and use a pre-built container image for the CI runner that comes with preinstalled system dependencies (protoc & just) to save on install time (~2 minutes per CI job)
- [ ] Use beefier multicore CI runners instead of the github default runners


flake.nix
```diff
@@ -85,7 +87,7 @@
                 shared_preload_libraries = "pg_stat_statements";
                 log_statement = "all";
                 default_text_search_config = "pg_catalog.english";
-                max_connections = 200;
+                max_connections = 500;
```
lutter, line 90: That's a lot of connections. Does the test setup really need that?

lutter (approved):

.github/workflows/ci.yml
```diff
@@ -25,105 +31,118 @@ jobs:
...
-      - name: Tune GitHub hosted runner to reduce flakiness
-        # https://github.com/smorimoto/tune-github-hosted-runner-network/blob/main/action.yml
-        run: sudo ethtool -K eth0 tx off rx off
-      - name: Checkout sources
-        uses: actions/checkout@v4
-      # Don't use the rust-cache as it leads to 'no space left on device' errors
-      # - uses: Swatinem/rust-cache@v2
-      - name: Install lld
-        run: sudo apt-get install -y lld protobuf-compiler
+      - uses: actions/checkout@08eba0b27e820071cde6df949e0beb9ba4906955 # v4
```
lutter: Out of curiosity, is updating to checkout@v5 causing issues?

fubhy: No, I was just too lazy to update the checksums for all projects and instead just copied these from another internal project.

.github/workflows/ci.yml
```diff
@@ -25,105 +31,118 @@ jobs:
...
+      - name: Alter postgres max connections
+        run: |
+          docker exec -i postgres bash << EOF
+            sed -i -e 's/max_connections = 100/max_connections = 1000/' /var/lib/postgresql/data/postgresql.conf
+            sed -i -e 's/shared_buffers = 128MB/shared_buffers = 2GB/' /var/lib/postgresql/data/postgresql.conf
```
lutter: It should be possible to set these via `POSTGRES_INITDB_ARGS` by including something like `-c max_connections=1000 -c shared_buffers=2GB`

flake.nix
fubhy: It was a bit flaky with 200, yes

.github/workflows/ci.yml
fubhy: That didn't work for me. Let me try though.

lutter: Actually, according to [this](https://hub.docker.com/r/bitnami/postgresql) the correct variable is `POSTGRESQL_INITDB_ARGS`

lutter: Wait, we use `bitnami/postgres` for the runner tests, and just `postgres` for the integration tests. Ugh

lutter: For the `postgres` image, [this](https://hub.docker.com/_/postgres#environment-variables) claims that the variable is called `POSTGRES_INITDB_ARGS`

PR #6126, opened 2025-08-23
feat: implement weighted RPC load balancing with traffic distribution 

DaMandal0rian:
## Summary

Adds weighted RPC load balancing with dynamic health-aware traffic distribution across Ethereum RPC providers.

## Key changes

- **Weighted adapter selection**: Probabilistic selection via `WeightedIndex` using configurable provider weights (0.0–1.0)
- **Real health checks**: Periodic `eth_blockNumber` calls (5s timeout, 10s interval) replace the previous stub; health scores adjust weights dynamically
- **Atomics over RwLock**: Health metrics use `AtomicU64`/`AtomicU32` to avoid lock poisoning
- **Per-chain health checkers**: Each chain gets its own set of health checkers instead of a shared global list
- **Cancellation support**: Health check task accepts a `CancellationToken` for graceful shutdown
- **O(1) health lookup**: `HashMap<String, Arc<Health>>` replaces linear scan
- **Encapsulated adapter field**: `EthereumNetworkAdapter.adapter` is now private with a getter

## Configuration

```toml
weighted_rpc_steering = true

[chains.mainnet]
provider = [
  { label = "primary", url = "http://rpc1.io", weight = 0.7 },
  { label = "backup",  url = "http://rpc2.io", weight = 0.3 },
]
```

- Weights must be between 0.0 and 1.0; they don't need to sum to 1.0
- Weight of 0.0 disables a provider from weighted selection while keeping it available for error-retesting
- Weight is only meaningful for RPC providers (ignored for firehose)

Closes [OPS-727](https://linear.app/thegraph/issue/OPS-727/add-weighted-load-balancing-and-fallback-to-graph-node-for-rpc)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3805,6 +3805,7 @@ dependencies = [
  "tiny-keccak 1.5.0",
  "tokio",
  "tokio-stream",
+ "tokio-util",
  "tonic-prost-build",
  "tower 0.5.2",
 ]
--- a/chain/ethereum/Cargo.toml
+++ b/chain/ethereum/Cargo.toml
@@ -18,6 +18,7 @@ semver = { workspace = true }
 thiserror = { workspace = true }
 tokio = { workspace = true }
 tokio-stream = { workspace = true }
+tokio-util = { workspace = true }
 tower = { workspace = true }
 
 itertools = "0.14.0"
--- a/chain/ethereum/src/ethereum_adapter.rs
+++ b/chain/ethereum/src/ethereum_adapter.rs
@@ -57,7 +57,7 @@ use std::convert::TryFrom;
 use std::iter::FromIterator;
 use std::pin::Pin;
 use std::sync::Arc;
-use std::time::Instant;
+use std::time::{Duration, Instant};
 use tokio::sync::RwLock;
 use tokio::time::timeout;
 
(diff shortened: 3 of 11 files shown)


DaMandal0rian:
@lutter I have done the rebase for this branch. PR is ready.

lutter:
Still my question from the other PR: it seems that you can now both configure weights for an RPC provider, but weights are also adjusted dynamically. Why are weights configurable if we can figure them out on the fly? What's the interplay between static and dynamic weights? Which one takes precedence and why?

DaMandal0rian:
@lutter 

The two aren't competing — they compose multiplicatively: `effective_weight = static_weight × health_score`.

Static weights set the operator's intended baseline distribution (e.g., 80/20 split between a premium and budget provider based on cost, capacity, or latency characteristics). The dynamic health score (0.0–1.0, derived from latency + error rate + consecutive failures) then modulates that baseline in real-time.

When all providers are healthy (score=1.0), traffic follows the configured ratios exactly. When a provider degrades, its effective weight drops proportionally — so an 0.8-weight provider with a 0.25 health score gets the same effective weight (0.2) as a healthy 0.2-weight provider.

Without static weights, all healthy providers would get equal traffic regardless of their capacity or cost. Without dynamic scores, a failing provider would keep getting its full share of traffic. You need both to express "prefer this provider, but back off when it's struggling."

PR #6142, opened 2025-09-03
support querying multiple subgraphs' features in single query

shiyasmohd:


server/index-node/src/schema.graphql
```diff
@@ -36,7 +36,7 @@ type Query {
   publicProofsOfIndexing(
     requests: [PublicProofOfIndexingRequest!]!
   ): [PublicProofOfIndexingResult!]!
-  subgraphFeatures(subgraphId: String!): SubgraphFeatures!
+  subgraphFeatures(subgraphs: [String!]!): [SubgraphFeatures!]!
```
lutter, line 39: This is a breaking change - existing queries for subgraph features will break after this change. I am not sure how big a deal that is and how many people are using this, but it could be a problem

PR #6143, opened 2025-09-03
handle empty array in indexing status query

shiyasmohd:


lutter (approved):

snigdha920:
was using this to get a list of subgraph hashes deployed to my graph node, how can i do this now?

shiyasmohd:
you can get it using `graphman info -s --all`

snigdha920:
> you can get it using `graphman info -s --all`

thanks, is that the same as this graphql query to graphman?

```graphql
query {
    deployment {
        __typename
        info {
            hash
            name
            isActive
            versionStatus
            nodeId
        }
    }
}
```

shiyasmohd:
Oh, Yes, You can use the graphaman graphql api too. The query is correct.

PR #6160, opened 2025-09-12
node: pause subgraphs before reassigning in graphman

shiyasmohd:
When reassigning subgraph and not pausing, it gives the below error and creates a race condition.
```
Failed to transact block operations: subgraph `Qm.....` has already processed block `xxx`; there are most likely two (or more) nodes indexing this subgraph
```
This PR solves the issue by pausing subgraphs  and waits before reassigning subgraphs.

### DIFF
--- a/node/src/manager/commands/deployment/reassign.rs
+++ b/node/src/manager/commands/deployment/reassign.rs
@@ -1,4 +1,5 @@
 use std::sync::Arc;
+use std::time::Duration;
 
 use anyhow::Result;
 use graph::prelude::NodeId;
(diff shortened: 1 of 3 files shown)


node/src/manager/commands/deployment/reassign.rs
```diff
@@ -12,10 +13,22 @@ use graphman::deployment::DeploymentSelector;
...
+    super::pause::run(
+        primary_pool.clone(),
+        notification_sender.clone(),
+        deployment_selector.clone(),
+    )?;
+
+    println!(
+        "Waiting {}s to make sure pausing was processed ...",
+        delay.as_secs()
+    );
```
incrypto32, line 29: The delay is not actually used, i think what you meant to do is to call sleep() here too with the delay

node/src/manager/commands/deployment/reassign.rs
```diff
@@ -28,8 +41,8 @@ pub fn run(
     println!("{}", reassign_msg);
 
     let reassign_result = reassign_deployment(
-        primary_pool,
-        notification_sender,
+        primary_pool.clone(),
+        notification_sender.clone(),
         &deployment,
         node,
         curr_node,
```
incrypto32, line 48: minor nitpick: if the `pause::run`  succeeds and actual reassign fails the resume will never run and the subgraph would be paused unintentionally.

incrypto32 (changes requested):

PR #6162, opened 2025-09-14
Fix thread contention during subgraph syncing

DaMandal0rian:
When multiple subgraphs are syncing concurrently, the node can become unresponsive due to thread pool contention. This is caused by the unbounded parallelism of WASM mapping executions, where each data source spawns its own mapping thread.

This PR introduces a semaphore to limit the number of concurrent mapping executions across all subgraphs. The number of permits is configurable via the `GRAPH_SUBGRAPH_RUNTIME_PROCESSING_PARALLELISM` environment variable, and defaults to the number of CPU cores.

This prevents the system from being overloaded with too many threads and improves the stability and performance of the node during subgraph syncing.


lutter:
I've stared at this for a bit and am still not quite sure what this is trying to achieve. There's no semaphore in our code that I can find that is being sharded - the PR just introduces a sharded semaphore. 

I also don't understand the rationale in the intro - how is it getting from 32 workers to 1024 workers? In general, parallelism is constrained by the number of CPUs; tokio takes care of scheduling the much larger amount of work that is usually ready across those. This PR seems to introduce another layer of scheduling on top of tokio's. I am very confused by all this.

DaMandal0rian:
@lutter 

> I've stared at this for a bit and am still not quite sure what this is trying to achieve. There's no semaphore in our code that I can find that is being sharded - the PR just introduces a sharded semaphore.
> 
> I also don't understand the rationale in the intro - how is it getting from 32 workers to 1024 workers? In general, parallelism is constrained by the number of CPUs; tokio takes care of scheduling the much larger amount of work that is usually ready across those. This PR seems to introduce another layer of scheduling on top of tokio's. I am very confused by all this.

Thanks for reviewing! I realize I need to better explain the problem and solution.

  The Current Bottleneck

  The bottleneck isn't in the code you see in master branch - it's in my earlier commits on this branch. In the first 2 commits, I introduced a single global semaphore to limit concurrent trigger
  processing (this was my first solution attempted):

  With 2500 continuously syncing subgraphs all competing for 32 permits from a single semaphore, it would create a regressive and massive contention - 98% of subgraphs were blocked waiting.

  The Solution: Sharding

  This PR replaces that single bottleneck semaphore with multiple sharded semaphores:

  Each subgraph is consistently assigned to one shard, so instead of 2500 subgraphs competing for 32 permits or tokio threads, we have ~78 subgraphs per shard competing for 32 permits.

  About CPU Constraints

  You're  right that we're ultimately constrained by CPUs. The semaphores aren't about creating more parallelism than we have cores - they're about:

  1. Preventing thread and possibly memory exhaustion: Without limits, 2500 subgraphs could queue up unbounded work
  2. Fair scheduling: Ensuring each subgraph gets processing time
  3. Backpressure: Slowing down subgraphs that are queuing too much work

Tokio still schedules these across available CPU cores.

---
About "Another Layer of Scheduling"

  My earlier commit (0c820f563): Added a single global semaphore for trigger processing:

  This PR: Shards that single bottleneck into multiple semaphores:
  // Distributed - ~78 subgraphs compete per shard instead of 2500
  let shard_id = get_shard_for(subgraph);
  let _permit = self.semaphores[shard_id].acquire().await;

  The semaphores do admission control, not scheduling:

  - Tokio's job: Schedule ready tasks across CPU cores
  - Semaphore's job: Limit how many tasks can be in-flight (prevent contention and memory exhaustion)

  Without admission control, 2500 continuously syncing subgraphs would spawn unbounded Futures.

This PR doesn't introduce a new scheduling layer. It uses semaphores + sharded connection pools to reduce contention. I had earlier tried using a global semaphore approach but realised it wouldn't scale. It's more of a WIP and idea on how to better scale the workers/threads and make graph-node manage tokio threads better as the amount of subgraphs per machine grows. I have also updated the description to avoid the confusion with semaphores.

lutter:
I am closing this for now; I don't think this is needed and with the async store changes we are already reducing the number of blocking threads we need, with some more reductions to come.

PR #6169, opened 2025-09-16
Log more details about how assignments are processed

lutter:


fordN:
I'm curious about how the cargo.lock change snuck in there.

lutter:
Woops .. yes, me too.

PR #6172, opened 2025-09-19
SQL query interface (very experimental)

lutter:
This PR picks up the work done in PR https://github.com/graphprotocol/graph-node/pull/5382. It's main purpose is to help evaluate how such an interface would work. There's no expectation that this will ever make it into a graph-node release at this point.

Documentation can be found in docs/implementation/sql-interface.md in this PR

This is a continuation of PR #5799 but had to be a new PR because of commit signing


graphql/src/runner.rs
```diff
@@ -251,4 +253,51 @@ where
...
+
+        let store = self
+            .store
+            .query_store(QueryTarget::Deployment(
+                req.deployment.clone(),
+                ApiVersion::default(),
+            ))
+            .await?;
+
+        let query_hash = req.query_hash();
```
isum, line 277: This hash is not very useful in this context, since the SQL query can be manipulated in a lot of ways to produce a different hash every time.

It is definetly not a problem for testing, but we should normalize the SQL queries before hashing them if we plan to use this feature in production

store/postgres/src/sql/validation.rs
```diff
@@ -0,0 +1,283 @@
...
+            }
+        }
+        ControlFlow::Continue(())
+    }
+}
+
+impl VisitorMut for Validator<'_> {
+    type Break = Error;
+
+    fn pre_visit_statement(&mut self, _statement: &mut Statement) -> ControlFlow<Self::Break> {
```
isum: The parameter name does not need to start with `_`

Cargo.toml
```diff
@@ -81,11 +81,8 @@ serde_derive = "1.0.125"
 serde_json = { version = "1.0", features = ["arbitrary_precision"] }
 serde_regex = "1.1.0"
 serde_yaml = "0.9.21"
-slog = { version = "2.7.0", features = [
-    "release_max_level_trace",
-    "max_level_trace",
-] }
-sqlparser = "0.46.0"
+slog = { version = "2.7.0", features = ["release_max_level_trace", "max_level_trace"] }
+sqlparser = { version = "0.46.0", features = ["visitor"] }
```
isum: The latest version _(0.59.0)_ has an additional variant for `SetExpr` for DELETE statements. It's probably worth updating and covering the new variant in the validator.

store/postgres/src/sql/validation.rs
```diff
@@ -0,0 +1,283 @@
...
+
+    fn pre_visit_query(&mut self, query: &mut Query) -> ControlFlow<Self::Break> {
+        // Add common table expressions to the set of known tables
+        if let Some(ref with) = query.with {
+            self.ctes.extend(
+                with.cte_tables
+                    .iter()
+                    .map(|cte| cte.alias.name.value.to_lowercase()),
+            );
+        }
```
isum, line 207: Nested queries might have their own scoped CTEs what are not available to parent queries. 
CTEs are never removed from the list when a query scope ends and this is a problem, because nested CTEs allow parent queries to access arbitrary tables.

This query passes the validation:
`SELECT *, (WITH pg_user AS (SELECT 1) SELECT 1) AS one FROM pg_user;`

Also, maybe the `pre_` and `post_` access usage should be revisited to ensure consistent validations.

isum (changes requested):
This looks very nice! Blocking because I have found one security issue that needs to be fixed

graphql/src/runner.rs
lutter: Absolutely true; we'd need to hash the AST in some form. But for now, I think we can leave that; it'll help a bit with programmatically generated SQL, but could definitely be improved.

lutter:
Addressed all review comments

isum (approved):

Issue #4503, opened 2023-03-28
[Feature] Implement a way to clean up the call cache


lutter:
### Description

Currently, we only ever write to the call cache, but never remove unneeded entries from it. The call cache can grow quite large, and removing unneeded entries from that will help keep its size in check.

We already have a table `call_meta` that tracks how long ago a cache entry was last accessed. We should add an internal job that periodically removes entries that haven't been used in a while.

### Some information to help us out

- [ ] Tick this box if you plan on implementing this feature yourself.
- [X] I have searched the issue tracker to make sure this issue is not a duplicate.

PR #6186, opened 2025-10-02
Feat/graphman/clear stale call cache

dimitrovmaksim:
Partialy addresses https://github.com/graphprotocol/graph-node/issues/4503

What this PR does:
1. Adds a new option `--ttl-days <days>` to the `graphman chain call-cache <chain> remove` command
Executing the command using the `--ttl-days` option, will remove call cache entries for contracts that haven't been accessed within a user-defined number of days
2. Updates the documentation to reflect the new options
3. Adds a test to verify query correctness.

### DIFF
--- a/node/src/bin/manager.rs
+++ b/node/src/bin/manager.rs
@@ -555,14 +555,18 @@ pub enum ChainCommand {
 pub enum CallCacheCommand {
     /// Remove the call cache of the specified chain.
     ///
-    /// Either remove entries in the range `--from` and `--to`, or remove
-    /// the entire cache with `--remove-entire-cache`. Removing the entire
+    /// Either remove entries in the range `--from` and `--to`,
+    /// remove the cache for contracts that have not been accessed for the specified duration --ttl_days,
+    /// or remove the entire cache with `--remove-entire-cache`. Removing the entire
     /// cache can reduce indexing performance significantly and should
     /// generally be avoided.
     Remove {
         /// Remove the entire cache
         #[clap(long, conflicts_with_all = &["from", "to"])]
         remove_entire_cache: bool,
+        /// Remove the cache for contracts that have not been accessed in the last <TTL_DAYS> days
+        #[clap(long, conflicts_with_all = &["from", "to", "remove-entire-cache"])]
+        ttl_days: Option<i32>,
         /// Starting block number
         #[clap(long, short, conflicts_with = "remove-entire-cache", requires = "to")]
         from: Option<i32>,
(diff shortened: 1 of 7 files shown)


store/postgres/src/chain_store.rs
```diff
@@ -1398,6 +1399,126 @@ mod data {
...
+                        .get_results::<Vec<u8>>(conn)?;
+
+                    if stale_contracts.is_empty() {
+                        return Ok(());
+                    }
+
+                    loop {
+                        let next_batch = cache::table
+                            .select(cache::id)
+                            .filter(cache::contract_address.eq_any(&stale_contracts))
```
incrypto32: Would it be better to batch by contracts rather than cache entries? Im wondering because the number of stale_contracts itself can be very huge, in our production i just checked and the number for >7days is 9624844.
So for each row postgres would need to do a linear scan through this number of address to filter.

node/src/bin/manager.rs
```diff
@@ -555,14 +555,18 @@ pub enum ChainCommand {
...
+    /// or remove the entire cache with `--remove-entire-cache`. Removing the entire
     /// cache can reduce indexing performance significantly and should
     /// generally be avoided.
     Remove {
         /// Remove the entire cache
         #[clap(long, conflicts_with_all = &["from", "to"])]
         remove_entire_cache: bool,
+        /// Remove the cache for contracts that have not been accessed in the last <TTL_DAYS> days
+        #[clap(long, conflicts_with_all = &["from", "to", "remove-entire-cache"])]
+        ttl_days: Option<i32>,
```
incrypto32, line 569: We shouldn't allow negative values here

incrypto32:
Another nice feature would be if we can give a small progress bar for the user indicating the deletion progress if its going to take time

store/postgres/src/chain_store.rs
dimitrovmaksim: If this is this case, probably both have to be batched

dimitrovmaksim:
> Another nice feature would be if we can give a small progress bar for the user indicating the deletion progress if its going to take time

We could print info about each batch. Not sure about progress bars, because that will require fetching the count of all records to be deleted. Also I'm not sure how this will play out later, when the call cache cleaning is implemented as a recurring job, because this file is used by both `graph-node` and `graphman` and the `graph-node` logs via logger, while `graphman` prints with `println!`

store/postgres/src/chain_store.rs
dimitrovmaksim: Btw is that number in the public `eth_call_meta` table?

incrypto32: no its just for ethereum in the private schema

dimitrovmaksim: I'll look into adjusting the queries to make them more efficient, but In general I doubt this approach is preferable, even if optimized for such a number of records. Depending on the avg number of call cache records per contract, it could take days to clean up this many.

incrypto32:
LGTM, feel free to merge

Issue #3499, opened 2022-04-26
chain.call_cache will cache the data whose return_value='' is empty


yongjun925:
**Do you want to request a *feature* or report a *bug*?**
bug

**What is the current behavior?**
chain.call_cache will cache the data whose return_value='' is empty, causing the subsequent returned data to be wrong

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem.**
select count(*) from chain1.call_cache where return_value='';
{
  "errors": [
    {
      "message": "indexing_error"
    }
  ]
}

**What is the expected behavior?**
Don't cache empty data

`chain/ethereum/src/ethereum_adapter.rs`
`if !for_cache.is_empty() {`

madumas:
Re-upping this issue, as I just stumbled into it.

Following an outage of an RPC node, graph-node was restarted while the RPC node was still recovering:

```
2024-11-22 07:46:57.644	Nov 22 12:46:57.644 DEBG Trying again after eth_call RPC call for block #64589850 (c8fb260eaf4269ea0a17bd687ef52ca1f876c8e239fb618965a57beb3f50bba6) failed (attempt #4) with result Err(Web3Error(Rpc(Error { code: InternalError, message: "hash c8fb260eaf4269ea0a17bd687ef52ca1f876c8e239fb618965a57beb3f50bba6 is not currently canonical", data: None }))), provider: dshackle-polygon-ing, sgd: 1854, subgraph_id: QmQEYSGSD8t7jTw4gS2dwC4DLvyZceR9fYQ432Ff1hZpCp, component: SubgraphInstanceManager
2024-11-22 07:47:06.335	Nov 22 12:47:06.335 DEBG Contract call finished, source: rpc, block_number: 64589850, block_hash: c8fb260eaf4269ea0a17bd687ef52ca1f876c8e239fb618965a57beb3f50bba6, result: none, time_ms: 9348ms, args: [], signature: totalFeeGrowth0Token():(uint256), contract: Pool, address: 0xac4494e30a85369e332bdb5230d6d694d4259dbc, sgd: 1854, subgraph_id: QmQEYSGSD8t7jTw4gS2dwC4DLvyZceR9fYQ432Ff1hZpCp, component: SubgraphInstanceManager
[...]

PR #6187, opened 2025-10-02
store: Do not cache calls that return 0x

dimitrovmaksim:
Related to https://github.com/graphprotocol/graph-node/issues/3499


incrypto32 (approved):
LGTM

PR #6194, opened 2025-10-11
Make all store/db interactions async

lutter:
We have a lot of places where we do blocking database work from a non-blocking tokio task. This PR addresses this longstanding issue by using async database connections and `diesel_async`. With that, we don't need to rely on error-prone approaches that require spawning blocking tasks whenever we do database work.

Most of this PR is concerned with sprinkling async/await throughout the code base. There are _a lot_ of commits, but most of them are boring, and I tried to structure it so that many of them just add async/await in a few places to make them easy to review individually.

Since switching to async connections also requires changing the connection pool, towards the end of the PR there are changes to recreate the instrumentation we had with r2d2 with deadpool.

There are still places in the codebase where `spawn_blocking` is used, but those can be removed in separate PR's as this one is already crazy long. My apologies to the reviewer(s)

I made sure that `cargo check` passed after each commit; in addition, I ran CI after every few (say 5-10) commits and made sure that passed, too, up to the commit marked `SQUASH FROM HERE` as it was not possible to pass CI from that point on until `CI PASSES AGAIN` as the `AsyncConnectionWrapper` uses `block_on` for db interactions, and too many of them happen on the main tokio threads. For those, tokio panics as you are not allowed to use `block_on` on those threads.

Fixes https://github.com/graphprotocol/graph-node/issues/905


isum (approved):
This is a huge improvement! 👏

isum (approved):

lutter:
Squashed all commits into one to make rollback easier if that's needed

lutter:
Just for posterity's sake: I ran this in the integration cluster, and there were no PoI differences. I also checked performance of a single subgraph, and there are no perf regressions for that, either

PR #6199, opened 2025-10-16
node: Avoid duplicate key constraints when changing chain shard

### LINKED ISSUE #6196: [Bug] Cannot Revert Chain Shard Change Due to Duplicate Key Constraint
### Description

When changing the shard for a chain using `graphman chain change-shard`, the existing chain is correctly renamed to `<chain>-old` and persists in the database.

However, if a subsequent attempt is made to revert the chain's shard back to the original shard (which would cause the currently active chain to be renamed to `<chain>-old` again), the operation fails due to a **unique constraint violation** on the `chains` table. The error indicates that the key value (`<chain>-old`) already exists, as it was created during the initial shard change.

### Steps to Reproduce

- Start with chain A on shard S1 (e.g., matic on polygon_chain).
- Change the shard for chain A to S2 (e.g., primary):
```
graphman chain change-shard A S2
# Result: Successful. A is on S2. A-old (the original chain) is on S1.
```

- Attempt to change the shard for chain A back to the original shard S1:
```
graphman chain change-shard A S1
```

### Expected Behavior

The shard change should complete successfully. The current chain A (on S2) should be renamed to A-old (or a more uniquely named backup), and the old chain A-old (on S1) should be re-promoted/renamed to A.

### Actual Behavior

The command fails with a database constraint violation:
```
Error: database constraint violated: duplicate key value violates unique constraint "chains_name_key": Key (name)=(matic-old) already exists.
```

### Additional

The current logic for `change-shard` unconditionally attempts to rename the current chain to `<chain>-old`. This approach is fragile because it doesn't account for the persistence of a previously archived `<chain>-old` entity. The failure occurs because the database enforces a unique constraint on the `name` column.

### Relevant log output

```Shell
root@b5c46359fc62:/# graphman chain change-shard matic primary
Changing block cache shard for matic to primary
Current shard: polygon_chain
Error: database constraint violated: duplicate key value violates unique constraint "chains_name_key": Key (name)=(matic-old) already exists.
```

### IPFS hash

_No response_

_No response_

### OS information

Linux

erayack:
## Description

This PR fixes issue #6196 by implementing a robust backup naming system that prevents duplicate key constraint violations when reverting chain shard changes.

## Problem

When changing a chain's shard using `graphman chain change-shard`, the existing chain is renamed to `<chain>-old`. However, if a subsequent attempt is made to revert the chain's shard back to the original shard, the operation fails due to a unique constraint violation because `<chain>-old` already exists in the database.

## Solution

- **Robust backup naming**: Implemented `next_backup_name()` function that generates unique backup names by appending numeric suffixes when conflicts exist
- **Backup reuse optimization**: When reverting to the original shard, the system now reuses the existing backup instead of creating a new one
- **Backup preservation**: Previous backups are preserved with unique names to maintain data integrity
- **Logging**: Added logging to track backup operations and outcomes

## Changes

- Added `ChainSwapOutcome` struct to track the results of chain swap operations
- Implemented `next_backup_name()` function for conflict-free backup naming
- Enhanced `change_block_cache_shard()` function with robust backup handling logic
- Added detailed logging for backup operations

## Testing

The fix handles the following scenarios:
1. Initial shard change (creates `<chain>-old`)
2. Reverting to original shard (reuses existing backup)
3. Multiple shard changes (preserves all backups with unique names)

Fixes #6196

### DIFF
--- a/node/src/manager/commands/chain.rs
+++ b/node/src/manager/commands/chain.rs
@@ -31,6 +31,7 @@ use graph_store_postgres::find_chain;
 use graph_store_postgres::update_chain_name;
 use graph_store_postgres::{ConnectionPool, command_support::catalog::block_store};
 
+use crate::manager::prompt::prompt_for_confirmation;
 use crate::network_setup::Networks;
 
 pub async fn list(primary: ConnectionPool, store: BlockStore) -> Result<(), Error> {
@@ -249,6 +250,10 @@ pub async fn change_block_cache_shard(
         .await?
         .ok_or_else(|| anyhow!("unknown chain: {}", chain_name))?;
     let old_shard = chain.shard;
+    let canonical_backup_name = format!("{chain_name}-old");
+
+    let existing_backup = find_chain(&mut conn, &canonical_backup_name).await?;
+    let existing_backup_store = store.chain_store(&canonical_backup_name).await;
 
     println!("Current shard: {}", old_shard);
 
@@ -256,30 +261,92 @@ pub async fn change_block_cache_shard(
         .chain_store(&chain_name)
         .await
         .ok_or_else(|| anyhow!("unknown chain: {}", &chain_name))?;
-    let new_name = format!("{}-old", &chain_name);
     let ident = chain_store.chain_identifier().await?;
+    let target_shard = Shard::new(shard)?;
+
+    let reuse_existing_backup = match existing_backup.as_ref() {
+        None => false,
+        Some(backup) if backup.shard != target_shard => {
+            bail!(
+                "`{}` already exists on shard `{}`. Remove it with `graphman chain remove {}` before changing `{}` to shard `{}`",
+                canonical_backup_name,
+                backup.shard,
+                canonical_backup_name,
+                chain_name,
+                target_shard,
+            );
+        }
+        Some(backup) => {
+            let backup_ident = backup.network_identifier()?;
+            if backup_ident != ident {
+                bail!(
+                    "`{}` has a different chain identifier ({}) than `{}` ({}). Remove it with `graphman chain remove {}` before changing `{}` to shard `{}`",
+                    canonical_backup_name,
+                    backup_ident,
+                    chain_name,
+                    ident,
+                    canonical_backup_name,
+                    chain_name,
+                    target_shard,
+                );
+            }
+            let prompt = format!(
+                "`{}` already exists on shard `{}` and will be reused as the active `{}` chain.\nProceed?",
+                canonical_backup_name, target_shard, chain_name
+            );
+            if !prompt_for_confirmation(&prompt)? {
+                println!(
+                    "Aborting. Remove `{}` with `graphman chain remove {}` if you want to create a fresh cache on shard `{}`.",
+                    canonical_backup_name, canonical_backup_name, target_shard
+                );
+                return Ok(());
+            }
+            true
+        }
+    };
+
+    let existing_backup_store = if reuse_existing_backup {
+        store.chain_store(&canonical_backup_name).await
+    } else {
+        None
+    };
+
+    if !reuse_existing_backup {
+        let chain =
+            BlockStore::allocate_chain(&mut conn, &chain_name, &target_shard, &ident).await?;
+        store.add_chain_store(&chain, true).await?;
+    }
+
+    let temp_backup_name = format!("{chain_name}-old-temp");
+    if reuse_existing_backup && find_chain(&mut conn, &temp_backup_name).await?.is_some() {
+        bail!(
+            "`{}` already exists. Remove it with `graphman chain remove {}` before changing `{}` to shard `{}`",
+            temp_backup_name,
+            temp_backup_name,
+            chain_name,
+            target_shard,
+        );
+    }
 
     conn.transaction::<(), StoreError, _>(async |conn| {
-        let shard = Shard::new(shard.to_string())?;
-
-        let chain = BlockStore::allocate_chain(conn, &chain_name, &shard, &ident).await?;
-
-        store.add_chain_store(&chain, true).await?;
-
-        // Drop the foreign key constraint on deployment_schemas
         sql_query(
             "alter table deployment_schemas drop constraint deployment_schemas_network_fkey;",
         )
         .execute(conn)
         .await?;
 
-        // Update the current chain name to chain-old
-        update_chain_name(conn, &chain_name, &new_name).await?;
+        if let Some(backup) = existing_backup.as_ref() {
+            update_chain_name(conn, &backup.name, &temp_backup_name).await?;
+        }
 
-        // Create a new chain with the name in the destination shard
-        let _ = add_chain(conn, &chain_name, &shard, ident).await?;
+        update_chain_name(conn, &chain_name, &canonical_backup_name).await?;
 
-        // Re-add the foreign key constraint
+        if reuse_existing_backup {
+            update_chain_name(conn, &temp_backup_name, &chain_name).await?;
+        } else {
+            add_chain(conn, &chain_name, &target_shard, ident.clone()).await?;
+        }
+
         sql_query(
             "alter table deployment_schemas add constraint deployment_schemas_network_fkey foreign key (network) references chains(name);",
         )
@@ -289,11 +356,15 @@ pub async fn change_block_cache_shard(
     })
     .await?;
 
-    chain_store.update_name(&new_name).await?;
+    chain_store.update_name(&canonical_backup_name).await?;
+
+    if reuse_existing_backup && let Some(backup_store) = existing_backup_store.as_ref() {
+        backup_store.update_name(&chain_name).await?;
+    }
 
     println!(
         "Changed block cache shard for {} from {} to {}",
-        chain_name, old_shard, shard
+        chain_name, old_shard, target_shard
     );
 
     Ok(())
--- a/node/src/manager/commands/chain.rs
+++ b/node/src/manager/commands/chain.rs
@@ -253,7 +253,6 @@ pub async fn change_block_cache_shard(
     let canonical_backup_name = format!("{chain_name}-old");
 
     let existing_backup = find_chain(&mut conn, &canonical_backup_name).await?;
-    let existing_backup_store = store.chain_store(&canonical_backup_name).await;
 
     println!("Current shard: {}", old_shard);
 


lutter:
We really need to review this, this shouldn't have been closed

erayack:
Hi @lutter, I rebased PR #6199 onto current master and resolved the conflict in node/src/manager/commands/chain.rs.

The PR is now mergeable. I kept the original fix intent, but ported it to the current async chain-store code, preserved the safe chain-store creation ordering, simplified the backup-name selection flow, and added focused tests for backup-name formatting.

node/src/manager/commands/chain.rs
```diff
@@ -250,51 +288,112 @@ pub async fn change_block_cache_shard(
...
-
-            // Drop the foreign key constraint on deployment_schemas
             sql_query(
                 "alter table deployment_schemas drop constraint deployment_schemas_network_fkey;",
             )
             .execute(conn).await?;
 
-            // Update the current chain name to chain-old
-            update_chain_name(conn, &chain_name, &new_name).await?;
+            let previous_backup_final_name = if let Some(backup) = existing_backup.as_ref() {
```
dimitrovmaksim: Your code will hit the same constraint in the following case:
mainnet on shard_a
mainnet-old on shard_b

Then running:
`graphman chains change-shard mainnet shard_b`
 
- Line 326 computes the next free backup name, e.g. `mainnet-old-1`.
- Line 327 renames `mainnet-old` -> `mainnet-old-1`.
- Line 330 then tries to rename `mainnet-old-1` -> `mainnet`.

At that point the original `mainnet` row still exists, so the rename to mainnet violates the unique constraint on chains.name. The current `mainnet` -> `mainnet-old-{next}` rename only happens later on lines 339-340.

More generally, I’m not sure preserving multiple old caches is the right approach here. Moving a chain cache between shards should be a rare maintenance operation, and old caches are only useful as an immediate rollback. Reusing an existing backup on the target shard could make sense for an immediate rollback, although in other cases it may be too outdated. For a sequence like A -> B -> C, keeping the old cache on A does not seem useful, so probably dropping that previous backup before preserving the current B cache as the new mainnet-old would make the behavior simpler and avoid having multiple stale caches around the shards.

dimitrovmaksim (changes requested):

node/src/manager/commands/chain.rs
erayack: I updated the flow to keep only one rollback backup and to order the reuse case as mainnet-old -> temp, mainnet -> mainnet-old, temp -> mainnet so the unique constraint cannot fire. 

One thing I’m less certain about is policy: when an older mainnet-old exists and we are not reusing it, do you want that old backup dropped only after the new change-shard succeeds, or should it be deleted earlier as part of simplifying the behavior? Also, are you okay with updating the in-memory BlockStore mapping explicitly during these renames, or would you prefer a simpler approach even if it leaves cleanup to a reload?

dimitrovmaksim: Sorry for the late reply, I was contemplating what we really want to happen here, so we don't make you go back and forth. What i'm thinking is:
- Add a check if `{chain}-old` already exists and:
  - If its shard matches the target shard, prompt the user that it will re-use that backup (they can decline and remove it with `chain remove` if they want)
  - If target shard is a different shard, print hint to use `chain remove`.

The point is to give more control to the indexers rather than doing stuff for then in the background

erayack:
Implemented the behavior you suggested. @dimitrovmaksim 

Current flow is now:

- if {chain}-old exists on a different shard, the command aborts and tells the user to remove it explicitly with graphman chain remove
- if {chain}-old exists on the target shard, the command prompts before reusing it
- otherwise it keeps the current cache as {chain}-old and creates a fresh cache on the target shard

I also changed the reuse ordering to avoid the unique constraint issue: {chain}-old -> temp, {chain} -> {chain}-old, temp -> {chain}

I added focused unit tests for the boundary decision logic.

node/src/manager/commands/chain.rs
```diff
@@ -250,51 +305,134 @@ pub async fn change_block_cache_shard(
...
+            }
 
-            // Re-add the foreign key constraint
             sql_query(
                 "alter table deployment_schemas add constraint deployment_schemas_network_fkey foreign key (network) references chains(name);",
             )
             .execute(conn).await?;
-            Ok(())
+
+            Ok(ChainSwapOutcome {
```
dimitrovmaksim: I think this return value is redundant. The transaction does not compute any new state; ChainSwapOutcome just returns `reuse_existing_backup` and `allocated_chain.is_some()`, which are already known before entering the transaction. This can return () and the struct can be completelly removed.

node/src/manager/commands/chain.rs
```diff
@@ -250,51 +305,134 @@ pub async fn change_block_cache_shard(
...
+    let existing_backup = find_chain(&mut conn, &canonical_backup_name).await?;
 
     println!("Current shard: {}", old_shard);
 
     let chain_store = store
         .chain_store(&chain_name)
         .await
         .ok_or_else(|| anyhow!("unknown chain: {}", &chain_name))?;
-    let new_name = format!("{}-old", &chain_name);
     let ident = chain_store.chain_identifier().await?;
```
dimitrovmaksim, line 263: One thing I missed is that the current chain and the backup Identities may differ. `graphman` provides a command to update the genesis hash `graphman chain update-genesis`, which is part of the ChainIdentifier, so we should fetch the backup identifier as well and compare them and abort If they are different.

node/src/manager/commands/chain.rs
```diff
@@ -250,51 +305,134 @@ pub async fn change_block_cache_shard(
...
+            );
+            if !prompt_for_confirmation(&prompt)? {
+                println!(
+                    "Aborting. Remove `{}` with `graphman chain remove {}` if you want to create a fresh cache on shard `{}`.",
+                    canonical_backup_name, canonical_backup_name, target_shard
+                );
+                return Ok(());
+            }
+        }
+    }
```
dimitrovmaksim: The `ExistingBackupDisposition` enum seems a little over-modeled for this control flow. It only wraps the None / same shard / wrong shard cases. 
It can be inlide to something like
```rust
let reuse_existing_backup = match existing_backup.as_ref() {
      None => false,
      Some(backup) if backup.shard != target_shard => {
          bail!(...);
      }
      Some(backup) => {
         ...
      }
  };
```
This will remove the need to manage the lifetime on the enum.

node/src/manager/commands/chain.rs
```diff
@@ -250,51 +305,134 @@ pub async fn change_block_cache_shard(
...
                 "alter table deployment_schemas drop constraint deployment_schemas_network_fkey;",
             )
             .execute(conn).await?;
 
-            // Update the current chain name to chain-old
-            update_chain_name(conn, &chain_name, &new_name).await?;
-
-            // Create a new chain with the name in the destination shard
-            let _ = add_chain(conn, &chain_name, &shard, ident).await?;
+            let temp_backup_name = if let Some(backup) = existing_backup.as_ref() {
```
dimitrovmaksim: The temp backup name helpers may be too much for what we need. It's only a transient name for the three-way rename, so something hardcoded like `{chain}-old-temp` should probably be enough.

node/src/manager/commands/chain.rs
```diff
@@ -250,51 +305,134 @@ pub async fn change_block_cache_shard(
...
+                update_chain_name(conn, &backup.name, &temp_name).await?;
+                Some(temp_name)
+            } else {
+                None
+            };
+
+            update_chain_name(conn, &chain_name, &canonical_backup_name).await?;
+
+            if reuse_existing_backup {
+                debug_assert!(temp_backup_name.is_some());
```
dimitrovmaksim: I would not use the `debug_assert!` here, for --release build it will be compiled out, and it will panic on the unwrap() after that anyways.

erayack:
Hi @dimitrovmaksim again. Force-pushed 334940b → 704e10d addressing the latest round:

- Removed redundant transaction outcome plumbing (ChainSwapOutcome) and made the transaction return ().
- Inlined the backup disposition control flow and removed the extra enum/helper layering.
 - Added chain-identifier safety checks for backup reuse ({chain}-old must match the active chain identifier), switched transient rename to {chain}-old-temp, and removed the debug_assert!/unwrap() path.

  Tests still pass locally. Ready for another look when you have a moment.

node/src/manager/commands/chain.rs
```diff
@@ -250,51 +251,138 @@ pub async fn change_block_cache_shard(
...
+    println!("Latest backup recorded as `{}`", canonical_backup_name);
+
+    if reuse_existing_backup {
+        println!(
+            "Reused existing backup `{}` as the active `{}` chain",
+            canonical_backup_name, chain_name
+        );
+    }
+
+    if allocated_chain.is_some() {
```
dimitrovmaksim: nit: Same as the `allocated_chain = ...` comment. I think this print is redundant.

node/src/manager/commands/chain.rs
```diff
@@ -250,51 +251,138 @@ pub async fn change_block_cache_shard(
...
+
+    if reuse_existing_backup && let Some(backup_store) = existing_backup_store.as_ref() {
+        backup_store.update_name(&chain_name).await?;
+    }
 
     println!(
         "Changed block cache shard for {} from {} to {}",
         chain_name, old_shard, shard
     );
+    println!("Latest backup recorded as `{}`", canonical_backup_name);
```
dimitrovmaksim: nit: I would remove this println!, seems like a left-over from the sufixed backup names

node/src/manager/commands/chain.rs
```diff
@@ -250,51 +251,138 @@ pub async fn change_block_cache_shard(
...
-        async {
-            let shard = Shard::new(shard.to_string())?;
+    let existing_backup_store = if reuse_existing_backup {
+        store.chain_store(&canonical_backup_name).await
+    } else {
+        None
+    };
 
-            let chain = BlockStore::allocate_chain(conn, &chain_name, &shard, &ident).await?;
+    let allocated_chain = if reuse_existing_backup {
```
dimitrovmaksim: nit: `allocated_chain` could be boolean, as it is used only for the conditional printing here https://github.com/erayack/graph-node/blob/8b649af8184d6a83e730b4ddb74289c07d3c5cd8/node/src/manager/commands/chain.rs#L380-L385. Tbf I think this message may be redundant as we already notify the user that the chain has been moved to another shard. Personally i would remove that print and change this to 
```
if !reuse_existing_backup {
    let chain = BlockStore::allocate_chain(&mut conn, &chain_name, &target_shard, &ident).await?;
    store.add_chain_store(&chain, true).await?;
}
```

node/src/manager/commands/chain.rs
```diff
@@ -250,51 +251,138 @@ pub async fn change_block_cache_shard(
...
 
     println!("Current shard: {}", old_shard);
 
     let chain_store = store
         .chain_store(&chain_name)
         .await
         .ok_or_else(|| anyhow!("unknown chain: {}", &chain_name))?;
-    let new_name = format!("{}-old", &chain_name);
     let ident = chain_store.chain_identifier().await?;
+    let target_shard = Shard::new(shard.clone())?;
```
dimitrovmaksim: nit: You can drop the `clone()` here and let Shard::new consume the String, then you can use `target_shard` directly later as print arguments instead of `shard`

dimitrovmaksim:
Hey @erayack sorry for the delay, I added couple of nit comment. I'll ask another team member to take a look before approving, just to be sure I'm not missing something.

Btw could you drop the merge commit and instead rebase your branch ontop of master? 🙏🏻

erayack:
Hi @dimitrovmaksim  Addressed the nits, dropped the merge commit, and rebased onto master. 

Ready for another look whenever you have a moment, and happy for you to loop in the second reviewer in parallel.

incrypto32 (approved):
LGTM

dimitrovmaksim (approved):

dimitrovmaksim:
Hi @erayack, thank you for your contribution! Merging now

erayack:
Thanks @dimitrovmaksim and @incrypto32 for the reviews, and thanks @lutter for bringing this back up. Appreciate the guidance on the rollback behavior and cleanup along the way.

Issue #5470, opened 2024-06-06
expose table sizes in the indexing status endpoint


cargopete:
### Description

There should be a new resolver added (for Subgraph size), the resolver should map to the appropriate SQL query.

### Some information to help us out

- [X] Tick this box if you plan on implementing this feature yourself.
- [X] I have searched the issue tracker to make sure this issue is not a duplicate.

PR #6201, opened 2025-10-20
Feat: Expose subgraph table sizes in SubgraphIndexingStatus

dimitrovmaksim:
Resolves https://github.com/graphprotocol/graph-node/issues/5470

What this PR does:

1. Exposes (sum of) the deployments table sizes and rows through `SubgraphIndexingStatus`. The data is collected from the `subgraph_sizes` materialized view.
Screenshot 2025-10-20 at 15 25 41

2. Makes the `subgraphs` argument for ther `indexingStatuses` query required. After the changes in https://github.com/graphprotocol/graph-node/pull/6143, calling the query without an argument is redundant because it will always return an empty array.

### DIFF
--- a/store/postgres/src/detail.rs
+++ b/store/postgres/src/detail.rs
@@ -437,7 +437,7 @@ pub(crate) async fn deployment_statuses(
         .collect()
     };
 
-    let mut deployment_sizes = deployment_sizes(conn, sites)?;
+    let mut deployment_sizes = deployment_sizes(conn, sites).await?;
 
     details_with_fatal_error
         .into_iter()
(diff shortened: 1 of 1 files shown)


store/postgres/src/detail.rs
```diff
@@ -422,17 +425,93 @@ pub(crate) fn deployment_statuses(
...
+    } else {
+        query.push_str(" WHERE ds.id = ANY($1)");
+        diesel::sql_query(query)
+            .bind::<Array<Integer>, _>(sites.iter().map(|site| site.id).collect::<Vec<_>>())
+            .load::<SubgraphSizeRow>(conn)
+    };
+
+    let rows = match result {
+        Ok(rows) => rows,
+        Err(e) if e.to_string().contains("has not been populated") => Vec::new(),
```
incrypto32, line 507: Minor nitpick: Is there no typed error for this? string matching might be fragile if diesel changes the error message.

incrypto32: Feel free to ignore if there is no typed error for this

dimitrovmaksim: This is a postgre error, and it is returned as `Unknown`
```
Unknown(materialized view "subgraph_sizes" has not been populated)
```
So I don't think there's another way

dimitrovmaksim:
Let's wait for this PR https://github.com/graphprotocol/graph-node/pull/6194 to be merged before this one, so I can update to async logic

Issue #4579, opened 2023-04-27
[Feature] Automate setting of account-like optimization


lutter:
The account-like optimization is turned on with `graphman stats account-like <sgdNNN> <table>` and changes query generation for that table by adding to the clause `block_range @> $block` that all our queries contain the redundant clauses `lower(block_range <= $block and coalesce(upper(block_range), 2147483647) > $block`. That makes the BRIN index on these expressions that we have on all tables usable by the query planner; it can bring about a dramatic improvement in query speed when it works, but unfortunately, it can also slow down queries a lot in some cases. That's why we don't use it by default.

The general rule of thumb is that if the ratio of entities to versions (`count(distinct id)/count(*)`) is very low, say < 0.1%, that the account-like optimization can be useful. Because of the BRIN index, it is very dependent on the exact layout of data on disk. In Uniswap-like subgraphs, it usually is very effective for tables like `token` and `pair`, since in those, versions that are visible at a certain block tend to cluster. It's usually not effective for tables like `token_day_data` or `pair_hour_data` where versions with an unlimited upper bound on their block range are interspersed with very old versions throughout the heap. It's also not effective for entities that are immutable, as they all have an unlimited upper bound on their block range - if subgraph authors declare these entities as immutable, this isn't an issue as we don't store a block range in that case, just a start block and can use BTree indexes to search. But for entities that are immutable but not declared as such it can become an issue.

It would be great if we could have a reliable indication of when the account-like optimization will be effective. For example, we could look at how selective the BRIN index is at higher block numbers and adapt query generation accordingly. It's not clear to me how best we would gauge the selectiveness of the BRIN index; we might have to resort to `page_inspect` or some such.

avandras:
If by selectivity you mean non-overlappiness (I mean, how much the data is in order, therefore the value ranges of different index blocks _don't_ overlap), then `pageinspect` is indeed a good candidate.  

Otherwise, as I'm lacking the history of development, have you ever considered to keep the lower and upper bounds separate, and index them using `btree`?  Especially in the case when there is a lot of overlapping in the BRIN, this might help (depending on cardinality, of course).  It would also make it possible to replace the gist indexes (as far as I see, they are not used for actual exclusion).

PR #6209, opened 2025-11-06
Feat: Automate account-like optimization

dimitrovmaksim:
Resolves https://github.com/graphprotocol/graph-node/issues/4579

What this PR does:

1. Adds a new materialized view `info.table_stats` that aggregates statistics from pg_stats for subgraph entity tables, calculating estimated unique entities, total versions, and uniqueness ratio.
2. Adds a optional experimental background job querying the view to find eligible tables and setting the `is_account_like` flag. The job can be enabled by setting the following env vars 

- `GRAPH_STORE_ACCOUNT_LIKE_SCAN_INTERVAL_HOURS`
- `GRAPH_STORE_ACCOUNT_LIKE_MIN_VERSION_COUNT`
- `GRAPH_STORE_ACCOUNT_LIKE_MAX_UNIQUE_RATIO`

3. Updates the ENV vars documentation with the new variables and their usage.

### DIFF
--- a/store/postgres/migrations/2025-10-30-094807-0000_table_stats_view/down.sql
+++ b/store/postgres/migrations/2025-10-30-094807-0000_table_stats_view/down.sql
@@ -1,1 +1,0 @@
-DROP MATERIALIZED VIEW IF EXISTS info.table_stats;
--- a/store/postgres/migrations/2025-12-18-104956-0000_create_entity_version_stats_view/down.sql
+++ b/store/postgres/migrations/2025-12-18-104956-0000_create_entity_version_stats_view/down.sql
@@ -1,0 +1,1 @@
+DROP MATERIALIZED VIEW IF EXISTS info.entity_version_stats;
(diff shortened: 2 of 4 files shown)


store/postgres/migrations/2025-10-30-094807-0000_table_stats_view/up.sql
```diff
@@ -0,0 +1,35 @@
+CREATE MATERIALIZED VIEW info.table_stats AS
```
lutter: It's a little unfortunate that we already have a `subgraphs.table_stats` which has a completely different meaning. It might be good to use a different name here, maybe `entity_version_stats`, though more than happy with something else

lutter (commented):
Looks good, but would be nice to avoid the nameclash I noted.

store/postgres/migrations/2025-10-30-094807-0000_table_stats_view/up.sql
dimitrovmaksim: I guess it is true that the hardest thing in development is naming stuff :D 

Removed the previous migration and added a new one with the name of the view changed to your suggestion

Issue #5513, opened 2024-06-26
[Feature] Make call cache a config option for Indexers with locally hosted RPC nodes


alex-pakalniskis:
### Description

Feature request from Vincent T. of Data Nexus during June 25, 2024 edition of IOH
![image](https://github.com/graphprotocol/graph-node/assets/43630382/f93ec8e6-24b8-40d6-a0dc-e1b9954f7e64)

### Are you aware of any blockers that must be resolved before implementing this feature? If so, which? Link to any relevant GitHub issues.

No

### Some information to help us out

- [ ] Tick this box if you plan on implementing this feature yourself.
- [X] I have searched the issue tracker to make sure this issue is not a duplicate.

PR #6214, opened 2025-11-12
Feat: Add option to disable store call cache

dimitrovmaksim:
Resolves https://github.com/graphprotocol/graph-node/issues/5513

What this PR does:

Introduces an environment variable `GRAPH_STORE_DISABLE_CALL_CACHE` that allows disabling the store call cache. The main motivation is to provide indexers, who use locally hosted RPC nodes, with an option to bypass the graph-node call cache.

### DIFF
--- a/graph/src/env/store.rs
+++ b/graph/src/env/store.rs
@@ -160,8 +160,7 @@ pub struct EnvVarsStore {
     /// Set by the environment variable `GRAPH_STORE_ACCOUNT_LIKE_MAX_UNIQUE_RATIO`
     /// Defines the maximum share of unique entities (e.g. 0.01 for a 1:100 entity-to-version ratio).
     pub account_like_max_unique_ratio: Option<f64>,
-    /// Disables the store call cache entirely. Graph node will skip writing and reading from the
-    /// call cache. The buffered block call cache will still be enabled.
+    /// Disables storing or reading `eth_call` results from the store call cache.
     /// Set by `GRAPH_STORE_DISABLE_CALL_CACHE`. Defaults to false.
     pub disable_call_cache: bool,
 }


docs/environment-variables.md
```diff
@@ -294,3 +294,7 @@ those.
   `GRAPH_STORE_ACCOUNT_LIKE_MIN_VERSIONS_COUNT` and `GRAPH_STORE_ACCOUNT_LIKE_MAX_UNIQUE_RATIO`.
 - `GRAPH_STORE_ACCOUNT_LIKE_MIN_VERSIONS_COUNT`: Sets the minimum total number of versions a table must have to be considered for account-like flagging. Expects a positive integer value. No default value.
 - `GRAPH_STORE_ACCOUNT_LIKE_MAX_UNIQUE_RATIO`: Sets the maximum unique entities to version ratio (e.g., 0.01 ≈ 1:100 entity-to-version ratio).
+- `GRAPH_STORE_DISABLE_CALL_CACHE`: Disables the store call cache entirely. Graph node will skip writing and reading from the
+  call cache. The buffered block call cache will still be enabled. This option may be useful
+  for indexers who are running their own RPC nodes. Disabling the store call cache may have
+  significant performance impact. (default: false)
```
lutter: What do you mean by `The buffered block call cache` - do you mean just the block cache?

Also, I would reword 'Disabling the store call cache may have significant performance impact' to something like 'Disabling the store call cache may have significant performance impact; the precise impact depends on how long an eth_call takes on average compared to how long a database lookup of a cached call result takes'

dimitrovmaksim: > What do you mean by The buffered block call cache - do you mean just the block cache?

I meant the in memory call cache that buffers the call results for a duration of a block, before the results is persisted in the DB. Will make this line more clear with something like `Call results will still be cached in memory for the duration of a block`

lutter: Aah; I wouldn't mention that here, it just complicates the description of the environment variable

PR #6215, opened 2025-11-14
graph: emit a block only when there's a block_ptr

cjorge-graphops:
After this was merged https://github.com/graphprotocol/graph-node/pull/5160 substreams fatal errors surface and can write SubgraphError{ block_ptr: None }. This introduces a bug that shows in Status API as those records can have fatalError.block = null violating the schema of type Block { hash: Bytes! number: BigInt! }

A later commit https://github.com/graphprotocol/graph-node/pull/5971 makes it much easier to trigger this when calling status API to return all subgraphs (empty [] array for deployment ID) as failed subgraphs linger in the indexing status set as paused instead of being unassigned.

### DIFF
--- a/graph/src/data/subgraph/status.rs
+++ b/graph/src/data/subgraph/status.rs
@@ -191,7 +191,8 @@ mod tests {
 
     #[test]
     fn subgraph_error_block_is_null_without_pointer() {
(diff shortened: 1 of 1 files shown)


lutter (approved):
Sorry for the long delay. If you are still interested in getting this merged, can you rebase it to latest master and squash the two commits into one? I will merge it then

PR #6218, opened 2025-11-21
Amp-powered subgraphs

isum:
[Amp](https://github.com/edgeandnode/amp)-powered subgraphs are a new kind of subgraphs with SQL data sources that query and index data from the Amp servers. They are significantly more efficient than the standard subgraphs, and the indexing time can be reduced from days and weeks, to minutes and hours in most cases.

### Documentation

Documentation on creating Amp-powered subgraphs is available [here](https://github.com/graphprotocol/graph-node/blob/ion/amp-powered-subgraphs/docs/amp-powered-subgraphs.md).

### Examples

Complete examples of how to create, deploy and query Amp-powered subgraphs are available in a separate repository:
https://github.com/edgeandnode/amp-subgraph-examples

### DIFF
--- a/core/src/amp_subgraph/metrics.rs
+++ b/core/src/amp_subgraph/metrics.rs
@@ -35,7 +35,7 @@ impl Metrics {
         let stopwatch = StopwatchMetrics::new(
             logger.cheap_clone(),
             deployment.cheap_clone(),
-            "amp-process",
+            "process",
             metrics_registry.cheap_clone(),
             store.shard().to_string(),
         );
@@ -123,7 +123,11 @@ impl DeploymentHead {
             .new_int_gauge(
                 "deployment_head",
                 "Tracks the most recent block number processed by a deployment",
-                const_labels,
+                const_labels
+                    .into_iter()
+                    .map(|(k, v)| (k.to_string(), v.to_string()))
+                    // TODO: Pass the network from the Amp manifest
+                    .chain([("network".to_string(), "".to_string())]),
             )
             .expect("failed to register `deployment_head` gauge");
 
@@ -139,7 +143,7 @@ impl DeploymentHead {
     }
 }
 
-/// Tracks the target block number of a deployment.
+/// Tracks the maximum block number currently available for indexing within a deployment.
 pub(super) struct DeploymentTarget(IntGauge);
 
 impl DeploymentTarget {
@@ -149,8 +153,8 @@ impl DeploymentTarget {
     ) -> Self {
         let int_gauge = metrics_registry
             .new_int_gauge(
-                "amp_deployment_target",
-                "Tracks the target block number of a deployment",
+                "deployment_target",
+                "Tracks the maximum block number currently available for indexing within a deployment",
                 const_labels,
             )
             .expect("failed to register `amp_deployment_target` gauge");
@@ -217,7 +221,7 @@ impl IndexingDuration {
     ) -> Self {
         let int_counter = metrics_registry
             .new_int_counter(
-                "amp_deployment_indexing_duration_seconds",
+                "deployment_indexing_duration_seconds",
                 "Tracks the total duration in seconds of deployment indexing",
                 const_labels,
             )
--- a/core/src/amp_subgraph/monitor.rs
+++ b/core/src/amp_subgraph/monitor.rs
@@ -439,13 +439,6 @@ impl Monitor {
                         error!(logger, "Subgraph panicked";
                             "error" => ?e
                         );
-
-                        // TODO: Maybe abort the entire process on panic and require a full graph-node restart.
-                        //       Q: Should a bug that is triggered in a specific subgraph affect everything?
-                        //       Q: How to make this failure loud enough so it is not missed?
-                        //
-                        // println!("Subgraph panicked");
-                        // std::process::abort();
                     }
                     Err(e) => {
                         error!(logger, "Subgraph failed";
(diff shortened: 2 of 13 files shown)


core/src/amp_subgraph/runner/compat.rs
```diff
@@ -0,0 +1,50 @@
...
+
+mod legacy {
+    pub(super) use graph::{
+        blockchain::{BlockHash, BlockPtr, BlockTime},
+        components::store::BlockNumber,
+        data::store::scalar::Timestamp,
+    };
+}
+
+pub(super) trait Compat<T> {
```
lutter, line 14: Not a big deal, but `Compat` is maybe a bit of a misnomer, maybe `FromAlloy` would be better

core/src/nozzle_subgraph/runner/reorg_handler.rs
```diff
@@ -0,0 +1,162 @@
...
+        common::Ident,
+        Client,
+    },
+};
+use itertools::Itertools;
+use slog::debug;
+
+use super::{Compat, Context, Error, LatestBlocks};
+
+pub(super) async fn check_and_handle_reorg<NC>(
```
lutter: This is really important. It would be good to have a comment on how this works and why. It might also be worth explaining what the other two methods here do.

core/src/nozzle_subgraph/metrics.rs
```diff
@@ -0,0 +1,36 @@
...
+    pub(super) fn new(
+        logger: &Logger,
+        metrics_registry: Arc<MetricsRegistry>,
+        store: Arc<dyn WritableStore>,
+        deployment: DeploymentHash,
+    ) -> Self {
+        let stopwatch = StopwatchMetrics::new(
+            logger.cheap_clone(),
+            deployment,
+            "nozzle-process",
```
lutter: Just make that `process`, it's easier on existing dashboards and it's clear from the manifest what kind of subgraph it is

core/src/amp_subgraph/runner/data_processing.rs
```diff
@@ -0,0 +1,231 @@
...
+        entity_lfu_cache,
+        evict_stats: _,
+    } = entity_cache
+        .as_modifications(block_number.compat())
+        .map_err(Error::from)
+        .map_err(|e| e.context("failed to extract entity modifications from the state"))?;
+
+    let is_close_to_chain_head = latest_block.saturating_sub(block_number) <= 100;
+
+    cx.store
```
lutter, line 151: This should be marked in the stopwatch as `transact_block`

core/src/amp_subgraph/runner/data_processing.rs
```diff
@@ -0,0 +1,231 @@
...
+        )
+        .await
+        .map_err(|e| {
+            e.context(format!(
+                "failed to process record batch for stream '{stream_index}'"
+            ))
+        })?;
+    }
+
+    let ModificationsAndCache {
```
lutter, line 137: This should be marked in the stopwatch as `as_modifications`

core/src/amp_subgraph/monitor.rs
```diff
@@ -0,0 +1,573 @@
...
+                        error!(logger, "Subgraph panicked";
+                            "error" => ?e
+                        );
+
+                        // TODO: Maybe abort the entire process on panic and require a full graph-node restart.
+                        //       Q: Should a bug that is triggered in a specific subgraph affect everything?
+                        //       Q: How to make this failure loud enough so it is not missed?
+                        //
+                        // println!("Subgraph panicked");
+                        // std::process::abort();
```
lutter: We shouldn't do that; yes, it's hard to spot a panic in lots of logs, but taking down the process would mean that one subgraph can affect lots of other subgraphs.

core/src/amp_subgraph/metrics.rs
```diff
@@ -25,12 +34,227 @@ impl Metrics {
...
+    pub(super) fn update(&self, new_most_recent_block_number: BlockNumber) {
+        self.0.set(
+            i64::try_from(new_most_recent_block_number)
+                .expect("new most recent block number does not fit into `i64`"),
+        );
+    }
+}
+
+/// Tracks the target block number of a deployment.
+pub(super) struct DeploymentTarget(IntGauge);
```
lutter, line 147: I don't understand what 'target' means here - is this basically the latest block that Amp has available?

core/src/amp_subgraph/metrics.rs
```diff
@@ -25,12 +34,227 @@ impl Metrics {
...
+pub(super) struct DeploymentHead(IntGauge);
+
+impl DeploymentHead {
+    fn new(
+        metrics_registry: &MetricsRegistry,
+        const_labels: impl IntoIterator<Item = (impl ToString, impl ToString)>,
+    ) -> Self {
+        let int_gauge = metrics_registry
+            .new_int_gauge(
+                "amp_deployment_head",
```
lutter: _I think that was addressed in a later commit, so ignore_

I think prefixing all these with `amp` when they correspond to an existing metric like `deployment_head` makes using them in dashboards much harder. When metrics mean the same thing regardless of subgraph flavor, we should use the same metric name

docs/amp-powered-subgraphs.md
```diff
@@ -0,0 +1,407 @@
...
+
+Amp-powered subgraphs fully support the subgraph composition feature.
+This allows applying complex subgraph mappings on top of data indexed from the Amp servers.
+
+For more information on using the powerful subgraph composition feature,
+refer to the [documentation](https://github.com/graphprotocol/example-composable-subgraph).
+
+## ENV variables
+
+Amp-powered subgraphs feature introduces the following new ENV variables:
```
lutter, line 542: While this is still experimental, it's fine to have that here, but at some point we should move this to `docs/environment-variables.md`

PR #6219, opened 2025-12-01
graph, runtime: Do not panic when converting negative number to U256

lutter:
Return an error instead

### DIFF
--- a/graph/src/data/store/scalar/bigint.rs
+++ b/graph/src/data/store/scalar/bigint.rs
@@ -1,3 +1,4 @@
+use anyhow::bail;
 use num_bigint;
 use serde::{self, Deserialize, Serialize};
 use stable_hash::utils::AsInt;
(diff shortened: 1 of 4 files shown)


graph/src/data/store/scalar/bigint.rs
```diff
@@ -207,14 +208,15 @@ impl BigInt {
...
-    pub fn to_unsigned_u256(&self) -> U256 {
+    pub fn to_unsigned_u256(&self) -> Result<U256, anyhow::Error> {
         let (sign, bytes) = self.to_bytes_le();
-        assert!(
-            sign == BigIntSign::NoSign || sign == BigIntSign::Plus,
-            "negative value encountered for U256: {}",
-            self
-        );
-        U256::from_little_endian(&bytes)
+        if sign != BigIntSign::Plus && sign != BigIntSign::NoSign {
```
isum, line 213: A `Minus` variant is available. It may be clearer to use it

```suggestion
        if BigIntSign::Minus {
```

PR #6224, opened 2025-12-09
Avoid deadlocks under heavy load, especially during index node startup

lutter:


store/postgres/src/pool/mod.rs
```diff
@@ -42,6 +43,35 @@ type AsyncPool = deadpool::managed::Pool<ConnectionManager>;
...
+    conn: AsyncPgConnection,
+    _permit: tokio::sync::OwnedSemaphorePermit,
+}
+
+impl PermittedConnection {
+    /// Returns the underlying connection, consuming the permit.
+    /// The permit is released when this method is called.
+    pub fn into_inner(self) -> AsyncPgConnection {
+        self.conn
+    }
```
isum: Why is this necessary? If this method is used a lot, it seems to defeat the purpose of using the permits. Maybe it is possible to always use a reference to the inner connection, so that the permit is alive exactly as long as the connection?

store/postgres/src/pool/mod.rs
```diff
@@ -42,6 +43,35 @@ type AsyncPool = deadpool::managed::Pool<ConnectionManager>;
 /// A database connection for asynchronous diesel operations
 pub type AsyncPgConnection = deadpool::managed::Object<ConnectionManager>;
 
+/// A database connection bundled with a semaphore permit.
+/// The permit is held for the lifetime of the connection, providing
+/// backpressure to prevent pool exhaustion during mass operations.
```
isum, line 49: This does not seem to be accurate because of the `into_inner` method

store/postgres/src/pool/mod.rs
lutter: I just added a commit that gets rid of that method

isum (approved):

PR #6225, opened 2025-12-15
Extend supported types for `first` and `last` aggregations

isum:
This PR makes it possible to use the `first` and `last` aggregations with the following types: `String`, `Bytes`, and references to entity types.


store/postgres/migrations/2025-12-09-164812_extend_arg_min_max/generate.sh
```diff
@@ -0,0 +1,98 @@
+#! /bin/bash
```
lutter: I don't think you need to check this in - just use the original generate.sh and locally change the `types` variable. Having up/down checked in is enough. If you want to be really thorough, add a note to this directory that that's how the migrations were created

Issue #6220, opened 2025-12-01
[Bug] Grafting Fails Due to Entity ID Type Mismatch (`ID!` vs. `Bytes!`)


0x19dG87:
## Summary

Subgraph syncing fails during the **grafting** process because of a type mismatch for the **ID attribute of an entity** between the source (graft) and the destination (new subgraph). Specifically, the destination schema defines the attribute as `Bytes!`, while the source schema defines it as `ID!`. This causes the data copy operation to fail immediately with a `store error`, preventing the new subgraph from ever syncing.

## Steps to Reproduce

1.  Start indexing subgraph `QmWZnioK8S9HL45kQqKFd67NxeqM1uEoVhYrrbcPDxuQSV` and its graft `QmQfYe5Ygg9A3mAiuBZYj5a64bDKLF4gF6sezfhgxKvb9y`.
2.  Wait graft to reach copy point
3.  Check in graphman copying process `graphman copy list` started.
4. Open graph node logs and wait for error. You can restart subgraph in graphman to catch error (`graphman restart QmWZnioK8S9HL45kQqKFd67NxeqM1uEoVhYrrbcPDxuQSV`).

## Observed Behavior

The subgraph fails during the data copy phase of the graft with a `store error`. The relevant logs show the following exception:

```
The attribute User.id has type Bytes!, but its type in the source is ID!
```

**Relevant Logs:**

```
Dec 01 21:08:53.909 INFO Initializing graft by copying data from sgd91 to sgd108, runner_index: 28, sgd: 108, subgraph_id: QmWZnioK8S9HL45kQqKFd67NxeqM1uEoVhYrrbcPDxuQSV, component: SubgraphInstanceManager
...
Dec 01 21:08:55.953 INFO Starting to copy `User` entities from "sgd91"."user" to "sgd108"."user", dst: sgd108, runner_index: 28, sgd: 108, subgraph_id: QmWZnioK8S9HL45kQqKFd67NxeqM1uEoVhYrrbcPDxuQSV, component: SubgraphInstanceManager
...
Dec 01 21:08:55.956 ERRO copy worker had an error: store error: The attribute User.id has type Bytes!, but its type in the source is ID!, dst: sgd108, runner_index: 28, sgd: 108, subgraph_id: QmWZnioK8S9HL45kQqKFd67NxeqM1uEoVhYrrbcPDxuQSV, component: SubgraphInstanceManager
[...]

PR #6228, opened 2025-12-16
Bugfix: Ensure can_copy_from failures fully roll back deployment creation

dimitrovmaksim:
Resolves https://github.com/graphprotocol/graph-node/issues/6220

Currently, the `can_copy_from` validation is only performed when a deployment is considered **new**—that is, when no existing `deployment_schema` record exists for the given Qm hash.

In the code [here](https://github.com/graphprotocol/graph-node/blob/792a3fe5a7f2b461731efda7ed0d89215f10f27f/store/postgres/src/subgraph_store.rs#L366), the layout is assigned if the deployment is new. If a `deployment_schema` record already exists, the code assumes the deployment was previously created and copied successfully, and therefore sets `graft_base_layout` to `None`.

However, if `can_copy_from` later fails (see [here](https://github.com/graphprotocol/graph-node/blob/master/store/postgres/src/deployment_store.rs#L230)), the transaction rolls back all changes **except** the record in the `deployment_schemas` table. This leaves the deployment in a partially created state and copy was never attempted

When the same deployment is attempted again, it is no longer treated as new. As a result, `graft_base_layout` is set to `None`, and the `can_copy_from` check is [skipped](https://github.com/graphprotocol/graph-node/blob/master/store/postgres/src/deployment_store.rs#L229). The deployment then proceeds as if it were valid, the copy process starts, but it fails again—ultimately leaving the deployment stuck in a **"never synced"** state.

### What this PR changes:
Moves the `can_copy_from` check right after site allocation. Adds an aditional check to detect orphaned `Sites` (deployment exists in `deployment_schemas` table but the actual deployment is not present in the shard, which indicates the deployment has failed somewhere between Site allocation and deployment creation) and re-run the graft `can_copy_from` check on redeploys. This ensures failed deployments that left orphaned sites get validated again.

Remove the duplicate compatibility check from deployment_store.rs

Minor cleanups (string formatting improvements, unused variables removed).

Removed the `can_copy_from` check from `copy_deployment` because it passed the source layout as the graft base and then compared it against the destination layout, which seems as a redundant check since the two layouts are the same entity.

### DIFF
--- a/store/postgres/src/subgraph_store.rs
+++ b/store/postgres/src/subgraph_store.rs
@@ -365,23 +365,21 @@ impl SubgraphStore {
                 .get(&site.shard)
                 .ok_or_else(|| StoreError::UnknownShard(site.shard.to_string()))?;
 
-            let mut shard_conn = deployment_store.get_replica_conn(ReplicaId::Main).await?;
-            let needs_check = if site_was_created {
-                true
-            } else {
-                // If deployment does not exist, but site exists it means
-                // that we are recovering from a failed deployment creation with an orphaned site.
-                // In that case, we should check graft compatibility again.
-                let exists = crate::deployment::exists(&mut shard_conn, &site).await?;
-                !exists
-            };
-
             if let Some(graft_base) = graft_base {
-                let base_layout = self.layout(graft_base).await?;
+                // Perform schema compatibility validation when:
+                // - the site was just created (first-time deployment), or
+                // - the site exists but the deployment is missing (recovering from a failed create).
+                // The `exists` DB call is only made if `site_was_created` is false.
+                let should_validate = {
+                    let mut shard_conn = deployment_store.get_replica_conn(ReplicaId::Main).await?;
+                    site_was_created || !crate::deployment::exists(&mut shard_conn, &site).await?
+                };
 
-                if needs_check {
+                if should_validate {
+                    let base_layout = self.layout(graft_base).await?;
                     let entities_with_causality_region =
                         deployment.manifest.entities_with_causality_region.clone();
+                    let mut shard_conn = deployment_store.get_replica_conn(ReplicaId::Main).await?;
                     let catalog = Catalog::for_creation(
                         &mut shard_conn,
                         site.cheap_clone(),


store/postgres/src/catalog.rs
```diff
@@ -1110,7 +1110,7 @@ pub(crate) async fn histogram_bounds(
     table: &SqlName,
     column: &str,
 ) -> Result<Vec<i64>, StoreError> {
-    const QUERY: &str = "select histogram_bounds::text::int8[] bounds \
+    const QUERY: &str = "select coalesce(histogram_bounds::text::int8[], '{}'::int8[]) as bounds \
```
lutter, line 1113: woops .. nice!

store/postgres/src/subgraph_store.rs
```diff
@@ -351,33 +352,60 @@ impl SubgraphStore {
...
+                Some(base) => Some(self.layout(base).await?),
                 None => None,
             };
 
-            if let Some(graft_base) = &graft_base {
+            if let Some(graft_base_layout) = &graft_base_layout {
                 self.primary_conn()
                     .await?
-                    .record_active_copy(graft_base.site.as_ref(), site.as_ref())
+                    .record_active_copy(graft_base_layout.site.as_ref(), site.as_ref())
```
lutter: Shouldn't this also happen in the transaction above? Otherwise, can't we end up in a situation where we set up everything except recording the active copy if `graph-node` gets killed atthe wrong moment?

store/postgres/src/subgraph_store.rs
```diff
@@ -351,33 +352,60 @@ impl SubgraphStore {
...
+                        .await?;
+                    let node_id = conn.assigned_node(&site).await?.unwrap_or(node_id);
+                    let site = Arc::new(site);
+
+                    if let Some(graft_base) = graft_base {
+                        // Ensure that the graft base exists
+                        let base_layout = self.layout(graft_base).await?;
+                        let entities_with_causality_region =
+                            deployment.manifest.entities_with_causality_region.clone();
+                        let catalog = Catalog::for_tests(
```
lutter: This is wrong - `for_tests` doesn't actually check the database for its capabilities (and the method should be marked as `#[cfg(debug_assertions)]`). Instead, this needs to use `Catalog::for_creation` where the connection is a connection to the `shard` in which we will create the subgraph. It's best to create the `catalog` object outside of this transaction so that we don't hold multiple db connections at once.

dimitrovmaksim: > for_tests doesn't actually check the database for its capabilities

For this specific case the catalog is needed just to create the layout object, so the `can_copy_from` helpers can be used to compare the `src` and `dst` schemas. Later in the code, when creating the deployment relational schema, we construct the catalog with the proper checks. Although I agree using a `for_tests` function does not look good.

> Instead, this needs to use Catalog::for_creation where the connection is a connection to the shard

At this point the only option (i see) to get a shard conn is using `deployment_store.get_replica_conn(ReplicaId::Main)` because the `pool` field is private. 

> It's best to create the catalog object outside of this transaction so that we don't hold multiple db connections at once.

Catalog creation requires to have the site already created, which happens in the transaction, and the point of adding the transaction is to revert the creation of the site if the `can_copy_from` fails (which is the cause of the original issue) so I'm not sure if holding a primary and a shard conn can be avoided. Unless, instead of using a transaction, we don't just execute `drop_site` if the `can_copy_from` check fails. But I may be missing something.

I wanted to make this fix with minimal changes to the current workflow, but maybe I should rethink the whole workflow instead of just patching in.

store/postgres/src/subgraph_store.rs
dimitrovmaksim: Good catch, I was only focusing on the incompatible schemas case

store/postgres/src/subgraph_store.rs
```diff
@@ -354,57 +354,63 @@ impl SubgraphStore {
...
+            let mut shard_conn = deployment_store.get_replica_conn(ReplicaId::Main).await?;
+            let needs_check = if site_was_created {
+                true
+            } else {
+                // If deployment does not exist, but site exists it means
+                // that we are recovering from a failed deployment creation with an orphaned site.
+                // In that case, we should check graft compatibility again.
+                let exists = crate::deployment::exists(&mut shard_conn, &site).await?;
+                !exists
+            };
```
lutter: Simpler:
```rust
    // If !site_was_created but deployment::exists, we are recovering from a failed
    // deployment creation with an orphaned site. In that case, check graft compatibility again
   let needs_check = site_was_created || crate::deployment::exists(&mut shard_conn, &site).await?;
```

lutter: This is getting very nit-picky, but we could further reduce the time we hold `shard_conn` by doing something like
```rust
    let needs_check = {
        let mut shard_conn = deployment_store.get_replica_conn(ReplicaId::Main).await?;
        site_was_created || crate::deployment::exists(&mut shard_conn, &site).await?;
    }
```

and something similar around the `Catalog::for_creation` call, but probably not a huge improvement.

dimitrovmaksim: I initially did shorten it like that, but decided it may be too ambiguous and confusing. 
-Edit-
I addressed the comments and moved the code inside the `if let` block to perform the validation related checks only if there's a graft_base

PR #6240, opened 2026-01-01
Fix compatibility issues between rust-web3 and alloy, Fallback to firehose or RPC on deserialization errors

incrypto32:
This PR addresses a few issues.
1) Workarounds for backward compatibility - ABI's which had non standard fields were accepted by rust-web3 but alloy is more strict with this so subgraphs that has ABI's with these compatibility issues failed. 
2) Fallback to use RPC or firehose to fetch blocks when it fails to deseriliaze from cache
3) Few other bug fixes

### DIFF
--- a/chain/ethereum/src/chain.rs
+++ b/chain/ethereum/src/chain.rs
@@ -355,6 +355,38 @@ impl std::fmt::Debug for Chain {
     }
 }
 
+/// Walk back from a block pointer by following parent pointers.
+/// This is the core logic used as a fallback when the cache doesn't have ancestor block.
+///
+async fn walk_back_ancestor<F, Fut, E>(
+    start_ptr: BlockPtr,
+    offset: BlockNumber,
+    root: Option<BlockHash>,
+    mut parent_getter: F,
+) -> Result<Option<BlockPtr>, E>
+where
+    F: FnMut(BlockPtr) -> Fut,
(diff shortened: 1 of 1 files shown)


graph/src/data_source/common.rs
```diff
@@ -24,6 +24,43 @@ use slog::Logger;
 use std::collections::HashMap;
 use std::{str::FromStr, sync::Arc};
 
+fn normalize_abi_json(json_bytes: &[u8]) -> Result<Vec<u8>, anyhow::Error> {
```
dwerner, line 48: Because this is kinda magic (bytes to bytes), maybe just an internal doc comment regarding the bug would be good to add here?

graph/src/data_source/common.rs
```diff
@@ -364,11 +401,14 @@ impl UnresolvedMappingABI {
                     self.name, self.file.link
                 )
             })?;
-        let contract = serde_json::from_slice(&*contract_bytes)
+        let normalized_bytes = normalize_abi_json(&contract_bytes)
```
dwerner, line 443: Similar to above, just suggesting we call out the normalization as a cover for the abi disparity.

dwerner (approved):
lgtm! (couple comments, just around calling out the patching/normalization)

incrypto32:
@dwerner i just added two more normalizations and added comments.

dwerner (approved):

PR #6262, opened 2026-01-08
Add fallback for ancestor_block to handle cache truncation

incrypto32:
Subgraphs would stall when block cache is truncated because ancestor_block returned None when the requested ancestor wasn't in cache.
 
 This PR adds RPC/Firehose fallback for ancestor_block logic

### DIFF
--- a/chain/ethereum/src/chain.rs
+++ b/chain/ethereum/src/chain.rs
@@ -1153,14 +1153,10 @@ impl TriggersAdapterTrait<Chain> for TriggersAdapter {
             }
             ChainClient::Rpc(adapters) => {
                 match self
-                    .fetch_light_block_with_rpc(adapters, &block_ptr)
+                    .fetch_full_block_with_rpc(adapters, &block_ptr)
                     .await?
                 {
-                    Some(light_block) => {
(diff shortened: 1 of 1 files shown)


incrypto32:
@lutter Integration cluster caught an issue i have updated the PR with the fix, i was using a Light Block for the rpc fall back for ancestor_block and was keeping an empty array for transaction_receipts, this was not right. So i've added a commit that uses the full block instead.

chain/ethereum/src/chain.rs
```diff
@@ -1040,56 +1086,83 @@ impl TriggersAdapterTrait<Chain> for TriggersAdapter {
...
+                            ptr_for_log.hash_hex(),
+                            e
+                        );
+                        ptr
+                    }
+                }
+            }
+            None => {
+                // Cache miss - fall back to walking the chain via parent_ptr() calls.
+                // This provides resilience when the block cache is empty (e.g., after truncation).
```
lutter, line 1122: The one downside with this is that if `ancestor_block` is called with an invalid hash, we run both a db query and make an rpc call. But that's probably not very common

PR #6264, opened 2026-01-09
Fix all clippy warnings

lutter:
This PR fixes all clippy warnings and requires that 'cargo clippy' passes in CI.

The commits with title "Run 'cargo clippy --fix'" were generated by running just that; since clippy only applies safe mechanical transforms with `--fix`, they should not require much review. Most of the other commits are very small and address one specific lint.

The first half of the PR (up to the first change to `justfile`) only changes actual code since clippy is run against the default target (binary and library)

The second half only changes test code by now running `cargo clippy --all-targets`

### DIFF
--- a/justfile
+++ b/justfile
@@ -8,7 +8,7 @@ format *EXTRA_FLAGS:
 
 # Run Clippy linting (cargo clippy)
 lint *EXTRA_FLAGS:
-    cargo clippy {{EXTRA_FLAGS}}
+    cargo clippy --all-targets {{EXTRA_FLAGS}}
 
 # Check Rust code (cargo check)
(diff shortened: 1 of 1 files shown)


graph/src/components/metrics/stopwatch.rs
```diff
@@ -229,6 +229,7 @@ impl StopwatchInner {
                 .write(true)
                 .append(false)
                 .create(true)
+                .truncate(true)
```
isum, line 232: This does not seem to be a lint fix. Why is this necessary?

lutter: It was .. there is some lint that complains when this line isn't there

PR #6278, opened 2026-01-15
*: Expose subgraph logs via subgraph GraphQL 

fordN:
This PR introduces a subgraph log storage and querying system for Graph Node. Subgraph logs can be queried through the GraphQL API via a new `_logs` field. The implementation supports multiple storage backends (File, Elasticsearch, Loki) with a consistent query interface.

## What's new

**GraphQL Query API**
- New `_logs` query field on all subgraph deployments
- Filter by log level, timestamp range, and text search
- Structured log entries with metadata (source location, arguments)
- Pagination via `first`/`skip` and sort order via `orderDirection`

**Storage Backends**
- **File**: JSON Lines format, one file per subgraph (good for development)
- **Elasticsearch**: Enterprise search and analytics (production)
- **Loki**: Grafana log aggregation, with basic auth support (production)
- **Disabled**: Default when no `[log_store]` section is configured

**Configuration via `graph-node.toml`**

Log store is configured through a `[log_store]` section in the TOML config file, following the same pattern as `[store]`, `[chains]`, and `[deployment]`.

**Architecture**
- `LogDrain`: Write-side sink for each backend (File, Loki, Elasticsearch)
- `LogStore`: Read-side query interface for each backend
- `LoggerFactory`: Refactored for multi-backend log routing
- Uses `slog::Level` directly (no custom LogLevel enum)
- Parse failures are logged as warnings, not silently dropped

### Querying logs

```graphql
{
  _logs(
    level: ERROR
    search: "timeout"
    from: "2024-01-15T00:00:00Z"
    to: "2024-01-16T00:00:00Z"
    first: 100
    orderDirection: desc
  ) {
    id
    timestamp
    level
    text
    arguments { key value }
    meta { module line column }
  }
}
```

### Configuring log store backends

File-based (development):
```toml
[log_store]
backend = "file"
directory = "/var/log/graph-node/subgraph-logs"
retention_hours = 72
```

Loki (production):
```toml
[log_store]
backend = "loki"
url = "https://loki.example.com"
tenant_id = "my-tenant"
username = "user"
password = "secret"
```

Elasticsearch (production):
```toml
[log_store]
backend = "elasticsearch"
url = "http://localhost:9200"
username = "elastic"
password = "changeme"
index = "subgraph"
timeout_secs = 10
```


lutter:
One thing I wonder about: should this be configured via environment variables or through `graph-node.toml`? I would lean towards the latter since it's a configuration that is unlikely to change often, and having it in the config file would let's us express more complicated configuration (though admittedly, right now the config is not very complex)

graph/src/components/log_store/config.rs
```diff
@@ -0,0 +1,216 @@
...
+///
+/// # Arguments
+/// * `logger` - Logger for emitting deprecation warnings
+/// * `new_key` - The new environment variable name
+/// * `old_key` - The deprecated environment variable name
+/// * `default` - Default value to return if parsing fails or neither key is set
+///
+/// # Returns
+/// The parsed u64 value, or the default if parsing fails or neither key is set
+pub fn read_u64_with_fallback(logger: &Logger, new_key: &str, old_key: &str, default: u64) -> u64 {
```
dwerner: Nit: could lift the generic style of `str::parse` and avoid needing multiple functions, get a specialization of `FromStr` implementer for free.

i.e.
```
    pub fn parse<F: FromStr>(&self) -> Result<F, F::Err> {
        FromStr::from_str(self)
    }
```

graph/src/components/log_store/elasticsearch.rs
```diff
@@ -0,0 +1,217 @@
...
+        &self,
+        query_body: serde_json::Value,
+    ) -> Result<Vec<LogEntry>, LogStoreError> {
+        let url = format!("{}/{}/_search", self.endpoint, self.index);
+
+        let mut request = self
+            .client
+            .post(&url)
+            .json(&query_body)
+            .timeout(self.timeout);
```
dwerner, line 105: :+1: for timeout.

graph/src/components/log_store/file.rs
```diff
@@ -0,0 +1,363 @@
...
+use std::io::{BufRead, BufReader};
+use std::path::PathBuf;
+
+use crate::prelude::DeploymentHash;
+
+use super::{LogEntry, LogMeta, LogQuery, LogStore, LogStoreError};
+
+pub struct FileLogStore {
+    directory: PathBuf,
+    // TODO: Implement log rotation when file exceeds max_file_size
```
dwerner: Todos left, not sure if intended to be implemented this pass.

graph/src/components/log_store/file.rs
```diff
@@ -0,0 +1,363 @@
...
+        })
+    }
+
+    /// Get log file path for a subgraph
+    fn log_file_path(&self, subgraph_id: &DeploymentHash) -> PathBuf {
+        self.directory.join(format!("{}.jsonl", subgraph_id))
+    }
+
+    /// Parse a JSON line into a LogEntry
+    fn parse_line(&self, line: &str) -> Option<LogEntry> {
```
dwerner, line 58: Design: idiomatically, parse methods have a `Result<T, SomeErrorType>` and I would expect that here but I get the practical. Worth considering a custom error type for log store? Particularly because we just ignore the line entirely it's effectively eaten if there's a parse error.

Edit: further reading later suggests there is a `LogStoreError` enum already. Use here?

graph/src/components/log_store/loki.rs
```diff
@@ -0,0 +1,283 @@
...
+                    .values
+                    .into_iter()
+                    .filter_map(move |value| self.parse_log_entry(value, &stream_labels))
+            })
+            .collect();
+
+        Ok(entries)
+    }
+
+    fn parse_log_entry(
```
dwerner, line 143: See previous comment - error enum useful here?

graph/src/components/log_store/mod.rs
```diff
@@ -0,0 +1,333 @@
...
+
+use async_trait::async_trait;
+use std::path::PathBuf;
+use std::str::FromStr;
+use std::sync::Arc;
+use thiserror::Error;
+
+use crate::prelude::DeploymentHash;
+
+#[derive(Error, Debug)]
```
dwerner, line 15: Continuing my suggestion/question: maybe a child error enum that captures errors on retrieval/parsing?

graph/src/components/log_store/mod.rs
```diff
@@ -0,0 +1,333 @@
...
+
+    /// File-based logs (JSON lines format)
+    File {
+        directory: PathBuf,
+        max_file_size: u64,
+        retention_days: u32,
+    },
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
```
dwerner, line 60: Suggestion: Is there a need for a custom LogLevel impl? I feel like you'd get all of this for free if you used `tracing::Level` including the `FromStr` impls.

graph/src/log/common.rs
```diff
@@ -0,0 +1,237 @@
...
+#[derive(Clone, Debug, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct LogMeta {
+    pub module: String,
+    pub line: i64,
+    pub column: i64,
+}
+
+/// Converts an slog Level to a string representation
+pub fn level_to_str(level: Level) -> &'static str {
```
dwerner: Related to last comment on custom log level type.

graph/src/log/elastic.rs
```diff
@@ -33,28 +32,14 @@ pub struct ElasticLoggingConfig {
 }
 
 /// Serializes an slog log level using a serde Serializer.
-fn serialize_log_level<S>(level: &Level, serializer: S) -> Result<S::Ok, S::Error>
+fn serialize_log_level<S>(level: &str, serializer: S) -> Result<S::Ok, S::Error>
```
dwerner: Was this previously `Level` coming in from the `slog::*` log import? Related to the log level stuff above.

dwerner (commented):
Lgtm! I have a few questions/suggestions, but yolo

graph/src/components/log_store/file.rs
fordN: Addressed these TODOs now.

DaMandal0rian:
Great work on this! We're planning to use the Loki drain in production as part of our ES decommission.

One blocker: our production Loki endpoint (`loki-logs.thegraph.com`) requires HTTP basic auth. The current `LokiDrain` implementation only supports `GRAPH_LOG_STORE_LOKI_URL` and `GRAPH_LOG_STORE_LOKI_TENANT_ID` — no credentials.

Could we add support for basic auth? Something like:
```
GRAPH_LOG_STORE_LOKI_USER=<username>
GRAPH_LOG_STORE_LOKI_PASSWORD=<password>
```

This would allow the reqwest client to attach an `Authorization: Basic ...` header on push requests. Without this, we'd need to deploy a sidecar proxy or an unauthenticated push endpoint, which adds operational complexity.

graph/src/components/log_store/config.rs
fordN: Good call! I ended up just removing this code altogether though when I moved to file based configuration.

graph/src/components/log_store/file.rs
fordN: Good point about losing visibility. I've added `warn!` logging to all three backends (file, loki, elasticsearch). 

Parse failures now emit warnings with the specific error before skipping the entry. The approach is still resilient (malformed entries are skipped, not query-fatal), but failures are no longer silent.

graph/src/components/log_store/mod.rs
fordN: Same as above. Went with `warn!` logging for log parse failures.

graph/src/components/log_store/mod.rs
fordN: 👍 Removed the custom `LogLevel` enum and now just use `slog::Level` directly throughout.

graph/src/log/elastic.rs
fordN: Yep! Restored this back to `slog::Level` along with the other `Level` changes.

fordN:
> Great work on this! We're planning to use the Loki drain in production as part of our ES decommission.
> 
> One blocker: our production Loki endpoint (`loki-logs.thegraph.com`) requires HTTP basic auth. The current `LokiDrain` implementation only supports `GRAPH_LOG_STORE_LOKI_URL` and `GRAPH_LOG_STORE_LOKI_TENANT_ID` — no credentials.
> 
> Could we add support for basic auth? Something like:
> 
> ```
> GRAPH_LOG_STORE_LOKI_USER=<username>
> GRAPH_LOG_STORE_LOKI_PASSWORD=<password>
> ```
> 
> This would allow the reqwest client to attach an `Authorization: Basic ...` header on push requests. Without this, we'd need to deploy a sidecar proxy or an unauthenticated push endpoint, which adds operational complexity.

Added! ✔️

fordN:
> One thing I wonder about: should this be configured via environment variables or through `graph-node.toml`? I would lean towards the latter since it's a configuration that is unlikely to change often, and having it in the config file would let's us express more complicated configuration (though admittedly, right now the config is not very complex)

Yeah, might as well keep configs in the toml file, easier to manage. I've made the change to that effect in commit `64fa5278`.

dwerner (approved):
looking good

PR #6282, opened 2026-01-19
Rewrite graph-cli in Rust

lutter:
This is all Claude-generated code, I haven't even looked at it or tried it. It's mostly here for informational purposes

### DIFF
--- a/gnd/Cargo.toml
+++ b/gnd/Cargo.toml
@@ -64,9 +64,6 @@ semver = { workspace = true }
 sha1 = "0.10"
 wasmparser = { workspace = true }
 
-# Test command
-which = "7"
-
 # Interactive prompts
 inquire = "0.7"
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3167,12 +3167,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "env_home"
-version = "0.1.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c7f84e12ccf0a7ddc17a6c41c93326024c42920d7ee630d04950e6926645c0fe"
-
-[[package]]
 name = "env_logger"
 version = "0.11.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3672,7 +3666,6 @@ dependencies = [
  "url",
  "walkdir",
  "wasmparser 0.118.2",
- "which 7.0.3",
 ]
 
 [[package]]
@@ -6209,7 +6202,7 @@ dependencies = [
  "protobuf-support",
  "tempfile",
  "thiserror 1.0.61",
- "which 4.4.2",
+ "which",
 ]
 
 [[package]]
@@ -9076,18 +9069,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "which"
-version = "7.0.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "24d643ce3fd3e5b54854602a080f34fb10ab75e0b813ee32d00ca2b44fa74762"
-dependencies = [
- "either",
- "env_home",
- "rustix 1.0.7",
- "winsafe",
-]
-
-[[package]]
 name = "whoami"
 version = "1.5.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -9462,12 +9443,6 @@ checksum = "21a0236b59786fed61e2a80582dd500fe61f18b5dca67a4a067d0bc9039339cf"
 dependencies = [
  "memchr",
 ]
-
-[[package]]
-name = "winsafe"
-version = "0.0.19"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d135d17ab770252ad95e9a872d365cf3090e3be864a34ab46f48555993efc904"
 
 [[package]]
 name = "wiremock"
--- a/gnd/src/commands/test.rs
+++ b/gnd/src/commands/test.rs
@@ -56,6 +56,14 @@ pub fn run_test(opt: TestOpt) -> Result<()> {
     }
 }
 
+/// Check if a binary with the given name exists in any PATH directory.
+fn is_in_path(name: &str) -> bool {
+    let Some(path_var) = std::env::var_os("PATH") else {
+        return false;
+    };
+    std::env::split_paths(&path_var).any(|dir| dir.join(name).is_file())
+}
+
 /// Find the Matchstick binary.
 fn find_matchstick() -> Result<PathBuf> {
     // First, check node_modules/.bin/graph-test (graph-cli's matchstick wrapper)
@@ -71,7 +79,7 @@ fn find_matchstick() -> Result<PathBuf> {
     }
 
     // Check if matchstick is in PATH
-    if which::which("matchstick").is_ok() {
+    if is_in_path("matchstick") {
         return Ok(PathBuf::from("matchstick"));
     }
 


dimitrovmaksim:
Couple of things I noticed for the commands that i'm familiar with:
1. `test`
    - It expects matchstick to be a `node_module` installed via `matchstick-as` (the equivalent of graph-ts for matchstick). In the ts graph-cli we used a lib to download and execute binaries, so it probably got confused by that.
    EDIT: I now see that the binaryinstall lib was removed and the matchstick bin was directly fetched from the repo into `node_modules/.bin/`, but I don't see the `test` command doing it here
    - I would probably remove the binary path and leave only the Docker path, or at least make it the default (or phase out the test command completelly) 
2. `add`
   - By default the `add` command indexes events as entities. This caused some issues with duplicated entities and abi overwrites. If there were collision in event/entity names with existing events/entity names it would prepend the new entity name with the contract name. It seems that Claude has cut that whole logic off and just ignores collsions, unless the `--merge-entities` flag is passed. IMO this descision back then caused more trouble than benefit, so I would remove the code that touches the `schema`  with the `add` command and leave this to the developer. Anyways it seems rare that entites match events 1:1.
   - the `to_kebab_case` function seems to not handle some cases correctly https://github.com/graphprotocol/graph-node/blob/5c086e691a9993737a0b91905a6b2ce4c0d77a20/gnd/src/commands/add.rs#L658 (i would expect the result in this case to be either `erc-20-token` or `erc20-token` in the best case scenario). The `TS` cli uses the gluegun/loadash implementation for `kebabCase`

dimitrovmaksim (commented):
Differences and regressions I noticed compared to the `graph-cli` when running `init` interactively with the Smart Contract source

1. Step ordering
Some prompts are ordered differently (for example, subgraph slug and directory).
I actually think the `gnd` order makes more sense here, since it may allow suggestions based on the fetched contract information.

2. Contract info fetching (regression)
In `graph-cli`, contract information is fetched immediately after the contract address is entered. The CLI then suggests the fetched contract name, start block, and would prompt for ABI path only if fetching fails.
In `gnd`, these values are prompted before fetching the contract info, and later, it appears that the fetched data is not used and the defaults or user input are kept instead.
(I’ve added code snippets below showing the step order in both CLIs.)

3. Adding additional contracts
`graph-cli` prompts the user at the end of `init` to `add` another contract, while gnd does not.

4. Directory name collisions
If the target directory already exists, `graph-cli` asks whether it should be overwritten.
`gnd` exits with an error instead. While this may not strictly be a regression, I think it would make more sense to prompt the user to choose a different directory name to avoid the collision.

5. Obsolete `spkg` flag
Not a regression, but gnd still exposes the spkg flag, which appears to be obsolete now that SPS is deprecated and the Substreams option has been removed.

graph-cli:
```bash
$ graph init
✔ Network · Arbitrum One Mainnet · arbitrum-one · https://arbiscan.io
✔ Source · Smart Contract · ethereum
✔ Subgraph slug · graph-token
✔ Directory to create the subgraph in · graph-token
✔ Contract address · 0x9623063377AD1B27544C965cCd7342f7EA7e88C7
✔ Fetching ABI from Sourcify API...
✔ Fetching ABI from Contract API...
✔ Fetching start block from Contract API...
✔ Fetching contract name from Contract API...
✔ Start block · 42449274
✔ Contract name · GraphProxy
✔ Index contract events as entities (Y/n) · false
  Generate subgraph
  Write subgraph to directory
✔ Create subgraph scaffold
✔ Initialize networks config
✔ Initialize subgraph repository
✔ Install dependencies with yarn
✔ Generate ABI and schema types with yarn codegen
✔ Add another contract? (y/N) · false

Subgraph graph-token created in graph-token
```

gnd:
```bash
$ gnd init
Creating a new subgraph...

> Source: Smart contract
> Network: Arbitrum One Mainnet (arbitrum-one)
> Contract address: 0x9623063377AD1B27544C965cCd7342f7EA7e88C7
> Contract name: Contract
> Subgraph slug: contract
> Directory: contract
> Do you have an ABI file? No
> Start block: 0
> Index contract events as entities? No
→ Fetching contract info from 0x9623063377AD1B27544C965cCd7342f7EA7e88C7 on arbitrum-one
✔ Found contract: GraphProxy
→ Generating scaffold files
→ Initializing Git repository
✔ Install dependencies with pnpm
✔ Subgraph created at contract

Next steps:
  cd contract
  gnd codegen
  gnd build
```

I’ll follow up with additional comments if I find any more differences or regressions.

gnd/src/commands/init.rs
dimitrovmaksim, line 1: Differences and regressions I noticed compared to the `graph-cli` when running `init` interactively with the Smart Contract source

1. Step ordering
Some prompts are ordered differently (for example, subgraph slug and directory).
I actually think the `gnd` order makes more sense here, since it may allow suggestions based on the fetched contract information.

2. Contract info fetching (regression)
In `graph-cli`, contract information is fetched immediately after the contract address is entered. The CLI then suggests the fetched contract name, start block, and would prompt for ABI path only if fetching fails.
In `gnd`, these values are prompted before fetching the contract info, and later, it appears that the fetched data is not used and the defaults or user input are kept instead.
(I’ve added code snippets below showing the step order in both CLIs.)

3. Adding additional contracts
`graph-cli` prompts the user at the end of `init` to `add` another contract, while gnd does not.

graph-cli:
```bash
$ graph init
✔ Network · Arbitrum One Mainnet · arbitrum-one · https://arbiscan.io
✔ Source · Smart Contract · ethereum
✔ Subgraph slug · graph-token
✔ Directory to create the subgraph in · graph-token
✔ Contract address · 0x9623063377AD1B27544C965cCd7342f7EA7e88C7
✔ Fetching ABI from Sourcify API...
✔ Fetching ABI from Contract API...
✔ Fetching start block from Contract API...
✔ Fetching contract name from Contract API...
✔ Start block · 42449274
✔ Contract name · GraphProxy
✔ Index contract events as entities (Y/n) · false
  Generate subgraph
  Write subgraph to directory
✔ Create subgraph scaffold
✔ Initialize networks config
✔ Initialize subgraph repository
✔ Install dependencies with yarn
✔ Generate ABI and schema types with yarn codegen
✔ Add another contract? (y/N) · false

Subgraph graph-token created in graph-token
```

gnd:
```bash
$ gnd init
Creating a new subgraph...

> Source: Smart contract
> Network: Arbitrum One Mainnet (arbitrum-one)
> Contract address: 0x9623063377AD1B27544C965cCd7342f7EA7e88C7
> Contract name: Contract
> Subgraph slug: contract
> Directory: contract
> Do you have an ABI file? No
> Start block: 0
> Index contract events as entities? No
→ Fetching contract info from 0x9623063377AD1B27544C965cCd7342f7EA7e88C7 on arbitrum-one
✔ Found contract: GraphProxy
→ Generating scaffold files
→ Initializing Git repository
✔ Install dependencies with pnpm
✔ Subgraph created at contract

Next steps:
  cd contract
  gnd codegen
  gnd build
```

4. Directory name collisions
If the target directory already exists, `graph-cli` asks whether it should be overwritten.
`gnd` exits with an error instead. While this may not strictly be a regression, I think it would make more sense to prompt the user to choose a different directory name to avoid the collision.

5. Obsolete `spkg` flag
Not a regression, but gnd still exposes the spkg flag, which appears to be obsolete now that SPS is deprecated and the Substreams option has been removed.

6. `gnd` does not create the `networks.json` file with the chain info, that is used for deploying the subgraph to multiple networks

```json
{
  "arbitrum-one": {
    "GraphProxy": {
      "address": "0x9623063377AD1B27544C965cCd7342f7EA7e88C7",
      "startBlock": 42449274
    }
  }
}
```

I’ll follow up with additional comments if I find any more differences or regressions.

gnd/src/commands/add.rs
dimitrovmaksim, line 1: For the `add` command (in addition to what is mentioned here https://github.com/graphprotocol/graph-node/pull/6282#issuecomment-3777597396)

1. In `gnd` the `add` command accepts a `--network` flag, which in the case of `add` does not make sense, as subgraphs do not support multichain datasources, so the chain should always be derived from the manifest file. I guess Claude got confused here and added `--network` instead of `--network-file`, which should be the path to the networks config file (if a custom one is used).
2. Similar to `init`, `add` does not update the `networks` file and this functionality has been ignored completely.

Other than that seems to be working fine.

gnd/src/codegen/schema.rs
```diff
@@ -0,0 +1,1148 @@
...
+            // For Int8 and Bytes, check if id is null/unset and use "auto" as the key
+            Method::new(
+                "save",
+                vec![],
+                Some(NamedType::new("void").into()),
+                format!(
+                    r#"
+        let id = this.get('id')
+        if (id == null || id.kind == ValueKind.NULL) {{
+          store.set('{}', 'auto', this)
```
dimitrovmaksim, line 411: I'm not sure where this comes from. I'm not familiar with timeseries and aggregation entities, which seem to have auto incremented `Int8` ids (at least according to the docs)

`gnd codegen` will produce this for a regular entity from the schema. The comment before the constructor seems incorrect, at least for regular entities. Also I don't think regular entities with and id of type ID or Bytes, should be auto-incremented. Also (2) `if (id !== null)` <--- this won't work with AS < 0.20.0, before `===` and `==` were unified https://github.com/AssemblyScript/assemblyscript/releases/tag/v0.20.0
```ts
export class ExampleEntity extends Entity {
  /**
   * Leaving out the id argument uses an autoincrementing id.
   */
  constructor(id: Bytes | null = null) {
    super();
    if (id !== null) {
      this.set("id", Value.fromBytes(id));
    }
  }

  save(): void {
    let id = this.get("id");
    if (id == null || id.kind == ValueKind.NULL) {
      store.set("ExampleEntity", "auto", this);
    } else {
      assert(
        id.kind == ValueKind.BYTES,
        `Entities of type ExampleEntity must have an ID of type Bytes but the id '${id.displayData()}' is of type ${id.displayKind()}`,
      );
      store.set("ExampleEntity", id.toBytes().toHexString(), this);
    }
  }
```

Also this is what `gnd` generates for a `timeseries` Entity

```ts
export class Data extends Entity {
  /**
   * Leaving out the id argument uses an autoincrementing id.
   */
  constructor(id: i64 = i64.MIN_VALUE) {
    super();
    if (id != i64.MIN_VALUE) {
      this.set("id", Value.fromI64(id));
    }
  }

  save(): void {
    let id = this.get("id");
    if (id == null || id.kind == ValueKind.NULL) {
      store.set("Data", "auto", this);
    } else {
      assert(
        id.kind == ValueKind.INT8,
        `Entities of type Data must have an ID of type Int8 but the id '${id.displayData()}' is of type ${id.displayKind()}`,
      );
      store.set("Data", id.toI64().toString(), this);
    }
  }
...
}
```

and this is from `graph-cli v0.98.1`:
```ts
export class Data extends Entity {
  constructor(id: Int8) {
    super();
    this.set("id", Value.fromI64(id));
  }

  save(): void {
    let id = this.get("id");
    assert(id != null, "Cannot save Data entity without an ID");
    if (id) {
      assert(
        id.kind == ValueKind.INT8,
        `Entities of type Data must have an ID of type Int8 but the id '${id.displayData()}' is of type ${id.displayKind()}`,
      );
      store.set("Data", id.toI64().toString(), this);
    }
  }
...
}
```
I'm a little bit confused here :D Both are specVersion: 1.3.0 and apiVersion: 0.0.9

gnd/src/commands/init.rs
dimitrovmaksim: Two more things i noticed for `init`:

1. `gnd` will add the contract address as a hex number in the manifest, instead of as a string

```yaml
# gnd
source:
    abi: Graph
    address: 0x9623063377AD1B27544C965cCd7342f7EA7e88C7

# graph-cli
    source:
      address: "0x9623063377AD1B27544C965cCd7342f7EA7e88C7"
```

2. `gnd`, when scaffolding the manifest, will add the mappings file without prependign `./src` so it will only contain the mappings file name. https://github.com/graphprotocol/graph-node/blob/lutter/gnd-cli/gnd/src/scaffold/manifest.rs#L9
The `add` command on the other hand, correctly prepends `src`

```yaml
# gnd
...
     file: ./graph.ts
...

# graph-cli
...
      file: ./src/graph-proxy.ts
...
```

So `graph-cli` won't be able to work with `gnd` generated subgraphs:
```bash
✖ Failed to load subgraph from subgraph.yaml: Error in subgraph.yaml:

  Path: dataSources > 0 > source > address
  Expected string, found number:
  8.571296818927177e+47

  Path: dataSources > 0 > mapping > file
  File does not exist: graph.ts

  Path: dataSources > 1 > source > address
  Expected string, found number:
  1.0021244402728634e+48
```

Also `gnd build` will fail, as it expects the path to contain `./src`
```bash
Manifest validation errors:
  - Mapping file for data source 'Graph' not found: ./graph.ts
 ```

File: gnd/src/codegen/schema.rs (hunk 1 above)
Comment (lutter) [reply gnd/src/codegen/schema.rs:411]: A long time ago, these generated constructors took strings; since PR #5029, `graph-node` accepts `"auto"` for these constructors if the type of the id is `Int8` or `Bytes` to mean "generate an id". This was meant as a general facility to keep people from generating insanely long ids when they don't really need them/don't care about them.

At some point, `graph-cli` was changed so that these constructors took `Int8` and `Bytes`, making it impossible to use `"auto"`. This updated codegen tries to make that possible again.

As for the `id !== null` comparison, that seems to work. I am attaching a little example that demonstrates that. When I run that all the asserts are successful. But I might be missing some big footgun

[asctest.zip](https://github.com/user-attachments/files/25028568/asctest.zip)

Comment (dimitrovmaksim) [reply gnd/src/codegen/schema.rs:411]: Yeah i'm not completelly sure about `==` vs `===` either, what the exact conditions/cases are. I remember people had issues when loading an enitity and comparing with `null` using `===`, or comparing values coming from an event against some pre-defined values. One case that I could reproduce is:

```ts
test("=== vs ==", () => {
    let helloWorld = "hello world";
    let hello = "hello";
    let hello_2 = helloWorld.slice(0, 5);

    log.info("Test ===", []);
    if (hello_2 === hello) {
        log.info("Strings are equal with ===: {}", [hello_2]);
    } else {
        log.info("Strings are not equal with ===: {}", [hello_2]);
    }

    log.info("Test ==", []);
    if (hello_2 == hello) {
        log.info("Strings are equal with ==: {}", [hello_2]);
    } else {
        log.info("Strings are not equal with ==: {}", [hello_2]);
    }
});
```

and the result is

```bash
Compiling...

💬 Compiling equality...

Igniting tests 🔥

equality
--------------------------------------------------
  √ === vs == - 0.029ms
      💬 Test ===
    💬 Strings are not equal with ===: hello
    💬 Test ==
    💬 Strings are equal with ==: hello
```

gnd/src/commands/init.rs
```diff
@@ -0,0 +1,1362 @@
...
+    }
+
+    let network = opt.network.as_deref().unwrap_or("mainnet");
+
+    step(
+        Step::Load,
+        &format!("Fetching contract info from {} on {}", address, network),
+    );
+
+    let contract_info = {
```
dimitrovmaksim, line 267: After the last changes, `init` interactive will fetch the contract info twice, first time from `prompt.rs` to populate the form/suggestions and the second time here.

lutter: Woops .. fixed that. As far as I can tell, all the issues you found are now fixed except for anything to do with `gnd test` - not quite sure what to do there. I did remove schema editing from `gnd add` though.

dimitrovmaksim: I'll test the changes tomorrow and see if I will find something else. About `test`, I may try again to update matchstick and make it work, although I'm not sure there's anyone that can do a release(I don't have permissions anymore).

lutter: This would be a much bigger task, but I wonder if we shouldn't integrate matchstick into gnd - not sure how to prioritize it, but I am thinking that `gnd init` should also put down some sample tests and have the facilities to run tests, especially since it now also handles running Postgres etc. Maybe leave `test` out for now and do this separately. Seeing how matchstick hasn't had a release in more than 2 years, I wonder how much use it actually gets.

dimitrovmaksim: > but I am thinking that gnd init should also put down some sample tests 

Yeah, graph-cli actually does this, but I did not mention it as I wasn't sure if it's worth adding

> This would be a much bigger task, but I wonder if we shouldn't integrate matchstick into gnd

I actually did that in my attempt to write graph-cli in Rust, and it was quite easy to have matchstick as a lib that can be executed directly from within the cli.

> Seeing how matchstick hasn't had a release in more than 2 years, I wonder how much use it actually gets.

By the lack of open issues/comments in discord about it being broken/outdated, I would assume not much

fubhy:
David is plowing through

lutter:
> David is plowing through

With huge help from @dimitrovmaksim who's been doing the hard work of finding discrepancies with graph-cli

gnd/src/commands/init.rs
dimitrovmaksim: Couple of things:
1. It seems the `add` command messes the addresses up when updating the manifest, removing the surrounding `""` for all datasources.

2. More of a nit: the `next steps` hint will be shown before prompting to add another contract/datasource and again after adding a new datasource, probably will do it on each itteration, I haven't tested with more than 2 datasources

Honestly the `gnd` flow feels way better than the current graph-cli flow. Can't see anything wrong with `build` and `codegen` atm, will have to test to deploy with `gnd` and see how this goes

dimitrovmaksim: Tested `init` from example. It seems it runs `codegen` through the `package.json` script.
```
$ ../graph-node/target/release/gnd init                                                   
Creating a new subgraph...

> Source: Example subgraph
> Subgraph slug: example
> Directory: example
✔ Cloning example subgraph
✔ Initialize networks config
✔ Update subgraph name and commands in package.json
✔ Initialize subgraph repository
✔ Install dependencies with pnpm
✔ Generate ABI and schema types with pnpm codegen

Subgraph example created in example

Next steps:

  1. Run `graph auth` to authenticate with your deploy key.

  2. Type `cd example` to enter the subgraph.

  3. Run `yarn deploy` to deploy the subgraph.

Make sure to visit the documentation on https://thegraph.com/docs/ for further information.
```
(for reference `gnd init` from contract does not run codegen at the end)

gnd/src/commands/build.rs
```diff
@@ -0,0 +1,878 @@
...
+    })?;
+
+    Ok(())
+}
+
+/// Copy an ABI file to the specified output directory.
+///
+/// This unified function handles both data source and template ABIs.
+/// The caller specifies the output directory.
+fn copy_abi_to_dir(abi_name: &str, abi_path: &Path, output_subdir: &Path) -> Result<()> {
```
dimitrovmaksim, line 707: When copying the ABI to the `build` folder, it will not check if the json also includes the abi metadata. For example the GravatarRegistry abi from the [example subgraph](https://github.com/graphprotocol/graph-tooling/blob/main/examples/ethereum-gravatar/artifacts/contracts/Gravity.sol/GravatarRegistry.json) includes other metadata
```json
{
  "_format": "hh-sol-artifact-1",
  "contractName": "GravatarRegistry",
  "sourceName": "contracts/Gravity.sol",
  "abi": [{..}]
  ...
}
```
which results in an error, as the json is expected to contain only the value of the `abi` field,
```
expected a valid JSON ABI sequence at line 1 column 0
```

gnd/src/commands/build.rs
```diff
@@ -0,0 +1,878 @@
...
+            abi_path.display(),
+            output_path.display()
+        )
+    })?;
+
+    Ok(())
+}
+
+/// Write the output manifest file.
+fn write_output_manifest(
```
dimitrovmaksim, line 734: Nit: It seems the manifest written to the `build` folder is slightly misformatted (not properly indented), other than that it seems it's working fine and the issue was actually the contact abi containing metadata, not the formatting of the copied manifest

```yaml
specVersion: 0.0.4
description: Gravatar for Ethereum
repository: https://github.com/graphprotocol/graph-toolings
schema:
  file: schema.graphql
dataSources:
- kind: ethereum/contract
  name: Gravity
  network: mainnet
  source:
    address: 0x2E645469f354BB4F5c8a05B3b30A929361cf77eC
    abi: Gravity
  mapping:
    kind: ethereum/events
    apiVersion: 0.0.6
    language: wasm/assemblyscript
    entities:
    - Gravatar
    abis:
    - name: Gravity
      file: Gravity/Gravity.json
    eventHandlers:
    - event: NewGravatar(uint256,address,string,string)
      handler: handleNewGravatar
    - event: UpdatedGravatar(uint256,address,string,string)
      handler: handleUpdatedGravatar
    file: Gravity/Gravity.wasm
```

lutter:
I think this addresses all the open issues. AFAICT, the indenting of the YAML is correct, even if it's different from `graph-cli`

gnd/src/codegen/schema.rs
```diff
@@ -0,0 +1,1095 @@
...
+    /// Get the value type string for a field.
+    ///
+    /// Returns the GraphQL-style value type string:
+    /// - Scalars: `String`, `Int`, `BigInt`, etc.
+    /// - Arrays: `[String]`, `[Int]`, etc.
+    /// - Nested arrays: `[[String]]`, `[[Int]]`, etc.
+    /// - Entity references are converted to `String` (their ID type)
+    fn value_type_from_field(&self, field: &FieldInfo) -> String {
+        let base = if self.entity_names.contains(&field.base_type) {
+            "String".to_string() // Entity references are stored as string IDs
```
dimitrovmaksim: Direct entity references default to String for the referenced entity id type. 
```graphql
type Balance @entity(immutable: false) {
  id: Bytes! # Account address + token address
  token: Token!
  account: Account!
```
`token` and `account` fields where `Token` and `Account` entities have `Bytes` for id, have their setters generated to expect `string` as argument, and their getters will return the id as string as well.
```ts
get token(): string {
    let value = this.get("token");
    if (!value || value.kind == ValueKind.NULL) {
      throw new Error("Cannot return null for a required field.");
    } else {
      return value.toString();
    }
  }

  set token(value: string) {
    this.set("token", Value.fromString(value));
  }

  get account(): string {
    let value = this.get("account");
    if (!value || value.kind == ValueKind.NULL) {
      throw new Error("Cannot return null for a required field.");
    } else {
      return value.toString();
    }
  }

  set account(value: string) {
    this.set("account", Value.fromString(value));
  }
```

gnd/src/scaffold/mod.rs
```diff
@@ -0,0 +1,279 @@
...
+        .subgraph_name
+        .split('/')
+        .next_back()
+        .unwrap_or(&options.subgraph_name);
+
+    serde_json::to_string_pretty(&serde_json::json!({
+        "name": basename,
+        "license": "UNLICENSED",
+        "scripts": {
+            "codegen": "graph codegen",
```
dimitrovmaksim, line 194: I wonder if the scripts should run `gnd` instead of `graph-cli`. Potentially also drop `graph-cli` from the dependecies, Usually people have it installed globally anyways (unless the subgraph is a cloned repo)

gnd/Cargo.toml
dimitrovmaksim, line 1: It may be a good idea to review all new dependencies

gnd/src/codegen/schema.rs
lutter: Fixed in ec21bfd

gnd/Cargo.toml
lutter: I did some of that, trying to use workspace packages as much as possible. I'd like to get rid of ethabi in favor of alloy, but that requires a rebase

dimitrovmaksim: Tested "codegen", "build", "create" and "deploy" commands with an existing subgraph and was able to deploy to a locally running node. So if you want, go ahead with the alloy change and rebase. I guess we can close the current comments. I'll retest after that and if there are regressions I'll review again.

lutter:
> I don't know which file to refer for this one. After some more testing, `codegen` and `build` do not validate the manifest file correctly. They did not catch:
> 
>     * missing `abi` property under `source`
> 
>     * missing `abis` property under `mapping`

Getting better validations in there turned out to be a bit more involved than I thought. On the plus side, the `gnd` validations are now based on the `graph-node` validations which should reduce the danger of drift between the two (aka it builds but fails to deploy)

I'll rebase all this and merge and we'll take it from there

PR #6293, opened 2026-01-20
Query the current aggregation bucket

isum:
This PR adds support for querying the current, partially filled aggregation bucket with GraphQL.

### How does it work?

When the `current: include` is specified in the GraphQL request for an aggregation, the graph-node loads the most recent time series entities that do not currently form a complete bucket and calculates the aggregates on the fly. Then, the results are merged with the relevant complete buckets to produce one response list.

### DIFF
--- a/store/postgres/src/relational/rollup.rs
+++ b/store/postgres/src/relational/rollup.rs
@@ -238,7 +238,7 @@ pub(crate) struct Rollup {
 
     /// The SQL query that loads the current, partially filled bucket.
     ///
-    /// Contains a `--FILTERS;` comment that can be replaced with additional filters like `and c.block$ <= $1`.
+    /// Contains a `/*FILTERS*/` comment that can be replaced with additional filters like `and c.block$ <= $1`.
     /// The filters are applied to the SQL query that loads the time series entities, not to the aggregated values.
     pub(crate) select_current_sql: String,
 }
@@ -549,7 +549,7 @@ impl<'a> RollupSql<'a> {
 
     /// Generates the SQL query that loads the current, partially filled bucket.
     ///
-    /// The generated SQL query contains a `--FILTERS;` comment that can be replaced
+    /// The generated SQL query contains a `/*FILTERS*/` comment that can be replaced
     /// with additional filters like `and c.block$ <= $1`. The filters are applied to
     /// the SQL query that loads the time series entities, not to the aggregated values.
     fn select_current(&self, w: &mut dyn fmt::Write) -> fmt::Result {
@@ -658,7 +658,7 @@ impl<'a> RollupSql<'a> {
     ///     {dimensions},
     ///     {aggregates}
     ///   from
-    ///     ({select timeseries entities} where timestamp >= {last rollup timestamp} {--FILTERS;})
+    ///     ({select timeseries entities} where timestamp >= {last rollup timestamp} {/*FILTERS*/})
     ///   group by
     ///     {dimensions};
     fn select_current_bucket(&self, w: &mut dyn fmt::Write) -> fmt::Result {
@@ -694,9 +694,11 @@ impl<'a> RollupSql<'a> {
             src_table = self.src_table,
             last_rollup = self.last_rollup(),
         )?;
-        write!(w, " --FILTERS;) c")?;
-        write!(w, " group by timestamp")?;
-        write_dims(self.dimensions, w, true)?;
+        write!(w, " /*FILTERS*/) c")?;
+        if !self.dimensions.is_empty() {
+            write!(w, " group by ")?;
+            write_dims(self.dimensions, w, false)?;
+        }
         Ok(())
     }
 }
@@ -894,8 +896,8 @@ mod tests {
                        "token", "amount", "price" \
                   from "sgd007"."data" as c \
                  where c.timestamp >= coalesce((select max(timestamp) + '3601 s'::interval as last_rollup \
-                       from "sgd007"."stats_hour"), '-infinity'::timestamptz) --FILTERS;) c \
-         group by timestamp, "token""#;
+                       from "sgd007"."stats_hour"), '-infinity'::timestamptz) /*FILTERS*/) c \
+         group by "token""#;
 
         const STATS_DAY_CURRENT_SQL: &str = r#"\
         select max(id) as id, max(vid) as vid, max(block$) as block$, max(timestamp) as timestamp, \
@@ -904,8 +906,8 @@ mod tests {
                        "token", "amount", "price" \
                   from "sgd007"."data" as c \
                  where c.timestamp >= coalesce((select max(timestamp) + '86401 s'::interval as last_rollup \
-                       from "sgd007"."stats_day"), '-infinity'::timestamptz) --FILTERS;) c \
-         group by timestamp, "token""#;
+                       from "sgd007"."stats_day"), '-infinity'::timestamptz) /*FILTERS*/) c \
+         group by "token""#;
 
         const TOTAL_CURRENT_SQL: &str = r#"\
         select max(id) as id, max(vid) as vid, max(block$) as block$, max(timestamp) as timestamp, \
@@ -914,8 +916,7 @@ mod tests {
                        "amount", "price" \
                   from "sgd007"."data" as c \
                  where c.timestamp >= coalesce((select max(timestamp) + '86401 s'::interval as last_rollup \
-                       from "sgd007"."total_stats_day"), '-infinity'::timestamptz) --FILTERS;) c \
-         group by timestamp"#;
+                       from "sgd007"."total_stats_day"), '-infinity'::timestamptz) /*FILTERS*/) c"#;
 
         const OPEN_CLOSE_CURRENT_SQL: &str = r#"\
         select max(id) as id, max(vid) as vid, max(block$) as block$, max(timestamp) as timestamp, \
@@ -926,8 +927,7 @@ mod tests {
                        "amount", "price" \
                   from "sgd007"."data" as c \
                  where c.timestamp >= coalesce((select max(timestamp) + '86401 s'::interval as last_rollup \
-                       from "sgd007"."open_close_day"), '-infinity'::timestamptz) --FILTERS;) c \
-         group by timestamp"#;
+                       from "sgd007"."open_close_day"), '-infinity'::timestamptz) /*FILTERS*/) c"#;
 
         const LIFETIME_CURRENT_SQL: &str = r#"
         with bucket as (select max(id) as id, max(vid) as vid, max(block$) as block$,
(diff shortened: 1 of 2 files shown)


graph/src/amp/sql/query_builder/parser.rs
```diff
@@ -42,15 +42,15 @@ struct AllowOnlySelectQueries;
 
 impl AllowOnlySelectQueries {
     /// Returns an error if the `set_expr` is not a `SELECT` expression.
-    fn visit_set_expr(&self, set_expr: &ast::SetExpr) -> Result<()> {
+    fn visit_set_expr(set_expr: &ast::SetExpr) -> Result<()> {
```
lutter, line 45: The changes in this file seem unrelated, right?

graph/src/components/store/mod.rs
```diff
@@ -465,6 +465,8 @@ pub struct EntityQuery {
     pub query_id: Option<String>,
 
     pub trace: bool,
+
+    pub aggregation_current: Option<AggregationCurrent>,
```
lutter, line 470: When is this `None`? I assume when the query does not access aggregations

graph/src/data_source/common.rs
```diff
@@ -190,9 +190,11 @@ impl AbiJson {
...
                                                     }
                                                     // Recursively resolve the nested path
-                                                    return self
-                                                        .resolve_field_path(components, nested_path)
-                                                        .map(Some);
+                                                    return Self::resolve_field_path(
+                                                        components,
+                                                        nested_path,
+                                                    )
+                                                    .map(Some);
```
lutter, line 277: These changes also seem unrelated

graphql/src/execution/ast.rs
```diff
@@ -364,6 +364,26 @@ impl Field {
...
+        let Some(value) = self.argument_value(kw::CURRENT) else {
+            return Ok(None);
+        };
+
+        if let r::Value::Enum(current) = value {
+            match current.as_str() {
+                "exclude" => return Ok(Some(AggregationCurrent::Exclude)),
+                "include" => return Ok(Some(AggregationCurrent::Include)),
+                _ => {}
+            }
```
lutter, line 378: This could be a `FromStr` implementation on `AggregationCurrent`

graphql/src/store/prefetch.rs
```diff
@@ -644,6 +645,10 @@ impl<'a> Loader<'a> {
                     let field_type = object_type
                         .field(&field.name)
                         .expect("field names are valid");
+
+                    // Loading the current bucket is not supported for nested queries
+                    aggregation_current = None;
```
lutter: One case where that might be wanted is when you query e.g., for a list of tokens and want stats for each of them, i.e., some structure like `{ tokens { id tokenStats(..) { .. } } }`. 

But it's fine to leave that as a TODO for now.

I am not entirely sure whether silently suppressing `current: include` is the right approach. I think we should produce an error during query validation for this case.

graphql/src/store/prefetch.rs
```diff
@@ -728,6 +732,10 @@ impl<'a> Loader<'a> {
             );
         }
 
+        if child_type.is_aggregation() {
+            query.aggregation_current = Some(aggregation_current.unwrap_or_default());
+        }
```
lutter, line 734: I think `child_type.is_aggregation() && aggregation_current.is_some()` indicates an implementation error somewhere that we should detect (similar for the opposite case)

Also, I don't understand why this isn't simply `query.aggregation_current = aggregation_current`?

store/postgres/src/relational/rollup.rs
```diff
@@ -229,13 +229,18 @@ impl<'a> Agg<'a> {
...
     /// A query that determines the last time a rollup was done. The query
     /// finds the latest timestamp in the aggregation table and adds the
     /// length of the aggregation interval to deduce the last rollup time
     last_rollup_sql: String,
+
+    /// The SQL query that loads the current, partially filled bucket.
+    ///
+    /// Contains a `--FILTERS;` comment that can be replaced with additional filters like `and c.block$ <= $1`.
+    /// The filters are applied to the SQL query that loads the time series entities, not to the aggregated values.
+    pub(crate) select_current_sql: String,
```
lutter, line 243: Nice! It would be good to have tests in the `tests` module that check that we generate the SQL that we expect to generate. That will be helpful if we ever need to change query generation (and helps understand what these queries look like)

store/postgres/src/relational_queries.rs
```diff
@@ -4284,20 +4293,59 @@ impl<'a> FilterQuery<'a> {
...
+    fn find_rollup(
+        collection: &'a FilterCollection,
+        layout: &'a Layout,
+        aggregation_current: Option<AggregationCurrent>,
+    ) -> Result<Option<&'a Rollup>, QueryExecutionError> {
+        if !matches!(aggregation_current, Some(AggregationCurrent::Include)) {
+            return Ok(None);
+        }
+
+        // Supporting window and/or multiple entity queries would make the existing SQL queries even more complicated.
```
lutter: It's been a long time since I looked at this, but as I remember for nested queries, the `EntityQuery` will have an `EntityWindow`; it will be more than one only if the `child_type` is an interface (i.e., not in our case) so that it should be possible to have a query `select .. from (select <whatever we do for the only window> from agg_table union all <current_rollup_sql>) order by ..` 

But as I said, that can be done in a separate PR

store/postgres/src/relational_queries.rs
```diff
@@ -4284,20 +4293,59 @@ impl<'a> FilterQuery<'a> {
...
+        // It is also unclear whether supporting that would be a valid use case.
+        let entity = match collection {
+            FilterCollection::All(entities) if entities.len() == 1 => &entities[0],
+            _ => return Err(QueryExecutionError::NotSupported(
+                "The current aggregation bucket can only be queried in a root query for one aggregation entity".to_string())
+            ),
+        };
+
+        // This is not supported because the use of `SELECT *` does not always produce the expected results when combined with `UNION ALL`.
```
lutter, line 4366: We can get rid of `AttributesNames::All` and simplify a bunch of stuff (also a separate PR)

store/postgres/src/relational_queries.rs
```diff
@@ -4284,20 +4293,59 @@ impl<'a> FilterQuery<'a> {
...
+        if matches!(entity.column_names, AttributeNames::All) {
+            return Err(QueryExecutionError::NotSupported(
+                "The current aggregation bucket can only be queried when fields are explicitly selected".to_string())
+            );
+        }
+
+        Ok(layout
+            .rollups
+            .iter()
+            .find(|rollup| rollup.agg_table.object.as_str() == entity.table.meta.object.as_str()))
```
lutter: This would be nicer as a method on `Layout`

graphql/src/store/query.rs
```diff
@@ -552,6 +553,14 @@ fn build_order(
...
             }
         }
+        // Apply a default ordering to the aggregations so that the most recent buckets are returned first
+        (None, _) if entity.is_aggregation() => {
+            let ts = entity
+                .field(kw::TIMESTAMP)
+                .expect("aggregation entities have timestamps");
+
+            EntityOrder::Descending(ts.name.to_string(), ts.value_type)
+        }
```
lutter, line 563: I assume we need to resolve that here because the code in `relational_queries` doesn't have enough information to do it there. It's a little nasty that in some cases we turn the default into a real order, and in others we pass `EntityOrder::Default' and let query generation resolve it. But I am not sure how to make that more sensible (and definitely out of scope for this PR)

lutter (changes requested):

graph/src/amp/sql/query_builder/parser.rs
isum: These are clippy lint fixes

graph/src/components/store/mod.rs
isum: Yes, the `aggregation_current` is `None` for non-aggregation queries and `Some(_)` for aggregation queries

graph/src/data_source/common.rs
isum: These are clippy lint fixes

graphql/src/execution/ast.rs
isum: I decided to keep it simple since it's only needed in one place and it's only a few lines of code

graphql/src/store/prefetch.rs
isum: Fixed

store/postgres/src/relational/rollup.rs
isum: Added tests

store/postgres/src/relational_queries.rs
isum: Added support for nested queries

graphql/src/store/query.rs
isum: Even if this is a real order, it's still the default order for the aggregations, but I agree that we should find a better place for such things

graphql/src/store/prefetch.rs
isum: Initially, I wanted to make this a boolean, but then decided to go with a custom type just to make code more readable, and easier to understand. Now, when `query.aggregation_current` is `None`, this means that the query is not related to aggregations, and when it's `Some(_)` it is an aggregation query

store/postgres/src/relational_queries.rs
```diff
@@ -4361,19 +4421,61 @@ impl<'a> FilterQuery<'a> {
...
+                write_column_names(&wh.column_names, wh.table, Some("c."), out)?;
+                self.filtered_rows(wh, out)?;
+                out.push_sql("\n ");
+                self.limit.sort_key.order_by(out, false)?;
+                out.push_sql(" limit ");
+                out.push_sql(&self.limit.max_num_rows().to_string());
+                out.push_sql(") union all (select ");
+                write_column_names(&wh.column_names, wh.table, Some("c."), out)?;
+                out.push_sql(" from (");
+                let mut current_sql_split = rollup.select_current_sql.split("--FILTERS;");
```
lutter: It might be good to make sure the split results in exactly 2 components (maybe as a `debug_assert!`) Or, the `select_current_sql` could be stored pre-split as `(String, String)`

lutter (approved):
Nice! This will be very useful.

Some comments from Claude - not entirely convinced they are necessary; feel free to address or ignore before merging:

```
3. nested_current_include_count test is redundant

 File: aggregation.rs

 This test verifies the exact same thing as nested_current_include (6 total rows, 3 per token). It provides no additional signal. Consider replacing it with a more useful test — e.g., testing pagination (first/skip) with current: include, or testing what happens
 when the current argument is omitted.

 4. Missing test coverage for key scenarios

 File: aggregation.rs, runner_tests.rs

 - current argument omission: No test verifies that omitting current defaults to exclude behavior through the full GraphQL → store path. The unwrap_or_default() in prefetch.rs handles this, but it's not tested end-to-end.
 - first/last with multiple data points in current bucket: Store tests have only 1 data point per token in the current bucket, so first == last trivially. The runner test checks sum only.
 - Empty current bucket (all data rolled up, no new data): Not tested — does current: include gracefully return the same as exclude?
 - Error paths: The NotSupported error for AttributeNames::All and the ParentLink rejection are untested.

 5. max_num_rows() could theoretically overflow

 File: relational_queries.rs

 fn max_num_rows(&self) -> u32 {
     self.range.0.first.unwrap_or(EntityRange::FIRST) + self.range.0.skip
 }

 If both first and skip are near u32::MAX, this overflows. In practice the values are small (default first=100), but saturating_add would be defensive. 
```

PR #6300, opened 2026-01-24
perf: remove blocking calls from POI queries

MoonBoi9001:
## Summary

- Makes the `proofOfIndexing` resolver async, removing `graph::block_on` calls that blocked the Tokio runtime
- Adds a new `blockForPoi` query that finds which block number produced a given proof of indexing, supporting dispute investigation where an indexer may have submitted a POI for block N that actually corresponds to block X < N

## `blockForPoi` design

The resolver queries all historical POI digest entries from `poi2$` in a single DB call, then searches backwards from `endBlock` in 1M-block chunks. Block hash fetching is pipelined (the next chunk is prefetched while the current one is being processed), and POI computation within each chunk is parallelized across all CPU cores via rayon's `find_map_any`.

Key decisions:
- Searches backwards from `endBlock` towards `startBlock`
- `indexer` is required so the caller specifies which address to use in the POI computation
- No artificial cap on the search range; internally processes in 1M-block chunks

---
Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/server/index-node/src/resolver.rs
+++ b/server/index-node/src/resolver.rs
@@ -352,7 +352,10 @@ where
         ))
     }
 
-    fn resolve_proof_of_indexing(&self, field: &a::Field) -> Result<r::Value, QueryExecutionError> {
+    async fn resolve_proof_of_indexing(
+        &self,
+        field: &a::Field,
+    ) -> Result<r::Value, QueryExecutionError> {
         let deployment_id = field
             .get_required::<DeploymentHash>("subgraph")
             .expect("Valid subgraphId required");
@@ -381,7 +384,7 @@ where
         let poi_fut = self
             .store
             .get_proof_of_indexing(&deployment_id, &indexer, block.clone());
-        let poi = match graph::futures03::executor::block_on(poi_fut) {
+        let poi = match poi_fut.await {
             Ok(Some(poi)) => r::Value::String(format!("0x{}", hex::encode(poi))),
             Ok(None) => r::Value::Null,
             Err(e) => {
@@ -791,7 +794,7 @@ where
             field.name.as_str(),
             scalar_type.name.as_str(),
         ) {
-            ("Query", "proofOfIndexing", "Bytes") => self.resolve_proof_of_indexing(field),
+            ("Query", "proofOfIndexing", "Bytes") => self.resolve_proof_of_indexing(field).await,
             ("Query", "blockData", "JSONObject") => self.resolve_block_data(field).await,
             ("Query", "blockHashFromNumber", "Bytes") => {
                 self.resolve_block_hash_from_number(field).await
--- a/server/index-node/src/resolver.rs
+++ b/server/index-node/src/resolver.rs
@@ -1,6 +1,7 @@
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, HashMap};
 
 use async_trait::async_trait;
+use graph::components::subgraph::ProofOfIndexingFinisher;
 use graph::data::query::Trace;
 use graph::data::store::Id;
 use graph::prelude::alloy::primitives::Address;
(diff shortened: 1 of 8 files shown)


lutter:
The first commit is great and something that absolutely needs to happen. For the second commit, I am not quite clear on what the motivation is - are PoI requests slow? The worry I have is that when `graph-node` is under a lot of load, parallelizing these requests makes the situation worse. The code limits the number of requests to 10, which is good, but I'd still like to understand what the motivation for this change is.

MoonBoi9001:
The motivation is POI verification tooling for dispute investigation.                                                                                   
                                                                                                                                                          
When investigating POI discrepancies (e.g., for arbitration disputes), we need to compute POIs across large block ranges to identify where divergence occurred.

With sequential processing, throughput is bottlenecked at ~2k POIs/second even though the database connection pool has capacity. The publicProofsOfIndexing endpoint accepts batch requests (up to 10), but processing them sequentially means we're not utilizing available parallelism. Checking every block sequentially on a network with 100m blocks takes 13+ hours at ~2k blocks/second.

MoonBoi9001:
> The worry I have is that when graph-node is under a lot of load, parallelizing these requests makes the situation worse.

The third commit addresses this, now pre-fetch all block hashes in a single batch query before parallel processing, reducing DB round-trips.

lutter:
> With sequential processing, throughput is bottlenecked at ~2k POIs/second even though the database connection pool has capacity. The publicProofsOfIndexing endpoint accepts batch requests (up to 10), but processing them sequentially means we're not utilizing available parallelism. Checking every block sequentially on a network with 100m blocks takes 13+ hours at ~2k blocks/second.

Have you tried a binary search for PoI's? I assume you are looking for the point where the PoI's for some subgraph diverge between two indexers. Instead of a binary search, you could actually request multiple PoI's across some range (start with `[0, current subgraph head]`) and then refine the range so that the point of divergence is always in the range. That should yield results very rapidly.

MoonBoi9001:
Binary search works well for finding divergence points between two indexers. This PR speeds up computations for finding which block a given POI actually corresponds to if it doesnt correspond to the block that it is claimed to (can be used to prove an indexer claimed rewards for block N, but only synced to block X). Under the assumption that an indexer force closes with a valid POI for a block x < n, but submits that block x POI for block n.

lutter:
Thanks for the explanation - if you use this to find the block for which a given POI was generated, how about we add a route in the resolver that just does that? That might be much faster than just speeding up POI lookup.

MoonBoi9001:
Great suggestion @lutter I dropped those last two commits and implemented a route in [b8bd47d](https://github.com/graphprotocol/graph-node/pull/6300/commits/b8bd47d14e01fcfde6ab61573b8c6d1cfad7b781)

server/index-node/src/resolver.rs
```diff
@@ -454,6 +455,237 @@ where
...
+            .expect("Valid endBlock required");
+
+        let indexer = Some(
+            field
+                .get_required::<Address>("indexer")
+                .expect("Valid indexer required"),
+        );
+
+        if end_block <= start_block {
+            return Ok(r::Value::Null);
```
lutter, line 484: I think this should be a `QueryExecutionError::ValidationError`

server/index-node/src/resolver.rs
```diff
@@ -454,6 +455,237 @@ where
...
+
+        let target_bytes: [u8; 32] = match target_poi_hash.as_slice().try_into() {
+            Ok(bytes) => bytes,
+            Err(_) => {
+                error!(
+                    self.logger,
+                    "Invalid targetPoi: expected 32 bytes";
+                    "got_bytes" => target_poi_hash.as_slice().len()
+                );
+                return Ok(r::Value::Null);
```
lutter, line 495: Similarly, this should return an error, too

server/index-node/src/resolver.rs
```diff
@@ -454,6 +455,237 @@ where
         Ok(r::Value::List(public_poi_results))
     }
 
+    async fn resolve_block_for_poi(
+        &self,
+        field: &a::Field,
+    ) -> Result<r::Value, QueryExecutionError> {
```
lutter, line 461: I think it's true, but just wanted to make extra sure: this can't be used to exfiltrate a POI for a specific indexer, right? From reading the code, it seems you only get anything back when you hit the POI that gets passed in.

server/index-node/src/resolver.rs
```diff
@@ -454,6 +455,237 @@ where
...
+                );
+                return Ok(r::Value::Null);
+            }
+        };
+
+        // Fetch the full digest history for the block range
+        let history = match self
+            .store
+            .get_poi_digest_history(&deployment_id, start_block..end_block)
+            .await
```
lutter, line 517: We should limit the range of blocks that can be checked. There could be millions of POI entries, and the `history` could become gigantic

server/index-node/src/resolver.rs
```diff
@@ -454,6 +455,237 @@ where
...
+                    } else {
+                        None
+                    }
+                })
+                .collect();
+
+            // Parallel POI computation across all cores via rayon
+            let re = region_entries.clone();
+            let did = deployment_id.clone();
+            let result = graph::spawn_blocking_allow_panic(move || {
```
lutter, line 618: We shouldn't do this - we should not spawn an unlimited number of threads. In fact, we shouldn't spawn _any_ threads as this is all running in a shared server and we should leave some room for other requests to go through

server/index-node/src/resolver.rs
```diff
@@ -454,6 +455,237 @@ where
...
+        let network = match self.store.network_for_deployment(&deployment_id).await {
+            Ok(n) => n,
+            Err(e) => {
+                error!(
+                    self.logger,
+                    "Failed to resolve network for deployment";
+                    "subgraph" => &deployment_id,
+                    "error" => format!("{:?}", e)
+                );
+                return Ok(r::Value::Null);
```
lutter, line 509: Should return an error, too

server/index-node/src/resolver.rs
```diff
@@ -454,6 +455,237 @@ where
...
+        for entry in history.entries {
+            region_entries.entry(entry.id).or_default().push((
+                entry.start_block,
+                entry.end_block,
+                entry.digest,
+            ));
+        }
+        for entries in region_entries.values_mut() {
+            entries.sort_by_key(|(start, _, _)| *start);
+        }
```
lutter, line 547: We could skip this sort if `store.get_poi_digest_history` returned the entries already sorted by start block

lutter (changes requested):

lutter:
Not sure how much it matters for these investigations, but I just realized that for pruned subgraphs, the PoI table also gets pruned

lutter:
Seems like this has gone stale. Closing that for now. Feel free to reopen when you get a chance to work on this.

PR #6301, opened 2026-01-25
chore: fixed broken link

operagxoksana:


lutter:
Thanks for this! I just applied these changes directly.

PR #6310, opened 2026-01-28
Remove web3 from integration tests and other test simplifications

lutter:
* Use alloy instead of web3 in tests
* Make integration tests a little more robust when rerunning against existing anvil
* Do not hardcode PoIs in grafted test
* Make topic-filter test independent of txn hashes


lutter:
Screwed up some git commands bu this is merged as of 4db9be2

PR #6330, opened 2026-02-05
 Fix block cache deserialization for old ethereum blocks

incrypto32:


chain/ethereum/src/chain.rs
```diff
@@ -1076,22 +1077,42 @@ impl TriggersAdapterTrait<Chain> for TriggersAdapter {
...
+        // 1. Full RPC format: {"block": {...}, "transaction_receipts": [...]}
+        // 2. Shallow/header-only: {"timestamp": "...", "data": null} - only timestamp, no block data
+        // 3. Legacy direct: block fields at root level {hash, number, transactions, ...}
+        // We need full format with receipts for ancestor_block (used for trigger processing).
         let block_ptr = match cached {
             Some((json, ptr)) => {
-                if json.get("block").is_none() {
-                    warn!(
+                // Shallow blocks have "data": null - no block data to deserialize
+                if json.get("data") == Some(&json::Value::Null) {
```
lutter: Since this check is repeated a few times and a little noisy, I wonder if it wouldn't be better to introduce a newtype `JsonBlock(json::Value)` so that you could have methods like `is_shallow(&self)` and `patch_transactions` on that

chain/ethereum/src/chain.rs
```diff
@@ -1076,22 +1077,42 @@ impl TriggersAdapterTrait<Chain> for TriggersAdapter {
...
                     ptr
                 } else {
-                    match json::from_value::<EthereumBlock>(json.clone()) {
+                    // Some cached blocks are missing the transaction `type` field
+                    // because graph-node's rust-web3 fork didn't have that field. Patch them
+                    // with type: 0x0 (legacy) so alloy can deserialize them.
+                    let mut json = json;
+                    if let Some(block) = json.get_mut("block") {
+                        json_patch::patch_block_transactions(block);
+                    }
```
lutter: I would put the entire if statement into a `json_block.path_block_transactions()` method

chain/ethereum/src/chain.rs
```diff
@@ -1076,22 +1077,42 @@ impl TriggersAdapterTrait<Chain> for TriggersAdapter {
...
+                    if let Some(receipts) = json.get_mut("transaction_receipts") {
+                        json_patch::patch_receipts(receipts);
+                    }
```
lutter: Same for this if statement

chain/ethereum/src/chain.rs
```diff
@@ -1076,22 +1077,42 @@ impl TriggersAdapterTrait<Chain> for TriggersAdapter {
...
+                    match json::from_value::<EthereumBlock>(json) {
```
lutter: or since patching happens here and on line 1205, just have a `JsonBlock::patched_block<T>(self) -> Result<T, SomeError>` that does the patching and conversion

chain/ethereum/src/json_block.rs
```diff
@@ -0,0 +1,51 @@
+use graph::prelude::serde_json::{self as json, Value};
+use serde::de::DeserializeOwned;
+
+use crate::json_patch;
+
+#[derive(Debug)]
+pub struct EthereumJsonBlock(Value);
```
incrypto32, line 7: Let me know if we should call it just `JsonBlock` named it this because we already have a `JsonBlock` in chainstore and this one is very ethereum specific

lutter: I am fine with either, but since we are in the ethereum crate, the `Ethereum` prefix isn't strictly necessary. But not a big deal either way.

chain/ethereum/src/json_block.rs
```diff
@@ -0,0 +1,51 @@
...
+        }
+        if let Some(receipts) = self.0.get_mut("transaction_receipts") {
+            json_patch::patch_receipts(receipts);
+        }
+    }
+
+    pub fn into_full_block<T: DeserializeOwned>(mut self) -> Result<T, json::Error> {
+        self.patch();
+        json::from_value(self.0)
+    }
```
lutter, line 40: When I suggested the type parameter `T`, I assumed we'd use this for different block forms. But since we need to have different methods for light and full blocks, I think we don't need `T`.

chain/ethereum/src/json_block.rs
```diff
@@ -0,0 +1,51 @@
...
+
+    pub fn is_shallow(&self) -> bool {
+        self.0.get("data") == Some(&Value::Null)
+    }
+
+    pub fn is_legacy_format(&self) -> bool {
+        self.0.get("block").is_none()
+    }
+
+    pub fn patch(&mut self) {
```
lutter, line 27: It would be good to have a brief comment on this and the following methods about what they do. It's noted in the `json_patch` module, but when looking through the code using this, you'd have to dig through the implementation

chain/ethereum/src/json_block.rs
```diff
@@ -0,0 +1,51 @@
...
+        json_patch::patch_block_transactions(&mut inner);
+        json::from_value(inner)
+    }
+}
+
+impl From<Value> for EthereumJsonBlock {
+    fn from(value: Value) -> Self {
+        Self::new(value)
+    }
+}
```
lutter: This is definitely a matter of taste, but I generally try to not have `From` implementations that just call `new` or other constructors, since when you're reading code `value.into()` is much less clear than `EthereumJsonBlock::new(value)`

Issue #6335, opened 2026-02-07
Bug: Duplicate VID constraint violation when multiple offchain triggers fire in the same block


hudsonhrh:
# Bug: Duplicate VID constraint violation when multiple offchain triggers fire in the same block

> **Submitter's note:** This issue was researched and written with the assistance of Claude Code. The bug was discovered while indexing the POP (Perpetual Organization Protocol) subgraph on hoodi, and the root cause and fix were identified by tracing the graph-node source code.

## Summary

When `specVersion >= 1.3.0` (which enables deterministic VID generation), multiple offchain triggers (e.g. file/ipfs data sources) processed in the same block each create a fresh `EntityCache` with `vid_seq` reset to `RESERVED_VIDS` (100). If two or more offchain triggers in the same block write entities to the same table, they produce identical VIDs, causing a PostgreSQL unique constraint violation.

## Affected Subgraph

This bug was discovered while indexing the POP subgraph (`QmVDwncSF2SKkux73Gv2hDjS3jBgNF54mP7DMzBMzTmNUi`) on the hoodi testnet. The subgraph uses `TaskMetadata @entity(immutable: false)` entities populated by `file/ipfs` data source handlers. When multiple tasks are created in the same transaction, multiple IPFS file data sources fire in the same block, triggering the VID collision.

The issue was identified and root-caused with the assistance of Claude Code (Anthropic's CLI tool).

## Error Message

```
Subgraph writer failed, error: database constraint violated: duplicate key value
violates unique constraint "task_metadata_pkey": Key (vid)=(9293978516062308) already exists
```

Decoding the VID: `9293978516062308 = (2163923 << 32) + 100`, confirming the sequence number is always 100 (`RESERVED_VIDS`) — each offchain trigger resets to the same starting point.

## Environment

- graph-node version: v0.36.0 (master at commit `3ca739f`)
- `specVersion: 1.3.0` (deterministic VID generation enabled)
- Multiple `file/ipfs` data sources triggered in the same block

### The Bug Location

**File:** `core/src/subgraph/runner.rs`, method `handle_offchain_triggers`

**Problem: Each offchain trigger creates a fresh `BlockState` with `vid_seq` reset to `RESERVED_VIDS` (100)**

```rust
async fn handle_offchain_triggers(
    &mut self,
    triggers: Vec<offchain::TriggerData>,
    block: &Arc<C::Block>,
) -> Result<(Vec<EntityModification>, ...), Error> {
[...]

PR #6336, opened 2026-02-07
core: Fix duplicate VID when multiple offchain triggers fire in the same block

hudsonhrh:
Each iteration of the offchain trigger loop in `handle_offchain_triggers` creates a fresh `BlockState` with `vid_seq` reset to `RESERVED_VIDS` (100). When two or more offchain triggers (e.g. file/ipfs data sources) fire in the same block and write to the same entity table, they produce identical VIDs, causing a unique constraint violation.

This threads `vid_seq` from the onchain `EntityCache` through the offchain trigger loop so each trigger continues the sequence where the previous one left off. Also adds unit tests demonstrating both the collision and the fix.

There are tests for illustrative purposes showing the bug and the fix

Fixes https://github.com/graphprotocol/graph-node/issues/6335

### DIFF
--- a/store/test-store/tests/graph/entity_cache.rs
+++ b/store/test-store/tests/graph/entity_cache.rs
@@ -413,9 +413,18 @@ async fn offchain_trigger_vid_collision_without_fix() {
     // "duplicate key value violates unique constraint"
     // vid = (block << 32) + 100 for BOTH triggers
     let expected_vid = ((block as i64) << 32) + 100;
-    assert_eq!(vid1, expected_vid, "first trigger vid should be (block << 32) + 100");
-    assert_eq!(vid2, expected_vid, "second trigger vid should ALSO be (block << 32) + 100 — the bug!");
-    assert_eq!(vid1, vid2, "VIDs collide — this is the bug that causes the DB constraint violation");
+    assert_eq!(
+        vid1, expected_vid,
+        "first trigger vid should be (block << 32) + 100"
+    );
+    assert_eq!(
+        vid2, expected_vid,
+        "second trigger vid should ALSO be (block << 32) + 100 — the bug!"
+    );
+    assert_eq!(
+        vid1, vid2,
+        "VIDs collide — this is the bug that causes the DB constraint violation"
+    );
 }
 
 // Test that demonstrates the fix: threading vid_seq from one trigger's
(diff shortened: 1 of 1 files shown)


hudsonhrh:
@lutter fixed the styling issue with https://github.com/graphprotocol/graph-node/pull/6336/commits/8ab10ca7d619703fb7e9455bdfbd59c9d92e3bde 
 thanks for triggering the workflow run

lutter:
> @lutter fixed the styling issue with [8ab10ca](https://github.com/graphprotocol/graph-node/commit/8ab10ca7d619703fb7e9455bdfbd59c9d92e3bde) thanks for triggering the workflow run

No Problem. One thing I am not entirely sure about is that now everywhere an `EntityCache` is created, we need to remember to preserve the `vid_seq`. I would like to have something more automatic that is harder to screw up. Maybe we introduce an `EntityCacheBuilder` or some such and have `as_modifications` return that. The builder would then be responsible for preserving state from one cache to the next. But I haven't fully wrapped my head around how the cache and its data move through the code.

hudsonhrh:
> > @lutter fixed the styling issue with [8ab10ca](https://github.com/graphprotocol/graph-node/commit/8ab10ca7d619703fb7e9455bdfbd59c9d92e3bde) thanks for triggering the workflow run
> 
> No Problem. One thing I am not entirely sure about is that now everywhere an `EntityCache` is created, we need to remember to preserve the `vid_seq`. I would like to have something more automatic that is harder to screw up. Maybe we introduce an `EntityCacheBuilder` or some such and have `as_modifications` return that. The builder would then be responsible for preserving state from one cache to the next. But I haven't fully wrapped my head around how the cache and its data move through the code.

Good call. I looked into it a bit more and the blast radius is actually pretty narrow. onchain triggers share a single `EntityCache` passed by value, so `vid_seq` accumulates naturally. It's only the offchain trigger loop that needs manual threading since each trigger gets an isolated cache.

the manual read-before-consume / write-after-create pattern is definitely fragile. I think the simplest fix is adding `vid_seq `to `ModificationsAndCache`  that struct already carries the LFU cache forward, so `vid_seq `should travel with it. then you can’t forget it because it’s part of the return value you’re already destructuring. No new abstractions needed just using existing patterns. 

how does that approach sound to you i could go ahead and implement it if you think it sounds good

lutter:
Reading the code in more detail, I think this is fine as-is

hudsonhrh:
> Can you rebase this on top of latest master? I'll merge it then

should be good now

lutter:
Thanks for the fix!

PR #6353, opened 2026-02-12
chore(deps): upgrade actions/checkout from v4 to v6

operagxoksana:


lutter:
What is the reason/advantage for this change?

operagxoksana:
I think it's good to always have the latest stable version, we avoid detected bugs

PR #6361, opened 2026-02-13
gnd(test): Add mock-based subgraph test runner

dimitrovmaksim:
# Summary

Implements a native test runner for `gnd test` that executes JSON-defined tests through real graph-node infrastructure (store, WASM runtime, trigger processing) with only the blockchain layer mocked. Provides a faster, integrated alternative to Matchstick while maintaining backward compatibility.

## Key Features

- **JSON test format** - Define blocks, events, eth_calls, and GraphQL assertions
- **Real graph-node pipeline** - Tests run through actual indexing infrastructure
- **Auto block triggers** - All block handler filters work (`once`, `polling`, no filter)
- **startBlock support** - Auto-extracts from manifest and numbers blocks accordingly
- **eth_call mocking** - Pre-populate call cache before indexing
- **Smart assertions** - Order-insensitive comparison, type coercion, detailed diffs
- **Database isolation** - Fresh pgtemp database per test
- **Backward compatible** - `--matchstick` flag preserves legacy behavior

## Architecture

```
JSON test → Parse → ABI encode events → StaticStreamBuilder (mock stream)
  → Real graph-node indexing (WASM + triggers + storage)
  → GraphQL assertions → Report results
```

**Design decisions:**
- Reuses actual graph-node stores, runtime, and trigger processing
- Only blockchain RPC is mocked (via `StaticStreamBuilder`)
- Isolated database per test - Each test gets a pgtemp database dropped on completion (default), or a shared persistent database with post-test cleanup (`--postgres-url`)  
- Pre-populated eth_call cache for deterministic results

## Usage

```bash
gnd test                    # Run all tests in tests/
gnd test tests/foo.json     # Run specific test
gnd test --skip-build       # Skip auto-build
gnd test --matchstick       # Legacy Matchstick mode
gnd test --help             # To see usage and all options 
```

## Test Format Example

```json
{
  "name": "Transfer creates entity",
  "blocks": [
    {
      "number": 1,
      "events": [
        {
          "address": "0x1234...",
          "event": "Transfer(address indexed from, address indexed to, uint256 value)",
          "params": {"from": "0xaaaa...", "to": "0xbbbb...", "value": "1000"}
        }
      ],
      "ethCalls": [
        {
          "address": "0x1234...",
          "function": "balanceOf(address)(uint256)",
          "params": ["0xaaaa..."],
          "returns": ["1000000000000000000"]
        }
      ]
    }
  ],
  "assertions": [
    {
      "query": "{ transfer(id: \"1\") { from to value } }",
      "expected": {"transfer": {"from": "0xaaaa...", "to": "0xbbbb...", "value": "1000"}}
    }
  ]
}
```

### Modified Graph-Node Files

- `graph/src/data_source/subgraph.rs` - Minor exports for startBlock support

## Supported Features

| Feature | Status |
|---------|--------|
| Log events | ✅ Supported |
| Block handlers (`once`, `polling`, no filter) | ✅ Supported |
| eth_call mocking | ✅ Supported |
| Dynamic/template data sources | ✅ Supported |
| Transaction receipts | ⚠️ Not implemented |
| File data sources / IPFS | ⚠️ Not implemented |
| Call triggers & Block Handler (`call`) | ❌ TBD if will be supported |

## Breaking Changes

None. Adds native testing while preserving `--matchstick` for backward compatibility.

## Code Attribution

⚠️ Implementation was assisted by Claude Code after a detailed planning and investigation phase.

Human contributions: requirements, design direction, investigation of Graph Node's indexing pipeline, review, validation with real subgraphs, bug fixes, and integration

## Checklist

- [x] Implementation complete
- [x] User documentation (776-line README)
- [x] Backward compatible via `--matchstick`
- [x] Validated with real subgraphs
- [x] Support for BlockHandler filters (except `call`)
- [x] startBlock handling working
- [x] No breaking changes


gnd/README.md
```diff
@@ -301,25 +301,47 @@ Keys are stored in `~/.graph-cli.json`.
...
 | Flag | Short | Description |
 |------|-------|-------------|
-| `--coverage` | `-c` | Run with coverage reporting |
-| `--docker` | `-d` | Run in Docker container |
-| `--force` | `-f` | Force redownload of Matchstick binary |
-| `--logs` | `-l` | Show debug logs |
-| `--recompile` | `-r` | Force recompilation before testing |
-| `--version` | `-v` | Matchstick version to use |
+| `--manifest` | `-m` | Path to subgraph manifest (default: `subgraph.yaml`) |
+| `--skip-build` | | Skip building the subgraph before testing |
```
lutter: I would think that people will use this a lot when developing tests; maybe it should have a short option, too?

gnd/README.md
```diff
@@ -301,25 +301,47 @@ Keys are stored in `~/.graph-cli.json`.
...
+| `--postgres-url` | | PostgreSQL connection URL (env: `POSTGRES_URL`) |
+| `--matchstick` | | Use legacy Matchstick runner |
+| `--docker` | `-d` | Run Matchstick in Docker (requires `--matchstick`) |
+| `--coverage` | `-c` | Run with coverage reporting (requires `--matchstick`) |
+| `--recompile` | `-r` | Force recompilation (requires `--matchstick`) |
+| `--force` | `-f` | Force redownload of Matchstick binary (requires `--matchstick`) |
```
lutter, line 324: It would be nice to communicate somehow that we want to get rid of Matchstick and give people a bit of time to migrate their tests - it really feels like nobody is using it.

gnd/tests/gnd_test.rs
```diff
@@ -0,0 +1,217 @@
...
+fn fixture_path() -> PathBuf {
+    PathBuf::from(env!("CARGO_MANIFEST_DIR"))
+        .join("tests")
+        .join("fixtures")
+        .join("gnd_test")
+        .join("subgraph")
+}
+
+/// Assert that `asc` (AssemblyScript compiler) is available in PATH or in local node_modules.
+fn verify_asc_available(subgraph_dir: &Path) {
```
lutter, line 118: `gnd/src/compiler/asc.rs` has `is_asc_available` and `get_asc_version`. Maybe they can be reused here?

gnd/tests/gnd_test.rs
```diff
@@ -0,0 +1,217 @@
...
+    assert!(
+        fixture.exists(),
+        "Fixture not found at {}",
+        fixture.display()
+    );
+
+    copy_dir_recursive(&fixture, &subgraph_dir).expect("Failed to copy fixture to temp directory");
+
+    // Install npm dependencies (graph-ts, graph-cli)
+    let npm_output = Command::new("npm")
```
lutter: In other places like integration tests, we already require `pnpm`, maybe we should do that here, too

gnd/src/commands/test/runner.rs
```diff
@@ -0,0 +1,789 @@
...
+    // post-test cleanup of the current run's deployment.
+    if db.needs_cleanup() {
+        cleanup(
+            &ctx.store,
+            &manifest_info.subgraph_name,
+            &manifest_info.hash,
+        )
+        .await
+        .ok();
+    }
```
lutter, line 332: At least for failed tests, it might be nicer to not clean up afterwards and leave the data in place to make it easier to troubleshoot what went wrong

gnd/src/commands/test/assertion.rs
```diff
@@ -0,0 +1,229 @@
...
+    }
+}
+
+/// Reorder `actual` arrays to align with `expected`'s element ordering.
+///
+/// When a test fails, the raw diff can be misleading if array elements appear
+/// in a different order — every line shows as changed even if only one field
+/// differs. This function reorders `actual` so that elements are paired with
+/// their closest match in `expected`, producing a diff that highlights only
+/// real value differences.
```
lutter, line 86: Nice!

gnd/src/commands/test/assertion.rs
```diff
@@ -0,0 +1,229 @@
...
+            expected: assertion.expected.clone(),
+            actual: actual_json,
+        }))
+    }
+}
+
+/// Convert graph-node's internal `r::Value` (GraphQL result) to `serde_json::Value`.
+///
+/// Graph-node uses its own value type for GraphQL results. This converts to
+/// standard JSON for comparison with the expected values in the test file.
```
lutter: `r::Value` implements `Serialize`, so this should not be needed, and by using the existing implementation, we catch any quirks that that serialization might have.

gnd/docs/gnd-test.md
```diff
@@ -0,0 +1,752 @@
...
+| Field | Required | Default | Description |
+|-------|----------|---------|-------------|
+| `address` | Yes | — | Contract address (lowercase hex with 0x prefix) |
+| `event` | Yes | — | Full event signature with `indexed` keywords |
+| `params` | No | Empty object | Event parameter values |
+| `txHash` | No | `keccak256(block_number \|\| log_index)` | Transaction hash |
+
+### Event Signature Format
+
+**Important:** Include `indexed` keywords in the signature:
```
lutter, line 114: That's the exact same form in which the event must be listed in the manifest, right? I wonder if there is a way to make the notation more concise, maybe by allowing the use of `<ABI>.<Event>`, i.e. something like `ERC20.Transfer` in addition to the exact event signature. But definitely not something that needs to go into this PR

lutter (approved):
Very nice! I think this is a really useful tool

gnd/README.md
dimitrovmaksim: I can add a warn notification when using matchstick, also add a note about that in the docs.

gnd/tests/gnd_test.rs
dimitrovmaksim: It seems I'm overwriting is_asc_available in my PR with find_asc_binary

gnd/docs/gnd-test.md
dimitrovmaksim: Yeah, it is needed to know if the params should be set as topics (and potentially padded to 32 bytes or keccak hashed if dynamically sized) in case they are indexed. But this should be doable, scan the abi, look at the event definition and encode accordignly. This will be a good QOL improvement.

gnd/tests/gnd_test.rs
dimitrovmaksim: Ah yes, I remember now. `is_asc_available` was looking onl for a globally installed asc. I added a lookup into the local `node_modules` as well.

gnd/src/commands/test/runner.rs
dimitrovmaksim: This works for persisted DBs. For PgTemp, it is doable, it could be persisted, but then on success needs to be manually deleted, so it kind of becomes an inverted logic.

PR #6368, opened 2026-02-15
Make test IPFS URLs configurable via environment variable

isum:
### Summary

  - Rename `ServerAddress::local_rpc_api()` to `ServerAddress::test_rpc_api()` and add support for the `GRAPH_TEST_IPFS_URL` environment variable to override the default IPFS address (`http://127.0.0.1:5001`)
  - Update `ipfs_upload_files` in test utilities to derive its upload URL from `ServerAddress::test_rpc_api()` instead of a hardcoded URL
  - Make `GraphNodeConfig::ipfs_uri` (used by integration tests) also read from `GRAPH_TEST_IPFS_URL`, falling back to `http://127.0.0.1:3001`
  - Update all test call sites across `core`, `graph`, `runtime/test`, and `tests` to use the renamed method

### Motivation

The IPFS URL was hardcoded in multiple places across unit tests, runner tests, and integration tests. This made it impossible to run tests against an IPFS instance on a non-default address without modifying source code. A single `GRAPH_TEST_IPFS_URL` env var now controls the IPFS endpoint for all test contexts.

### DIFF
--- a/graph/src/ipfs/server_address.rs
+++ b/graph/src/ipfs/server_address.rs
@@ -75,9 +75,9 @@ impl ServerAddress {
     }
 
     pub fn test_rpc_api() -> Self {
-        match std::env::var("GRAPH_TEST_IPFS_URL") {
+        match std::env::var("GRAPH_NODE_TEST_IPFS_URL") {
             Ok(value) if !value.is_empty() => Self::new(&value).unwrap_or_else(|e| {
-                panic!("GRAPH_TEST_IPFS_URL contains an invalid URL: {value}: {e}")
+                panic!("GRAPH_NODE_TEST_IPFS_URL contains an invalid URL: {value}: {e}")
             }),
             _ => Self::new("http://127.0.0.1:5001").unwrap(),
         }
--- a/tests/src/config.rs
+++ b/tests/src/config.rs
@@ -143,7 +143,7 @@ impl GraphNodeConfig {
         Self {
             bin,
             ports: GraphNodePorts::default(),
-            ipfs_uri: std::env::var("GRAPH_TEST_IPFS_URL")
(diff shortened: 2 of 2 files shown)


lutter:
`graph-node` and `graphman` already support using the environment variable `IPFS` for this. Should we use that one instead?

graph/src/ipfs/server_address.rs
```diff
@@ -74,8 +74,13 @@ impl ServerAddress {
...
     }
 
-    pub fn local_rpc_api() -> Self {
-        Self::new("http://127.0.0.1:5001").unwrap()
+    pub fn test_rpc_api() -> Self {
+        match std::env::var("GRAPH_TEST_IPFS_URL") {
+            Ok(value) if !value.is_empty() => Self::new(&value).unwrap_or_else(|e| {
+                panic!("GRAPH_TEST_IPFS_URL contains an invalid URL: {value}: {e}")
+            }),
+            _ => Self::new("http://127.0.0.1:5001").unwrap(),
```
lutter, line 82: Maybe the best here would be to use `GRAPH_TEST_IPFS_URL`, if that is not set use `IPFS`, and if that is not set fall back to localhost. `IPFS` is here because `graph-node` and `graphman` accept it

lutter (approved):
Looks good; just the minor naming comment.

One final thing: we have a environment variable `GRAPH_NODE_TEST_CONFIG`, should we use `GRAPH_NODE_TEST_IPFS_URL` instead so that test variables all start with `GRAPH_NODE_TEST` ?

graph/src/ipfs/server_address.rs
isum: The problem with `IPFS` env variable is that it might be set to the IPFS gateway URL _(port :8080)_ and it is read-only API

PR #6383, opened 2026-02-18
Amp-powered subgraphs: Move Amp configs from ENV to TOML

isum:
## Summary

Amp Flight service configuration moves from a single global endpoint _(via `GRAPH_AMP_FLIGHT_SERVICE_ADDRESS` env var or `--amp-flight-service-address` CLI flag)_ to per-chain TOML configuration under `[chains.<name>.amp]`. This allows each chain to connect to its own Amp Flight service with independent credentials and context tables, which is necessary for multi-chain deployments where different chains are served by different Amp instances.

The per-chain config also makes the context table _(used to join block hashes and timestamps into query results)_ explicit rather than discovered at runtime by probing source tables. Previously, manifest resolution would iterate over `source.tables` and try each one as a potential context source, silently skipping failures. Now the operator declares `context_dataset` and `context_table` in the TOML config, and resolution fails clearly if the context table is missing or lacks required columns.

A secondary change adds Amp-based start block resolution: when an Amp subgraph has `start_block > 0`, the registrar queries the Amp Flight service for the block hash at `start_block - 1` instead of making an RPC call to the chain. This falls back to RPC on failure.

### Breaking changes

- **Removed**: `GRAPH_AMP_FLIGHT_SERVICE_ADDRESS` env var and `--amp-flight-service-address` CLI flag
- **Removed**: `GRAPH_AMP_FLIGHT_SERVICE_TOKEN` env var (now `token` field in TOML)
- **Changed**: The `amp` field on a chain section changes from a plain string (network alias) to a TOML table with `address`, `context_dataset`, `context_table`, and optional `token` and `network` fields

### New config format

```toml
[chains.mainnet.amp]
address = "http://amp-flight.example.com:50051"
token = "my-secret-token"
context_dataset = "ethereum"
context_table = "blocks"
network = "ethereum-mainnet"  # optional, defaults to chain name

### DIFF
--- a/core/src/amp_subgraph/manager.rs
+++ b/core/src/amp_subgraph/manager.rs
@@ -9,7 +9,7 @@ use graph::{
     components::{
         link_resolver::{LinkResolver, LinkResolverContext},
         metrics::MetricsRegistry,
-        network_provider::{AmpChainConfig, AmpClients},
+        network_provider::{AmpChainConfig, AmpClients, ChainName},
         store::{DeploymentLocator, SubgraphStore},
         subgraph::SubgraphInstanceManager,
     },
@@ -35,7 +35,7 @@ pub struct Manager<SS, NC> {
     subgraph_store: Arc<SS>,
     link_resolver: Arc<dyn LinkResolver>,
     amp_clients: AmpClients<NC>,
-    amp_chain_configs: HashMap<String, AmpChainConfig>,
+    amp_chain_configs: HashMap<ChainName, AmpChainConfig>,
 }
 
 impl<SS, NC> Manager<SS, NC>
@@ -52,7 +52,7 @@ where
         subgraph_store: Arc<SS>,
         link_resolver: Arc<dyn LinkResolver>,
         amp_clients: AmpClients<NC>,
-        amp_chain_configs: HashMap<String, AmpChainConfig>,
+        amp_chain_configs: HashMap<ChainName, AmpChainConfig>,
     ) -> Self {
         let logger = logger_factory.component_logger("AmpSubgraphManager", None);
         let logger_factory = logger_factory.with_parent(logger);
@@ -145,7 +145,7 @@ where
                         }
                     };
 
-                    let amp_context = network_name.as_deref().and_then(|chain| {
+                    let amp_context = network_name.as_ref().and_then(|chain| {
                         manager
                             .amp_chain_configs
                             .get(chain)
--- a/core/src/subgraph/instance_manager.rs
+++ b/core/src/subgraph/instance_manager.rs
@@ -273,7 +273,7 @@ impl<S: SubgraphStore, AC: amp::Client> SubgraphInstanceManager<S, AC> {
         // Look up the per-chain Amp client based on the network from the
         // raw manifest (before the manifest is moved into parse).
         let amp_client = network_name_from_raw_manifest(&raw_manifest)
-            .as_deref()
+            .as_ref()
             .and_then(|network| self.amp_clients.get(network));
 
         let manifest =
--- a/core/src/subgraph/registrar.rs
+++ b/core/src/subgraph/registrar.rs
@@ -3,6 +3,7 @@ use std::collections::{HashMap, HashSet};
 use async_trait::async_trait;
 use graph::amp;
 use graph::blockchain::{Blockchain, BlockchainKind, BlockchainMap};
+use graph::components::network_provider::ChainName;
 use graph::components::{
     link_resolver::LinkResolverContext,
     network_provider::{AmpChainConfig, AmpChainNames, AmpClients},
@@ -26,7 +27,7 @@ pub struct SubgraphRegistrar<P, S, SM, AC> {
     store: Arc<S>,
     subscription_manager: Arc<SM>,
     amp_clients: AmpClients<AC>,
-    amp_chain_configs: HashMap<String, AmpChainConfig>,
+    amp_chain_configs: HashMap<ChainName, AmpChainConfig>,
     chains: Arc<BlockchainMap>,
     node_id: NodeId,
     version_switching_mode: SubgraphVersionSwitchingMode,
@@ -49,7 +50,7 @@ where
         store: Arc<S>,
         subscription_manager: Arc<SM>,
         amp_clients: AmpClients<AC>,
-        amp_chain_configs: HashMap<String, AmpChainConfig>,
+        amp_chain_configs: HashMap<ChainName, AmpChainConfig>,
         chains: Arc<BlockchainMap>,
         node_id: NodeId,
         version_switching_mode: SubgraphVersionSwitchingMode,
@@ -302,15 +303,13 @@ where
 
         // Extract the network name from the raw manifest and resolve the
         // per-chain Amp client (if any).
-        let resolved_amp_chain = network_name_from_raw_manifest(&raw).map(|network| {
-            let resolved = self.amp_chain_names.resolve(&Word::from(network));
-            resolved.to_string()
-        });
+        let resolved_amp_chain = network_name_from_raw_manifest(&raw)
+            .map(|network| self.amp_chain_names.resolve(&network));
         let amp_client = resolved_amp_chain
-            .as_deref()
+            .as_ref()
             .and_then(|chain| self.amp_clients.get(chain));
         let amp_context = resolved_amp_chain
-            .as_deref()
+            .as_ref()
             .and_then(|chain| self.amp_chain_configs.get(chain))
             .map(|cfg| cfg.context());
 
--- a/graph/src/components/network_provider/mod.rs
+++ b/graph/src/components/network_provider/mod.rs
@@ -55,18 +55,18 @@ impl AmpChainConfig {
 /// instead of a single global `Option<Arc<AC>>`. Use `get(chain_name)` to
 /// retrieve the client for a specific chain.
 pub struct AmpClients<AC> {
-    clients: HashMap<String, Arc<AC>>,
+    clients: HashMap<ChainName, Arc<AC>>,
 }
 
 impl<AC> AmpClients<AC> {
     /// Creates a new `AmpClients` from a map of chain names to clients.
-    pub fn new(clients: HashMap<String, Arc<AC>>) -> Self {
+    pub fn new(clients: HashMap<ChainName, Arc<AC>>) -> Self {
         Self { clients }
     }
 
     /// Returns the Amp client for the given chain, or `None` if no client
     /// is configured for that chain.
-    pub fn get(&self, chain_name: &str) -> Option<Arc<AC>> {
+    pub fn get(&self, chain_name: &ChainName) -> Option<Arc<AC>> {
         self.clients.get(chain_name).cloned()
     }
 
@@ -166,18 +166,18 @@ mod tests {
     #[test]
     fn amp_clients_returns_client_for_configured_chain() {
         let mut map = HashMap::new();
-        map.insert("mainnet".to_string(), Arc::new(42u32));
+        map.insert(ChainName::from("mainnet"), Arc::new(42u32));
         let clients = AmpClients::new(map);
-        let client = clients.get("mainnet");
+        let client = clients.get(&ChainName::from("mainnet"));
         assert!(client.is_some());
         assert_eq!(*client.unwrap(), 42);
     }
 
     #[test]
     fn amp_clients_returns_none_for_unconfigured_chain() {
-        let map: HashMap<String, Arc<u32>> = HashMap::new();
+        let map: HashMap<ChainName, Arc<u32>> = HashMap::new();
         let clients = AmpClients::new(map);
-        assert!(clients.get("mainnet").is_none());
+        assert!(clients.get(&ChainName::from("mainnet")).is_none());
     }
 
     /// Verifies the condition that causes Amp manager registration:
@@ -185,7 +185,7 @@ mod tests {
     #[test]
     fn amp_manager_registered_when_chain_has_config() {
         let mut map = HashMap::new();
-        map.insert("mainnet".to_string(), Arc::new(42u32));
+        map.insert(ChainName::from("mainnet"), Arc::new(42u32));
         let clients = AmpClients::new(map);
         assert!(
             !clients.is_empty(),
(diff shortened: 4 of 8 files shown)

### CODE REVIEW

**Comment by lutter** [comment]
I feel this would be more logical to configure as a new kind of provider so that in the `graph-node.toml` you would say
```toml
[chains.mainnet]
shard = "blocks"
provider = [
  { label = "mainnet-amp", details = { type = "amp", url = "https://..", token = "..", pointers = "ethereum.blocks", name = "ethereum-mainnet" },
  { label = "mainnet-rpc", details = { type = "web3", url = "http://..", features = [..] },
 ..
]
```

For now, we would only allow one `amp` provider, but it leaves the door open to allowing multiple if that would ever make sense. I also renamed the `context_*` settings to `pointers` - I believe that's what we expect from that table, right? That it contains columns for block number, hash, parent hash, and timestamp? I am open to other names, but `context` doesn't seem very clear to me

graph/src/components/network_provider/mod.rs
```diff
@@ -26,10 +27,24 @@ pub type ChainName = crate::data::value::Word;
...
 pub type ProviderName = crate::data::value::Word;
 
-/// Maps AMP network names to internal graph-node chain names.
+/// Resolved per-chain Amp configuration, with the address parsed as a `Uri`.
 ///
-/// AMP-powered subgraphs may use different network names than graph-node
-/// (e.g., AMP uses `"ethereum-mainnet"` while graph-node uses `"mainnet"`).
+/// This struct is the *runtime* counterpart of the TOML-level `AmpConfig`
+/// (which stores the address as a plain `String`). The `Config::amp_chain_configs()`
+/// method bridges the two by parsing each address string into a `Uri`.
```
lutter, line 34: You could avoid this almost identical struct by deserializing the Url in the config (have a look at annotations like `#[serde(with = ..)]` and `#[serde(deserialize_with = ..)]` in `config.rs`

core/src/amp_subgraph/manager.rs
```diff
@@ -112,13 +113,44 @@ where
...
+                            }
+                        },
+                        None => {
+                            anyhow::bail!(
+                                "no network name found in manifest for Amp subgraph '{}'",
+                                deployment.hash
+                            );
+                        }
+                    };
+
```
lutter: This is a lot of mechanical code - might be better to define a helper function for it

core/src/subgraph/instance_manager.rs
```diff
@@ -267,6 +268,17 @@ impl<S: SubgraphStore, AC: amp::Client> SubgraphInstanceManager<S, AC> {
...
+        // raw manifest (before the manifest is moved into parse).
+        let amp_client = raw_manifest
+            .get(serde_yaml::Value::String("dataSources".to_owned()))
+            .and_then(|ds| ds.as_sequence())
+            .and_then(|ds| ds.first())
+            .and_then(|ds| ds.as_mapping())
+            .and_then(|ds| ds.get(serde_yaml::Value::String("network".to_owned())))
+            .and_then(|n| n.as_str())
+            .and_then(|network| self.amp_clients.get(network));
+
```
lutter: There should definitely be a helper to get the network from the manifest

core/src/subgraph/registrar.rs
```diff
@@ -297,6 +297,13 @@ where
             SubgraphRegistrarError::ResolveError(SubgraphManifestResolveError::ResolveError(e))
         })?;
 
+        // Extract the network name from the raw manifest and resolve the
+        // per-chain Amp client (if any).
+        let amp_client = network_name_from_raw(&raw).and_then(|network| {
```
lutter, line 302: There is a helper!

graph/src/components/network_provider/mod.rs
```diff
@@ -41,6 +42,58 @@ pub struct AmpChainConfig {
     pub network: Option<String>,
 }
 
+/// Holds per-chain Amp Flight clients, keyed by chain name.
+///
+/// This wrapper is used to pass per-chain Amp clients through the system
+/// instead of a single global `Option<Arc<AC>>`. Use `get(chain_name)` to
+/// retrieve the client for a specific chain.
+pub struct AmpClients<AC> {
```
lutter, line 50: Isn't this pretty much a `HashMap`? Just a curried type like `type AmpClients<V> = HashMap<String, V>;`

node/src/manager/commands/run.rs
```diff
@@ -144,41 +143,59 @@ pub async fn run(
...
-                &logger_factory,
-                metrics_registry.cheap_clone(),
-                env_vars.cheap_clone(),
-                &cancel_token,
-                network_store.subgraph_store(),
-                link_resolver.cheap_clone(),
-                amp_client.cheap_clone(),
+    let amp_clients = {
+        let amp_chain_configs = config
+            .amp_chain_configs()
```
lutter, line 148: I believe this is the point where the different config syntax we discussed doesn't matter - `Config` accepts a different format, but `amp_chain_configs` can return the same structs as it does now

node/src/manager/commands/run.rs
```diff
@@ -144,41 +143,59 @@ pub async fn run(
...
-            Some(amp_client)
+            match amp::FlightClient::new(amp_chain_config.address.clone()).await {
+                Ok(mut client) => {
+                    if let Some(token) = &amp_chain_config.token {
+                        client.set_auth_token(token);
+                    }
+                    info!(logger, "Amp Flight client connected";
+                        "chain" => &chain_name,
+                        "host" => ?amp_chain_config.address.host()
+                    );
```
lutter, line 165: Since the new config syntax has a label, we should only log the label anywhere where we need to identify the amp server we are talking to. That completely obscures any operational details, like, e.g., host addresses or names

node/src/launcher.rs
```diff
@@ -505,35 +505,40 @@ pub async fn run(
...
+                        "host" => ?amp_chain_config.address.host()
+                    );
+                    clients.insert(chain_name, Arc::new(client));
+                }
+                Err(e) => {
+                    error!(logger, "Failed to connect Amp Flight client";
+                        "chain" => &chain_name,
+                        "error" => e.to_string()
+                    );
+                }
```
lutter, line 535: Code is duplicated from `node/src/manager/commands/run.rs`. This should go into a common setup helper

server/index-node/src/resolver.rs
```diff
@@ -520,6 +521,17 @@ where
...
+        // Extract the network name from the raw yaml to look up the
+        // per-chain Amp client.
+        let amp_client = raw_yaml
+            .get(serde_yaml::Value::String("dataSources".to_owned()))
+            .and_then(|ds| ds.as_sequence())
+            .and_then(|ds| ds.first())
+            .and_then(|ds| ds.as_mapping())
+            .and_then(|ds| ds.get(serde_yaml::Value::String("network".to_owned())))
+            .and_then(|n| n.as_str())
+            .and_then(|network| self.amp_clients.get(network));
```
lutter: Again extracting network name from raw manifest the hard way

node/src/manager/commands/run.rs
```diff
@@ -147,36 +147,51 @@ pub async fn run(
...
-                        client.set_auth_token(token);
+            AmpClients::new(std::collections::HashMap::new())
+        } else {
+            let mut clients = std::collections::HashMap::new();
+            for (chain_name, amp_chain_config) in &amp_chain_configs {
+                debug!(logger, "Connecting to Amp Flight service";
+                    "chain" => chain_name.as_str(),
+                    "host" => ?amp_chain_config.address.host(),
+                    "port" => ?amp_chain_config.address.port()
+                );
```
lutter, line 164: I am not going to find every place where we log to identify the amp server, but this should also log the label

node/src/manager/commands/run.rs
```diff
@@ -147,36 +147,51 @@ pub async fn run(
...
-        for (chain_name, amp_chain_config) in amp_chain_configs {
-            debug!(logger, "Connecting to Amp Flight service";
-                "chain" => &chain_name,
-                "host" => ?amp_chain_config.address.host(),
-                "port" => ?amp_chain_config.address.port()
+
+        if amp_chain_configs.is_empty() {
+            info!(
+                logger,
+                "Amp support disabled — no chains have [amp] configuration"
```
lutter, line 154: I don't think we need to log this - it will only be confusing to users who don't know what amp is

node/src/manager/commands/run.rs
```diff
@@ -147,36 +147,51 @@ pub async fn run(
...
-                        "chain" => &chain_name,
-                        "error" => e.to_string()
-                    );
                 }
             }
+            if clients.is_empty() {
+                warn!(
+                    logger,
+                    "Amp-powered subgraphs disabled — all configured chains failed to connect"
+                );
```
lutter, line 188: I think this logic is not great; since amp clients are per-chain, even one failing to connect is very bad. We should at least log separately for each chain where connection fails. I also think that connecting at startup is the right thing to do in the long term (it's ok for now) What we really need is a more dynamic way to deal with this so that we don't need to have the amp server up when graph-node starts and automatically connect when the amp server comes up.

core/src/subgraph/registrar.rs
```diff
@@ -299,10 +302,17 @@ where
 
         // Extract the network name from the raw manifest and resolve the
         // per-chain Amp client (if any).
-        let amp_client = network_name_from_raw(&raw).and_then(|network| {
+        let resolved_amp_chain = network_name_from_raw(&raw).map(|network| {
             let resolved = self.amp_chain_names.resolve(&Word::from(network));
-            self.amp_clients.get(resolved.as_str())
+            resolved.to_string()
```
lutter, line 307: Just a thought: since it can become confusing whether we are talking about graph-node's or amp's chain name, we might want to introduce newtypes `ChainName(Word)` and `AmpChainName(Word)` but doing that everywhere for the existing chain name might be a pretty big and annoying undertaking

graph/src/amp/manifest/data_source/raw.rs
```diff
@@ -320,23 +331,27 @@ impl RawTransformer {
...
-                    input_schema,
-                    source,
-                    abis,
-                )
-                .await
-                .map_err(|e| e.source_context(format!("invalid `tables` at index {i}")))
+        let table_futs = tables.into_iter().enumerate().map(|(i, table)| {
+            let amp_context = amp_context.clone();
+            async move {
+                let logger = logger.new(slog::o!("table_name" => table.name.clone()));
```
lutter, line 337: I haven't followed where `logger` comes from, but it will help debugging a lot if all loggers that have to do with amp also log the provider label. We don't do that consistently for RPC, and it's really annoying when you need to understand provider errors

graph/src/amp/manifest/data_source/raw.rs
```diff
@@ -571,79 +587,52 @@ impl RawTable {
...
-                dataset,
-                table,
-                columns,
-            ));
-            break;
+        let (context_dataset, context_table) = amp_context.ok_or_else(|| {
+            Error::InvalidQuery(anyhow!(
+                "query requires context columns (block hash/timestamp) but no Amp context config is available"
+            ))
+        })?;
```
lutter, line 594: I think we can catch that when we read the config file

graph/src/data/subgraph/mod.rs
```diff
@@ -971,12 +973,20 @@ impl<C: Blockchain> SubgraphManifest<C> {
         raw: serde_yaml::Mapping,
         resolver: &Arc<dyn LinkResolver>,
         amp_client: Option<Arc<AC>>,
+        amp_context: Option<(String, String)>,
```
lutter, line 976: Since we seem to be passing `amp_context` around with `amp_client` a lot, could we carry the context in the client? It would mostly require an additional method on the `Client` trait.

core/src/subgraph/registrar.rs
```diff
@@ -420,6 +420,54 @@ where
     }
 }
 
+/// Resolves a block pointer for an Amp subgraph by querying the Amp Flight
+/// service for the block hash at the given `block_number`.
+async fn resolve_amp_start_block<AC: amp::Client>(
```
lutter, line 425: This would be a good candidate for a method on `amp::Client`

core/src/subgraph/registrar.rs
```diff
@@ -550,7 +602,40 @@ async fn create_subgraph_version<C: Blockchain, S: SubgraphStore, AC: amp::Clien
...
+                                "block_number" => min - 1
+                            );
+                            resolve_start_block(&manifest, &*chain, &logger).await?
+                        }
+                    }
+                }
+                // Non-Amp subgraph — use RPC.
+                _ => resolve_start_block(&manifest, &*chain, &logger).await?,
+            }
+        }
```
lutter, line 638: The overall `create_subgraph_version` function is already not great, but it would be great if these 30 lines of fairly straightforward code lived somewhere in a helper. I am always worried about the readability of code that mixes control logic with basic "get me this thing" code.

node/src/config.rs
```diff
@@ -155,8 +156,8 @@ impl Config {
                     AmpChainConfig {
                         address: uri,
                         token: amp.token.clone(),
-                        context_dataset: amp.context_dataset.clone(),
-                        context_table: amp.context_table.clone(),
+                        context_dataset: normalize_sql_ident(&amp.context_dataset),
+                        context_table: normalize_sql_ident(&amp.context_table),
```
lutter, line 160: It's definitely good to do the normalization early - I am just not sure if we should do this, since SQL identifiers are case-sensitive, it's just that dealing with identifiers that are not all-lowercase is highly annoying since, at least in Postgres, you need to write `select * from "MyTable"` since it would interpret `select * from MyTable` as the same as `select * from mytable`. An alternative would be to use what the user gave us verbatim but enclose each component in double quotes (if it is not already in double-quotes, it gets complicated)

Either way, fine to leave this as-is for now, just something to think about.

core/src/subgraph/registrar.rs
```diff
@@ -604,7 +605,7 @@ async fn create_subgraph_version<C: Blockchain, S: SubgraphStore, AC: amp::Clien
                 // Genesis block — no resolution needed.
                 (0, _, _) => None,
                 // Amp subgraph with start_block > 0 — try Amp-based resolution.
-                (min, Some(client), Some((dataset, table))) => {
+                (min, Some(client), Some((dataset, table))) if is_amp_subgraph => {
```
lutter, line 608: Isn't there some bigger issue if `is_amp_subgraph` is true and client or context are `None` (and vice versa)? I think we should produce some sort of internal error for those cases rather than silently do something else (like talk to RPC)

core/src/subgraph/registrar.rs
lutter: I see that the last commit uses `ChainName` - but since it's just a type alias and not a newtype we won't get compiler errors when it's used in the wrong place. I am fine with leaving that part to a later PR.

isum:
I will open another PR to make Amp a new kind of provider

Issue #6183, opened 2025-09-30
[Bug] the chain NNN is stored in shard XXX but is configured for shard YYY


madumas:
### Bug report

After the graphman command `change-shard` is used to move a chain NNN to a different shard, graph-node complains on start with the error `the chain NNN is stored in shard XXX but is configured for shard YYY`.

Other than the error message at startup, graph-node runs as expected, using the chain stored on the new shard.

`graph-node 0.40.0`

### Relevant log output

```Shell
graphman -c /etc/graph-node/config.toml chain change-shard avalanche shard18
Changing block cache shard for avalanche to shard18
Current shard: primary
Changed block cache shard for avalanche from primary to shard18
Segmentation fault (core dumped)



graphman -c /etc/graph-node/config.toml chain list
        name         |   shard    | namespace  | version | head block
---------------------+------------+------------+---------+-----------
avalanche            | shard18    | chain35    |   43114 |   69541463
avalanche-old        | primary    | chain3     |   43114 |   69541151



graph-node output:

Sep 30 12:55:53.868 DEBG Database setup finished, component: ConnectionPool > Coordinator
Sep 30 12:55:53.868 INFO Setup finished, shards: 21
Sep 30 12:55:53.868 DEBG Cleaning up large notifications after about 300s, channel: chain_head_updates, component: ChainHeadUpdateListener > NotificationListener
Sep 30 12:55:53.983 ERRO the chain avalanche is stored in shard shard18 but is configured for shard primary, component: BlockStore
```

### Some information to help us out

- [ ] Tick this box if this bug is caused by a regression found in the latest release.
- [ ] Tick this box if this bug is specific to the hosted service.
- [x] I have searched the issue tracker to make sure this issue is not a duplicate.

### OS information

None

dimitrovmaksim:
This is not a bug per se, there's a check if chain config matches the db config, but it should log a warning rather than error.

PR #6385, opened 2026-02-18
node, store: Remove unused ChainStatus enum.

dimitrovmaksim:
Closes https://github.com/graphprotocol/graph-node/issues/6183

- Removes unused ChainStatus enum
- Changes the shard configuration mismatch log from `error` to `warn`

### DIFF
--- a/store/postgres/src/block_store.rs
+++ b/store/postgres/src/block_store.rs
@@ -10,7 +10,7 @@ use graph::{
     blockchain::ChainIdentifier,
     components::store::{BlockStore as BlockStoreTrait, QueryPermit},
     derive::CheapClone,
-    prelude::{error, info, BlockNumber, BlockPtr, Logger, ENV_VARS},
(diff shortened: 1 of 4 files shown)


store/postgres/src/block_store.rs
```diff
@@ -278,39 +270,22 @@ impl BlockStore {
...
-                let status = if chain_ingestible(&block_store.logger, chain, &shard) {
-                    ChainStatus::Ingestible
-                } else {
-                    ChainStatus::ReadOnly
-                };
-                block_store.add_chain_store(chain, status, false).await?;
+                if chain.shard != shard {
+                    warn!(
+                        &block_store.logger,
+                        "the chain {} is stored in shard {} but is configured for shard {}",
```
lutter: The log message would be even clearer if it said
```
"the chain {} is stored in shard {} but is configured for shard {}; ignoring config and using shard {chain.shard}"
```

dimitrovmaksim: Updated and force pushed 🫡

PR #6387, opened 2026-02-22
chore: remove redundant if statement

cuiweixie:
these two statements is duplicated.  
https://github.com/graphprotocol/graph-node/blob/a9b3cf00c59c7bdec9a8c1adcd678c84ebeb242d/chain/ethereum/src/polling_block_stream.rs#L369-L375


lutter:
Nice catch! Thanks for the contribution.

Since the commit wasn't signed, I couldn't merge the PR. But I committed an identical change as 68fc7f6b

PR #6397, opened 2026-02-24
Dump/restore for subgraphs

lutter:
This PR implements `graphman dump` and `graphman restore` commands that do what it says on the tin.

Dumps are consistent, but they are taken using a single transaction; that may have very bad effects on the overall system if dumps take a very long time. Restore is a little nicer in that it splits the data import into multiple jobs.

There are lots of ways in which this could be improved, but I feel this is a useful starting point, at the very least for development and test systems. The help text for the dump and restore command have scary warnings about not using them in production - and they shouldn't, even though I think at least for smaller subgraphs they might work ok.

Here's the help text for `graphman dump`:

```
Dump a subgraph deployment into a directory

EXPERIMENTAL - NOT FOR PRODUCTION USE

This will create a dump of the subgraph deployment in the specified directory. The dump includes
the subgraph manifest, the mapping, and the data in the database as parquet files. The dump can
be used to restore the subgraph deployment later with the `restore` command.

Usage: graphman --config <CONFIG> dump <DEPLOYMENT> <DIRECTORY>

Arguments:
  <DEPLOYMENT>
          The deployment (see `help info`)

  <DIRECTORY>
          The name of the directory to dump to
```

and `grpahman restore`:

```
Restore a subgraph deployment from a dump directory

EXPERIMENTAL - NOT FOR PRODUCTION USE

Restore a subgraph deployment from a dump created with the `dump` command.

Usage: graphman --config <CONFIG> restore [OPTIONS] <DIRECTORY>

Arguments:
  <DIRECTORY>
          Path to the dump directory

Options:
      --shard <SHARD>
          The database shard to restore into (default: primary)

      --name <NAME>
          Subgraph name for deployment rule matching and node assignment. If omitted, uses an
          existing name from the database; errors if none found

      --replace
          Drop and recreate if the deployment already exists in the target shard

      --add
          Create a copy in a shard that doesn't have this deployment (requires --shard)

      --force
          Restore no matter what: replace if exists in target shard, add if not
```

### DIFF
--- a/docs/dump.md
+++ b/docs/dump.md
@@ -23,6 +23,62 @@ in the dump. This can lead to very long restore times for large subgraphs.
 The restore process will be optimized in the future to only create indexes
 that are present in the dump's metadata.
 
+### Usage
+
+#### Dumping a deployment
+
+```bash
+graphman dump <deployment> <directory>
+```
+
+`<deployment>` identifies the subgraph deployment to dump. It can be a
+subgraph name, a deployment hash (`Qm...`), or a database namespace
+(`sgdNNN`). `<directory>` is the path where the dump will be written; it
+will be created if it does not exist.
+
+Running `graphman dump` against an existing dump directory performs an
+**incremental dump**: only rows added since the last dump are exported, and
+new chunk files are appended rather than rewriting existing ones.
+
+```bash
+# Full dump
+graphman dump my-subgraph /backups/my-subgraph
+
+# Incremental update of the same dump
+graphman dump my-subgraph /backups/my-subgraph
+```
+
+#### Restoring a deployment
+
+```bash
+graphman restore <directory> [options]
+```
+
+`<directory>` is the path to a dump previously created with `graphman
+dump`.
+
+| Option      | Description                                                                                         |
+| ----------- | --------------------------------------------------------------------------------------------------- |
+| `--shard`   | Target database shard. Uses deployment rules (or primary shard) when omitted. Required with `--add` |
+| `--name`    | Subgraph name for deployment rule matching and node assignment. Falls back to an existing name       |
+| `--replace` | Drop and recreate if the deployment already exists in the target shard                               |
+| `--add`     | Create a copy in a shard that doesn't already have this deployment (requires `--shard`)              |
+| `--force`   | Replace if the deployment exists in the target shard, add if it doesn't                              |
+
+`--replace`, `--add`, and `--force` are mutually exclusive. When none is
+given, restore fails if the deployment already exists in the target shard.
+
+```bash
+# Restore into the default shard
+graphman restore /backups/my-subgraph
+
+# Restore into a specific shard, replacing if it already exists
+graphman restore /backups/my-subgraph --shard shard1 --replace
+
+# Force-restore (replace or add as needed)
+graphman restore /backups/my-subgraph --force
+```
+
 ### Directory layout
 
 A dump directory has the following structure:


store/postgres/src/subgraph_store.rs
```diff
@@ -735,6 +737,35 @@ impl Inner {
         }
     }
 
+    /// Determine the target node for a deployment using the configured
+    /// deployment rules, ignoring the shard selection. Returns an error
+    /// if no rule matches.
+    async fn node_for_deployment(
```
dimitrovmaksim: The documentation says that this function ignores the shard from `place()`. If `--shard` is not passed, the shard will default to the primary one, but wouldn't this break the deployment rule for that node if a different shard is defined for subgraph deployments in the config?

store/postgres/src/parquet/reader.rs
```diff
@@ -0,0 +1,309 @@
...
+use arrow::array::RecordBatch;
+use graph::components::store::StoreError;
+use parquet::arrow::arrow_reader::ParquetRecordBatchReaderBuilder;
+
+/// Read all record batches from a Parquet file.
+///
+/// Opens the file, reads all row groups, and returns them as a vector
+/// of `RecordBatch`es. The batches retain the schema embedded in the
+/// Parquet file.
+pub fn read_batches(path: &Path) -> Result<Vec<RecordBatch>, StoreError> {
```
dimitrovmaksim: Wouldn't this function load the whole table into the memory? I know it says it's not production ready and mostly for local dev, where probably tables aren't big, but could be a problem for large tables.

store/postgres/src/relational/dump.rs
```diff
@@ -0,0 +1,694 @@
...
+            ColumnType::Bytes => add_typed::<Binary>(select, table, name, is_list, is_nullable),
+            ColumnType::Int => add_typed::<Integer>(select, table, name, is_list, is_nullable),
+            ColumnType::Int8 => add_typed::<BigInt>(select, table, name, is_list, is_nullable),
+            ColumnType::Timestamp => {
+                add_typed::<Timestamptz>(select, table, name, is_list, is_nullable)
+            }
+            ColumnType::String => add_typed::<Text>(select, table, name, is_list, is_nullable),
+            ColumnType::Enum(_) => {
+                // Cast enum to text for dump
+                let alias = table.meta.name.as_str();
```
dimitrovmaksim: This alias is never used, but then the `c` alias is hardcoded into the query (also on row 334 and 335). Probably this `alias` definition should be moved somewhere above and then used in all 3 places.

store/postgres/src/relational/dump.rs
```diff
@@ -0,0 +1,694 @@
...
+            has_causality_region: table.has_causality_region,
+            chunks: vec![],
+            max_vid: -1,
+        });
+    }
+
+    let relative_path = format!("{}/chunk_000000.parquet", table_dir_name);
+    let abs_path = table_dir.join("chunk_000000.parquet");
+    let mut writer = ParquetChunkWriter::new(abs_path, relative_path, &arrow_schema)?;
+
```
dimitrovmaksim, line 691: It would be nice to have some kind of progression logging, something like

```suggestion
println!("Dumping table {} into {}", table_name, relative_path)
```

store/postgres/src/relational/dump.rs
```diff
@@ -0,0 +1,694 @@
...
+use crate::relational::index::IndexList;
+use crate::relational::value::OidRow;
+use crate::relational::{ColumnType, SqlName, Table as RelTable};
+use crate::vid_batcher::{VidBatcher, VidRange};
+use crate::AsyncPgConnection;
+
+use super::Layout;
+
+#[derive(Serialize, Deserialize)]
+pub(crate) struct Manifest {
```
dimitrovmaksim, line 34: Just a thought, but this Manifest structs and SubgraphManifestEntity seem pretty similar, is it possible to use the SubgraphManifestEntity direclty?

node/src/manager/commands/dump.rs
```diff
@@ -0,0 +1,30 @@
...
+
+use crate::manager::deployment::DeploymentSearch;
+
+pub async fn run(
+    subgraph_store: Arc<SubgraphStore>,
+    primary_pool: ConnectionPool,
+    deployment: DeploymentSearch,
+    directory: String,
+) -> Result<()> {
+    let directory = fs::canonicalize(&directory)?;
```
dimitrovmaksim, line 142: Just a suggestion: `dump` does not create the output directory and expects it to already exists or will throw an error, but maybe the command should create it instead.

store/postgres/src/relational/restore.rs
```diff
@@ -0,0 +1,465 @@
...
+        .bind::<Nullable<Text>, _>(row.context.as_deref())
+        .bind::<Nullable<Integer>, _>(row.done_at)
+        .execute(conn)
+        .await
+        .map_err(StoreError::from)?;
+    Ok(())
+}
+
+/// Import the `data_sources$` table from Parquet chunks.
+async fn import_data_sources(
```
dimitrovmaksim, line 153: Not sure if there's something specific about data_sources$ that requires it, but `import_data_sources` seems to insert rows one at a time rather than in batches like the entity tables do. Spec file suggest it has to be batched as well
https://github.com/graphprotocol/graph-node/blob/8f6b0dc8eca7bb1e92bc86e4a1a01f030c74b7dc/docs/specs/dump-restore.md?plain=1#L207

node/src/bin/manager.rs
```diff
@@ -327,6 +327,48 @@ pub enum Command {
...
+    Restore {
+        /// Path to the dump directory
+        directory: String,
+        /// The database shard to restore into (default: primary)
+        #[clap(long)]
+        shard: Option<String>,
+        /// Subgraph name for deployment rule matching and node assignment.
+        /// If omitted, uses an existing name from the database; errors if none found.
+        #[clap(long)]
+        name: Option<String>,
```
dimitrovmaksim, line 364: Ooops, just discovered this one.
If `--name` points to a subgraph that doesn't exist in the DB, restore still runs to completion but the deployment ends up orphaned — no name association, only reachable by hash, and is hard to clean up as `unused record` and `unused remove` does not see it. This should be validated, before restore is executed.
Also, docs should probably mention that in a fresh environment you need to run `graphman create <NAME>` first.

store/postgres/src/relational/dump.rs
lutter: I don't think that's a good idea, as a change to `SubgraphManifestEntity` would inadvertently change the format of the `metadata.json`, and therefore how subgraphs are dumped.

store/postgres/src/parquet/reader.rs
lutter: Nice catch! Yes, it will, fixing that.

store/postgres/src/subgraph_store.rs
lutter: Good point. I am changing this so that `--shard` takes priority, but if not given, use the deployment rules first, and only fall back to the primary when that doesn't yield anything. For `--add`, passing `--shard` is still required since you are essentially creating a copy.

store/postgres/src/relational/dump.rs
lutter: Changed this

node/src/manager/commands/dump.rs
lutter: Good point - changed

store/postgres/src/relational/restore.rs
lutter: Yeah, that was bad; I changed it to batch the inserts

PR #6398, opened 2026-02-24
Replace AnyNetwork with AnyNetworkBare to remove WithOtherFields deserialization overhead

incrypto32:
AnyNetwork wraps every RPC response (block, transaction, receipt) in `WithOtherFields<T>`, which uses `#[serde(flatten)] `to capture unknown JSON keys into a BTreeMap. This forces serde to buffer entire JSON objects into intermediate Value maps and re-serialize them to diff keys on every block and every transaction. graph-node never reads these extra fields.

Define `AnyNetworkBare`, a custom Network impl that uses the same `AnyTxEnvelope`/`AnyReceiptEnvelope` types (for L2/sidechain tx support) but without the WithOtherFields wrapper on response types. This eliminates the flatten buffering, BTreeMap allocation, and re-serialization from the deserialization hot path.


graph/src/components/ethereum/network.rs
```diff
@@ -0,0 +1,212 @@
...
+/// `AnyNetwork` wraps every response (block, transaction, receipt) in `WithOtherFields<T>`,
+/// which captures unknown JSON keys into a `BTreeMap`. This requires `#[serde(flatten)]`,
+/// forcing serde to buffer entire JSON objects into intermediate `Value` maps and
+/// re-serialize them to diff keys — on every block and every transaction.
+/// graph-node never uses these extra fields, so we define a bare network that deserializes
+/// directly into the inner types.
+///
+/// Uses `AnyTxEnvelope` (not `TxEnvelope`) to support non-standard transaction types from
+/// L2s and sidechains. `TransactionRequest` retains `WithOtherFields` because upstream
+/// `From` impls for `AnyTxEnvelope` only exist on `WithOtherFields<TransactionRequest>`.
```
lutter, line 28: Would it be worth to ask upstream to change that?

incrypto32: Upstream is right . AnyTxEnvelope can carry unknown tx type fields that a bare TransactionRequest has no place for. Just documenting why this one retains the wrapper.

PR #6400, opened 2026-02-27
Serve typed blocks from recent block cache to avoid repeated deserialization

incrypto32:


chain/ethereum/src/chain.rs
```diff
@@ -66,6 +65,7 @@ use graph::blockchain::block_stream::{
     BlockStream, BlockStreamBuilder, BlockStreamError, BlockStreamMapper, FirehoseCursor,
     TriggersAdapterWrapper,
 };
+use graph::components::ethereum::EthereumJsonBlock;
```
lutter: Very minor: since this is in a module called `ethereum`, it's probably enough to call this `JsonBlock`

graph/src/components/ethereum/types.rs
```diff
@@ -259,3 +261,55 @@ impl<'a> From<&'a EthereumCall> for BlockPtr {
...
+    pub fn light_block(&self) -> &LightEthereumBlock {
+        match self {
+            CachedBlock::Full(block) => &block.block,
+            CachedBlock::Light(block) => block,
+        }
+    }
+
+    pub fn into_light_block(self) -> LightEthereumBlock {
+        match self {
+            CachedBlock::Full(block) => block.block.as_ref().clone(),
```
lutter: If you change this to
```rust
pub fn into_light_block(self) -> Arc<LightEthereumBlock> {
        match self {
            CachedBlock::Full(block) => block.block,
            CachedBlock::Light(block) => Arc::new(block),
        }
}
```

you can avoid a deep clone of the block

chain/ethereum/src/ethereum_adapter.rs
```diff
@@ -1641,23 +1640,7 @@ impl EthereumAdapterTrait for EthereumAdapter {
...
-                        warn!(
-                            &logger,
-                            "Failed to deserialize cached block: {}. Block will be re-fetched from RPC.",
-                            e
-                        );
-                    })
-                    .ok()
-            })
-            .map(Arc::new)
+            .map(|cached| Arc::new(cached.into_light_block()))
```
lutter: This then becomes `.map(CachedBlock::into_light_block)`

graph/src/components/ethereum/types.rs
```diff
@@ -259,3 +261,55 @@ impl<'a> From<&'a EthereumCall> for BlockPtr {
         BlockPtr::from((call.block_hash, call.block_number))
     }
 }
+
+/// Typed cached block for Ethereum. Stores the deserialized block so that
+/// repeated reads from the in-memory cache avoid `serde_json::from_value()`.
+#[derive(Clone, Debug)]
```
lutter, line 267: It would be nice to get rid of `Clone` for this and all other block structs; that's probably a bigger puzzle and requires using `Arc` judiciously in various data structures, i.e., something for another PR; but conceptually, blocks are readonly and it should therefore be possible to `Arc` them.

graph/src/components/store/traits.rs
```diff
@@ -553,8 +554,12 @@ pub trait ChainStore: ChainHeadStore {
...
     ) -> Result<Option<B256>, Error>;
 
-    /// Returns the blocks present in the store.
-    async fn blocks(
+    /// Returns the blocks present in the store as typed cached blocks.
+    async fn blocks(self: Arc<Self>, hashes: Vec<BlockHash>) -> Result<Vec<CachedBlock>, Error>;
+
+    /// Returns blocks as raw JSON. Used by callers that need the original
+    /// JSON representation (e.g., GraphQL block queries, CLI tools).
+    async fn blocks_as_json(
```
lutter, line 562: We should try and get rid of this one, but that's also for another PR

store/postgres/src/chain_store.rs
```diff
@@ -2519,6 +2543,22 @@ fn json_block_to_block_ptr_ext(json_block: &JsonBlock) -> Result<ExtendedBlockPt
     Ok(ptr)
 }
 
+fn cache_block_to_block_ptr_ext(block: &CacheBlock) -> Result<ExtendedBlockPtr, Error> {
```
lutter: This could be a method on `CacheBlock`

store/postgres/src/chain_store.rs
```diff
@@ -2645,8 +2685,8 @@ impl ChainStoreTrait for ChainStore {
     async fn upsert_block(&self, block: Arc<dyn Block>) -> Result<(), Error> {
         // We should always have the parent block available to us at this point.
         if let Some(parent_hash) = block.parent_hash() {
-            let block = JsonBlock::new(block.ptr(), parent_hash, block.data().ok());
-            self.recent_blocks_cache.insert_block(block);
+            let json_block = JsonBlock::new(block.ptr(), parent_hash, block.data().ok());
+            self.recent_blocks_cache.insert_json_block(json_block);
```
lutter, line 2689: `upsert_block` is only used in 2 places (apart from tests) It should be possible to tighten the type of `block` to a concrete type here so we don't have to go through JSON. It's kinda low priority since this code path is only used on the block ingestor, which isn't that performance critical.

chain/ethereum/src/chain.rs
incrypto32: In the next commit it was moved to the `graph` crate, hence keeping it as it is.

graph/src/components/ethereum/types.rs
incrypto32: Thanks! addressed this

graph/src/components/ethereum/types.rs
incrypto32: Noted

graph/src/components/store/traits.rs
incrypto32: Noted

store/postgres/src/chain_store.rs
incrypto32: Noted. will write up an issue for a follow up pR

PR #6418, opened 2026-03-04
docker: Add profiling support to graph-node-debug image

incrypto32:
- Build with frame pointers for perf stack unwinding
- Rebase `graph-node-debug` on slim runtime, add `linux-perf`
- Remove unused objcopy debug link dance

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -22,34 +22,24 @@ ARG TAG_NAME=unknown
 
 ADD . /graph-node
 
-RUN apt-get update \
-    && apt-get install -y cmake protobuf-compiler && \
-    cd /graph-node && \
-    RUSTFLAGS="-g" cargo build --release --package graph-node \
+RUN apt-get update && apt-get install -y cmake protobuf-compiler
+
+RUN cd /graph-node \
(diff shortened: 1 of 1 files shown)


docker/Dockerfile
```diff
@@ -107,3 +96,12 @@ COPY --from=envsubst /go/bin/envsubst /usr/local/bin/
...
 
+# Debug and profiling image
+# Build with: docker build --target graph-node-debug -f docker/Dockerfile .
+FROM graph-node AS graph-node-debug
+COPY --from=graph-node-build /usr/local/bin/graph-node-debug /usr/local/bin/graph-node
+COPY --from=graph-node-build /usr/local/bin/graphman-debug /usr/local/bin/graphman
+RUN apt-get update \
+    && apt-get install -y curl gdb postgresql-client linux-perf procps binutils
+COPY docker/Dockerfile /Dockerfile
+COPY docker/bin/* /usr/local/bin/
```
lutter, line 107: Maybe we should also copy `/etc/image-info` to help identify what exact `graph-node` version is in the image

incrypto32: Its already inherited from the `graph-node` base stage

PR #6420, opened 2026-03-04
Clean up vid generation

lutter:
The way `vid`'s were created was very brittle and relied on passing a counter around in the right places. We encapsulate this logic in a `SeqGenerator` that can be shared freely for the duration of a block and makes sure that generated `vid`s are unique

@incrypto32 I am asking for your review mostly for the first commit which affects vid generation proper
@isum I am asking for your review mostly for the second commit which changes amp subgraphs

### DIFF
--- a/store/test-store/tests/graph/entity_cache.rs
+++ b/store/test-store/tests/graph/entity_cache.rs
@@ -463,6 +463,37 @@ async fn offchain_trigger_vid_no_collision_with_shared_generator() {
     );
 }
 
+// Simulate the ipfs.map() pattern: multiple EntityCache instances each create
+// an entity using a shared SeqGenerator. VIDs must be unique and sequential.
+#[graph::test]
+async fn ipfs_map_pattern_vid_uniqueness() {
+    let block: i32 = 42;
+    let vid_gen = SeqGenerator::new(block);
+
+    let mut all_vids = Vec::new();
+    for i in 0..5u32 {
+        let store = Arc::new(MockStore::new(BTreeMap::new()));
+        let mut cache = EntityCache::new(store, vid_gen.cheap_clone());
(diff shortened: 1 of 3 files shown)


graph/src/components/store/entity_cache.rs
```diff
@@ -152,19 +190,26 @@ impl EntityCache {
         self.schema.make_entity(iter)
     }
 
-    pub fn with_current(store: Arc<dyn s::ReadStore>, current: EntityLfuCache) -> EntityCache {
+    pub fn with_current(
+        store: Arc<dyn s::ReadStore>,
+        current: EntityLfuCache,
+        vid_gen: SeqGenerator,
```
incrypto32: Minor:
Maybe the it should be `seq_gen`

graph/src/components/subgraph/instance.rs
```diff
@@ -100,6 +100,10 @@ impl BlockState {
             write_capacity_remaining: ENV_VARS.block_write_capacity,
         }
     }
+
+    pub fn vid_gen(&self) -> SeqGenerator {
```
incrypto32: Naming inconsitency here too`vid_gen` -> `seq_gen`

graph/src/components/store/entity_cache.rs
```diff
@@ -152,19 +190,26 @@ impl EntityCache {
...
             in_handler: false,
             schema: store.input_schema(),
             store,
-            seq: 0,
-            vid_seq: RESERVED_VIDS,
+            seq_gen: vid_gen,
         }
     }
 
+    pub fn vid_gen(&self) -> SeqGenerator {
```
incrypto32: Here too

incrypto32 (approved):
Nice refactor, just some minor comments

PR #6422, opened 2026-03-06
core: Restart store on revert errors to clear poisoned writer

incrypto32:
#6298 mapped `Action::Restart` from `handle_revert` to `RestartReason::DataSourceExpired`, which skips `store.restart()`. This meant a poisoned writer from a revert error would never be cleared, causing the subgraph to loop indefinitely until process restart.

- Use `RestartReason::StoreError` for revert errors so `store.restart()` is always called (it's a no-op when not poisoned)
- Remove the unused `DataSourceExpired` variant from `RestartReason`

### DIFF
--- a/core/src/subgraph/runner/state.rs
+++ b/core/src/subgraph/runner/state.rs
@@ -74,8 +74,6 @@ pub enum RunnerState<C: Blockchain> {
 pub enum RestartReason {
     /// New dynamic data source was created that requires filter updates.
     DynamicDataSourceCreated,
-    /// A data source reached its end block.
-    DataSourceExpired,
     /// Store error occurred and store needs to be restarted.
     StoreError,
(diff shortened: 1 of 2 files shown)


core/src/subgraph/runner/state.rs
```diff
@@ -74,8 +74,6 @@ pub enum RunnerState<C: Blockchain> {
 pub enum RestartReason {
     /// New dynamic data source was created that requires filter updates.
     DynamicDataSourceCreated,
-    /// A data source reached its end block.
-    DataSourceExpired,
```
lutter, line 78: I assume this is unused since we already have to handle the end block of data sources by just ignoring any matching triggers.

incrypto32: right

PR #6423, opened 2026-03-06
core: Remove error return from IndexingContext.revert_data_sources

lutter:
The unnecessary return makes callers look like they could fail

Just a small code clarification as I was looking at #6422


lutter:
I screwed up some git commands. This was merged as 4db386b

Issue #6213, opened 2025-11-11
Block ingestor retries failing provider indefinitely


dimitrovmaksim:
When multiple network providers are configured, the block ingestor will pick a provider at start up, if at some point the provider fails, it will retry it indefinetely, instead of falling over to another configured provider.

Issue #5575, opened 2024-07-30
Smarter provider choice and fallback between providers


fordN:
- Ensure subgraphs are assigned to a provider that supports its features (if available)
 - Support switching providers if issues with currently assigned provider

Issue #5313, opened 2024-03-29
[Bug] graph-node stops retrying when block ingestion stops due to RPC issues


Shiyasmohd:
### Bug report

When block ingestion stops due to RPC issues, graph-node stops retrying block ingestion.
Even after an RPC issue is resolved graph node block ingestor doesn’t recover on itself without restarting the ingestor.

### Some information to help us out

- [ ] Tick this box if this bug is caused by a regression found in the latest release.
- [ ] Tick this box if this bug is specific to the hosted service.
- [X] I have searched the issue tracker to make sure this issue is not a duplicate.

### OS information

None

azf20:
thanks @Shiyasmohd! @mangas @leoyvens is this expected, or should we already be retrying in this case?

rlgns98kr:
Is there any progess of this issue ?
I'm a operator of graph-node. I think it have to resolve...
I always restart graph-node manually. And sometime if I missed to check the graph node, there is some downtime to our services.... 🙇🏼‍♂️

azf20:
@mangas do the lazy provider checks help here? #5337

PR #6430, opened 2026-03-10
Add RPC provider failover for block ingestor

dimitrovmaksim:
Resolves https://github.com/graphprotocol/graph-node/issues/6213
Related open issues: https://github.com/graphprotocol/graph-node/issues/5313, https://github.com/graphprotocol/graph-node/issues/5575

When the block ingestor's polling fails and the current RPC provider is unreachable, the ingestor now automatically switches to a healthy alternative provider. If no alternative providers are configured, the current one will be retried indefinetelly, similar to the current logic.

#### How it works

  1. On do_poll() failure, the current provider is probed via eth_blockNumber first — do_poll() can fail for non-RPC reasons (DB errors, chain reorgs), so switching would not help in those cases.
  2. If the current provider is unreachable, all other validated providers are probed in parallel. The first to respond is selected.
  3. If all providers are unreachable, the ingestor stays on the current provider and re-probes on the next failure.
  4. There is no automatic return to the original provider — the ingestor stays on whatever provider it switched to until that one fails.

 #### Other changes

  - latest_block_ptr retries are now limited to ENV_VARS.request_retries instead of infinite, so failures surface to the failover logic.
  - retry_strategy uses saturating_sub to avoid underflow when limit is 0.
  - Added is_reachable() to the EthereumAdapter trait for lightweight provider health checks.
  - Added all_cheapest() to EthereumNetworkAdapters to expose all validated providers.

### DIFF
--- a/chain/ethereum/src/ingestor.rs
+++ b/chain/ethereum/src/ingestor.rs
@@ -223,8 +223,7 @@ impl PollingBlockIngestor {
         current_provider: &mut Option<String>,
     ) {
         // Resolve by name; resets to first provider if the tracked one left the list.
-        let idx = resolve_provider_idx(providers, current_provider, &self.logger);
-        let eth_adapter = providers[idx].clone();
+        let eth_adapter = resolve_provider(providers, current_provider, &self.logger).clone();
         // Pin the name so the next iteration knows which provider is active.
         let provider_name = eth_adapter.provider().to_string();
         let logger = self.logger.new(o!("provider" => provider_name.clone()));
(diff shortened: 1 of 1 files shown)


DaMandal0rian:
@dimitrovmaksim seems related to https://github.com/graphprotocol/graph-node/pull/6126 ?

dimitrovmaksim:
> @dimitrovmaksim seems related to #6126 ?

@DaMandal0rian yeah, I noticed this PR today while going through the issues/PRs and started wondering if they are not conflicting (i haven't checked it myself). I'm okay with this one being closed, if the other one handles this issue and gets approved/merged

PR #6431, opened 2026-03-10
Reduce clones and cleanup in block cache path

incrypto32:
Addresses review feedback from #6400.

- Migrate `check_blocks` to use typed `blocks()` instead of raw JSON
- Remove `blocks_as_json` from `ChainStore` trait
- Wrap `LightEthereumBlock` in `Arc` inside `CachedBlock::Light` to avoid expensive deep clones
- Inline `EthereumJsonBlock` into `CachedBlock::from_json` and delete the wrapper

### DIFF
--- a/chain/ethereum/src/ethereum_adapter.rs
+++ b/chain/ethereum/src/ethereum_adapter.rs
@@ -1639,8 +1639,8 @@ impl EthereumAdapterTrait for EthereumAdapter {
             .await
             .map_err(|e| error!(&logger, "Error accessing block cache {}", e))
             .unwrap_or_default()
-            .into_iter()
-            .map(|b| b.into_light_block())
+            .iter()
+            .map(|b| b.to_light_block())
             .collect();
 
         let missing_blocks = Vec::from_iter(
--- a/graph/src/components/ethereum/types.rs
+++ b/graph/src/components/ethereum/types.rs
@@ -278,7 +278,7 @@ impl CachedBlock {
         }
     }
 
-    pub fn into_light_block(&self) -> Arc<LightEthereumBlock> {
+    pub fn to_light_block(&self) -> Arc<LightEthereumBlock> {
         match self {
             CachedBlock::Full(block) => block.block.clone(),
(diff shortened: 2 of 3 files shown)


server/index-node/src/resolver.rs
```diff
@@ -237,7 +236,24 @@ where
             }
             Ok(mut blocks) => {
                 assert!(blocks.len() == 1, "Multiple blocks with the same hash");
-                blocks.pop().unwrap().into()
+                let block = blocks.pop().unwrap();
+                let json = match block {
+                    CachedBlock::Full(ref b) => serde_json::to_value(b),
+                    CachedBlock::Light(ref b) => serde_json::to_value(b),
+                };
```
lutter: This could be a `to_json` method on `CachedBlock`

chain/ethereum/src/ethereum_adapter.rs
```diff
@@ -1640,7 +1640,7 @@ impl EthereumAdapterTrait for EthereumAdapter {
             .map_err(|e| error!(&logger, "Error accessing block cache {}", e))
             .unwrap_or_default()
             .into_iter()
-            .map(CachedBlock::into_light_block)
+            .map(|b| b.into_light_block())
```
lutter: This seems like a noop, but I assume it's because `b` isn't a reference here. Maybe changing `into_iter()` to `iter()` would be better to make it clear that some copying is going on

graph/src/components/ethereum/types.rs
```diff
@@ -279,10 +278,10 @@ impl CachedBlock {
         }
     }
 
-    pub fn into_light_block(self) -> Arc<LightEthereumBlock> {
+    pub fn into_light_block(&self) -> Arc<LightEthereumBlock> {
```
lutter: Very minor, but in Rust, `into_` methods usually take ownership; `to_` methods usually operate on references.

lutter (approved):
Nice! Just a few very minor comments, feel free to ignore

PR #6434, opened 2026-03-11
store: Consolidate index creation using CreateIndex and add postponed index support

lutter:
Currently, `GRAPH_POSTPONE_ATTRIBUTE_INDEX_CREATION` when copying subgraphs; but subgraphs whose performance is limited by the speed at which we can write to the database, such as amp subgraphs, will also benefit from deferring index creation.

Besides postponing index creation for syncing subgraphs, this PR also refactors how indexes are created for subgraph tables, replacing scattered raw SQL generation with a unified `CreateIndex` abstraction.

This PR will also be the basis for speeding up `graphman restore` by having it defer index creation until after the data import.

  ### Index creation consolidation
  - Introduce `Table::indexes()` returning all indexes (time-travel, attribute, aggregate) as structured `CreateIndex` objects instead of raw SQL strings
  - Parse and round-trip our own index definitions including BRIN indexes with `minmax_multi_ops` operator classes and various  `WHERE` clauses
  - Remove the `index_def: Option<IndexList>` parameter threading and simplify callers across `copy.rs`, `prune.rs`, and `deployment_store.rs`
  - Add a `create_index` example tool for testing index definition parsing

  ### Postponed index creation
  - Allow deferring index creation during initial sync via `CreateIndex::to_postpone()`, controlled by the `GRAPH_POSTPONE_INDEXES` env var
  - Trigger creation of postponed indexes when a subgraph gets within a configurable number of blocks of the chain head (`GRAPH_POSTPONE_INDEXES_CREATION_THRESHOLD`, default 10000)
  - Re-create any missing postponed indexes on subgraph restart as a safety net, using `IF NOT EXISTS` + `CONCURRENTLY`

  ## Test plan
  - [X] Unit tests pass (`just test-unit`)
  - [X] DDL test constants updated to match new single-line index format
  - [ ] Verify postponed index creation triggers correctly near chain head
  - [ ] Verify indexes are recreated on subgraph restart

### DIFF
--- a/core/src/subgraph/runner/mod.rs
+++ b/core/src/subgraph/runner/mod.rs
@@ -109,6 +109,7 @@ where
                 ),
                 entity_lfu_cache: LfuCache::new(),
                 cached_head_ptr: None,
+                postponed_indexes_created: false,
             },
             logger,
             metrics,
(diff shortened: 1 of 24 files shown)


isum (approved):
This is a nice improvement!

_One note:_
- Amp subgraphs have a separate runner so it should also be updated to trigger the postponed index creation when a subgraph reaches the chain head

PR #6494, opened 2026-04-12
graph: Remove debug prefix from Timestamp value parse errors

cuiweixie:
## Summary

Removes an accidental `"xxx"` debug prefix from the `TIMESTAMP_SCALAR` branch of `Value::from_graphql_value`, so `QueryExecutionError::ValueParseError` reports the invalid timestamp string directly (consistent with other scalars).

## Testing

Please run `just lint`, `just check --release`, and relevant unit tests locally before merge.


lutter:
The PR fails CI. Pushed commit 757ff94b which addresses the same thing correctly

PR #6495, opened 2026-04-12
Handle missing output field in Ethereum trace results

JunaidCD:
## Problem

Graph Node fails to deserialize traces when the `result.output` field is missing.

This occurs for certain trace types (e.g. `suicide`) where the RPC response legitimately omits the `output` field, leading to errors like:

data did not match any variant of untagged enum TraceOutput

## Cause

After switching to alloy, the `output` field is treated as required during deserialization, while some valid traces do not include it.

## Solution

Before deserialization, a default value ("0x") is inserted when the `output` field is missing.

## Scope

- Localized change in `ethereum_adapter.rs`
- No changes to external dependencies

## Impact

Prevents indexing failures for chains that return traces without `output`, restoring compatibility with existing RPC behavior.


lutter:
Why does this PR add protoc.exe, various proto files, and an error.log? Please reduce this to the minimum necessary to address the problem

lutter:
Closing this for now; feel free to reopen once you've removed the extraneous files and rebased the PR

PR #6502, opened 2026-04-14
store: Clamp adaptive batch size for stale cleanup

mk0walsk:
## Summary

Clamp adaptive batch sizes so stale call cache cleanup never degrades to a zero-sized batch.

## Changes

- clamp `AdaptiveBatchSize` to a minimum of 1 when it is created, adapted, or set explicitly
- guard `clear_stale_call_cache` before issuing each delete batch so the loop always makes progress
- add a unit test covering the zero-sized batch regression path

### DIFF
--- a/store/postgres/src/chain_store.rs
+++ b/store/postgres/src/chain_store.rs
@@ -3319,7 +3319,7 @@ impl ChainStoreTrait for ChainStore {
         let mut last_log = Instant::now();
 
         loop {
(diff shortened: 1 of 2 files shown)


lutter:
Is this something you encountered in practice? I'd be curious to hear how that happened.

Also, I don't think we need all that `clamp_size` business in most places; it's fine to assume that constructors etc. set the batch size to something sane

PR #6520, opened 2026-04-17
graph: Remove manual Eq::assert_receiver_is_total_eq impl

incrypto32:
Manually implementing `Eq::assert_receiver_is_total_eq` is now a hard error in recent Rust toolchains. Remove the manual impl and use the empty impl block instead.

This is currently breaking CI on master and all open PRs.

### DIFF
--- a/graph/src/util/intern.rs
+++ b/graph/src/util/intern.rs
@@ -510,9 +510,7 @@ impl<V: PartialEq> PartialEq for Object<V> {
     }
 }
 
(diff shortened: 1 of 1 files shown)


incrypto32:
Closing — this is addressed by #6517 (Rust 2024 edition update) which fixes both the `assert_receiver_is_total_eq` hard error and the `queries overflow the depth limit` issue on Rust 1.95.

PR #6548, opened 2026-04-30
Decide load_related membership against final state

incrypto32:
`EntityCache::load_related` decided derived field membership against intermediate
cache layers (`self.updates` and `self.handler_updates` separately), so same-block
changes to an entity's parent reference could leave it in the wrong derived
collection or omit it from one it had moved into.

Rewrite `load_related` to:

- Build a candidate set from the store baseline plus any in-block writes whose
  key is compatible with the query (entity_type + causality_region).
- Resolve each candidate's final state by layering store → updates →
  handler_updates the same way `EntityCache::get()` does.
- Apply `query.matches` exactly once against the final state.

Closes three previously-distinct failure modes: entities wrongly retained after
moving out of a collection, wrongly retained after a cross-handler revert, and
wrongly omitted after being moved in across handlers.

## Scenarios covered by the new tests

- Single-handler reassignment removes entity from old parent's collection
- Cross-handler reassign-then-revert leaves intermediate parent's collection unchanged
- Cross-handler touch-then-reassign adds entity to new parent's collection
- Removal via `self.updates` drops entity from collection
- Removal via `self.handler_updates` drops entity from collection
- Fresh in-handler creation appears in new parent's collection

All six tests are in `store/test-store/tests/graph/entity_cache.rs`.

### DIFF
--- a/graph/src/components/store/entity_cache.rs
+++ b/graph/src/components/store/entity_cache.rs
@@ -306,12 +306,12 @@ impl EntityCache {
         // region are compatible with the query. The latter catches
         // entities an in-block write has moved into the matching set or
         // created fresh in this block.
-        let mut candidates: BTreeSet<EntityKey> = stored.keys().cloned().collect();
+        let mut candidates: BTreeSet<&EntityKey> = stored.keys().collect();
         for key in self.updates.keys().chain(self.handler_updates.keys()) {
             if key.entity_type == query.entity_type
                 && key.causality_region == query.causality_region
             {
-                candidates.insert(key.clone());
+                candidates.insert(key);
             }
         }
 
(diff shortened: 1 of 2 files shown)


graph/src/components/store/entity_cache.rs
```diff
@@ -291,114 +291,52 @@ impl EntityCache {
...
-            } else {
-                // if entity_arc is None, it means that the entity was removed by an update
-                // mark the key for removal from the map
-                keys_to_remove.push(key.clone());
+        // Candidate set: keys that were matching at baseline, plus keys
+        // any in-block write has touched whose entity_type and causality
+        // region are compatible with the query. The latter catches
+        // entities an in-block write has moved into the matching set or
+        // created fresh in this block.
+        let mut candidates: BTreeSet<EntityKey> = stored.keys().cloned().collect();
```
lutter: Minor nit: would this also work with references so we don't have to clone the keys?

lutter (approved):
Nice! Subtle bug.

graph/src/components/store/entity_cache.rs
incrypto32: done made the change

PR #6558, opened 2026-05-05
chain/ethereum: OR-merge receipt requirements in EthereumLogFilter

incrypto32:
## Summary

`EthereumLogFilter` merged per-handler `receipt: bool` flags using overwrite semantics, so when two handlers collided on the same filter key with different `receipt` flags, a later `false` would silently drop a prior `true`. Handlers that declared `receipt: true` then ran with `event.receipt == null`.

Wrap the three internal collections in `MergeMap` / `MergeGraph` newtypes whose only mutators OR-combine values, and update `from_data_sources`, `from_mapping`, and `extend` to use them.

Triggers when a manifest has two handlers at the same effective filter key (e.g. same `(contract, event)`) with different `receipt` flags.

### DIFF
--- a/chain/ethereum/src/adapter.rs
+++ b/chain/ethereum/src/adapter.rs
@@ -1849,7 +1849,7 @@ fn log_filter_require_transacion_receipt_method() {
     .into_iter()
     .collect();
 
-    let events_with_topic_filters = HashMap::new(); // TODO(krishna): Test events with topic filters
+    let events_with_topic_filters = HashMap::new();
 
     let alien_event_signature = b256(8); // those will not be inserted in the graph
     let alien_contract_address = address(9);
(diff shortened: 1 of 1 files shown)


chain/ethereum/src/adapter.rs
```diff
@@ -1877,25 +1964,29 @@ fn log_filter_require_transacion_receipt_method() {
...
-    contracts_and_events_graph.add_edge(event_a_id, contract_b_id, false);
-    contracts_and_events_graph.add_edge(event_b_id, contract_a_id, false);
-    contracts_and_events_graph.add_edge(event_c_id, contract_c_id, true);
-
-    let filter = EthereumLogFilter {
-        contracts_and_events_graph,
-        wildcard_events,
-        events_with_topic_filters,
-    };
+    // TODO(krishna): Test events with topic filters
```
lutter: Is this TODO still needed?

lutter (approved):
That's been a very longstanding bug then, right? Looks good though

chain/ethereum/src/adapter.rs
incrypto32: no, just removed

PR #6610, opened 2026-06-01
graph, store: Reject grafts below the base's earliest block

### LINKED ISSUE #6581: Grafting onto a pruned base silently loses data
## Bug

`Graft::validate` (`graph/src/data/subgraph/mod.rs:549`) checks that the graft block is below `base.latest_block`, above the reorg threshold, and that the base is healthy. It does not check that the graft block is at or above `base.earliest_block_number`.

When the base has pruned past the graft block, the copy step's filter (`WHERE lower(block_range) <= target_block`) finds no rows for mutable entities whose `[lo, hi)` versions were deleted by pruning. The destination is silently created missing all pre-graft mutable state. Immutable entities are unaffected — their `[lo, ∞)` ranges are never pruned.

The destination reports `synced: true, healthy: true` and returns wrong data with no error.

## Scope

Affects any subgraph with:

- `indexerHints.prune` enabled on the base
- A `graft.block` older than `base.earliest_block_number` at deploy time
- At least one mutable entity updated between the graft block and the prune horizon

### LINKED ISSUE #6607: [Bug] Grafting from a Pruned Source at a Block Below `earliest_block_number` Silently Produces Incomplete State
## Summary

When a subgraph that has been continuously pruned is used as a graft base, and the `graft.block` specified in the manifest falls **below** the source's `earliest_block_number`, graph-node accepts the graft without any warning or error. The resulting subgraph silently inherits an inconsistent state: `data_sources$` rows survive pruning (they are never deleted), but the entity data they reference has been removed. The grafted subgraph then fails with `unexpected null in handler` on every event that touches an entity whose row no longer exists in the pruned source.

---

## Background

Pruning removes historical entity rows older than `history_blocks` blocks from a deployment's store tables. However, `data_sources$` rows — which record which dynamic contract addresses the subgraph is watching — **are never pruned**. This is intentional: deleting them would cause the subgraph to stop listening to those contracts.

This creates a structural gap after sustained pruning: `data_sources$` may contain entries for contracts whose corresponding entity rows have been deleted. As long as the subgraph continues indexing forward from its pruned state this is harmless, since the live entities exist for recent blocks. But if a new subgraph is grafted from this source at a block *below* `earliest_block_number`, the grafted copy inherits the orphaned `data_sources$` references with no entity rows to back them.

Graft validation does not check whether the requested graft block is at or above the source's `earliest_block_number`. The graft is accepted silently.

By contrast, `graphman copy` was updated in PR #6384 (Feb 2026) to reject copies where the offset block falls below the source's `earliest_block_number`. The manifest graft path has no equivalent guard.

---

### Prerequisites

- A subgraph using a factory-contract / dynamic data sources pattern (e.g. Uniswap V3 style: a factory deploys pool contracts, each pool becomes a dynamic data source)
- The source subgraph has been running for an extended period with continuous pruning configured (`history_blocks: N`)

### Steps

1. **Deploy and sync the source subgraph with pruning enabled.** After extended operation, `earliest_block_number` advances to a block significantly above the subgraph's genesis. All factory-created entity rows (pools, tokens) from early in the subgraph's history are deleted from the store tables; their `data_sources$` rows remain.

2. **Create a grafted subgraph at a block below `earliest_block_number`:**
   ```yaml
   graft:
     base: <source-deployment-hash>
     block: <block-number-below-source-earliest_block_number>
   ```
   graph-node accepts the manifest. No error or warning is emitted at deploy time or at startup.

3. **The grafted subgraph begins indexing** from the graft block. It immediately encounters events on factory-created contracts whose entity rows were pruned from the source. Every such event produces a deterministic `unexpected null in handler` error, because the expected entity (e.g. a `Pool`) does not exist.

4. **Errors accumulate silently as non-fatal.** The subgraph remains `unhealthy` indefinitely, producing incorrect or missing query results for all affected entities with no automated recovery path.

---

### Reference case

- Chain: **Scroll**
- Source subgraph (grafted from): `QmR6VP1qoF8nxhtMaGEg9VMmkaTDFqyeM8nJWkakP6nHes`
- Grafted subgraph (affected): `QmNqxqVfBETuMhV91BfquULGBLFvPZwr3ADVFTMgGZcqNf`
- Graft block: **14,146,768**
- Source `earliest_block_number`: **33,660,708**
- The graft block sits ~19.5 million blocks below the source's prune boundary

### Error log (representative entries)

```
ERRO Subgraph error 1/1,
  code: SubgraphSyncingFailure,
  error: transaction b459efa6523aead5fe69fc50945b2c10137f7555f3dba0b6a57d725e9e951c30:
    error while executing at wasm backtrace:
      0: 0x8aaf - <unknown>!src/mappings/core/handleSwap:
    Mapping aborted at src/mappings/core.ts, line 263, column 16,
    with message: unexpected null in handler `handleSwap`
  block_hash: 0x106299e06bbce9ce2b214cdaf46373c8f212f1b4f2d8893733923b4fc1957c01,
  block_number: 14252708, handler: handleSwap, deterministic: true,
  sgd: 1296, subgraph_id: QmNqxqVfBETuMhV91BfquULGBLFvPZwr3ADVFTMgGZcqNf,
  component: SubgraphInstanceManager
```

```
ERRO Subgraph error 1/1,
  code: SubgraphSyncingFailure,
  error: transaction d6b53305b97f91b336281727ac4aabc85dc66ad3a002087f81b093ed26d84f94:
    Mapping aborted at src/mappings/core.ts, line 263, column 16,
    with message: unexpected null in handler `handleSwap`
  block_number: 14252759, deterministic: true,
  sgd: 1296, subgraph_id: QmNqxqVfBETuMhV91BfquULGBLFvPZwr3ADVFTMgGZcqNf
```

```
ERRO Subgraph error 1/1,
  code: SubgraphSyncingFailure,
  error: transaction bda5e0f0830f1daec60ec6c2c82598ab1403ae97788391edbc46e99a7d5f04e2:
    Mapping aborted at src/mappings/core.ts, line 263, column 16,
    with message: unexpected null in handler `handleSwap`
  block_number: 14253040, deterministic: true,
  sgd: 1296, subgraph_id: QmNqxqVfBETuMhV91BfquULGBLFvPZwr3ADVFTMgGZcqNf
```

---

## Impact

- **Silent data corruption.** No warning, no validation failure, no startup log. There is no signal to the operator that the graft will produce missing entities.
- **Permanent `unhealthy` state.** Every event on a pruned entity generates a deterministic non-fatal error. These errors cannot be cleared without reconstructing the missing rows from the underlying chain.
- **Incorrect query results.** Queries over entities whose rows were not copied return null or incomplete data.

---

## Similar Issue
https://github.com/graphprotocol/graph-node/issues/6581

incrypto32:
`Graft::validate` checks that the graft block is below `base.latest_block`, above the reorg threshold, and that the base is healthy. It does not check that the graft block is at or above `base.earliest_block_number`.

When the base has been pruned past the graft block (or its start block is above it), the copy step silently leaves the destination missing all pre-graft mutable entity versions. Dynamic data source registrations in `data_sources$` survive (they are never pruned), so the destination keeps listening to factory-created contracts whose backing entities are gone, producing `unexpected null in handler` errors on every event.

This adds the missing check, mirroring the equivalent guard added for `graphman copy` in #6384.

Closes #6581
Closes #6607

### DIFF
--- a/graph/src/components/store/traits.rs
+++ b/graph/src/components/store/traits.rs
@@ -219,6 +219,12 @@ pub trait SubgraphStore: Send + Sync + 'static {
     /// being set up
     async fn least_block_ptr(&self, id: &DeploymentHash) -> Result<Option<BlockPtr>, StoreError>;
 
+    /// Return the earliest block for which the deployment with the given
+    /// `id` still has entity data. Blocks below this are either pre-`startBlock`
+    /// or have been removed by pruning, and cannot serve as a graft or copy
+    /// source.
+    async fn earliest_block_number(&self, id: &DeploymentHash) -> Result<BlockNumber, StoreError>;
+
     async fn is_healthy(&self, id: &DeploymentHash) -> Result<bool, StoreError>;
 
     /// Find all deployment locators for the subgraph with the given hash.
--- a/graph/src/data/subgraph/mod.rs
+++ b/graph/src/data/subgraph/mod.rs
@@ -600,7 +600,24 @@ impl Graft {
                 self.block,
                 ptr.number - 1
             ))),
-            (Some(_), _) => Ok(()),
+            (Some(_), _) => {
+                // The graft block must be at or above the base's earliest
+                // available block. Below that the base store no longer has
+                // the entity versions the copy step would need: either the
+                // base started above the graft block, or pruning has
+                // removed the history.
+                let earliest_block = store
+                    .earliest_block_number(&self.base)
+                    .await
+                    .map_err(|e| GraftBaseInvalid(e.to_string()))?;
+                if self.block < earliest_block {
+                    return Err(GraftBaseInvalid(format!(
+                        "failed to graft onto `{}` at block {} since its earliest available block is {}",
+                        self.base, self.block, earliest_block
+                    )));
+                }
+                Ok(())
+            }
         }
     }
 }
--- a/store/postgres/src/subgraph_store.rs
+++ b/store/postgres/src/subgraph_store.rs
@@ -1888,6 +1888,12 @@ impl SubgraphStoreTrait for SubgraphStore {
         store.block_ptr(site.cheap_clone()).await
     }
 
+    async fn earliest_block_number(&self, id: &DeploymentHash) -> Result<BlockNumber, StoreError> {
+        let (store, site) = self.store(id).await?;
+        let state = store.deployment_state(site.cheap_clone()).await?;
+        Ok(state.earliest_block_number)
+    }
+
     async fn is_healthy(&self, id: &DeploymentHash) -> Result<bool, StoreError> {
         let (store, site) = self.store(id).await?;
         let health = store.health(&site).await?;


0x19dG87:
@incrypto32 Why it should reject? Should not subgraph that uses graft follow graft structure and data and therefore just copy graft data and start from the latest graft block?

incrypto32:
@0x19dG87  My reasoning was that `graft.block` today means "the block the new deployment starts at, with all entity state up to and including that block inherited from the base."

If `graft.block < earliest_block_number`, the copy step matches nothing for mutable entities, since pruning deleted them. The destination starts without the inherited state the manifest implies it has, and mappings hit null for entities they expect (see #6607).

Also there is implications for POI: two indexers grafting the same deployment would diverge depending on whether they grafted before or after the base was pruned. The one with the full copy reads real pre-graft state into handlers; the one with the empty copy reads null and writes different entities based on how mappings handle it. 

Happy to discuss this further if there's uncertainities. Yet to be discussed internally and waiting for a review too from the team.

lutter (approved):

PR #6614, opened 2026-06-03
graph: Re-enable postponed index creation in debug builds

incrypto32:
Follow-up to v0.44.0 (#6611): the disable patch hardcoded
`postpone_attribute_index_creation` to `false`, which broke CI in
`relational::ddl_tests` (`generate_ddl`, `generate_postponed_indexes`,
`postponed_indexes_with_block_column`).

Switching to `cfg!(debug_assertions)` keeps the production behavior
(release builds: postponement off, env var inert) and re-enables it
for debug builds so the tests pass.


lutter:
I reenabled that as part of #6608

Issue #6360, opened 2026-02-13
gnd: Improvements thread


dimitrovmaksim:
I'll add issues that I find as different comments, so I don't flood the repo

dimitrovmaksim:
There are several places where the manifest is loaded directly with `serde_yaml` instead through the `load_manifest()` function from `manifest.rs`.
Some of them may be intentional, but maybe worth checking.

```sh
  add.rs                                                                                                                                             
                                                                                                                                                     
  Line 84-85: Loading manifest to extract network           
  let manifest: serde_yaml::Value = serde_yaml::from_str(&manifest_content)
      .with_context(|| format!("Failed to parse manifest: {}", opt.manifest.display()))?;

  Line 372-373: In update_manifest function
  let mut manifest: serde_yaml::Value =
      serde_yaml::from_str(&content).context("Failed to parse manifest")?;

  build.rs

  Line 489: In create_ipfs_manifest function (no .context() but direct ?)
  let mut value: serde_yaml::Value = serde_yaml::from_str(&manifest_str)?;

  Line 751: In write_output_manifest function (no .context() but direct ?)
  let mut value: serde_yaml::Value = serde_yaml::from_str(&manifest_str)?;

  Line 895-896: In apply_network_to_manifest function
  let mut manifest: serde_yaml::Value = serde_yaml::from_str(&manifest_str)
      .with_context(|| format!("Failed to parse manifest: {}", manifest_path.display()))?;

  init.rs

  Line 908: In init_from_subgraph function
  let manifest: serde_yaml::Value =
      serde_yaml::from_str(&manifest_yaml).context("Failed to parse subgraph manifest YAML")?;
```

Some of the cases and also the `load_manifest` function uses `.with_context()`, but only the context is printed and the underlying error is silenced, because the error are printed with `{}`.
```sh
gnd build  
→ Load subgraph from subgraph.yaml
Error: Failed to parse manifest: "subgraph.yaml"
```

dimitrovmaksim:
This is probably more of `graph-node` related issue, because `gnd` uses the `graph-node` manifest validation, but the error are returned one-by-one (if there are multiple missing fields for example) which may annoy developers:
```sh
 $ gnd build
→ Load subgraph from subgraph.yaml
Error: Failed to parse manifest: "subgraph.yaml": parse error: missing field `file`: missing field `file`

$ graph build                              
✔ Apply migrations
[...]

PR #6622, opened 2026-06-10
fix(gnd): surface anyhow error chains in error output

mvanhorn:
## Summary
`gnd` error output now surfaces the full anyhow error chain instead of only the top-level message. The bare `{}` Display formattings in `main.rs` and `watch.rs` printed just the outermost context (for example `failed to load manifest`), hiding the underlying cause a developer needs to act on; they now use the alternate `{:#}` chain rendering. The two bare `serde_yaml::from_str()?` sites in `commands/build.rs` and `manifest.rs` gain context so YAML parse failures name the file being parsed.

## Why this matters
Issue #6360 reports that gnd swallows error causes, leaving developers with messages like `Error: failed to start watch mode` and no underlying reason. The silenced chains were verified at `main.rs:266` and `watch.rs:49`/`104` before the change. This covers the error-display half of the issue; the issue also discusses restructuring some error sites to attach more context at creation time, which is a larger refactor left for a follow-up.

## Testing
A new unit test asserts a manifest parse failure renders with its cause chain (file name + YAML error) rather than the bare context string. `cargo fmt --check` and `cargo check -p gnd` pass; the new manifest error-chain test passes.

Refs #6360

### DIFF
--- a/gnd/src/commands/build.rs
+++ b/gnd/src/commands/build.rs
@@ -486,7 +486,8 @@ fn create_ipfs_manifest(
     let manifest_str = fs::read_to_string(&manifest_path)
         .with_context(|| format!("Failed to read manifest: {}", manifest_path.display()))?;
 
-    let mut value: serde_yaml::Value = serde_yaml::from_str(&manifest_str)?;
+    let mut value: serde_yaml::Value = serde_yaml::from_str(&manifest_str)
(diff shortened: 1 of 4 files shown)


lutter:
Thanks for the contribution. Can you rebase this to latest master?

mvanhorn:
Rebased onto latest master in 06108c6 - clean rebase, no conflicts, same single commit.

PR #6646, opened 2026-06-25
store: split dump batches by byte size to avoid 2 GiB Arrow column overflow (#6609)

### LINKED ISSUE #6609: [Bug] Graphman dump/restore cant handle larger subgraphs
### Bug report

So we are testing/working with graphman dump and restore. It works perfectly fine for smaller subgraphs, but on the larger ones it fails.

### Relevant log output

```Shell
This is a known issue with arrow-array hitting a 2GB limit on byte array buffers. This typically happens when graphman (or graph-node) is trying to process a very large result set in one shot.
What's causing it:
Arrow's GenericBytesBuilder uses 32-bit offsets internally — once the accumulated byte data exceeds ~2GB, it overflows. This is a hard limit in the arrow-array crate when using i32 offsets (vs i64 for "large" variants).
Likely triggers in your case:
Running a graphman command that fetches a large dataset (entity counts, history, stats across many subgraphs/blocks)
A subgraph with very large string fields or a huge number of entities being exported/queried at once
Workarounds to try:
Filter/limit the query — if the command accepts filters, scope it down (e.g. specific deployment hash, block range, or shard)
```

### IPFS hash

_No response_

_No response_

- [x] Tick this box if this bug is caused by a regression found in the latest release.

### OS information

Linux

cargopete:
## Summary

Fixes #6609. `graphman dump` panics on large tables. Each fetched batch is turned
into a single Arrow `RecordBatch` whose string/binary columns use the 32-bit-offset
`Utf8`/`Binary` types, which overflow once a single column accumulates more than
~2 GiB. The `VidBatcher` sizes fetches by elapsed time with no byte ceiling, so on a
large table a string/binary column crosses the limit and Arrow panics on offset
overflow.

## Fix

Decouple the Arrow batch from the SQL fetch: split each fetched batch into
contiguous byte-bounded sub-slices (≤ 256 MiB each) and write one `RecordBatch`
per slice. Because a batch's total payload stays well under 2 GiB, no individual
column can reach the i32 offset limit, and Postgres' ~1 GiB per-field cap keeps even
a single oversized row safe. The on-disk Parquet format and the restore/reader path
are unchanged — a dump simply produces more row groups.

- `parquet/convert.rs`: `estimate_row_bytes` — conservative per-row payload estimate.
- `relational/dump.rs`: `MAX_RECORD_BATCH_BYTES` (256 MiB) + `byte_bounded_slices`;
  the entity-table dump loop writes one `RecordBatch` per byte-bounded slice.

The clamp loops (`vid` + `block_range_end`, fixed-width) and the tiny `data_sources$`
table are not subject to the 2 GiB limit and are left unchanged.

## Testing

- New unit test `byte_bounded_slices_split_and_preserve_rows` (packing, no-data-loss,
  per-slice cap, oversized-single-row).
- Existing parquet dump/restore module tests still pass (29).
- `cargo fmt`, `cargo clippy --all-targets`, and `cargo check --release` clean for
  `graph-store-postgres`.

⚠️ Not yet validated end-to-end against a real >2 GiB-per-column table — that requires
a large populated database. @mindstyle85 this is the case you hit; a test run against
your large subgraph would be much appreciated.

### DIFF
--- a/store/postgres/src/parquet/convert.rs
+++ b/store/postgres/src/parquet/convert.rs
@@ -58,6 +58,39 @@ pub fn rows_to_record_batch(schema: &Schema, rows: &[OidRow]) -> Result<RecordBa
         .map_err(|e| StoreError::InternalError(format!("failed to build RecordBatch: {e}")))
 }
 
+/// Conservative estimate of the number of payload bytes a row contributes to an
+/// Arrow `RecordBatch`. The dumper uses this to bound batch size so that no
+/// single 32-bit-offset string/binary column can exceed Arrow's 2 GiB offset
+/// limit while a large table is dumped (issue #6609). Variable-length payloads
+/// (`String`/`Bytes`) dominate that limit and are counted exactly; fixed-width
+/// values use their in-memory width.
+pub(crate) fn estimate_row_bytes(row: &OidRow) -> usize {
+    row.into_iter().map(estimate_value_bytes).sum()
+}
+
+fn estimate_value_bytes(value: &OidValue) -> usize {
+    match value {
+        OidValue::Null => 0,
+        OidValue::Bool(_) => 1,
+        OidValue::Int(_) => 4,
+        OidValue::Int8(_) => 8,
+        OidValue::Timestamp(_) => 8,
+        OidValue::Int4Range(_, _) => 8,
+        OidValue::Bytes(v) => v.as_ref().len(),
+        OidValue::String(v) => v.len(),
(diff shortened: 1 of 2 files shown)


mindstyle85:
hey, sure will test it out, we are still in the process of moving some subgraphs around, but we are using what Vincent made right now and it works fine 

once we are done with migrations we can try on the old db :)

lutter (approved):
Thanks for the fix

PR #6647, opened 2026-06-25
chain/ethereum: send eth_getBlockReceipts block hash as plain string param (#5835)

### LINKED ISSUE #5835: [Bug] Get Block Receipts not working in the latest version
### Bug report

Hello,

We just implemented the `eth_getBlockReceipts` RPC call in our node but we noticed it's not working properly in the latest version of `graph-node` (using graphprotocol/graph-node:v0.36.1 docker image).

In all implementations of the call the params look like this:

```
{"jsonrpc":"2.0","method":"eth_getBlockReceipts","params":["0xa4c3c1414373ed64c1b80d007d8edab0335193d87788ebfddc59c79a5feb1ec2"],"id":1}
```

But `graph-node` sends this request:
```
{"jsonrpc":"2.0","method":"eth_getBlockReceipts","params":[{"blockHash":"0xa4c3c1414373ed64c1b80d007d8edab0335193d87788ebfddc59c79a5feb1ec2"}],"id":1}
```
Guessing the problem is in the `rust-web3` fork where the BlockId is serialized directly to json instead of having separate cases for block number and hash - https://github.com/graphprotocol/rust-web3/blob/graph-patches-onto-0.18/src/api/eth.rs#L144

I see there's a change by @incrypto32 here https://github.com/graphprotocol/rust-web3/commit/b9ba7c72316761ae23e95d8d6be1135a5f9e9c01 and that's why I marked this as a regression.

Thanks,

### Relevant log output

```Shell
Graph node:

graph-node-1  | Feb 18 22:19:18.918 WARN Skipping use of block receipts, reason: Error fetching block receipts: RPC error: Error { code: InvalidParams, message: "INVALID_PARAMS: Invalid method parameters (invalid name and/or type) recognised: Type is not convertible to string", data: None }, provider: mainnet-rpc-0, component: EthereumPollingBlockIngestor

Node:
blockchain-taraxa-node-3 taraxa-node {"jsonrpc":"2.0","method":"eth_getBlockReceipts","params":[{"blockHash":"0xa4c3c1414373ed64c1b80d007d8edab0335193d87788ebfddc59c79a5feb1ec2"}],"id":5}
```

### IPFS hash

_No response_

_No response_

- [x] Tick this box if this bug is caused by a regression found in the latest release.

### OS information

Linux

cargopete:
## Summary

Fixes #5835. graph-node sends `eth_getBlockReceipts` with the block hash as an
EIP-1898 object param (`[{"blockHash":"0x.."}]`) rather than the plain string
form (`["0x.."]`). Some clients (e.g. taraxa-node) only accept the plain string
and reject the object with `INVALID_PARAMS`.

This happens because both block-receipts call sites go through alloy's typed
`Provider::get_block_receipts(BlockId)`, and `BlockId::from(block_hash)`
serializes a `BlockId::Hash` as `{"blockHash": ".."}`.

When the capability probe (`check_block_receipt_support`) fails on such a node,
graph-node caches "block receipts unsupported" and permanently falls back to one
`eth_getTransactionReceipt` per transaction — correct data, but materially slower
ingestion and far more RPC load.

## Fix

Add a small helper that issues a raw request with the hash as a plain string:

```rust
alloy.client().request("eth_getBlockReceipts", (block_hash,)) // B256 -> "0x.." -> ["0x.."]
```

Both call sites (`check_block_receipt_support` and `fetch_block_receipts_with_retry`)
now route through it. The plain-string form is accepted by all known
implementations, and remains valid for nodes that currently work (they accept
both forms).

## Testing

- New unit test `block_receipts_param_is_plain_hash_string` asserting the new
  param serializes to a string array `["0x.."]`, and pinning the old object form
  as a regression guard.
- Existing `ethereum_adapter` unit tests pass (incl. `test_check_block_receipts_support`).
- `cargo fmt`, `cargo clippy --all-targets`, and `cargo check --release` clean for
  `graph-chain-ethereum`.

## Trade-off

A node that accepts *only* the EIP-1898 object form and *not* the plain string
would regress. In practice this is the rarer (and arguably non-conformant) case —
the plain hash string is the canonical Geth form and, per the issue reporter, the
form "all implementations" accept. Flagging it explicitly for maintainer review.

### DIFF
--- a/chain/ethereum/src/ethereum_adapter.rs
+++ b/chain/ethereum/src/ethereum_adapter.rs
@@ -2728,30 +2728,6 @@ mod tests {
     use std::sync::Arc;
 
     #[test]
-    fn block_receipts_param_is_plain_hash_string() {
-        use graph::prelude::alloy::rpc::types::BlockId;
-
-        let hash = B256::repeat_byte(0xab);
-
-        // The fix: the block hash is sent as a bare string param, i.e. `["0x.."]`.
-        // This is the form all known clients accept (issue #5835).
-        let new_params: Value = serde_json::to_value((hash,)).unwrap();
-        let arr = new_params.as_array().expect("params must be an array");
-        assert_eq!(arr.len(), 1);
-        assert!(
-            arr[0].is_string(),
-            "param must be a plain hash string, got {new_params}"
-        );
-        assert_eq!(arr[0].as_str().unwrap(), format!("{hash:#x}"));
-
-        // Regression guard: alloy's typed call serializes the hash to an
-        // EIP-1898 object `[{"blockHash":".."}]`, which strict nodes reject.
-        let old_params: Value = serde_json::to_value((BlockId::from(hash),)).unwrap();
-        assert!(old_params[0].is_object());
-        assert!(old_params[0].get("blockHash").is_some());
-    }
-
-    #[test]
     fn parse_block_triggers_every_block() {
         let block = create_minimal_block_for_test(2, hash(2));
 
--- a/chain/ethereum/src/ethereum_adapter.rs
+++ b/chain/ethereum/src/ethereum_adapter.rs
@@ -2315,13 +2315,31 @@ async fn batch_get_transaction_receipts(
     Ok(results)
 }
 
+/// Fetch block receipts by hash, sending the block hash as a plain string
+/// param (`["0x.."]`) rather than alloy's default EIP-1898 object
+/// (`[{"blockHash":".."}]`).
+///
+/// `eth_getBlockReceipts` accepts both forms on most clients, but some (e.g.
+/// taraxa-node) only accept the plain-string hash and reject the object with
+/// `INVALID_PARAMS`. The string form is accepted by all known implementations,
+/// so we always use it; alloy's typed `get_block_receipts` would always emit
+/// the object form. See issue #5835.
+async fn get_block_receipts_by_hash(
+    alloy: &AlloyProvider,
+    block_hash: B256,
+) -> Result<Option<Vec<AnyTransactionReceiptBare>>, RpcError<TransportErrorKind>> {
+    alloy
+        .client()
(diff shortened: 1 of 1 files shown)


lutter:
Have you reported this upstream to alloy? I'd be curious what their take is

incrypto32 (commented):
Fix looks right.

nit: On the test it doesn’t actually call the helper, so it’d still pass even if the fix got reverted. It’s not really testing our code. I’d just drop it, test_check_block_receipts_support already covers this.

I’d prefer this fixed upstream in alloy rather than worked around here. Feels like an alloy-side fix. Did you get a chance to report it upstream? if they push back let’s just land this and rip it out once alloy ships the fix.

cargopete:
Thanks both.

@incrypto32 - dropped the serialization test; you're right that it pinned `serde_json`'s behaviour rather than our code path. `test_check_block_receipts_support` exercises the real path now (it routes through the new helper).

@lutter @incrypto32 - reported upstream: alloy-rs/alloy#4049, asking whether `get_block_receipts` should serialize a bare `BlockId::Hash` as a plain string for broader node compatibility, or whether they consider such nodes non-conformant.

Happy to treat this PR as a stopgap and rip it out once alloy ships a fix. If they decide it's the node's problem, this keeps the affected indexers on single-call block receipts in the meantime rather than degrading to per-transaction fetching.

cargopete:
Upstream fix is now open as a PR too: alloy-rs/alloy#4050 (introduces a `BlockReceiptsParams` wrapper that serializes a block-hash param as a plain string for `eth_getBlockReceipts`). If/when that lands and we bump alloy, this workaround can be reverted.

incrypto32:
Thanks @cargopete for digging into this and reporting it upstream. Seems alloy fix is now merged (alloy-rs/alloy#4052) and covers our exact call path, so we won’t need to carry this workaround. Going to close it out, we’ll pick the fix up for free on the next alloy bump. Tracking in #5835 until then.

PR #6648, opened 2026-06-25
store: skip chains in pool_size = 0 shards instead of panicking at startup (#6195)

### LINKED ISSUE #6195: [Bug] Node Startup Fails When Using pool_size = 0 for Unused Shard Connections
### Description

We are attempting to configure **multiple indexer nodes**, each dedicated to indexing a specific chain, within a multi-chain environment. To minimize resource consumption, we are utilizing the `pool_size = 0` configuration option (introduced in PR [#3513](https://github.com/graphprotocol/graph-node/pull/3513)) for any shard connections that are not relevant to the current index node.

The documentation states that the primary shard is shared between chains. When a node starts, it connects to the primary shard, which in turn seems to discover the existence of all configured shards.

The intention of setting `pool_size = 0` is to avoid establishing unnecessary Postgres connections for these unused shards. However, the node startup process appears to fail later when attempting to use or initialize a component (specifically the BlockStore) that expects a connection pool for every configured shard, even when its pool size is zero.

The log clearly shows the shard being ignored as expected, but a later component fails because a pool for the shard is missing:

```
Oct 12 21:00:05.465 WARN pool size for shard eth_chain is 0, ignoring this shard, shard: eth_chain
...
thread 'main' panicked at node/src/store_builder.rs:203:14:
Creating the BlockStore works: InternalError("there is no pool for shard eth_chain")
```

### Expected Behavior

If a shard is configured with `pool_size = 0`, the Graph Node should:

- Log a warning indicating the shard is ignored (which currently happens).
- Successfully complete startup and operate without requiring or attempting to establish a connection pool for that shard, especially during components like BlockStore initialization.

This behavior would allow us to list all potential shards in a `config.toml` (even though we do not understand why we need to do this) but only consume Postgres connections for the shards actively used by the specific node instance.

### Steps to Reproduce

- Configure a config.toml with the standard primary shard and at least two chain-specific shards (e.g., optimism and eth_chain).
- Set `pool_size = 0` for one of the chain-specific shards (e.g., eth_chain).
- Start Graph Node using this configuration.

### Environment

Graph Node Version: 0.41.0 (from the logs)
Docker
PostgreSQL Version: 18

### Relevant Configuration Snippet

```
# Main node shard, used by this indexer
[store.primary]
connection = "..."
pool_size = 10

# Chain shard for this indexer
[store.optimism]
connection = "..."
pool_size = 10

# Chain shard for an *other* indexer node, but included here to satisfy configuration discovery
[store.eth_chain]
connection = "..." # This connection URL doesn't even need to be valid/reachable, as we expect it to be ignored
pool_size = 0 
```

### Relevant log output

```Shell
Oct 12 21:00:05.465 WARN pool size for shard eth_chain is 0, ignoring this shard, shard: eth_chain
...
thread 'main' panicked at node/src/store_builder.rs:203:14:
Creating the BlockStore works: InternalError("there is no pool for shard eth_chain")
```

### IPFS hash

_No response_

_No response_

### OS information

Linux

cargopete:
## Summary

Fixes #6195. A node configured with `pool_size = 0` for an unused shard (a
documented option from #3513) panics on startup:

```
WARN pool size for shard eth_chain is 0, ignoring this shard
thread 'main' panicked at node/src/store_builder.rs:203:
Creating the BlockStore works: InternalError("there is no pool for shard eth_chain")
```

`store_builder` correctly drops `pool_size = 0` shards from the pool map, but
`DieselBlockStore::new` then iterates **every** chain — both configured chains
and all chains present in the database — and calls `add_chain_store` for each.
`add_chain_store` errors when a chain's shard has no pool, and that error is
`.expect()`ed in `store_builder`, killing startup. So `pool_size = 0` is only
half-implemented: the pool is skipped, but the chains living in that shard are
not.

## Fix

In both chain-enumeration loops in `DieselBlockStore::new`, skip chains whose
shard has no connection pool (logging a `WARN`) instead of calling
`add_chain_store`. This lets a node list all shards in `config.toml` but only
consume Postgres connections for the shards it actually uses, which is the
intent of `pool_size = 0`.

`add_chain_store`'s contract is unchanged — it still errors when *creating* a
new chain in a shard with no pool, which remains a genuine misconfiguration.

## Testing

- `cargo fmt`, `cargo clippy --all-targets`, and `cargo check --release` clean
  for `graph-store-postgres`.

I was not able to add an automated test: `DieselBlockStore::new` and
`ConnectionPool` require a live multi-shard Postgres, so exercising the
multi-shard startup path isn't possible in a unit test. The change is small and
verified by inspection, but I'd appreciate a maintainer (or @0x19dG87) confirming
startup against a real multi-shard config with `pool_size = 0` set on an unused
shard.

### DIFF
--- a/store/postgres/src/block_store.rs
+++ b/store/postgres/src/block_store.rs
@@ -253,7 +253,27 @@ impl BlockStore {
         const CHAIN_HEAD_CACHE_TTL: Duration = Duration::from_secs(2);
 
         let mirror = PrimaryMirror::new(&pools);
-        let existing_chains = mirror.read_async(primary::load_chains).await?;
+        // Treat a chain whose shard has no connection pool (i.e. a shard
+        // configured with `pool_size = 0`, which `store_builder` drops) as if it
+        // doesn't exist: skip it here so we neither create a chain store for it
+        // nor fail startup. See issue #6195.
+        let existing_chains: Vec<_> = mirror
+            .read_async(primary::load_chains)
+            .await?
+            .into_iter()
+            .filter(|chain| {
+                let has_pool = pools.contains_key(&chain.shard);
+                if !has_pool {
+                    warn!(
+                        &logger,
+                        "Ignoring chain `{}`: its shard `{}` has no connection pool (pool_size = 0)",
+                        chain.name,
+                        chain.shard,
+                    );
+                }
+                has_pool
+            })
+            .collect();
         let chain_head_cache = TimedCache::new(CHAIN_HEAD_CACHE_TTL);
         let chains = shards.clone();
 
@@ -275,18 +295,6 @@ impl BlockStore {
                 .iter()
                 .find(|chain| chain.name == chain_name)
             {
-                // A shard configured with `pool_size = 0` is intentionally
-                // ignored and has no connection pool. Skip chains that live in
-                // such a shard rather than failing startup. See issue #6195.
-                if !block_store.pools.contains_key(&chain.shard) {
-                    warn!(
-                        &block_store.logger,
-                        "Skipping chain `{}`: its shard `{}` has no connection pool (pool_size = 0)",
-                        chain.name,
-                        chain.shard,
-                    );
-                    continue;
-                }
                 if chain.shard != shard {
                     warn!(
                         &block_store.logger,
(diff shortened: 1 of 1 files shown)


store/postgres/src/block_store.rs
```diff
@@ -275,6 +275,18 @@ impl BlockStore {
...
+                // such a shard rather than failing startup. See issue #6195.
+                if !block_store.pools.contains_key(&chain.shard) {
+                    warn!(
+                        &block_store.logger,
+                        "Skipping chain `{}`: its shard `{}` has no connection pool (pool_size = 0)",
+                        chain.name,
+                        chain.shard,
+                    );
+                    continue;
+                }
```
lutter: It would be better to do this filtering when we construct `existing_chains` rather than every time we iterate over it - essentially, this logic treats a chain with `pool_size == 0` as not exsiting/configured.

lutter (changes requested):
Nice, just one small cleanup

cargopete:
Thanks @lutter - good call. I've moved the filtering to where `existing_chains` is constructed, so a chain whose shard has no pool (`pool_size = 0`) is dropped once and treated as not existing, rather than being skipped on every iteration. Both chain-enumeration loops are back to their original form.

lutter (approved):

PR #6660, opened 2026-06-30
gnd: Fix `add` and unify subgraph scaffolding generators

incrypto32:
`gnd add` was a drifted copy of the `init`/scaffold code generation, and the copy never wrote `schema.graphql` — so adding a data source produced a subgraph that fails `codegen`/`build`. This unifies the two paths and fixes a cluster of related codegen bugs.

The scaffolder now decides every event's names once (a `ResolvedEvent`) and feeds one set of generators (schema, mapping, manifest), instead of `init` and `add` each rolling their own.

What's fixed:
- `add` now writes the event entities to `schema.graphql` (the main bug)
- Event-name collisions across data sources: renamed with the contract prefix by default, or merged into the existing entity with `--merge-entities` (previously a dead flag), keeping the handler so the new contract's events are still indexed
- Overloaded events (same name, different params) are disambiguated (`Transfer`, `Transfer1`) in both `add` and `init`
- Small integer types (`uint8`, `int32`, …) map to GraphQL `Int` instead of `BigInt`
- Reserved-word parameter names (`new`, `class`, …) are escaped so the generated code compiles
- `add` errors on a duplicate data-source name, and only updates `networks.json` when it exists

Cleanup:
- One field-name sanitizer (was three), one mapping generator (was two), shared version constants

Notes:
- Kept gnd's existing field-name conventions (camelCase, `eventId`) rather than graph-cli's (`internal_id`) — deliberate.
- Tuple/struct params and a codegen-time entity-vs-schema check are intentionally left for follow-up PRs.

### DIFF
--- a/gnd/src/scaffold/schema.rs
+++ b/gnd/src/scaffold/schema.rs
@@ -2,6 +2,7 @@
 
 use super::ScaffoldOptions;
 use super::manifest::{EventInput, extract_events_from_abi};
+use super::sanitize_field_name;
 
 /// Generate the schema.graphql content.
 pub fn generate_schema(options: &ScaffoldOptions) -> String {
@@ -55,7 +56,7 @@ fn generate_example_entity(inputs: &[EventInput]) -> String {
 }
 
 /// Generate an entity type for an event.
-fn generate_event_entity(event_name: &str, inputs: &[EventInput]) -> String {
+pub fn generate_event_entity(entity_name: &str, inputs: &[EventInput]) -> String {
     let mut fields = String::new();
 
     // ID field
@@ -76,7 +77,7 @@ fn generate_event_entity(event_name: &str, inputs: &[EventInput]) -> String {
     format!(
         "# Declare entity types as immutable when possible for better performance\n\
          type {} @entity(immutable: true) {{\n{}\n}}",
-        event_name, fields
+        entity_name, fields
     )
 }
 
@@ -117,47 +118,6 @@ fn solidity_to_graphql(solidity_type: &str) -> &'static str {
 
         // Default to Bytes for unknown types
         _ => "Bytes",
-    }
-}
-
-/// Sanitize a field name to be a valid GraphQL identifier.
-fn sanitize_field_name(name: &str) -> String {
-    if name.is_empty() {
-        return "value".to_string();
-    }
-
-    // GraphQL field names must start with a letter or underscore
-    let mut result = String::new();
-
-    for (i, c) in name.chars().enumerate() {
-        if i == 0 && c.is_ascii_digit() {
-            result.push('_');
-        }
-        if c.is_alphanumeric() || c == '_' {
-            result.push(c);
-        } else {
-            result.push('_');
-        }
-    }
-
-    // Convert to camelCase if starts with uppercase
-    if result
-        .chars()
-        .next()
-        .map(|c| c.is_uppercase())
-        .unwrap_or(false)
-    {
-        let mut chars = result.chars();
-        if let Some(first) = chars.next() {
-            result = first.to_lowercase().collect::<String>() + chars.as_str();
-        }
-    }
-
-    // Avoid reserved words
-    match result.as_str() {
-        "id" => "eventId".to_string(),
-        "type" => "eventType".to_string(),
-        _ => result,
     }
 }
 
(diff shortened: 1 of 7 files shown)


gnd/src/scaffold/schema.rs
```diff
@@ -113,14 +114,37 @@ fn solidity_to_graphql(solidity_type: &str) -> &'static str {
         | "bytes23" | "bytes24" | "bytes25" | "bytes26" | "bytes27" | "bytes28" | "bytes29"
         | "bytes30" | "bytes31" | "bytes32" => "Bytes",
 
-        // Integer types - all map to BigInt for simplicity
-        t if t.starts_with("uint") || t.starts_with("int") => "BigInt",
+        // Integers: small widths fit in an i32 (GraphQL Int), the rest need BigInt.
+        t if t.starts_with("uint") || t.starts_with("int") => int_to_graphql(t),
```
lutter, line 118: Should we do something similar for `Int8`?

lutter (approved):

lutter:
Part of me thinks we should not do the complicated renaming on collision, but I think it would force users to not use `gnd add`, and adding parameters to `gnd add` that explicitly changes the name (like `gnd add --rename Token:MyToken`) seems a bit clunky.

Here's also what Claude had to say, I think it caught a few corner cases of the renaming machinery; I think they are all pretty esoteric corner cases, not sure if it's worth addressing them.

> What it does: gnd add was a drifted copy of the init scaffolding logic that never wrote event entities to schema.graphql, so adding a data source produced a subgraph that failed codegen/build. This PR introduces a single ResolvedEvent model — names (handler alias, entity name, whether to declare in schema) are decided once and fed to one set of schema/mapping/manifest generators. It also fixes overloaded-event disambiguation, cross-datasource entity-name collisions (rename or --merge-entities), small-int → GraphQL Int mapping, reserved-word escaping, and duplicate-datasource-name detection. It's well-structured and well-tested; the accessor generation correctly mirrors the ABI codegen's disambiguate_event_params.
>
> The findings below are edge cases that survived verification — all produce the exact "invalid schema → build fails" outcome the PR is trying to eliminate.
>
> Findings
>
> 1. gnd/src/commands/add.rs (resolve_events, ~line 303) — a renamed colliding entity isn't checked against sibling events in the same batch, producing a duplicate @entity type. (CONFIRMED)
When an event collides with an existing entity and --merge-entities is off, it's renamed to <contract_name><alias>, but that name is only checked against the pre-existing existing set — never against the other events being added. Concrete: existing schema declares Transfer; the new ABI for contract Token has events Transfer and TokenTransfer. Transfer collides → renamed to TokenTransfer; the ABI's own TokenTransfer event keeps entity_name = TokenTransfer. Both ResolvedEvents end up with entity_name = TokenTransfer and declare_in_schema = true, so add_schema_entities emits type TokenTransfer @entity twice → invalid schema, codegen/build fails. A HashSet of already-assigned entity names threaded through the loop (and inserted as each name is finalized) would close this.
>
> 2. gnd/src/scaffold/schema.rs (generate_event_entity, ~line 65) & mapping.rs (generate_single_handler) — params that sanitize to the same identifier produce a duplicate entity field. sanitize_field_name has no de-duplication (unlike the new event_param_accessors, which suffixes collisions). Two unnamed params both sanitize to value; Value and value both camelCase to value. The schema then declares value: T! twice (invalid GraphQL) and the mapping emits two entity.value = … assignments. Pre-existing behavior, but it lives in the touched, now-unified generators and defeats the PR's goal; the accessor side was fixed while the field-name side was not.
>
> 3. gnd/src/scaffold/manifest.rs (disambiguate_events, ~line 780) — the overload suffix can collide with a real event whose name already carries that suffix. An ABI with events Transfer, Transfer, Transfer1 yields aliases Transfer, Transfer1, Transfer1 (the counter for Transfer doesn't know Transfer1 also exists as a distinct name). Result: duplicate handleTransfer1 and duplicate Transfer1 entity. Rare, but it's the same broken-build failure mode and now runs on both init and add.
>
> 4. gnd/src/commands/add.rs (schema_entity_types / entity_types_in_schema, ~line 226) — a schema that fails to parse silently drops collision detection. If graphql_tools::parser::schema can't parse the subgraph's schema.graphql, an empty set is returned and the code falls back to manifest entities lists only. A hand-written type Foo @entity not listed in any mapping's entities would then be missed, and add_schema_entities re-declares it → duplicate type. Graceful-ish (manifest-listed entities are still caught), so lowest severity, but worth a log line rather than a silent empty fallback.
> 
> Findings 1 and 2 are the ones most worth acting on before merge; 3 and 4 are low-probability hardening.

gnd/src/scaffold/schema.rs
incrypto32: Its doable, but the codegen generated binding currently only ever return i32 or BigInt, so an Int8 field would mismatch the getter on the mapping side. ABI codegen needs to change too. Will do it in a follow-up, the width cutoff is duplicated in a few places now and I'd rather consolidate it there.

incrypto32:
Agree these are rare corner cases. Noted them all down, will fix them in the same follow-up as the Int8 change since it touches the same files. The rename collision and the field name dedup are the most likely to actually happen, the other two need really weird ABIs.

PR #6687, opened 2026-07-16
store: Populate fulltext columns when copying a subgraph

lutter:
`CopyEntityBatchQuery::new` looked up source columns via `Table::column`, which deliberately hides `tsvector` (fulltext) columns. Because fulltext columns are nullable, they were silently dropped from the copy instead of hitting the non-nullable error path, so the destination's fulltext search column and its GIN index were left empty after a copy/graft.

Add `Table::column_including_fulltext` and use it in the copy so fulltext columns are matched and their already-computed tsvector value is copied verbatim; document that `Table::column` skips fulltext search columns. Add a regression test that copies the `User` table and verifies a fulltext query against the copy returns the expected entity.

### DIFF
--- a/store/postgres/src/relational_queries.rs
+++ b/store/postgres/src/relational_queries.rs
@@ -5104,7 +5104,13 @@ impl<'a> CopyEntityBatchQuery<'a> {
     ) -> Result<Self, StoreError> {
         let mut columns = Vec::new();
         for dcol in &dst.columns {
-            if let Some(scol) = src.column(&dcol.name) {
+            // Match against all source columns, INCLUDING fulltext (tsvector)
+            // columns, which `Table::column` deliberately hides. The stored
+            // tsvector is copied verbatim; without this, fulltext columns
+            // would be silently dropped from the copy (they are nullable, so
+            // the error branch below is not taken) and the destination's
+            // search index would be left empty.
+            if let Some(scol) = src.column_including_fulltext(&dcol.name) {
                 if let Some(msg) = dcol.is_assignable_from(scol, &src.object) {
                     return Err(anyhow!("{}", msg).into());
                 } else {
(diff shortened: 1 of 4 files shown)


store/postgres/src/relational_queries.rs
```diff
@@ -5104,7 +5104,13 @@ impl<'a> CopyEntityBatchQuery<'a> {
...
         let mut columns = Vec::new();
         for dcol in &dst.columns {
-            if let Some(scol) = src.column(&dcol.name) {
+            // Match against all source columns, INCLUDING fulltext (tsvector)
+            // columns, which `Table::column` deliberately hides. The stored
+            // tsvector is copied verbatim; without this, fulltext columns
+            // would be silently dropped from the copy (they are nullable, so
+            // the error branch below is not taken) and the destination's
+            // search index would be left empty.
+            if let Some(scol) = src.column_including_fulltext(&dcol.name) {
```
incrypto32, line 5113: Should `can_copy_from` also use  `column_including_fulltext` it uses `column()` currently?

store/postgres/src/relational_queries.rs
```diff
@@ -5104,7 +5104,13 @@ impl<'a> CopyEntityBatchQuery<'a> {
...
                 if let Some(msg) = dcol.is_assignable_from(scol, &src.object) {
```
incrypto32, line 5114: One thing that claude flagged during the review:

> `is_assignable_from `only compares FulltextConfig, which is language and algorithm, not fulltext_fields. So a graft that adds a field to include: copies the old tsvectors and searches on the new field quietly return nothing. Compare `fulltext_fields` here too?

incrypto32 (approved):
Apart from the comments, LGTM

